PR #24808, opened 2026-01-06
Ensure all immutable claims are complete and accurate

arun-koshy:
## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/consensus_validator.rs
+++ b/crates/sui-core/src/consensus_validator.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeSet, HashMap, HashSet},
     sync::Arc,
 };
 
@@ -316,7 +316,7 @@ impl SuiTxValidator {
             .map(|obj_ref| (obj_ref.0, *obj_ref))
             .collect();
 
-        // First check: all claimed objects must be among the input objects
+        // First check: all claimed object IDs must be among the input object IDs
         for claimed_id in claimed_ids {
             if !input_refs_by_id.contains_key(claimed_id) {
                 return Err(SuiErrorKind::ImmutableObjectClaimNotFoundInInput {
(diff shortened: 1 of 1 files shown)


crates/sui-core/src/consensus_validator.rs
```diff
@@ -299,48 +299,52 @@ impl SuiTxValidator {
...
-                    // Object exists but is NOT immutable - invalid claim
-                    let object_ref = o.compute_object_reference();
-                    if !owned_object_refs.contains(&o.compute_object_reference()) {
-                        return Err(SuiErrorKind::ImmutableObjectClaimNotFoundInInput {
-                            object_id: *id,
-                        }
-                        .into());
-                    }
-                    if !o.is_immutable() {
+                    let is_immutable = o.is_immutable();
```
mwtian: I don't think we should skip the comparison between o.compute_object_reference() and the input object ref.

crates/sui-core/src/consensus_validator.rs
```diff
@@ -299,48 +299,52 @@ impl SuiTxValidator {
...
+            owned_object_refs.iter().map(|(id, _, _)| *id).collect();
 
         let objects = self
             .authority_state
             .get_object_cache_reader()
-            .get_objects(claimed_ids);
+            .get_objects(&input_object_ids);
 
-        for (obj, id) in objects.into_iter().zip(claimed_ids.iter()) {
+        for (obj, obj_ref) in objects.into_iter().zip(owned_object_refs.iter()) {
```
mwtian: How about just collect the set of immutable objects based on owned_object_refs, checking for existence / version / digest. Then compare the computed list with the claimed_ids to ensure exact match?

crates/sui-core/src/consensus_validator.rs
```diff
@@ -349,6 +353,18 @@ impl SuiTxValidator {
...
+        // since all claimed objects must be in owned_object_refs.
+        // This loop only ensures no claimed objects are missing from inputs.
+        for claimed_id in claimed_ids {
+            if !input_object_ids.contains(claimed_id) {
+                return Err(SuiErrorKind::ImmutableObjectClaimNotFoundInInput {
+                    object_id: *claimed_id,
+                }
+                .into());
+            }
+        }
```
akichidis: nit: I think it would be cleaner to have that first as it's also kind of an important check to do

crates/sui-core/src/consensus_validator.rs
```diff
@@ -299,56 +302,94 @@ impl SuiTxValidator {
...
                         },
                     }
                     .into());
                 }
             }
         }
 
+        // Compare claimed_ids with actual immutable objects - must match exactly
+        for claimed_id in claimed_ids {
+            if !actual_immutable_ids.contains(claimed_id) {
```
mystenmark: this is quadratic - can we just make two sorted vectors and do a strict equality check instead?

mwtian (approved):
I did some cleanups. Ptal at the last commit.

PR #24823, opened 2026-01-07
fix(indexer-alt-framework): Fix client_args not being parsed from CLI

tpham-mysten:
## Description 

Fixes a bug where `client_args` was None when parsed from CLI arguments, causing Error: `client_args is required` even when `--remote-store-url` was provided.

Root cause: https://github.com/clap-rs/clap/issues/5809 - when using `#[clap(flatten)]` with `Option<T>` and the inner type contains nested #[clap(flatten)], the parsed values are discarded and the Option stays None.

After #24261 added nested flattening inside `ClientArgs` (for `IngestionClientArgs` and `StreamingClientArgs`), the clap bug was triggered.

Fix: Remove the Option wrapper from client_args in the Args struct. The #[group(required = true)] on IngestionClientArgs already enforces that one source option is provided.

## Test plan 

Verified the basic-sui-indexer example now runs successfully with 

`cargo run -- --remote-store-url https://checkpoints.testnet.sui.io`

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/cluster.rs
+++ b/crates/sui-indexer-alt-framework/src/cluster.rs
@@ -31,7 +31,7 @@ pub struct Args {
 
     /// Where to get checkpoint data from.
     #[clap(flatten)]
-    pub client_args: Option<ClientArgs>,
+    pub client_args: ClientArgs,
 
     /// How to expose metrics.
     #[clap(flatten)]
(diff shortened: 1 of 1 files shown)


tpham-mysten:
Thanks for the recall. You're right that `ClientArgs` now derives default, so we don't need Option<ClientArgs> hack here anymore.

PR #24824, opened 2026-01-07
Do not cancel effects wait on epoch change

mwtian:
## Description 

Waiting on effects for a transaction digest should not need to be cancelled on epoch change in `TransactionOrchestrator`.

Also, increase epoch length from 3s to 5s in `test_simulated_load_with_reconfig`.

## Test plan 

simtest

### DIFF
--- a/scripts/simtest/seed-search.py
+++ b/scripts/simtest/seed-search.py
@@ -22,6 +22,7 @@ def safe_print(*args, **kwargs):
 parser = argparse.ArgumentParser(description='Run the simulator with different seeds')
 parser.add_argument('testname', type=str, help='Name of test to run')
 parser.add_argument('--test', type=str, help='Name of the test binary run', required=True)
+parser.add_argument('--no-capture', action='store_true', help='Whether to output logs as the test runs', default=False)
 parser.add_argument('--exact', action='store_true', help='Use exact matching for test name', default=False)
 parser.add_argument('--num-seeds', type=int, help='Number of seeds to run', default=200)
 parser.add_argument(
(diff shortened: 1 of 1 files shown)


scripts/simtest/seed-search.py
```diff
@@ -105,9 +105,10 @@ def main(commands):
 
     for i in range(1, args.num_seeds + 1):
         next_seed = args.seed_start + i
-        commands.append(("%s --test-threads 1 %s %s" % (binary, '--exact' if args.exact else '', args.testname), {
+        commands.append(("%s --no-capture --test-threads 1 %s %s" % (binary, '--exact' if args.exact else '', args.testname), {
```
mystenmark: can you add an option for this instead of making it the default?

crates/sui-core/src/transaction_orchestrator.rs
```diff
@@ -542,14 +542,12 @@ where
...
-                self.validator_state
-                    .get_transaction_cache_reader()
-                    .notify_read_executed_effects(
-                    "TransactionOrchestrator::notify_read_execute_transaction_with_effects_waiting",
-                    &digests,
-                ),
+        let mut local_effects_future = self
+            .validator_state
+            .get_transaction_cache_reader()
+            .notify_read_executed_effects(
```
mystenmark, line 548: this will make the `digests must exist` error more likely to occur, not less. But since we've disabled pruning on fullnodes in most tests we could try this. But please keep an eye on simtest nightly to see if this makes things worse

crates/sui-benchmark/tests/simtest.rs
```diff
@@ -96,7 +96,7 @@ mod test {
     #[sim_test(config = "test_config()")]
     async fn test_simulated_load_with_reconfig() {
         sui_protocol_config::ProtocolConfig::poison_get_for_min_version();
-        let test_cluster = build_test_cluster(2, 3000, 1).await;
+        let test_cluster = build_test_cluster(2, 5_000, 1).await;
```
mystenmark, line 99: I've been using 10s for most tests but if 5s is empirically stable we can do that.

mwtian: It seems to work and it also used by a few other tests.

crates/sui-core/src/transaction_orchestrator.rs
mwtian: Hopefully that bug should be fixed separately.

scripts/simtest/seed-search.py
mwtian: Sure.

PR #24829, opened 2026-01-07
[tree-sitter] Fix `spec` variable-name parsing. Update kws

### LINKED ISSUE #24783: Prettier plugin fails on variables named 'spec'
## Bug
`@mysten/prettier-plugin-move@0.3.3` crashes when code contains a variable named `spec`.

## Error
```
Error: tree-sitter failure on (spec)
```

## Repro
```move
public fun example(spec: &ActionSpec) {
    call(spec);  // crashes here
}
```

The parser sees `(spec)` and thinks it's a Move specification block (`spec { ... }`) rather than a function argument.

## Expected
`spec` should be valid as a variable/parameter name.

tzakian:
## Description 

Treesitter was failing to parse `spec` as a variable name since it's an (older) contextual keyword. This fixes this parsing issue.

This also removes `spec` as a keyword.

This fixes #24783 

## Test plan 

Added a repro test case and verified the fix against it. 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/tooling/tree-sitter/src/grammar.json
```diff
@@ -7080,6 +7084,10 @@
     ],
     [
       "abort_expression"
+    ],
+    [
+      "spec_block",
```
damirka, line 7089: Should this one be in there?

damirka (approved):
Thanks! Though I don't understand why it got saved in a different file.

external-crates/move/tooling/tree-sitter/src/grammar.json
tzakian: yep... we need it so treesitter knows how to handle things when it sees a `spec` and how to backtrack properly

PR #24832, opened 2026-01-07
Tweaks to checkpoint summary handling in state sync

aschran:
## Description 

Adds max lookahead & batching options

## Test plan 

Unit tests updated

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-config/src/p2p.rs
+++ b/crates/sui-config/src/p2p.rs
@@ -313,6 +313,34 @@ impl StateSyncConfig {
         self.max_checkpoint_summary_size
             .unwrap_or(DEFAULT_MAX_CHECKPOINT_SUMMARY_SIZE)
     }
+
+    /// Returns a StateSyncConfig with randomized settings for testing.
+    pub fn randomized_for_testing() -> Self {
+        use rand::Rng;
+        let mut rng = rand::thread_rng();
+        let config = Self {
+            mailbox_capacity: Some(rng.gen_range(16..=2048)),
+            synced_checkpoint_broadcast_channel_capacity: Some(rng.gen_range(16..=2048)),
+            checkpoint_header_download_concurrency: Some(rng.gen_range(10..=500)),
+            checkpoint_content_download_concurrency: Some(rng.gen_range(10..=500)),
+            max_checkpoint_lookahead: Some(rng.gen_range(100..=2000)),
+            // Tests sync up to 100 checkpoints, so minimum must be >= 100
+            max_checkpoint_sync_batch_size: Some(rng.gen_range(100..=500)),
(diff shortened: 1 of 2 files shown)


mystenmark (commented):
I think the settings here might benefit from randomization during testing, like how we do `randomize_cache_capacity_in_tests`. But we probably want a new function that just uses a uniform value rather than the logic of `randomize_cache_capacity_in_tests`

crates/sui-network/src/state_sync/builder.rs
```diff
@@ -116,7 +116,14 @@ where
                 .add_layer_for_get_checkpoint_contents(InboundRequestLayer::new(layer));
         }
 
-        (builder, state_sync_server)
+        let router = anemo::Router::new()
+            // Size limit layer applied before deserialization.
+            .route_layer(SizeLimitLayer::new(
+                state_sync_config.max_checkpoint_summary_size(),
+            ))
```
bmwill, line 123: Specifically this only applies to limiting the size of *incoming* requests, of which only CheckpointSummaries are *pushed* to other nodes, so this effectively limits only that and doesn't impact the fetching of checkpoint contents which can and will be larger than this set size.

crates/sui-network/src/state_sync/mod.rs
```diff
@@ -226,17 +220,57 @@ impl PeerHeights {
...
     pub fn cleanup_old_checkpoints(&mut self, sequence_number: CheckpointSequenceNumber) {
         self.unprocessed_checkpoints
             .retain(|_digest, checkpoint| *checkpoint.sequence_number() > sequence_number);
         self.sequence_number_to_digest
             .retain(|&s, _digest| s > sequence_number);
     }
 
+    /// Inserts a checkpoint into the unprocessed checkpoints store.
+    /// Only one checkpoint per sequence number is stored. If a checkpoint with the same
+    /// sequence number but different digest already exists, the new one is dropped.
```
bmwill, line 253: If we fail to validate the summary we already have do we drop it and allow the service to attempt to validate a different one still?

aschran: Yes we do, that happens in the checkpoint sync task

crates/sui-network/src/state_sync/builder.rs
aschran: Right, this one is treated specially bc it's the only one pushed to us without a request and potentially stored without immediate validation

PR #24840, opened 2026-01-08
chore(imports): Standardize imports among indexing and RPC crates.

amnn:
## Description 

Standardize imports according to the following rules:

- Each import belongs on its own line.
- When a trait is imported, it is imported `as _` unless it needs to be referred to by name.
- Split imports into three blocks, `std`, external crates, and `crate::`
- Do not use `this::` or `super::`, except in the case of inline test modules which are allowed to `use super::*`.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-rpc-resolver/src/json_visitor.rs
+++ b/crates/sui-rpc-resolver/src/json_visitor.rs
@@ -9,17 +9,19 @@
 //! collections. It should not be used in memory-constrained contexts like RPC
 //! handlers where the size of the data is unbounded.
 
-use move_core_types::{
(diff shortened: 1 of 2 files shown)


henryachen (approved):
Got some merge conflicts but looks good!

wlmyng (approved):
holy moly

PR #24841, opened 2026-01-08
chore(deps): remove use of unmaintained rustls-pemfile

amnn:
## Description 

As in title.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/balances.move
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/balances.move
@@ -3,8 +3,7 @@
 
 // Simple e2e test of coin, address, and total balance queries.
 
-//# init --accounts A B --addresses T=0x0 --simulator --enable-accumulators
-
+//# init --protocol-version 106 --accounts A B --addresses T=0x0 --simulator --enable-accumulators
 
 //# publish --sender A
 #[allow(deprecated_usage)]
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/balances.snap
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/balances.snap
@@ -6,13 +6,13 @@ processed 6 tasks
 init:
 A: object(0,0), B: object(0,1)
 
-task 1, lines 9-43:
+task 1, lines 8-42:
 //# publish --sender A
 created: object(1,0), object(1,1), object(1,2)
 mutated: object(0,0)
 gas summary: computation_cost: 1000000, storage_cost: 12144800,  storage_rebate: 0, non_refundable_storage_fee: 0
 
-task 2, line 45:
+task 2, line 44:
 //# view-object 1,1
 Owner: Account Address ( A )
 Version: 2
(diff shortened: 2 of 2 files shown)


bmwill (approved):

crates/sui-config/src/local_ip_utils.rs
amnn, line 1: @bmwill, I ended up having to make this change because I needed to update `hyper-rustls` past `0.27.5` for it to drop its dependency on `rustls-pemfile`, and that pulled in this fix:

https://github.com/rustls/hyper-rustls/pull/295

I believe the issue is that the client configs specify `https_only` and previously that was being ignored, but now it is not. My fix lacks finesse -- I just kept changing `http -> https` in multiaddrs used for tests, but I know that in real set-ups, we often just drop the protocol part altogether, so I wanted to make sure this seemed right, and check with you whether this new enforcement of `https_only` could cause issues in real deployments (in which case we need to loosen the `https_only` requirement).

crates/sui-indexer-alt-e2e-tests/tests/graphql/epochs/validator_set/active_validators.snap
```diff
@@ -34,10 +34,10 @@ Response: {
                     "description": "",
                     "image_url": "",
                     "project_url": "",
-                    "net_address": "/ip4/127.0.0.1/tcp/8000/http",
-                    "p2p_address": "/ip4/127.0.0.1/udp/8001/http",
-                    "primary_address": "/ip4/127.0.0.1/udp/8004/http",
-                    "worker_address": "/ip4/127.0.0.1/udp/8005/http",
+                    "net_address": "/ip4/127.0.0.1/tcp/8000/https",
```
amnn, line 37: This change is what has caused all the IDs and digests to change in the transactional tests.

crates/sui-config/src/local_ip_utils.rs
bmwill: Yeah we rewrite the urls to https here: https://github.com/mystenlabs/sui/blob/658ff31cf95c78106d6760b1a3791314b6e593ae/crates/sui-core/src/authority_client.rs#L257-L261

bmwill: This was a larger change by @aschran to ensure that when we send traffic to the validators that we establish a secure link with them. The fact that you had to change this may point to us not properly re-writing those urls everywhere, if we are missing places then we need to fix those as well so this should highlight those places. 

I think unconditionally changing this to always use https may address that https://github.com/mystenlabs/sui/blob/658ff31cf95c78106d6760b1a3791314b6e593ae/crates/mysten-network/src/client.rs#L64-L87

I did a quick look around and we do seem to always require a TLS config to be passed in so we're generally expecting to use https over http.

amnn: Ah okay great -- then if it looks good still, could I get a re-accept?

bmwill (approved):

PR #24844, opened 2026-01-08
[cli] Add a no-tree-shaking flag to sui move build command

stefan-mysten:
## Description 

This PR adds the `--no-tree-shaking flag` to `sui move build` command when used in connection with `dump-bytecode-as-base64`. It will avoid removing unused dependencies for which it requires an RPC connection, thus it can be used offline.

## Test plan 

Existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [x] CLI: Added `--no-tree-shaking` flag that can only be used with `--dump-bytecode-as-base64`. This will ensure that all dependencies will be kept in the list of dependencies in the json output, regardless if they're used or not used in the source code. In contrast, by default, the CLI will remove any unused dependencies from the dependency list on publication/upgrade and when `--no-tree-shaking` flag is false.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-move/src/build.rs
+++ b/crates/sui-move/src/build.rs
@@ -17,24 +17,24 @@ const STRUCT_LAYOUTS_FILENAME: &str = "struct_layouts.yaml";
 pub struct Build {
     /// Include the contents of packages in dependencies that haven't been published (only relevant
     /// when dumping bytecode as base64)
-    #[clap(short = 'u', long, global = true)]
+    #[clap(long, global = true)]
     pub with_unpublished_dependencies: bool,
     /// Dump the compiled bytecode as base64-encoded strings in a JSON object, together with the
     /// digest and list of dependencies.
-    #[clap(short = 'd', long, global = true)]
+    #[clap(long, global = true)]
     pub dump_bytecode_as_base64: bool,
     /// By default, the CLI will drop any unused dependencies from the output and makes calls to
     /// the RPC. This flag disables that behavior and avoids any RPC calls, keeping all
     /// dependencies in the output (the dependency list). This is useful for offline compilation.
     /// Only to be used with `--dump-bytecode-as-base64`.
-    #[clap(short = 't', long, global = true, requires = "dump_bytecode_as_base64")]
+    #[clap(long, global = true, requires = "dump_bytecode_as_base64")]
     pub no_tree_shaking: bool,
     /// If true, generate struct layout schemas for
     /// all struct types passed into `entry` functions declared by modules in this package
     /// These layout schemas can be consumed by clients (e.g.,
     /// the TypeScript SDK) to enable serialization/deserialization of transaction arguments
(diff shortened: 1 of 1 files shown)


stefan-mysten:
> Are the added shortcuts generally safe here? 👀

For this particular command, yes. Maybe it's better to remove them, not sure.

```
Usage: sui move build [OPTIONS]

Options:
  -p, --path <PACKAGE_PATH>                     Path to a package which the command should be run with respect to
  -u, --with-unpublished-dependencies           Include the contents of packages in dependencies that haven't been published (only relevant when dumping bytecode as base64)
  -d, --dump-bytecode-as-base64                 Dump the compiled bytecode as base64-encoded strings in a JSON object, together with the digest and list of dependencies
  -t, --no-tree-shaking                         By default, the CLI will drop any unused dependencies from the output and makes calls to the RPC. This flag disables that behavior and avoids any RPC calls,
                                                keeping all dependencies in the output (the dependency list). This is useful for offline compilation. Only to be used with `--dump-bytecode-as-base64`
  -s, --generate-struct-layouts                 If true, generate struct layout schemas for all struct types passed into `entry` functions declared by modules in this package These layout schemas can be
                                                consumed by clients (e.g., the TypeScript SDK) to enable serialization/deserialization of transaction arguments and events
      --test                                    Compile in 'test' mode and include the code in the 'tests' directory
      --doc                                     Generate documentation for packages
      --disassemble                             Save disassembly for generated bytecode along with bytecode maps (source maps for disassembled bytecode)
      --install-dir <INSTALL_DIR>               Installation directory for compiled artifacts. Defaults to current directory
      --force                                   Force recompilation of all packages
      --allow-dirty                             Allow building, even if some cached git dependencies in `~/.move` are modified
      --default-move-flavor <DEFAULT_FLAVOR>    Default flavor for move compilation, if not specified in the package's config
      --default-move-edition <DEFAULT_EDITION>  Default edition for move compilation, if not specified in the package's config
      --silence-warnings                        If set, ignore any compiler warnings
  -q, --quiet                                   Display less output
      --warnings-are-errors                     If set, warnings become errors
      --json-errors                             If set, reports errors at JSON
      --no-lint                                 If `true`, disable linters
      --lint                                    If `true`, enables extra linters
      --mode <MODE>                             Arbitrary mode -- this will be used to enable or filter user-defined `#[mode(<MODE>)]` annotations during compilation
  -e, --environment <ENVIRONMENT>               Environment to use for building packages
  -h, --help                                    Print help
  -V, --version                                 Print version
```

manolisliolios:
> > Are the added shortcuts generally safe here? 👀
> 
> For this particular command, yes. Maybe it's better to remove them, not sure.

Probably better to skip those and decide later tbh but up to you

PR #24851, opened 2026-01-08
simulate: add support for address balances to select_gas

bmwill:
## Description 

This patch adds support for selecting gas from an account's address
balance if the GasCoin isn't used as an argument in the transaction and
if there is sufficient balance to pay for the required budget, otherwise
we fallback to normal coin selection.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -2485,6 +2485,7 @@ impl AuthorityState {
         &self,
         mut transaction: TransactionData,
         checks: TransactionChecks,
+        allow_mock_gas_coin: bool,
     ) -> SuiResult<SimulateTransactionResult> {
         if transaction.kind().is_system_tx() {
             return Err(SuiErrorKind::UnsupportedFeatureError {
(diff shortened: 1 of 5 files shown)


crates/sui-rpc-api/src/reader.rs
```diff
@@ -200,6 +200,29 @@ impl StateReader {
...
+
+        let key = AccumulatorKey { owner };
+        let key_type_tag = AccumulatorKey::get_type_tag(&[balance_type]);
+
+        DynamicFieldKey(SUI_ACCUMULATOR_ROOT_OBJECT_ID, key, key_type_tag)
+            .into_unbounded_id()
+            .unwrap()
+            .load_object(self.inner())
+            .and_then(|o| o.load_value::<u128>().ok())
+            .map(|balance| balance as u64)
```
mystenmark, line 224: feel free to add a version of `AccumulatorValue::load` that takes ObjectStore and move this code there. Up to you.

PR #24856, opened 2026-01-09
[consensus] improve direct finalization

mwtian:
## Description 

Do not directly finalize transactions when a block is below the GC round as computed from its commit leader.

Reduce GC round in antithesis to exercise GC more often in its tests.

## Test plan 

antithesis

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/commit_finalizer.rs
+++ b/consensus/core/src/commit_finalizer.rs
@@ -265,19 +265,24 @@ impl CommitFinalizer {
 
     // Tries directly finalizing transactions in the commit.
     // Direct commit means every transaction in the commit can be considered to have a quorum of post-commit certificates,
-    // unless the transaction has reject votes that do not reach quorum, or the block containing the transaction is outside
-    // of GC bound.
+    // unless (1) the transaction has reject votes that do not reach quorum, or
+    // (2) the block containing the transaction is outside the GC bound of the commit's leader.
+    // In the 2nd case, when the blocks voting and certifying this commit's leader were proposed, there is a chance
+    // that some of these voting and certifying blocks do not include votes for the transactions below the leader's GC bound.
+    // So conservatively, these transactions are not directly finalized. The logic here matches the GC logic in
+    // try_indirect_finalize_pending_transactions_in_block().
     fn try_direct_finalize_commit(&mut self, index: usize) {
         let metrics = &self.context.metrics.node_metrics;
         let num_commits = self.pending_commits.len();
         let commit_state = self
             .pending_commits
             .get_mut(index)
-            .unwrap_or_else(|| panic!("Commit {} does not exist. len = {}", index, num_commits,));
-        let gc_round = self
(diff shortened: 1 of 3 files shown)


consensus/core/src/commit_finalizer.rs
```diff
@@ -264,19 +264,43 @@ impl CommitFinalizer {
...
+            {
+                // The block is outside of GC bound.
+                let transactions =
+                    (0..(num_transactions as TransactionIndex)).collect::<BTreeSet<_>>();
+                commit_state
+                    .pending_transactions
+                    .entry(block_ref)
+                    .or_default()
+                    .extend(transactions);
+                continue;
```
akichidis, line 318: Can we add a metric to figure out how frequently we hit this case? Or even a debug log? It might be useful for future debug

consensus/core/src/commit_finalizer.rs
```diff
@@ -1284,6 +1308,115 @@ mod tests {
         assert!(fixture.commit_finalizer.is_empty());
     }
 
+    // Test direct finalization when some blocks are below GC bound.
```
akichidis: Can we make it clear that we are talking about the commit's leader gc round?

consensus/core/src/commit_finalizer.rs
```diff
@@ -264,19 +264,43 @@ impl CommitFinalizer {
     }
 
     // Tries directly finalizing transactions in the commit.
+    // Direct commit means every transaction in the commit can be considered to have a quorum of post-commit certificates,
+    // unless the transaction has reject votes that do not reach quorum, or the block containing the transaction is outside
+    // of GC bound.
```
akichidis: Maybe it would be good to explain what GC bound we consider here. As every committed sub dag will always respect the GC bounds (the one that has been set by the previous leader) it _might_ be a bit confusing when talking about GC bounds here. Also I think it would be good to explain why we can't/don't want to certify blocks that legitimately ended up in a commit but are bellow by the committed sub dag's leader gc round.

consensus/core/src/commit_finalizer.rs
```diff
@@ -264,19 +264,43 @@ impl CommitFinalizer {
...
         let pending_blocks = std::mem::take(&mut commit_state.pending_blocks);
+
         for (block_ref, num_transactions) in pending_blocks {
+            if self
+                .context
+                .protocol_config
+                .consensus_skip_gced_blocks_in_direct_finalization()
+                && block_ref.round <= gc_round
+                && num_transactions > 0
+            {
```
akichidis, line 298: Do I understand correctly that this means we could face a potential slow down in finalization if we do have commits that carry over blocks that are always below the leader's GC round? (and they do cary over transactions)

mwtian: Yes.

PR #24857, opened 2026-01-09
[bella ciao] Fix value `Rc` usage 

tnowacki:
## Description 

- Fix `Rc` value usage to never fail, and to not check the count. 

## Test plan 
- 🤷 
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
@@ -2600,7 +2600,15 @@ impl Value {
     }
 
     pub fn typed_serialize(&self, layout: &MoveTypeLayout) -> Option<Vec<u8>> {
-        Some(bcs::to_bytes(&AnnotatedValue { layout, val: self }).expect("BCS failed"))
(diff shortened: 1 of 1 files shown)


external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -2629,7 +2600,15 @@ impl Value {
     }
 
     pub fn typed_serialize(&self, layout: &MoveTypeLayout) -> Option<Vec<u8>> {
-        Some(bcs::to_bytes(&AnnotatedValue { layout, val: self }).expect("BCS failed"))
```
tnowacki, line 2632: I can revert this, but should this really be an `expect` here? Should this be a `Result<Option<...>>`?

tzakian: Good spot. This _should_ be an `ok()` here and that matches what is in `main`.

PR #24859, opened 2026-01-09
[pkg-system] Improve error message when trying to republish

manolisliolios:
## Description 

Better and more descriptive error msg when trying to republish an already published package.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/tests/shell_tests/with_network/ephemeral/ephemeral_publish_fail.snap
+++ b/crates/sui/tests/shell_tests/with_network/ephemeral/ephemeral_publish_fail.snap
@@ -30,7 +30,7 @@ sui client --client.config $CONFIG \
   test-publish --build-env testnet --pubfile-path Pub.local.toml a \
   > /dev/null || echo "failed to publish a"
 
-# trying to republsh should fail now.
+# trying to republish should fail now.
 sui client --client.config $CONFIG test-publish --build-env testnet --pubfile-path Pub.local.toml a
 
 ----- results -----
--- a/crates/sui/tests/shell_tests/with_network/ephemeral/ephemeral_publish_fail.sh
+++ b/crates/sui/tests/shell_tests/with_network/ephemeral/ephemeral_publish_fail.sh
@@ -26,5 +26,5 @@ sui client --client.config $CONFIG \
   test-publish --build-env testnet --pubfile-path Pub.local.toml a \
   > /dev/null || echo "failed to publish a"
 
-# trying to republsh should fail now.
+# trying to republish should fail now.
 sui client --client.config $CONFIG test-publish --build-env testnet --pubfile-path Pub.local.toml a


crates/sui/src/client_commands.rs
```diff
@@ -2023,15 +2023,12 @@ async fn compatibility_checks(
         }
     }
 
-    if !compiled_package.is_system_package()
-        && let Some(already_published) = compiled_package.published_root_module()
-    {
+    if !compiled_package.is_system_package() && compiled_package.published_root_module().is_some() {
```
mdgeorge4153, line 2026: I think it would be better to do this check up front, before doing all the compilation, etc. This is OK for now if we want to rush to cherry pick it, but maybe add a TODO and task?

crates/sui/tests/cli_tests.rs
```diff
@@ -4809,6 +4809,7 @@ async fn test_clever_errors() -> Result<(), anyhow::Error> {
             build_env: Some("testnet".to_string()),
             pubfile_path: Some(tempdir()?.path().join("localnet.toml")),
         },
+        ..Default::default()
```
mdgeorge4153:
```suggestion
        ..TestPublishArgs::default()
```

mdgeorge4153: slight preference

crates/sui/tests/shell_tests/with_network/ephemeral/ephemeral_publish_fail.snap
```diff
@@ -24,11 +24,26 @@ sui client --client.config $CONFIG \
   test-publish --build-env testnet --pubfile-path Pub.local.toml e \
   > /dev/null || echo "failed to build e"
 
+
+# publish a
+sui client --client.config $CONFIG \
+  test-publish --build-env testnet --pubfile-path Pub.local.toml a \
+  > /dev/null || echo "failed to publish a"
+
+# trying to republsh should fail now.
```
mdgeorge4153:
```suggestion
# trying to republish should fail now.
```

crates/sui/tests/shell_tests/with_network/ephemeral/ephemeral_publish_fail.snap
```diff
@@ -24,11 +24,26 @@ sui client --client.config $CONFIG \
   test-publish --build-env testnet --pubfile-path Pub.local.toml e \
   > /dev/null || echo "failed to build e"
 
+
+# publish a
+sui client --client.config $CONFIG \
+  test-publish --build-env testnet --pubfile-path Pub.local.toml a \
+  > /dev/null || echo "failed to publish a"
```
mdgeorge4153, line 31: One idea I thought of is that we can get the output in case the test fails by doing
```
 > output.log || cat output.log
```

crates/sui/tests/shell_tests/with_network/ephemeral/ephemeral_publish_fail.snap
```diff
@@ -24,11 +24,26 @@ sui client --client.config $CONFIG \
...
+success: false
+exit_code: 1
 ----- stdout -----
 === expect to fail when publishing e because prereqs aren't published ===
 failed to build e
+INCLUDING DEPENDENCY MoveStdlib
+INCLUDING DEPENDENCY Sui
+BUILDING a
+Failed to publish the Move module(s), reason: Your package is already published. You have to manually remove the publication entry to publish again.
+ - If you are doing a regular publish, you can remove the entry for your environment from `Published.toml`.
```
mdgeorge4153, line 46: Maybe we should list `upgrade` as the first option here?

crates/sui/tests/shell_tests/with_network/ephemeral/ephemeral_publish_fail.snap
manolisliolios: I'll defer to another Pr but I agree that's great idea.

crates/sui/tests/shell_tests/with_network/ephemeral/ephemeral_publish_fail.snap
manolisliolios: Given there's an explicit command, I think people do try to publish (so they wouldn't want to upgrade).

crates/sui/src/client_commands.rs
manolisliolios: Created an issue in our quality improvement backlog

PR #24862, opened 2026-01-09
Add EndpointManager for peer address updates

aschran:
## Description 

Replaces existing epoch-end channel in discovery protocol, allowing for intra-epoch updates.

Can be extended to support consensus address updates as well.

## Test plan 

Updated existing discovery test

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-network/src/discovery/mod.rs
+++ b/crates/sui-network/src/discovery/mod.rs
@@ -251,17 +251,19 @@ impl DiscoveryEventLoop {
         debug!(?peer_info, "Add committee member as preferred peer.");
         if let Some(old_peer_info) = self.network.known_peers().insert(peer_info.clone())
             && old_peer_info.address != peer_info.address
-            && let Some(address) = peer_info.address.first().cloned()
         {
             // Forcibly reconnect to the peer if its address(es) changed.
+            // If no valid new address is provided, we will remain disconnected.
             let _ = self.network.disconnect(peer_info.peer_id);
-            let network = self.network.clone();
-            self.tasks.spawn(async move {
-                // If this fails, ConnectionManager will retry.
-                let _ = network
-                    .connect_with_peer_id(address, peer_info.peer_id)
-                    .await;
-            });
+            if let Some(address) = peer_info.address.first().cloned() {
+                let network = self.network.clone();
+                self.tasks.spawn(async move {
+                    // If this fails, ConnectionManager will retry.
+                    let _ = network
+                        .connect_with_peer_id(address, peer_info.peer_id)
+                        .await;
+                });
+            }
         }
     }
 


crates/sui-network/src/discovery/mod.rs
```diff
@@ -218,28 +247,21 @@ impl DiscoveryEventLoop {
...
-                    // If this fails, ConnectionManager will retry.
-                    let _ = network
-                        .connect_with_peer_id(address, peer_info.peer_id)
-                        .await;
-                });
-            }
+    fn handle_peer_address_change(&mut self, peer_info: PeerInfo) {
+        debug!(?peer_info, "Add committee member as preferred peer.");
+        if let Some(old_peer_info) = self.network.known_peers().insert(peer_info.clone())
+            && old_peer_info.address != peer_info.address
```
akichidis, line 253: probably very minor, but since this is a vec even if it's the same addresses but in different order we'll disconnect and reconnect the peer. It is probably fine but leaving as comment. I guess a simple sort (if not already) when we create the `PeerInfo` might improve this

crates/sui-network/src/endpoint_manager.rs
```diff
@@ -0,0 +1,59 @@
...
+                                warn!(
+                                    ?addr,
+                                    "Skipping peer address: can't convert to anemo address"
+                                )
+                            })
+                            .ok()
+                    })
+                    .collect();
+                if anemo_addresses.is_empty() {
+                    warn!(?peer_id, "No valid addresses for peer after conversion");
```
akichidis, line 43: Do we still want to call the `peer_address_change` in this case?

crates/sui-network/src/discovery/mod.rs
aschran: I think we don't want to sort because it should be possible to provide addresses in a priority order. arguably if you change the priority order then trying to reconnect makes sense? in any case this is kind of moot, because for the most part we only have one address in here rn

crates/sui-network/src/endpoint_manager.rs
aschran: I am of two minds here. do you think if they reconfigure with no valid addresses we should update and disconnect, or keep to the old set (which might then be lost anyway on restart?)

akichidis: Well depends on how we want to make the endpoint work semantically overall.

* Provide empty addresses array: force the node disconnect from current peers
* Provide invalid addresses array: force the node disconnect from current peers

those two actions will have the same effect. If we don't consider it confusing to the user , then we can always attempt to update and disconnect. So we know that the endpoint will always do this. The user will have to be careful in this case to provide valid addresses. As you said a simple restart will "fix" this anyways

aschran: Good points, I updated to ensure consistent disconnect behavior. let's hope this is an edge case that doesn't matter much in practice! but it's probably good to have the option to forcibly disconnect from a peer if you want to

PR #24863, opened 2026-01-09
[pkg-alt] Enable recursive publishing

manolisliolios:
## Description 

This PR enables recursive publishing through `test-publish` command.

## Test plan 

Added new tests to cover some cases & unit tests for the package alt changes

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-package-alt/src/package/root_package.rs
```diff
@@ -226,6 +226,12 @@ impl<F: MoveFlavor + fmt::Debug> RootPackage<F> {
         self.filtered_graph.packages()
     }
 
+    /// Return the list of all packages in the root package's package graph (including itself and all
+    /// transitive dependencies). This includes the non-duplicate addresses only, sorted in topological order.
+    pub fn sorted_packages(&self) -> Vec<PackageInfo<'_, F>> {
```
mdgeorge4153, line 231: I wonder if we should just return this from `packages` and specify that it's sorted in the comment?

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
```diff
@@ -70,4 +70,8 @@ impl MoveFlavor for Vanilla {
     fn validate_manifest(_: &ParsedManifest) -> Result<(), String> {
         Ok(())
     }
+
+    fn is_system_address(address: &crate::schema::OriginalID) -> bool {
+        address == &OriginalID::from(0xBEEF)
```
mdgeorge4153, line 75: Still don't love this but don't see a more expedient option. Maybe we should add a comment

external-crates/move/crates/move-package-alt/src/package/root_package.rs
manolisliolios: We actually use `packages` sorted by name for another purpose iirc. We could always sort after getting them though.

Let's keep it as is, and when we do the API narrowing, we can decide to change?

PR #24868, opened 2026-01-09
address-balance: use TypeTag instead of TypeInput

bmwill:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
+++ b/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
@@ -499,7 +499,7 @@ impl<'pc, 'vm, 'state, 'linkage> Env<'pc, 'vm, 'state, 'linkage> {
     }
 
     /// Converts a VM runtime Type to an adapter Type.
-    pub fn adapter_type_from_vm_type(
+    fn adapter_type_from_vm_type(
         &self,
         vm_type: &vm_runtime_type::Type,
     ) -> Result<Type, ExecutionError> {


sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
```diff
@@ -494,7 +499,7 @@ impl<'pc, 'vm, 'state, 'linkage> Env<'pc, 'vm, 'state, 'linkage> {
     }
 
     /// Converts a VM runtime Type to an adapter Type.
-    fn adapter_type_from_vm_type(
+    pub fn adapter_type_from_vm_type(
```
tnowacki: I don't think this should be public, since we need to be careful about VM types being exposed

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
```diff
@@ -242,6 +242,11 @@ impl<'pc, 'vm, 'state, 'linkage> Env<'pc, 'vm, 'state, 'linkage> {
         self.adapter_type_from_vm_type(&runtime_type)
     }
 
+    pub fn load_type_tag(&self, idx: usize, ty: &TypeTag) -> Result<Type, ExecutionError> {
+        let runtime_type = self.load_vm_type_from_type_tag(Some(idx), ty)?;
+        self.adapter_type_from_vm_type(&runtime_type)
+    }
+
```
tnowacki, line 249: I'll leave @tzakian to ponder this one

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
bmwill: i think this was accidentally left in can remove

PR #24872, opened 2026-01-10
Add alt binary that uses framework for ingestion

nickvikeras:
## Description

This is a minimal change to use the alt framework for ingestion for the kv store indexer.

Design notes:

- The new binary and old binary share a watermark so they can can run in parallel.
- Similar to the blob store indexer, we use a batch size of 1 to ensure a throttled write doesn't cause a retry storm by retrying a large number of successful writes.
- It's a concurrent pipeline so checkpoints can be delivered out-of-order. This means there is a race condition where the last checkpoint of an epoch is indexed before the last checkpoint of the previous epoch. It's unlikely to occur under normal operations, but to deal with it I return an error if the previous epoch record doesn't exist on read. This way the pipeline will retry indefinitely, stalling that final update until the epoch record actually exists.
- The code in store.rs that writes out all the records is copy-pasted from worker.rs. The only change is to handle the race condition described in the above point. Keeping them decoupled will make further refactoring easier.
- I am ignoring tasks for now.

The trickiest thing was that the legacy binary's watermark stores "next checkpoint to process", while the new framework
uses "checkpoint_hi_inclusive" (last processed). I added code to translate between these to the new Store impl. I can rip that out when we turn off the legacy infra.

We can deploy this ASAP to get off of the legacy infra, and then I will make further changes to split up the tables into independent pipelines with their own watermarks, do some load testing, and make schema adjustments.

## Test plan

Let it bake in testnet for a while?

### DIFF
--- a/docker/sui-kvstore/Dockerfile
+++ b/docker/sui-kvstore/Dockerfile
@@ -1,0 +1,19 @@
+FROM rust:1.90-bullseye as build
+
+ARG PROFILE=release
+WORKDIR /work
+
+RUN apt-get update && apt-get install -y cmake clang
+
+COPY .git/ .git/
+COPY Cargo.toml Cargo.lock ./
+COPY consensus consensus
+COPY crates crates
+COPY sui-execution sui-execution
+COPY external-crates external-crates
+
+RUN cargo build --profile ${PROFILE} --bin sui-kvstore-alt
+
+FROM debian:12-slim as deploy
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates libjemalloc-dev
+COPY --from=build --chmod=755 /work/target/release/sui-kvstore-alt /opt/sui/bin/sui-kvstore-alt


crates/sui-kvstore/src/bigtable/store.rs
```diff
@@ -0,0 +1,198 @@
...
+                            "previous epoch {} not found when processing epoch {}",
+                            epoch_info.epoch - 1,
+                            epoch_info.epoch
+                        )
+                    })?;
+                prev.end_checkpoint = epoch_info.start_checkpoint.map(|sq| sq - 1);
+                prev.end_timestamp_ms = epoch_info.start_timestamp_ms;
+                self.client.save_epoch(prev).await?;
+            }
+            self.client.save_epoch(epoch_info).await?;
```
bmwill, line 102: Is this fall through case correct?

amnn: It looks to be -- it is writing the current checkpoint always, and the `if` is guarding updates to the previous checkpoint.

FWIW, the postgres schema for epochs (where we split writes to the start and end of the epoch) is designed that way exactly to avoid this race condition.

crates/sui-kvstore/src/bigtable/store.rs
```diff
@@ -0,0 +1,198 @@
...
+            .await?;
+        self.client.save_transactions(&transactions).await?;
+        self.client.save_checkpoint(checkpoint).await?;
+        if let Some(epoch_info) = checkpoint.epoch_info()? {
+            if epoch_info.epoch > 0 {
+                let mut prev = self
+                    .client
+                    .get_epoch(epoch_info.epoch - 1)
+                    .await?
+                    .ok_or_else(|| {
```
amnn: `with_context`?

amnn (approved):
Nice!

crates/sui-kvstore/src/bigtable/store.rs
bmwill: yep, i expect we'd just write to particular columns of the same table in two workflows in the rework

PR #24877, opened 2026-01-12
perf: avoid extra clones in sui transaction builder

lisenokdonbassenok:
Removed unnecessary clone() calls on address and signature byte vectors before passing them to ProgrammableTransactionBuilder::pure. The builder only needs a serializable view of the data and internally copies into a new BCS buffer, so taking references avoids redundant allocations and byte copies without changing the serialized wire format.


Bridgerz (approved):
LGTM

PR #24878, opened 2026-01-12
perf: cache tx digest in `OverloadTracker::remove_pending_certificate`

MoNyAvA:
Cache the transaction digest once at the beginning of remove_pending_certificate and reuse it inside the loop instead of recomputing it for every object. This removes redundant work on the hot path without changing the behavior of the overload tracker or its public interface.


stefan-mysten (approved):
Thanks @MoNyAvA

PR #24882, opened 2026-01-12
[prettier] bugfixes - format on save + tree-sitter nits

### LINKED ISSUE #24783: Prettier plugin fails on variables named 'spec'

damirka:
## Description 

- extension: mass replace no longer messes up files
- plugin: variables can be named "spec"
- closes #24783 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/tooling/prettier-extension/CHANGELOG.md
+++ b/external-crates/move/tooling/prettier-extension/CHANGELOG.md
@@ -1,7 +1,0 @@
-# Change Log
-
-All notable changes to the "prettier-move" extension will be documented in this file.
-
-## v0.1.0
-
-- alpha release of the Prettier Move extension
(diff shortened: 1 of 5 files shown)


damirka:
cc @Bridgerz — your annoyances will be gone with this release

PR #24892, opened 2026-01-12
[consensus] improve randomized tests

mwtian:
## Description 

- Add Linearizer, TransactionCertifier and CommitFinalizer to randomized tests.
- Share test fixture across randomized_tests and commit_finalizer tests.
- In randomized_tests, generate one DAG and ensure all runs produce the same output.

## Test plan 

CI

### DIFF
--- a/consensus/core/src/tests/randomized_tests.rs
+++ b/consensus/core/src/tests/randomized_tests.rs
@@ -1,19 +1,22 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::env;
+use std::{env, sync::Arc};
 
+use consensus_types::block::{Round, TransactionIndex};
 use rand::{Rng, SeedableRng, prelude::SliceRandom, rngs::StdRng};
 
 use crate::{
-    block::{BlockAPI, Slot},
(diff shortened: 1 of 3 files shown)


consensus/core/src/tests/randomized_tests.rs
```diff
@@ -79,131 +65,272 @@ async fn test_randomized_dag_all_direct_commit() {
...
-        let mut sequenced_leaders_2 = vec![];
+        let mut finalized_commits = vec![];
         let mut last_decided = Slot::new_for_test(0, 0);
         let mut i = 0;
         while i < all_blocks.len() {
             let chunk_size = random_test_setup
                 .seeded_rng
-                .gen_range(1..=(all_blocks.len() - i));
-            let chunk = &all_blocks[i..i + chunk_size];
+                .gen_range(1..=(all_blocks.len() - i).min(5 * num_authorities as usize));
```
akichidis, line 188: why `5 * num_authorities` ?

mwtian: I want to limit the degree of shuffling. If we shuffle round 1-100 blocks together, the test will be more about block manager than commit. In my next PR, I will use a different algorithm to pick next block from the DAG to process.

PR #24897, opened 2026-01-12
Clean up address balance test code to make tests more concise

mystenmark:
Big clean up to make tests more concise. (Note that I haven't pursued every possible cleanup yet in order to keep this PR from getting any bigger.)

### DIFF
--- a/crates/test-cluster/src/addr_balance_test_env.rs
+++ b/crates/test-cluster/src/addr_balance_test_env.rs
@@ -118,13 +118,14 @@ impl TestEnv {
         let tx = TestTransactionBuilder::new(address, gas, self.rgp)
             .transfer_sui_to_address_balance(FundSource::coin(gas), vec![(amount, address)])
             .build();
-        let (_, effects) = self
+        let (digest, effects) = self
             .cluster
             .sign_and_execute_transaction_directly(&tx)
             .await
             .unwrap();
         // update the gas object we used.
         self.gas_objects.get_mut(&address).unwrap()[0] = effects.gas_object().0;
+        self.cluster.wait_for_tx_settlement(&[digest]).await;
     }
 
     #[allow(dead_code)]


crates/test-cluster/src/addr_balance_test_env.rs
```diff
@@ -0,0 +1,347 @@
...
+        Option<Box<dyn Fn(ProtocolVersion, ProtocolConfig) -> ProtocolConfig + Send>>,
+}
+
+impl Default for TestEnvBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl TestEnvBuilder {
```
alex-mysten, line 39: Why not build this functionality directly into TestClusterBuilder / TestCluster? It seems many of these methods either delegate to TestCluster or are generally useful additions.

crates/sui-e2e-tests/tests/address_balance_compatibility_tests.rs
```diff
@@ -0,0 +1,168 @@
...
+use sui_test_transaction_builder::FundSource;
+use sui_types::{
+    base_types::{FullObjectRef, ObjectID, SequenceNumber, SuiAddress},
+    coin_reservation::ParsedObjectRefWithdrawal,
+};
+
+use test_cluster::addr_balance_test_env::TestEnvBuilder;
+
+#[sim_test]
+async fn test_coin_reservation_validation() {
```
alex-mysten, line 14: Do we have positive tests for coin reservation or are these TODO?

crates/test-cluster/src/addr_balance_test_env.rs
mystenmark: We could, I suppose - but I'm not sure I want TestCluster to become a kitchen sink. For instance TestEnv tracks accounts / gas objects, which doesn't seem like TestCluster's job.

crates/sui-e2e-tests/tests/address_balance_compatibility_tests.rs
mystenmark: Coming Soon to a pull request near you

PR #24901, opened 2026-01-12
Return address balance specific balances in json rpc

mystenmark:
- Add address balance to get_balance / get_all_balances
- Verify that json rpc balances agree with direct reads

### DIFF
--- a/crates/sui-open-rpc/spec/openrpc.json
+++ b/crates/sui-open-rpc/spec/openrpc.json
@@ -3193,7 +3193,8 @@
                 "coinType": "0x2::sui::SUI",
                 "coinObjectCount": 15,
                 "totalBalance": "3000000000",
-                "lockedBalance": {}
+                "lockedBalance": {},
+                "fundsInAddressBalance": "42"
               }
             ]
           }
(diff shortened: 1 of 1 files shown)


bmwill (approved):

crates/sui-json-rpc-types/src/sui_coin.rs
```diff
@@ -33,6 +33,13 @@ pub struct Balance {
     #[schemars(with = "HashMap<BigInt<u64>, BigInt<u128>>")]
     #[serde_as(as = "HashMap<BigInt<u64>, BigInt<u128>>")]
     pub locked_balance: HashMap<EpochId, u128>,
+
+    /// The portion of `total_balance` that resides in the address balance
+    /// rather than in the coin objects. `total_balance` is the total amount
+    /// of funds owned by the address. That is, do not add these two fields together.
+    #[schemars(with = "BigInt<u128>")]
+    #[serde_as(as = "BigInt<u128>")]
+    pub funds_in_address_balance: u128,
```
bmwill, line 42: Why u128 vs u64

mystenmark: Because all the other fields use u128?

bmwill: I guess in practice it doesn't matter...odd that the other fields are u128 though.

PR #24903, opened 2026-01-12
[move][move-vm][bella-ciao] Add an LRU for virtual tables

cgswords:
## Description 

Add an LRU for caching virtual tables by linkage as a huge perf win.

## Test plan 

Tim is going to test perf; the other tests all pass.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
@@ -11,16 +11,17 @@ use crate::{
     jit,
     runtime::telemetry::MoveCacheTelemetry,
     shared::{
-        constants::VIRTUAL_DISPATCH_TABLE_LRU_SIZE, linkage_context::LinkageHash, types::VersionId,
+        constants::VIRTUAL_DISPATCH_TABLE_CACHE_SIZE, linkage_context::LinkageHash,
+        types::VersionId,
     },
     validation::verification,
 };
 
-use lru::LruCache;
+use dashmap::DashMap;
 use move_vm_config::runtime::VMConfig;
-use parking_lot::RwLock;
+use quick_cache::sync::Cache as QCache;
 
-use std::{collections::HashMap, num::NonZero, sync::Arc};
+use std::sync::Arc;
 
 // -------------------------------------------------------------------------------------------------
 // Types
@@ -32,15 +33,15 @@ pub struct Package {
     pub runtime: Arc<jit::execution::ast::Package>,
 }
 
-type PackageCache = HashMap<VersionId, Arc<Package>>;
+type PackageCache = DashMap<VersionId, Arc<Package>>;
 
 /// The loader for the VM. This is the data structure is used to resolve packages and cache them
 /// and their types. This is then used to create the VTables for the VM.
 #[derive(Debug)]
 pub struct MoveCache {
     pub(crate) vm_config: Arc<VMConfig>,
-    pub(crate) package_cache: Arc<RwLock<PackageCache>>,
-    pub(crate) linkage_vtables: Arc<RwLock<LruCache<LinkageHash, VMDispatchTables>>>,
+    pub(crate) package_cache: Arc<PackageCache>,
+    pub(crate) linkage_vtables: Arc<QCache<LinkageHash, VMDispatchTables>>,
     pub(crate) interner: Arc<IdentifierInterner>,
 }
 
@@ -60,11 +61,9 @@ impl MoveCache {
     pub fn new(vm_config: Arc<VMConfig>) -> Self {
         Self {
             vm_config,
-            package_cache: Arc::new(RwLock::new(HashMap::new())),
+            package_cache: Arc::new(DashMap::new()),
             interner: Arc::new(IdentifierInterner::new()),
-            linkage_vtables: Arc::new(RwLock::new(LruCache::new(
-                NonZero::new(VIRTUAL_DISPATCH_TABLE_LRU_SIZE).unwrap(),
-            ))),
+            linkage_vtables: Arc::new(QCache::new(VIRTUAL_DISPATCH_TABLE_CACHE_SIZE)),
         }
     }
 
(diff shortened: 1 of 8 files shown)


external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
```diff
@@ -149,13 +189,16 @@ impl MoveCache {
...
+        let vtable_lru_count = self.linkage_vtables.read().len() as u64;
+
         MoveCacheTelemetry {
             package_cache_count,
             total_arena_size,
             module_count,
             function_count,
             type_count,
             interner_size,
+            vtable_lru_count,
```
tzakian: ❤️

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -125,31 +129,42 @@ impl MoveRuntime {
...
-                    package_store,
-                    txn_telemetry,
-                    &self.cache,
-                    &self.natives,
-                    all_packages,
-                )?;
+                let linkage_hash = match link_context.to_linkage_hash(&self.cache.interner) {
+                    Ok(hash) => hash,
+                    Err(err) => {
+                        // Anything better to do here?
```
tzakian: Not really I don't think

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -125,31 +129,42 @@ impl MoveRuntime {
...
+                        .finish(Location::Undefined)
+                    );
+                }
 
-                let virtual_tables = VMDispatchTables::new(
-                    self.vm_config.clone(),
-                    self.cache.interner.clone(),
-                    runtime_packages,
-                )?;
+                if !(link_context == *virtual_tables.link_context) {
```
tzakian:
```suggestion
                if link_context != *virtual_tables.link_context {
```

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -125,31 +129,42 @@ impl MoveRuntime {
...
 
-                let runtime_packages = packages
-                    .into_values()
-                    .map(|pkg| (pkg.runtime.original_id, Arc::clone(&pkg.runtime)))
-                    .collect::<BTreeMap<OriginalId, Arc<jit::execution::ast::Package>>>();
+                // This is more a sanity check than anything else. The VMDispatchTables should
+                // never have precomputed type depths, as those are computed on-demand.
+                if !virtual_tables.type_depths.is_empty() {
+                    return Err(
+                        move_binary_format::errors::PartialVMError::new(
```
tzakian: Nit: import `PartialVMError`?

external-crates/move/crates/move-vm-runtime/src/shared/linkage_context.rs
```diff
@@ -107,4 +115,17 @@ impl LinkageContext {
             .cloned()
             .collect::<BTreeSet<_>>())
     }
+
+    pub(crate) fn to_linkage_hash(
+        &self,
+        interner: &cache::identifier_interner::IdentifierInterner,
```
tzakian: Nit: import `IdentifierInterner`?

external-crates/move/crates/move-vm-runtime/src/cache/identifier_interner.rs
```diff
@@ -59,21 +60,27 @@ impl IdentifierInterner {
...
     }
 
-    #[allow(dead_code)]
+    /// Get the interned address value. This may raise an invariant error if `try_get_or_intern`
+    /// fails, but that's likely a serious OOM issue.
+    pub(crate) fn intern_address(
+        &self,
+        address: &AccountAddress,
+    ) -> PartialVMResult<IdentifierKey> {
+        self.get_or_intern_str_internal(&address.short_str_lossless())
```
tzakian: I'd probably use `address.to_hex()` or `address.into()` (which will call `hex::encode` on the address) here instead of `short_str_lossless`. 

Main thing is that `short_str_lossless` tries to shorten the address. It's unlikely to have any issues since the function is pretty simple, but I don't want a bug in that to cause possible issues for keying and looking up the vtables. 

(I also realize this will be somewhat of a moot point after your next PR, so if you'd prefer to just move ahead with this since it's going to be replaced soon anyways that's fine).

external-crates/move/crates/move-vm-runtime/src/shared/constants.rs
```diff
@@ -27,3 +27,13 @@ pub const HISTORICAL_MAX_TYPE_TO_LAYOUT_NODES: u64 = 256;
...
+
+/// Size of the type depth LRU
+/// TODO(vm-rewrite): find a good bound for this
+pub const TYPE_DEPTH_LRU_SIZE: usize = 16_384;
+
+/// Size of the linkage-cahge virtual dispatch LRU
+/// TODO(vm-rewrite): find a good bound for this
+/// This number is currently 1 GB / 128 bytes (size of VMDispatchTables), giving approximately
+/// a gigabytes of storage to VTables (though this disregards key and LRU overhead).
+pub const VIRTUAL_DISPATCH_TABLE_LRU_SIZE: usize = 7_812_500;
```
tzakian: This isn't used anywhere. Did you mean to use this when initializing the linkage vtable lru? (but if so that seems kinda big to me :p)

tzakian (approved):
Couple comments, otherwise looks pretty good

external-crates/move/crates/move-vm-runtime/src/shared/constants.rs
cgswords: It can be big because this is only like 1gb overall, but I reduced it to 1 million. We are just on a sliding scale of performance now: big cache or big jit time. Maybe your replay can validate the 1-million number, or indicate if we direly need an LFU.

external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
cgswords: Another good metric would be to record the number of misses.

external-crates/move/crates/move-vm-runtime/src/shared/linkage_context.rs
```diff
@@ -22,6 +25,11 @@ pub struct LinkageContext {
     pub linkage_table: BTreeMap<OriginalId, VersionId>,
 }
 
+/// A hashable representation of a linkage context, for caching purposes.
+/// This is a vector of (key, value) pairs representing the linkage table.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) struct LinkageHash(Vec<(AccountAddress, AccountAddress)>);
```
tzakian: can this be `pub`? I will want to use this for VM reuse in execution probably.

external-crates/move/crates/move-vm-runtime/src/shared/linkage_context.rs
```diff
@@ -107,4 +115,13 @@ impl LinkageContext {
...
     }
+
+    pub(crate) fn to_linkage_hash(&self) -> PartialVMResult<LinkageHash> {
+        Ok(LinkageHash(
+            self.linkage_table
+                .iter()
+                .map(|(k, v)| (*k, *v))
+                .collect::<Vec<_>>(),
+        ))
+    }
```
tzakian: can this also be `pub`?

Also, this no longer needs to be fallible, and making it infallible will make the callsite the runtime a bit nicer too.

external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
tzakian: It would be super interesting to look at this and from this get a sense of how "warm" the VM at any given moment.

tzakian: I actually really like this metric. Feel free to add it, but if you don't I will here ;)

external-crates/move/crates/move-vm-runtime/src/shared/linkage_context.rs
tzakian: I actually wonder if we should have `LinkageHash` at all? As the vector of pairs gives us less of a guarantee than a `BTreeMap` of k,v (one has a canonical order, the other does not).

cgswords: I have a plan to eventually replace it with something more-efficient. It is currently a `BTreeMap`, but my hope is to one day do something better than that.

PR #24905, opened 2026-01-13
Add APIs for reading pending balance changes

mystenmark:


crates/sui-framework/packages/sui-framework/sources/balance.move
```diff
@@ -125,6 +125,38 @@ public fun settled_funds_value<T>(root: &sui::accumulator::AccumulatorRoot, addr
...
+public fun positive_pending_funds_value<T>(
+    root: &sui::accumulator::AccumulatorRoot,
+    address: address,
+): Option<u64> {
+    let settled = settled_funds_value<T>(root, address);
+    let deposits = pending_funds_deposits<T>(address);
+    let withdrawals = pending_funds_withdrawals<T>(address);
+    let total_available = (settled as u128) + (deposits as u128);
+    if (total_available >= (withdrawals as u128)) {
+        std::option::some(((total_available - (withdrawals as u128)) as u64))
```
tnowacki:
```suggestion
        std::option::some((total_available - (withdrawals as u128)) as u64)
```

crates/sui-framework/packages/sui-framework/sources/balance.move
```diff
@@ -125,6 +125,38 @@ public fun settled_funds_value<T>(root: &sui::accumulator::AccumulatorRoot, addr
...
+    } else {
+        std::option::none()
+    }
```
tnowacki:
```suggestion
    let settled = settled_funds_value<T>(root, address) as u128;
    let deposits = pending_funds_deposits<T>(address) as u128;
    let withdrawals = pending_funds_withdrawals<T>(address) as u128;
    let total_available = settled + deposits;
    if (total_available >= withdrawals) option::some((total_available - withdrawals) as u64)
    else option::none()
```

crates/sui-framework/packages/sui-framework/sources/balance.move
```diff
@@ -125,6 +125,38 @@ public fun settled_funds_value<T>(root: &sui::accumulator::AccumulatorRoot, addr
...
+/// This is the sum of all `Balance<T>` values being withdrawn from the address during this
+/// transaction, before they are settled.
+public fun pending_funds_withdrawals<T>(address: address): u64 {
+    sui::accumulator::pending_withdrawals<Balance<T>>(address)
+}
+
+/// Read the expected balance for the given address after the current transaction settles.
+/// Returns `Some(settled + deposits - withdrawals)` if the result is non-negative,
+/// or `None` if the result would underflow (withdrawals exceed settled + deposits).
+public fun positive_pending_funds_value<T>(
```
tnowacki, line 157: I'm not sure how to feel about this. I _think_ I'd rather just wait for `i64`? But at the same time `i64` technically can be too small for this and error, where that isn't true for `Option<u64>`

crates/sui-framework/packages/sui-framework/sources/balance.move
```diff
@@ -114,6 +114,49 @@ public fun withdraw_funds_from_object<T>(obj: &mut UID, value: u64): Withdrawal<
...
+/// Read the total pending deposits for the given address within the current transaction.
+/// This is the sum of all `Balance<T>` values sent via `send_funds` to the address during this
+/// transaction, before they are settled.
+public fun pending_funds_deposits<T>(address: address): u64 {
+    sui::accumulator::pending_deposits<Balance<T>>(address)
+}
+
+/// Read the total pending withdrawals for the given address within the current transaction.
+/// This is the sum of all `Balance<T>` values being withdrawn from the address during this
+/// transaction, before they are settled.
```
manolisliolios: I assume the scope of that is PTB (so the sum of balances withdrawn across the commands of the PTB?)

crates/sui-framework/packages/sui-framework/sources/balance.move
```diff
@@ -114,6 +114,49 @@ public fun withdraw_funds_from_object<T>(obj: &mut UID, value: u64): Withdrawal<
...
+        return 0
+    };
+    let val: u128 = root.u128_read<Balance<T>>(address);
+    let val = std::u128::min(std::u64::max_value!() as u128, val);
+    val as u64
+}
+
+/// Read the total pending deposits for the given address within the current transaction.
+/// This is the sum of all `Balance<T>` values sent via `send_funds` to the address during this
+/// transaction, before they are settled.
```
manolisliolios: That means that the value would change depending on when (during a consensus commit) you read it (which is out of control for the user), even from un-related tx deposits?

crates/sui-framework/packages/sui-framework/sources/balance.move
mystenmark: yes, any deposit or withdraw made during the current tx (not the current command)

crates/sui-framework/packages/sui-framework/sources/balance.move
mystenmark: ah, i need to clarify this. This can only tell you about deposits made _by the current transaction_. Deposits made by other transactions are not visible until the next settlement.

PR #24907, opened 2026-01-13
[move-stdlib] saturating arithmetic 

tnowacki:
## Description 

- Following changes to add checked arithmetic, this adds the saturating versions
- There is no saturating_div since it would not perform any function, and could be confusing that it still errors even when the rhs is 0

## Test plan 

- New tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority/snapshots/sui_core__authority__execution_time_estimator__tests__execution_time_observer_v109.snap
+++ b/crates/sui-core/src/authority/snapshots/sui_core__authority__execution_time_estimator__tests__execution_time_observer_v109.snap
@@ -1,0 +1,156 @@
+---
+source: crates/sui-core/src/authority/execution_time_estimator.rs
+expression: snapshot_data
+---
+protocol_version: 109
+consensus_observations:
+  - - MakeMoveVec
+    - observations:
+        - - 8
+          - secs: 0
+            nanos: 26000000
+        - - 10
+          - secs: 0
+            nanos: 16000000
+        - - 5
+          - secs: 0
+            nanos: 19000000
+        - - 1
+          - secs: 0
+            nanos: 25000000
+      stake_weighted_median:
+        secs: 0
+        nanos: 25000000
+  - - MergeCoins
+    - observations:
+        - - 6
+          - secs: 0
+            nanos: 63000000
+        - - 1
+          - secs: 0
+            nanos: 21000000
+        - - 2
+          - secs: 0
+            nanos: 32000000
+        - - 7
+          - secs: 0
+            nanos: 0
+      stake_weighted_median:
+        secs: 0
+        nanos: 32000000
+  - - SplitCoins
+    - observations:
+        - - 9
+          - secs: 0
+            nanos: 56000000
+        - - 0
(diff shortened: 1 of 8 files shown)


crates/sui-framework/packages/move-stdlib/sources/macros.move
```diff
@@ -108,6 +108,28 @@ public macro fun num_checked_div<$T>($x: $T, $y: $T): Option<$T> {
     if (y == 0) option::none() else option::some(x / y)
 }
 
+public macro fun num_saturating_add<$T>($x: $T, $y: $T, $max_t: $T): $T {
+    let x = $x;
+    let y = $y;
+    let max_t = $max_t;
+    if (y > max_t - x) max_t else x + y
```
tzakian, line 115: I'm guessing it's OK to abort if `max_t > x`? (should never be the case for nums, but since this is public).

crates/sui-framework/packages/move-stdlib/sources/macros.move
```diff
@@ -108,6 +108,28 @@ public macro fun num_checked_div<$T>($x: $T, $y: $T): Option<$T> {
     if (y == 0) option::none() else option::some(x / y)
 }
 
+public macro fun num_saturating_add<$T>($x: $T, $y: $T, $max_t: $T): $T {
```
tzakian: Thoughts on making these macros be `public(package)` as otherwise you may run into some sharp(ish) edges.

crates/sui-framework/packages/move-stdlib/sources/macros.move
cgswords, line 115: It might be nice to use parens in case someone messes up precedence in the future.

crates/sui-framework/packages/move-stdlib/sources/macros.move
```diff
@@ -108,6 +108,28 @@ public macro fun num_checked_div<$T>($x: $T, $y: $T): Option<$T> {
...
+    let y = $y;
+    if (x < y) 0 else x - y
+}
+
+public macro fun num_saturating_mul<$T>($x: $T, $y: $T, $max_t: $T): $T {
+    let x = $x;
+    let y = $y;
+    let max_t = $max_t;
+    if (x == 0 || y == 0) 0
+    else if (y > max_t / x) max_t
```
cgswords, line 129: Same comment about precedence parsing

crates/sui-framework/packages/move-stdlib/sources/macros.move
tnowacki: Who is messing up precedence?

crates/sui-framework/packages/move-stdlib/sources/macros.move
tnowacki: why not. gonna do em all, let's see who complains

PR #24909, opened 2026-01-13
[docs] ix Sui CLI replay command and command count

Forostovec:
Updated the Sui CLI reference to match the current CLI behavior. The replay section previously documented the command as sui client replay, but the actual top-level command implemented in SuiCommand is sui replay. I also fixed the “five most useful” wording to reflect that there are six CLI groups listed.


jessiemongeon1 (approved):

jessiemongeon1:
Thanks @Forostovec , this fix has been addressed in a larger batch PR.

PR #24910, opened 2026-01-13
[benchmark] add gas w/ address balances to composite probabalistic workload

alex-mysten:
## Description 

Composite workload supports paying gas with address balances

## Test plan 

Simtested

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/src/workloads/composite/mod.rs
+++ b/crates/sui-benchmark/src/workloads/composite/mod.rs
@@ -25,6 +25,7 @@ use operations::{InitRequirement, Operation, OperationResources, ResourceRequest
 use rand::rngs::SmallRng;
 use rand::{Rng, SeedableRng, thread_rng};
 use std::collections::HashMap;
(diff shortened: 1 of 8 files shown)


crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -363,6 +372,9 @@ impl Payload for CompositePayload {
 
         let pool = self.pool.blocking_read();
 
+        let use_address_balance_gas = self.config.address_balance_gas_probability > 0.0
+            && self.rng.gen_range(0.0..1.0) < self.config.address_balance_gas_probability;
```
mystenmark: nit: can use https://docs.rs/rand/latest/rand/trait.Rng.html#method.random_bool

PR #24912, opened 2026-01-13
reduce public interface to package-alt

mdgeorge4153:
## Description 

This makes everything in move-package-alt private except a few things.

## Test plan 

Reran existing tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-package-alt/src/git/cache.rs
+++ b/external-crates/move/crates/move-package-alt/src/git/cache.rs
@@ -271,7 +271,7 @@ impl GitTree {
     /// The path to the folder containing the cached repo (without the addition of the path within
     /// the repo)
     #[cfg(test)]
-    pub fn repo_fs_path(&self) -> &Path {
+    fn repo_fs_path(&self) -> &Path {
         &self.path_to_repo
     }
 }
--- a/external-crates/move/crates/move-package-alt/src/git/errors.rs
+++ b/external-crates/move/crates/move-package-alt/src/git/errors.rs
@@ -56,21 +56,25 @@ pub enum CommandErrorKind {
 }
 
 impl GitError {
-    pub fn dirty(repo: &str) -> Self {
+    pub(crate) fn dirty(repo: &str) -> Self {
         Self::Dirty {
             repo: repo.to_string(),
         }
     }
 
-    pub fn io_error(cmd: &Command, cwd: &Option<&PathBuf>, error: std::io::Error) -> Self {
+    pub(crate) fn io_error(cmd: &Command, cwd: &Option<&PathBuf>, error: std::io::Error) -> Self {
         Self::command_error(cmd, cwd, CommandErrorKind::IoError(error))
     }
 
-    pub fn nonzero_exit_status(cmd: &Command, cwd: &Option<&PathBuf>, code: ExitStatus) -> Self {
+    pub(crate) fn nonzero_exit_status(
+        cmd: &Command,
+        cwd: &Option<&PathBuf>,
+        code: ExitStatus,
+    ) -> Self {
         Self::command_error(cmd, cwd, CommandErrorKind::ErrorCode(code))
     }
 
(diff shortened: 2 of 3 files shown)


external-crates/move/crates/move-package-alt/src/cli/build.rs
stefan-mysten, line 1: Why not remove the CLI code in `pkg-alt`, is it still needed?

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -93,9 +72,7 @@ impl MoveFlavor for SuiFlavor {
         IndexMap::from([(testnet.name, testnet.id), (mainnet.name, mainnet.id)])
     }
 
-    fn implicit_dependencies(
-        environment: &EnvironmentID,
-    ) -> BTreeMap<PackageName, ReplacementDependency> {
+    fn implicit_dependencies(_: &EnvironmentID) -> BTreeMap<PackageName, ReplacementDependency> {
```
stefan-mysten, line 75: Why not remove the `envId` param instead?

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -108,7 +85,7 @@ impl MoveFlavor for SuiFlavor {
         ])
     }
 
-    fn system_deps(environment: &EnvironmentID) -> BTreeMap<SystemDepName, LockfileDependencyInfo> {
+    fn system_deps(_: &EnvironmentID) -> BTreeMap<SystemDepName, LockfileDependencyInfo> {
```
stefan-mysten, line 88: Ditto ^

external-crates/move/crates/move-package-alt/src/compatibility/mod.rs
```diff
@@ -198,28 +194,6 @@ fn strip_comments(source: &str) -> String {
     result
 }
 
-/// Return legacy package metadata; this is needed for tests in sui side
-pub fn parse_legacy_package_info(
```
stefan-mysten, line 202: not needed anymore?

external-crates/move/crates/move-package-alt/src/package/lockfile.rs
```diff
@@ -23,7 +22,7 @@ pub struct Lockfiles {
 
 impl Lockfiles {
     /// Read `Move.lock` from `path`; returning [None] if it doesn't exist
-    pub fn read_from_dir<F: MoveFlavor>(
+    pub fn read_from_dir(
```
stefan-mysten, line 25: Shouldn't this be a `pub(crate)`?

manolisliolios (approved):
good stuff

crates/sui-package-alt/src/sui_flavor.rs
mdgeorge4153: I think it's sensible that the implicit dependencies could depend on the environment, even though they currently don't, so I think it makes sense to leave the argument in the interface even though it's not currently used.

external-crates/move/crates/move-package-alt/src/compatibility/mod.rs
mdgeorge4153: Correct, we were just using this to pull out the name of the legacy framework packages, and this information is already exposed for all packages using the `display_name` method. This way we prevent information about legacy packages from leaking to the rest of the system

external-crates/move/crates/move-package-alt/src/package/lockfile.rs
mdgeorge4153: The `Lockfiles` type itself is `pub(crate)`, so this is effectively `pub(crate)` anyhow

PR #24921, opened 2026-01-13
docs: fix Validator Tools card link

VolodymyrBg:
The Validator Tools card on the Sui Validators index page was linking to the validator configuration guide instead of the validator node tools guide. This change updates the href to /guides/operator/validator/node-tools so the tools card points to the correct documentation and the node tools guide is discoverable from the validator overview.


jessiemongeon1:
Thanks @VolodymyrBg , we've included this fix in a larger batch update.

PR #24924, opened 2026-01-13
Include user signatures in proto conversion

nickvikeras:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-types/src/rpc_proto_conversions.rs
+++ b/crates/sui-types/src/rpc_proto_conversions.rs
@@ -146,8 +146,13 @@ impl TryFrom<&Checkpoint> for crate::full_checkpoint_content::Checkpoint {
             .transactions()
             .iter()
             .zip(user_signatures)
-            .map(|(tx, user_signatures)| tx_with_signatures_try_from_proto(tx, user_signatures))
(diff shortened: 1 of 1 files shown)


crates/sui-types/src/rpc_proto_conversions.rs
```diff
@@ -153,6 +160,40 @@ impl TryFrom<&Checkpoint> for crate::full_checkpoint_content::Checkpoint {
...
+                    .map_err(|e| TryFromProtoError::invalid("events.bcs", e))
+            })
+            .transpose()?,
+        unchanged_loaded_runtime_objects: value
+            .effects()
+            .unchanged_loaded_runtime_objects()
+            .iter()
+            .map(TryInto::try_into)
+            .collect::<Result<_, _>>()?,
+    })
```
bmwill: this could be simplified to do the try_into call (from the function below this) and then assigning the signatures to the signatures field to dedup this conversation logic.

PR #24925, opened 2026-01-13
Add jitter to watermark interval

nickvikeras:
## Description

I enabled 3 checkpoint bucket writers and they were racing to update the watermark file every minute. This didn't cause any actual problems, but it did cause a bunch of error level log spam which can be confusing/misleading to folks looking at the dashboard.

Adding jitter spaced out the updates and I am no longer seeing the 429 conflict errors in the log.

## Test plan

Deployed to testnet.

---

## Release notes

- [x] Indexing Framework: Adding optional jitter to watermark update interval.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
@@ -75,7 +75,7 @@ pub(super) fn commit_watermark<H: Handler + 'static>(
             next_checkpoint, "Starting commit watermark task"
         );
 
-        let mut next_wake = config.watermark_interval_with_jitter();
+        let mut next_wake = tokio::time::Instant::now();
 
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
```diff
@@ -80,9 +75,16 @@ pub(super) fn commit_watermark<H: Handler + 'static>(
             next_checkpoint, "Starting commit watermark task"
         );
 
+        let mut next_wake =
+            tokio::time::Instant::now() + config.watermark_interval_with_jitter();
```
amnn: Could we have the function on `config` return the Instant directly, instead of the Duration?

nickvikeras: done

PR #24929, opened 2026-01-14
[consensus] add RandomDag and RandomDagIterator for testing commit logic

mwtian:
## Description 

Improve the pattern coverage in generated random DAG. The new tests are able to uncover issues previously not found with other tests.

## Test plan 

```
python ./scripts/simtest/seed-search.py test_randomized_dag_with_4_authorities --test consensus_dag_tests --num-seeds 100
python ./scripts/simtest/seed-search.py test_randomized_dag_with_7_authorities --test consensus_dag_tests --num-seeds 100
```

### DIFF
--- a/consensus/core/src/commit_test_fixture.rs
+++ b/consensus/core/src/commit_test_fixture.rs
@@ -7,18 +7,14 @@
 use std::sync::Arc;
 
 use consensus_config::{AuthorityIndex, Stake};
-#[cfg(msim)]
 use consensus_types::block::BlockRef;
 use consensus_types::block::{Round, TransactionIndex};
 use mysten_metrics::monitored_mpsc::unbounded_channel;
 use parking_lot::RwLock;
-#[cfg(msim)]
 use rand::prelude::SliceRandom;
 use rand::{Rng, rngs::StdRng};
 
-#[cfg(msim)]
 use crate::Transaction;
-#[cfg(msim)]
 use crate::block::{BlockTransactionVotes, TestBlock, genesis_blocks};
 use crate::{
     block::{BlockAPI, VerifiedBlock},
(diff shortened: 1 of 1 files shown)


consensus/core/src/commit_test_fixture.rs
```diff
@@ -195,3 +208,369 @@ impl CommitTestFixture {
...
+            authorities.shuffle(rng);
+            let selected_authorities = &authorities[..n];
+
+            let mut current_round_blocks = Vec::new();
+
+            for &authority in selected_authorities {
+                // First, select exactly quorum blocks from the previous round.
+                let mut prev_round_blocks = last_round_blocks.clone();
+                prev_round_blocks.shuffle(rng);
+                prev_round_blocks.truncate(quorum_threshold);
```
akichidis: We assume that the committee members have all equal stake right?

consensus/core/src/commit_test_fixture.rs
```diff
@@ -195,3 +208,369 @@ impl CommitTestFixture {
...
+/// `visited_round + 1` to `quorum_round + max_step`, simulating arrival with delays.
+pub struct RandomDagIterator<'a> {
+    dag: &'a RandomDag,
+    rng: &'a mut StdRng,
+    quorum_threshold: Stake,
+    max_step: Round,
+    // Highest round where all prior rounds have quorum stake visited.
+    quorum_round: Round,
+    // Highest round where all prior rounds have all blocks visited.
+    visited_round: Round,
```
akichidis: Maybe that could be something like `completed_round` ?

consensus/core/src/commit_test_fixture.rs
mwtian: In tests indeed all authorities have 1 stake. But changing to use Stake properly.

PR #24779, opened 2026-01-05
graphql: NameRecord

amnn:
## Description 

Replace `Query.suinsName` and `IAddressable.defaultSuinsName` with `Query.nameRecord` and `IAddressable.defaultNameRecord` to match the API offered by gRPC, and required by the TS SDK.

Also adds a way to fetch an `Address` by its SuiNS name as part of `Query.address`.

## Test plan 

Updated E2E tests

```
$ cargo nextest run            \
  -p sui-indexer-alt-e2e-tests \
  --test graphql_name_service_tests
```

## Scope

- #24621 
- #24652 
- #24768 
- #24769 
- #24770 
- #24771 
- #24772 
- #24773 
- #24774 
- #24775 
- #24776 
- #24777 
- #24778

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Replace `Query.suinsName(name: ...)` with `Query.address(name: ...)`, replace `IAddressable.defaultSuinsName` with `IAddressable.defaultNameRecord.target`, and add `Query.nameRecord` for fetching the SuiNS NameRecord for a given SuiNS name.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #24778, opened 2026-01-05
docs(graphql): Explain the concept of scopes for GraphQL

amnn:
## Description

Document the concept of scopes, which sets the checkpoint being viewed and bounds on objects being loaded at any point in a GraphQL request.

## Test plan

:eyes:

## Stack

- #24621 
- #24652 
- #24768 
- #24769 
- #24770 
- #24771 
- #24772 
- #24773 
- #24774 
- #24775 
- #24776 
- #24777

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #24776, opened 2026-01-05
graphql: Simplify SystemState and ValidatorSet APIs

amnn:
## Description 

With the introduction of `MoveValue.extract` and `MoveValue.asAddress`, the system state and validator set APIs can be simplified: Most of the fields that were exposed through GraphQL can be fetched through `MoveValue.extract`.

The only exception is `ValidatorSet.activeValidators`, as Display v2 doesn't have a way of extracting some fields under a vector.

## Test plan 

E2E tests:

```
$ cargo nextest run -p sui-indexer-alt-e2e-tests
```

## Stack

- #24621
- #24652
- #24768
- #24769 
- #24770 
- #24771 
- #24772 
- #24773 
- #24774 
- #24775 

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Remove fields related to the system state from `Epoch`, in favour of exposing the whole system state as `Epoch.systemState: MoveValue`. Similarly replace most fields on `ValidatorSet` with `ValidatorSet.contents: MoveValue`, and `Validator` with `Validator.contents: MoveValue`.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #24774, opened 2026-01-05
graphql: MoveValue.format

amnn:
## Description 

Add a field to GraphQL for evaluating a single Display v2 format string against a `MoveValue`, without having to publish a `Display<T>` object.

## Test plan 

New E2E tests:

```
$ cargo nextest run -p sui-indexer-alt-e2e-tests
```

## Stack

- #24621
- #24652
- #24768
- #24769
- #24770 
- #24771 
- #24772 
- #24773 

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Add `MoveValue.format` to evaluate a single format string against a Move value.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #24935, opened 2026-01-14
graphql: partial errors for Display-related fields

amnn:
## Description

Update the fields that are newly added as of the Display stack to support partial errors.

## Test plan

E2E tests:

```
$ cargo nextest run -p sui-indexer-alt-e2e-tests -j3
```

## Stack

- #24621 
- #24652 
- #24768 
- #24769 
- #24770 
- #24771 
- #24772 
- #24773 
- #24774 
- #24775 
- #24776 
- #24777 
- #24778
- #24779 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/error.rs
+++ b/crates/sui-indexer-alt-graphql/src/error.rs
@@ -31,7 +31,7 @@ pub(crate) mod code {
     pub const RESOURCE_EXHAUSTED: &str = "RESOURCE_EXHAUSTED";
 }
 
-#[derive(thiserror::Error, Debug, Clone)]
+#[derive(thiserror::Error, Debug)]
 pub(crate) enum RpcError<E: std::error::Error = Infallible> {
     /// An error that is the user's fault.
     BadUserInput(Arc<E>),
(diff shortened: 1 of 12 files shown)


crates/sui-indexer-alt-graphql/src/error.rs
```diff
@@ -128,6 +128,20 @@ impl<E: std::error::Error> From<RpcError<E>> for async_graphql::Error {
     }
 }
 
+impl<E: std::error::Error> Clone for RpcError<E> {
```
tpham-mysten, line 131: Why is this needed?

amnn: This was needed to allow `Result<_, RpcError<E>>` to be allowable as an output type even when `E` itself doesn't implement `Clone` (which is not necessary for `RpcError<E>` to implement `Clone` because the inner user error is wrapped in an `Arc`.

PR #24939, opened 2026-01-14
Apply consensus_skip_gced_blocks_in_direct_finalization

mwtian:


consensus/core/src/commit_observer.rs
```diff
@@ -149,9 +149,11 @@ impl CommitObserver {
     }
 
     async fn recover_and_send_commits(&mut self, commit_consumer: &CommitConsumerArgs) {
+        self.truncate_bad_finalized_commits();
```
mystenmark, line 152: if the validators crash or restart for any reason, they will truncate again, right? should we write a sentinel file to prevent this?

mwtian: Once the validator restarts, it will skip truncation because the bad commit will have the correct info populated.

PR #24945, opened 2026-01-14
[bella-ciao][perf]: resolution vm in sptbs

tzakian:
## Description 

Add a pre-loading pass to compute a "resolution VM" for resolving any types during compilation or that may appear in the PTB.

This also adds caching for VMs that are created during execution of the PTBs. It also uses the same VM for execution and taking the user events generated from the PTB.

This also computes final writeout linkage to determine type layouts for all types that need to be written out.

## Test plan 

Existing tests + perf testing + added a test to ensure that we don't reuse the resolution VM for function resolution during sptb compilation.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -135,6 +140,30 @@ macro_rules! charge_gas {
     ($context:ident, $case:ident, $value_view:expr) => {{ charge_gas_!($context.gas_charger, $context.env, $case, $value_view) }};
 }
 
+macro_rules! with_vm {
```
tnowacki, line 144: Maybe a comment or two here

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -205,6 +234,8 @@ pub struct Context<'env, 'pc, 'vm, 'state, 'linkage, 'gas> {
     user_events: Vec<(ModuleId, StructTag, Vec<u8>)>,
     // runtime data
     locations: Locations,
+    // LRU cache of Move VMs created this transaction for different linkage contexts so that we can reuse them.
+    tearoffs: lru::LruCache<LinkageHash, MoveVM<'env>>,
```
cgswords: Nit: Prefer `quick_cache` as `lru` is evidently not well-maintained.

cgswords: Also, name this something more-sensible.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -517,17 +559,75 @@ impl<'env, 'pc, 'vm, 'state, 'linkage, 'gas> Context<'env, 'pc, 'vm, 'state, 'li
         let new_events = events
             .into_iter()
             .map(|(tag, value)| {
-                let layout = self.env.type_layout_for_struct(&tag)?;
```
cgswords, line 520: What is the relative utility of keeping the VM cache on the context instead of the `env`, to avoid this argument structure? I suppose instead you would need to pass the linkage?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -517,17 +559,75 @@ impl<'env, 'pc, 'vm, 'state, 'linkage, 'gas> Context<'env, 'pc, 'vm, 'state, 'li
...
+                let layout = vm
+                    .runtime_type_layout(&type_tag)
+                    .map_err(|e| self.env.convert_linked_vm_error(e, linkage))?;
                 let Some(bytes) = value.typed_serialize(&layout) else {
                     invariant_violation!("Failed to serialize Move event");
                 };
-                Ok((version_mid.clone(), tag, bytes))
+                let TypeTag::Struct(tag) = type_tag else {
+                    unreachable!()
+                };
```
cgswords, line 573: This code is reflexively true due to the construction just a few lines up. Not sure what it's doing.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -709,60 +809,38 @@ impl<'env, 'pc, 'vm, 'state, 'linkage, 'gas> Context<'env, 'pc, 'vm, 'state, 'li
...
-            args,
-            linkage,
-            tracer,
-        )
+        with_vm!(self, &function.linkage, |vm: &mut MoveVM<'env>| {
+            let ty_args = function
+                .type_arguments
+                .iter()
+                .map(|ty| {
+                    let tag: TypeTag = ty.clone().try_into().map_err(|e| {
```
cgswords, line 819: I wonder if we shouldn't decorate these functions with typetag-based arguments earlier.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/analysis.rs
```diff
@@ -171,3 +183,138 @@ impl LinkageAnalyzer {
...
+                let tag: StructTag = ty.clone().into();
+                let ids = tag.all_addresses().into_iter().map(ObjectID::from);
+                resolution_table.add_type_linkages_to_table(ids, package_store)?;
+            }
+            CallArg::FundsWithdrawal(f) => {
+                let FundsWithdrawalArg { type_arg, .. } = f;
+                match type_arg {
+                    WithdrawalTypeArg::Balance(inner) => {
+                        let Ok(tag) = inner.to_type_tag() else {
+                            invariant_violation!(
```
cgswords: Not a fan of `invariant_violation`, as mentioned in the other PR.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/analysis.rs
```diff
@@ -171,3 +183,138 @@ impl LinkageAnalyzer {
...
+                        resolution_table.add_type_linkages_to_table(ids, package_store)?;
+                    }
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    fn command(
```
cgswords, line 271: I assume @tnowacki  already okayed this?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -517,17 +559,75 @@ impl<'env, 'pc, 'vm, 'state, 'linkage, 'gas> Context<'env, 'pc, 'vm, 'state, 'li
...
+        I: IntoIterator<Item = MoveObjectType>,
+    {
+        let tys_addrs = writes
+            .into_iter()
+            .flat_map(|ty| StructTag::from(ty).all_addresses())
+            .map(ObjectID::from)
+            .collect::<BTreeSet<_>>();
+
+        let ty_linkage = ExecutableLinkage::type_linkage(&tys_addrs, env.linkable_store)?;
+        env.vm
```
cgswords, line 603: Shouldn't we check the LRU first?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
```diff
@@ -288,16 +242,15 @@ impl<'pc, 'vm, 'state, 'linkage> Env<'pc, 'vm, 'state, 'linkage> {
             })?;
         let runtime_signature = subst_signature(runtime_signature, &loaded_type_arguments)
             .map_err(|e| self.convert_linked_vm_error(e, &linkage))?;
-        let vm_opt = Some(vm);
         let parameters = runtime_signature
             .parameters
             .into_iter()
-            .map(|ty| self.adapter_type_from_vm_type(&vm_opt, &ty))
+            .map(|ty| self.adapter_type_from_vm_type(&vm, &ty))
```
cgswords, line 248: The lack of uniform interface here troubles me a bit. Not that we are doing it wrong, but some of these methods take a VM and some (like `fully_annotated_layout`) do not. Would it be prudent to take a VM in every case, and then add explicit `fully_annotated_layout_with_input_vm` or similar? Let someone (me) accidentally calls `fully_annotated_layout` not realizing their mistake?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
tzakian: Twofold:
1. We need/want mutable access to the execution context (and need to grab mutable VMs out of its VM cache).
2. The VMs differ in their native contexts -- the VMs in the execution context have the full native context extensions (which are created after the env is created) as they need to be able to execute code, whereas the resolution VM has an empty set of native extensions since all it is used for is type resolution.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
tzakian: Just trying to be fancy and avoid a clone... (we needed to box the struct tag into a type tag for a call into the VM, and we now want that inner struct tag back, so to avoid cloning we just destructure it at this point)

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
tzakian: We do (ish) -- we decorate with an adapter `Type` which is basically a more copyable `TypeTag` with some additional information that useful for typechecking (e.g., abilities and the like).

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/analysis.rs
tzakian: ah...different type of invariant violation here (this is a Sui/adapter invariant violation). I'll let you and Todd duke it out on that one :p

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
tzakian: The likelihood of finding one is probably fairly low, but doesn't hurt to do that. Will add the check

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
tzakian: I think it's the exception and not the rule that the env should take the vm -- basically this is only needed in cases where we are calling into the VM from execution somewhere in a place where values/events may have been created from newly published code (i.e., for events, or for final writeout).

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
tzakian: The likelihood of being able to reuse the input resolution VM is significantly higher (and becomes even higher if we allow subset behavior -- but I pinged you on Slack to discuss this).

PR #24952, opened 2026-01-14
[move] Run bytecode verifier transactional tests with the new regex reference safety verifier

tnowacki:
## Description 

- Generated alternative snapshots for the bytecode verifier tests that utilize exclusively the new reference safety verifier

## Test plan 

- The PR is tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/bytecode-verifier-tests/src/unit_tests/mod.rs
+++ b/external-crates/move/crates/bytecode-verifier-tests/src/unit_tests/mod.rs
@@ -4,8 +4,8 @@
 
 use move_binary_format::file_format_common::VERSION_MAX;
 use move_vm_config::verifier::{
-    MeterConfig, VerifierConfig, DEFAULT_MAX_CONSTANT_VECTOR_LEN, DEFAULT_MAX_IDENTIFIER_LENGTH,
-    DEFAULT_MAX_VARIANTS,
+    DEFAULT_MAX_CONSTANT_VECTOR_LEN, DEFAULT_MAX_IDENTIFIER_LENGTH, DEFAULT_MAX_VARIANTS,
+    MeterConfig, VerifierConfig,
 };
 
 pub mod binary_samples;


external-crates/move/crates/bytecode-verifier-transactional-tests/tests/reference_safety/expressivity/alias_writes@regex.snap
```diff
@@ -0,0 +1,4 @@
+---
+source: crates/move-transactional-test-runner/src/framework.rs
+---
+processed 4 tasks
```
tzakian, line 4: fancy!

Does this also work if you write them out of declaration order too? (I'm guessing not?)

external-crates/move/crates/move-vm-config/src/verifier.rs
```diff
@@ -36,6 +36,8 @@ pub struct VerifierConfig {
     pub sanity_check_with_regex_reference_safety: Option</* meter limit */ u128>,
     pub deprecate_global_storage_ops: bool,
     pub disable_entry_point_signature_check: bool,
+    /// If true, the very will run only the regex reference safety check.
```
cgswords:
```suggestion
    /// If true, the verifier will run only the regex reference safety check.
```

external-crates/move/crates/move-bytecode-verifier/src/code_unit_verifier.rs
```diff
@@ -179,63 +179,73 @@ impl<'env> CodeUnitVerifier<'env, '_> {
...
-            let regex_res = regex_reference_safety::verify(
+        if verifier_config.switch_to_regex_reference_safety {
+            regex_reference_safety::verify(
+                verifier_config,
+                self.module,
+                &self.function_context,
+                meter,
+            )
+        } else {
+            let reference_safety_res = reference_safety::verify(
```
cgswords: It might be easier to follow if this whole body was pulled out into a helper.

external-crates/move/crates/bytecode-verifier-transactional-tests/tests/reference_safety/expressivity/alias_writes@regex.snap
tnowacki: Of course it works then too! Why wouldn't it?

tnowacki: An alias is an alias

tnowacki: The test also shows all the different orderings are allowed :)

tzakian: ah! And indeed it does..missed those for some reason when looking at this the first time 🎉

PR #24955, opened 2026-01-15
Fix object scheduler bug by reading exact versions

mystenmark:


crates/sui-core/src/execution_scheduler/execution_scheduler_impl.rs
```diff
@@ -711,6 +712,7 @@ impl ExecutionScheduler {
         if effects.status().is_err() {
             // This transaction already failed. It does not matter any more
             // whether it has sufficient object funds or not.
+            debug!("Transaction failed, committing effects");
```
lxfind, line 715: Should include tx_digest and maybe a bit more text. ditto for other debug logs

lxfind (approved):

crates/sui-core/src/execution_scheduler/execution_scheduler_impl.rs
mystenmark: the span includes the digest

PR #24958, opened 2026-01-15
[move][move-vm][bella-ciao] Swap out caches

cgswords:
## Description 

See title

## Test plan 

No testing changes, plus Tim will verify perf offline.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
@@ -89,18 +89,24 @@ impl MoveCache {
         verified: verification::ast::Package,
         runtime: jit::execution::ast::Package,
     ) -> bool {
-        // Check if the package is already present, and if so, return false early to avoid
-        // re-inserting the already-cached package.
-        if self.package_cache.contains_key(&package_key) {
-            return false;
+        use dashmap::mapref::entry::Entry;
+        // Grab the entry at the top, so we can figure out which flag to return while holding the
+        // lock on the shard of dashmap we are modifying (so this does not change out from under us
+        // mid-write).
+        let entry = self.package_cache.entry(package_key);
+        match entry {
+            Entry::Occupied(_) => {
+                // Package is already present.
+                false
+            }
+            Entry::Vacant(vacant_entry) => {
+                let verified = Arc::new(verified);
+                let runtime = Arc::new(runtime);
+                let package = Package { verified, runtime };
+                vacant_entry.insert(Arc::new(package));
+                true
+            }
         }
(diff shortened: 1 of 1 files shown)


external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
```diff
@@ -90,44 +89,38 @@ impl MoveCache {
...
         // re-inserting the already-cached package.
-        if package_cache.contains_key(&package_key) {
+        if self.package_cache.contains_key(&package_key) {
             return false;
         }
         let verified = Arc::new(verified);
         let runtime = Arc::new(runtime);
         let package = Package { verified, runtime };
-        package_cache.insert(package_key, Arc::new(package));
+        self.package_cache.insert(package_key, Arc::new(package));
```
tzakian: I'm not sure if I'm sold here. This could lead to eviction of a previously compiled and cached package, which seems wrong to me.

cgswords: Swapped to use `entry`.

external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
```diff
@@ -90,44 +89,42 @@ impl MoveCache {
         verified: verification::ast::Package,
         runtime: jit::execution::ast::Package,
     ) -> bool {
-        // NB: We grab a write lock here to ensure that we don't double-insert a package.
-        let mut package_cache = self.package_cache.write();
-
         // Check if the package is already present, and if so, return false early to avoid
         // re-inserting the already-cached package.
-        if package_cache.contains_key(&package_key) {
+        if self.package_cache.contains_key(&package_key) {
```
tzakian: Sorry to be so nitty on this function...

The return value of this function is incorrect now, even with the `entry` since if it gets cached between the check and insertion, it could not be present on this check, but then at line 100 the `entry` would be filled and we wouldn't insert the new package, but nevertheless return `true`. 

We only use this boolean to track cache loads I think, so it's not a big deal, but might be worth either calling out in a comment, or removing the `bool` return entirely.

tzakian (approved):
Looks great!

external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
cgswords: Fixed here and the other one.

PR #24960, opened 2026-01-15
fix: stop commit syncer after core shutdown

SashaMalysehko:
Ensure CommitSyncer exits when CoreThreadDispatcher rejects certified commits. This aligns the shutdown log with actual behavior and avoids continued fetch work after core termination.

### DIFF
--- a/consensus/core/src/commit_syncer.rs
+++ b/consensus/core/src/commit_syncer.rs
@@ -171,7 +171,10 @@ impl<C: NetworkClient> CommitSyncer<C> {
                         return;
                     }
                     let (target_end, commits) = result.unwrap();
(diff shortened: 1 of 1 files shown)


consensus/core/src/commit_syncer.rs
```diff
@@ -171,7 +171,10 @@ impl<C: NetworkClient> CommitSyncer<C> {
                         return;
                     }
                     let (target_end, commits) = result.unwrap();
-                    self.handle_fetch_result(target_end, commits).await;
+                    if !self.handle_fetch_result(target_end, commits).await {
+                        self.inflight_fetches.shutdown().await;
```
mwtian, line 175: Maybe add an info log?

PR #24963, opened 2026-01-15
Expose address balances from consistent service up through graphql

wlmyng:
## Description 

New `Balance.address_balance` and `Balance.coin_balance` fields to reflect the two possible sources of balance for an address. `Balance.total_balance` will return the sum of address and coin balances.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: `Balance.totalBalance` now returns the sum of balances from owned coins and from the accumulator object. The individual coin or address balances can be retrieved through `Balance.coinBalance` and `Balance.addressBalance` respectively. For the previous behavior, select the `Balance.coinBalance` field for coin balances only.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/balances.snap
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/balances.snap
@@ -63,7 +63,7 @@ Response: {
           },
           {
             "coinType": {
-              "repr": "0x871b5b72dbbd9a32fcba584fadeea5df73efc78541fb1f89f4da918b67a6b827::test::TEST"
+              "repr": "0xbaa3060844f46b8ac670ce4945a2abc6aae36b9d9514d51bde88bec13447bf1c::test::TEST"
             },
             "totalBalance": "1002000",
             "coinBalance": "2000",


emmazzz (commented):
The core changes look good to me. But could we add a transactional test? Nothing fancy, just a simple one validating that the API works.

crates/sui-indexer-alt-graphql/src/api/types/balance.rs
```diff
@@ -6,6 +6,7 @@ use async_graphql::SimpleObject;
 use async_graphql::connection::Connection;
 use async_graphql::connection::CursorType;
 use async_graphql::connection::Edge;
+use sui_indexer_alt_reader::consistent_reader::Balance as StoredBalance;
```
amnn: nit: when I see `Stored*` it makes me think of the types we get out from diesel. I'm wondering if there's a different name/convention we can come up with here? What I'm tempted to suggest is to avoid the two step conversion (from `proto::Balance` to the new `Balance` type, to the GraphQL `Balance` type), and have `ConsistentReader` produces pages of `proto::Balance`, and have a `try_from_proto(balance: proto::Balance, scope: Scope) -> RpcError<Self>` here?

amnn (approved):
Looks good -- minor suggestion about avoiding an intermediate type, thanks @wlmyng !

crates/sui-indexer-alt-graphql/src/api/types/balance.rs
wlmyng: right, that's reasonable

PR #24969, opened 2026-01-16
[move][move-compiler] Fix module name resolution crash, plus replace several unwraps with expect in the compiler

cgswords:
## Description 

This adds a name map where there should have been one in expansion, which caused a crash before.

## Test plan 

New tests -- and they produce a number of errors instead of panicking now.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-compiler/src/expansion/translate.rs
```diff
@@ -990,12 +998,13 @@ fn into_modules_and_extenions(
...
                 let addr = top_level_address_(
                     &mut context.defn_context,
-                    na_map,
+                    na_map.clone(),
                     /* suggest_declaration */ false,
                     addr,
                 );
 
-                modules
+                context.defn_context.named_address_mapping = Some(na_map);
```
cgswords, line 1006: This is the actual fix.

external-crates/move/crates/move-compiler/src/expansion/translate.rs
```diff
@@ -990,12 +998,13 @@ fn into_modules_and_extenions(
                 }
                 let addr = top_level_address_(
                     &mut context.defn_context,
-                    na_map,
+                    na_map.clone(),
```
tnowacki, line 1001: This is an arc right?

tnowacki (approved):

external-crates/move/crates/move-compiler/src/expansion/translate.rs
cgswords: Yep!

PR #24977, opened 2026-01-16
Remove allow deprecated

akmysti:
## Description 

## Test plan 

cargo move-clippy is clean.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/dev_utils/dbg_print.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/dev_utils/dbg_print.rs
@@ -4,11 +4,11 @@
 // Debug print code for development. Remove when moving into production mode.
 pub const DEBUG_PRINT: bool = false;
 
-#[allow(dead_code)]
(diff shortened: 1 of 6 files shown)


tzakian (commented):
Not sure if all of these should be removed? If I run `cargo move-clippy` on this I get a bunch of unused warnings. IMO we should either keep the `#[(allow(dead_code)]` on it, or even better remove the fields/functions (although we should discuss the ones that are being removed in that case).

akmysti:
> Not sure if all of these should be removed? If I run `cargo move-clippy` on this I get a bunch of unused warnings. IMO we should either keep the `#[(allow(dead_code)]` on it, or even better remove the fields/functions (although we should discuss the ones that are being removed in that case).

I added back the attributes on the ones that are dead.

PR #24981, opened 2026-01-16
jsonrpc: add index to track existence of address balance for a coin type

bmwill:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/accumulators/balances.rs
+++ b/crates/sui-core/src/accumulators/balances.rs
@@ -1,19 +1,10 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use mysten_common::debug_fatal;
 use sui_types::{
-    TypeTag,
-    accumulator_metadata::AccumulatorOwner,
-    accumulator_root::AccumulatorValue,
-    balance::Balance,
-    base_types::{ObjectID, SuiAddress},
-    error::SuiResult,
-    storage::error::Error as StorageError,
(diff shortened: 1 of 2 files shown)


crates/sui-core/src/jsonrpc_index.rs
```diff
@@ -789,6 +793,15 @@ impl IndexStore {
...
         let cache_updates = self.index_coin(digest, &mut batch, &object_index_changes, tx_coins)?;
 
+        // update address balances index
+        let address_balance_updates = accumulator_events.into_iter().filter_map(|event| {
+            let ty = &event.write.address.ty;
+            // Only process events with Balance<T> types
+            let coin_type = sui_types::balance::Balance::maybe_get_balance_type_param(ty)?;
+            Some(((event.write.address.address, coin_type), ()))
+        });
+        batch.insert_batch(&self.tables.address_balances, address_balance_updates)?;
```
bmwill, line 803: @mystenmark I did this slightly different from how you suggested by just looking at the accumulator events directly instead. This should be sufficient right?

mystenmark, line 803: i don't think this works because it will never delete anything. If you index the settlement txn you will be able to observe deletions

bmwill: Yeah thats true, jsonrpc indexes already grow unbounded and this likely wont add to that much over the next few months till we remove it. And its something that we can hook in pruning for if we find it to be an issue

PR #24999, opened 2026-01-17
simtest workload improvements

mystenmark:
- Unify soft_bundle and batch payload types. 
- Allow batches to be split into one more smaller soft-bundles.
- Submit transactions to random validators

### DIFF
--- a/crates/sui-benchmark/src/drivers/bench_driver.rs
+++ b/crates/sui-benchmark/src/drivers/bench_driver.rs
@@ -1082,7 +1082,7 @@ async fn run_bench_worker(
                                             let (_, exec_result) =
                                                 proxy.execute_transaction_block(tx).await;
                                             exec_result.map(|effects| {
-                                                vec![(digest, BundleItemResponse::DirectEffects(effects))]
+                                                vec![(digest, BundleItemResponse::DirectEffects(effects.into()))]
                                             })
                                         } else {
                                             proxy.execute_soft_bundle(bundle_txs).await.map(|results| {
@@ -1223,7 +1223,7 @@ async fn run_bench_worker(
 /// (WaitForEffectsResponse) or direct execution (ExecutionEffects).
 enum BundleItemResponse {
     WaitForEffects(WaitForEffectsResponse),
-    DirectEffects(ExecutionEffects),
+    DirectEffects(Box<ExecutionEffects>),
 }
 
 type BundleResults = Vec<(
@@ -1256,9 +1256,7 @@ fn process_bundle_results(
                                 payload,
                                 effects.status()
                             );
-                            BatchedTransactionStatus::Success {
-                                effects: Box::new(effects),
-                            }
+                            BatchedTransactionStatus::Success { effects }
                         }
                         BundleItemResponse::WaitForEffects(wait_response) => match wait_response {
                             WaitForEffectsResponse::Executed { details, .. } => {


crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -372,45 +372,68 @@ impl Payload for CompositePayload {
     }
 
     fn make_transaction(&mut self) -> Transaction {
+        self.make_soft_bundle_transactions().pop().unwrap()
+    }
+
+    fn is_soft_bundle(&self) -> bool {
+        true
```
alex-mysten: Do we lose anything by enabling this unconditionally?

crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -372,45 +372,68 @@ impl Payload for CompositePayload {
     }
 
     fn make_transaction(&mut self) -> Transaction {
+        self.make_soft_bundle_transactions().pop().unwrap()
```
alex-mysten: This would be unreachable, correct?

crates/sui-benchmark/src/drivers/bench_driver.rs
```diff
@@ -996,202 +1040,45 @@ async fn run_bench_worker(
...
-                        futures.push(Box::pin(res));
-                    } else if payload.is_concurrent_batch() {
-                        // Concurrent batch: submit multiple transactions separately but concurrently
-                        let txs = payload.make_concurrent_transactions();
-                        let num_txs = txs.len();
-                        let start = Arc::new(Instant::now());
-                        let metrics_clone = Arc::clone(&metrics);
-                        let proxy = worker.proxy.clone();
+                        // Partition transactions into random bundles
+                        let bundles = partition_into_random_bundles(txs, max_bundles);
```
alex-mysten, line 1053: Seems misleading that transactions from make_soft_bundle_transactions() are not necessarily included in the same soft bundle

mystenmark: yeah i think i will rename the whole concept to just be "batch" or something

crates/sui-benchmark/src/workloads/composite/mod.rs
mystenmark: yes, fixed in #25052

crates/sui-benchmark/src/workloads/composite/mod.rs
mystenmark: i don't think so. it can and will generate batches of size 1, so it should be a superset of the behavior of the previous code

crates/sui-benchmark/src/workloads/payload.rs
```diff
@@ -85,6 +77,12 @@ pub trait Payload: Send + Sync + std::fmt::Debug + Display {
         false // Default: not a soft bundle
     }
 
+    /// Returns the maximum number of bundles that can be created for a batch of transactions.
+    /// If set to 1, all transactions will always be executed as a single bundle.
+    fn max_bundles(&self) -> NonZeroUsize {
```
mwtian: The change makes understanding the logic more confusing to me. We should consider clarifying and renaming some of the methods, for example
is_soft_bundle() -> Maybe name this use_soft_bundle() and only set it in the specific workload, or remove the method since it always returns the same value.
max_bundles(): mention that the batch of transactions will be randomly divided into 1..=max_bundles number of bundles.
handle_soft_bundle_results() -> handle_batch_results() // or something related.

crates/sui-benchmark/src/drivers/bench_driver.rs
```diff
@@ -996,202 +1040,45 @@ async fn run_bench_worker(
...
-                            let futures: Vec<_> = txs
+                            // Submit each bundle concurrently
+                            let bundle_futures: Vec<_> = bundles
                                 .into_iter()
-                                .map(|tx| {
-                                    let proxy = proxy.clone();
-                                    async move { proxy.execute_transaction_block(tx).await }
+                                .map(|(start_idx, bundle_txs)| {
+                                    let proxy = proxy.clone_new();
+                                    let bundle_size = bundle_txs.len();
```
mwtian: Can you also test not using soft bundle submission when bundle_size is 1?

crates/sui-benchmark/src/workloads/composite/mod.rs
mwtian: If this affects how the request is specified, this does loose the coverage for non soft bundle transactions.

mwtian:
Btw, is the plan to migrate most workloads (e.g. randomized transaction, randomness, etc) to CompositeWorkload, and only leave payment / counter / batched payment for perf testing?

crates/sui-benchmark/src/drivers/bench_driver.rs
mystenmark: done.

crates/sui-benchmark/src/workloads/payload.rs
mystenmark: i renamed everything, PTAL

mystenmark:
> Btw, is the plan to migrate most workloads (e.g. randomized transaction, randomness, etc) to CompositeWorkload, and only leave payment / counter / batched payment for perf testing?

That seems reasonable to me, yes. We'll go bit by bit, ensuring that we aren't losing coverage.

crates/sui-benchmark/src/drivers/bench_driver.rs
```diff
@@ -993,205 +1039,73 @@ async fn run_bench_worker(
...
-                                    async move { proxy.execute_transaction_block(tx).await }
+                                .map(|(start_idx, bundle_txs)| {
+                                    let proxy = proxy.clone_new();
+                                    let digests: Vec<TransactionDigest> =
+                                        bundle_txs.iter().map(|tx| *tx.digest()).collect();
+                                    async move {
+                                        info!("executing bundle {:?}", digests);
+                                        // For single-transaction bundles, randomly use execute_transaction_block
+                                        let result = if bundle_txs.len() == 1
+                                            && rand::thread_rng().gen_bool(0.5)
```
mwtian, line 1078: I think we should give TransactionOrchestrator & TransactionDriver good coverage, so maybe with probably 0.9 or remove this condition, since softbundle should already have enough coverage from the randomized transaction workload.
The issue with execute_soft_bundle() is that it doesn't wait for a quorum of responses, which can mask issues.

mystenmark: Ok, we can add that. For now I'm more concerned with exposing as many possible orderings of transactions in mysticeti blocks, hence this approach. While we want to test TransactionOrchestrator/TransactionDriver, they can't cause safety issues so are a lower priority for me.

mwtian: I don't see how the changes here test more ordering of transactions.

PR #25000, opened 2026-01-17
[move-compiler] fix borrow checker invariant violation

tnowacki:
## Description 

- Fix invariant violation 
- The error message is... a mess but this is an edge case that is going away in the new borrow checker so I'm not sure it is worth too much time fixing right now 

## Test plan 

- New test 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-compiler/src/cfgir/borrows/state.rs
+++ b/external-crates/move/crates/move-compiler/src/cfgir/borrows/state.rs
@@ -380,12 +380,25 @@ impl BorrowState {
         );
     }
 
-    fn local_borrowed_by(&self, local: &Var) -> (BTreeMap<RefID, Loc>, BTreeMap<RefID, Loc>) {
+    /// returns full borrows of the root and any borrows specific to the specified `local`
+    fn root_borrowed_by(&self, local: &Var) -> (BTreeMap<RefID, Loc>, BTreeMap<RefID, Loc>) {
         let (full_borrows, mut field_borrows) = self.borrows.borrowed_by(Self::LOCAL_ROOT);
         let local_borrows = field_borrows
             .remove(&Self::local_label(local))
             .unwrap_or_default();
         (full_borrows, local_borrows)
+    }
+
+    /// Returns specific borrows on the `local`,
+    /// Also includes any full borrows on the root that might have occurred via the "overflow"
+    /// optimization in the borrow graph
+    fn local_borrowed_by(&self, local: &Var) -> BTreeMap<RefID, Loc> {
+        let (mut full_borrows, mut field_borrows) = self.borrows.borrowed_by(Self::LOCAL_ROOT);
+        let local_borrows = field_borrows
+            .remove(&Self::local_label(local))
+            .unwrap_or_default();
+        full_borrows.extend(local_borrows);
+        full_borrows
     }
 
     // returns empty errors if borrowed_by is empty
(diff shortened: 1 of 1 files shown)


external-crates/move/crates/move-compiler/src/cfgir/borrows/state.rs
```diff
@@ -502,11 +503,12 @@ impl BorrowState {
         let mut diags = Diagnostics::new();
         for (local, stored_value) in self.locals.key_cloned_iter() {
             if let Value::NonRef = stored_value {
-                let borrowed_by = self.local_borrowed_by(&local);
+                let (mut full_borrows, local_borrows) = self.local_borrowed_by(&local);
+                full_borrows.extend(local_borrows);
```
cgswords: This being repeated all of these places feels like a bit of an anti-pattern; could we add a different function alongside `local_borrowed_by` that does this extension and hands back a single value? I would hate to mess this concatenation up in a single place.

external-crates/move/crates/move-compiler/src/cfgir/borrows/state.rs
```diff
@@ -668,26 +672,24 @@ impl BorrowState {
...
-                local,
-                &mut_borrows,
-                ReferenceSafety::RefTrans,
-                "borrow",
-            )
-            .into()
-        } else {
-            Diagnostics::new()
+                .filter(|(id, _loc)| borrows.is_mutable(*id));
+            full_borrows.extend(mut_borrows);
```
cgswords, line 691: Why not also filter `full_borrows`? Because they may also contain mutable borrows as a result of filling?

external-crates/move/crates/move-compiler/tests/move_2024/borrows/borrow_owned_enum.snap
```diff
@@ -0,0 +1,214 @@
...
+17 │     x = &y;
+   │         -- It is still being borrowed by this reference
+   ·
+33 │     if (test) x = p;
+   │         ^^^^
+   │         │
+   │         Ambiguous usage of variable 'test'
+   │         Try an explicit annotation, e.g. 'move test' or 'copy test'
+   │
+   = Ambiguous inference of 'move' or 'copy' for a borrowed variable's last usage: A 'move' would invalidate the borrowing reference, but a 'copy' might not be the expected implicit behavior since this the last direct usage of the variable.
```
cgswords, line 20: This error is very hard to follow---perhaps it is receiving the wrong location?

tnowacki: It's not... its just producing an error for every possible branch. This particular error/limitation is going away with the new borrow checker, so I'm not sure it is worth spending much time fixing.

external-crates/move/crates/move-compiler/src/cfgir/borrows/state.rs
tnowacki: Because you don't know which local the full borrow is actually rooted in. So in a way, it would be extending the full borrow. This is the same cause of the funky code in copy loc.

But again, luckily this is all going away at some point

PR #25001, opened 2026-01-17
Always accept system transactions

mwtian:
## Description 

Do not skip system transactions even when they are rejected. System transactions can be rejected due to GC even when there is no reject vote against the transactions. But they should still be processed by consensus handler in this case.

Remove consensus block handler.

## Test plan 

CI

### DIFF
--- a/consensus/core/src/commit.rs
+++ b/consensus/core/src/commit.rs
@@ -20,6 +20,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     block::{BlockAPI, Slot, VerifiedBlock},
+    context::Context,
     leader_scoring::ReputationScores,
     storage::Store,
 };
@@ -379,6 +380,8 @@ pub struct CommittedSubDag {
     ///
     /// Indices of rejected transactions in each block.
     pub rejected_transactions_by_block: BTreeMap<BlockRef, Vec<TransactionIndex>>,
+    /// Used by consensus to communicate whether to always accept system transactions in this commit.
+    pub always_accept_system_transactions: bool,
 }
 
 impl CommittedSubDag {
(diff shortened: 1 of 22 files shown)


crates/sui-core/src/consensus_manager/mod.rs
```diff
@@ -157,23 +155,15 @@ impl ConsensusManager {
         let replay_after_commit_index =
             last_processed_commit_index.saturating_sub(num_prior_commits);
 
-        let (commit_consumer, commit_receiver, block_receiver) =
+        let (commit_consumer, commit_receiver, _block_receiver) =
```
akichidis: Why not remove the `_block_receiver` completely?

crates/sui-core/src/consensus_handler.rs
akichidis, line 3466: bye bye 👋

consensus/core/src/commit.rs
```diff
@@ -379,6 +380,8 @@ pub struct CommittedSubDag {
     ///
     /// Indices of rejected transactions in each block.
     pub rejected_transactions_by_block: BTreeMap<BlockRef, Vec<TransactionIndex>>,
+    /// Used by consensus to communicate whether to always accept system transactions in this commit.
+    pub always_accept_system_transactions: bool,
```
akichidis, line 384: Although I see why you added here, it seems to me that this is a decision of interpretation that consensus it self doesn't have much of a saying to it. It's literally the otherwise that interprets it that way. Do you believe that alternatively the `always_accept_system_transactions` config protocol parameter could be provided directly to the `transactions()` method when reading the commited sub dag on the consensus handler?

crates/sui-core/src/consensus_manager/mod.rs
mwtian: Good point

consensus/core/src/commit.rs
mwtian: Adding this to transactions() method will create a lot of churn in the codebase. This option will be removed after rolling out and CommittedSubDag is not persisted, so this temporary change should be ok.

PR #25003, opened 2026-01-17
[move-stlib] Checked and lossless helpers for << and >>

tnowacki:
## Description 

- Adds helpers for shift left and shift right
- The << and >> abort if the rhs is too big for the bitsize, e.g. 63 is fine for u64, but 64 aborts. 
- As such, `checked_shl` and `checked_shr` return `option::none` iff the rhs is too large. 
  - There are no `checked_*` variants for `u256` since the rhs is always valid 
- `lossless_shl` and `lossless_shr` return `option::some` iff the operation is reversible (and in that sense lossless). That is,  if `(x << y) >> y == x` (and similarly for `>>`.
  - Note this means they also return `none` if the rhs is too large 

- We will use the `lossless_` prefix for integer division that does not round (when the remainder is 0) 

## Test plan 

- Added tests 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/balances.snap
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/balances.snap
@@ -63,7 +63,7 @@ Response: {
           },
           {
             "coinType": {
-              "repr": "0xbaa3060844f46b8ac670ce4945a2abc6aae36b9d9514d51bde88bec13447bf1c::test::TEST"
+              "repr": "0x37d5b103ff1b7d4413ea820b3b0736f9073c54d2004b75405e76964be215a526::test::TEST"
             },
             "totalBalance": "1002000",
             "coinBalance": "2000",


crates/sui-framework/packages/move-stdlib/sources/u128.move
```diff
@@ -129,6 +129,32 @@ public fun saturating_mul(x: u128, y: u128): u128 {
     std::macros::num_saturating_mul!(x, y, max_value!())
 }
 
+/// Shifts `x` left by `shift` bits.
+/// Returns `None` if the shift is greater than or equal to the bit size of 128.
+public fun checked_shl(x: u128, shift: u8): Option<u128> {
+    std::macros::num_checked_shl!(x, shift, 128)
```
tzakian, line 135: Thoughts on adding a macro for integers called `num_bits!()` and/or `num_bytes!()` that gives this number/number of bytes for the number?

crates/sui-framework/packages/move-stdlib/tests/integer_tests.move
```diff
@@ -515,3 +515,71 @@ public(package) macro fun test_saturating_mul<$T>($max: $T, $cases: vector<$T>)
...
+    let max = $max;
+    let bit_size = $bit_size;
+    // exact shifts (no bits lost)
+    assert_eq!((0: $T).lossless_shl(0), option::some(0));
+    assert_eq!((0: $T).lossless_shl(bit_size - 1), option::some(0));
+    assert_eq!((1: $T).lossless_shl(0), option::some(1));
+    assert_eq!((1: $T).lossless_shl(1), option::some(2));
+    assert_eq!((1: $T).lossless_shl(bit_size - 1), option::some(1 << (bit_size - 1)));
+    // inexact shifts (bits lost)
+    assert_eq!((1: $T).lossless_shl(bit_size), option::none());
```
tzakian, line 562: Having a test for `0.lossless_shl(bit_size)` (and for `lossless_shr` as well) might be useful too (to test the boundary checking)

crates/sui-framework/packages/move-stdlib/sources/u128.move
tnowacki: We don't have `$crate` style invocations, if that is what you are after?

tnowacki: Or do you just want these values in constants?

tnowacki: Or as public macros??

tzakian: I was thinking of as public macros in a similar vein to `max_value` above?

PR #25009, opened 2026-01-18
Fix important typo in docs - Anna to Alice.

PrisionMike:
## Description 

The wrong name breaks the flow of someone trying to understand how transactions work.

## Test plan  
  
Re-read the example and diagram. The change make more sense.

---

## Release notes

N/A


jessiemongeon1:
Thanks @PrisionMike , we have included this fix in a larger batch update PR.

PR #25014, opened 2026-01-19
fix: pg_isready error message

Forostovec:
Cause:
The error message in pg_isready referenced pg_ctl, which is a different command and misleads diagnostics when pg_isready is missing.

Change:
Updated the error text to correctly reference pg_isready, improving clarity for troubleshooting.


amnn (approved):
Looks good, thanks @Forostovec !

PR #25016, opened 2026-01-20
Fix --sender flag ignored in publish/upgrade commands

tuky191:
## Description

Fixes a bug where the `--sender` flag was ignored when using `--serialize-unsigned-transaction` with `sui client publish` and `sui client upgrade` commands.

The `TxProcessingArgs::sender` field is documented as:

> Set the transaction sender to this address. **When not specified, the sender is inferred by finding the owner of the gas payment.**

However, `publish_command` and `upgrade_command` were always inferring the sender from gas objects, never checking `processing.sender`. This caused the UpgradeCap to be transferred to the inferred address instead of the specified sender.

This was problematic for workflows where:
- The transaction is serialized for later signing by a different party
- The signer address differs from the address owning the gas objects

The fix makes the implementation match the documented behavior: check `processing.sender` first and only fall back to gas inference if not specified.

## Test plan

Added `test_publish_sender_flag_respected_in_serialized_transaction` which:
1. Creates a test cluster with two addresses (address_0 as active, address_1 as specified sender)
2. Calls `TestPublish` with `--serialize-unsigned-transaction` and `--sender address_1`
3. Verifies the transaction sender is address_1
4. Verifies the UpgradeCap recipient in the PTB is address_1, not address_0

```bash
cargo test -p sui --test cli_tests test_publish_sender_flag_respected_in_serialized_transaction
```

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: The `--sender` flag is now correctly respected in `sui client publish` and `sui client upgrade` commands when used with `--serialize-unsigned-transaction`. Previously, the sender was incorrectly inferred from gas objects, ignoring the `--sender` flag.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/src/client_commands.rs
+++ b/crates/sui/src/client_commands.rs
@@ -3503,7 +3503,9 @@ async fn publish_command(
         processing,
     } = args;
 
-    let sender = processing.sender.unwrap_or(context.infer_sender(&payment.gas).await?);
+    let sender = processing
+        .sender
+        .unwrap_or(context.infer_sender(&payment.gas).await?);
     let client = context.get_client().await?;
     let read_api = client.read_api();
     let chain_id = read_api.get_chain_identifier().await?;
@@ -3587,7 +3589,9 @@ async fn upgrade_command(
         processing,
     } = args;
 
-    let sender = processing.sender.unwrap_or(context.infer_sender(&payment.gas).await?);
+    let sender = processing
+        .sender
+        .unwrap_or(context.infer_sender(&payment.gas).await?);
     let client = context.get_client().await?;
     let read_api = client.read_api();
     let chain_id = read_api.get_chain_identifier().await?;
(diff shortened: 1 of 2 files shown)


stefan-mysten:
Thanks @tuky191, I'll do a few updates and try to merge asap.

stefan-mysten:
@tuky191 it appears that I cannot push on this PR. There's two things to fix
1. run `cargo fmt` in `crates/sui` folder and commit those changes
2. add 
```                
let sender = processing
    .sender
    .unwrap_or(context.infer_sender(&payment.gas).await?);
```
to `SuiClientCommands::SerializedTx {` and `SuiClientCommands::Call` cases as well.

Thanks again for your fix.

tuky191:
Hey @stefan-mysten, thanks for looking into this. I have fixed the formatting, the CI should pass now.

PR #25019, opened 2026-01-20
fix(bridge-cli): drop redundant committee RPC call

Galoretka:
Reason:
Before this change, committee() was called twice during CLI config load, overwriting the same address and causing an unnecessary extra RPC call that adds latency without any functional gain.

Summary:
Remove the redundant committee() RPC call and reuse the initially fetched committee proxy address when building the CLI config.


Bridgerz (approved):
LGTM

PR #25022, opened 2026-01-20
add doc for address aliases

aschran:
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
+---
+title: Address Aliases
+description: Address aliases can be used to configure the set of keys that are allowed to sign transactions for a particular address on the Sui network.
+keywords: [ address aliases, aliases, key rotation, changing keys ]
+---
+
+Address aliases can be used to configure the set of keys that are allowed to sign transactions for a particular address on the Sui network.
+
+One can use address aliases to:
```
jessiemongeon1:
```suggestion
You can use address aliases to:
```

docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
...
+* perform simple key rotation for an address.
```
jessiemongeon1:
```suggestion
- Perform simple key rotation for an address.
```

docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
...
+* grant multiple signers full access to everything owned by an address.
```
jessiemongeon1:
```suggestion
- Grant multiple signers full access to everything owned by an address.
```

docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
...
+* implement more complicated authentication schemes, such as smart wallets with passkeys.
```
jessiemongeon1:
```suggestion
- Implement more complicated authentication schemes, such as smart wallets with passkeys.
```

docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+All aliases for a given address have the ability to individually sign transactions acting as that address, and to control all resources owned by it. Be extremely careful when making any changes to your aliases.
+
+:::
+
+## How aliases work
+
+Each address on Sui has a set of aliases. New addresses begin as their own alias. That is, a new address `A` has the alias set `{A}`.
+
+Aliases for an address can be modified by using the Move API in the <UnsafeLink href="/references/framework/sui-framework/sui_sui/address_alias">`sui::address_alias` module</UnsafeLink>. These modification functions must be invoked as top-level PTB commands and cannot be called from other Move functions.
```
jessiemongeon1:
```suggestion
You can modify aliases for an address using the Move API in the <UnsafeLink href="/references/framework/sui-framework/sui_sui/address_alias">`sui::address_alias` module</UnsafeLink>. These modification functions must be invoked as top-level PTB commands and cannot be called from other Move functions.
```

docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+Some constraints for modifying aliases:
+* An address must always have at least one alias.
+* An address can be removed from its own alias set. For example, one can have an address `A` which only allows signatures from `{B, C}`.
+* There is a maximum number of allowed aliases for an address (currently 8).
```
jessiemongeon1:
```suggestion
- An address must always have at least one alias.
- An address can be removed from its own alias set. For example, you can have an address `A` which only allows signatures from `{B, C}`.
- The maximum number of allowed aliases for an address is 8.
```

docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+### Authenticating transactions with aliased addresses
+
+Transactions on Sui are configured with two addresses: a `sender` and a `gas_owner` (which may or may not be the same). For each configured address, a signature is required from one of the signers in its alias set.
```
jessiemongeon1:
```suggestion
Transactions on Sui are configured with two addresses: a `sender` and a `gas_owner` (which might or might not be the same). For each configured address, a signature is required from one of the signers in its alias set.
```

docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
...
+  ...
+}
+
+aliases[A] = {A, B}
+aliases[X] = {Y}
+```
+
+To be valid, the transaction `T` would require a signature from either `A` or `B`, *and* a signature from `Y`.
+
+The transaction "acts as" its `sender` for the purposes of execution. In the example above, even if the transaction had signatures from `[B, Y]` and not from `A`, during execution it would nonetheless have access to everything owned by `A`, and during Move call execution its `TxContext::sender` would be `A`.
```
jessiemongeon1:
```suggestion
The transaction acts as its `sender` for the purposes of execution. In this example, even if the transaction had signatures from `[B, Y]` and not from `A`, during execution it would nonetheless have access to everything owned by `A`, and during Move call execution its `TxContext::sender` would be `A`.
```

docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
...
+Address aliases can be used to configure the set of keys that are allowed to sign transactions for a particular address on the Sui network.
+
+You can use address aliases to:
+- Perform simple key rotation for an address.
+- Grant multiple signers full access to everything owned by an address.
+- Implement more complicated authentication schemes, such as smart wallets with passkeys.
+
+:::warning
+
+All aliases for a given address have the ability to individually sign transactions acting as that address, and to control all resources owned by it. Be extremely careful when making any changes to your aliases.
```
mystenmark: i would expand and add something like:

IF YOU ADD A NEW ADDRESS AS AN ALIAS FOR YOUR ADDRESS, ANYONE WITH THE KEYS FOR THAT ADDRESS CAN STEAL ALL YOUR FUNDS IRREVOCABLY

maybe surrounded by siren emojis

docs/content/guides/developer/transactions/address-aliases.mdx
```diff
@@ -0,0 +1,49 @@
...
+  ...
+}
+
+aliases[A] = {A, B}
+aliases[X] = {Y}
+```
+
+To be valid, the transaction `T` would require a signature from either `A` or `B`, *and* a signature from `Y`.
+
+The transaction acts as its `sender` for the purposes of execution. In this example, even if the transaction had signatures from `[B, Y]` and not from `A`, during execution it would nonetheless have access to everything owned by `A`, and during Move call execution its `TxContext::sender` would be `A`.
```
mystenmark: did you mean "The transaction sender acts as `sender`"? this isn't quite clear

aschran: I guess I was imprecisely trying to convey that for purposes of execution, the transaction acts as though `A` sent it, even though `A` did not sign it. in other words, after signing checks, nothing about the signatures is ever relevant again (I think there's no way for Move to know that it was actually signed by `B`?). if you have any suggestions how to better phrase this I'm open!

aschran: I just removed the first sentence and I think it still works fine and is hopefully less confusing

docs/content/guides/developer/transactions/address-aliases.mdx
aschran: I feel like that mostly repeats what I already wrote (minus the siren emojis)?

I rephrased to try to emphasize the ability to remove resources, idk if I want to use the word "steal" though...there are plenty of ways things can be stolen with or without aliases, and if the feature is really so dangerous that we need to use the word "steal" in a warning box at the top of its docs, then why are we adding it?

PR #25029, opened 2026-01-20
Report consistent service health status as NOT_SERVING until snapshots are available

wlmyng:
## Description 

The `add_service` method to register a new gRPC service now also expects a ready future. Until ready, the service will report NOT_SERVING.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-consistent-store/src/rpc/mod.rs
+++ b/crates/sui-indexer-alt-consistent-store/src/rpc/mod.rs
@@ -205,22 +205,27 @@ impl<'d> RpcService<'d> {
                 .await;
         }
 
+        // Create a Service to be attached as secondary to the main service
+        let mut readiness_checks = Service::new();
+
         for (name, ready) in service_futures {
             health_reporter
                 .set_service_status(name, ServingStatus::NotServing)
                 .await;
 
             let reporter = health_reporter.clone();
-            tokio::spawn(async move {
+            readiness_checks = readiness_checks.spawn(async move {
                 ready.await;
                 reporter
                     .set_service_status(name, ServingStatus::Serving)
                     .await;
                 info!("gRPC service {name} is now SERVING");
+                Ok(())
             });
         }
 
         let mut service = Service::new();
+        service = service.attach(readiness_checks);
 
         // Start HTTPS server if TLS is configured
         if let (Some(listen_address), Some(config)) = (rpc_tls_listen_address, tls_config) {


crates/sui-indexer-alt-consistent-store/src/rpc/mod.rs
```diff
@@ -197,11 +199,40 @@ impl<'d> RpcService<'d> {
...
+        // Set empty string service to NotServing (this is what the health probe will check)
+        health_reporter
+            .set_service_status("", ServingStatus::NotServing)
+            .await;
+
+        // Flip to Serving once snapshots exist
+        {
+            let reporter = health_reporter.clone();
+            tokio::spawn(async move {
+                loop {
```
amnn: `while`

crates/sui-indexer-alt-consistent-store/src/rpc/mod.rs
```diff
@@ -197,11 +199,40 @@ impl<'d> RpcService<'d> {
             ))
             .layer(CatchPanicLayer::new(metrics));
 
-        for service_name in service_names {
+        for service_name in &service_names {
```
amnn: I don't think this is technically right -- the health check is itself a service, and by doing this, we are saying that it is not ready until the Db has a snapshot, which is not true. Similar for the introspection service.

The right way to do this is to supply a `ready` future with each service when we add it, and then here, we spawn a separate task for each (user) service, that awaits the `ready` service and then sets it as serving.

Then we can add this health check just for the consistent service, and the health probe should check the health of the consistent service.

wlmyng: Okay, that makes sense. Today in the operations repo we don't check the health of the specific service, but you're right that it's just the ConsistentService grpc that is NOT_SERVING until it sees db snapshots. So we'll update both places

crates/sui-indexer-alt-consistent-store/src/rpc/mod.rs
```diff
@@ -197,12 +199,27 @@ impl<'d> RpcService<'d> {
...
                 .await;
         }
 
+        for (name, ready) in service_futures {
+            health_reporter
+                .set_service_status(name, ServingStatus::NotServing)
+                .await;
+
+            let reporter = health_reporter.clone();
+            tokio::spawn(async move {
```
amnn: nit: spawn these into a `Service` and attach that as a secondary to the service you create below, so that they are correctly cleaned up in case the service as a whole is instructed to wind down before all of these `ready` futures complete.

amnn (approved):
Just one more additional change for clean up purposes!

PR #25030, opened 2026-01-21
[move][move-vm][bella-ciao] Remove many VM pointer operations, including mutability

cgswords:
## Description 

Revise `VMPointer` to remove a lot of the unused functionality. Also refactor how jump tables are computed to avoid needing mutable references to VM pointers.

## Test plan 

All tests still pass, plus will be subject to replay after landing.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/jit/execution/translate.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/jit/execution/translate.rs
@@ -20,7 +20,6 @@ use crate::{
         vm_pointer::VMPointer,
     },
 };
-use indexmap::IndexMap;
 use move_binary_format::{
     errors::{PartialVMError, PartialVMResult},
     file_format::{
@@ -32,6 +31,9 @@ use move_core_types::{
     identifier::Identifier, language_storage::ModuleId, resolver::IntraPackageName,
     vm_status::StatusCode,
 };
+
+use indexmap::IndexMap;
+
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 
 // -------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


external-crates/move/crates/move-vm-runtime/src/shared/vm_pointer.rs
```diff
@@ -18,97 +19,48 @@ pub struct VMPointer<T>(*const T);
...
-pub fn to_mut_ref<'a, T>(value: *const T) -> &'a mut T {
-    unsafe { &mut *(value as *mut T) }
-}
-
 // -------------------------------------------------------------------------------------------------
 // Trait Implementations
 // -------------------------------------------------------------------------------------------------
 
+// [SAFETY] VMPointer<T> is Send and Sync because it only allows immutable access to T.
+
```
tnowacki, line 57: Do we need `Send` too? Does `Send` even make sense here? I feel like it doesn't

external-crates/move/crates/move-vm-runtime/src/shared/vm_pointer.rs
```diff
@@ -3,13 +3,14 @@
 
 #![allow(unsafe_code)]
 
-use std::mem::MaybeUninit;
```
tzakian, line 6: It's a sea of red 🥳

tzakian (approved):
Ok, this _is_ nicer.

external-crates/move/crates/move-vm-runtime/src/jit/execution/translate.rs
```diff
@@ -1150,32 +1132,58 @@ fn alloc_function(
...
+        fn_bytecode
             .into_iter()
-            .map(|bc| bytecode(context, jump_tables, bc))
+            .map(|bc| bytecode(context, &jump_table_ptrs, bc))
             .collect::<PartialVMResult<Vec<Bytecode>>>()?
             .into_iter(),
     )?;
-    Ok(result)
+
+    // Retunr the final bytecode and jump tables
```
akmysti: typo

cgswords:
```suggestion
    // Return the final bytecode and jump tables
```

external-crates/move/crates/move-vm-runtime/src/shared/vm_pointer.rs
cgswords: Sadly, according to rust spec.

cgswords:
> nit: flatten_and_renumber_blocks seems to do two separate things. maybe we could split that into two functions 1. for jump tables and 2. for code offsets.

To do that, we would need to return the code offset map -- only to use it to update the jump tables. I will break the function up further, though.

PR #25033, opened 2026-01-21
[consensus] fix randomized dag tests

mwtian:
## Description 

Fix test logic to avoid flakiness.

## Test plan 

CI
`python ./scripts/simtest/seed-search.py test_randomized_dag_with_7_authorities --test consensus_dag_tests --num-seeds 1000`

### DIFF
--- a/consensus/core/src/commit_test_fixture.rs
+++ b/consensus/core/src/commit_test_fixture.rs
@@ -4,6 +4,7 @@
 //! Shared test fixture for commit-related tests.
 //! Used by both commit_finalizer.rs tests and randomized_tests.rs.
 
+use std::collections::{BTreeMap, BTreeSet};
 use std::sync::Arc;
 
 use consensus_config::{AuthorityIndex, Stake};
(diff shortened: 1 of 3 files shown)


consensus/core/src/block.rs
```diff
@@ -622,7 +622,24 @@ impl TestBlock {
         self
     }
 
-    pub fn set_ancestors(mut self, ancestors: Vec<BlockRef>) -> Self {
+    /// Sorts then sets ancestors in the TestBlock.
```
akichidis, line 625:
```suggestion
    /// Sorts then sets ancestors in the TestBlock to make sure that author's own block is always first as expected by the rest of the system.
```

consensus/core/src/commit_test_fixture.rs
```diff
@@ -206,34 +211,40 @@ impl CommitTestFixture {
...
+    let (shortest_idx, shortest_sequence) = commit_sequences
+        .iter()
+        .enumerate()
+        .min_by_key(|(_, seq)| seq.len())
+        .expect("commit_sequences should not be empty");
+
+    for (run, commit_sequence) in commit_sequences.iter().enumerate() {
+        assert!(
+            commit_sequence.len() <= shortest_sequence.len() + 3,
+            "Commit sequence at run {run} is more than 3 commits longer than shortest (run {shortest_idx}): {} vs {}",
```
akichidis, line 231: How is `3` been picked?

mwtian: Added a comment. It is based on the maximum number of commits in INDIRECT_REJECT_DEPTH, which is 3 rounds.

PR #25051, opened 2026-01-22
[move-stdlib] Lossless div

tnowacki:
## Description 

- Following the naming convention of lossless_shl/shr 
- Adds a function that returns some(x / y) only if x % y == 0

## Test plan 

- new tests 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-swarm-config/tests/snapshots/snapshot_tests__populated_genesis_snapshot_matches-2.snap
+++ b/crates/sui-swarm-config/tests/snapshots/snapshot_tests__populated_genesis_snapshot_matches-2.snap
@@ -240,13 +240,13 @@ validators:
         next_epoch_worker_address: ~
         extra_fields:
           id:
-            id: "0xe5a73b2a49b1855e5e5762bf2b104b8791d02e07990230a2713fb06ab2476980"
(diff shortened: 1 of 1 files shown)


crates/sui-framework/packages/move-stdlib/sources/macros.move
```diff
@@ -166,6 +166,14 @@ public macro fun num_lossless_shr<$T>($x: $T, $shift: u8, $bit_size: u8): Option
     }
 }
 
+public macro fun num_lossless_div<$T>($x: $T, $y: $T): Option<$T> {
+    let x = $x;
+    let y = $y;
+    if (y == 0) option::none()
```
tzakian, line 172: Should we have a checked div that does this check? I vaguely remember us discussing this and deciding against it I believe? But it feels sane to have this (and it would then match with the other lossless_yyy functions being a checked_yyy + reversible)

tnowacki: Landed in #24834

PR #25052, opened 2026-01-22
Further batch workload changes

mystenmark:
- Create batches in composite workload
- change soft bundle => batch where appropriate

### DIFF
--- a/crates/sui-benchmark/src/workloads/composite/mod.rs
+++ b/crates/sui-benchmark/src/workloads/composite/mod.rs
@@ -41,6 +41,8 @@ use tracing::{debug, info};
 
 use super::MultiGas;
 
+const MAX_BATCH_SIZE: usize = 4;
+
 macro_rules! update_gas {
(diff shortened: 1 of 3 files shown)


crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -918,6 +1005,66 @@ impl Workload<dyn Payload> for CompositeWorkload {
                 info!("TestCoinCap not available - skipping TEST_COIN address balance seeding");
             }
         }
+
+        // split remaining gas coins into 4 equal parts
```
alex-mysten, line 1011: Why hardcoded 4 parts?

alex-mysten (approved):
LGTM. One comment

crates/sui-benchmark/src/workloads/composite/mod.rs
mystenmark: will make this configurable

PR #25060, opened 2026-01-22
interning cur_key to only when needed

akmysti:
## Description 

Since the interner is multithreaded, it might be better to avoid interning cur_key  when it is not used (when the state is State::NotVisited ).

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


akmysti:
> The cost/benefit here seems low to me. We have to move the interner call three different places, ensuring each one is accurate, to avoid a single call in the else case of unvisited modules. And, frankly, interning things early seems preferable to me. If we are so intent on saving the cycles, we should simply enqueue them interned, as `while let Some((cur_id, cur_key)) = stack.pop()` or similar.

agreed. i guess it is better to leave as is for now. This may not be worth optimizing.

PR #25061, opened 2026-01-22
add alias versions to signed_data_cache key

aschran:
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/signature_verifier.rs
+++ b/crates/sui-core/src/signature_verifier.rs
@@ -93,13 +93,6 @@ impl CertBuffer {
     }
 }
 
-// Signed data cache is also keyed on alias versions so that signatures are re-verified
-// if aliases change.
-type SignedDataCacheKey = (
-    SenderSignedDataDigest,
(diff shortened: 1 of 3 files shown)


crates/sui-core/src/signature_verifier.rs
```diff
@@ -466,10 +473,12 @@ impl SignatureVerifier {
     fn verify_tx(
         &self,
         signed_tx: &SenderSignedData,
+        alias_versions: Vec<(SuiAddress, Option<SequenceNumber>)>,
         aliased_addresses: Vec<(SuiAddress, NonEmpty<SuiAddress>)>,
     ) -> SuiResult {
+        let cache_key = (signed_tx.full_message_digest(), alias_versions);
```
mystenmark: if we just hashed the SignedDataCacheKey then we could avoid cloning the vector. I suppose most transactions will have an empty alias set so it may not matter much in practice.

PR #25067, opened 2026-01-22
[indexer-alt] Quick highlight of storage requirements for 30 and 90d retention 

wlmyng:
## Description 

Update the GraphQL and General-Purpose setup guide with estimated postgres db footprint requirements.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


abhinavg6 (approved):
The only general comment is that we should probably indicate that the numbers are based on where the networks is today at the start of 2026, and those may differ at different stages in future depending on the network growth. Basically people should use those as directional rather than exact figures.

jessiemongeon1 (approved):

PR #25069, opened 2026-01-22
Separate settle and commit in object funds checker

lxfind:
## Description 

Previously we rely on settlements to garbage collect pending settlement information in the object fund withdraw scheduler. That doesn't work because settlements can come out of order with execution of the withdraws. So we cannot yet remove those pending withdraw information even when we see settlements. This PR introduces another interface besides settlement, which is the commitment of transactions, triggered from checkpoint executor. This is the only reliable signal where we could garbage collect old pending withdraw information. Settlement signals still come from execution for liveness.
The scheduler is also moved and renamed. It now lives under accumulator sub crate, and named object_funds_checker instead of scheduler. It lives in AuthorityState directly now.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/accumulators/object_funds_checker/mod.rs
+++ b/crates/sui-core/src/accumulators/object_funds_checker/mod.rs
@@ -245,10 +245,6 @@ impl ObjectFundsChecker {
         accumulator_version: SequenceNumber,
     ) -> bool {
         for (obj_id, amount) in object_withdraws {
-            // It is safe to get the latest funds here because this function is called during execution,
-            // which means this transaction is not committed yet,
-            // so the settlement transaction at the end of the same consensus commit cannot have settled yet.
-            // That is, we must be blocked by this transaction in order to make progress.
             let funds = funds_read.get_account_amount_at_version(obj_id, accumulator_version);
             // Reading inner without a top-level lock is safe because no two transactions can be withdrawing
             // from the same account at the same time.


crates/sui-core/src/accumulators/object_funds_checker/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+        }
+        let Some(accumulator_version) = execution_env.assigned_versions.accumulator_version else {
+            // Fastpath transactions that perform object funds withdraws
+            // must wait for consensus to assign the accumulator version.
+            // We cannot optimize the scheduling by processing fastpath object withdraws
+            // sooner because these may get reverted, and we don't want them
+            // pollute the scheduler tracking state.
+            // TODO: We could however optimize execution by caching
+            // the execution state to avoid re-execution.
+            return false;
```
mystenmark, line 134: i think this branch can be removed now that we do not execute on the fast path. The Option can be removed as well

crates/sui-core/src/accumulators/object_funds_checker/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    fn try_withdraw(
+        &self,
+        funds_read: &dyn AccountFundsRead,
+        object_withdraws: &BTreeMap<AccumulatorObjId, u64>,
+        accumulator_version: SequenceNumber,
+    ) -> bool {
+        for (obj_id, amount) in object_withdraws {
+            // It is safe to get the latest funds here because this function is called during execution,
+            // which means this transaction is not committed yet,
```
mystenmark: i don't understand this comment - we aren't getting the latest funds, we're getting the funds at the given version, right?

crates/sui-core/src/accumulators/object_funds_checker/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+            // so the settlement transaction at the end of the same consensus commit cannot have settled yet.
+            // That is, we must be blocked by this transaction in order to make progress.
+            let funds = funds_read.get_account_amount_at_version(obj_id, accumulator_version);
+            // Reading inner without a top-level lock is safe because no two transactions can be withdrawing
+            // from the same account at the same time.
+            let unsettled_withdraw = self
+                .inner
+                .read()
```
mystenmark, line 253: would it better to acquire `inner.read()` once, instead of every loop iteration?

mystenmark: oh but i guess if we do that we should do the `get_account_amount_at_version` in a separate loop first. up to you.

PR #25072, opened 2026-01-23
Improvements to composite workload to trigger insufficient balance condition in scheduler

mystenmark:
This is obviously a big change. Besides the introduction of the overflow operation, there was a ton of work to make soft bundle submission reliable.

- Add fullnode_proxies to benchmark workloads for RPC reads
- Make SystemStateObserver reliable in simtest
- fixup proxy
- Create address balance over-draws
- Fixes for reliable handling a soft bundles
- format

### DIFF
--- a/crates/sui-benchmark/src/lib.rs
+++ b/crates/sui-benchmark/src/lib.rs
@@ -30,7 +30,7 @@ use sui_json_rpc_types::{
     SuiTransactionBlockEffects, SuiTransactionBlockEffectsAPI, SuiTransactionBlockResponseOptions,
 };
 use sui_protocol_config::ProtocolConfig;
-use sui_sdk::{SuiClient, SuiClientBuilder};
+use sui_sdk::{SuiClient, SuiClientBuilder, error::Error as SuiSdkError};
 use sui_types::sui_system_state::sui_system_state_summary::SuiSystemStateSummary;
 use sui_types::transaction::Argument;
 use sui_types::transaction::CallArg;
@@ -515,7 +515,7 @@ impl ValidatorProxy for LocalValidatorAggregatorProxy {
         &self,
         txs: Vec<Transaction>,
     ) -> anyhow::Result<Vec<(TransactionDigest, WaitForEffectsResponse)>> {
-        execute_soft_bundle_with_retries(&self.td, txs).await
+        execute_soft_bundle_with_retries(&self.td, &txs).await
     }
 
     fn get_chain_identifier(&self) -> ChainIdentifier {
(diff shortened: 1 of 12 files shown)


crates/sui-benchmark/src/lib.rs
```diff
@@ -293,6 +313,8 @@ impl ExecutionEffects {
 pub trait ValidatorProxy {
     async fn get_object(&self, object_id: ObjectID) -> Result<Object, anyhow::Error>;
 
+    async fn get_sui_address_balance(&self, address: SuiAddress) -> Result<u64, anyhow::Error>;
```
alex-mysten, line 320: Should this be more general, since we also have other currency based testing?

crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -483,15 +526,15 @@ impl Payload for CompositePayload {
...
-            if i >= self.current_batch_op_sets.len() {
-                break;
-            }
             let op_set = self.current_batch_op_sets[i];
             match &result.status {
                 BatchedTransactionStatus::Success { effects } => {
                     if effects.is_cancelled() {
                         metrics.record_cancellation(op_set);
+                    } else if effects.is_insufficient_funds() {
+                        metrics.record_insufficient_funds(op_set);
```
alex-mysten, line 588: Are we just validating with these metrics that we hit the "sometimes" reachability assertion? 

It feels a bit strange to special case and collect metrics about this particular type of error ~ the other metrics types feel more general-purpose (Success | Failure | Invalid_Request)

crates/sui-benchmark/src/benchmark_setup.rs
```diff
@@ -47,44 +48,48 @@ impl BenchmarkSetup {
...
+        // Always create fullnode proxies for RPC reads
+        let mut fullnode_proxies: Vec<Arc<dyn ValidatorProxy + Send + Sync>> = vec![];
+        for fullnode_rpc_url in fullnode_rpc_urls.iter() {
+            info!("Creating FullNodeProxy: {:?}", fullnode_rpc_url);
+            fullnode_proxies.push(Arc::new(
+                FullNodeProxy::from_url(fullnode_rpc_url, registry).await?,
+            ));
+        }
+
+        // Create execution proxies - either fullnode proxies or validator proxies
```
alex-mysten, line 65: Imagining we have these so that we have nice coverage over read / execute paths, but it's not immediately clear why we provide both options to each method. Maybe we should rename the fullnode / rpc_read_proxy path accordingly to make this distinction more clear?

crates/sui-benchmark/src/workloads/composite/mod.rs
mystenmark: Well, the cancellation condition was already pretty special purpose so I was following its lead. Open to making things less specific though

crates/sui-benchmark/src/benchmark_setup.rs
mystenmark: Yeah, the motivation was to give workloads the ability to do reads. Renaming is a good idea, we may want to eventually provide other ways of reading besides fullnode. Also, its probably a good idea not to pollute the ValidatorProxy trait and instead have a RPCProxy trait or something

crates/sui-benchmark/src/lib.rs
mystenmark: I plan to add another method with a coin type arg - the usual pattern in all our RPCs is to have a default for SUI + a general API for others.

crates/sui-benchmark/src/lib.rs
```diff
@@ -530,78 +555,192 @@ async fn execute_soft_bundle_impl(
...
-            let request = WaitForEffectsRequest {
-                transaction_digest: Some(*digest),
-                consensus_position: Some(*position),
-                include_details: true,
-                ping_type: None,
-            };
-            safe_client.wait_for_effects(request, None)
-        })
-        .collect();
+        if should_retry {
```
alex-mysten, line 669: Should the loop terminate after this block?

mystenmark: It will! The loop always terminates at the end unless it hits a continue.

PR #25073, opened 2026-01-23
Track total withdraw amount per account for object funds withdraw

lxfind:
## Description 

When we check whether object funds withdraw can be satisfied in an effects, it is important that we use the total withdraw amount instead of the merged results. For instance, if in the same transaction, we withdraw X from an object account and then deposit Y into it. We must ensure that there are X amount prior to this transaction, instead of just X - Y. This is important so that it cannot be exploited and do flash loans.

## Test plan 

Added unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/accumulators/object_funds_checker/integration_tests.rs
+++ b/crates/sui-core/src/accumulators/object_funds_checker/integration_tests.rs
@@ -294,4 +294,109 @@ async fn test_object_withdraw_multiple_withdraws() {
     );
 }
 
-// FIXME: More tests coming.
+#[tokio::test]
+async fn test_object_withdraw_and_deposit_same_transaction() {
+    telemetry_subscribers::init_for_testing();
+    let env = TestEnv::new().await;
+    env.fund_address(env.vault_obj.into(), 2).await;
+
+    // In the same transaction, we are withdrawing from the object account
+    // and depositing back to the same object account.
+    // The max net withdraws for this account should be 3, because at any given moment,
+    // the net withdraws is at most 3.
+    // Since the account has a balance of 2, the transaction should fail.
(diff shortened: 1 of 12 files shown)


crates/sui-core/src/accumulators/object_funds_checker/integration_tests.rs
```diff
@@ -292,4 +292,50 @@ async fn test_object_withdraw_multiple_withdraws() {
...
+#[tokio::test]
+async fn test_object_withdraw_and_deposit_same_transaction() {
+    let env = TestEnv::new().await;
+    env.fund_address(env.vault_obj.into(), 1).await;
+
+    let gas = env.oref(&env.gas_obj).await;
+    // In the same transaction, we are withdrawing from the object account
+    // and depositing back to the same object account twice.
+    // This should result in a total withdraw amount of 2, even though
+    // if we merge all accumulator events they will balance out to 0.
```
mystenmark: do you think there's any argument to be made for tracking the running total withdrawal? i.e. if you sum `[-1, 1 , -1, 1]` in order, you never drop lower than -1.

I'm not sure its worth handling this but I could imagine someone asking about this someday.

lxfind: What is the purpose to track the running total withdrawal?

mystenmark: it would be the most liberal behavior that prevents temporary minting - the current code is more conservative than strictly necessary

PR #25081, opened 2026-01-23
[transaction] validate transaction serialization

joyqvq:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-types/src/crypto.rs
+++ b/crates/sui-types/src/crypto.rs
@@ -285,6 +285,49 @@ impl ZkLoginPublicIdentifier {
 
         Ok(Self(bytes))
     }
+
+    /// Validates zkLogin public identifier structure: iss_len || iss || padded_32_byte_address_seed.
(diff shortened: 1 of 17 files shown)


crates/sui-types/src/crypto.rs
```diff
@@ -285,6 +285,49 @@ impl ZkLoginPublicIdentifier {
...
+
+        if address_seed_bytes.len() > 32 {
+            return Err(SuiErrorKind::InvalidSignature {
+                error: format!(
+                    "address seed must be at most 32 bytes, got {}",
+                    address_seed_bytes.len()
+                ),
+            }
+            .into());
+        }
```
bmwill, line 327: This should require that its exactly 32 bytes long since we expect this to be the full padded version.

PR #25082, opened 2026-01-23
Enable ephemeral --dump-bytecode-as-base64

mdgeorge4153:
## Description 

See the release notes below

## Test plan 

Shell tests were added to ensure that `--dump-bytecode-as-base64 --pubfile-path` produces bytecode containing the ephemeral addresses.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [X] CLI: 

You can now do `sui move build --dump-bytecode-as-base64 --pubfile-path <file>` to use ephemeral addresses for dumped bytecode.

The `--dump` flag can be used as a shorthand for `--dump-bytecode-as-base64`.

The full flag for `-e` has been changed to `--build-env` (the shorthand `-e` remains the same)

Fixed a bug when `--pubfile-path` is used with `-p`

- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/src/sui_commands.rs
+++ b/crates/sui/src/sui_commands.rs
@@ -632,7 +632,11 @@ impl SuiCommand {
                         .await?;
 
                         let mut root_pkg = if let Some(pubfile_path) = pubfile_path {
-                            let chain_id = context.grpc_client()?.get_chain_identifier().await?.to_string();
+                            let chain_id = context
+                                .grpc_client()?
+                                .get_chain_identifier()
+                                .await?
+                                .to_string();
                             let modes = build_config.mode_set();
                             load_root_pkg_for_ephemeral_publish_or_upgrade(
                                 &rerooted_path,


docs/content/guides/developer/packages/move-package-management.mdx
```diff
@@ -334,6 +334,12 @@ The command `sui client test-publish --pubfile-path <pubfile> --build-env <env>`
 
 If `<pubfile>` is omitted, it defaults to `Pub.<env-name>.toml` where `<env-name>` is the name of the active CLI environment, irrespective of the environments defined in the manifest. This file contains the addresses to use for publication.
 
+You can also prepare bytecode for publication without actually publishing it by passing `--pubfile-path <pubfile>` to `sui move build --dump-bytecode-as-base64`:
+
+```sh
+sui move build --dump-bytecode-as-base64 --pubfile-path Pub.localnet.toml
```
manolisliolios: Shouldn't this also need `build-env` (-e)? Or do we extract it from the pubfile? (but i guess the latter only works if the pubfile is not fresh)

crates/sui/tests/shell_tests/with_network/ephemeral_dump/dump_with_pubfile.snap
manolisliolios, line 1: Would be cool to also test the case where pubfile path does not exist (so it also requires build-env)

mdgeorge4153:
Note: the --allow-dirty flag is not being passed down to the PackageLoader constructor when converting a `BuildConfig`

PR #25086, opened 2026-01-24
fix: Allow nodes to start with snapshots from pre-quarantine epochs

EasonC13:
## Description

We encountered an issue where our sui-node failed to start with snapshots from certain epochs, and we're opening this PR hoping the MystenLabs team can review our proposed fix and ensure it's safe and appropriate for the codebase.

**After applying this fix, our newly set up validator node works well** and we wanted to share this solution in case others face the same issue.

**We're not certain if this is the best approach, and we'd value feedback from the MystenLabs team on whether there are any risks we haven't considered.**

### Problem We Encountered
Nodes fail to start with the error:
```
This version of sui-node can only run after data quarantining has been enabled.
Please run version 1.45.0 or later to the end of the current epoch and retry
```

This occurs even with recent snapshots because the flag is stored in the snapshot's `epoch_start_configuration`, representing the state when that epoch **started**, not the current node version.

### Root Cause (Our Understanding)
After investigating, we found that:
- Epoch flags are stored in snapshots and loaded directly from `perpetual_tables`
- Different networks (mainnet/testnet/devnet) enable protocol flags at different times
- Snapshots from epochs that started before this flag was enabled on that network won't have it
- The hard assertion prevented nodes from starting, blocking legitimate use cases

**We'd appreciate if the MystenLabs team could verify our analysis is correct.**

### Our Proposed Solution
We propose converting the hard assertion to a warning log message. Our reasoning:
- The actual data quarantining code in v1.63.3+ appears to be fully implemented and functional
- Only the epoch metadata flag check would be bypassed
- This would allow nodes to start with older snapshots while maintaining data safety
- Production validators would still be encouraged to use recent snapshots but won't be blocked

### Our Changes
**File**: `crates/sui-core/src/authority/consensus_quarantine.rs`
- Line 437-445: Changed `assert!()` to `if !flag { warn!() }`
- Added detailed warning message explaining the situation
- Preserved all actual quarantining functionality

### Tested scenarios:
1. **Build verification**: Successfully compiled sui-node with the fix
   ```bash
   cargo build --release -p sui-node
   ```

2. **Snapshot compatibility**: Verified that nodes can now start with snapshots from various epochs, including those before v1.45.0. **Our node is now running successfully with this fix applied.**

3. **Functionality verification**: Confirmed that all quarantining logic still operates correctly:
   - Deferred transactions are properly tracked
   - Consensus output is safely persisted
   - Transaction ordering is maintained

4. **Backward compatibility**: Cherry-picked cleanly onto framework/mainnet, ensuring compatibility with other versions

### Expected behavior:
- Nodes start successfully with older snapshots
- Warning logged when flag is missing:
  ```
  WARN: Data quarantining not active from beginning of epoch.
        This snapshot may be from an epoch before v1.45.0.
  ```
- All quarantining mechanisms function normally

---

## Release notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes):
  **Fix**: Nodes can now start with snapshots from epochs before the DataQuarantineFromBeginningOfEpoch flag was enabled. Previously, nodes would fail to start with a hard assertion error. The fix converts this to a warning while maintaining all data quarantining safety features. Node operators using snapshots from older epochs will see a warning log but the node will start and operate normally.

  **Action Required**: None. This is a bug fix that improves snapshot compatibility. Node operators are still encouraged to use recent snapshots for production validators.

- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority/consensus_quarantine.rs
+++ b/crates/sui-core/src/authority/consensus_quarantine.rs
@@ -434,10 +434,15 @@ impl ConsensusOutputCache {
             .get_all_deferred_transactions_v2()
             .expect("load deferred transactions cannot fail");
 
-        assert!(
-            epoch_start_configuration.is_data_quarantine_active_from_beginning_of_epoch(),
-            "This version of sui-node can only run after data quarantining has been enabled. Please run version 1.45.0 or later to the end of the current epoch and retry"
-        );
+        // RELAY-VALIDATOR: Downgrade assertion to warning to support snapshots from pre-1.45.0 epochs
+        // This is safe for relay-validator since it's not running on production mainnet
+        if !epoch_start_configuration.is_data_quarantine_active_from_beginning_of_epoch() {
+            tracing::warn!(
+                "Data quarantining not active from beginning of epoch. \
+                This snapshot may be from an epoch before v1.45.0. \
+                For production use, please run version 1.45.0 or later to the end of the current epoch."
+            );
+        }
 
         let executed_in_epoch_cache_capacity = 50_000;
 


mystenmark:
I think this PR needs to be rebased, there are many unrelated commits in here

PR #25088, opened 2026-01-24
[decompiler] Fix FixedBitSet panic on trivial functions with empty CFG

sblackshear:
Framework packages contain many `native` functions with no basic blocks, resulting in empty input maps. Previously, `Graph::new` would create a 0-node CFG, causing petgraph's dominator algorithm to panic with "put at index 0 exceeds fixedbitset size 0".

Fixes:
1. Ensure start_node exists in CFG even when input is empty
2. Handle missing nodes in structure_acyclic by creating empty sequences

Example package to repro this: 0x0000000000000000000000000000000000000000000000000000000000000001

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-decompiler/src/structuring/mod.rs
+++ b/external-crates/move/crates/move-decompiler/src/structuring/mod.rs
@@ -28,6 +28,11 @@ pub(crate) fn structure(
     mut input: BTreeMap<D::Label, D::Input>,
     entry_node: D::Label,
 ) -> D::Structured {
+    // Native functions have empty basic blocks - return early to avoid panicking in Graph::new
+    if input.is_empty() {
+        return D::Structured::Seq(vec![]);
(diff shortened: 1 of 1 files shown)


sblackshear:
> I am okay with this as-is, but I would prefer to handle this in the AST layer: add some `is_native: bool` field to the `Function` form, check it before ever calling into the structuring code. We could even check it in `structuring/mod.rs` in `fn structure`, returning if the input map is empty. Pushing this check this deep into the algorithm seems less-ideal.

Agree--there's no reason for a function body to be empty other than `native` functions, so that seems better

PR #25092, opened 2026-01-24
Improve test reliability

mystenmark:
- Send initial system state to SystemStateObserver
- wip
- Disable CompositeWorkload for fork test
- Fix retry error retry behavior
- make test reliable

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/src/workloads/payload.rs
+++ b/crates/sui-benchmark/src/workloads/payload.rs
@@ -121,8 +121,12 @@ pub trait Payload: Send + Sync + std::fmt::Debug + Display {
         NonZeroUsize::MAX
     }
 
-    // TODO: Implement this to allow limiting the size of each soft bundle.
-    // fn max_soft_bundle_size(&self) -> NonZeroUsize
+    /// Maximum size of any individual soft bundle.
+    fn max_soft_bundle_size(&self) -> NonZeroUsize {
+        // TODO: we could get this from the protocol config but a) its unlikely to change
+        // b) it would be very hard to do that
+        NonZeroUsize::new(5).unwrap()
+    }
 
     /// Creates a batch of transactions for concurrent execution.
     /// Only called when `is_batched()` returns true.
--- a/crates/sui-benchmark/src/drivers/bench_driver.rs
+++ b/crates/sui-benchmark/src/drivers/bench_driver.rs
@@ -64,6 +64,7 @@ use super::{BenchmarkStats, StressStats};
 fn partition_into_random_bundles<T>(
     items: Vec<T>,
     max_bundles: NonZeroUsize,
+    max_bundle_size: usize,
 ) -> Vec<(usize, Vec<T>)> {
     if items.is_empty() {
         return vec![];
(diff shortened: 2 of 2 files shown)


crates/sui-benchmark/tests/simtest.rs
```diff
@@ -1582,14 +1589,14 @@ mod test {
 
         let metrics_sum = metrics.sum_all();
 
+        println!("metrics: {:#?}", metrics.sum_all());
```
alex-mysten: remove println

crates/sui-benchmark/src/workloads/payload.rs
```diff
@@ -40,6 +40,9 @@ pub enum BatchedTransactionStatus {
         /// Error message describing the failure.
         error: String,
     },
+    /// We didn't get a specific error message, so the failure could be
+    /// retriable or permanent.
```
alex-mysten, line 44: When would we hit this branch?

crates/sui-benchmark/src/lib.rs
```diff
@@ -832,6 +817,14 @@ impl FullNodeProxy {
     }
 }
 
+fn is_retryable_sdk_error(err: &SuiSdkError) -> bool {
+    let err_str = format!("{:?}", err);
+    !(err_str.contains("Error checking transaction input objects")
+        || err_str.contains("Transaction Expired")
+        || err_str.contains("already locked by a different transaction")
+        || err_str.contains("is not available for consumption"))
```
alex-mysten, line 825: The other errors have an is_retriable() method, is this something we should consider adding more generally, outside of the scope of this PR?

mystenmark: normally I would say yes, but a) this is an SDK error and b) SDK errors seem kind of broken because although they have a rich error enum https://github.com/MystenLabs/sui/blob/8b38457ce18291909559624268b5c1d89639d24f/crates/sui-sdk/src/error.rs#L12 everything shows up as `RPCError`, so it seems like adding anything useful to this error type is going to take me down a deep rabbithole

crates/sui-benchmark/tests/simtest.rs
mystenmark: really? seemed quite useful for eyeballing what was going on in the test

mystenmark: i guess i should log it though

crates/sui-benchmark/src/workloads/payload.rs
mystenmark: If the transaction is rejected by vote in consensus, but the validator that you are connected to did not itself cast a no vote, it doesn't know what the reason for rejection is.

alex-mysten: Reasonable ~ but can you add a comment here? It's not obvious without knowing that fact

PR #25093, opened 2026-01-24
Properly map user sigs to alias versions in checkpoint contents

aschran:
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-types/src/transaction.rs
```diff
@@ -3904,15 +3919,20 @@ mod nonempty_as_vec {
...
     ),
 
     /// Object IDs that are claimed to be immutable.
     /// Used to filter out immutable objects from lock acquisition in consensus handler.
     ImmutableInputObjects(Vec<ObjectID>),
+
+    /// Address aliases used for signature verification.
+    /// Length must equal the number of `required_signers`. Each element maps the corresponding
+    /// signer to the signature index and alias object version (if any) used to verify it.
+    AddressAliasesV2(#[serde(with = "nonempty_as_vec")] NonEmpty<(u8, Option<SequenceNumber>)>),
```
mystenmark, line 3964: why have an index instead of just stipulating that the vector is sorted?

lxfind:
What was wrong with using SuiAddress?

aschran:
> What was wrong with using SuiAddress?

This fixes a bug about signature ordering in the checkpoint contents after address aliases is enabled. The removal of `SuiAddress` from some internal structs is incidental, it was simply not needed

(edit: internal as in, not part of the fullnode interface, but it was admittedly in a message that is exchanged between validators through consensus)

PR #24900, opened 2026-01-12
graphql-alt: scan transactions api

henryachen:
## Description 

Adds a new transactionsScan query that scans checkpoints using bloom filters to find transactions matching filter criteria.

  Changes

  GraphQL (sui-indexer-alt-graphql)
  - New transactionsScan query endpoint
  - maxScanLimit service config to limit checkpoints scanned per query
  - Transaction filter and scan logic in scan.rs
  
  Schema (sui-indexer-alt-schema)
  - `bloom_contains` SQL function for checking probe membership in bloom filters

  Reader (sui-indexer-alt-reader)
  - `cp_blooms` loader for batch-loading bloom filter data

## Test plan 

```
cargo nextest run -p sui-indexer-alt-graphql graphql_scan_limit_tests
cargo nextest run -p sui-indexer-alt-graphql
cargo nextest run -p sui-indexer-alt
cargo nextest run -p sui-indexer-alt-schema
```

Query Plan:
```
Limit  (cost=1911.69..1358704.33 rows=62 width=8) (actual time=14.454..14.836 rows=62 loops=1)
  Buffers: shared hit=7287
  CTE block_bit_probes
    ->  ProjectSet  (cost=0.00..8.20 rows=1635 width=18) (actual time=0.004..0.431 rows=1635 loops=1)
          ->  Result  (cost=0.00..0.01 rows=1 width=0) (actual time=0.001..0.001 rows=1 loops=1)
  ->  Nested Loop  (cost=1903.49..1358696.13 rows=62 width=8) (actual time=14.453..14.830 rows=62 loops=1)
        Buffers: shared hit=7287
        ->  Limit  (cost=1903.06..1903.22 rows=62 width=24) (actual time=14.411..14.414 rows=1 loops=1)
              Buffers: shared hit=6976
              CTE block_lookup
                ->  Nested Loop Left Join  (cost=41.30..1723.38 rows=200 width=131) (actual time=0.654..2.899 rows=327 loops=1)
                      Buffers: shared hit=1348
                      ->  HashAggregate  (cost=40.88..42.88 rows=200 width=10) (actual time=0.620..0.682 rows=327 loops=1)
                            Group Key: block_bit_probes_1.cp_block_index, block_bit_probes_1.bloom_idx
                            Batches: 1  Memory Usage: 61kB
                            ->  CTE Scan on block_bit_probes block_bit_probes_1  (cost=0.00..32.70 rows=1635 width=10) (actual time=0.000..0.159 rows=1635 loops=1)
[...]

PR #24788, opened 2026-01-05
graphql-alt: scanning api bloom filter pipelines

henryachen:
## Description 

Adds two new indexer pipelines to support transaction scanning queries via bloom filters:

  cp_blooms concurrent pipeline: Per-checkpoint bloom filters
  - Indexes function calls, affected objects, senders, and recipients for each checkpoint
  - Folds bloom filter if it is sparse up to a certain min size or density
  - Stored in cp_blooms table with one row per checkpoint

  cp_bloom_blocks sequential pipeline: Blocked bloom filters spanning 1000 checkpoints
  - Splits each 256KB bloom into 128 separate smaller 2KB blocks
  - Stored in cp_bloom_blocks table with ~128 rows per cp_block (sparse, only non-zero blocks)
  - ORs bloom blocks on conflicts

  Schema changes:
  - cp_blooms: Per-checkpoint bloom filters
  - cp_bloom_blocks: Blocked bloom filters with (cp_block_id, bloom_block_index) primary key

## Test plan
```
cargo nextest run -p sui-indexer-alt
cargo nextest run -p sui-indexer-alt-schema
```

## Stack:
https://github.com/MystenLabs/sui/pull/24788 👈 
https://github.com/MystenLabs/sui/pull/24900
https://github.com/MystenLabs/sui/pull/25097

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL:   add bloom filter pipelines for scanning APIs
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #25097, opened 2026-01-26
graphql-alt: add event scanning

henryachen:
## Description 

  - Added scanEvent GraphQL query endpoint for scanning events using bloom filters
  - Reuse existing scan infrastructure (bloom querying, pagination) from
 Transaction Scanning
  - Added e2e tests covering filtering by sender, module, type, and pagination

## Test plan 

```
cargo nextest run -p sui-indexer-alt-graphql
cargo nextest run -p sui-indexer-alt-graphql -- schema_sdl
cargo nextest run -p sui-indexer-alt-graphql --features staging -- schema_sdl
cargo nextest run -p sui-indexer-alt-e2e-tests -- graphql/
```

## Stack:
https://github.com/MystenLabs/sui/pull/24788 
https://github.com/MystenLabs/sui/pull/24900 
https://github.com/MystenLabs/sui/pull/25097 👈 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/events/filter/module.snap
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/events/filter/module.snap
@@ -625,24 +625,19 @@ Response: {
 task 9, lines 141-148:
 //# run-graphql
 Response: {
-  "data": {
-    "eventsByModuleAndTypeIsUnavailable": null
-  },
+  "data": null,
   "errors": [
     {
-      "message": "Filtering by both emitting module and event type is not supported not available",
+      "message": "Failed to parse \"EventFilter\": Filtering by both emitting module and event type is not supported",
       "locations": [
         {
           "line": 2,
-          "column": 3
+          "column": 65
         }
       ],
       "path": [
         "eventsByModuleAndTypeIsUnavailable"
-      ],
-      "extensions": {
-        "code": "FEATURE_UNAVAILABLE"
-      }
+      ]
     }
   ]
 }
--- a/crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
@@ -18,7 +18,6 @@ use crate::api::scalars::type_filter::TypeFilter;
 use crate::api::scalars::uint53::UInt53;
 use crate::api::types::event::CEvent;
 use crate::api::types::lookups::CheckpointBounds;
-use crate::api::types::transaction::filter::ScanFilterValidator;
 use crate::error::RpcError;
 use crate::pagination::Page;
 
@@ -204,38 +203,6 @@ impl CustomValidator<EventFilter> for EventFilterValidator {
                 "Filtering by both emitting module and event type is not supported",
             ));
         }
-        Ok(())
-    }
-}
-
-impl CustomValidator<EventFilter> for ScanFilterValidator {
-    fn check(&self, filter: &EventFilter) -> Result<(), InputValueError<EventFilter>> {
-        let (cp_lo, cp_hi) = if let Some(at) = filter.at_checkpoint.map(u64::from) {
-            (at, at.saturating_add(1))
-        } else if let Some(before) = filter.before_checkpoint.map(u64::from) {
-            (
-                filter
-                    .after_checkpoint
-                    .map_or(0, |a| u64::from(a).saturating_add(1)),
-                before,
-            )
-        } else {
-            return Err(InputValueError::custom(format!(
-                "Unbounded scan, add beforeCheckpoint or atCheckpoint filters (Scan limit: {}).",
-                self.max_scan_limit,
-            )));
-        };
-
-        let cps_scanned = cp_hi.saturating_sub(cp_lo);
-        if cps_scanned > self.max_scan_limit {
-            return Err(InputValueError::custom(format!(
-                "Scan of {cps_scanned} checkpoints exceeds maximum of {}. \
-                Use afterCheckpoint and beforeCheckpoint or atCheckpoint filters \
-                to reduce the range.",
-                self.max_scan_limit,
-            )));
-        }
-
         Ok(())
     }
 }
--- a/crates/sui-indexer-alt-graphql/src/api/types/transaction/filter.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/transaction/filter.rs
@@ -5,6 +5,7 @@ use async_graphql::Context;
 use async_graphql::CustomValidator;
 use async_graphql::Enum;
 use async_graphql::InputObject;
+use async_graphql::InputType;
 use async_graphql::InputValueError;
 use sui_indexer_alt_reader::kv_loader::TransactionContents;
 use sui_indexer_alt_schema::blooms::should_skip_for_bloom;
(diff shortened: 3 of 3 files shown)


crates/sui-indexer-alt-graphql/src/api/types/scan/cursor.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/// Extension trait for scan cursors that also track event position within a transaction.
+pub(crate) trait ScanCursorWithEvent: ScanCursor {
+    /// The event index within the transaction for this cursor position.
+    fn ev_sequence_number(&self) -> u64;
+}
+
+/// The transaction index bounds `[tx_lo, tx_hi)` within a checkpoint based on cursor positions.
+pub(crate) fn cp_tx_bounds<C: ScanCursor>(
```
wlmyng: Looking at its callsites, when we get here, we've already bloom-filtered for the candidate_cps. It's impossible for candidates_cps to contain checkpoints outside the cursor range. So at its core we're checking if the `after` cursor is a slice into the first checkpoint's txns, and the `before` cursor is a slice somewhere into the middle of the last checkpoint's txns. For most other checkpoints, we will return `Range<checkpoint_digests.len()>`

I think this was a bit obscured by the impl., but I don't think we need to modify the impl, could we add some calrification to the docs with the intent of what this fn is doing?

crates/sui-indexer-alt-graphql/src/api/types/scan/paginate.rs
```diff
@@ -20,36 +28,29 @@ use crate::error::RpcError;
...
-                page,
-                cp_sequence_number,
-                checkpoint_digests.len(),
-            ))
-        } else {
-            Either::Right(cp_tx_bounds(page, cp_sequence_number, checkpoint_digests.len()).rev())
-        };
+        let bounds = directional_iter(
+            page,
+            cp_tx_bounds(page, cp_sequence_number, checkpoint_digests.len()),
```
wlmyng: tbh i feel like cp_tx_bounds and directional_iter belong in the paginate module, rather than disbursed in cursor and scan

crates/sui-indexer-alt-graphql/src/api/types/available_range.rs
```diff
@@ -350,7 +350,7 @@ collect_pipelines! {
         pipelines.insert("consistent".to_string());
         pipelines.insert("obj_versions".to_string());
     };
-    Query.[events] |pipelines, filters| {
+    Query.[events, eventsScan] |pipelines, filters| {
```
wlmyng: eventsScan only depends on the cp_blooms + block tables right? they have entierly different retentions?

crates/sui-indexer-alt-graphql/src/api/types/scan/cursor.rs
henryachen: Agreed cp clamping here makes the function harder to reason able, this was simplified this based on changes in transaction scans!

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -187,6 +193,57 @@ impl Event {
...
+        ctx: &Context<'_>,
+        scope: Scope,
+        page: Page<CEvent>,
+        filter: EventFilter,
+    ) -> Result<Connection<String, Event>, RpcError<ScanError>> {
+        let limits: &Limits = ctx.data()?;
+        let watermarks: &Arc<Watermarks> = ctx.data()?;
+        let available_range_key = AvailableRangeKey {
+            type_: "Query".to_string(),
+            field: Some("eventsScan".to_string()),
```
amnn:
```suggestion
            field: Some("scanEvents".to_string()),
```

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -187,6 +193,57 @@ impl Event {
...
+            filters: Some(filter.active_filters()),
+        };
+        let reader_lo = available_range_key.reader_lo(watermarks).map_err(upcast)?;
+
+        let Some(checkpoint_viewed_at) = scope.checkpoint_viewed_at() else {
+            return Ok(Connection::new(false, false));
+        };
+
+        let Some(cp_bounds) = checkpoint_bounds(
```
amnn, line 226: I mentioned on the transaction scanning PR that we should be failing earlier if we aren't supplied with the appropriate checkpoint bounds -- practically what that means is that the `scan_range > limits.max_scan_limit` check below gets moved above this line, and looks something like this:

```rust
match filter {
    // Assume the max scan limit is >= 1
    EventFilter { at_checkpoint: Some(_), .. } => { /* nop */ },

    EventFilter {
        after_checkpoint: Some(after),
        before_checkpoint: Some(before),
        ..
    } => {
        let after: u64 = after.into();
        let before: u64 = before.into();
        let requested = before.saturating_sub(after).saturating_sub(1);

        if range > limits.max_scan_limit {
            return Err(bad_use_input(ScanError::LimitExceeded {
                requested,
                max: limits.max_scan_limit,
            });
        }
    }

    // New error case: complain that the query is missing checkpoint bounds.
    _ => { /* ... */ }
}
```

You still need to run things through `checkpoint_bounds`, for consistency reasons, but this makes the nature of the constraint less dynamic and therefore clearer for the user.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -95,16 +101,94 @@ pub(crate) async fn transactions(
         .collect()
 }
 
+pub(super) type EventsBySequenceNumbers = BTreeMap<EventCursor, Event>;
```
amnn: nit: types to the top. Also, should `bloom/mod.rs` be moved to just `bloom.rs` now that it's the only module here?

crates/sui-indexer-alt-graphql/src/api/types/available_range.rs
```diff
@@ -362,6 +362,10 @@ collect_pipelines! {
             pipelines.insert("ev_struct_inst".to_string());
         }
     };
+    Query.[scanEvents] |pipelines, _filters| {
+        pipelines.insert("cp_blooms".to_string());
+        pipelines.insert("cp_bloom_blocks".to_string());
+    };
```
amnn: nit: Maybe just merge this with `scanTransactions` below?

crates/sui-indexer-alt-graphql/src/api/query.rs
```diff
@@ -308,6 +309,29 @@ impl Query {
...
+    /// Note that this differs from `Query.events`, which supports a single filter with an optional sender filter, with no checkpoint range restriction and generally a shorter retention.
+    async fn scan_events(
+        &self,
+        ctx: &Context<'_>,
+        first: Option<u64>,
+        after: Option<CEvent>,
+        last: Option<u64>,
+        before: Option<CEvent>,
+        filter: Option<EventFilter>,
+    ) -> Result<Option<Connection<String, Event>>, RpcError<ScanError>> {
```
amnn: `Option<Result<...>>`

PR #25113, opened 2026-01-27
[consensus] test equivocations with randomized dag

mwtian:
## Description 

Test commit logic with equivocating blocks.

## Test plan 

`python ./scripts/simtest/seed-search.py test_randomized_dag_with_4_authorities_1_equivocator --test consensus_dag_tests --num-seeds 1000`

### DIFF
--- a/consensus/core/src/commit_test_fixture.rs
+++ b/consensus/core/src/commit_test_fixture.rs
@@ -311,7 +311,9 @@ pub struct RandomDagConfig {
 
 /// A randomly generated DAG for testing commit patterns with reject votes.
 pub struct RandomDag {
+    context: Arc<Context>,
     pub blocks: Vec<VerifiedBlock>,
+    num_rounds: Round,
 }
 
 impl RandomDag {
@@ -413,18 +415,29 @@ impl RandomDag {
         }
 
         RandomDag {
+            context,
             blocks: all_blocks.values().cloned().collect(),
+            num_rounds,
         }
     }
 
     /// Creates a RandomDag from existing blocks.
-    pub fn from_blocks(blocks: Vec<VerifiedBlock>) -> Self {
-        RandomDag { blocks }
+    pub fn from_blocks(context: Arc<Context>, blocks: Vec<VerifiedBlock>) -> Self {
+        let num_rounds = blocks.iter().map(|b| b.round()).max().unwrap_or(0);
+        RandomDag {
+            context,
+            blocks,
+            num_rounds,
+        }
     }
 
-    /// Creates an iterator yielding blocks in dependency-respecting random order.
-    pub fn random_iter<'a>(&'a self, rng: &'a mut StdRng) -> RandomDagIterator<'a> {
-        RandomDagIterator::new(self, rng)
+    /// Creates an iterator yielding blocks in constrained random order.
+    pub fn random_iter<'a>(
+        &'a self,
+        rng: &'a mut StdRng,
+        max_step: Round,
+    ) -> RandomDagIterator<'a> {
+        RandomDagIterator::new(self, rng, max_step)
     }
 }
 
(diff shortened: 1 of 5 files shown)


consensus/core/src/commit_test_fixture.rs
```diff
@@ -288,196 +292,290 @@ pub fn assert_commit_sequences_match(
...
+    // Number of distinct authorities creating blocks.
+    pub num_authorities: usize,
+    // Number of rounds to generate.
+    pub num_rounds: Round,
+    // Number of transactions per block.
+    pub num_transactions: u32,
+    // Percentage chance of transactions to randomly reject.
+    pub reject_percentage: u8,
+    // Each element specifies the authority index which the equivocators are acting on behalf of,
+    // and the number of equivocators acting for that authority.
```
akichidis:
```suggestion
    // Each element specifies the authority index and the number of equivocators that are acting for this authority.
```

consensus/core/src/commit_test_fixture.rs
```diff
@@ -288,196 +292,290 @@ pub fn assert_commit_sequences_match(
...
+    // under the specified authority index.
+    pub equivocators: Vec<(AuthorityIndex, u16)>,
```
akichidis, line 309: Maybe use the `InstanceID` instead of the tuple `(AuthorityIndex, u16)` ?

consensus/core/src/commit_test_fixture.rs
```diff
@@ -288,196 +292,290 @@ pub fn assert_commit_sequences_match(
...
         RandomDag { blocks }
     }
+
+    /// Creates an iterator yielding blocks in dependency-respecting random order.
+    pub fn random_iter<'a>(&'a self, rng: &'a mut StdRng) -> RandomDagIterator<'a> {
+        RandomDagIterator::new(self, rng)
+    }
+}
+
+// Identifies a consensus instance, by its authority index and sub-index within the authority.
```
akichidis: maybe `and sub-index within the authority.` -> `and sub-index that equivocates for this authority`

consensus/core/src/commit_test_fixture.rs
```diff
@@ -288,196 +292,290 @@ pub fn assert_commit_sequences_match(
...
+
         for r in 1..=num_rounds {
-            // Select random quorum-or-more authorities to produce blocks this round.
+            // Select random quorum-or-more stake to produce blocks this round.
             let target_stake = rng.gen_range(quorum_threshold..=total_stake);
-            let mut authorities: Vec<_> = committee.authorities().map(|(a, _)| a).collect();
-            authorities.shuffle(rng);
-            let selected_authorities: Vec<_> = authorities
+
+            // Select random instances to produce blocks this round.
```
akichidis, line 371: Maybe let's extend the comment that this is the point where equivocations might happen as it's possible to choose multiple equivocated proposers for same authorities

consensus/core/src/commit_test_fixture.rs
```diff
@@ -288,196 +292,290 @@ pub fn assert_commit_sequences_match(
...
+                .take_while(|instance| {
+                    if selected_stake >= target_stake {
+                        return false;
                     }
-                    *acc += committee.stake(a);
-                    Some(a)
+                    if !selected_authorities[instance.0.value()] {
+                        selected_authorities[instance.0.value()] = true;
+                        selected_stake += committee.stake(instance.0);
+                    }
```
akichidis, line 388: Maybe also leave a comment that we are doing this due to the possibility of choosing equivocating authorities. In that case the selected stake should not increase.

consensus/core/src/commit_test_fixture.rs
mwtian: The 2nd field is a count and not strictly an ID.

PR #25129, opened 2026-01-27
[bella-ciao] Remove `TypeView`

tzakian:
## Description 

Removes `TypeView` and its usages.

The gas changes are because of the gas charger changes in `dev_utils` gas schedule. Note that there are no semantic changes to the real gas meter on the Sui side though.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


cgswords (approved):
LGTM, but: are we ever going to regret removing this information from the gas?

external-crates/move/crates/move-vm-runtime/src/dev_utils/gas_schedule.rs
```diff
@@ -306,14 +305,10 @@ impl GasMeter for GasStatus<'_> {
         &mut self,
         _module_id: &ModuleId,
         _func_name: &str,
-        ty_args: impl ExactSizeIterator<Item = impl TypeView>,
         args: impl ExactSizeIterator<Item = impl ValueView>,
         _num_locals: NumArgs,
     ) -> PartialVMResult<()> {
-        self.charge_instr_with_size(
-            Opcodes::CALL_GENERIC,
-            ((ty_args.len() + args.len() + 1) as u64).into(),
```
cgswords, line 318: Do we care about this cost change?

dariorussi:
> LGTM, assuming we did not want this info anyway. @dariorussi might have opinions, too.

LGTM too, we can chat more but it seems reasonable?

dariorussi (approved):

external-crates/move/crates/move-vm-runtime/src/dev_utils/gas_schedule.rs
tzakian: Not really -- this is only used by the Move CLI outside of the Sui CLI.

I honestly think we should consider removing this entirely and just using the `tiered_gas_schedule` as that tracks most closely to the metering we do on the Sui side.

PR #25132, opened 2026-01-27
Allow multiple address groups of different priority in EndpointManager

aschran:
## Description 

Addresses from different sources are all saved, and the highest-priority source is used. This allows us to e.g. retain address overrides from admin server, even after epoch change when new committee is updated.

## Test plan 

Added unit tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-network/src/discovery/tests.rs
+++ b/crates/sui-network/src/discovery/tests.rs
@@ -904,8 +904,6 @@ async fn test_address_source_clear_all() -> Result<()> {
     let (network_2, key_2) = build_network_and_key(|router| router.add_rpc_service(server));
     let (event_loop_2, _handle_2) = builder.build(network_2.clone(), key_2);
 
-    let state_1 = event_loop_1.state.clone();
-
     tokio::spawn(event_loop_1.start());
     tokio::spawn(event_loop_2.start());
 


crates/sui-network/src/endpoint_manager.rs
```diff
@@ -57,3 +56,9 @@ pub enum EndpointId {
     // TODO: Implement support for updating consensus addresses via EndpointManager.
     // Consensus(NetworkPublicKey),
 }
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum AddressSource {
```
akichidis, line 62: Maybe leave a comment that the order of the enums here is important as it designates priority as well?

akichidis (approved):

aschran:
> While we retain addresses from various sources, do we try them in order or do the high priority sources always override the lower priority ones?

We intentionally only try addresses from the highest priority group, so that e.g. an override coming from admin server would take precedence over new committee addresses coming in after an epoch change

PR #25134, opened 2026-01-28
[jsonrpc] Change cursor type from ObjectID to String for ObjectsPage used in owned objects queries.

stefan-mysten:
## Description 

This PR updates the JSON RPC to use a String for cursor type instead of an ObjectID.

## Test plan 

Existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [x] JSON-RPC: The cursor type for querying owned objects (`suix_getOwnedObjects`) has changed from `ObjectID` to `String`. In practice, if cursors are passed to the query requests as they're received, the client should see no difference and would not require any changes.
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16126,7 +16126,6 @@ name = "sui-open-rpc"
 version = "1.65.0"
 dependencies = [
  "anyhow",
- "base64 0.21.7",
  "bcs",
  "clap",
  "fastcrypto",
--- a/crates/sui-open-rpc/Cargo.toml
+++ b/crates/sui-open-rpc/Cargo.toml
@@ -10,7 +10,6 @@ edition = "2024"
 workspace = true
 
 [dependencies]
-base64.workspace = true
 schemars.workspace = true
 serde.workspace = true
 serde_json.workspace = true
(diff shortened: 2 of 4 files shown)


crates/sui/tests/cli_tests.rs
```diff
@@ -3629,7 +3629,7 @@ async fn test_get_owned_objects_owned_by_address_and_check_pagination() -> Resul
                             .with_previous_transaction(),
                     ),
                 )),
-                cursor,
+                cursor.clone(),
```
stefan-mysten, line 3632: ObjectID has `Copy`, so the cursor now needs to be cloned as `String` does not have `Copy`.

crates/sui-json-rpc/src/indexer_api.rs
```diff
@@ -164,6 +165,16 @@ impl<R: ReadApiServer> IndexerApiServer for IndexerApi<R> {
             self.metrics.get_owned_objects_limit.observe(limit as f64);
             let SuiObjectResponseQuery { filter, options } = query.unwrap_or_default();
             let options = options.unwrap_or_default();
+            let cursor = match cursor {
```
amnn, line 168: Same comment as on the other PR -- it should still be a stringified `ObjectID`.

PR #25135, opened 2026-01-28
[jsonrpc] Change cursor type from ObjectID to String for DynamicFieldPage used in dynamic field queries.

stefan-mysten:
## Description 

This PR updates the JSON RPC to use a String for cursor type instead of an ObjectID for dynamic field queries.

## Test plan 

Existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [x] JSON-RPC: The cursor type for querying dynamic fields (`suix_getDynamicFields`) has changed from `ObjectID` to `String`. In practice, if cursors are passed to the query requests as they're received, the client should see no difference and would not require any changes.
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-open-rpc/src/examples.rs
+++ b/crates/sui-open-rpc/src/examples.rs
@@ -1169,9 +1169,10 @@ impl RpcExampleProvider {
             .map(Into::into)
             .collect::<Vec<_>>();
 
-        let next_cursor =
-            fastcrypto::encoding::Base64::from_bytes(&ObjectID::new(self.rng.r#gen()).into_bytes())
-                .encoded();
+        let next_cursor = fastcrypto::encoding::Base64::from_bytes(
+            &bcs::to_bytes(&ObjectID::new(self.rng.r#gen())).unwrap(),
+        )
+        .encoded();
 
         let page = DynamicFieldPage {
             data: dynamic_fields,


crates/sui-json-rpc/src/indexer_api.rs
```diff
@@ -345,19 +346,31 @@ impl<R: ReadApiServer> IndexerApiServer for IndexerApi<R> {
...
         parent_object_id: ObjectID,
         // If `Some`, the query will start from the next item after the specified cursor
-        cursor: Option<ObjectID>,
+        cursor: Option<String>,
         limit: Option<usize>,
     ) -> RpcResult<DynamicFieldPage> {
         with_tracing!(async move {
             let limit = cap_page_limit(limit);
             self.metrics.get_dynamic_fields_limit.observe(limit as f64);
+            let cursor = match cursor {
```
amnn, line 355: IIUC, previously the cursor format was the `ObjectID` in canonical (string) form, and this is changing it to a Base64-, BCS-encoded form. We don't want to do that. The idea is to change the Rust type to `String`, but the format remains the same, so that clients who were (admittedly incorrectly) relying on the cursor format could continue to do so.

amnn:
Note though that with Brandon's recent changes to the CLI, to support gRPC, we may not need to make this change after all, depending on what partners need vis-a-vis JSONRPC support (it's no longer a requirement for talking with the CLI).

stefan-mysten:
__

> Note though that with Brandon's recent changes to the CLI, to support gRPC, we may not need to make this change after all, depending on what partners need vis-a-vis JSONRPC support (it's no longer a requirement for talking with the CLI).

Indeed, I was not aware of the move to gRPC under the hood. That's awesome!

PR #25143, opened 2026-01-28
KV Store / Archival indexer refactor + schema changes

nickvikeras:
## Description

* Splitting into pipeline-per-table
* Making (backward-compatible) schema changes
  * Write txn signature and data as separate columns (keep writing old column)
  * Write epoch start/end as separate columns (keep writing old column)
  * New object type table (this data is already in object table, but this allows us to avoid loading the entire object for rendering type information in grpc/graphql's txn apis)
  * Continuing to write old watermark (as min of all pipeline watermarks)
* Use Bytes instead of Vec<u8> in the prost-generated code, because the framework sort of forces you to clone all of your data in the commit path to deal with retries.
 
This can be deployed without breaking any readers. The reader updates to read these new columns/watermarks will follow in a separate PR, and then we can finally stop writing and delete the old columns.

## Test plan

* New test that uses a mocked grpc server to test partial write failures.
* New emulator-based test (had to install the gcloud cli on the CI container to make this work).

### DIFF
--- a/crates/sui-kvstore/src/handlers/object_types.rs
+++ b/crates/sui-kvstore/src/handlers/object_types.rs
@@ -22,16 +22,18 @@ impl Processor for ObjectTypesPipeline {
 
     async fn process(&self, checkpoint: &Arc<Checkpoint>) -> anyhow::Result<Vec<Self::Value>> {
         let timestamp_ms = checkpoint.summary.timestamp_ms;
-        let mut entries = Vec::with_capacity(checkpoint.object_set.len());
+        let mut entries = Vec::new();
 
-        for obj in checkpoint.object_set.iter() {
-            let object_type = ObjectType::from(obj);
-            let entry = tables::make_entry(
-                tables::object_types::encode_key(&obj.id()),
-                tables::object_types::encode(&object_type)?,
-                Some(timestamp_ms),
-            );
-            entries.push(entry);
+        for tx in &checkpoint.transactions {
+            for obj in tx.created_objects(&checkpoint.object_set) {
+                let object_type = ObjectType::from(obj);
+                let entry = tables::make_entry(
+                    tables::object_types::encode_key(&obj.id()),
+                    tables::object_types::encode(&object_type)?,
+                    Some(timestamp_ms),
+                );
+                entries.push(entry);
+            }
         }
 
         Ok(entries)
--- a/crates/sui-types/src/full_checkpoint_content.rs
+++ b/crates/sui-types/src/full_checkpoint_content.rs
@@ -392,6 +392,16 @@ impl ExecutedTransaction {
                     .and_then(|version| object_set.get(&ObjectKey(change.id, version)))
             })
     }
+
+    pub fn created_objects<'a>(
+        &self,
+        object_set: &'a ObjectSet,
+    ) -> impl Iterator<Item = &'a Object> + 'a {
+        self.effects
+            .created()
+            .into_iter()
+            .filter_map(move |((id, version, _), _)| object_set.get(&ObjectKey(id, version)))
+    }
 }
 
 impl From<Checkpoint> for CheckpointData {


crates/sui-kvstore/src/lib.rs
```diff
@@ -1,29 +1,155 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Checkpoint {
+    pub summary: CheckpointSummary,
+    pub contents: CheckpointContents,
+    pub signatures: AuthorityStrongQuorumSignInfo,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct TransactionData {
+    pub transaction: Transaction,
```
bmwill, line 95: can we explicitly separate out `TransactionData` and `Signatures` here? This is both for clarity as well as to make sure we don't make the mistake of serializing these together again.

crates/sui-kvstore/src/lib.rs
```diff
@@ -1,29 +1,155 @@
...
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Checkpoint {
+    pub summary: CheckpointSummary,
+    pub contents: CheckpointContents,
+    pub signatures: AuthorityStrongQuorumSignInfo,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
bmwill: Do we actually serialize/deserialize these types you've defined here? If we don't (i think we just serialize/deserialize the individual parts of these structs) then we should remove the impls (and maybe add a comment indicating that we shouldn't implement serialize/deserialize on these types)

crates/sui-kvstore/src/lib.rs
```diff
@@ -1,29 +1,155 @@
...
+            system_state: info.system_state.clone(),
+        }
+    }
+}
+
+/// Data written at epoch end (after the last transaction of the epoch)
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct EpochEndData {
+    pub end_timestamp_ms: Option<u64>,
+    pub end_checkpoint: Option<u64>,
```
bmwill, line 122: For indexing these shouldn't ever be `None` right?

crates/sui-kvstore/src/lib.rs
```diff
@@ -1,29 +1,155 @@
...
+
+/// Data written at epoch start (before transactions are processed)
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct EpochStartData {
+    pub epoch: u64,
+    pub protocol_version: Option<u64>,
+    pub start_timestamp_ms: Option<u64>,
+    pub start_checkpoint: Option<u64>,
+    pub reference_gas_price: Option<u64>,
+    pub system_state: Option<sui_types::sui_system_state::SuiSystemState>,
```
bmwill, line 120: For indexing these shouldn't ever be `None` right?

crates/sui-kvstore/src/tables/epochs.rs
```diff
@@ -0,0 +1,47 @@
...
+use sui_types::storage::EpochInfo;
+
+use crate::tables::DEFAULT_COLUMN;
+use crate::{EpochEndData, EpochStartData};
+
+pub const NAME: &str = "epochs";
+
+pub mod col {
+    pub const START: &str = "start";
+    pub const END: &str = "end";
```
bmwill: Do we want to break out the individual components into their own columns vs having a `start` and `end` column which data that can't be expanded and evolved in the future?

crates/sui-kvstore/src/tables/transactions.rs
```diff
@@ -0,0 +1,135 @@
...
+pub fn decode(row: &[(Bytes, Bytes)]) -> Result<TransactionData> {
+    let mut transaction = None;
+    let mut effects = None;
+    let mut events = None;
+    let mut timestamp = 0;
+    let mut checkpoint_number = 0;
+
+    for (column, value) in row {
+        match column.as_ref() {
+            b"tx" => transaction = Some(bcs::from_bytes(value)?),
```
bmwill, line 93: Shouldn't we be reading from the new columns? if we need to do this for now while we're doing a backfill then we should add a comment here indicating that we need to change this eventually.

crates/sui-kvstore/src/handlers/objects.rs
```diff
@@ -0,0 +1,44 @@
...
+#[async_trait::async_trait]
+impl Processor for ObjectsPipeline {
+    const NAME: &'static str = "kvstore_objects";
+    type Value = Entry;
+
+    async fn process(&self, checkpoint: &Arc<Checkpoint>) -> anyhow::Result<Vec<Self::Value>> {
+        let timestamp_ms = checkpoint.summary.timestamp_ms;
+        let mut entries = Vec::with_capacity(checkpoint.object_set.len());
+
+        for object in checkpoint.object_set.iter() {
```
bmwill, line 27: doing this we'll end up with some redundant writes since this includes input and loaded unchanged objects as well. But this should be effectively idempotent so maybe not worth the optimization yet?

crates/sui-kvstore/src/lib.rs
nickvikeras: Yeah, I left that as a TODO somewhere in the code. I was gonna tackle it when I update the grpc server to actually request the fields separately based on read masks.

crates/sui-kvstore/src/tables/transactions.rs
nickvikeras: I change that in the stacked PR. It reads from the new columns and falls back to the old column and it will need to stay like that until the backfill is done.

crates/sui-kvstore/src/lib.rs
nickvikeras: I didn't add these types, I just reorganized the files to match the indexing & RPC's style preferences when refactoring. But I can try deleting the serde stuff and see if it still compiles. I don't think we actually serialize/deserialize it so you are right it shouldn't be there.

crates/sui-kvstore/src/lib.rs
nickvikeras: Yeah good catch, there is no reason these need to be Options.

crates/sui-kvstore/src/lib.rs
nickvikeras: ditto.

crates/sui-kvstore/src/tables/epochs.rs
nickvikeras: Ah yeah yeah I can split it up. My brain for some reason still hasn't fully registered that BCS isn't extensible. I was just looking at the query in the rpc API and saw the data was small and wasn't sure if it made sense to split it up into multiple columns to read separately.

crates/sui-kvstore/src/lib.rs
nickvikeras: Actually, if we are storing each field as it's own column in bigtable, I think we do want these to be Options so that we can query just a subset of the columns.

bmwill: yep i agree, my comment only makes sense if we keep the current schema

crates/sui-kvstore/src/bigtable/client.rs
```diff
@@ -807,11 +553,207 @@ impl BigTableClient {
...
+        transaction_digests: &[TransactionDigest],
+    ) -> Result<Vec<(TransactionDigest, TransactionEventsData)>> {
+        let query = self.multi_get(
+            tables::transactions::NAME,
+            transaction_digests
+                .iter()
+                .map(tables::transactions::encode_key)
+                .collect(),
+            Some(RowFilter {
+                filter: Some(Filter::ColumnQualifierRegexFilter(
```
bmwill, line 732: is it always a regex or is there a way to just provide exact columns?

crates/sui-kvstore/src/handlers/epochs_end.rs
amnn, line 1: The Postgres handler for epoch end information is indexing a lot more stuff, which we would need in order to switch over to the archival store instead of relying on this index in GraphQL:

https://github.com/MystenLabs/sui/blob/965fc3495c59ab67a5b3e4be37a384bf25eff440/crates/sui-indexer-alt-schema/src/schema.rs#L75-L93

crates/sui-kvstore/src/handlers/epochs_start.rs
amnn, line 1: This works well schema-wise for GraphQL, but does it record a record for genesis (epoch 0)?

crates/sui-kvstore/src/handlers/object_types.rs
amnn, line 1: IIUC, this pipeline assumes that an object's type cannot change once it is created, we make this assumption in one other place -- the `obj_info` pipeline of `sui-indexer-alt`, but this is one of the reasons why we're getting rid of this pipeline:

This assumption prevents us from allowing people to reclaim derived object IDs (because it would allow the same UID to be re-used under a different type).

For that reason, we should not introduce new pipelines that rely on this assumption. In this case, I think the fix is pretty simple: Adapt the pipeline to check whether the object was created or it was mutated and the mutation involves a type change.

Note that even today, dynamic fields may fail this test: a dynamic field's ID is derived from the parent ID, name type and name content. The value type and content don't impact the ID, which means a transaction can modify a field to change its value's type and it will be treated as a field mutation. If you query the type of the `Field<..., ...>` object corresponding to the dynamic field, you will then get a stale response.

crates/sui-kvstore/src/handlers/objects.rs
amnn: The efficient implementation of this is fairly simple:

https://github.com/MystenLabs/sui/blob/965fc3495c59ab67a5b3e4be37a384bf25eff440/crates/sui-indexer-alt/src/handlers/kv_objects.rs#L25-L52

(the `kv_objects` pipeline also records a sentinel for deleted objects, but that can be ignored).

crates/sui-kvstore/src/lib.rs
```diff
@@ -1,29 +1,137 @@
...
+pub use crate::handlers::EpochLegacyPipeline;
+pub use crate::handlers::EpochStartPipeline;
+pub use crate::handlers::ObjectTypesPipeline;
+pub use crate::handlers::ObjectsPipeline;
+pub use crate::handlers::PrevEpochUpdate;
+pub use crate::handlers::TransactionsPipeline;
+pub use crate::handlers::set_max_mutations;
+
+/// All pipeline names registered by the indexer. Single source of truth used for:
+/// - Pipeline registration in `BigTableIndexer::new()`
```
amnn, line 22: This at least doesn't seem to be true (this const is not used for pipeline registration) is it needed, and if so, is it meant to be here in the file? It's flanked on either side by imports.

crates/sui-kvstore/src/handlers/epochs_start.rs
nickvikeras: Yeah it does, confirmed the record is there in my test db which started out empty and ran from genesis

PR #25150, opened 2026-01-28
Remove logic unused after removing disable_preconsensus_locking flag

mwtian:
## Description 

Remove logic now unused, after `disable_preconsensus_locking` becomes `true`.
Tests are only removed if they don't make sense anymore, or are already covered by other tests.

## Test plan 

CI

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -1099,29 +1099,6 @@ impl AuthorityState {
         Ok(())
     }
 
-    /// This is a private method and should be kept that way. It doesn't check whether
-    /// the provided transaction is a system transaction, and hence can only be called internally.
-    #[instrument(level = "trace", skip_all)]
-    fn handle_transaction_impl(
-        &self,
-        transaction: VerifiedTransaction,
-        epoch_store: &Arc<AuthorityPerEpochStore>,
-    ) -> SuiResult<()> {
-        // Ensure that validator cannot reconfigure while we are validating the transaction.
-        let _execution_lock = self.execution_lock_for_validation()?;
-
-        let checked_input_objects =
-            self.handle_transaction_deny_checks(&transaction, epoch_store)?;
-
-        let owned_objects = checked_input_objects.inner().filter_owned_objects();
-
-        // Validate owned object versions without acquiring locks. Locking happens post-consensus
-        // in the consensus handler. Validation still runs to prevent spam transactions with
-        // invalid object versions, and is necessary to handle recently created objects.
-        self.get_cache_writer()
-            .validate_owned_object_versions(&owned_objects)
-    }
-
     /// Vote for a transaction, either when validator receives a submit_transaction request,
     /// or sees a transaction from consensus. Performs the same types of checks as
     /// transaction signing, but does not explicitly sign the transaction.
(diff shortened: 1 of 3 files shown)


crates/sui-core/src/authority.rs
```diff
@@ -1108,15 +1108,14 @@ impl AuthorityState {
         sign: bool,
```
akichidis: Since don't sign anymore maybe it's an opportunity to cleanup this as well?

akichidis (approved):
Changes look good to me and I approve to unblock. But @mwtian others should take a look as well as it's touching some sensitive parts - although unused now - of the system to make sure we are not missing something.

crates/sui-core/src/authority.rs
```diff
@@ -1105,52 +1105,21 @@ impl AuthorityState {
     fn handle_transaction_impl(
```
aschran: the name of this function is a bit confusing now that `handle_transaction` isn't a thing anymore and it looks like this is only called from a single spot within `handle_vote_transaction`. could the functions maybe be combined?

crates/sui-core/src/authority_server.rs
```diff
@@ -868,58 +867,29 @@ impl ValidatorService {
...
+                claims.push(TransactionClaim::ImmutableInputObjects(
+                    immutable_object_ids,
+                ));
+            }
 
-                let (tx, aliases) = verified_transaction.into_inner();
-                if epoch_store.protocol_config().address_aliases() {
-                    if epoch_store
-                        .protocol_config()
-                        .fix_checkpoint_signature_mapping()
```
aschran, line 891: thanks for doing this, if you want the cleanup for this flag to also be part of this PR can you also do the TODO in `transaction.rs`?

```
// TODO: Remove once `fix_checkpoint_signature_mapping` flag is enabled in testnet.
```

or if you'd rather I can also take care of the cleanup for this "fix" flag in a separate PR and you could put this part back for now, either way works

aschran: sui-protocol-config/src/lib.rs also has a small related TODO

```
// TODO: remove old code and deprecate once in mainnet.
```

could just replace that with a deprecation notice probably

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2719,12 +2697,8 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
                 // for transactions that will be dropped (e.g., during epoch change).
                 // Only applies to UserTransactionV2 - other transaction types don't need lock acquisition.
-                if self
-                    .epoch_store
-                    .protocol_config()
-                    .disable_preconsensus_locking()
-                    && let ConsensusTransactionKind::UserTransactionV2(tx_with_claims) =
-                        &parsed.transaction.kind
+                if let ConsensusTransactionKind::UserTransactionV2(tx_with_claims) =
+                    &parsed.transaction.kind
```
aschran, line 2701: if you're going to keep the `fix_checkpoint_signature_mapping` cleanup in this PR there's another spot where it is referenced below in this file

crates/sui-core/src/consensus_validator.rs
```diff
@@ -24,7 +24,8 @@ use sui_types::{
     error::{SuiError, SuiErrorKind, SuiResult, UserInputError},
     messages_consensus::{ConsensusPosition, ConsensusTransaction, ConsensusTransactionKind},
     transaction::{
-        InputObjectKind, PlainTransactionWithClaims, TransactionDataAPI, TransactionWithClaims,
+        CertifiedTransaction, InputObjectKind, PlainTransactionWithClaims, TransactionDataAPI,
```
aschran, line 27: (once more) if you're going to keep the `fix_checkpoint_signature_mapping` cleanup in this PR there's another spot where it is referenced below in this file

crates/sui-core/src/authority.rs
mwtian: Good idea.

crates/sui-core/src/authority_server.rs
mwtian: Chatted offline. I will leave the flag here for now.

PR #25159, opened 2026-01-29
[jsonrpc-alt] Migrate to read from consistent-store instead of `obj_info`

wlmyng:
## Description 

jsonrpc-alt uses `obj_info` to filter for owned objects by a type prefix, and to support queries for owned objects not of a certain type (slush requirement.) However, the latter is quite inefficient as it requires a sequential scan through the postgres db, and we have a consistent-store that can satisfy these queries, so this PR migrates jsonrpc-alt over to read from consistent-store instead of `obj_info` to unblock getting rid of the last consistent pipelines in postgres.

This PR implements a basic form of `MatchNone`, which just supports "owned objects not of this type".

We plan to support type inclusion/exclusion filters eventually.

From consistent-store up:
- `skip_prefix` in the forward and reverse iter to skip past a provided prefix
- in pagination.rs, `paginate_from_front` and its equivalent additionally skips past entries starting with `exclude_prefix`
- in consistent-store's `list_owned_objects.rs`, we assume a base prefix of `OwnerKind`, and a single exclusion filter on object type
- `ListOwnedObjectsRequest` has new field `optional string exclude_object_type = 3;`
- consistent service returns new error if both `object_type` and `exclude_object_type` are provided
- consistent reader `list_owned_objects` method has additional `exclude_object_type`

## Test plan 

Existing tests in `jsonrpc_compound_object_filter.rs`, unit tests in `iter.rs` and `pagination.rs`. 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-consistent-api/proto/sui/rpc/consistent/v1alpha/consistent_service.proto
```diff
@@ -145,6 +146,18 @@ message ListOwnedObjectsRequest {
...
+  // with `object_type`.
+  //
+  // Accepts filters by the type's package, module, fully-qualified name, or a
+  // type instantiation:
+  //
+  //  - `0x2`
+  //  - `0x2::coin`
+  //  - `0x2::coin::Coin`
+  //  - `0x2::coin::Coin<0x2::sui::SUI>`
+  optional string exclude_object_type = 3;
```
amnn: Let's not add this new field. Instead, extend support to `object_type` so that if it begins with an `!` it means it's a negated filter. That way when we extend the schema to support multiple queries, we only need to replace `optional` with `repeated` on `object_type` (i.e. these are all compatible schema changes).

crates/sui-indexer-alt-consistent-store/src/db/iter.rs
```diff
@@ -38,6 +38,23 @@ impl<'d, K, V> FwdIter<'d, K, V> {
         }
     }
 
+    /// Skip past all keys that start with the given `prefix`. The iterator will be positioned at
+    /// the first key that does not start with the `prefix`, or will be invalid if no such key
+    /// exists.
+    pub(crate) fn skip_prefix(&mut self, prefix: impl AsRef<[u8]>) {
```
amnn: Let's just call this `skip` because that's really what it is -- it just happens to work on prefixes.

crates/sui-indexer-alt-consistent-store/src/db/iter.rs
```diff
@@ -72,6 +89,21 @@ impl<'d, K, V> RevIter<'d, K, V> {
...
+    /// Skip past all keys that start with the given `prefix` in reverse order. The iterator will be
+    /// positioned at the last key that does not start with `prefix`, or will be invalid if no such
+    /// key exists.
+    pub(crate) fn skip_prefix(&mut self, prefix: impl AsRef<[u8]>) {
+        if let Some(inner) = &mut self.inner {
+            let end = prefix.as_ref();
+            inner.seek_for_prev(end);
+            // The only way to land on something in the prefix range is if the prefix itself is an
+            // exact key, so an exact match check is sufficient.
+            if inner.key() == Some(end) {
```
amnn, line 101: Why not add a `key::prev` to match `key::next`? It would save you an extra comparison.

crates/sui-indexer-alt-consistent-store/src/db/iter.rs
amnn: Ditto for the reverse iterator.

crates/sui-indexer-alt-consistent-store/src/db/mod.rs
```diff
@@ -1721,4 +1721,66 @@ pub(crate) mod tests {
...
+
+        // Insert keys: 0, 2, 4, 6, 8 with values 1, 3, 5, 7, 9
+        let mut batch = rocksdb::WriteBatch::default();
+        for i in (0u64..10).step_by(2) {
+            batch.put_cf(&cf, key::encode(&i), bcs::to_bytes(&(i + 1)).unwrap());
+        }
+        db.write("test", wm(0), batch).unwrap();
+        db.take_snapshot(wm(0));
+
+        let mut iter: iter::FwdIter<'_, u64, u64> = db.iter(0, &cf, (U::<u64>, U)).unwrap();
```
amnn: Would this work?


```suggestion
        let mut iter = db.iter(0, &cf, (U::<u64>, U::<u64>)).unwrap();
```

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/dynamic_fields/dynamic_fields.move
amnn, line 1: Why change the dynamic field tests?

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/objects/query/by_module.move
amnn, line 1: Ditto, why has this changed?

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/objects/query/filter_errors.snap
amnn, line 1: Just confirming we've double checked with Slush that this set of features is sufficient for their current needs.

crates/sui-indexer-alt-consistent-store/src/rpc/pagination.rs
```diff
@@ -102,12 +103,47 @@ impl<'r> Page<'r> {
...
             self.paginate_from_back(
                 map.prefix_rev(checkpoint, prefix)
                     .map_err(|e| db_error(e, "failed to create reverse iterator"))?,
                 pred,
+                None,
+            )
+        }
+    }
+
+    /// Paginate over the key-value pairs in `map` that start with the `base_prefix`, excluding
```
amnn, line 118: Let's add some kind of note here with the TODO for the more general thing (where we are iterating/paginating over multiple inclusion/exclusion regions.

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/dynamic_fields/dynamic_fields.move
amnn: Answering this and others for myself: these tests have been moved into a Rust test where it's easier to deal with the more complicated cursor format, is that right?

wlmyng: Yes that's right - it's a bit of a nuisance to cook these by hand

crates/sui-indexer-alt-consistent-store/src/db/iter.rs
wlmyng: I can do that to be more symmetric with the forward case and avoid the conditional check. my initial observation was that we could just use `seek_for_prev`

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/objects/query/filter_errors.snap
wlmyng: Confirmed with @tlmysten, only usage is

```
const response = await client.getOwnedObjects({
			owner: owner.address,
			filter: {
				MatchNone: [{ StructType: '0x2::coin::Coin' }],
			},
			options: {
				showDisplay: true,
				showType: true,
				showContent: true,
			},
			limit,
			cursor: currentCursor || undefined,
		});
```

crates/sui-indexer-alt-consistent-store/src/db/iter.rs
wlmyng: oh but this clashes with Iterator::skip(self, usize)

```
error[E0308]: mismatched types
    --> crates/sui-indexer-alt-consistent-store/src/rpc/pagination.rs:199:27
     |
 199 |                 iter.skip(excl);
     |                      ---- ^^^^ expected `usize`, found `&[u8]`
     |                      |
     |                      arguments to this method are incorrect
     |
note: method defined here
    --> /Users/williamyang/.rustup/toolchains/1.92-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:1316:8
     |
1316 |     fn skip(self, n: usize) -> Skip<Self>
     |        ^^^^

```

wlmyng: maybe `skip_past`? a little awkward for the rev case but hopefully excusable

crates/sui-indexer-alt-consistent-store/src/db/mod.rs
wlmyng: It seems I need the type annotation on `iter`

crates/sui-indexer-alt-consistent-store/src/db/iter.rs
wlmyng: Oh, there's an asymmetry ... if we look for the last key `<= [0, 1]`, with keys `[0, 1, 0, 1]`, `[0, 1, 0, 2]` etc, since `[0, 1, 0, 1] > [0, 1]`, no key `<= [0, 1]` exists, so the iterator becomes invalidated.

PR #25162, opened 2026-01-29
[Consensus] update peer address via EndpointManager

akichidis:
## Description 

Add the ability to update a consensus peer address via the EndpointManager and the Admin server. The new address is stored separately as an override. Once the new address is provided then the node will attempt to re-subscribe to the new address and also invalidate the previous connection and establish new one for the TonicClient.

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/consensus_manager/mod.rs
+++ b/crates/sui-core/src/consensus_manager/mod.rs
@@ -19,7 +19,7 @@ use fastcrypto::traits::KeyPair as _;
 use mysten_metrics::{RegistryID, RegistryService};
 use mysten_network::Multiaddr;
 use prometheus::{IntGauge, Registry, register_int_gauge_with_registry};
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
@@ -48,11 +48,11 @@ enum Running {
 
 /// Stores address updates that should be persisted across epoch changes.
 /// We store the consensus NetworkPublicKey to avoid repeated conversions.
-/// Using BTreeMap since ConsensusNetworkPublicKey implements Ord but not Hash.
 struct AddressOverridesMap {
+    // We store the AddressSource on a BTreeMap as it helps accessing the keys in priority order.
     map: BTreeMap<
         ConsensusNetworkPublicKey,
-        HashMap<sui_network::endpoint_manager::AddressSource, Vec<Multiaddr>>,
+        BTreeMap<sui_network::endpoint_manager::AddressSource, Vec<Multiaddr>>,
     >,
 }
 
@@ -97,7 +97,7 @@ impl AddressOverridesMap {
     ) -> Option<Multiaddr> {
         self.map
             .get(&network_pubkey)
-            .and_then(|sources| sources.iter().min_by_key(|(source, _)| *source))
+            .and_then(|sources| sources.first_key_value())
             .and_then(|(_, addresses)| addresses.first().cloned())
     }
 
@@ -107,7 +107,7 @@ impl AddressOverridesMap {
         let mut result = Vec::new();
 
         for (network_pubkey, sources) in self.map.iter() {
-            if let Some((_source, addresses)) = sources.iter().min_by_key(|(source, _)| *source)
+            if let Some((_source, addresses)) = sources.first_key_value()
                 && let Some(address) = addresses.first()
             {
                 result.push((network_pubkey.clone(), address.clone()));


consensus/core/src/network/tonic_network.rs
```diff
@@ -357,13 +361,50 @@ struct ChannelPool {
...
+
+            // Treat the empty list as a clear operation.
+            if addresses.is_empty() {
+                overrides.remove(&peer);
+                info!("Cleared address override for peer {}", peer);
+                return;
+            }
+
+            // Otherwise, set the first address as the override.
+            // TODO: support multiple addresses. For now, we only support one address per peer.
```
mwtian: Let's file a task because this is a big missing piece.

mwtian (approved):
I think later we will need an integration test to make sure nodes running at the new address can get connected.

crates/sui-network/src/endpoint_manager.rs
```diff
@@ -1,29 +1,61 @@
...
 /// other nodes in the network.
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct EndpointManager {
+    inner: Arc<Inner>,
+}
+
+struct Inner {
     discovery_handle: discovery::Handle,
+    consensus_address_updater: ArcSwapOption<Arc<dyn ConsensusAddressUpdater>>,
```
aschran, line 24: unless we think we are going to add a lot more non-clonable things to the EndpointManager struct, it might be simpler to just wrap this in an `Arc` rather than have a separate `Inner` struct thing? wdyt?

crates/sui-node/src/lib.rs
```diff
@@ -2426,7 +2433,7 @@ fn update_peer_addresses(
     for (peer_id, address) in
         epoch_start_state.get_validator_as_p2p_peers(config.protocol_public_key())
     {
-        endpoint_manager.update_endpoint(EndpointId::P2p(peer_id), vec![address]);
+        let _ = endpoint_manager.update_endpoint(EndpointId::P2p(peer_id), vec![address]);
```
aschran: we prob should not be ignoring errors here?

crates/sui-network/src/discovery/tests.rs
```diff
@@ -239,7 +239,7 @@ async fn peers_are_added_from_endpoint_manager() -> Result<()> {
     let peer2_addr: Multiaddr = format!("/dns/localhost/udp/{}", network_2.local_addr().port())
         .parse()
         .unwrap();
-    endpoint_manager_1.update_endpoint(
+    let _ = endpoint_manager_1.update_endpoint(
```
aschran, line 242: we probably should not be ignoring errors here?

crates/sui-core/src/consensus_manager/mod.rs
```diff
@@ -316,6 +319,25 @@ impl ConsensusManager {
...
+// Implementing the interface so we can update the consensus peer addresses when requested.
+impl ConsensusAddressUpdater for ConsensusManager {
+    fn update(&self, network_pubkey: NetworkPublicKey, addresses: Vec<Multiaddr>) -> SuiResult<()> {
+        if let Some(authority) = self.authority.load_full() {
+            let network_pubkey = consensus_config::NetworkPublicKey::new(network_pubkey);
+            authority.0.update_peer_address(network_pubkey, addresses);
+            Ok(())
+        } else {
+            warn!(
+                "Consensus authority node is not running, ignoring update of peer addresses for network public key {network_pubkey:?} and addresses {addresses:?}"
```
aschran: is it safe to ignore updates here? could you get into a race where you have an update that you want to be applied once consensus starts, but it gets dropped here because it came too soon?

consensus/core/src/network/tonic_network.rs
```diff
@@ -357,13 +361,50 @@ struct ChannelPool {
     context: Arc<Context>,
     // Size is limited by known authorities in the committee.
     channels: RwLock<BTreeMap<AuthorityIndex, Channel>>,
+    // Address overrides for peers, indexed by AuthorityIndex
+    address_overrides: RwLock<BTreeMap<AuthorityIndex, Multiaddr>>,
```
aschran, line 365: this will probably need some refactoring to work with the source-based priorities

maybe it's worth moving that logic to track which address group currently has priority into `EndpointManager`, instead of duplicating it in both the p2p network code and the consensus code? but I will leave that up to you

crates/sui-network/src/endpoint_manager.rs
akichidis: That's how I had this initially, but especially with the `EndpointManager` deriving `Clone`, I believe it would be simple to have the `Inner` struct now and extend as we add more things. If it's not too much of an issue from your point of view, I would suggest to keep as is.

crates/sui-core/src/consensus_manager/mod.rs
akichidis: The method returns an error and basically I am thinking it's the responsibility of the caller to retry or take further decisions. Now, for the use cases we are looking at I guess this will be quite an edge case?

consensus/core/src/network/tonic_network.rs
akichidis: Ah ok let me rebase and see what can I do here

crates/sui-core/src/consensus_manager/mod.rs
aschran: It's fine if retries go in the caller, I guess they just have to go _somewhere_. My question about how to handle the source-based priorities (and whether to move the logic into the EndpointManager) might be relevant to this too. Discovery/p2p runs continuously so it does not have any problem with maintaining high priority overrides across epochs, but for consensus you're presumably going to have to save the overrides somewhere to reapply them after epoch change, is that right?

akichidis: OK this might need a bit more thought/design then. Because I assumed that all the address update changes - even after epoch change - will be driven by the node it self given that consensus is ephemeral . If those _should_ be persisted across epoch change then yes this will need some redesign.

consensus/core/src/network/mod.rs
```diff
@@ -193,8 +193,8 @@ where
     async fn stop(&mut self);
 
     /// Updates the network address for a peer identified by their authority index.
-    /// If addresses is empty, any address override is cleared.
-    fn update_peer_address(&self, peer: AuthorityIndex, addresses: Vec<Multiaddr>);
+    /// If address is None, the override is cleared and the committee address will be used.
```
aschran: there are multiple override sources and clearing one source should use the next-lower priority source that has addresses, which may or may not be the committee (for example it could be that an `Admin` address is cleared and the next-lower priority available is `Config`)

consensus/core/src/network/tonic_network.rs
```diff
@@ -374,31 +374,26 @@ impl ChannelPool {
...
-                info!(
-                    "Updated address override for peer {} from {} to {}",
-                    peer, previous_address, address
-                );
-            } else {
-                info!("Set address override for peer {} to {}", peer, address);
             }
         }
 
+        // Clear the cached channel so that the next connection attempt uses the updated address.
```
aschran, line 396: will this proactively reconnect, or only if the existing connection is lost?

crates/sui-core/src/consensus_manager/mod.rs
```diff
@@ -42,6 +46,77 @@ enum Running {
     False,
 }
 
+/// Stores address updates that should be persisted across epoch changes.
+/// We store the consensus NetworkPublicKey to avoid repeated conversions.
+/// Using BTreeMap since ConsensusNetworkPublicKey implements Ord but not Hash.
```
aschran: seems like it should be trivial to impl Hash on this if you need? `Ed25519PublicKey` implements `Hash`

consensus/core/src/network/tonic_network.rs
akichidis: we basically do a lazy initialisation whenever we make a request. So next time we try to perform any sort of request clean up will make sure that the override logic will re-run and if there is an address override it will be applied.

crates/sui-core/src/consensus_manager/mod.rs
akichidis: ah please ignore the comment - I was actually meant to use BTreeMap anyways. Any performance diff is negligible.

PR #24653, opened 2025-12-16
sui-indexer-alt: update version of bb8, add connect_started, connect_pending, connect_completed metrics

evan-wall-mysten:
## Description 

This PR updates bb8 which allows publishing `connect_started`, `connect_pending`, and `connect_completed` metrics. 

Currently `connect_pending` does not account for connection acquiring futures that were started ([here](https://github.com/djc/bb8/blob/19635b56a4cd9d1ef728b6f0df26d959d06d9bb3/bb8/src/inner.rs#L122)), but canceled before one of the completed metrics is published ([here](https://github.com/djc/bb8/blob/19635b56a4cd9d1ef728b6f0df26d959d06d9bb3/bb8/src/inner.rs#L132)).

## Test plan 

Verified metrics are being published in grafana
image

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #25163, opened 2026-01-29
sui-indexer-alt: Add metrics for connection pool acquisition attempts

evan-wall-mysten:
## Description 

This PR adds new connection pool metrics:
* `connections_requested` - Attempts to acquire a connection
* `connections_acquired` - Attempts that successfully acquired a connection.
* `connections_unacquired_error` - Attempts that failed to acquire a connection because of an error from the pool.
* `connections_unacquired_canceled` - Attempts that failed to acquire a connection because they were canceled before acquiring a connection.

These metrics are useful for knowing if the connection pool is able to handle the request throughput:
* If the connection pool is rightsized for the request throughput, `connections_requested` should be equal to or slightly above `connections_acquired + connections_unacquired_error + connections_unacquired_canceled` as there is a very small delay between requesting and acquiring a connection.
* If the connection pool is undersized for the request throughput, `connections_requested` will increase faster than `connections_acquired + connections_unacquired_error + connections_unacquired_canceled` until eventually `connections_unacquired_error` or `connections_unacquired_canceled` begin to increase as requests time out before connections become available.

## Test plan 

New unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-pg-db/src/lib.rs
+++ b/crates/sui-pg-db/src/lib.rs
@@ -4,6 +4,7 @@
 use std::ops::Deref;
 use std::ops::DerefMut;
 use std::path::PathBuf;
+use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::anyhow;
@@ -21,6 +22,7 @@ use diesel_async::pooled_connection::bb8::PooledConnection;
 use diesel_migrations::EmbeddedMigrations;
 use diesel_migrations::embed_migrations;
 use futures::FutureExt;
+use prometheus::Registry;
 use tracing::info;
 use url::Url;
 
@@ -28,6 +30,7 @@ use crate::tls::AsyncPgConnectionWithId;
 use crate::tls::build_tls_config;
 use crate::tls::establish_tls_connection;
 
+mod metrics;
 mod model;
 pub mod query;
 pub mod schema;
@@ -35,6 +38,7 @@ pub mod store;
 pub mod temp;
 mod tls;
 
+use crate::metrics::PoolMetrics;
 pub use sui_field_count::FieldCount;
 pub use sui_sql_macro::sql;
 
@@ -65,7 +69,16 @@ pub struct DbArgs {
 }
 
 #[derive(Clone)]
-pub struct Db(Pool<AsyncPgConnectionWithId>);
+pub struct Db {
+    pool: Pool<AsyncPgConnectionWithId>,
+    pool_metrics: Option<Arc<PoolMetrics>>,
+}
+
+// Ensures that unacquired_canceled is incremented if neither acquired nor unacquired_error are so
+// that one of acquired, unacquired_error, unacquired_canceled is incremented for every requested
+// connection. This is used to be able to calculate the number of pending connections:
+// pending = requested - (acquired + unacquired_error + unacquired_canceled)
+struct CancelGuard<'m>(Option<&'m PoolMetrics>);
 
 /// Wrapper struct over the remote `PooledConnection` type for dealing with the `Store` trait.
 pub struct Connection<'a>(PooledConnection<'a, AsyncPgConnectionWithId>);
@@ -84,25 +97,60 @@ impl Db {
     /// Construct a new DB connection pool talking to the database at `database_url` that supports
     /// write and reads. Instances of [Db] can be cloned to share access to the same pool.
     pub async fn for_write(database_url: Url, config: DbArgs) -> anyhow::Result<Self> {
-        Ok(Self(pool(database_url, config, false).await?))
+        Self::new(database_url, config, false).await
     }
 
     /// Construct a new DB connection pool talking to the database at `database_url` that defaults
     /// to read-only transactions. Instances of [Db] can be cloned to share access to the same
     /// pool.
     pub async fn for_read(database_url: Url, config: DbArgs) -> anyhow::Result<Self> {
-        Ok(Self(pool(database_url, config, true).await?))
+        Self::new(database_url, config, true).await
+    }
+
+    async fn new(database_url: Url, db_args: DbArgs, read_only: bool) -> anyhow::Result<Self> {
+        Ok(Db {
+            pool: pool(database_url, db_args, read_only).await?,
+            pool_metrics: None,
+        })
+    }
+
+    pub fn register_metrics(
+        mut self,
+        prefix: Option<&str>,
+        registry: &Registry,
+    ) -> anyhow::Result<Self> {
+        let pool_metrics = PoolMetrics::new(prefix, registry)?;
+        self.pool_metrics = Some(pool_metrics);
+        Ok(self)
     }
 
     /// Retrieves a connection from the pool. Can fail with a timeout if a connection cannot be
     /// established before the [DbArgs::connection_timeout] has elapsed.
     pub async fn connect(&self) -> anyhow::Result<Connection<'_>> {
-        Ok(Connection(self.0.get().await?))
+        if let Some(pool_metrics) = &self.pool_metrics {
+            let guard = CancelGuard::request(pool_metrics);
+            match self.pool.get().await {
+                Ok(c) => {
+                    guard.acquired();
+                    Ok(Connection(c))
+                }
+                Err(e) => {
+                    guard.unacquired_error();
+                    Err(e.into())
+                }
+            }
+        } else {
+            Ok(Connection(self.pool.get().await?))
+        }
     }
 
     /// Statistics about the connection pool
     pub fn state(&self) -> bb8::State {
-        self.0.state()
+        self.pool.state()
+    }
+
+    pub fn pool_metrics(&self) -> Option<&PoolMetrics> {
+        self.pool_metrics.as_ref().map(|p| p.as_ref())
     }
 
     async fn clear_database(&self) -> anyhow::Result<()> {
@@ -167,7 +215,7 @@ impl Db {
         let merged_migrations = merge_migrations(migrations);
 
         info!("Running migrations ...");
-        let conn = self.0.dedicated_connection().await?;
+        let conn = self.pool.dedicated_connection().await?;
         let mut wrapper: AsyncConnectionWrapper<AsyncPgConnectionWithId> =
             AsyncConnectionWrapper::from(conn);
 
@@ -184,6 +232,25 @@ impl Db {
     }
 }
 
+impl<'m> CancelGuard<'m> {
+    fn request(stats: &'m PoolMetrics) -> Self {
+        stats.requested.inc();
+        Self(Some(stats))
+    }
+
+    fn acquired(mut self) {
+        if let Some(m) = self.0.take() {
+            m.acquired.inc()
+        }
+    }
+
+    fn unacquired_error(mut self) {
+        if let Some(m) = self.0.take() {
+            m.unacquired_error.inc()
+        }
+    }
+}
+
 impl Default for DbArgs {
     fn default() -> Self {
         Self {
@@ -196,13 +263,21 @@ impl Default for DbArgs {
     }
 }
 
+impl<'m> Drop for CancelGuard<'m> {
+    fn drop(&mut self) {
+        if let Some(m) = self.0.take() {
+            m.unacquired_canceled.inc()
+        }
+    }
+}
+
 /// Drop all tables, and re-run migrations if supplied.
 pub async fn reset_database(
     database_url: Url,
-    db_config: DbArgs,
+    db_args: DbArgs,
     migrations: Option<&'static EmbeddedMigrations>,
 ) -> anyhow::Result<()> {
-    let db = Db::for_write(database_url, db_config).await?;
+    let db = Db::for_write(database_url, db_args).await?;
     db.clear_database().await?;
     if let Some(migrations) = migrations {
         db.run_migrations(Some(migrations)).await?;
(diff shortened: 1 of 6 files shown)


crates/sui-indexer-alt-metrics/src/db.rs
```diff
@@ -71,6 +73,34 @@ impl DbConnectionStatsCollector {
...
                 ),
             ),
+            (
+                MetricType::COUNTER,
+                desc(
+                    name("connections_requested"),
+                    "Total requested connections from the pool",
+                ),
+            ),
+            (
```
wlmyng: maybe I'm missing something, but I think some of these metrics are covered by bb8's statistics:

- `requested <- get_direct + get_waited + get_timed_out`
- acquired = get_direct + get_waited

`unacquired_error` might be interesting, although perhaps also covered by `Statistics` depending on the intended use case

`unacquired_canceled` is interesting as this is a caller-side metric, since bb8 wouldn't know when a caller drops the future mid-flight

crates/sui-pg-db/src/store.rs
```diff
@@ -222,7 +222,7 @@ impl store::Store for Db {
     type Connection<'c> = Connection<'c>;
 
     async fn connect<'c>(&'c self) -> anyhow::Result<Self::Connection<'c>> {
-        Ok(Connection(self.0.get().await?))
+        self.connect().await
```
wlmyng, line 225: oh, nice catch

crates/sui-pg-db/src/lib.rs
```diff
@@ -65,7 +73,65 @@ pub struct DbArgs {
...
+
+#[pinned_drop]
+impl<F> PinnedDrop for PoolFuture<F> {
+    fn drop(self: Pin<&mut Self>) {
+        inc_if_false(&self.completed, &self.pool_stats.unacquired_canceled);
+    }
+}
+
+// Ensure the counter is only incremented once per bool instance.
+fn inc_if_false(bool: &AtomicBool, counter: &AtomicU64) {
```
wlmyng: it seems this largely makes the second increment in `fn connect()` a no-op? `PoolFuture::poll()` increments stats

crates/sui-pg-db/src/lib.rs
```diff
@@ -84,28 +150,57 @@ impl Db {
...
     /// established before the [DbArgs::connection_timeout] has elapsed.
     pub async fn connect(&self) -> anyhow::Result<Connection<'_>> {
-        Ok(Connection(self.0.get().await?))
+        self.pool_stats.requested.fetch_add(1, Ordering::Relaxed);
+        let future = PoolFuture {
+            completed: Default::default(),
+            pool_stats: self.pool_stats.clone(),
+            inner: self.pool.get(),
+        };
+        let completed = future.completed.clone();
```
wlmyng: I think we can avoid the `Future impl`, `AtomicBool`, `pin_project` with some kind of drop guard? May be incorrect but I think when a future is canceled mid-await Rust will still call drop

Could use scopeguard or something homemade

```
  struct CancelGuard<'a> {                                                                                                                                                                                                                                                                                                                                                   
      stats: &'a PoolStats,                                                                                                                                                                                                                                                                                                                                                  
      disarmed: bool,                                                                                                                                                                                                                                                                                                                                                        
  }                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                             
  impl Drop for CancelGuard<'_> {                                                                                                                                                                                                                                                                                                                                            
      fn drop(&mut self) {                                                                                                                                                                                                                                                                                                                                                   
          if !self.disarmed {                                                                                                                                                                                                                                                                                                                                                
              self.stats.unacquired_canceled.fetch_add(1, Ordering::Relaxed);                                                                                                                                                                                                                                                                                                
          }                                                                                                                                                                                                                                                                                                                                                                  
      }                                                                                                                                                                                                                                                                                                                                                                      
  }                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                             
  pub async fn connect(&self) -> anyhow::Result<Connection<'_>> {                                                                                                                                                                                                                                                                                                            
      self.pool_stats.requested.fetch_add(1, Ordering::Relaxed);                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                             
      let mut guard = CancelGuard { stats: &self.pool_stats, disarmed: false };                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                             
      match self.pool.get().await {                                                                                                                                                                                                                                                                                                                                          
          Ok(c) => {                                                                                                                                                                                                                                                                                                                                                         
              self.pool_stats.acquired.fetch_add(1, Ordering::Relaxed);                                                                                                                                                                                                                                                                                                      
              guard.disarmed = true;                                                                                                                                                                                                                                                                                                                                         
              Ok(Connection(c))                                                                                                                                                                                                                                                                                                                                              
          }                                                                                                                                                                                                                                                                                                                                                                  
          Err(e) => {                                                                                                                                                                                                                                                                                                                                                        
              self.pool_stats.unacquired_error.fetch_add(1, Ordering::Relaxed);                                                                                                                                                                                                                                                                                              
              guard.disarmed = true;                                                                                                                                                                                                                                                                                                                                         
              Err(e.into())                                                                                                                                                                                                                                                                                                                                                  
          }                                                                                                                                                                                                                                                                                                                                                                  
      }                                                                                                                                                                                                                                                                                                                                                                      
  }
```

wlmyng: I think `InflightGuard` in mysten-metrics would be a good ref, only we have the `disarmed` flag to distinguish cancel from success/ error

crates/sui-indexer-alt-metrics/src/db.rs
evan-wall-mysten: `get_direct + get_waited + get_timed_out` are only triggered after the connection is acquired: https://github.com/djc/bb8/blob/v0.8.5/bb8/src/inner.rs#L133

A newer version of bb8 has a `get_started` metric which is the equivalent of `connections_requested`, but that version does not have an equivalent of `connections_unacquired_canceled` so cancellations are unaccounted for. I tested it out in https://github.com/MystenLabs/sui/pull/24653, but canceled that in favor of this PR.

crates/sui-pg-db/src/lib.rs
evan-wall-mysten: This prevents `drop` from incrementing `connections_unacquired_canceled` if `connections_acquired` or  `connections_unacquired_error` was already incremented. The extra bool is needed because it is impossible to compare the old and new value of `connections_acquired` and `connections_unacquired_error` since they may have been incremented by another request.

crates/sui-pg-db/src/lib.rs
evan-wall-mysten: I modeled this after `MetricsFuture`: https://github.com/MystenLabs/sui/blob/de97e292c9d2b884a03e9b5c0aa2bf59b44a9801/crates/sui-indexer-alt-graphql/src/extensions/logging.rs#L195-L248

The `CancelGuard` impl looks simpler though. @amnn is there a reason to use a `Future` with `PinnedDrop` instead of this?

crates/sui-pg-db/src/store.rs
amnn: Huh -- how is this not an infinite loop?

crates/sui-pg-db/src/lib.rs
amnn: No good reason -- a guard type is better in this case. You can still adopt the pattern of holding an `Option<PoolMetrics>` that you then `take` from, instead of holding a separate `complete` flag.

I think the same is also true of `MetricsFuture` in the logging extension. It looks like that because it's probably derived from a tower middleware which needs to define a future, but middleware works differently for GraphQL and you could just wrap the metrics in a guard.

crates/sui-indexer-alt-metrics/src/db.rs
```diff
@@ -114,6 +150,16 @@ impl Collector for DbConnectionStatsCollector {
...
+            counter(&self.desc[7].1, requested.load(Ordering::Relaxed) as f64),
+            counter(&self.desc[8].1, acquired.load(Ordering::Relaxed) as f64),
+            counter(
+                &self.desc[9].1,
+                unacquired_error.load(Ordering::Relaxed) as f64,
+            ),
+            counter(
+                &self.desc[10].1,
+                unacquired_canceled.load(Ordering::Relaxed) as f64,
+            ),
```
amnn: We only introduce metrics like this when we don't control where they are incremented/decremented, which is not the case here. We can/should use regular metrics for this, rather than the `DbConnectionStatsCollector`.

crates/sui-pg-db/src/lib.rs
```diff
@@ -65,7 +73,65 @@ pub struct DbArgs {
...
+impl<F> PinnedDrop for PoolFuture<F> {
+    fn drop(self: Pin<&mut Self>) {
+        inc_if_false(&self.completed, &self.pool_stats.unacquired_canceled);
+    }
+}
+
+// Ensure the counter is only incremented once per bool instance.
+fn inc_if_false(bool: &AtomicBool, counter: &AtomicU64) {
+    if !bool.swap(true, Ordering::Relaxed) {
+        counter.fetch_add(1, Ordering::Relaxed);
```
amnn: This is not safe -- the purpose of atomic orderings is to provide some guarantees about ordering relative to an operation across threads. `Relaxed` provides no guarantees, but it seems like you are explicitly using this bool to communicate a state change across threads. That implies that you should be using `AcqRel` ordering for this swap.

I suspect that you don't need to use an atomic bool at all here. Following @wlmyng's suggestion to introduce a guard type would address that.

crates/sui-pg-db/src/lib.rs
```diff
@@ -12,7 +19,6 @@ use diesel::migration::Migration;
 use diesel::migration::MigrationSource;
 use diesel::migration::MigrationVersion;
 use diesel::pg::Pg;
-use diesel_async::RunQueryDsl;
```
amnn, line 15: Why remove this here?

crates/sui-pg-db/src/store.rs
evan-wall-mysten: I think it is because of this https://doc.rust-lang.org/book/ch20-02-advanced-traits.html#disambiguating-between-identically-named-methods which says rust will prefer the method from the struct (`impl Db` which also has a method `connect`) over the trait (`impl Store for Db`).

crates/sui-pg-db/src/lib.rs
evan-wall-mysten: This was interferring with the `load` method on `AtomicU64`, but I'm not using that anymore so I'll put it back.

crates/sui-indexer-alt-metrics/src/db.rs
evan-wall-mysten: I switched to use the prometheus metrics directly.

crates/sui-pg-db/src/lib.rs
evan-wall-mysten: > Relaxed provides no guarantees, but it seems like you are explicitly using this bool to communicate a state change across threads. That implies that you should be using AcqRel ordering for this swap.

Thanks, that's good to know for future changes.

> I suspect that you don't need to use an atomic bool at all here. Following @wlmyng's suggestion to introduce a guard type would address that.

After switching to @wlmyng's suggestion and using metrics directly instead of atomic + db collector, this problem no longer exists.

crates/sui-pg-db/src/lib.rs
```diff
@@ -40,6 +45,14 @@ pub use sui_sql_macro::sql;
 
 pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!("migrations");
 
+#[derive(Debug, Clone)]
+pub struct DbConfig<'a> {
+    database_url: Url,
+    db_args: DbArgs,
+    read_only: bool,
+    registry: Option<&'a Registry>,
```
amnn: It is a bit confusing to introduce the prometheus registry here -- everywhere else `Config` types are plain old data.

It looks like the reason to do this is to offer an API, `Db::register_stats(&Registry)` to optionally fill out the `PoolStats`. That way you can avoid updating all the `Db` constructor call sites as well.

crates/sui-pg-db/src/lib.rs
```diff
@@ -81,28 +138,81 @@ impl DbArgs {
...
+            database_url,
+            db_args,
+            read_only,
+            registry,
+        }: DbConfig<'_>,
+    ) -> anyhow::Result<Self> {
+        Ok(Db {
+            pool: pool(database_url, db_args, read_only).await?,
+            pool_stats: registry.map(|registry| {
+                let register = |name: &str, help: &str| -> GenericCounter<AtomicU64> {
```
amnn: Let's align this with how other metrics sets are defined, for consistency/discoverability purposes:

- Can we use the macros supplied with `prometheus` instead of defining a new closure? (Check out how the other metrics structs are defined).
- Maybe call this `PoolMetrics` for consistency as well.
- ...and define a `PoolMetrics::new(&Registry): Arc<PoolMetrics>` 
- ...and perhaps move this into a `metrics` module for this crate.

crates/sui-pg-db/src/lib.rs
```diff
@@ -65,11 +78,55 @@ pub struct DbArgs {
...
+    pool_stats: Option<Arc<PoolStats>>,
+}
+
+// Ensures that unacquired_canceled is incremented if neither acquired nor unacquired_error are so
+// that one of acquired, unacquired_error, unacquired_canceled is incremented for every requested
+// connection. This is used to be able to calculate the number of pending connections:
+// pending = requested - (acquired + unacquired_error + unacquired_canceled)
+struct CancelGuard<'a> {
+    stats: &'a PoolStats,
+    disarmed: bool,
```
amnn: Here's a way to abstract this in Rust so the type system is doing more work for you to prove that it's impossible to break the invariants here (exactly one of `acquired`, `unacquired_error`, `unacquired_canceled` called by the time the guard is dropped).

```rust
struct CancelGuard<'m>(Option<&'m PoolMetrics>);

impl<'m> CancelGuard<'m> {
    fn request(metrics: &'m PoolMetrics) -> Self { 
        metrics.requested.inc();
        Self(Some(metrics))
    }

    fn acquired(mut self) {
        self.0.take().map(|m| m.acquired.inc());
    }

    fn errored(mut self) {
        self.0.take().map(|m| m.unacquired_error.inc());
    }
}

impl<'m> Drop for CancelGuard<'m> {
    fn drop(&mut self) {
        self.0.take().map(|m| m.unacquired_canceled.inc());
    }
}
```

PR #25188, opened 2026-01-30
[consensus] disable periodic sync only when commit sync is not progressing

mwtian:
## Description 

Use periodic sync as the fallback when commit sync is not progressing for various reasons.

## Test plan 

CI

### DIFF
--- a/consensus/core/src/synchronizer.rs
+++ b/consensus/core/src/synchronizer.rs
@@ -30,7 +30,7 @@ use tokio::{
 use tracing::{debug, error, info, trace, warn};
 
 use crate::{
-    BlockAPI,
+    BlockAPI, CommitIndex,
     block::{ExtendedBlock, SignedBlock, VerifiedBlock},
     block_verifier::BlockVerifier,
     commit_vote_monitor::CommitVoteMonitor,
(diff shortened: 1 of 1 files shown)


consensus/core/src/synchronizer.rs
```diff
@@ -908,8 +916,22 @@ impl<C: NetworkClient, V: BlockVerifier, D: CoreThreadDispatcher> Synchronizer<C
 
         // If we are commit lagging, then we don't want to enable the scheduler. As the node is sycnhronizing via the commit syncer, the certified commits
         // will bring all the necessary blocks to run the commits. As the commits are certified, we are guaranteed that all the necessary causal history is present.
+        // However, if local commit index has not progressed for too long, we run periodic sync anyway to help unblock progress, in case large commit sync
+        // requests are failing too frequently.
```
akichidis, line 920: Have we actually observed this in any of our environments? Because I would expect this to not significantly help and _maybe_ even create some race between the synchronizer and commit syncer

akichidis (approved):
I am approving but I would suggest giving this a good test before rollout.

consensus/core/src/synchronizer.rs
mwtian: Yes I have observed this in tests where sometimes commit sync took a long time. Let me optimize and test it out a bit more.

mwtian:
Subsumed by #26021

PR #25191, opened 2026-01-30
Update kv-store readers to read the new schema with fallback

nickvikeras:
## Description

Follow up to https://github.com/MystenLabs/sui/pull/25143 to actually read the new schema. Fallback to the old schema is left in place while we wait for the backfill to complete.

## Test plan

I tried running the e2e test with --write-legacy-data false and ensured everything could still be read when only writing the new schema.

---

## Release notes

- [ ] Protocol:
- [x] gRPC: Read the new BigTable schema and per-pipeline watermarks.
- [ ] JSON-RPC:
- [x] GraphQL: Read the new BigTable schema and per-pipeline watermarks.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-kvstore/src/tables/transactions.rs
+++ b/crates/sui-kvstore/src/tables/transactions.rs
@@ -9,8 +9,9 @@ use sui_types::balance_change::BalanceChange;
 use sui_types::digests::TransactionDigest;
 use sui_types::effects::{TransactionEffects, TransactionEvents};
 use sui_types::messages_checkpoint::CheckpointSequenceNumber;
+use sui_types::signature::GenericSignature;
 use sui_types::storage::ObjectKey;
-use sui_types::transaction::Transaction;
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-graphql/src/task/watermark.rs
amnn, line 1: This will be a larger change (and not blocking) but it would be good to update GraphQL to fetch specific watermarks based on the pipelines it is actively tracking from BigTable.

That change can wait until after we get rid of raw BigTable support, and are just using gRPC-based access though, and it needs to be coupled with a change to how we calculate the AvailableRange so that we pay attention to retention for KV lookups cc @henryachen, tracking here: https://linear.app/mysten-labs/issue/DVX-1987/per-pipeline-archival-watermarks

PR #25194, opened 2026-01-30
indexer-alt: track ConsensusAddressOwner for tx_affected_addresses and blooms pipelines

henryachen:
## Description 

Previously, the indexer only tracked `AddressOwner` as affected addresses, so queries like "show me all transactions affecting address X" would miss transactions that only touched consensus-owned objects affecting X. This change ensures those addresses are indexed the same way.

Changes:
 - extracted recipient match logic into a shared helper and added `ConsensusAddressOwner` to match arm
 - use recipient shared helped in `tx_affected_address` and bloom pipelines

## Test plan 
```
cargo nextest run -p sui-indexer-alt
cargo nextest run -p sui-indexer-alt-e2e-tests -- graphql/
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt/src/handlers/mod.rs
+++ b/crates/sui-indexer-alt/src/handlers/mod.rs
@@ -7,9 +7,12 @@ use std::collections::btree_map::Entry;
 
 use anyhow::Context;
 use sui_indexer_alt_framework::types::base_types::ObjectID;
+use sui_indexer_alt_framework::types::base_types::SuiAddress;
+use sui_indexer_alt_framework::types::effects::TransactionEffects;
 use sui_indexer_alt_framework::types::effects::TransactionEffectsAPI;
 use sui_indexer_alt_framework::types::full_checkpoint_content::Checkpoint;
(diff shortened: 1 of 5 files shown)


crates/sui-indexer-alt/src/handlers/mod.rs
```diff
@@ -89,6 +92,23 @@ pub(crate) fn checkpoint_input_objects(
     Ok(checkpoint_input_objects)
 }
 
+/// The recipient addresses from changed objects in a transaction's effects.
+///
+/// Returns addresses from `AddressOwner` and `ConsensusAddressOwner` owners,
+/// skipping other owner types.
+pub(crate) fn recipient_addresses(
```
wlmyng: Why not just `affected_addresses`? that seems more correct given some object changes come from the gas of the sender, for example

PR #25198, opened 2026-01-31
Fix flaky test_reconfig_with_committee_change_stress test

mystenmark:
## Summary
- Add retry logic to `try_request_add_validator` to handle cases where new validators join consensus late and need time to catch up before their transactions can be sequenced
- Add `execute_transaction_directly` method to `TestCluster` for executing already-signed transactions directly via validators

## Test plan
- Verified with the originally failing seed: `MSIM_TEST_SEED=1769814155031`
- Ran seed-search.py with 20 seeds, all passed

### DIFF
--- a/crates/test-cluster/src/lib.rs
+++ b/crates/test-cluster/src/lib.rs
@@ -623,6 +623,18 @@ impl TestCluster {
         Ok(res.pop().unwrap())
     }
 
+    /// Execute an already-signed transaction via direct validator submission, bypassing the fullnode.
+    pub async fn execute_transaction_directly(
+        &self,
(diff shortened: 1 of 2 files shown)


crates/test-cluster/src/lib.rs
```diff
@@ -623,6 +623,18 @@ impl TestCluster {
         Ok(res.pop().unwrap())
     }
 
+    /// Execute an already-signed transaction via direct validator submission, bypassing the fullnode.
```
mwtian, line 626: I'm ok with this being a temporary workaround. I can look into the full node side.

PR #25213, opened 2026-02-02
Fix FundsWithdrawScheduler DB reference leak

lxfind:
## Description 

The FundsWithdrawScheduler spawned background tasks that held clones of the entire scheduler, including channel senders. This created a self-referential pattern where the tasks held senders while waiting on receivers, preventing the channels from ever closing and keeping DB references alive.

Fix by passing only the inner schedulers (wrapped in Arc) to the spawned tasks instead of the whole scheduler. When the scheduler is dropped, the senders are dropped, channels close, and the tasks exit properly.

## Test plan 

This allows test_authority_persist to run with accumulators enabled.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


alex-mysten (approved):
LGTM!

PR #25218, opened 2026-02-02
[Consensus] Block streaming server

akichidis:
## Description 

This PR is introducing the observer server. According to the design doc the Observers are the nodes that are "following" the consensus DAG but are not proposing. The whole observer server API will run on a separate port from the "authority" one to offer maximum flexibility from the operators.

Also this PR is introducing architectural changes around the networking. Instead of having only one client that implements the `NetworkClient` interface to communicate between nodes, now each component (ex the `synchronizer`, `commit_syncer` ) will have its own client defining their own high level APIs. Those clients will "wrap" internally the clients to talk either to validator or observer nodes basically encapsulating the corresponding logic.

Those are key changes as networking is one of the first things that block streaming affects.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/network/mod.rs
+++ b/consensus/core/src/network/mod.rs
@@ -40,7 +40,7 @@ pub(crate) type NodeId = NetworkPublicKey;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) enum PeerId {
     /// A validator node identified by its authority index.
-    Authority(AuthorityIndex),
+    Validator(AuthorityIndex),
     /// An observer node identified by its network public key.
     #[allow(dead_code)]
     Observer(NodeId),
--- a/consensus/core/src/commit_syncer.rs
+++ b/consensus/core/src/commit_syncer.rs
@@ -553,7 +553,7 @@ where
         let (serialized_commits, serialized_blocks) = inner
             .network_client
             .fetch_commits(
-                crate::network::PeerId::Authority(target_authority),
+                crate::network::PeerId::Validator(target_authority),
                 commit_range.clone(),
                 timeout,
             )
@@ -596,7 +596,7 @@ where
                     let serialized_blocks = inner
                         .network_client
                         .fetch_blocks(
-                            crate::network::PeerId::Authority(target_authority),
+                            crate::network::PeerId::Validator(target_authority),
                             request_block_refs.to_vec(),
                             vec![],
                             false,
--- a/consensus/core/src/network/clients.rs
+++ b/consensus/core/src/network/clients.rs
@@ -57,7 +57,7 @@ where
         // A validator node will always talk via the validator interface to another authority.
         // Otherwise, the only way to communicate with the other peer is via the observer interface.
         if self.context.is_validator()
-            && let PeerId::Authority(authority) = peer
+            && let PeerId::Validator(authority) = peer
         {
             let client = self.validator_client.as_ref().ok_or_else(|| {
                 ConsensusError::NetworkConfig("Validator client not available".to_string())
@@ -132,7 +132,7 @@ where
         // A validator node will always talk via the validator interface to another authority.
         // Otherwise, the only way to communicate with the other peer is via the observer interface.
         if self.context.is_validator()
-            && let PeerId::Authority(authority) = peer
+            && let PeerId::Validator(authority) = peer
         {
             let client = self.validator_client.as_ref().ok_or_else(|| {
                 ConsensusError::NetworkConfig("Validator client not available".to_string())
@@ -157,7 +157,7 @@ where
         // A validator node will always talk via the validator interface to another authority.
         // Otherwise, the only way to communicate with the other peer is via the observer interface.
         if self.context.is_validator()
-            && let PeerId::Authority(authority) = peer
+            && let PeerId::Validator(authority) = peer
         {
             let client = self.validator_client.as_ref().ok_or_else(|| {
                 ConsensusError::NetworkConfig("Validator client not available".to_string())
--- a/consensus/core/src/synchronizer.rs
+++ b/consensus/core/src/synchronizer.rs
@@ -721,7 +721,7 @@ where
         let resp = timeout(
             request_timeout,
             network_client.fetch_blocks(
-                crate::network::PeerId::Authority(peer),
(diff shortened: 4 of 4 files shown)


consensus/core/src/network/mod.rs
```diff
@@ -31,15 +31,27 @@ use crate::{
     error::ConsensusResult,
 };
 
+/// Identifies an observer node by its network public key.
+#[allow(dead_code)]
+pub(crate) type NodeId = NetworkPublicKey;
```
mwtian, line 37:
```suggestion
pub(crate) type ObserverNodeID = NetworkPublicKey;
```

consensus/core/src/network/mod.rs
```diff
@@ -31,15 +31,27 @@ use crate::{
...
+
 // Tonic generated RPC stubs.
 mod tonic_gen {
     include!(concat!(env!("OUT_DIR"), "/consensus.ConsensusService.rs"));
+    include!(concat!(
+        env!("OUT_DIR"),
+        "/consensus.ObserverConsensusService.rs"
```
mwtian:
```suggestion
        "/consensus.ConsensusObserverService.rs"
```

consensus/core/src/network/mod.rs
```diff
@@ -171,6 +183,124 @@ pub(crate) trait NetworkService: Send + Sync + 'static {
     ) -> ConsensusResult<(Vec<Round>, Vec<Round>)>;
 }
 
+/// A stream item for observer block streaming that includes both the block and highest commit index.
+#[allow(dead_code)]
+pub(crate) struct ObserverBlockStreamItem {
+    pub(crate) block: Bytes,
+    pub(crate) highest_commit_index: u64,
```
mwtian, line 199: Is this going to be used?

consensus/core/src/network/mod.rs
```diff
@@ -171,6 +183,124 @@ pub(crate) trait NetworkService: Send + Sync + 'static {
...
+    async fn handle_stream_blocks(
+        &self,
+        peer: NodeId,
+        request_stream: BlockRequestStream,
+    ) -> ConsensusResult<ObserverBlockStream>;
+
+    /// Handles the bidirectional commit streaming request from an observer peer.
+    /// Returns a stream of batched commits with the highest commit index.
+    /// The input stream contains CommitStreamRequest items with Start/Stop commands for flow control.
+    async fn handle_stream_commits(
```
mwtian: As discussed, we can leave this but probably it should be consolidated with commit sync.

consensus/core/src/network/tonic_network.rs
```diff
@@ -713,15 +721,32 @@ impl<S: NetworkService> NetworkManager<S> for TonicManager {
...
-            .metrics
-            .network_metrics
-            .network_type
-            .with_label_values(&["tonic"])
-            .set(1);
-
         info!("Starting tonic service");
 
+        let service_arc = service.clone();
+        let (own_address, http_config) = self.start_validator_server(service).await;
```
mwtian: The more natural way migtht be to expose start_validator_server() and start_observer_server() separately from NetworkManager trait? I believe these services don't always start together, for example on full nodes.

consensus/core/src/network/tonic_network.rs
```diff
@@ -757,12 +782,7 @@ impl<S: NetworkService> NetworkManager<S> for TonicManager {
                     .on_failure(DefaultOnFailure::new().level(tracing::Level::DEBUG)),
             )
             .layer_fn(|service| {
-                mysten_network::grpc_timeout::GrpcTimeout::new(
-                    service,
-                    // This should only bound the unary and initial response time,
```
mwtian, line 811: Keep the comment?

consensus/core/src/authority_service.rs
```diff
@@ -635,10 +622,57 @@ impl<C: CoreThreadDispatcher> NetworkService for AuthorityService<C> {
...
+        // Own blocks do not go through the core dispatcher, so they need to be set separately.
+        highest_received_rounds[self.context.own_index] =
+            highest_accepted_rounds[self.context.own_index];
+
         Ok((highest_received_rounds, highest_accepted_rounds))
     }
 }
 
+#[async_trait]
+impl<C: CoreThreadDispatcher> ObserverNetworkService for AuthorityService<C> {
```
mwtian, line 646: ObserverNetworkService should be implemented for a different struct in a different file?

consensus/core/src/network/mod.rs
akichidis: Yes, that is meant to be used for control flow so the client can easily switch between block and commit sync. Again, that is linked with the discussion of making the API friendlier to generic clients, but as my comment on the original design notion doc: `// Note: Could potentially extract from block commit votes, but explicit is clearer`

consensus/core/src/network/mod.rs
akichidis: Following up our earlier discussion, I am fine for now leaving this endpoint aside and revisit at a later time. We can modify the commit syncer to support targeted synchronization and we can benchmark how much of an issue that might be.

consensus/core/src/network/tonic_network.rs
akichidis: Yes, you are right on this. And the validator server , as you mentioned, will not start for the full nodes. I was just thinking that the logic on which servers to start could be all wrapped up behind the `install_service` as it will make little difference to the caller. I mean, the call it self will need to do similar detection to figure out which servers to start.

consensus/core/src/authority_service.rs
akichidis: Yeah we can totally do that. As I said I've avoided doing too much refactoring for now and just to showcase what could be done re-using as much as possible existing components. But we can totally separate those. One benefit of reusing the `AuthorityService` though is that we won't necessary need to refactor other components using it (ex the Subscriber). But yes, it will be cleaner

akichidis:
Closing this PR as I did break down further the changes so they are more easy to review https://github.com/MystenLabs/sui/pull/25527

PR #25219, opened 2026-02-02
sui-indexer-alt: add timeout error tracing

evan-wall-mysten:
## Description 

While running a benchmark, some requests were timing out (`query_timeout_ms`) with a response message containing `"message": "Query timed out after 40.00s"`, but no more details indicating the root cause. I eventually pinpointed it to the database being underprovisioned and not being able to handle the benchmark load.

This PR adds tracing to GraphQL and logs a warning containing the trace of where the thread(s) were when the request timed out.

## Test plan 

Added unit tests.

Example log output from the unit test:
```
[WARN] Request timed out: timeout elapsed at:
trace 0:
   0: async_graphql::graphql::field
           with path: op, parent_type: Root, return_type: Boolean!
             at /Users/evanwall/.cargo/git/checkouts/async-graphql-7336e61dcafca7ed/7be9351/src/extensions/tracing.rs:135
   1: async_graphql::graphql::execute
             at /Users/evanwall/.cargo/git/checkouts/async-graphql-7336e61dcafca7ed/7be9351/src/extensions/tracing.rs:120
   2: async_graphql::graphql::request
             at /Users/evanwall/.cargo/git/checkouts/async-graphql-7336e61dcafca7ed/7be9351/src/extensions/tracing.rs:55
 request_id=ffffffff-ffff-ffff-ffff-ffffffffffff kind=Query
```

This can be improved on by adding more span instrumentation: https://docs.rs/tracing/latest/tracing/attr.instrument.html.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
+++ b/crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
@@ -17,9 +17,12 @@ use async_graphql::extensions::NextExecute;
 use async_graphql::extensions::NextParseQuery;
 use async_graphql::parser::types::ExecutableDocument;
 use async_graphql::parser::types::OperationType;
-use tokio::time::timeout;
+use timeout_tracing::CaptureSpanTrace;
+use timeout_tracing::timeout;
+use tracing::warn;
 
 use crate::error::request_timeout;
+use crate::extensions::logging::Session;
 
 /// How long to wait for each kind of operation before timing out.
 pub(crate) struct TimeoutConfig {
@@ -87,10 +90,12 @@ impl Extension for TimeoutExt {
             self.config.query
         };
 
-        timeout(limit, next.run(ctx, operation_name))
+        timeout(limit, CaptureSpanTrace, next.run(ctx, operation_name))
             .await
-            .unwrap_or_else(|_| {
+            .unwrap_or_else(|e| {
                 let kind = if is_mutation { "Mutation" } else { "Query" };
+                let Session { uuid, .. } = ctx.data_unchecked();
+                warn!(request_id = %uuid, %kind, "Request timed out: {e}");
                 Response::from_errors(vec![ServerError::from(request_timeout(kind, limit))])
             })
     }
@@ -98,16 +103,26 @@ impl Extension for TimeoutExt {
 
 #[cfg(test)]
 mod tests {
+    use std::net::Ipv4Addr;
+    use std::net::SocketAddr;
+
     use async_graphql::EmptyMutation;
     use async_graphql::EmptySubscription;
     use async_graphql::Object;
     use async_graphql::Schema;
     use async_graphql::Value;
+    use async_graphql::extensions::Tracing;
+    use insta::assert_snapshot;
+    use itertools::Itertools;
+    use regex::Regex;
+    use telemetry_subscribers::TelemetryConfig;
+    use uuid::Uuid;
 
     use crate::error::code;
 
     use super::*;
 
+    #[derive(Clone)]
     struct Root(Duration);
 
     #[Object]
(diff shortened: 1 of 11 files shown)


crates/sui-pg-db/src/tls.rs
```diff
@@ -149,10 +148,10 @@ pub(crate) async fn establish_tls_connection(
                 error!(pid, "Database connection terminated with error: {e}");
             }
             Ok(Ok(())) => {
-                debug!(pid, "Database connection terminated without error");
+                info!(pid, "Database connection terminated without error");
```
wlmyng: why bump this up from debug to info?

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -170,10 +178,17 @@ mod tests {
...
+            extensions.get("code"),
             Some(&Value::String(code::REQUEST_TIMEOUT.into()))
-        )
+        );
+        let Some(Value::String(chain)) = extensions.get("chain") else {
+            panic!("no chain extensions")
+        };
+        assert!(chain.contains("timeout elapsed at"));
+        assert!(chain.contains("0: async_graphql::graphql::field"));
+        assert!(chain.contains("with path: op, parent_type: Root, return_type: Boolean!"));
```
wlmyng: is it possible to return the graphql query fragment specific to the field/resolver that was executing and timed out?

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -66,6 +72,8 @@ pub struct TelemetryConfig {
     pub trace_target: Option<Vec<String>>,
     /// Print unadorned logs from the target on standard error if this is a tty
     pub user_info_target: Vec<String>,
+    // Add error layer to capture trace spans. Defaults to true if not set.
```
henryachen: Do we want to default to True? Might be turning this on for a large set of services

evan-wall-mysten: I changed this to default to `false` in the `telemetry-subscribers` crate, but `true` for GraphQL.

evan-wall-mysten:
> The lack of details returned for a timeout is very much intentional, because a timeout could involve other clients' queries, and we are not generally careful with what we expose through tracing/logs (nor should we be), which means we could risk leaking details about other requests in that error message.
> 
> Let's share the trace as a `trace!` so operators can see it, but clients don't see it.

I removed the trace from the response, but logged it as warn instead of trace because IMO these timeouts indicate a server side issue (timeout set too low, number of levels/fields set too high, database overloaded). If properly configured, I don't think it should be possible for a client to make a query that times out.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -1,25 +1,27 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
+use std::sync::{Arc, OnceLock};
```
amnn: nit: one import per line please. I think this may be a Rust Rover specific issue as `rust-analyzer` will typically follow the convention in the file. Maybe there's a setting you can change so that Rust Rover will do the same.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -54,6 +58,19 @@ impl ExtensionFactory for Timeout {
...
+    /// Capture Session information from the Context so that the `request` handler can use it for
+    /// logging, once it has finished executing.
+    async fn prepare_request(
+        &self,
+        ctx: &ExtensionContext<'_>,
+        request: Request,
+        next: NextPrepareRequest<'_>,
+    ) -> ServerResult<Request> {
+        let session = ctx.data_unchecked();
+        let _ = self.session.set(*session);
```
amnn: Where do you end up using this? It looks like you fetch the session from the context when you need it, below.

crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -426,6 +430,15 @@ async fn graphiql(path: MatchedPath) -> Html<String> {
     Html(GraphiQLSource::build().endpoint(path.as_str()).finish())
 }
 
+pub fn telemetry_config_for_graphql() -> TelemetryConfig {
```
amnn: This is not really part of the library, it's part of `main`, just set it up there.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -155,13 +176,27 @@ mod tests {
     /// The request takes longer than the timeout to handle, so it should fail.
     #[tokio::test]
     async fn test_query_timeout_fail() {
+        // Enable tracing, configured by environment variables.
+        let (_guard, handle) = telemetry_config_for_graphql()
```
amnn: Ditto -- just inline the telemetry config here.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -173,7 +208,18 @@ mod tests {
         assert_eq!(
             error.extensions.as_ref().unwrap().get("code"),
             Some(&Value::String(code::REQUEST_TIMEOUT.into()))
-        )
+        );
+
+        let events = handle.get_test_layer_events();
+        // error message
+        assert!(events[0].contains("[WARN] Request timed out: timeout elapsed at:"));
```
amnn: Perhaps use an insta snapshot test here? This is going to be a pain to maintain otherwise.

crates/sui-pg-db/src/tls.rs
amnn, line 1: What's the motivation for this change v.s. turning on `debug` logging for this module when we want to see them? I.e. is it the case that these logs are generally useful to understand how the system is operating (and not too noisy), or are they something that help investigate a particular kind of issue in which case they can be enabled when that issue needs to be investigated.

crates/sui-indexer-alt-graphql/src/main.rs
```diff
@@ -32,9 +33,7 @@ async fn main() -> anyhow::Result<()> {
     let args = Args::parse();
 
     // Enable tracing, configured by environment variables.
-    let _guard = telemetry_subscribers::TelemetryConfig::new()
-        .with_env()
-        .init();
+    let _guard = telemetry_config_for_graphql().with_env().init();
```
amnn: IMO, abstracting this with a function is making it harder to follow, let's just write the config inline -- the function seems to imply that GraphQL must be started with tracing set a certain way which is not the case.

```suggestion
    let _guard = telemetry_subscribers::TelemetryConfig::new()
        .with_enable_error_layer(true)
        .with_env()
        .init();
```

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -586,6 +636,55 @@ pub fn init_for_testing() {
     Lazy::force(&LOGGER);
 }
 
+// Captures log events for tests.
```
amnn: Does this deserve its own module?

amnn (commented):
Thanks @evan-wall-mysten, this is much better -- and yes, agreed that it can/should be a warning as it's an issue with the server if the user can generate a timeout.

Can you also share an example of what this warning message will look like?

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
evan-wall-mysten: I was originally following what `LoggingExt` was doing, but ended up not using it. Removed.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
evan-wall-mysten: Yeah it keeps doing this and I missed this one. RustRover uses `rustfmt` so it is waiting on https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=import#imports_granularity becoming stable.

amnn: Yeah, `rust-analyzer` is the same, but also has some basic heuristics for its auto-fixes. I was hoping that Rust Rover would have the same.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
evan-wall-mysten: Oh nice I didn't realize it was as easy as using `assert_snapshot!`. Much more readable, but I did have to filter out the stack trace lines because they are not stable.

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: Yeah this `lib.rs` file seems to be getting large.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
evan-wall-mysten: I don't see anything built in, but I found this plugin that I'll try out: https://plugins.jetbrains.com/plugin/25600-rust-analyzer

crates/sui-pg-db/src/tls.rs
evan-wall-mysten: I was testing this change by running the benchmark env with an underprovisioned database and noticed two kinds of errors:
1. The database closes connections when it runs out of memory (`"Database connection terminated with error: {e}"`).
2. Query timeouts (what this PR helps with debugging).

I think upgrading these to `INFO` could help with issues similar to 1., but I didn't see these logs during my test. I don't know what would cause them, but I think it is ok to to make them `INFO` without being too noisy.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -155,73 +170,98 @@ mod tests {
...
+        let timeout = Duration::from_millis(200);
+        let event = test_timeout_fail(
+            timeout,
+            Duration::ZERO,
+            timeout * 2,
+            "query { op }",
+            "Query",
         )
+        .await;
+        assert_snapshot!(event);
```
amnn: I was thinking that you would use an inline snapshot here (add `@""` to the macro invocation and insta will populate the snapshot inline when you re-run the tests -- don't forget to delete the snapshots folder that was added before if you do this, otherwise it will stick around forever), because it's a bit easier to correlate the expected output with the operation that caused it.


```suggestion
        assert_snapshot!(event, @"");
```

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -155,73 +170,98 @@ mod tests {
...
             Some(&Value::String(code::REQUEST_TIMEOUT.into()))
-        )
+        );
+
+        let events = handle.get_test_layer_events();
+        assert_eq!(events.len(), 1);
+        // Remove line number strings to avoid test churn
+        // example: "             at /Users/evanwall/.cargo/git/checkouts/async-graphql-7336e61dcafca7ed/7be9351/src/extensions/tracing.rs:135"
+        let re = Regex::new(r"\s+at\s.+").unwrap();
+        events[0].split("\n").filter(|s| !re.is_match(s)).join("\n")
```
amnn, line 289: Insta does have a native feature for this, called [filters](https://insta.rs/docs/filters/), but this is fine too (it's harder to abstract away the filter settings, although it is possible).

evan-wall-mysten: Yeah it looks like that will require more boilerplate to get working because the `assert_snapshot!` needs to both be inside the `with_settings!` and inside the test function.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
evan-wall-mysten: Oh, nice I didn't know this was possible.

PR #25225, opened 2026-02-02
bridge: complete migration to grpc

bmwill:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-bridge/src/e2e_tests/basic.rs
+++ b/crates/sui-bridge/src/e2e_tests/basic.rs
@@ -26,6 +26,7 @@ use std::sync::Arc;
 use sui_types::bridge::{
     BridgeChainId, BridgeTokenMetadata, BridgeTrait, TOKEN_ID_ETH, get_bridge,
 };
+use sui_types::coin::Coin;
 use sui_types::crypto::get_key_pair;
 use sui_types::effects::TransactionEffectsAPI;
 use tracing::info;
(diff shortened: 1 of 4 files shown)


crates/sui-bridge/src/node.rs
```diff
@@ -344,6 +337,17 @@ async fn start_client_components(
...
     .await;
 
+    let (sui_monitor_tx, sui_monitor_rx) = mysten_metrics::metered_channel::channel(
+        10000,
+        &mysten_metrics::get_metrics()
+            .unwrap()
+            .channel_inflight
+            .with_label_values(&["sui_monitor_queue"]),
+    );
+    tokio::spawn(monitor::subscribe_bridge_events(
```
dlukegordon, line 346: Should we be using `spawn_logged_monitored_task!` here?

crates/sui-bridge/src/monitor.rs
```diff
@@ -468,6 +471,85 @@ async fn get_latest_bridge_pause_status_with_emergency_event<C: SuiClientInner>(
...
+            let checkpoint = match item {
+                Ok(checkpoint) => checkpoint,
+                Err(e) => {
+                    tracing::warn!("error in checkpoint stream: {e}");
+                    break;
+                }
+            };
+
+            let ckpt = checkpoint.cursor();
+            tracing::debug!("recieved checkpoint {ckpt}");
```
dlukegordon, line 516: Typo

crates/sui-bridge/src/node.rs
```diff
@@ -580,92 +572,6 @@ mod tests {
         );
     }
 
-    #[tokio::test]
-    async fn test_get_sui_modules_to_watch() {
```
dlukegordon, line 588: It looks like we still are using the `get_sui_modules_to_watch` function, should we still be testing it?

bmwill: We don't really use it, its just a convoluted way to read some values from the db now (which we can clean up later)

PR #25227, opened 2026-02-02
Add ObjectBalanceOverdraw benchmark operation and instrumentation

mystenmark:
## Summary
- Add new `ObjectBalanceOverdraw` operation to composite benchmark workloads that tests balance pool overdraw scenarios
- Track pool balance in `AccountState` for accurate overdraw calculations
- Add `assert_reachable!` instrumentation for pending object funds scenarios to monitor code path coverage

## Test plan
- [ ] Run composite benchmark workloads with the new operation enabled
- [ ] Verify `ObjectBalanceOverdraw` withdraws 50-100% of pool balance as expected

### DIFF
--- a/crates/sui-benchmark/src/workloads/composite/operations.rs
+++ b/crates/sui-benchmark/src/workloads/composite/operations.rs
@@ -997,7 +997,7 @@ impl Operation for AccumulatorBalanceRead {
 pub struct ObjectBalanceOverdraw;
 
 impl ObjectBalanceOverdraw {
-    pub const FLAG: u32 = 1 << 12;
+    pub const FLAG: u32 = 1 << 16;
     pub const DESCRIPTOR: OperationDescriptor = OperationDescriptor {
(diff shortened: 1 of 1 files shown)


crates/sui-core/src/accumulators/object_funds_checker/mod.rs
```diff
@@ -134,6 +145,7 @@ impl ObjectFundsChecker {
             // At that time we will have to re-enqueue the transaction for execution again.
             // Re-enqueue is handled here so the caller does not need to worry about it.
             ObjectFundsWithdrawStatus::Pending(receiver) => {
+                assert_reachable!("object funds pending");
```
lxfind: Might be better to put this in the two match branches below

crates/sui-core/src/accumulators/object_funds_checker/mod.rs
lxfind, line 137: Add an assert_reachable here as well?

PR #25228, opened 2026-02-02
[bridge] Handle storage layout mismatch during EVM contract upgrades

Bridgerz:
The testnet SuiBridge proxy was deployed without the __gap storage array
in CommitteeUpgradeable, placing vault at slot 4 and limiter at slot 5.
Mainnet was deployed with __gap, placing them at slots 54 and 55.

After the V2 upgrade on testnet, the new implementation expected the
mainnet layout, causing vault/limiter reads to return 0x0 and breaking
node startup.

Changes:
- utils.rs: Gracefully handle zero vault address by returning zero token
  addresses with warning logs instead of failing
- node.rs: Skip vault balance monitoring when addresses are zero, while
  keeping bridge pause status monitoring active
- CommitteeUpgradeable.sol: Document the storage layout divergence
- SUI_NATIVE_BRIDGE_PRIMER.md: Add section explaining mainnet vs testnet
  storage differences and upgrade procedures

This allows bridge nodes to start and sign governance actions (including
the fix upgrade) even when storage layout is mismatched.

### DIFF
--- a/crates/sui-bridge/src/node.rs
+++ b/crates/sui-bridge/src/node.rs
@@ -172,8 +172,22 @@ async fn start_watchdog(
         );
     }
 
-    let mut observables: Vec<Box<dyn Observable + Send + Sync>> = if vault_monitoring_enabled {
-        // Create vault balance monitors when addresses are valid
+    let eth_bridge_status = EthBridgeStatus::new(
+        eth_provider.clone(),
+        eth_bridge_proxy_address,
+        watchdog_metrics.eth_bridge_paused.clone(),
+    );
+
+    let sui_bridge_status = SuiBridgeStatus::new(
+        sui_client.clone(),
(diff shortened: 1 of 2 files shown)


crates/sui-bridge/src/node.rs
```diff
@@ -162,75 +162,102 @@ async fn start_watchdog(
...
+        } else {
+            None
+        };
 
-    let eth_bridge_status = EthBridgeStatus::new(
-        eth_provider,
-        eth_bridge_proxy_address,
-        watchdog_metrics.eth_bridge_paused.clone(),
-    );
+        let eth_bridge_status = EthBridgeStatus::new(
```
dlukegordon: Getting `eth_bridge_status` and `sui_bridge_status` is duplicated across both branches, can factor out.

dlukegordon (approved):

PR #25234, opened 2026-02-03
[move-ide] Support for incremental compilation of dependent modules

awelc:
## Description 

This the next step towards supporting module extensions. It does not yet implement a proper full fix, instead it focuses on support for incremental compilation that would work with module extensions.

The idea is to avoid caching dependent modules that have user-level extensions but rather compile them fully so that we can access extension-defined module-level definitions, in particular functions with their bodies. 

## Test plan 

All tests must pass. The recently added module extension test shows increased extension support coming with this PR by showing that a function definition in and extension defined for a dependent module is reported correctly by `move-analyzer`

### DIFF
--- a/external-crates/move/crates/move-analyzer/src/symbols/mod_extensions.rs
+++ b/external-crates/move/crates/move-analyzer/src/symbols/mod_extensions.rs
@@ -48,11 +48,11 @@ use std::{
 
 use move_command_line_common::files::FileHash;
 use move_compiler::{
+    Flags, PreCompiledProgramInfo,
     editions::Edition,
     expansion::ast as E,
     parser::{ast as P, syntax::parse_file_string},
     shared::{CompilationEnv, NamedAddressMap, PackageConfig},
-    Flags, PreCompiledProgramInfo,
 };
 use move_ir_types::location::sp;
 use move_symbol_pool::Symbol;


external-crates/move/crates/move-analyzer/src/symbols/compilation.rs
```diff
@@ -175,18 +174,36 @@ pub struct SymbolsComputationData {
...
+    /// Hashes of individual dependency files
     dep_hashes: Vec<FileHash>,
+    /// Paths of individual dependency packages
     dep_pkg_paths: BTreeMap<Symbol, PathBuf>,
+    /// Root package source files (for extension detection)
+    root_source_files: Vec<Symbol>,
+    /// Root package named addresses (for extension detection)
+    root_named_addresses: Arc<NamedAddressMap>,
+    /// Root package edition (for extension detection)
+    root_edition: Edition,
```
awelc, line 190: We need edition for extension handling now as well so it's also computed a bit differently now (and not cached as it does not have to be - it's computed from root package each time)

external-crates/move/crates/move-analyzer/tests/module_ext.snap
```diff
@@ -12,68 +12,18 @@ public fun ModuleExt::M2::bar(): u64
 
 -- test 1 -------------------
 use line: 6, use_ndx: 1
-ERROR: No use_line 6 in mod_symbols UseDefMap(
```
awelc, line 15: These changes look a bit weird as in the previous PR I updated the test but not fully (a function from the wrong module was called). If you look at the actual snapshot, though, it shows what it should. In particular,  functions defined in extensions are recognized for both extension defined for user module and for a dependent module.

external-crates/move/crates/move-analyzer/src/analysis/mod.rs
```diff
@@ -104,9 +104,17 @@ fn process_typed_modules<'a>(
     mod_to_alias_lengths: &'a BTreeMap<String, BTreeMap<Position, usize>>,
     typing_symbolicator: &mut typing_analysis::TypingAnalysisContext<'a>,
 ) {
+    // TODO(extensions): Extension modules are skipped during parsing analysis because
+    // their location-based address map lookup fails in `pkg_symbols`. When full extension
+    // support is added, alias lengths should be computed for extensions and this
+    // fallback removed.
+    static EMPTY_ALIAS_LENGTHS: std::sync::LazyLock<BTreeMap<Position, usize>> =
+        std::sync::LazyLock::new(BTreeMap::new);
```
cgswords, line 112: What is this doing? What is the point of the length storage? This is just confusing to me, and the comment is not clear.

external-crates/move/crates/move-analyzer/src/symbols/compilation.rs
```diff
@@ -414,12 +490,30 @@ pub fn get_compiled_pkg<F: MoveFlavor>(
...
+
+                // Remove dependencies that are already included in the cached package info,
+                // EXCEPT those in packages_to_keep (which need full compilation).
                 dependencies.retain(|d| {
-                    !cached_pkg_info
-                        .dep_names
-                        .contains(&Symbol::from(d.id().to_string()))
+                    let pkg_symbol = Symbol::from(d.id().to_string());
+                    !cached_pkg_info.dep_names.contains(&pkg_symbol)
+                        || packages_to_keep.contains(&pkg_symbol)
```
cgswords, line 534: This does not seem to reflect the comment above due to short circuiting.

external-crates/move/crates/move-analyzer/src/symbols/mod_extensions.rs
```diff
@@ -0,0 +1,130 @@
...
+        std::collections::BTreeMap::new(),
+        Some(PackageConfig {
+            edition,
+            ..Default::default()
+        }),
+        None,
+    );
+
+    // Parse file using the compiler's parser
+    let definitions = parse_file_string(&env, file_hash, &contents, None).ok()?;
```
cgswords, line 173: This is really strange to me. Why re-parse the file instead of pulling extensions out of the AST?

PR #23710, opened 2025-09-23
[framework] Display Registry

manolisliolios:
## Description 

Introduces new Display using a registry (similar to coin):
1. Display no longer is restricted to `T: key`. We can have display for non top-level objects.
2. Enable migration paths from `Display<T>` and `Publisher`, as well as legacy deletion after migration.
3. Enable "system" migration path using a singleton Cap, that will be destroyed after moving V1s to the registry.

**Why we are doing this?**
1. Display indexing currently relies on events, making it extremely hard to maintain proper indexing. This takes away this problem, by having a fixed point of lookup and rely on the live object set.
2. We wanna enable display for non-`key` objects too
3. Future-proof our way into an even better creation (e.g when args on init are landed)

**TODOS:**

- [x] Create the core framework code
- [x] Add move tests for all cases
- [x] Style / naming nits
- [x] Create new system object `0xd`(?)
- [ ] Leverage GQL's indexer's db to get a snapshot of V1 display objects and create the PTBs to migrate them
- [x] Come up with a secure key policy to store the migration cap (if we agree on doing it that way)
- [ ] Deprecate V1 package (`display.move`) [in a follow-up PR]

## Test plan 

How did you test the new or updated feature?

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [x] Protocol: Enables Display V2 (`0xd` system object is created)
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:

PR #25242, opened 2026-02-03
registry/display: GraphQL integration

amnn:
## Description
GraphQL will look for and preference `sui::display_registry::Display<T>` over `sui::display::Display<T>` (+ its events) when computing the Display output for a given Move Value.

## Test plan

New E2E tests:

```
$ cargo nextest run            \
  --test transactional_tests   \
  -- display_v2
```

## Stack

- #23710 
- #25240 
- #25241

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Adds support for Display Registry to GraphQL: `MoveValue.display` will look for a `Display<T>` stored in the Display Registry and will use that as the source of truth for its type's format. This takes precedence over any Display v1 formats that exist for this type.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/display.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/display.rs
@@ -1,9 +1,22 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::sync::Arc;
+
+use anyhow::Context as _;
+use async_graphql::Context;
 use async_graphql::SimpleObject;
+use async_graphql::dataloader::DataLoader;
+use move_core_types::language_storage::StructTag;
+use sui_indexer_alt_reader::displays::DisplayKey;
+use sui_indexer_alt_reader::pg_reader::PgReader;
+use sui_types::display::DisplayVersionUpdatedEvent;
+use sui_types::display_registry;
 
 use crate::api::scalars::json::Json;
+use crate::api::types::object::Object;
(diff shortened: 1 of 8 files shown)


crates/sui-indexer-alt-graphql/src/api/types/display.rs
```diff
@@ -14,3 +27,51 @@ pub(crate) struct Display {
...
+        return Ok(None);
+    };
+
+    let Some(native) = object.contents(ctx).await? else {
+        return Ok(None);
+    };
+
+    let Some(move_object) = native.data.try_as_move() else {
+        return Ok(None);
+    };
```
henryachen: nit: We usually throw an error `.context("Failed to retrieve contents for...")` if we fail to fetch contents after we have fetched something by address. Is `Ok(None)` here intentional? IIUC, `Ok(None)` is used here to fall through to V1, do we want all three of these cases to fall through?

amnn: Hmm, good question, I didn't really think about it.

In the past we took a very permissive stance (returning `null` instead of an error) so that if there was some issue fetching the Display format, it wouldn't mess with the rest of the GraphQL query.

Now we have partial errors, so we don't have that issue anymore -- so I think it's safe to return an error here, and that seems like the right call -- thanks for flagging!

PR #25250, opened 2026-02-03
[bella-ciao] Safe index trait + adopt usage

tzakian:
## Description 

This adds a `SafeIndex` trait over `AsRef<[T]>` to allow for indexing into collections in the VM and returning a VM error (invariant violation) if the lookup ever fails.

This also adds a `#![deny(clippy::indexing_slicing)]` to the new vm runtime. We only permit directs indices in two places in the VM now:
1. In tests/test-only code (seems fair)
2. In tracing (seems fair, needlessly complex to not allow this + not enabled in prod pathways).

## Test plan 

CI/existing tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/cache/arena.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/cache/arena.rs
@@ -140,16 +140,20 @@ impl<T> std::ops::Deref for ArenaVec<T> {
     }
 }
 
-impl<T> std::ops::Deref for ArenaBox<T> {
-    type Target = T;
-
-    fn deref(&self) -> &Self::Target {
+impl<T> AsRef<[T]> for ArenaVec<T> {
+    fn as_ref(&self) -> &[T] {
         &self.0
     }
 }
 
-impl<T> AsRef<[T]> for ArenaVec<T> {
-    fn as_ref(&self) -> &[T] {
+// -----------------------------------------------
+// ArenaBox Trait Implementations
+// -----------------------------------------------
+
+impl<T> std::ops::Deref for ArenaBox<T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
         &self.0
     }
 }
--- a/external-crates/move/crates/move-vm-runtime/src/jit/optimization/translate.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/jit/optimization/translate.rs
@@ -81,6 +81,7 @@ fn generate_basic_blocks(
             let start = cfg.block_start(label) as usize;
             let end = cfg.block_end(label) as usize;
             let label = label as ast::Label;
+            // TODO: Try and make this code a bit nicer
             let code = input
                 .safe_get(start..(end + 1))?
                 .iter()
--- a/external-crates/move/crates/move-vm-runtime/Cargo.toml
+++ b/external-crates/move/crates/move-vm-runtime/Cargo.toml
@@ -53,6 +53,9 @@ move-stdlib.workspace = true
 move-proc-macros.workspace = true
 insta.workspace = true
 
+[lints.clippy]
+indexing_slicing = "warn"
+
 [features]
 default = []
 fuzzing = []


external-crates/move/crates/move-vm-runtime/src/shared/mod.rs
```diff
@@ -55,3 +56,37 @@ pub fn unique_map<Key: Hash + Eq, Value>(
     }
     Ok(map)
 }
+
+/// A trait for safe indexing into collections that returns a PartialVMResult as long as the
+/// collection implements [`AsRef<[T]>`].
+/// This is useful for avoiding panics on out-of-bounds access, and instead returning a proper
+/// error.
+pub trait SafeIndex<T> {
```
tzakian, line 64: This is the trait/only implementation of it. Everything else just implements `AsRef<[T]>` and this will work over it.

external-crates/move/crates/move-vm-runtime/src/cache/arena.rs
```diff
@@ -140,14 +140,16 @@ impl<T> std::ops::Deref for ArenaVec<T> {
     }
 }
 
-// -----------------------------------------------
```
cgswords, line 143: Nit: This should read `ArenaBox`, and the `ArenaVec` impl should appear above it.

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
```diff
@@ -24,6 +24,7 @@ use crate::{
     natives::{extensions::NativeContextExtensions, functions::NativeContext},
     partial_vm_error,
     shared::{
+        SafeIndex as _,
```
cgswords, line 27: I hate this. 😄

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -448,19 +448,11 @@ impl std::iter::IntoIterator for FixedSizeVec {
...
-
-// Implement the `IndexMut` trait to allow mutable indexing.
-impl IndexMut<usize> for FixedSizeVec {
-    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
-        &mut self.0[index]
+// Implement `AsRef<[Membox<Value>]>` for `FixedSizeVec` to get access to the `SafeIndex` trait to
+// allow safe immutable indexing.
+impl AsRef<[MemBox<Value>]> for FixedSizeVec {
+    fn as_ref(&self) -> &[MemBox<Value>] {
+        self.as_slice()
```
cgswords, line 455: Why not just implement `SafeIndex` for `FixedSizeVec`? Is the idea to keep it all in one place?

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -968,7 +962,7 @@ impl StructRef {
         let container = &mut *self.0.borrow_mut();
         match container {
             Value::Struct(fixed_vec) => {
-                fixed_vec.0[index] = mem_box;
+                fixed_vec.0.0[index] = mem_box;
```
cgswords, line 965: This does not look particularly safe to me?

external-crates/move/crates/move-vm-runtime/src/jit/optimization/translate.rs
```diff
@@ -17,66 +22,73 @@ pub(crate) fn package(pkg: Input::Package) -> ast::Package {
...
             let end = cfg.block_end(label) as usize;
             let label = label as ast::Label;
-            let code = input[start..(end + 1)].iter().map(bytecode).collect();
-            (label, code)
+            let code = input
+                .safe_get(start..(end + 1))?
+                .iter()
+                .map(bytecode)
+                .collect();
+            Ok((label, code))
```
cgswords, line 90: This code was suspect before, and now it looks even worse. Can you add a `TODO` to investigate this?

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
tzakian: Yea, main thinking was to keep it all in one place so we can ensure a consistent error message/error type across all instances (+ this felt cleaner to me).

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
tzakian: 😉

PR #24575, opened 2025-12-10
Migrate native bridge from ethers to alloy

dlukegordon:
## Description 

Migrated all use of the `ethers` rust library to `alloy`.

[Linear issue](https://linear.app/mysten-labs/issue/BR-191/migrate-from-ethers-to-alloy).

Helpful documentation: https://alloy.rs/migrating-from-ethers/reference/

## Test plan 

Ran tests with `cargo test` locally as well as CI tests.

We may need to do more extensive testing, will consult with Bridger and Brandon.

PR #25257, opened 2026-02-03
[alloy][sui-bridge] Integrate Alloy multiprovider in sui bridge

0xsiddharthks:
## Description 

This PR builds on top of #24575 and replaces the native Alloy provider with the [multiprovider implementation](https://github.com/0xsiddharthks/alloy-multiprovider-strategy).

I have added the following fields to the sui bridge config:
- eth_rpc_url (deprecated. kept for backward compatibility)
- eth_rpc_urls
- eth_rpc_quorum
- eth_health_check_interval_secs

This change is backwords compatible with the existing config, but we will deprecate the eth_rpc_url field in the future once all node operators have migrated to using the new config fields.

## Test plan 

I have thoroughly tested the `alloy-multiprovider-strategy` crate via various test scripts.

Also all the CI tests are passing on the sui repo.

---

## Release notes

- [ ] Protocol: 
- [x] Nodes (Validators and Full nodes): Added multi-provider Ethereum RPC support to the bridge node. 

Operators can now configure multiple Ethereum RPC endpoints with quorum-based consensus for improved redundancy and fault tolerance. 

New optional YAML config fields (in sui bridge config): 
- `eth-rpc-urls` (list of RPC URLs), 
- `eth-rpc-quorum` (quorum size, defaults to 1)
- `eth-health-check-interval-secs` (health check interval, defaults to 300s).

The existing `eth-rpc-url` field continues to work for backward compatibility. When a single URL is configured, the multi-provider layer operates as a zero-overhead passthrough with no quorum, health-check, or locking machinery.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-bridge/Cargo.toml
+++ b/crates/sui-bridge/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2024"
 
 [dependencies]
 alloy = { version = "1.1.3", features = ["rand", "sol-types", "json-rpc", "provider-ws"] }
-alloy-multiprovider-strategy = { git = "https://github.com/0xsiddharthks/alloy-multiprovider-strategy", rev = "d94800292ffd393b7776817d01bca0c4f2533087" }
+alloy-multiprovider-strategy = { git = "https://github.com/0xsiddharthks/alloy-multiprovider-strategy", rev = "0f8d034e679631a715f9eed104089430ae57283a" }
 tokio = { workspace = true, features = ["full"] }
 sui-types.workspace = true
 sui-authority-aggregation.workspace = true
--- a/crates/sui-bridge/src/config.rs
+++ b/crates/sui-bridge/src/config.rs
@@ -5,7 +5,7 @@ use crate::abi::EthBridgeConfig;
 use crate::crypto::BridgeAuthorityKeyPair;
 use crate::error::BridgeError;
 use crate::eth_client::EthClient;
-use crate::metered_eth_provider::{new_metered_eth_provider, new_metered_multi_eth_provider};
+use crate::metered_eth_provider::new_metered_eth_multi_provider;
 use crate::metrics::BridgeMetrics;
 use crate::sui_client::SuiBridgeClient;
 use crate::types::{BridgeAction, is_route_valid};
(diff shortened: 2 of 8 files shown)


crates/sui-bridge/Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2024"
 
 [dependencies]
 alloy = { version = "1.1.3", features = ["rand", "sol-types", "json-rpc", "provider-ws"] }
+alloy-multiprovider-strategy = { git = "https://github.com/0xsiddharthks/alloy-multiprovider-strategy", rev = "0f8d034e679631a715f9eed104089430ae57283a" }
```
dlukegordon, line 11: Should we move this somewhere within the MystenLabs org?

crates/sui-bridge/src/config.rs
```diff
@@ -41,7 +41,19 @@ use tracing::info;
...
     /// Rpc url for Eth fullnode, used for query stuff.
-    pub eth_rpc_url: String,
+    /// @deprecated (use eth_rpc_urls instead)
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub eth_rpc_url: Option<String>,
+    /// Multiple RPC URLs for Eth fullnodes.
+    /// Quorum-based consensus is used across providers for redundancy.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub eth_rpc_urls: Option<Vec<String>>,
+    /// Quorum size for multi-provider consensus. Must be <= number of URLs.
```
dlukegordon, line 51: I don't think we're validating that the quorum is <= the number of URLs?

0xsiddharthks: we have this validation in the multiprovider itself, when it initializes.

crates/sui-bridge/Cargo.toml
0xsiddharthks: Yeah, I had talked to Brandon about it.
As of now, it's a fairly common pattern in the monorepo to have links to repositories / forks owned by personal accounts. 

But I'm totally open to transferring the repo ownership to the MystenLabs org if thats recommended.

dlukegordon: Cool, sounds good to me I just wanted to check

crates/sui-bridge/src/config.rs
bmwill: You can probably have this just be Vec<String> and change the skip serialize to Vec::is_empty

crates/sui-bridge/Cargo.toml
bmwill: Yeah this is ok for the time being.

crates/sui-bridge/src/config.rs
```diff
@@ -41,7 +41,19 @@ use tracing::info;
 #[serde(rename_all = "kebab-case")]
 pub struct EthConfig {
     /// Rpc url for Eth fullnode, used for query stuff.
-    pub eth_rpc_url: String,
+    /// @deprecated (use eth_rpc_urls instead)
```
dlukegordon, line 44: Does `@deprecated` in a comment actually trigger any deprecation warnings? Maybe we should use `#[deprecated]`?

bmwill: So this is more a note, we are not deprecating the actual rust field itself. And i wouldn't be surprised if ever after we role this out most folks won't update configs

PR #25261, opened 2026-02-03
[graphql-alt] Handle transaction execution/simulation errors as BAD_USER_INPUT

tpham-mysten:
## Description 

- gRPC `InvalidArgument` and `NotFound` errors from `executeTransaction` and `simulateTransaction` now return GraphQL `BAD_USER_INPUT` errors instead of being returned in an `error`/`errors` field
- Added `InvalidArgument` variant to `TransactionInputError` with message prefix "Invalid argument:"
- Removed `error` field from `SimulationResult` and `errors` field from `ExecutionResult`

## Example error:

```
{
  "data": null,
  "errors": [
    {
      "message": "Invalid argument: Invalid user signature: Signature is not valid: Fail to verify user sig Invalid signature was given to the function",
      "locations": [
        {
          "line": 3,
          "column": 17
        }
      ],
      "path": [
        "executeTransaction"
      ],
      "extensions": {
        "code": "BAD_USER_INPUT"
      }
    }
  ]
}
```

## Test plan 

- test_execute_transaction_grpc_errors verifies BAD_USER_INPUT code and ServerValidation message
- Existing e2e tests updated to remove error/errors field references

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Error field is no longer available in simulateResult and ExecutionResult. They will be propagated in GraphQL errors
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql_execute_transaction_tests.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql_execute_transaction_tests.rs
@@ -387,21 +387,8 @@ async fn test_execute_transaction_grpc_errors() {
         .await
         .unwrap();
 
-    // data should be null when there's a BAD_USER_INPUT error
-    assert_eq!(result.get("data"), Some(&serde_json::Value::Null));
-
-    // GraphQL errors should be returned with BAD_USER_INPUT code and InvalidArgument message
-    let code = result
-        .pointer("/errors/0/extensions/code")
-        .and_then(|v| v.as_str());
-    assert_eq!(code, Some("BAD_USER_INPUT"));
-
-    let message = result.pointer("/errors/0/message").and_then(|v| v.as_str());
-    assert!(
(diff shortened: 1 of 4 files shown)


crates/sui-indexer-alt-graphql/src/api/types/execution_result.rs
amnn, line 1: Do we even need an `ExecutionResults` type now?

amnn: Maybe it's worth it in case we need to add other fields (e.g. if we want to later include `outputs` like for simulation.

tpham-mysten:
Thanks for the review!

Sure. I have added the example error and also added a snapshot test to show how the error would look like in `crates/sui-indexer-alt-e2e-tests/tests/snapshots/graphql_execute_transaction_tests__execute_transaction_grpc_error.snap`.

PR #25282, opened 2026-02-04
[transactional-tests][regex reference safety] Added viewer for pre states

tnowacki:
## Description 

- Added a stringy version of the abstract state used just for a tests
- Added a command to dump the abstract states

TODO:
- Better labels for the code offsets
- Track non reference values 

## Test plan 

- Added tests 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-ir-compiler-transactional-tests/tests/bytecode-generation/statements/enums.snap
+++ b/external-crates/move/crates/move-ir-compiler-transactional-tests/tests/bytecode-generation/statements/enums.snap
@@ -20,93 +20,93 @@ B0:
 }
 
 f1(): T {
-L0:	loc0: T
-L1:	loc1: u64
+L0:	x: T
+L1:	f: u64
 B0:
 	0: LdU64(0)
 	1: PackVariant(VariantHandleIndex(1))
-	2: StLoc[0](loc0: T)
-	3: MoveLoc[0](loc0: T)
+	2: StLoc[0](x: T)
+	3: MoveLoc[0](x: T)
 	4: UnpackVariant(VariantHandleIndex(1))
-	5: StLoc[1](loc1: u64)
+	5: StLoc[1](f: u64)
 	6: PackVariant(VariantHandleIndex(0))
 	7: Ret
 }
 
 f2(): T {
-L0:	loc0: T
-L1:	loc1: u64
+L0:	x: T
+L1:	f: u64
 B0:
 	0: LdU64(0)
 	1: PackVariant(VariantHandleIndex(1))
-	2: StLoc[0](loc0: T)
-	3: ImmBorrowLoc[0](loc0: T)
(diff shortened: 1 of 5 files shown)


external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
```diff
@@ -745,3 +750,144 @@ fn borrowed_by_size(paths_map: &BTreeMap<Ref, Paths>) -> usize {
         .flat_map(|(_, paths)| paths)
         .fold(0, |acc, path| acc.saturating_add(path.abstract_size()))
 }
+
+//**************************************************************************************************
+// Serializable Abs Int
+//**************************************************************************************************
+
+pub trait StateSerializer {
```
tnowacki: I can move this to another file if we want, but unfortunately I can't really feature gate it due to strangeness with the transactional tests

external-crates/move/crates/move-transactional-test-runner/src/framework.rs
```diff
@@ -494,11 +500,18 @@ impl CompiledState {
...
+                self.source_maps.insert(id, source_map);
+            }
         }
     }
 
-    pub fn add_and_generate_interface_file(&mut self, module: CompiledModule) {
+    pub fn add_and_generate_interface_file(
+        &mut self,
+        module: CompiledModule,
+        source_map: Option<SourceMap>,
```
tnowacki, line 513: This really shouldn't be an Option, but I don't feel like fixing the sui side just for this

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
tzakian: This is a testing-only interface, so moving it into a separate module that conveys that would be nice (and bonus points if it can be e.g., only under `debug_assertions` or `test` or something like that, but not sure how easy that will be).

tzakian (approved):
Pretty cool! The main thing is seeing if the `StateSerializer` trait can be moved into a more aptly named module (and maybe gating that to make sure it can't make its way into prod pathways).

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
tnowacki: It would be a tough nut to crack. Though I don't think anything _that_ bad would happen if someone called this in prod other than perf. 
We really should figure out the issues around `feature = "testing"` and transactional tests, but maybe an issue for another day

PR #25314, opened 2026-02-05
[consensus] use fifo compaction by default

mwtian:
## Description 

Since most of the consensus data are never updated or deleted, and insertions are mostly append, FIFO compaction policy should work. 

## Test plan 

PT

### DIFF
--- a/consensus/core/src/storage/rocksdb_store.rs
+++ b/consensus/core/src/storage/rocksdb_store.rs
@@ -55,18 +55,22 @@ impl RocksDBStore {
 
     /// Creates a new instance of RocksDB storage.
     #[cfg(not(tidehunter))]
-    pub fn new(path: &str) -> Self {
+    pub fn new(path: &str, use_fifo_compaction: bool) -> Self {
         // Consensus data has high write throughput (all transactions) and is rarely read
         // (only during recovery and when helping peers catch up).
         let db_options = default_db_options().optimize_db_for_write_throughput(2);
         let mut metrics_conf = MetricConf::new("consensus");
         metrics_conf.read_sample_interval = SamplingInterval::new(Duration::from_secs(60), 0);
-        let cf_options = default_db_options().optimize_for_write_throughput();
+        let cf_options = if use_fifo_compaction {
+            default_db_options().optimize_for_no_deletion()
+        } else {
+            default_db_options().optimize_for_write_throughput()
+        };
         let column_family_options = DBMapTableConfigMap::new(BTreeMap::from([
             (
                 Self::BLOCKS_CF.to_string(),
-                default_db_options()
-                    .optimize_for_write_throughput_no_deletion()
+                cf_options
+                    .clone()
                     // Using larger block is ok since there is not much point reads on the cf.
                     .set_block_options(512, 128 << 10),
             ),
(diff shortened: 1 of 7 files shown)


crates/typed-store/src/rocks/options.rs
```diff
@@ -214,12 +213,12 @@ impl DBOptions {
...
         // Switch to universal compactions.
         self.options
-            .set_compaction_style(rocksdb::DBCompactionStyle::Universal);
-        let mut compaction_options = rocksdb::UniversalCompactOptions::default();
-        compaction_options.set_max_size_amplification_percent(10000);
-        compaction_options.set_stop_style(rocksdb::UniversalCompactionStopStyle::Similar);
+            .set_compaction_style(rocksdb::DBCompactionStyle::Fifo);
+        let mut compaction_options = rocksdb::FifoCompactOptions::default();
+        // Allow each consensus DB to grow to 1 PB, essentially unlimited.
+        compaction_options.set_max_table_files_size(1 << 50);
```
akichidis: I haven't managed to look into much detail to the API, but is this the only property taken into account to trigger a compaction? Asking because api suggests some caution with Fifo to avoid loss of data - and that's why I assume you used a practically unlimited option

consensus/core/src/storage/rocksdb_store.rs
```diff
@@ -55,18 +55,22 @@ impl RocksDBStore {
...
-    pub fn new(path: &str) -> Self {
+    pub fn new(path: &str, use_fifo_compaction: bool) -> Self {
         // Consensus data has high write throughput (all transactions) and is rarely read
         // (only during recovery and when helping peers catch up).
         let db_options = default_db_options().optimize_db_for_write_throughput(2);
         let mut metrics_conf = MetricConf::new("consensus");
         metrics_conf.read_sample_interval = SamplingInterval::new(Duration::from_secs(60), 0);
-        let cf_options = default_db_options().optimize_for_write_throughput();
+        let cf_options = if use_fifo_compaction {
+            default_db_options().optimize_for_no_deletion()
```
akichidis, line 65: In consensus we practically never delete any data. Isn't this an optimisation for write throughput as well?

crates/typed-store/src/rocks/options.rs
mwtian: This is the underlying option https://github.com/facebook/rocksdb/blob/8.10.fb/include/rocksdb/advanced_options.h#L224. This seems to be the most suitable one to turn off data deletions. We don't need to use the advanced options.

consensus/core/src/storage/rocksdb_store.rs
mwtian: Not sure I understand the question. Consensus is the only caller of this method. Is this a naming suggestion?

crates/typed-store/src/rocks/options.rs
```diff
@@ -214,12 +213,12 @@ impl DBOptions {
...
         // Switch to universal compactions.
         self.options
-            .set_compaction_style(rocksdb::DBCompactionStyle::Universal);
-        let mut compaction_options = rocksdb::UniversalCompactOptions::default();
-        compaction_options.set_max_size_amplification_percent(10000);
-        compaction_options.set_stop_style(rocksdb::UniversalCompactionStopStyle::Similar);
+            .set_compaction_style(rocksdb::DBCompactionStyle::Fifo);
+        let mut compaction_options = rocksdb::FifoCompactOptions::default();
+        // Allow each consensus DB column family to grow unlimited, and never drop data because of size limits.
+        compaction_options.set_max_table_files_size(u64::MAX);
```
akichidis, line 219: It looks that after all `u64::MAX` is safe to use right? That's probably better for practically undeletable data.

akichidis (approved):
I am approving to unblock given the tests showcase performance improvement and no noticeable degradation during other cases (ex node catch up).

crates/typed-store/src/rocks/options.rs
mwtian: Yes I checked and it is safe to use.

PR #25319, opened 2026-02-06
[consensus] remove consensus inflight tracking and fastpath certified status

mwtian:
## Description 

These are no longer used after removing fastpath execution.

## Test plan 

CI

### DIFF
--- a/crates/sui-core/src/unit_tests/wait_for_effects_tests.rs
+++ b/crates/sui-core/src/unit_tests/wait_for_effects_tests.rs
@@ -255,7 +255,7 @@ async fn test_wait_for_effects_ping_rejected_with_reason() {
     tokio::spawn(async move {
         tokio::time::sleep(Duration::from_millis(100)).await;
         let epoch_store = state_clone.epoch_store_for_testing();
-        epoch_store.set_consensus_tx_status(tx_position, ConsensusTxStatus::Dropped);
+        epoch_store.set_consensus_tx_status(tx_position, ConsensusTxStatus::Rejected);
         epoch_store.set_rejection_vote_reason(
             tx_position,
             &SuiErrorKind::UserInputError {
--- a/crates/sui-core/src/authority_server.rs
+++ b/crates/sui-core/src/authority_server.rs
@@ -1168,27 +1168,65 @@ impl ValidatorService {
         };
         let tx_digests = [tx_digest];
 
-        let effects = self
-            .state
-            .get_transaction_cache_reader()
-            .notify_read_executed_effects_may_fail(
-                "AuthorityServer::wait_for_effects::notify_read_executed_effects_finalized",
-                &tx_digests,
-            )
-            .await?
-            .pop()
-            .unwrap();
-        let effects_digest = effects.digest();
-        let details = if request.include_details {
-            Some(self.complete_executed_data(effects).await?)
-        } else {
-            None
+        // When consensus_position is provided, also watch the consensus status cache
+        // so rejected/dropped transactions get a timely response instead of waiting
+        // forever for effects that will never be produced.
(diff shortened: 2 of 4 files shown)


crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -524,10 +518,6 @@ pub struct AuthorityEpochTables {
     /// epoch change.
     consensus_message_processed: DBMap<SequencedConsensusTransactionKey, bool>,
 
-    /// Map stores pending transactions that this authority submitted to consensus
-    #[default_options_override_fn = "pending_consensus_transactions_table_default_config"]
-    pending_consensus_transactions: DBMap<ConsensusTransactionKey, ConsensusTransaction>,
```
akichidis, line 529: Is that backwards compatible? I kind of remember that removing tables is not, but I can't recall 100%

crates/sui-core/src/unit_tests/wait_for_effects_tests.rs
```diff
@@ -238,30 +234,28 @@ async fn test_wait_for_effects_timeout() {
...
+        include_details: false,
+        ping_type: Some(PingType::Consensus),
     };
 
     let state_clone = test_context.state.clone();
     tokio::spawn(async move {
         tokio::time::sleep(Duration::from_millis(100)).await;
         let epoch_store = state_clone.epoch_store_for_testing();
-        epoch_store.set_consensus_tx_status(tx_position, ConsensusTxStatus::Rejected);
+        epoch_store.set_consensus_tx_status(tx_position, ConsensusTxStatus::Dropped);
```
akichidis: Why dropped? Aren't we testing here the `reject` path?

crates/sui-core/src/validator_client_monitor/mod.rs
```diff
@@ -30,8 +30,8 @@ impl OperationType {
             OperationType::Submit => "submit",
             OperationType::Effects => "effects",
             OperationType::HealthCheck => "health_check",
-            OperationType::FastPath => "fast_path",
-            OperationType::Consensus => "consensus",
+            OperationType::SingleWriterFinality => "single_writer_finality",
+            OperationType::SharedObjectFinality => "shared_object_finality",
```
akichidis, line 34: Nice renaming 👍

crates/sui-types/src/messages_grpc.rs
```diff
@@ -481,15 +470,11 @@ pub enum PingType {
     /// Measures the end to end latency from when a transaction is included by a proposed block,
     /// to when the block is committed by consensus.
     Consensus = 0,
-    /// Measures the end to end latency from when a transaction is included by a proposed block,
-    /// to when the block is certified.
-    FastPath = 1,
```
akichidis, line 486: Is this backwards compatible with the clients?

crates/sui-core/src/authority/authority_per_epoch_store.rs
mwtian: Indeed this was the case before. Now there is `populate_missing_cfs()` so this should be fine.

crates/sui-types/src/messages_grpc.rs
mwtian: `PingType::FastPath` usage has dropped from standard clients. Validators will return an error when converting raw enum to `PingType` if they receive unrecognized variants.

crates/sui-core/src/unit_tests/wait_for_effects_tests.rs
mwtian: Good point. There is an issue with Rejected status handling. Updated.

PR #25324, opened 2026-02-06
Disable object_store internal retries for all cloud backends

nickvikeras:
## Description

object_store's default retry logic hides transient errors (429s, 5xx) from the framework's own retry/backoff, preventing callers from reacting to pressure signals. 

In order to build adaptive concurrency limits that react to upstream/downstream errors, we need the framework to _see_ the errors.

## Test plan

## Release notes

- [x] Indexing Framework: Disable object_store crate internal retries to make errors immediately visible to framework.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
@@ -12,7 +12,6 @@ use backoff::backoff::Constant;
 use bytes::Bytes;
 use object_store::ClientOptions;
 use object_store::ObjectStore;
-use object_store::RetryConfig;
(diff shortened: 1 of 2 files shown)


crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
```diff
@@ -133,6 +134,15 @@ impl IngestionClientArgs {
         };
         options
     }
+
+    /// Disable object_store's internal retries so that transient errors (429s, 5xx) propagate
+    /// immediately to the framework's own retry logic.
+    fn retry_config() -> RetryConfig {
```
amnn: nit: given this is specific to `StoreClient`, why not make it a privat top-level function or `const` there?

PR #25326, opened 2026-02-06
docs: properly enable indexer pruning in config

evan-wall-mysten:
## Description 

Fixes the indexer config toml files to properly enable pruning on supported pipelines.

Also groups indexer config in an `indexer` dir to more easily figure out which configs apply to the indexer.

Incorporates changes from https://github.com/MystenLabs/sui/pull/25265

## Test plan 

Tested locally with `pnpm start`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/docs/content/guides/operator/indexer-stack-setup.mdx
+++ b/docs/content/guides/operator/indexer-stack-setup.mdx
@@ -75,14 +75,18 @@ Example:
 
 ```sh
 $ sui-indexer-alt indexer \
-    --config misc.toml \
+    --config unpruned.toml \
     --database-url postgres://username:password@localhost:5432/database \
     --remote-store-url https://checkpoints.mainnet.sui.io
 ```
 
 ### Run config recommendations {#indexer-run-config}
 
-We recommend using the TOML files below which are grouped by pipeline speed. All pipelines in an instance are limited by the slowest pipeline in that instance so these files each contain pipelines that run at approximately the same speed.
+Use the TOML files below; they are grouped by pipeline speed. All pipelines in an instance are limited by the slowest pipeline in that instance so these files each contain pipelines that run at approximately the same speed.
+
+:::info
+When backfilling, you should set the `ingest-concurrency` to a higher value, e.g. 200, then reduce it to 20 for normal operation at network tip.
+:::
 
 <table>
 	<thead>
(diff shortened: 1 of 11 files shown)


docs/content/guides/operator/indexer-stack-setup.mdx
```diff
@@ -138,7 +138,7 @@ We recommend using the TOML files below which are grouped by pipeline speed. All
 			<td colspan={5}>
 				<details>
 					<summary>`misc.toml`</summary>
-					<ImportContent source="examples/prod-config/misc.toml" mode="code" />
+					<ImportContent source="examples/prod-config/indexer/misc.toml" mode="code" />
```
wlmyng: Could we rename this to something else that better conveys the significance of the tables and split out tx balance changes?

I think it'd also be nice if we could pick up some of the changes I made to the indexer stack setup doc https://github.com/MystenLabs/sui/pull/25265

Namely calling out bumping up ingestion concurrency and reclassifying tx balance changes (off the top of my head)

wlmyng: There's a section in the doc currently that says these tables have a 90 day retention, which is incorrect

evan-wall-mysten: We are now recommending to not prune these tables?

evan-wall-mysten: From @wlmyng on slack:
> Packages and displays can't be pruned 
> Graphql and other tables use cp sequence numbers for reference into epoch or tx number 
> And these are pretty lightweight but give important info ie relating to epoch

examples/prod-config/indexer/misc.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[pipeline.kv_epoch_ends.pruner]
+
+[pipeline.kv_epoch_starts.pruner]
+
+[pipeline.kv_feature_flags.pruner]
+
+[pipeline.kv_packages.pruner]
+
+[pipeline.kv_protocol_configs.pruner]
```
henryachen:
```suggestion
[pipeline.kv_epoch_ends]

[pipeline.kv_epoch_starts]

[pipeline.kv_feature_flags]

[pipeline.kv_packages]

[pipeline.kv_protocol_configs]
```
These should be indexed unpruned.

examples/prod-config/indexer/tx_balance_changes.toml
```diff
@@ -0,0 +1,6 @@
+[ingestion]
+ingest-concurrency = 20
+
+# Unpruned until graphql reads from KV
+
+[pipeline.tx_balance_changes]
```
emmazzz: Why do we have it in a separate toml? Maybe keep it in the unpruned.toml file?

evan-wall-mysten: I took this from @wlmyng's changes in https://github.com/MystenLabs/sui/pull/25265. I don't know why it was moved there. 

@henryachen / @amnn, do you know why this is being moved? Or should I undo this?

henryachen: It might be in preparation for being able to prune `tx_balance_changes` but we should leave it in `unpruned.toml` until we confirm that we can enable pruning for it.

docs/content/guides/operator/indexer-stack-setup.mdx
```diff
@@ -75,14 +75,15 @@ Example:
...
-    --config misc.toml \
+    --config unpruned.toml \
     --database-url postgres://username:password@localhost:5432/database \
     --remote-store-url https://checkpoints.mainnet.sui.io
 ```
 
 ### Run config recommendations {#indexer-run-config}
 
 We recommend using the TOML files below which are grouped by pipeline speed. All pipelines in an instance are limited by the slowest pipeline in that instance so these files each contain pipelines that run at approximately the same speed.
+Note: when backfilling, you should set the `ingest-concurrency` to a higher value, e.g. 200, then reduce it to 20 for normal operation at network tip.
```
jessiemongeon1:
```suggestion
Use the TOML files below; they are grouped by pipeline speed. All pipelines in an instance are limited by the slowest pipeline in that instance so these files each contain pipelines that run at approximately the same speed.

:::info
When backfilling, you should set the `ingest-concurrency` to a higher value, e.g. 200, then reduce it to 20 for normal operation at network tip.
:::
```

jessiemongeon1 (approved):

PR #25333, opened 2026-02-06
[to & td] log retry delay and error category

mwtian:
## Description 

Log retry delays and error categories.

## Test plan 

CI

### DIFF
--- a/crates/sui-core/src/transaction_driver/mod.rs
+++ b/crates/sui-core/src/transaction_driver/mod.rs
@@ -245,18 +245,20 @@ where
                                 .observe(attempts as f64);
                             if request.transaction.is_some() {
                                 tracing::info!(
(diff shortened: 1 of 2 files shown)


crates/sui-core/src/transaction_driver/mod.rs
```diff
@@ -276,6 +278,8 @@ where
                 } else {
                     backoff.next().unwrap()
                 };
+
+                tracing::debug!("Retrying after {:.3}s", delay.as_secs_f32());
```
akichidis, line 282: Do you want to include maybe the `overload` parameter as well in the log?

akichidis (approved):

crates/sui-core/src/transaction_driver/mod.rs
mwtian: Previous log lines should have included the error details and category.

PR #25334, opened 2026-02-06
Concurrent pipelines participate in backpressure

nickvikeras:
## Description

Previously only sequential pipelines reported commit_hi back to the
ingestion regulator. Concurrent pipelines were effectively invisible to
the broadcaster.

When I was trying to implement adaptive concurrency control without this in place,
the broadcaster would keep spawning new ingestion tasks even when the pipelines had
hit write throughput limits and eventually the system would just oom.

## Test plan

---

## Release notes

- [x] Indexing Framework: Enable ingestion backpressure for concurrent pipelines

### DIFF
--- a/crates/sui-indexer-alt-framework/src/lib.rs
+++ b/crates/sui-indexer-alt-framework/src/lib.rs
@@ -156,10 +156,6 @@ pub struct Indexer<S: Store> {
     /// unless overridden by [Self::default_next_checkpoint].
     first_ingestion_checkpoint: u64,
 
-    /// The minimum next_checkpoint across all sequential pipelines. This is used to initialize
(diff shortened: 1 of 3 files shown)


crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -237,7 +237,7 @@ impl<S: Store> Indexer<S> {
             },
             added_pipelines: BTreeSet::new(),
             first_ingestion_checkpoint: u64::MAX,
-            next_sequential_checkpoint: None,
+            initial_commit_hi: None,
```
amnn: Now that all pipelines participate in this, is its value ever different from `first_ingestion_checkpoint` (with the exception of using `u64::MAX` instead of `None` in the empty case)?

PR #25339, opened 2026-02-06
[bella-ciao][hardening] Checked addition + erroring on VM PC increments and one other place

tzakian:
## Description 

Make sure we can't overflow the PC no matter what.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


tzakian:
Absolutely, and that's going to be the next step I think. I'm going to go to checked add everywhere, but that's going to be a ton of tiny little places so figured it would be best to do this first and then I'll come through and do all the "less interesting" places in the next PR here.

PR #25340, opened 2026-02-06
[bella-ciao][hardening] disallow possible cast truncactions

tzakian:
## Description 

`as` can silently truncate numbers.

Nowhere that we are doing this today poses any issues, but towards being extremely paranoid and leaning into defense in depth, it's best to error in these places rather than silently overflowing.

This also adds in a 64 bit arch restriction, which is reasonable (although sadly the clippy lint is not able to pick up this restrictions 🙁).

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
@@ -1632,62 +1632,53 @@ macro_rules! comparison_op {
 }
 
 macro_rules! cast_integer {
-    ($func_name:ident, $target_type:ty, $max_value:expr, $unchecked_cast_method:ident) => {
+    ($func_name:ident, $target_type:ty) => {
         pub fn $func_name(self) -> PartialVMResult<$target_type> {
             use IntegerValue::*;
 
             match self {
-                U8(x) => Ok(move_binary_format::checked_as!(x, $target_type)?),
-                U16(x) => {
-                    if x > move_binary_format::checked_as!($max_value, u16)? {
-                        Err(partial_vm_error!(
-                            ARITHMETIC_ERROR,
-                            "Cannot cast u16({}) to {}",
-                            x,
-                            stringify!($target_type)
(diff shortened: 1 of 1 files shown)


external-crates/move/crates/move-vm-runtime/src/shared/mod.rs
```diff
@@ -39,6 +39,24 @@ macro_rules! partial_vm_error {
     }};
 }
 
+/// A helper macro for performing a checked cast from one type to another, returning a
+/// PartialVMError if the cast fails.
+#[macro_export]
```
cgswords: Nit: Should this go live with the others in `move-binary-format/src/lib.rs`? As `safe_as`?

tzakian: I like this idea. Let me go do that.

tzakian: Moved it, but kept the name as I preferred it to `safe_` (checked seems more appropriate than safe for what it's doing in this particular case)

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -1632,70 +1632,61 @@ macro_rules! comparison_op {
...
-                            stringify!($target_type)
-                        ))
-                    } else {
-                        Ok(move_binary_format::checked_as!(x, $target_type)?)
-                    }
-                }
+                U8(x) => <$target_type>::try_from(x).map_err(|_| {
+                    partial_vm_error!(
+                        ARITHMETIC_ERROR,
+                        "Cannot cast u16({}) to {}",
```
cgswords: Nit: `u8`

PR #25342, opened 2026-02-06
Decouple watermark advancement from DB writes

nickvikeras:
## Description

Watermark advancement (which feeds backpressure via commit_hi_tx) was
gated on the DB write timer. Now watermark progress through contiguous
entries happens on every loop iteration, while DB writes remain deferred
to the timer tick.

This was causing stalls in sui-checkpoint-blob-indexer, which uses a
1-minute watermark interval. The broadcaster's backpressure view of the
watermark was only updated once per minute, throttling ingestion far
below actual throughput.

The old code had inconsistent behavior on DB write failures: connect()
failures implicitly retried on the next tick (because watermark
advancement hadn't run yet, so entries stayed in the precommitted map),
while commit() failures dropped the watermark silently. The code already
documented the intended philosophy — "it's OK for the watermark to lag
from a correctness perspective" — so we now apply that consistently:
both failure types drop the pending watermark. A newer value will be
computed and written on the next successful tick when new parts arrive.

Previously, on shutdown the task would wait for the next timer tick to
write the final watermark. Now the shutdown path flushes any pending
watermark immediately after the loop exits. This makes shutdown faster
on average — particularly for sui-checkpoint-blob-indexer's 1-minute
watermark interval, where the old behavior could delay shutdown by up to
a minute.

Extracted write_watermark() helper to deduplicate the DB write + metrics
update logic between the loop body and the shutdown flush.

## Test plan

### DIFF
--- a/crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
@@ -20,6 +20,7 @@ use crate::pipeline::WARN_PENDING_WATERMARKS;
 use crate::pipeline::WatermarkPart;
 use crate::pipeline::concurrent::Handler;
 use crate::pipeline::logging::WatermarkLogger;
+use crate::store::CommitterWatermark;
 use crate::store::Connection;
 use crate::store::Store;
 use crate::store::pipeline_task;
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
```diff
@@ -195,56 +180,28 @@ pub(super) fn commit_watermark<H: Handler + 'static>(
...
-                            .with_label_values(&[H::NAME])
-                            .set(watermark.epoch_hi_inclusive as i64);
-
-                        metrics
-                            .watermark_transaction_in_db
-                            .with_label_values(&[H::NAME])
-                            .set(watermark.tx_hi as i64);
+            // DB writes are deferred to the timer interval to avoid excessive DB load.
+            if should_write_db {
+                if precommitted.len() > WARN_PENDING_WATERMARKS {
```
amnn: Should this warning be saved for the DB write? It seems like we should do it unconditionally after we've tried to clear the prefix of the watermark on every iteration of the loop.

nickvikeras: >Is there a reason to prefer the behaviour where we drop the watermark? It seems better to save it for next time, all things being equal.

I just went that route because of the comment in the code. I agree it probably is better to reset the pending watermark in case it doesn't update between the this tick and the next tick. I couldn't see any reason _not_ to do that, but I also wasn't 100% sure if there wasn't something subtle I was missing.

PR #25345, opened 2026-02-07
[move][move-vm][bella-ciao] Disallow `panic`, `expect`, and `unwrap` in the VM code.

cgswords:
## Description 

This disallows the listed operations, and carves out the few places we do use them. Note that this does not clean up the `unreachable` cases, or the issues with `panic` in the values implementation---both of those are already partially-handled by other PRs, and will get cleaned up as a separate step.

This also reintroduces panicking to the string interner, since it is the safest way to avoid forking.

This also revises some visibility on operations, though that work should also be extended in a follow-up PR.

## Test plan 

Tests all still pass (I hope).

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/dev_utils/mod.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/dev_utils/mod.rs
@@ -12,6 +12,7 @@
 // want to be able to use these constructs to write tests.
 #![allow(
     clippy::cast_possible_truncation,
+    clippy::unwrap_used,
     clippy::expect_used,
     clippy::indexing_slicing,
     clippy::new_without_default,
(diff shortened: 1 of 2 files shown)


external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -968,7 +972,17 @@ impl StructRef {
         let container = &mut *self.0.try_borrow_mut()?;
         match container {
             Value::Struct(fixed_vec) => {
-                fixed_vec.0.0[index] = mem_box;
+                let slots = &mut fixed_vec.0.0;
```
tzakian: Needed? This is a test function...

cgswords: No, but I was doing this as consistently as possible so it seems okay?

PR #25346, opened 2026-02-07
Move post_process_one_tx off critical path and batch index writes

mystenmark:
## Summary

- Spawn `post_process_one_tx` (JSON-RPC indexing + event emission) on a blocking thread instead of running inline during transaction execution
- Batch all index writes per-checkpoint into a single atomic RocksDB write via `StagedBatch` / `DBBatch::concat`, reducing write amplification
- Index batch is committed before transaction outputs, so crash recovery is safe (idempotent re-indexing on restart)
- `TransactionOrchestrator` waits on `transactions_executed_in_checkpoint_notify` instead of `await_post_processing`, since index data is guaranteed available at that point
- Backpressure via semaphore (`num_cpus` permits) prevents unbounded task accumulation
- `sync_post_process_one_tx` config flag preserves old inline behavior as a safety valve

## Test plan

- [x] `test_simulated_load_async_post_processing_consistency` simtest with crash injection — passed 200 seeds
- [x] `check_system_consistency` verifies all checkpointed transactions appear in `transactions_seq` at epoch boundaries
- [x] Dual-fullnode comparison (async vs sync) with rolling restarts and reconfiguration

## Implementation notes

Detailed design doc: https://gist.github.com/mystenmark/49b33c56ccd14f66564efd15c04a9c28

### DIFF
--- a/crates/sui-swarm-config/tests/snapshots/snapshot_tests__network_config_snapshot_matches.snap
+++ b/crates/sui-swarm-config/tests/snapshots/snapshot_tests__network_config_snapshot_matches.snap
@@ -26,6 +26,7 @@ validator_configs:
       submit-delay-step-override-millis: ~
       parameters: ~
     enable-index-processing: true
+    sync-post-process-one-tx: false
     jsonrpc-server-type: ~
     grpc-load-shed: ~
     grpc-concurrency-limit: 20000000000
@@ -193,6 +194,7 @@ validator_configs:
       submit-delay-step-override-millis: ~
       parameters: ~
     enable-index-processing: true
+    sync-post-process-one-tx: false
     jsonrpc-server-type: ~
     grpc-load-shed: ~
     grpc-concurrency-limit: 20000000000
@@ -360,6 +362,7 @@ validator_configs:
       submit-delay-step-override-millis: ~
       parameters: ~
     enable-index-processing: true
+    sync-post-process-one-tx: false
     jsonrpc-server-type: ~
     grpc-load-shed: ~
     grpc-concurrency-limit: 20000000000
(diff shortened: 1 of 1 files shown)


lxfind:
Are the indexes writes idempotent? For instance, if a node shuts down, and then restart with the new binary, we could end up re-indexing a few transactions, would that be an issue?

mystenmark:
> Are the indexes writes idempotent? For instance, if a node shuts down, and then restart with the new binary, we could end up re-indexing a few transactions, would that be an issue?

We can already re-index transactions in the current code. As far as I can tell this will not cause problems beyond the re-use of sequence numbers.

crates/sui-core/src/authority.rs
```diff
@@ -3225,69 +3264,205 @@ impl AuthorityState {
...
+            .indexes
+            .as_ref()
+            .map(|idx| idx.allocate_sequence_number());
+
+        if self.config.sync_post_process_one_tx {
+            // Synchronous mode: run post-processing inline on the calling thread
+            // and commit the index batch immediately with locks held.
+            // Used as a rollback mechanism and for testing correctness against async mode.
+            // TODO: delete this branch once async mode has shipped
+            let result = Self::post_process_one_tx_impl(
```
lxfind, line 3201: This will panic if `indexes` is `None`. Should we check here just in case?

crates/sui-core/src/authority.rs
```diff
@@ -3225,69 +3264,205 @@ impl AuthorityState {
...
-                    effects,
-                    events,
-                    timestamp_ms,
-                    tx_coins,
-                    written,
-                    inner_temporary_store,
-                    epoch_store,
+        let indexes = indexes
+            .as_ref()
+            .expect("post_process_one_tx_impl called without indexes");
```
lxfind: If we require `indexes` to be Some here, we might as well require caller to pass by direct reference instead of by Option.

crates/sui-core/src/authority.rs
mystenmark: we do check, at the top of the function

lxfind (approved):

PR #25348, opened 2026-02-07
Reduce transaction submission jitters

mwtian:
## Description 

- Start backup transaction submission requests when there is no response for 1s.
- Remove bias towards local execution. They are originally set to avoid submitting transactions when local results are available. Instead we can check for local execution first and poll select loop without bias. There might be unintended behavior with `biased`.

## Test plan 

CI

### DIFF
--- a/crates/sui-core/src/authority_server.rs
+++ b/crates/sui-core/src/authority_server.rs
@@ -1200,16 +1200,16 @@ impl ValidatorService {
                     "AuthorityServer::wait_for_effects::notify_read_executed_effects_finalized",
                     &tx_digests,
                 ) => {
-                tracing::Span::current().record("fast_path_effects", false);
                 let effects = effects_result?.pop().unwrap();
+                let effects_digest = effects.digest();
                 let details = if request.include_details {
-                    Some(self.complete_executed_data(effects.clone(), None).await?)
+                    Some(self.complete_executed_data(effects, None).await?)
                 } else {
                     None
                 };
 
                 Ok(WaitForEffectsResponse::Executed {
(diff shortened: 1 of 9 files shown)


crates/sui-core/src/authority_server.rs
lxfind: No longer need to record `fast_path_effects`?

crates/sui-core/src/authority_server.rs
```diff
@@ -1187,8 +1187,6 @@ impl ValidatorService {
             };
 
         tokio::select! {
-            // Ensure that finalized effects are always prioritized.
-            biased;
```
lxfind, line 1191: Why is this removed?

crates/sui-core/src/authority_server.rs
mwtian: Removed. Majority of fastpath cleanups will be done in separate PRs.

crates/sui-core/src/authority_server.rs
mwtian: On the client side, I observed the local effects future waiting for 28s, before the the 1st submission attempt was made. It can be a generic thread stall issue, but it can also be an issue with the `biased` specifier. So I'm removing all usages of `biased` for now. After fastpath execution is removed, this specific "biased" usage should be a no-op, because there is no transaction that is both finalized and fastpath executed.

lxfind (approved):

crates/sui-core/src/authority_server.rs
mystenmark: agreed that this was probably a no-op, but how could it possibly have caused any issues?

PR #25356, opened 2026-02-09
[gasless] support gasless transfers of balance

alex-mysten:
## Description 

Enable gasless balance transfers in testnet. 
* Users can set gas_price to 0 to indicate they intend to create a gasless txn. gas_budget must also be set to 0.
* We currently restrict gasless txns to three Move fns: withdrawal_split, redeem_funds, and send_funds
* We currently require net-zero storage; that is, we do not allow gasless transactions to write any object.

## Test plan 

simtests added

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-e2e-tests/tests/free_tier_tests.rs
```diff
@@ -0,0 +1,433 @@
...
+    let withdraw_arg = FundsWithdrawalArg::balance_from_sender(amount, GAS::type_tag());
+    let withdraw_arg = builder.funds_withdrawal(withdraw_arg).unwrap();
+    let amount_arg = builder.pure(U256::from(amount)).unwrap();
+    let split = builder.programmable_move_call(
+        SUI_FRAMEWORK_PACKAGE_ID,
+        Identifier::new("funds_accumulator").unwrap(),
+        Identifier::new("withdrawal_split").unwrap(),
+        vec!["0x2::balance::Balance<0x2::sui::SUI>".parse().unwrap()],
+        vec![withdraw_arg, amount_arg],
+    );
```
mystenmark: why split the full amount out of the withdrawal instead of passing the input withdrawal directly to balance::redeem_funds?

crates/sui-e2e-tests/tests/free_tier_tests.rs
```diff
@@ -0,0 +1,433 @@
...
```
mystenmark: ditto

crates/sui-e2e-tests/tests/free_tier_tests.rs
```diff
@@ -0,0 +1,433 @@
...
+    gas_coin::GAS,
+    programmable_transaction_builder::ProgrammableTransactionBuilder,
+    transaction::{
+        FundsWithdrawalArg, GasData, TransactionData, TransactionDataV1, TransactionExpiration,
+        TransactionKind,
+    },
+};
+use test_cluster::addr_balance_test_env::{TestEnv, TestEnvBuilder, get_sui_balance};
+
+fn create_free_tier_transaction(
```
mystenmark: can we move this to TestEnv so that we don't have to pass in rgp/chain_id/etc?

Also price should be 0, not rgp for a free tier tx

crates/sui-e2e-tests/tests/free_tier_tests.rs
```diff
@@ -0,0 +1,433 @@
...
+    TestEnvBuilder::new()
+        .with_proto_override_cb(Box::new(|_, mut cfg| {
+            cfg.enable_free_tier_for_testing();
+            cfg
+        }))
+        .build()
+        .await
+}
+
+fn db_sui_balance(test_env: &TestEnv, owner: SuiAddress) -> u64 {
```
mystenmark: we already have `TestEnv::get_sui_balance` and friends which can replace this func.

crates/sui-e2e-tests/tests/gasless_tests.rs
```diff
@@ -0,0 +1,433 @@
...
+    let result = test_env.exec_tx_directly(tx).await;
+
+    let (_, effects) = result.unwrap();
+
+    assert!(
+        effects.status().is_ok(),
+        "Free tier transfer should succeed: {:?}",
+        effects.status()
+    );
+    assert_zero_gas(effects.gas_cost_summary());
```
mystenmark: lets end all of these tests with `trigger_reconfiguration` just to check for conservation errors

crates/sui-e2e-tests/tests/free_tier_tests.rs
```diff
@@ -0,0 +1,433 @@
...
+    test_env
+        .fund_one_address_balance(sender, 10_000_000_000)
+        .await;
+
+    let tx_kind = build_gasless_send_funds_ptb(1000, sender);
+    let tx = create_free_tier_transaction(tx_kind, sender, test_env.rgp, test_env.chain_id, 0, 0);
+    let result = test_env.exec_tx_directly(tx).await;
+
+    assert!(
+        result.is_err(),
```
mystenmark: do you mind doing: `assert!(err.to_string().contains("expected error message"));` - need to be sure that if the check ever breaks the test doesn't continue passing because of some unrelated reason.

(same comment applies to any error test)

crates/sui-e2e-tests/tests/free_tier_tests.rs
```diff
@@ -0,0 +1,433 @@
...
+        "Should be rejected at signing (insufficient withdrawal balance)"
+    );
+
+    let sender_balance_after = db_sui_balance(&test_env, sender);
+    assert_eq!(sender_balance_after, sender_balance_before);
```
mystenmark: this test is a bit odd - if we are rejected at signing time its physically impossible to be charged gas. We need to see a TransactionEffects for this test to be meaningful

what we probably want is a test where we submit 2 txns via soft bundle so that one of them fails post scheduling.

crates/sui-framework/packages/sui-framework/sources/balance.move
```diff
@@ -103,6 +103,14 @@ public fun send_funds<T>(balance: Balance<T>, recipient: address) {
     sui::funds_accumulator::add_impl(balance, recipient);
 }
 
+/// Gasless (free tier) transfer:
+/// Send a `Balance` to an address's funds accumulator.
+/// Basis points (BPs) may be charged on the transfer amount
+/// in the future.
+public fun gasless_send_funds<T>(balance: Balance<T>, recipient: address) {
+    sui::funds_accumulator::add_impl(balance, recipient);
```
mystenmark: how is this gated? FYI we can call `public(package) native fun is_feature_enabled(feature_flag_name: vector<u8>): bool;` to check feature gates from move now if that helps

crates/sui-types/src/transaction.rs
```diff
@@ -1941,6 +1976,10 @@ pub fn is_gas_paid_from_address_balance(
         )
 }
 
+pub fn is_free_tier_transaction(gas_data: &GasData, transaction_kind: &TransactionKind) -> bool {
+    is_gas_paid_from_address_balance(gas_data, transaction_kind) && gas_data.budget == 0
```
mystenmark: ah i think we had a miscommunication: `price == 0` should be the signal that it is free tier. This way all gas  prioritization code (present and future) works without special cases.

crates/sui-e2e-tests/tests/free_tier_tests.rs
```diff
@@ -0,0 +1,433 @@
...
+fn build_gasless_send_funds_ptb(amount: u64, recipient: SuiAddress) -> TransactionKind {
+    let mut builder = ProgrammableTransactionBuilder::new();
+    let withdraw_arg = FundsWithdrawalArg::balance_from_sender(amount, GAS::type_tag());
+    let withdraw_arg = builder.funds_withdrawal(withdraw_arg).unwrap();
+    let amount_arg = builder.pure(U256::from(amount)).unwrap();
+    let split = builder.programmable_move_call(
+        SUI_FRAMEWORK_PACKAGE_ID,
+        Identifier::new("funds_accumulator").unwrap(),
+        Identifier::new("withdrawal_split").unwrap(),
+        vec!["0x2::balance::Balance<0x2::sui::SUI>".parse().unwrap()],
```
mystenmark: i think it would be best if we did these tests with an alternate currency, so that we can be sure that things work correctly for accounts that do not hold SUI.

This will also make various bugs into conservation check failures, for instance, if we allow an object to be written somehow, the storage rebate would have to come from nowhere.

lxfind:
Would this apply to all Coin types?

crates/sui-e2e-tests/tests/free_tier_tests.rs
alex-mysten: just to test all whitelisted functions. I've changed this, the "normal case" exercises one recipient, there's now a happy case testing multiple recipients. more than happy to drop withdrawal_split from the whitelist if that's better

crates/sui-e2e-tests/tests/free_tier_tests.rs
alex-mysten: ""

crates/sui-e2e-tests/tests/free_tier_tests.rs
alex-mysten: price == 0 is now the trigger for free tier

crates/sui-e2e-tests/tests/free_tier_tests.rs
alex-mysten: updated

crates/sui-e2e-tests/tests/gasless_tests.rs
alex-mysten: done

crates/sui-e2e-tests/tests/free_tier_tests.rs
alex-mysten: done

crates/sui-e2e-tests/tests/free_tier_tests.rs
alex-mysten: deleted for now - still need to add soft bundle test

crates/sui-framework/packages/sui-framework/sources/balance.move
alex-mysten: added is_feature_enabled check, but it's now just defensive:

nonfree transactions may never call `gasless_send_funds`
free tier transactions may only call `gasless_send_funds` iff free_tier_enabled = 1

crates/sui-types/src/transaction.rs
alex-mysten: price == 0 is now the directive to enable free tier. we then have to decide how to treat a nonzero budget - ignore or reject. i'm rejecting txns with nonzero budget as I think it is less ambiguous.

crates/sui-e2e-tests/tests/free_tier_tests.rs
alex-mysten: I've rewritten tests where it is useful to use non-sui

alex-mysten:
> Would this apply to all Coin types?

Added an allow list which presently contains only Sui - we can iterate here.

lxfind:
> Added an allow list which presently contains only Sui - we can iterate here.

Got it. Btw I thought we would not allow SUI here, but only a list of stable coins.

alex-mysten:
> > Added an allow list which presently contains only Sui - we can iterate here.
> 
> Got it. Btw I thought we would not allow SUI here, but only a list of stable coins.

This was news to me on Friday, anyway I've added a protocol-config defined token allow list (currently empty)

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -551,6 +551,11 @@ impl<'backing> TemporaryStore<'backing> {
             .fold(0, |sum, obj| sum + obj.object_size_for_gas_metering())
     }
 
+    pub fn has_non_accumulator_writes(&self) -> bool {
+        !self.execution_results.written_objects.is_empty()
+            || !self.execution_results.deleted_object_ids.is_empty()
```
mystenmark, line 569: this is good for now, although we will have to decide if we want to tackle the "coins can be consumed/deleted" in one go or add a flag later

mystenmark:
I guess with gas_price: 0, no changes to the gas code were required at all?

edit: i see there are changes to gas_model, but nothing to gas_charger.rs

crates/sui-framework/packages/sui-framework/sources/balance.move
```diff
@@ -103,6 +105,20 @@ public fun send_funds<T>(balance: Balance<T>, recipient: address) {
     sui::funds_accumulator::add_impl(balance, recipient);
 }
 
+/// Gasless (free tier) transfer:
+/// Send a `Balance` to an address's funds accumulator.
+/// Basis points (BPs) may be charged on the transfer amount
+/// in the future.
+/// Not `public` so that it cannot be called from other Move modules;
+/// only callable from free tier PTBs.
+entry fun gasless_send_funds<T>(balance: Balance<T>, recipient: address) {
```
dariorussi: is this still going in? I am confused as to what I am reviewing

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -404,6 +404,16 @@ mod checked {
             Ok((r, t)) => (Ok(r), t),
             Err((e, t)) => (Err(e), t),
         };
+        if gas_charger.is_free_tier()
+            && result.is_ok()
+            && temporary_store.has_non_accumulator_writes()
+        {
+            // safety net - we only allow ptb commands that cannot write objects.
+            result = Err(ExecutionError::new_with_source(
+                ExecutionErrorKind::InsufficientGas,
```
tzakian: Bit of a nit, but should this be a different error? 

If I squint hard enough I can believe this is an OK error, but also at the same time it seems like the wrong error in this case probably? (I'd expect something like `DisallowedOperation` or something along those lines)

sui-execution/latest/sui-adapter/src/temporary_store.rs
alex-mysten: I will leave this up to Andrew

alex-mysten:
> I guess with gas_price: 0, no changes to the gas code were required at all?
> 
> edit: i see there are changes to gas_model, but nothing to gas_charger.rs

Correct. `gas_price` is only used in `bucketize_computation`. Gasless transactions are charged as gassy transactions, except that all gas_costs are zero. 

Upstream gating prevents us from having positive storage or rebate values. It seems more future-proof to not special-case gas charging logic so it behaves unsurprisingly as rules evolve.

For the record, if a gasless transaction manages to create an object, it will fail with `InsufficientGas`, as `gas_budget == gas_price == 0`. If a gasless transaction manages to delete an object, the sender will receive a rebate.

crates/sui-types/src/gas_model/gas_v2.rs
```diff
@@ -307,10 +323,33 @@ mod checked {
             )
         }
 
+        pub(crate) fn new_free_tier(
+            reference_gas_price: u64,
+            config: &ProtocolConfig,
+        ) -> SuiGasStatus {
+            let metering_price = reference_gas_price.max(1);
+            let max_computation_units = config.free_tier_max_computation_units();
+            let compute_cap = max_computation_units * metering_price;
```
tzakian: Even with the `max(1)` above I always kinda like having `debug_assert!`s as checkable documentation of invariants, so it may be worth considering adding

```
debug_assert!(metering_price > 0);
``` 

before this.

crates/sui-types/src/gas_model/gas_v2.rs
```diff
@@ -279,20 +277,38 @@ mod checked {
...
+                config.storage_gas_price(),
                 config.storage_rebate_rate(),
                 gas_rounding_mode,
                 sui_cost_table,
             )
         }
 
+        pub(crate) fn new_with_budget(
+            gas_budget: u64,
+            gas_price: u64,
```
tzakian: Consider adding a `debug_assert!` here that gas_price (and maybe the others) are non-zero.

crates/sui-types/src/transaction.rs
```diff
@@ -964,6 +970,91 @@ impl ProgrammableTransaction {
...
+        let mut allowed_token_types = parse_free_tier_allowed_token_types(config);
+        fail_point_arg!("free_tier_extra_token_types", |extra: HashSet<
+            TypeInput,
+        >| {
+            allowed_token_types.extend(extra);
+        });
+        for command in &self.commands {
+            match command {
+                Command::MoveCall(call) if call.package == SUI_FRAMEWORK_PACKAGE_ID => {
+                    let is_balance_call = call.module.as_str() == BALANCE_MODULE_NAME.as_str()
```
tzakian: Nit: maybe consider seeing (or creating) a helper function like `is_call(call, package_id, module_name, function_name) -> bool` that does this type of checking.

crates/sui-types/src/gas_model/gas_v2.rs
```diff
@@ -219,6 +219,8 @@ mod checked {
         unmetered_storage_rebate: u64,
         /// Rounding mode for gas charges.
         gas_rounding_mode: GasRoundingMode,
+        // Max computation units for free tier. Set only for free tier transactions.
+        free_tier_max_computation_units: Option<u64>,
```
dariorussi: I am not clear this belongs here?
I would expect at this level to have no knowledge of the free tier?
Also I would look at what @tnowacki is doing for the charger and see if we can plug it there.
Thoughts?

crates/sui-types/src/transaction.rs
```diff
@@ -964,6 +970,91 @@ impl ProgrammableTransaction {
...
+                        "Free tier transactions can only call balance::send_funds, balance::redeem_funds, and funds_accumulator::withdrawal_split".to_string(),
+                    ));
+                }
+            }
+        }
+        Ok(())
+    }
+}
+
+fn parse_free_tier_allowed_token_types(config: &ProtocolConfig) -> HashSet<TypeInput> {
```
aschran: minor nit - Vec is probably simpler and faster here than HashSet, given the small size of the list?

crates/sui-framework/packages/sui-framework/sources/balance.move
alex-mysten: No - we removed this in [this commit]( https://github.com/MystenLabs/sui/pull/25356/changes/89e966db99f2c9b1075fe1aa00d2f2e2603f2c72) as per our decision last week.

crates/sui-types/src/gas_model/gas_v2.rs
alex-mysten: This is a member of the struct as we use the value in `bucketize_computation`, to handle the edge case where vm has barely enough compute budget but bucketizing bumps us above budget. We could also decide it's not important to handle this - gas_price is 0 so everything multiplies out to compute=0 without changes. Happy to discuss options

dariorussi: let's discuss further after Todd is back. I was hoping the gas charger knows "nothing" about free tiers but I may not be looking at this properly. 
Give me another day or two please

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -128,6 +128,14 @@ pub mod checked {
             self.gas_status.is_unmetered()
         }
 
+        pub fn is_address_balance(&self) -> bool {
+            self.payment_method.is_address_balance()
+        }
+
+        pub fn is_free_tier(&self) -> bool {
+            self.gas_status.gas_price() == 0 && self.payment_method.is_address_balance()
```
lxfind: Should we use gas status's `is_free_tier` for consistency? Could these two implementations diverge in the future?

sui-execution/latest/sui-adapter/src/execution_engine.rs
alex-mysten: This is actually more of an assertion - added debug_fatal + InvariantViolation. We can change this later as the allow list evolves.

crates/sui-types/src/gas_model/gas_v2.rs
alex-mysten: Added assertion

crates/sui-types/src/gas_model/gas_v2.rs
alex-mysten: Added assertion for budget. We assert on gas_price in new_metered

crates/sui-types/src/transaction.rs
alex-mysten: Added - much cleaner - thanks!

crates/sui-types/src/transaction.rs
alex-mysten: Changed to Vec<>. Thanks!

sui-execution/latest/sui-adapter/src/gas_charger.rs
alex-mysten: Yes - we now delegate to gas_status.is_gasless()

crates/sui-types/src/gas_model/gas_v2.rs
```diff
@@ -251,23 +253,20 @@ mod checked {
...
+                gasless_max_computation_units: None,
             }
         }
 
-        pub(crate) fn new_with_budget(
+        fn new_metered(
             gas_budget: u64,
             gas_price: u64,
+            metering_price: u64,
+            computation_budget: u64,
```
tnowacki: Having these two separate sets of fields confuses me, even with the context of the code below. Can we document this function?

crates/sui-types/src/transaction.rs
```diff
@@ -964,6 +970,106 @@ impl ProgrammableTransaction {
...
+                        SUI_FRAMEWORK_PACKAGE_ID,
+                        BALANCE_MODULE_NAME.as_str(),
+                        BALANCE_REDEEM_FUNDS_FUNCTION_NAME.as_str(),
+                    );
+                    let is_withdrawal_split = is_move_call(
+                        call,
+                        SUI_FRAMEWORK_PACKAGE_ID,
+                        FUNDS_ACCUMULATOR_MODULE_NAME.as_str(),
+                        FUNDS_ACCUMULATOR_WITHDRAWAL_SPLIT_FUNCTION_NAME.as_str(),
+                    );
```
tnowacki: I think we should try to write something a bit more robust here to deal with any future functions or changes to these rules in a quick manner. 
We also have a bad habit of fucking these sort of rules up. 

I would look at the private generics verifier (note the v2 in that I totally didn't fuck it up in v1 😉) https://github.com/MystenLabs/sui/blob/89b58bc27a382a638babb21664cc063ecc8a934f/sui-execution/latest/sui-verifier/src/private_generics_verifier_v2.rs#L121 

This let's us set up a more declarative approach of allowed functions, and rules for their generics. 

I could see just a simple bool for now, though if we add other functions I could see it becoming an enum to specify what exactly needs to be done with the type arg.

crates/sui-types/src/transaction.rs
```diff
@@ -964,6 +970,106 @@ impl ProgrammableTransaction {
...
+fn extract_balance_inner_type(ti: &TypeInput) -> Option<&TypeInput> {
+    if let TypeInput::Struct(s) = ti
+        && s.address == SUI_FRAMEWORK_ADDRESS
+        && s.module == BALANCE_MODULE_NAME.as_str()
+        && s.name == BALANCE_STRUCT_NAME.as_str()
+        && s.type_params.len() == 1
+    {
+        return Some(&s.type_params[0]);
+    }
+    None
```
tnowacki: Do we need a TypeInput here still? @tzakian I always forget this shit.

If we can have a tag, this is just `Balance::maybe_get_balance_type_param` from `crates/sui-types/src/balance.rs`

crates/sui-types/src/transaction.rs
```diff
@@ -964,6 +970,106 @@ impl ProgrammableTransaction {
...
+}
+
+fn parse_gasless_allowed_token_types(config: &ProtocolConfig) -> Vec<TypeInput> {
+    config
+        .gasless_allowed_token_types()
+        .iter()
+        .map(|s| {
+            let tag: TypeTag = s
+                .parse()
+                .unwrap_or_else(|e| panic!("invalid gasless token type {s:?}: {e}"));
```
tnowacki: Is this panic safe?

crates/sui-types/src/transaction.rs
```diff
@@ -964,6 +970,106 @@ impl ProgrammableTransaction {
...
+fn is_move_call(
+    call: &ProgrammableMoveCall,
+    package: ObjectID,
+    module: &str,
+    function: &str,
+) -> bool {
+    call.package == package && call.module.as_str() == module && call.function.as_str() == function
+}
+
+fn parse_gasless_allowed_token_types(config: &ProtocolConfig) -> Vec<TypeInput> {
```
tnowacki: BTreeSet?

tnowacki: Though I understand for small numbers of these it is going to probably be faster with a vec

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -128,6 +128,14 @@ pub mod checked {
...
         }
 
+        pub fn is_address_balance(&self) -> bool {
+            self.payment_method.is_address_balance()
+        }
+
+        pub fn is_gasless(&self) -> bool {
+            self.gas_status.is_gasless()
+        }
+
```
tnowacki: We will probably want a new PaymentMethod for this after `25728` lands... maybe. 
Or maybe we turn `Unmetered` into `Free` 🤷 not sure

crates/sui-types/src/transaction.rs
aschran: I changed this already in a follow up PR anyway btw

crates/sui-types/src/transaction.rs
alex-mysten: This was very helpful - function allow list is now declarative, and each accepts a fn to validate / extract the type param - this should be extensible as we support Coin @aschran

crates/sui-types/src/transaction.rs
alex-mysten: We're now using Balance::maybe_get_balance_type_param and we compare against a Vec of TypeTag

crates/sui-types/src/transaction.rs
alex-mysten: I was relying on CI to tell us if the protocol config is broken. `simtest-mainnet` would tell us, but there's nothing that would tell us if testnet was broken. I've changed this to an debug_assert and silent skip to make it a bit safer.

sui-execution/latest/sui-adapter/src/gas_charger.rs
alex-mysten: Happy to add a new PaymentMethod or not, whatever makes it easier to reason about. I intentionally avoided control flow changes to gas_charger, just touching how we compute totals in gas_v2. As is, we only touch bucketize_computation - otherwise gasless is exactly the same as address balance

crates/sui-types/src/gas_model/gas_v2.rs
alex-mysten: Added documentation for this function

crates/sui-types/src/gas_model/gas_v2.rs
```diff
@@ -307,10 +333,34 @@ mod checked {
...
+            );
+            status.gasless_max_computation_units = Some(max_computation_units);
+            status
+        }
+
         pub fn reference_gas_price(&self) -> u64 {
             self.reference_gas_price
         }
 
+        fn is_gasless(&self) -> bool {
```
lxfind: This function isn't needed since there is already a trait impl

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: I am not sure it is a different PaymentMethod or a "different" charger. I could not find a good way to put that "knowledge" on the payment method and it seemed more natural on the charger with some better API break up

crates/sui-types/src/gas_model/gas_v2.rs
alex-mysten: Method removed with latest rev

crates/sui-protocol-config/src/lib.rs
```diff
@@ -1030,6 +1030,12 @@ struct FeatureFlags {
     // When split-checkpoints enabled, merge randomness and non-randomness schedulables together.
     #[serde(skip_serializing_if = "is_false")]
     merge_randomness_into_checkpoint: bool,
+
+    #[serde(skip_serializing_if = "is_false")]
+    new_vm_enabled: bool,
+
```
tnowacki, line 1040: Bad merge? This flag was removed

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -391,13 +394,19 @@ mod checked {
...
+        is_gasless: bool,
     ) -> SuiResult<SuiGasStatus> {
         let gas_budget = transaction.gas_budget();
         let gas_price = transaction.gas_price();
         let gas_paid_from_address_balance = transaction.is_gas_paid_from_address_balance();
 
-        let gas_status =
-            SuiGasStatus::new(gas_budget, gas_price, reference_gas_price, protocol_config)?;
+        let gas_status = if is_gasless {
+            let rgp = reference_gas_price.max(1);
```
tnowacki, line 406: Is `reference_gas_price` ever 0?

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -414,6 +422,15 @@ mod checked {
...
             Err((e, t)) => (Err(e), t),
         };
+        if gas_charger.is_gasless()
+            && result.is_ok()
+            && temporary_store.has_non_accumulator_writes()
+        {
+            debug_fatal!(
+                "gasless transaction produced non-accumulator writes — command validation should have prevented this"
+            );
+            result = Err(ExecutionErrorKind::InvariantViolation.into());
```
tnowacki: Been a longggg time since I have had this code in my head, so apologies for the potentially dumb question:
Do we need to reset the writes in the temporary store at this point then? Is that handled somewhere else? Should this be handled inside of charge gas maybe...? though that definitely is a bit strange

And is there a test we can write for this case? I know it is hard since it is an invariant violation but just curious if it is possible--it would be good to know we aren't modifying objects if we get this far.

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -320,6 +325,10 @@ pub mod checked {
             self.smash_gas(temporary_store);
         }
 
+        pub fn is_gasless(&self) -> bool {
+            matches!(&self.payment, PaymentMetadata::Gasless)
+        }
```
tnowacki: Where is this used? And should we instead be asking it from the transaction itself or do we feel this is the correct abstraction?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -391,14 +400,20 @@ pub mod checked {
...
                     // If we don't have enough balance to withdraw, don't charge for gas
                     // TODO: consider charging gas if we have enough to reserve but not enough to cover all withdraws
                     return GasCostSummary::default();
             }
 
             self.compute_storage_and_rebate(temporary_store, execution_result);
+
             let cost_summary = self.gas_status.summary();
+
+            let Some(gas_payment_location) = gas_payment_location else {
```
tnowacki, line 410: I really don't like the way this is carved up. I mean that in that I didn't like how it was when I landed the last refactor, and I still don't like it lol. And now I like it a bit less with this additional payment kind. I'm not sure what to do here and it doesn't need to be done in this PR, but we should do _something_ to de-spaghettify this a bit.

tnowacki: In other words it just feels a bit forced right now, and the differences/similarities between the payment kinds does not feel well encapsulated.

crates/sui-types/src/transaction.rs
```diff
@@ -964,6 +972,117 @@ impl ProgrammableTransaction {
...
+        Ok(())
+    }
+}
+
+struct GaslessFunction {
+    package: AccountAddress,
+    module: &'static IdentStr,
+    function: &'static IdentStr,
+    extract: fn(&[TypeInput]) -> Result<TypeTag, UserInputError>,
+}
```
tnowacki: This is not quite what I had in mind, but I will need to dive in a bit more to see how exactly I feel about it (so give me a bit, sorry)

crates/sui-protocol-config/src/lib.rs
alex-mysten: Bad merge, will fix

crates/sui-transaction-checks/src/lib.rs
alex-mysten: In practice, no. This was defensive against testing scenarios. I can invariant violation if this is 0 - probably safer

crates/sui-types/src/transaction.rs
alex-mysten: The concern is with the extract function living on the struct? What I interpreted from your original comment was to have some enum to describe to the caller what must be done with the TypeInput to extract the Tag, which I tried - this was much simpler.

sui-execution/latest/sui-adapter/src/execution_engine.rs
alex-mysten: Writes are for sure dumped, just like for any other execution failure, in charge_gas:

```
fn charge_gas(result) {
  ...
  if result.is_err() {
     self.reset()
  }
  ....
fn reset() {
   temporary_store.drop_writes()
  ...
}
```

we actually err the execution result and dump writes today even later, within charge_gas, if bucketize_computation causes us to OOG. 

We'd have to bypass upstream checks to hit this defensive block, and the write-dumping logic is the same as in mainnet, which is why I didn't explicitly test this - but happy to add one if you think it would meaningfully add value.

sui-execution/latest/sui-adapter/src/gas_charger.rs
alex-mysten: Currently used in the `execution_engine`. We can also determine this from the transaction and kill this method. Will update

sui-execution/latest/sui-adapter/src/gas_charger.rs
alex-mysten: Yes - we need to carve out the variant and invariant behavior and figure out what kind of pattern we can apply here to better encapsulate the variant logic. I'm happy to explore this as a followup

sui-execution/latest/sui-adapter/src/execution_engine.rs
tnowacki: It's sort of wild this is done in the gas charger (cc @dariorussi for thinking about this)

crates/sui-types/src/balance.rs
```diff
@@ -28,6 +28,9 @@ pub const BALANCE_DESTROY_REBATES_FUNCTION_NAME: &IdentStr = ident_str!("destroy
 
 pub const BALANCE_REDEEM_FUNDS_FUNCTION_NAME: &IdentStr = ident_str!("redeem_funds");
 pub const BALANCE_SEND_FUNDS_FUNCTION_NAME: &IdentStr = ident_str!("send_funds");
+pub const FUNDS_ACCUMULATOR_MODULE_NAME: &IdentStr = ident_str!("funds_accumulator");
+pub const FUNDS_ACCUMULATOR_WITHDRAWAL_SPLIT_FUNCTION_NAME: &IdentStr =
+    ident_str!("withdrawal_split");
```
tnowacki: These should be in `crates/sui-types/src/funds_accumulator.rs` since they are a member of that module and not `balance`

crates/sui-types/src/transaction.rs
tnowacki: The concern is that this doesn't feel like a hardened set of validation checks, nor does it feel extensible for the future as these rules evolve over time. It will be easier to discuss outside of this PR

alex-mysten: Updated per offline feedback

crates/sui-types/src/transaction.rs
```diff
@@ -1082,6 +1118,96 @@ impl ProgrammableMoveCall {
...
+            (SUI_FUNDS_ACCUMULATOR_WITHDRAWAL_SPLIT, &[Some(BalanceType)]),
+        ];
+
+        let Some((_, type_arg_constraints)) =
+            GASLESS_FUNCTIONS
+                .iter()
+                .find(|((addr, module, function), _)| {
+                    *addr == AccountAddress::from(self.package)
+                        && module.as_str() == self.module.as_str()
+                        && function.as_str() == self.function.as_str()
```
tnowacki: nit, should only need the as_str on the IdentStr? The other one is a string? (IIRC)
So maybe just `&`?

crates/sui-e2e-tests/tests/gasless_tests.rs
```diff
@@ -0,0 +1,687 @@
+// Copyright (c) Mysten Labs, Inc.
```
tnowacki, line 1: Since we just need the gas budget to 0 and an empty gas vector, we should be able to have transactional tests for this then yeah? I think it would be a lot easier to reason through than the Rust tests personally

alex-mysten: All gasless_tests are now transactional tests except for a few scenarios. One happy-case e2e test that also runs transaction driver, and two scenarios that required additional protocol config manipulation that I didn't want to move to transactional tests.

crates/sui-types/src/transaction.rs
alex-mysten: Removed `as_str` on `self.module` and `self.function`

crates/sui-types/src/balance.rs
alex-mysten: Relocated

crates/sui-transactional-test-runner/src/args.rs
```diff
@@ -302,6 +305,12 @@ pub struct AuthenticatorStateUpdateCommand {
     pub authenticator_obj_initial_shared_version: Option<u64>,
 }
 
+#[derive(Debug, clap::Parser)]
+pub struct GaslessAllowTokenCommand {
+    #[clap(long = "type")]
```
tnowacki: We probably don't need a flag for this yeah? Like it can just be 
```
//# gasless-allow-token <type>
```
no need for the `--type` flag

crates/sui-transactional-test-runner/src/args.rs
```diff
@@ -302,6 +305,12 @@ pub struct AuthenticatorStateUpdateCommand {
...
+    pub token_type: String,
```
tnowacki: This should also be 
```suggestion
    #[clap(value_parser = ParsedType::parse)]
    pub token_type: ParsedType,
```

crates/sui-transactional-test-runner/src/test_adapter.rs
```diff
@@ -826,6 +832,17 @@ impl MoveTestAdapter<'_> for SuiTestAdapter {
...
             }
+            SuiSubcommand::GaslessAllowToken(GaslessAllowTokenCommand { token_type }) => {
+                let resolved = if let Some(prefix) = token_type.split("::").next() {
+                    let state = self.compiled_state();
+                    let addr = state.resolve_named_address(prefix);
+                    token_type.replacen(prefix, &format!("0x{addr}"), 1)
+                } else {
+                    token_type
+                };
+                sui_types::transaction::add_gasless_token_for_testing(resolved);
```
tnowacki: With `ParsedType` you will not need any of this parsing logic, though you will still need the named address resolving

crates/sui-types/src/transaction.rs
```diff
@@ -958,12 +966,59 @@ pub struct ProgrammableTransaction {
     pub commands: Vec<Command>,
 }
 
+static GASLESS_TOKENS_FOR_TESTING: RwLock<Vec<String>> = RwLock::new(Vec::new());
```
tnowacki, line 972: Can we `cfg` gate this?

crates/sui-types/src/transaction.rs
```diff
@@ -958,12 +966,59 @@ pub struct ProgrammableTransaction {
...
+                .iter()
+                .filter_map(|s| match s.parse() {
+                    Ok(tag) => Some(tag),
+                    Err(e) => {
+                        debug_fatal!("invalid gasless token type {s:?}: {e}");
+                        None
+                    }
+                })
+                .collect();
+            if mysten_common::in_test_configuration() {
```
tnowacki: If we can't `cfg` gate this, should we assert otherwise that `GASLESS_TOKENS_FOR_TESTING` is empty? Not sure it matters I suppose... just being paranoid

crates/sui-types/src/transaction.rs
```diff
@@ -958,12 +966,59 @@ pub struct ProgrammableTransaction {
...
+                            debug_fatal!("invalid gasless token override {s:?}: {e}");
+                        }
+                    }
+                }
+            }
+            types
+        }
+
+        let allowed_token_types = parse_gasless_allowed_token_types(config);
+
```
tnowacki, line 1044: As @tzakian suggested, this will be a good spot to assert that `commands` is non-empty

crates/sui-types/src/transaction.rs
```diff
@@ -958,12 +966,59 @@ pub struct ProgrammableTransaction {
...
+                        Ok(tag) => types.push(tag),
+                        Err(e) => {
+                            debug_fatal!("invalid gasless token override {s:?}: {e}");
+                        }
+                    }
+                }
+            }
+            types
+        }
+
```
tnowacki, line 1042: Should we be asserting/checking here that there aren't any objects? Or other weird things like `Receiving`? 
Can their be additional reservations/withdrawals? 
I think we are okay with pure inputs though?

crates/sui-adapter-transactional-tests/tests/gasless/gasless_reject.move
```diff
@@ -0,0 +1,77 @@
...
+//> 1: sui::balance::send_funds(Result(0), Input(1));
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs withdraw<sui::balance::Balance<test::usdc::USDC>>(500) @B
+// Reject: send_funds with two type args
+//> 0: sui::balance::redeem_funds<test::usdc::USDC>(Input(0));
+//> 1: sui::balance::send_funds<test::usdc::USDC, test::usdc::USDC>(Result(0), Input(1));
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs withdraw<sui::balance::Balance<test::usdc::USDC>>(500) 250u256
+// Reject: withdrawal_split with non-Balance type arg
+//> 0: sui::funds_accumulator::withdrawal_split<test::usdc::USDC>(Input(0), Input(1));
```
tnowacki, line 81: Fascinating. Maybe we should change the allowlist logic slightly. I had not considered this

crates/sui-adapter-transactional-tests/tests/gasless/gasless_multi_recipient.move
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Mysten Labs, Inc.
```
tnowacki, line 1: The tests here look fine but we probably want more, I might split up the reject file a bit and have a file for allowed/disallowed functions. 
Like showing you can't do even innocuous things right now like balance::value() or withdrawal::limit() 

(we should consider adding such things but we probably want a test for each allowed function in some separate file)

crates/sui-adapter-transactional-tests/tests/gasless/gasless_reject.move
alex-mysten: we guard against this with the type constraints we introduced - don't think we need to add anything immediately here. we will reject anything that is not Balance<T>

crates/sui-types/src/transaction.rs
alex-mysten: Added assertion above

crates/sui-types/src/transaction.rs
alex-mysten: Added input type checking above

crates/sui-adapter-transactional-tests/tests/gasless/gasless_multi_recipient.move
alex-mysten: split into allowed and rejected function tests - allowed methods are intended to be used in conjunction so dividing them further did not make sense

crates/sui-types/src/transaction.rs
alex-mysten: `cfg` gated with `debug_assertions` - TIL

crates/sui-types/src/transaction.rs
alex-mysten: Done

crates/sui-transactional-test-runner/src/test_adapter.rs
alex-mysten: Done

crates/sui-transactional-test-runner/src/args.rs
alex-mysten: Addressed

crates/sui-transactional-test-runner/src/args.rs
alex-mysten: Fixed

crates/sui-adapter-transactional-tests/tests/gasless/gasless_reject.move
tnowacki: Yeah its just I didn't consider the non-balance case really. I don't think we will need it ever, but if we do the code is in a spot that is easy enough to refactor

crates/sui-types/src/transaction.rs
tnowacki: Isn't it sort of incidental that debug_assertions are used for tests? It feels like a strange gating but maybe that is okay...

crates/sui-types/src/transaction.rs
```diff
@@ -958,12 +968,79 @@ pub struct ProgrammableTransaction {
...
 impl ProgrammableTransaction {
     pub fn has_shared_inputs(&self) -> bool {
         self.inputs
             .iter()
             .any(|input| matches!(input, CallArg::Object(ObjectArg::SharedObject { .. })))
     }
+
+    pub fn validate_gasless_transaction(&self, config: &ProtocolConfig) -> UserInputResult {
+        fp_ensure!(
+            !self.commands.is_empty(),
```
tnowacki, line 996: Do we have a limit somewhere on the number of commands?

crates/sui-types/src/transaction.rs
```diff
@@ -4936,3 +5145,254 @@ impl TransactionKey {
         }
     }
 }
+
+#[cfg(test)]
+mod gasless_validation_tests {
```
tnowacki: I think it is a bit strange to have these tests in this file directly. Just in that we haven't done before for anything else and I wouldn't expect to go looking for tests here

crates/sui-types/src/transaction.rs
```diff
@@ -958,12 +968,79 @@ pub struct ProgrammableTransaction {
...
+                    )));
+                }
+            }
+        }
+
+        fn parse_gasless_allowed_token_types(config: &ProtocolConfig) -> Vec<TypeTag> {
+            let mut types: Vec<TypeTag> = config
+                .gasless_allowed_token_types()
+                .iter()
+                .filter_map(|(s, _min_transfer_size)| match s.parse() {
```
tnowacki: Should we be ignoring the `_min_transfer_size` right now? Should we be asserting that it is `0`?

crates/sui-types/src/transaction.rs
alex-mysten: Discussed over Slack, Mark's pr will address this

crates/sui-types/src/transaction.rs
alex-mysten: Confirmed these are redundant with transactional tests and don't add value. deleted

crates/sui-types/src/transaction.rs
alex-mysten: added debug_assert that this is 0

PR #25357, opened 2026-02-09
[state-sync] Disconnect continuously failing peers

alex-mysten:
## Description 

Disconnect peers that are failing continuously. This will allow peer discovery to connect to another peer. Currently our concurrent connection pool is limited to 4 peers. Discovery selects a peer at random to establish a connection to, so it's possible we reconnect to the same peer.

## Test plan 

Simtest

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-network/src/discovery/mod.rs
+++ b/crates/sui-network/src/discovery/mod.rs
@@ -71,9 +71,8 @@ pub enum DiscoveryMessage {
     /// A spawned peer-query task discovered updated info for a configured peer,
     /// signaling the event loop to update the stored peer addresses.
     ConfiguredPeersUpdated,
-    PeerFailureReport {
-        peer_id: PeerId,
-    },
+    /// A peer has been reported as continuously failing, triggering disconnect and cooldown.
+    PeerFailureReport { peer_id: PeerId },
 }
 
 /// A Handle to the Discovery subsystem. The Discovery system will be shut down once all Handles


crates/sui-config/src/p2p.rs
```diff
@@ -255,6 +255,20 @@ pub struct StateSyncConfig {
     /// If unspecified, this will default to `0.3` (30%).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub peer_failure_rate: Option<f64>,
+
+    /// Duration in milliseconds that a peer must be continuously failing before it is
+    /// disconnected. Set to 0 to disable automatic disconnect of failing peers.
+    ///
+    /// If unspecified, this will default to `300,000` (5 minutes).
```
aschran: 5 minutes seems like a really long time to be loyal to a failing peer? wdyt?

crates/sui-network/src/state_sync/mod.rs
```diff
@@ -909,6 +964,37 @@ where
...
+
+        let same_chain_connected_count = peer_heights
+            .peers_on_same_chain()
+            .filter(|(peer_id, _)| self.network.peer(**peer_id).is_some())
+            .count();
+        if same_chain_connected_count <= min_peers {
+            return;
+        }
+
+        let Some(peer_id) = peer_heights.find_peer_to_disconnect(threshold) else {
```
aschran: this probably needs to exempt preconfigured peers from the disconnection

crates/sui-config/src/p2p.rs
alex-mysten: Lowered to 2 minutes. Note that peers are deprioritized within the connected peer selection first and only after this timeout are disconnected.

crates/sui-network/src/state_sync/mod.rs
alex-mysten: Done in discovery

crates/sui-config/src/p2p.rs
aschran: I think we likely still want to make it lower, bc there are a limited number of connection slots, so even if they are deprioritized, it's only relative to other connected peers and not the much larger pool of peers we might connect to but aren't attempting until this timeout expires. wdyt? admittedly it should be balanced against not wanting to do too much "sloshing"

crates/sui-config/src/p2p.rs
```diff
@@ -379,12 +393,24 @@ impl StateSyncConfig {
 
     pub fn exploration_probability(&self) -> f64 {
         const DEFAULT: f64 = 0.1;
-        self.exploration_probability.unwrap_or(DEFAULT)
+        self.exploration_probability
+            .unwrap_or(DEFAULT)
+            .clamp(0.0, 1.0)
```
aschran: nit - since it would be a mistake in the node config if this were outside [0, 1] i might actually panic here instead of silently clamping

crates/sui-network/src/discovery/mod.rs
```diff
@@ -644,37 +680,54 @@ impl DiscoveryEventLoop {
...
+            .discovery_config
+            .target_concurrent_connections()
+            .saturating_sub(number_of_connections);
 
-        // randomize the order
         use rand::seq::IteratorRandom;
-        for (peer_id, info) in eligible
+        let mut rng = rand::thread_rng();
+
+        let preferred_to_dial = std::cmp::min(preferred.len(), number_to_dial);
```
aschran: nit simplifications - 
choose_multiple already handles the case where you request more items than the iterator has, so I think the `min` is unnecessary? as well as the `if remaining > 0` because calling `choose_multiple` with 0 will just do a no-op extend

crates/sui-network/src/state_sync/mod.rs
```diff
@@ -126,6 +127,11 @@ impl Handle {
     pub fn subscribe_to_synced_checkpoints(&self) -> broadcast::Receiver<VerifiedCheckpoint> {
         self.checkpoint_event_sender.subscribe()
     }
+
+    /// Returns the number of peers disconnected due to consistent state sync failures.
+    pub fn get_peers_disconnected_for_failure(&self) -> u64 {
```
aschran: should this metric be moved to discovery now? (or at least name changed to "reported" vs "disconnected" if it stays here?)

crates/sui-network/src/state_sync/mod.rs
```diff
@@ -434,6 +464,31 @@ impl PeerHeights {
...
             .unwrap_or(false)
     }
+
+    pub fn update_failing_states(&mut self) {
+        for score in self.scores.values_mut() {
+            score.update_failing_state();
+        }
+    }
+
+    pub fn find_peer_to_disconnect(&self, threshold: Duration) -> Option<PeerId> {
```
aschran: to_report?

crates/sui-network/src/state_sync/mod.rs
```diff
@@ -909,6 +967,41 @@ where
         if let Some(layer) = self.download_limit_layer.as_ref() {
             layer.maybe_prune_map();
         }
+
+        self.maybe_disconnect_failing_peer();
+    }
+
+    fn maybe_disconnect_failing_peer(&mut self) {
```
aschran: maybe_report_...?

crates/sui-network/src/state_sync/mod.rs
```diff
@@ -909,6 +967,41 @@ where
...
+        let min_peers = self.config.min_peers_for_disconnect();
+
+        let mut peer_heights = self.peer_heights.write().unwrap();
+        peer_heights.update_failing_states();
+
+        let same_chain_connected_count = peer_heights
+            .peers_on_same_chain()
+            .filter(|(peer_id, _)| self.network.peer(**peer_id).is_some())
+            .count();
+        if same_chain_connected_count <= min_peers {
```
aschran: it seems to me that logic about tracking minimum peers we want to keep connected even in case of failures should be handled in discovery and not here? in my mind the division of responsibilities is: state sync just reports when peers are misbehaving, and discovery is what makes all the decisions about whom to connect to

crates/sui-config/src/p2p.rs
alex-mysten: You're right - lowered to 30s.

crates/sui-config/src/p2p.rs
alex-mysten: Agreed - now panicking on invalid config

crates/sui-network/src/discovery/mod.rs
alex-mysten: Addressed

crates/sui-network/src/state_sync/mod.rs
alex-mysten: Renamed the metric

crates/sui-network/src/state_sync/mod.rs
alex-mysten: Renamed

crates/sui-network/src/state_sync/mod.rs
alex-mysten: Renamed

crates/sui-network/src/state_sync/mod.rs
alex-mysten: Agreed - this separation of concerns is more sensible. I've moved the disconnection logic entirely to disovery. Thanks!

crates/sui-network/src/discovery/mod.rs
```diff
@@ -71,6 +71,9 @@ pub enum DiscoveryMessage {
     /// A spawned peer-query task discovered updated info for a configured peer,
     /// signaling the event loop to update the stored peer addresses.
     ConfiguredPeersUpdated,
+    PeerFailureReport {
```
aschran: nit-add doc comment?

PR #25364, opened 2026-02-09
[antithesis] randomize checkpoint split limit for reachability assert

alex-mysten:
## Description 

Lower the checkpoint split limit so we can assert reachability

## Test plan 

[steka-antithesis-cp-split-reachability 3430e844 4.0h](https://mystenlabs.antithesis.com/report/ZxSVmPkRhggT8uE06sAPf24U/8wWUOHsk4BFMv0hjPDJnvSqARNjVXdKcd5HoRcnWlho.html?auth=v2.public.eyJuYmYiOiIyMDI2LTAyLTE3VDAwOjUzOjA4Ljc4NDkyNTIyMloiLCJzY29wZSI6eyJSZXBvcnRTY29wZVYxIjp7ImFzc2V0IjoiOHdXVU9Ic2s0QkZNdjBoalBESm52U3FBUk5qVlhkS2NkNUhvUmNuV2xoby5odG1sIiwicmVwb3J0X2lkIjoiWnhTVm1Qa1JoZ2dUOHVFMDZzQVBmMjRVIn19fR9cbuhCp8uVXZPngeTBo1OwukkBAKAo7isIOEjEtILTkScnDPETk6NupSkMUVxkdHaKYn_LFKTeDxg3xv1l6gk)
[steka-antithesis-cp-split-reachability mainnet upgrade split-version 4.0h](https://mystenlabs.antithesis.com/report/8slJysoNBugHn_CwKruqbihm/UY_kqoRdK0Qlgz21DNY8u2dv5Dk_iGsQnDbUTfBf9fU.html?auth=v2.public.eyJzY29wZSI6eyJSZXBvcnRTY29wZVYxIjp7ImFzc2V0IjoiVVlfa3FvUmRLMFFsZ3oyMUROWTh1MmR2NURrX2lHc1FuRGJVVGZCZjlmVS5odG1sIiwicmVwb3J0X2lkIjoiOHNsSnlzb05CdWdIbl9Dd0tydXFiaWhtIn19LCJuYmYiOiIyMDI2LTAyLTE3VDAxOjE4OjI1Ljc4NTAxOTI1NVoifR7ITGuMFJFEytjMk7yCqfKaVN7zs_5tZYlGqFBUA9lSXdkNtJxLpc0c_2TXCEDdsXgaiGVzf1U5wirO7FekUQw)

---

## Release notes

- [x] Protocol: test only changes, no user impact
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/checkpoints/mod.rs
+++ b/crates/sui-core/src/checkpoints/mod.rs
@@ -2000,15 +2000,12 @@ impl CheckpointBuilder {
                 sorted.extend(settlement_effects);
             }
 
-            all_effects.extend(sorted);
-        }
+            #[cfg(msim)]
+            {
+                self.expensive_consensus_commit_prologue_invariants_check(&root_digests, &sorted);
+            }
 
-        #[cfg(msim)]
-        {
-            self.expensive_consensus_commit_prologue_invariants_check_v2(
-                &all_root_digests,
-                &all_effects,
-            );
+            all_effects.extend(sorted);
         }
         Ok((all_effects, all_root_digests.into_iter().collect()))
     }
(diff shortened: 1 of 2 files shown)


crates/sui-protocol-config/src/lib.rs
```diff
@@ -4602,15 +4609,30 @@ impl ProtocolConfig {
...
+        {
+            return;
         }
 
-        cfg
+        let should_apply = if cfg!(msim) {
+            true
+        } else {
+            mysten_common::in_antithesis()
+        };
```
mystenmark: you can use `mysten_common::in_test_configuration()`

crates/sui-protocol-config/src/lib.rs
```diff
@@ -989,6 +990,9 @@ struct FeatureFlags {
     // If true perform consistent verification of metadata
     #[serde(skip_serializing_if = "is_false")]
     validator_metadata_verify_v2: bool,
+
+    #[serde(skip_serializing_if = "is_false")]
+    randomize_checkpoint_tx_limit_in_tests: bool,
```
mystenmark, line 1005: let's make this a catch-all flag, `randomize_params_for_tests` and then below we will use this as a signal to randomize anything that it makes sense to randomize - we'll probably add more things than just this

alex-mysten: I think these need to be individualized for anything that is an existing protocol config. 

The reason I added this is because without this we do an antithesis upgrade test, we get a fork, because the max_txns_per_checkpoint would change mid-epoch.

If we generalize this protocol config, any additional randomization we add here would also fork antithesis, as the whole block is gated by one flag.

We could also not gate it, knowing that the only risk is potentially misleading antithesis upgrade forks when we introduce new randomization.

crates/sui-protocol-config/src/lib.rs
alex-mysten: Addressed

crates/sui-protocol-config/src/lib.rs
mystenmark: ok, good point, we can also revisit this later

mystenmark (approved):

PR #25378, opened 2026-02-09
Add V2 bridge CLI commands (deposit-on-eth, deposit-on-sui)

Bridgerz:
## Summary

Adds V2 bridge support to `sui-bridge-cli` for deposit commands, and adds a new `claim-on-sui` command.

### Changes

- **`deposit-native-ether-on-eth --v2`**: Calls `bridgeETHV2` on the Ethereum `SuiBridge` contract, which emits `TokensDepositedV2` including `block.timestamp` for the limiter-bypass flow.
- **`deposit-on-sui --v2`**: Calls `send_token_v2` in the Sui Move bridge module, passing the `Clock` object to include `timestamp_ms` in the V2 payload.
- **`claim-on-sui`** (new): Claims bridged tokens on Sui by calling `bridge::claim_and_transfer_token<T>`. Handles both V1 and V2 messages transparently.
- **`claim-on-eth`**: Already handles both V1 and V2 messages transparently (no `--v2` flag needed).

### Design Decisions

- Deposit commands use a `--v2` flag to opt into the V2 flow (which includes timestamp in the payload for limiter bypass).
- Claim commands do **not** have a `--v2` flag because the underlying bridge logic already handles V1/V2 messages transparently based on the message version field.
- Common ETH deposit logic is extracted into `deposit_native_ether_on_eth_internal` to avoid duplication.

### Context

See `bridge/SUI_NATIVE_BRIDGE_PRIMER.md` for details on the V2 limiter-bypass mechanism.

### DIFF
--- a/crates/sui-bridge-cli/src/lib.rs
+++ b/crates/sui-bridge-cli/src/lib.rs
@@ -21,7 +21,6 @@ use sui_bridge::abi::EthBridgeCommittee;
 use sui_bridge::abi::{EthSuiBridge, eth_sui_bridge};
 use sui_bridge::crypto::BridgeAuthorityPublicKeyBytes;
 use sui_bridge::encoding::TOKEN_TRANSFER_MESSAGE_VERSION_V2;
-use sui_bridge::error::BridgeResult;
 use sui_bridge::sui_client::SuiBridgeClient;
 use sui_bridge::types::BridgeAction;
 use sui_bridge::types::{
@@ -628,17 +627,13 @@ impl BridgeClientCommands {
                 .await
             }
             BridgeClientCommands::ClaimOnEth { seq_num, dry_run } => {
-                claim_on_eth(seq_num, config, sui_bridge_client, dry_run)
-                    .await
-                    .map_err(|e| anyhow!("{:?}", e))
+                claim_on_eth(seq_num, config, sui_bridge_client, dry_run).await
             }
             BridgeClientCommands::ClaimOnSui {
                 seq_num,
                 source_chain,
                 dry_run,
-            } => claim_on_sui(seq_num, source_chain, config, sui_bridge_client, dry_run)
-                .await
-                .map_err(|e| anyhow!("{:?}", e)),
+            } => claim_on_sui(seq_num, source_chain, config, sui_bridge_client, dry_run).await,
         }
     }
 }
@@ -785,9 +780,11 @@ async fn deposit_on_sui(
             );
             Ok(())
         }
-        sui_json_rpc_types::SuiExecutionStatus::Failure { error } => {
-            Err(anyhow!("Transaction {:?} failed: {:?}", tx_digest, error))
-        }
+        sui_json_rpc_types::SuiExecutionStatus::Failure { error } => Err(anyhow!(
+            "Deposit ({version:?}) transaction {:?} failed: {:?}",
+            tx_digest,
+            error
+        )),
     }
 }
 
@@ -796,11 +793,16 @@ async fn claim_on_eth(
     config: &LoadedBridgeCliConfig,
     sui_bridge_client: SuiBridgeClient,
     dry_run: bool,
-) -> BridgeResult<()> {
-    let sui_chain_id = sui_bridge_client.get_bridge_summary().await?.chain_id;
+) -> anyhow::Result<()> {
+    let sui_chain_id = sui_bridge_client
+        .get_bridge_summary()
+        .await
+        .map_err(|e| anyhow!("{:?}", e))?
+        .chain_id;
     let parsed_message = sui_bridge_client
         .get_parsed_token_transfer_message(sui_chain_id, seq_num)
-        .await?;
+        .await
+        .map_err(|e| anyhow!("{:?}", e))?;
     if parsed_message.is_none() {
         println!("No record found for seq_num: {seq_num}, chain id: {sui_chain_id}");
         return Ok(());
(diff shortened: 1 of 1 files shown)


crates/sui-bridge-cli/src/lib.rs
```diff
@@ -527,6 +527,9 @@ pub enum BridgeClientCommands {
         target_chain: u8,
         #[clap(long)]
         sui_recipient_address: SuiAddress,
+        /// Use the V2 bridge path (timestamp-aware, limiter bypass for mature messages)
+        #[clap(long, default_value_t = false)]
+        v2: bool,
```
dlukegordon: Nit: should we make this an enum instead? So the field could be something like bridge_version, with V1 and V2 variants. Also should we be defaulting to V2 instead of V1?

crates/sui-bridge-cli/src/lib.rs
```diff
@@ -601,20 +605,72 @@ impl BridgeClientCommands {
...
+) -> anyhow::Result<()> {
+    let eth_sui_bridge = EthSuiBridge::new(
+        config.eth_bridge_proxy_address,
+        config.eth_signer_provider().clone(),
+    );
+    // Note: even with f64 there may still be loss of precision even there are a lot of 0s
+    let int_part = ether_amount.trunc() as u64;
+    let frac_part = ether_amount.fract();
+    let int_wei = U256::from(int_part) * U256::from(10).pow(U256::from(18));
+    let frac_wei = U256::from((frac_part * 1_000_000_000_000_000_000f64) as u64);
```
dlukegordon: I think you can replace `1_000_000_000_000_000_000f64` with `1e18`.

crates/sui-bridge-cli/src/lib.rs
```diff
@@ -538,14 +541,30 @@ pub enum BridgeClientCommands {
...
     ClaimOnEth {
         #[clap(long)]
         seq_num: u64,
         #[clap(long, default_value_t = true, action = clap::ArgAction::Set)]
         dry_run: bool,
     },
+    /// Claim bridged tokens on Sui for an ETH→Sui transfer that has been approved but not yet claimed.
+    /// Auto-detects V1/V2 from the on-chain record (V2 messages >48h old bypass the rate limiter).
+    #[clap(name = "claim-on-sui")]
+    ClaimOnSui {
```
dlukegordon, line 575: Should we add a dry_run flag like we have for ClaimOnEth?

Bridgerz: not a bad idea. will add now

crates/sui-bridge-cli/src/lib.rs
```diff
@@ -754,13 +837,113 @@ async fn claim_on_eth(
...
+    let parsed_message = sui_bridge_client
+        .get_parsed_token_transfer_message(source_chain, seq_num)
+        .await?;
+    let Some(parsed_message) = parsed_message else {
+        println!("No record found for seq_num: {seq_num}, source chain: {source_chain}");
+        return Ok(());
+    };
+
+    let message_version = parsed_message.message_version;
+    let version_label = if message_version == TOKEN_TRANSFER_MESSAGE_VERSION_V2 {
```
dlukegordon, line 885: Looks like `message_version` and `version_label` are only used for log messages. Is this expected, we don't need to pass anything about the version into the move call?

crates/sui-bridge-cli/src/lib.rs
dlukegordon: Should we add `version_label` to the err message?

crates/sui-bridge-cli/src/lib.rs
```diff
@@ -601,20 +605,72 @@ impl BridgeClientCommands {
...
+
 async fn deposit_on_sui(
     coin_object_id: ObjectID,
     coin_type: TypeTag,
     target_chain: BridgeChainId,
     recipient_address: EthAddress,
     config: &LoadedBridgeCliConfig,
     sui_bridge_client: SuiBridgeClient,
+    use_v2: bool,
 ) -> anyhow::Result<()> {
```
dlukegordon, line 685: Should we be returning `BridgeResult` here to match the other functions?

crates/sui-bridge-cli/src/lib.rs
bmwill: i missed the floating point. we should probably use fixed-point arithmetic operations

crates/sui-bridge-cli/src/lib.rs
Bridgerz: no the claim function call determines what version to use dynamically (it looks up the token transfer record directly, something we can't do on the eth side)

crates/sui-bridge-cli/src/lib.rs
dlukegordon: I think 1e18 can resolve to either an f64 or u64 depending on context.

Bridgerz: tbh I don't ever use the CLI to kick off deposits, but good call. Will update

bmwill: the `ether_amount` is an f64. it should be a `String` that we then convert to a U256 via parsing the whole and fractional parts on either side of a `.` so that we have no loss in precision.

I do realize that some of this was effectively copied from the existing code but its probably worth improving here (and also make sure we always do fixed-point arithmetic in our other project)

dlukegordon: Alloy also has convenient parsing of strings into U256:

```
use alloy::utils::parse_units;
use alloy::primitives::U256;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    // Parse "1.04" ETH to wei (18 decimals)
    // The second argument can be a string "eth" or the number of decimals 18.
    let amount_wei = parse_units("1.04", "ether")?;
    
    // The result is a Unit struct, so we convert it to U256
    let value: U256 = amount_wei.into();

    println!("Wei value: {}", value); 
    // Output: 1040000000000000000
    
    Ok(())
}
```

PR #25389, opened 2026-02-10
[move][move-vm][bella-ciao] Remove all `unreachable` and `panic` calls, or otherwise allow them.

cgswords:
## Description 

See title

## Test plan 

All tests still pass

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
@@ -2004,7 +2004,20 @@ impl VectorRef {
     pub fn as_bytes_ref(&self) -> PartialVMResult<std::cell::Ref<'_, Vec<u8>>> {
         let borrowed = self.0.borrow();
         match &*borrowed {
-            Value::PrimVec(PrimVec::VecU8(_)) => (),
+            Value::PrimVec(PrimVec::VecU8(_)) => {
+                #[allow(clippy::unreachable)]
+                // We just checked the type, so this match should always succeed. If it fails, who knows:
+                // we should just panic full stop.
+                Ok(std::cell::Ref::map(borrowed, |value| {
+                    match value {
+                        Value::PrimVec(PrimVec::VecU8(vec)) => vec,
+                        // We already checked, but provide a safe fallback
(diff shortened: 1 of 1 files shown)


external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -3792,12 +3838,7 @@ impl Reference {
                                 (L::Address, PrimVec::VecAddress(xs)) => {
                                     Some(AV::Vector(xs.iter().map(|a| AV::Address(*a)).collect()))
                                 }
-                                (ty, vec) => {
-                                    panic!(
-                                        "Mismatched type {:?} for primitive vector {:?}",
-                                        ty, vec
-                                    )
-                                }
+                                (_ty, _vec) => None,
```
cgswords, line 3842: It was unclear what I should do here, but panicking seems worse than this.

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -2021,8 +2018,24 @@ impl VectorRef {
...
+        #[allow(clippy::unreachable)]
+        // We just checked the type, so this match should always succeed. If it fails, who knows:
+        // we should just panic full stop.
+        Ok(std::cell::Ref::map(borrowed, |value| {
+            match value {
+                Value::PrimVec(PrimVec::VecU8(vec)) => vec,
+                // We already checked, but provide a safe fallback
+                _ => unreachable!("INTERNAL PANIC: Value changed between matching and borrowing."),
+            }
+        }))
```
tzakian: Would it be better to move this into the `PrimVec::VecU8` match arm (if that is possible?). 

Seems a bit more direct/less error prone if a new value were to be introduced in the future.

cgswords: Done!

PR #24797, opened 2026-01-05
rpc-index: include address balances in the balance index

bmwill:
## Description 

Adapt rpc-indexes to include address balances in the balance index both on initialization as well normal checkpoint indexing. 

After this PR the balance index will include both Coin<T> as well as address balance T values summed together, effectively being an address's total balance for that coin_type T. In the RPC layer, when we want to break out that total into a coin and address balance balance, we can then just read the current value of that address balance and then derive the coin balance from the total - the address balance amount.

## Test plan 

How did you test the new or updated feature?

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #25392, opened 2026-02-10
rpc-index: deprecate and drop transaction info column

bmwill:
## Description 

Deprecates and drops the transaction info column in the rpc_indexes. This also fixes a bug that was introduced in #24797 that could lead to the balance index being incorrect if a fullnode restored indexes with the 1.64 release.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [x] gRPC: Fixes a bug that was introduced in #24797 that could lead to the balance index being incorrect if a fullnode restored indexes with the 1.64 release.
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


bmwill:
bypassing due to flaky test.

PR #25398, opened 2026-02-10
Add per-commit binary logging for congestion control

aschran:
## Description 

- Includes per-object and per-tx cost details.
- Saves logged proto files to disk.
- Default size limit 10x 100MB files.

## Test plan 

Unit tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority/shared_object_congestion_tracker.rs
+++ b/crates/sui-core/src/authority/shared_object_congestion_tracker.rs
@@ -219,14 +219,10 @@ impl SharedObjectCongestionTracker {
         let start_cost = self.compute_tx_start_at_cost(&shared_input_objects);
         let end_cost = start_cost.saturating_add(tx_cost);
 
-        let mut modified_objects = Vec::new();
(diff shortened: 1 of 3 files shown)


crates/sui-core/src/authority/shared_object_congestion_tracker.rs
```diff
@@ -215,29 +219,55 @@ impl SharedObjectCongestionTracker {
...
+                }
             }
         }
+
+        if let Some(entries) = &mut self.log_entries {
+            entries.push(TransactionCostLogEntry {
+                tx_digest: *cert.digest(),
+                start_cost,
+                end_cost,
+                modified_objects,
```
mystenmark: we probably don't need to log this since we can reconstruct it from effects during processing

PR #25405, opened 2026-02-10
cli(ptb): pure type inference looks through references

amnn:
## Description

Add support to the CLI to look "through" reference types when checking whether an argument is a pure type.

## Test plan

The following invocation would have failed before, and now it succeeds:

```
$ cargo run --bin sui -- client ptb \
    --move-call 0x3::sui_system::validator_address_by_pool_id @0x5 @0x0cf0a8d13f56cb4c675a74f9f8e646dc7c3e00c5e50da0d6c076b291f57f617f \
    --dev-inspect
```

It failed previously because the argument was an `&ID`, but now the CLI can understand that and still serialize it as an `address`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [x] CLI: Fix an issue where the CLI would fail to infer the type of a primitive argument to a Move call if that argument was accessed by reference or by mutable reference.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-json/src/lib.rs
+++ b/crates/sui-json/src/lib.rs
@@ -575,9 +575,11 @@ pub fn primitive_type(
         SignatureToken::U128 => MoveTypeLayout::U128,
         SignatureToken::U256 => MoveTypeLayout::U256,
         SignatureToken::Address => MoveTypeLayout::Address,
+
         SignatureToken::Vector(inner) => {
             MoveTypeLayout::Vector(Box::new(primitive_type(view, type_args, inner)?))
         }
+
         SignatureToken::Datatype(struct_handle_idx) => {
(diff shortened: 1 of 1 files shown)


stefan-mysten:
@amnn is this working now because `sui client ptb` uses `sui-json` under the hood? I was expecting that everything is migrated to gRPC, but maybe that still relies on `sui-json`?

amnn:
No I don't think so, unless I'm missing something, `sui-json` usage is limited to some of the transaction building APIs for JSON-RPC, and it should stay that way, it does some pretty whacky stuff, IIRC.

PR #25409, opened 2026-02-11
[move-ide] Added support for module extensions

awelc:
## Description 

This PR implements support for module extensions, following the previous [PR](https://github.com/MystenLabs/sui/pull/25234) that implemented compiler-side support for this and a skeleton of actual move-analyzer support.

The main previously unresolved problem was that some information about extensions does not survive until typing. Additionally, we cannot rely on typing-level module identity and locations the way we used to, as extensions share the same `ModuleIdent` as their "base" module and have separate locations than their base module.

As a result a few things needed to happen:
- we need to collect parsing-level information separately - this happened by splitting `ModuleDefs` struct into part that is collected at parsing (`ModuleParsingInfo`) and part that is still collected at typing (still called `ModuleDefs`)
- cursor computation needed to move to parsing to distinguish it being in one of the extensions from it being in the base module

Since in this PR we construct naming-level `ModuleIdent`-s from parsing-time info, we may be able to use them as keys to various maps directly (instead of constructing "module ID strings"). It's better pushed to a separate PR, though as this one is already non-trivial.

## Test plan 

New tests for symbols (use-defs, on-hover, etc.), inlay hints, and auto-imports/completion in presence of extensions have been added. All tests must pass

### DIFF
--- a/external-crates/move/crates/move-analyzer/tests/references.snap
+++ b/external-crates/move/crates/move-analyzer/tests/references.snap
@@ -27,7 +27,7 @@ References:
 -- test 3 -------------------
 use line: 26, use_ndx: 0
 References:
-  'match_mut' at mut_match.move:3:16
   'match_mut' at M1.move:23:31
   'match_mut_fun' at M1.move:23:44
   'match_mut_fun' at M1.move:26:9
+  'match_mut' at mut_match.move:3:16
--- a/external-crates/move/crates/move-analyzer/tests/references/sources/M1.move
+++ b/external-crates/move/crates/move-analyzer/tests/references/sources/M1.move
@@ -10,7 +10,7 @@ module References::M1 {
         MyStruct { value: 42 }
     }
 
-    public fun unoack(s: MyStruct): u64 {
+    public fun unpack(s: MyStruct): u64 {
         let MyStruct { value } = s;
         value
     }
--- a/external-crates/move/crates/move-analyzer/tests/module_ext_refs.ide
+++ b/external-crates/move/crates/move-analyzer/tests/module_ext_refs.ide
@@ -1,0 +1,22 @@
+// Tests references for basic Move constructs
+{
+  "References": {
+    "project": "tests/module-ext",
+    "file_tests": {
+      "M1.move": [
+        {
+          "use_line": 17,
+          "use_ndx": 0
+        },
+        {
+          "use_line": 28,
+          "use_ndx": 0
+        },
+        {
+          "use_line": 39,
+          "use_ndx": 0
+        }
+      ]
+    }
+  }
+}
--- a/external-crates/move/crates/move-analyzer/tests/module_ext_refs.snap
+++ b/external-crates/move/crates/move-analyzer/tests/module_ext_refs.snap
@@ -1,0 +1,45 @@
+---
+source: crates/move-analyzer/tests/ide_testsuite.rs
+---
+== M1.move ========================================================
+-- test 0 -------------------
+use line: 17, use_ndx: 0
+References:
+  'int_match' at int_match.move:3:16
+  'int_match' at M1.move:14:31
+  'int_match_function' at M1.move:14:44
+  'int_match_function' at M1.move:17:9
+  'int_match' at M1.move:25:31
+  'int_match_function' at M1.move:25:44
+  'int_match_function' at M1.move:28:9
+  'int_match' at M1.move:36:31
+  'int_match_function' at M1.move:36:44
+  'int_match_function' at M1.move:39:9
+
+-- test 1 -------------------
+use line: 28, use_ndx: 0
+References:
+  'int_match' at int_match.move:3:16
+  'int_match' at M1.move:14:31
+  'int_match_function' at M1.move:14:44
+  'int_match_function' at M1.move:17:9
+  'int_match' at M1.move:25:31
+  'int_match_function' at M1.move:25:44
+  'int_match_function' at M1.move:28:9
+  'int_match' at M1.move:36:31
+  'int_match_function' at M1.move:36:44
+  'int_match_function' at M1.move:39:9
+
+-- test 2 -------------------
+use line: 39, use_ndx: 0
+References:
+  'int_match' at int_match.move:3:16
+  'int_match' at M1.move:14:31
+  'int_match_function' at M1.move:14:44
+  'int_match_function' at M1.move:17:9
+  'int_match' at M1.move:25:31
+  'int_match_function' at M1.move:25:44
+  'int_match_function' at M1.move:28:9
+  'int_match' at M1.move:36:31
+  'int_match_function' at M1.move:36:44
+  'int_match_function' at M1.move:39:9


external-crates/move/crates/move-analyzer/src/analysis/mod.rs
```diff
@@ -31,31 +31,50 @@ pub mod typing_analysis;
 
 pub type DefMap = BTreeMap<Loc, DefInfo>;
 
+/// Context tracking the current location during analysis traversal.
+/// All fields are set together when entering a module and the entire
+/// context is cleared when exiting.
+#[derive(Debug, Clone)]
+pub struct CurrentLocationContext {
```
awelc, line 38: We need to maintain a few pieces of info about where we are - hence this new struct to track this

external-crates/move/crates/move-analyzer/src/analysis/mod.rs
```diff
@@ -104,17 +124,12 @@ fn process_typed_modules<'a>(
...
-    // fallback removed.
-    static EMPTY_ALIAS_LENGTHS: std::sync::LazyLock<BTreeMap<Position, usize>> =
-        std::sync::LazyLock::new(BTreeMap::new);
     for (module_ident, module_def) in typed_modules.key_cloned_iter() {
         let mod_ident_str = expansion_mod_ident_to_map_key(&module_ident.value);
+        // All typed modules must have alias lengths computed during parsing analysis.
         typing_symbolicator.alias_lengths = mod_to_alias_lengths
             .get(&mod_ident_str)
-            .unwrap_or(&EMPTY_ALIAS_LENGTHS);
+            .unwrap_or_else(|| panic!("no alias lengths for module {mod_ident_str}"));
```
awelc, line 132: This is how it worked originally - it's a critical error and it's best to signal early

external-crates/move/crates/move-analyzer/src/analysis/parsing_analysis.rs
```diff
@@ -89,27 +93,71 @@ impl<'a> ParsingAnalysisContext<'a> {
...
-            // TODO(extensions): This location-based lookup fails for extension modules because
-            // extensions have different locations than the merged typed module. Extensions share
-            // the same NamedAddressMap as their base module, so we could use pkg_def.named_address_map
-            // to look up directly from parsing-level NamedAddressMaps instead.
-            let Some(pkg_addresses) = typed_mod_named_address_maps.get(&mod_def.loc) else {
-                eprintln!(
-                    "no typing-level named address maps for module {}",
-                    mod_def.name.value()
-                );
+            // Get the address map directly from the package's own index - works for extensions too
```
awelc, line 99: Can't take `NamedAddressMaps` from typing-level modules anymore as they only have this data for the "base" module and not for its extensions

external-crates/move/crates/move-analyzer/src/analysis/parsing_analysis.rs
```diff
@@ -191,13 +239,9 @@ impl<'a> ParsingAnalysisContext<'a> {
                 earliest_loc
             }
         }
-        // parsing symbolicator is currently only responsible for processing use declarations
```
awelc, line 194: This comment was no longer accurate, even before this PR

external-crates/move/crates/move-analyzer/src/symbols/compilation.rs
```diff
@@ -1050,30 +1058,43 @@ fn merge_user_programs(
         if pkg_modified {
             unmodified_definitions.push(pkg_def);
         } else {
-            // find cached package definition with the same hash
-            // and update its named address map index
-            let pkg_hash = match &pkg_def.def {
+            // Update ALL cached package definitions from the same file. All modules in
```
awelc, line 1067: This was a bug prior to this PR that never manifested itself but now we switched to using `NamedAddressMaps` at parsing level all the time and things started to break

external-crates/move/crates/move-analyzer/src/symbols/mod.rs
```diff
@@ -671,23 +694,22 @@ fn pre_process_typed_modules(
 }
 
 /// Converts parsing AST's `LeadingNameAccess` to expansion AST's `Address` (similarly to
-/// expansion::translate::top_level_address but disregarding the name portion of `Address` as we
-/// only care about actual address here if it's available). We need this to be able to reliably
-/// compare parsing AST's module identifier with expansion/typing AST's module identifier, even in
```
awelc, line 676: This was no longer accurate (even prior to this PR)

external-crates/move/crates/move-analyzer/src/symbols/mod_extensions.rs
```diff
@@ -206,23 +206,19 @@ fn extract_module_info(
...
+                named_address_map,
+                file_path,
+                module_defs,
+                extension_targets,
+            );
         }
         P::Definition::Address(adef) => {
             for mdef in &adef.modules {
-                extract_module_info(
-                    &P::Definition::Module(mdef.clone()),
```
awelc, line 225: Helper function added to get rid of this `clone`

external-crates/move/crates/move-analyzer/src/symbols/mod.rs
```diff
@@ -1066,10 +1086,7 @@ impl Symbols {
     }
 
     pub fn file_hash(&self, path: &Path) -> Option<FileHash> {
-        let Some(mod_defs) = self.file_mods.get(path) else {
-            return None;
-        };
-        Some(mod_defs.first().unwrap().fhash)
+        self.files.file_hash(&path.to_path_buf())
```
awelc, line 1088: It makes more sene to get the hash from `files` anyways but the old way is now also incorrect

external-crates/move/crates/move-analyzer/tests/references.snap
```diff
@@ -27,7 +27,7 @@ References:
 -- test 3 -------------------
 use line: 26, use_ndx: 0
 References:
-  'match_mut' at mut_match.move:3:16
   'match_mut' at M1.move:23:31
   'match_mut_fun' at M1.move:23:44
   'match_mut_fun' at M1.move:26:9
+  'match_mut' at mut_match.move:3:16
```
awelc, line 33: It's unrelated change - I fixed a typo in the references test's source file and it caused this reorder (likely because file hash and thus location value changed and order in the references map)

external-crates/move/crates/move-analyzer/tests/references/sources/M1.move
```diff
@@ -10,7 +10,7 @@ module References::M1 {
         MyStruct { value: 42 }
     }
 
-    public fun unoack(s: MyStruct): u64 {
+    public fun unpack(s: MyStruct): u64 {
```
awelc, line 13: Unrelated typo fix

external-crates/move/crates/move-analyzer/src/analysis/mod.rs
cgswords, line 38: I'm pretty sure this is (intended to) live on the `Cursor` at `symbols/cursor.rs` --- it in fact already tracks the module it is in. We should extend that as possible here, IMO, instead of adding another new thing.

awelc: This is about where we are during analysis. We used to maintain "current module identifies string" so that it can be used during analysis but we need a few additional pieces. The cursor piece is about preserving where we are so that queries from the editor can be correctly handled. I will look into it some more but perhaps this explains a bit better why we have to track more info using this struct?

cgswords: Oh, that makes sense. Thanks!

PR #25411, opened 2026-02-11
[graphql-alt] transaction contents return balance changes for kv readers too

emmazzz:
## Description 

Balance changes are only returned for executed transactions from grpc in the kv loader while this data is available from other sources too. This PR makes sure balance changes are available for bigtable and ledger grpc reader variants too.

## Test plan 

Existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/balance_change.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/balance_change.rs
@@ -26,7 +26,6 @@ pub(crate) enum BalanceChangeContents {
     Stored(StoredBalanceChange),
 }
 
-/// Effects to the balance (sum of coin values per coin type) of addresses and objects.
 #[derive(Clone)]
 pub(crate) struct BalanceChange {
     pub(crate) scope: Scope,
@@ -34,14 +33,6 @@ pub(crate) struct BalanceChange {
 }
 
 impl BalanceChange {
-    /// Create a BalanceChange from a gRPC BalanceChange.
-    pub(crate) fn from_grpc(scope: Scope, grpc: GrpcBalanceChange) -> Self {
-        Self {
-            scope,
-            content: BalanceChangeContents::Grpc(grpc),
-        }
-    }
-
     /// Create a BalanceChange from a stored BalanceChange (database).
     pub(crate) fn from_stored(scope: Scope, stored: StoredBalanceChange) -> Self {
         Self {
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-graphql/src/api/types/balance_change.rs
```diff
@@ -93,4 +109,19 @@ impl BalanceChange {
         grpc.set_amount(amount.to_string());
         grpc
     }
+
+    /// Convert a native BalanceChange to a gRPC BalanceChange.
+    pub(crate) fn native_to_grpc(native: &NativeBalanceChange) -> GrpcBalanceChange {
```
amnn: This function doesn't belong here, (neither does `stored_to_gprc` above) -- it's a transformation from two types that are **not** `BalanceChange` despite the similar name. They point to the need for a wider refactor here:

- `BalanceChange` should no longer be a `SimpleObject`, it should be represented by a `Scope` and `content` which is an enum that can be one of `Stored`, `Native`, or `Grpc`.
- `BalanceChangeContents` should offer APIs to access its `address`, `coin_type` and `amount` that switch on the `enum`.
- We can also implement `From<BalanceChangeContents> for GrpcBalanceChange` to convert to the gRPC representation for `balanceChangeJson`, etc.

crates/sui-indexer-alt-graphql/src/api/types/transaction_effects.rs
```diff
@@ -235,10 +236,18 @@ impl EffectsContents {
...
                 // First try to get balance changes from execution context (content)
-                if let Some(grpc_balance_changes) = content.balance_changes() {
-                    return page.paginate_indices(grpc_balance_changes.len(), |i| {
-                        BalanceChange::from_grpc(self.scope.clone(), &grpc_balance_changes[i])
-                    });
+                if let Some(balance_changes) = content.balance_changes() {
+                    return page.paginate_indices(
+                        balance_changes.length(),
+                        |i| match balance_changes {
+                            BalanceChangeContents::Grpc(changes) => {
```
amnn: This is a sign of a leaky abstraction -- transaction effects should just be able to say "I have `BalanceChangeContents` in my hand, how do I make `BalanceChanges`, using a constructor on `BalanceChanges` (see my earlier comment about changing how `BalanceChanges` is structured to make this work).

emmazzz:
> After this change, the only reason we still have a StoredBalanceChange is to deal with e.g. the set-up we have for devnet, or localnet where we only run postgres, is that right? And that means that we and other RPC providers can stop indexing tx_balance_changes for production networks?

Actually there is a todo in `sui-kv-rpc` (which is the impl of bigtable's `LedgerService`) for supporting balance changes, and bigtable's balance changes are all just empty vectors. So these need to be supported and backfilled before we can get rid of `tx_balance_changes`.

amnn:
Ah okay, in that case, we need to hold off on landing this right? Currently we don't have the ability to control which configured data source GraphQL looks at on a case-by-case basis, so if we have Ledger gRPC or Bigtable configured, it will use that over `tx_balance_changes` (I suspect that Bigtable might also have a problem at the moment, because we haven't completed the backfill, right?).

PR #25422, opened 2026-02-11
[move-regex-borrow-graph] Improve perf with a custom graph

tnowacki:
## Description 

- Based on some initial results, this should be a good bit faster than using petgraph, though not as fast as the current borrow checker
- At the very least, this will make it easier to change/improve perf overtime, since we now have a layer of abstraction around the underlying graph (we could even switch it to petgraph without changing the "reference graph" implementation) 

## Test plan 

- Added tests
- Analyzed packages 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
+++ b/external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
@@ -138,10 +138,7 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
     }
 
     /// Returns the direct predecessors of the specified reference
-    fn predecessors(
-        &self,
-        r: Ref,
-    ) -> Result<impl Iterator<Item = (Ref, &Edge<Loc, Lbl>)> + '_> {
+    fn predecessors(&self, r: Ref) -> Result<impl Iterator<Item = (Ref, &Edge<Loc, Lbl>)> + '_> {
         let r_idx = self.node(&r)?.node_index();
         ensure!(
             self.graph.contains_node(r_idx),


external-crates/move/crates/bytecode-verifier-tests/src/unit_tests/reference_safety_tests.rs
```diff
@@ -360,7 +360,7 @@ fn test_merge_state() {
     );
     assert_eq!(
         result.unwrap_err().major_status(),
-        StatusCode::PROGRAM_TOO_COMPLEX
+        StatusCode::CONSTRAINT_NOT_SATISFIED,
```
cgswords, line 363: Is this a fork risk, or is it rewritten in the adapter?

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
```diff
@@ -146,8 +155,23 @@ impl AbstractState {
...
+            .0
+            .iter()
+            .chain(&function_context.locals().0)
+            .filter(|ty| {
+                matches!(
+                    ty,
+                    SignatureToken::Reference(_) | SignatureToken::MutableReference(_)
+                )
+            })
+            .count()
```
cgswords, line 170: Small nit: this re-processes `parameters`, when we have `param_refs` above we could get the size of directly.

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
```diff
@@ -607,7 +621,7 @@ impl AbstractState {
         let mut old_to_new = BTreeMap::new();
         old_to_new.insert(self.local_root, 0);
         for (&local, old_r) in &self.locals {
-            let new_r = (local as usize) + 1;
+            let new_r = safe_unwrap!((local as u32).checked_add(1));
```
cgswords, line 625: 🎉

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
```diff
@@ -646,6 +660,7 @@ impl AbstractState {
     pub(crate) fn check_invariants(&self) {
         #[cfg(debug_assertions)]
         {
+            let dummy_meter = &mut move_regex_borrow_graph::meter::DummyMeter;
```
cgswords, line 667: This is interesting, and may be worth calling out in a comment.

```suggestion
            // We use a dummy meter because these only happen in debug mode, and we
            // do not want to impact gas charging behavior to check them.
            let dummy_meter = &mut move_regex_borrow_graph::meter::DummyMeter;
```

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
```diff
@@ -722,42 +740,59 @@ impl AbstractDomain for AbstractState {
...
+    }
 }
 
-fn borrowed_by_size(paths_map: &BTreeMap<Ref, Paths>) -> usize {
-    paths_map
-        .iter()
-        .flat_map(|(_, paths)| paths)
-        .fold(0, |acc, path| acc.saturating_add(path.abstract_size()))
-}
+// fn charge_graph_size(size: usize, meter: &mut (impl Meter + ?Sized)) -> PartialVMResult<()> {
```
cgswords: Do we need this code?

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /// Clear the graph of all nodes and edges.
+    /// NOTE: Do not keep any `NodeIndex` values from before this call. They will be invalid and
+    /// may panic if used (at the very least they will give the wrong nodes/edges)
+    pub fn clear(&mut self) {
+        self.next = 0;
+        self.node_weights.clear();
+        self.edge_weights.clear();
+    }
```
cgswords: If we wanted to enforce this, we could record graph "generations" in the upper bits of the node ID and bump them on each clear, refusing to resolve any that do not match our generation:
```
pub fn clear(&mut self) {
    self.next = 0;
    self.generation += 1;
    ...
}

fn generation(&self) -> Index {
  self.generation << GENERATION_SHIFT
}

fn next(&mut self) -> NodeIndex {
   let next_id = next;
   self.next += 1;
   NodeIndex(self.generation() | next_id)
}

fn to_index(&self, ndx: NodeIndex) -> Option<Index> {
    let ndx = ndx.0;
    if (ndx & GENERATION_MASK) == self.generation()) {
        Some(ndx & !GENERATION_MASK)
    } else {
        None
    }
}
```

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Returns the number of nodes in the graph.
+    pub fn node_count(&self) -> usize {
+        self.node_weights.len()
+    }
+
+    /// Adds a node (with the given weight) to the graph and returns its index.
+    pub fn add_node(&mut self, weight: N) -> NodeIndex {
+        let index = NodeIndex(self.next);
+        self.next = self.next.checked_add(1).expect("NodeIndex overflow");
+        self.node_weights.insert(index, weight);
```
cgswords: Should we ensure this was empty before?

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Clear the graph of all nodes and edges.
+    /// NOTE: Do not keep any `NodeIndex` values from before this call. They will be invalid and
+    /// may panic if used (at the very least they will give the wrong nodes/edges)
+    pub fn clear(&mut self) {
+        self.next = 0;
+        self.node_weights.clear();
+        self.edge_weights.clear();
+    }
+
+    /// Recalculate the `next` field basedo on the current nodes. This should be called after
```
cgswords:
```suggestion
    /// Recalculate the `next` field based on the current nodes. This should be called after
```

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// edge does not exist.
+    pub fn edge_weight_mut(&mut self, from: NodeIndex, to: NodeIndex) -> Option<&mut E> {
+        self.edge_weights.get_mut(&(from, to))
+    }
+
+    /// Removes the specified node and all edges to/from it. Searching for the edges to remove
+    /// is O(E)
+    /// Panics if the node does not exist.
+    pub fn remove_node(&mut self, index: NodeIndex) {
+        let node = self.node_weights.swap_remove(&index);
```
cgswords: `swap_remove` here means iteration ordering is unstable, right? Since we cannot iterate nodes, this does not matter, but I wonder if the index map is doing anything for us cf. a less-stable map that might be faster.

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
```diff
@@ -0,0 +1,167 @@
...
+/// In the context of the borrow graph, the node weights will be the `Ref`, and the edge weights
+/// will be an `Edge<Loc, Lbl>`.
+pub struct GraphMap<N, E> {
+    next: u32,
+    node_weights: IndexMap<NodeIndex, N>,
+    edge_weights: IndexMap<(NodeIndex, NodeIndex), E>,
+}
+
+impl<N, E> GraphMap<N, E> {
+    /// Creates a  new graph with a given capacity for the nodes. This number is assumed to be
```
cgswords:
```suggestion
    /// Creates a new graph with a given capacity for the nodes. This number is assumed to be
```

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Panics if the node does not exist.
+    pub fn remove_node(&mut self, index: NodeIndex) {
+        let node = self.node_weights.swap_remove(&index);
+        assert!(node.is_some(), "Node {:?} does not exist", index);
+        self.edge_weights
+            .retain(|(p, s), _| *p != index && *s != index);
+    }
+
+    /// Returns an iterator over the outgoing edges from the specified node
+    pub fn outgoing_edges(&self, index: NodeIndex) -> impl Iterator<Item = (&E, NodeIndex)> + '_ {
```
cgswords: These linear scans could be a bit slow. Could we store this as `IndexMap<From,IndexMap<To,E>>`?

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Adds a node (with the given weight) to the graph and returns its index.
+    pub fn add_node(&mut self, weight: N) -> NodeIndex {
+        let index = NodeIndex(self.next);
+        self.next = self.next.checked_add(1).expect("NodeIndex overflow");
+        self.node_weights.insert(index, weight);
+        index
+    }
+
+    /// Adds an edge (with the given weight) to the graph. The nodes must already exist.
+    pub fn add_edge(&mut self, from: NodeIndex, weight: E, to: NodeIndex) {
```
cgswords: Should we check `from` and `to` are defined nodes?

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -256,127 +321,148 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
         Ok(new_refs)
     }
 
-    fn extend_by_extension(
+    /// For each source, and for each other node x (including the case where x == source),
+    /// consider all edges source --> x and x --> source,
+    /// for each new reference, determine the edge new --> x and x --> new based on the extension
+    /// provided
+    fn determine_all_new_edges<M: Meter>(
         &mut self,
```
cgswords: Does this need to be `&mut`?

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -420,71 +526,92 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
...
-            }
-            let self_edge_mut = self.graph.edge_weight_mut(p, s).unwrap();
-            let edge_size_increase = self_edge_mut.join(other_edge);
-            let node_mut = self.node_mut(&p)?;
-            node_mut.abstract_size = node_mut.abstract_size.saturating_add(edge_size_increase);
-            size_increase = size_increase.saturating_add(edge_size_increase);
+            let p_self_idx = self.node(&p)?.node_index();
+            let s_self_idx = self.node(&s)?.node_index();
+            let self_edge_size_increase =
+                if let Some(self_edge) = self.graph.edge_weight_mut(p_self_idx, s_self_idx) {
```
cgswords: Maybe an `entry` API?

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
```diff
@@ -0,0 +1,167 @@
...
+
+impl<N, E> GraphMap<N, E> {
+    /// Creates a  new graph with a given capacity for the nodes. This number is assumed to be
+    /// the maximum number of canonical references at the end of a block
+    pub fn new(canonical_reference_capacity: usize) -> Self {
+        debug_assert!(canonical_reference_capacity < 512);
+        Self {
+            next: 0,
+            node_weights: IndexMap::with_capacity(canonical_reference_capacity),
+            edge_weights: IndexMap::with_capacity(canonical_reference_capacity * 3 / 2),
```
cgswords, line 41: This number seems low, esp. in the context of all the self-edges we add.

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -420,71 +526,92 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
...
-        for (p, s, other_edge) in other
+        let other_all_edges = other
             .graph
             .all_edges()
-            .filter(|(p, s, _)| self_keys.contains(p) && self_keys.contains(s))
+            .map(|(p_other_idx, e, s_other_idx)| {
+                let p = *other.graph.node_weight(p_other_idx).unwrap();
+                let s = *other.graph.node_weight(s_other_idx).unwrap();
+                (p, e, s)
+            });
```
cgswords: Maybe `graph.all_edge_weights()`? Here and in `check_invariants`.

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -256,127 +321,148 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
...
-        for (&n, node) in self.nodes.iter_mut() {
-            self.graph.remove_edge(r, n);
-            if let Some(e) = self.graph.remove_edge(n, r) {
-                debug_assert_ne!(n, r);
-                node.abstract_size = node.abstract_size.saturating_sub(e.abstract_size());
-                self.abstract_size = self.abstract_size.saturating_sub(e.abstract_size());
-            }
-        }
-        self.graph.remove_node(r);
+        ensure!(
```
cgswords: This logic is a bit odd; maybe `remove_node` could indicate if it removed successfully instead of panicking, and that could be used here? Not a huge deal, just thinking out loud.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/memory_safety.rs
```diff
@@ -85,10 +85,22 @@ impl Context {
...
+        let (mut graph, _locals) =
+            Graph::new::<()>(canonical_reference_capacity, []).map_err(graph_err)?;
         let local_root = graph
-            .extend_by_epsilon((), std::iter::empty(), /* is_mut */ true)
-            .map_err(graph_err)?;
+            .extend_by_epsilon(
+                (),
+                std::iter::empty(),
+                /* is_mut */ true,
+                &mut DummyMeter,
```
cgswords, line 101: I wonder if it would be better to shove a meter in the context and use it, putting a `DummyMeter` there for now?

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -71,169 +75,230 @@ impl<Loc, Lbl> Path<Loc, Lbl> {
...
-    }
-
     /// Returns the direct successors of the specified reference
     pub(crate) fn successors(
         &self,
         r: Ref,
     ) -> Result<impl Iterator<Item = (&Edge<Loc, Lbl>, Ref)> + '_> {
+        let r_idx = self.node(&r)?.node_index();
+        Ok(self.successors_idx(r_idx)?.map(move |(e, s_idx)| {
+            let s = *self.graph.node_weight(s_idx).unwrap();
```
cgswords: I appreciate this _should_ be there, given where these values are coming from, but the `unwrap` still seems odd.

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
```diff
@@ -646,6 +660,7 @@ impl AbstractState {
     pub(crate) fn check_invariants(&self) {
```
cgswords, line 662: This likely wants to be marked `#[inline(always)]` since in most builds it will be completely empty and we do not want to push an empty frame in those cases. (LLVM _should_ optimize it away, but who knows.)

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -256,127 +321,148 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
...
-            self.determine_new_edges(&mut edges_to_add, x, &ext, new_ref, exclude)?;
-        }
-        for (p, r, s) in edges_to_add {
-            debug_assert!(p == new_ref || s == new_ref);
-            self.add_edge(p, loc, r, s)?;
+        let mut nodes_visited = 0usize;
+        let mut total_edge_size = 0usize;
+        let mut edge_to_add = |p: NodeIndex, r: Regex<Lbl>, s: NodeIndex| {
+            total_edge_size = total_edge_size.saturating_add(r.abstract_size());
+            acc.entry((p, s)).or_default().push(r);
```
cgswords, line 361: Do we ever care to set-ify these?

external-crates/move/crates/bytecode-verifier-tests/src/unit_tests/reference_safety_tests.rs
tnowacki: Its a swap of where the meter ends. There is an oddity in the meter on the external-crates side that issues this error since the folks that wrote the meter did not know what the status code meant 🙃

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
tnowacki: This one also includes the locals in addition to the parameters.

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/abstract_state.rs
tnowacki: Whoops, had it there for reference. Will remove

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
tnowacki: I kind of like it... but I would need to bump it up to a u64. Let me give it a go

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
tnowacki: Correct it is unstable. I can run the numbers again, but if I recall correctly the BTreeMap was a tad slower. Let me re-run the numbers with it and check. A good call out

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
tnowacki: I tried it and was slower overall (on average over main net packages)

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
tnowacki: It's very very crudely PGO-d. Running experiments this seemed to perform slightly better than `* 2`

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
tnowacki: The name is a bit funky, but sure!

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
tnowacki: I will take a look

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/memory_safety.rs
tnowacki: Let me do this in a separate PR because it would be a breaking change and needs to be gated

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
tnowacki: It gets set-ified downstream so I figured it would be better to do that once rather than many times, but I can test it each way!

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
tnowacki: This doesn't quite work because I think you really want to bump the generation after calling minimize. It at lease reduces the risk of holding onto an index so I like it

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -114,14 +114,17 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
     pub(crate) fn successors(
         &self,
         r: Ref,
-    ) -> Result<impl Iterator<Item = (&Edge<Loc, Lbl>, Ref)> + '_> {
+    ) -> Result<impl Iterator<Item = Result<(&Edge<Loc, Lbl>, Ref)>> + '_> {
```
cgswords: Gnarly!

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
```diff
@@ -8,14 +8,18 @@ use indexmap::IndexMap;
 pub struct Error;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct NodeIndex(u32);
+pub struct NodeIndex {
```
cgswords, line 13: Might want to `#[repr(transparent)]` to ensure Rust does not do something weird with this.

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
cgswords: I would have guessed much higher?

tnowacki: Most references aren't extended, just passed around

external-crates/move/crates/move-regex-borrow-graph/src/graph_map.rs
tnowacki: I have a test that asserts the size of this is the same as the size as a u64. It yells at me with transparent needing a single field. And repr(u64) being for enums

tnowacki: And the documentation on repr packed scared me

PR #25427, opened 2026-02-12
[move][move-vm][bella-ciao] Check ref pointers at calls

cgswords:
## Description 

Harden callsites a tad

## Test plan 

Tests are unchanged

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
@@ -1211,7 +1211,10 @@ pub(crate) fn check_reference_args_unique(
         // also the same, then we have two references (at least one of which is mutable) that point
         // to the same location, which is a violation.
         let i = all_ref_ptrs.partition_point(|p| p < mptr);
-        if i + 1 < ref_ptrs_len && all_ref_ptrs.safe_get(i)? == all_ref_ptrs.safe_get(i + 1)? {
+        let i_plus_1 = 1.safe_add(i)?;
+        if i_plus_1 < ref_ptrs_len
+            && all_ref_ptrs.safe_get(i)? == all_ref_ptrs.safe_get(i_plus_1)?
+        {
             return Err(partial_vm_error!(
                 UNKNOWN_INVARIANT_VIOLATION_ERROR,
                 "mutable reference argument aliases another reference in function call"


external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
```diff
@@ -1128,9 +1127,52 @@ fn push_call_frame(
...
+    for (arg, param_ty) in args.iter().zip(parameters.iter()) {
+        if let Value::Reference(reference) = arg {
+            let ptr = reference.ref_ptr();
+            all_ref_ptrs.push(ptr);
+            if matches!(param_ty, ArenaType::MutableReference(_)) {
+                mut_ref_ptrs.push(ptr);
+            }
+        }
+    }
+    all_ref_ptrs.sort_unstable();
```
tzakian, line 1205: Nit: optimization -- if `mut_ref_ptrs` is empty you can early return here and skip needing to sort `all_ref_ptrs`

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
```diff
@@ -1128,9 +1127,52 @@ fn push_call_frame(
...
+    for mptr in &mut_ref_ptrs {
+        let first = all_ref_ptrs.partition_point(|p| p < mptr);
```
tnowacki: Not sure if it is wroth commenting that `partition_point` is implemented via a binary search. It helps explain the `sort` above, and avoids maybe having to go look up `partition_point` to see if perf here needs to e improved

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -546,6 +546,13 @@ impl Reference {
             }
         }
     }
+
+    pub(crate) fn ref_ptr(&self) -> usize {
```
tnowacki, line 551: A small comment about this being used for debugging/invariants?

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
```diff
@@ -1128,9 +1127,52 @@ fn push_call_frame(
...
+        let last = all_ref_ptrs.partition_point(|p| p <= mptr);
+        if last - first > 1 {
```
tnowacki: `abs_diff` here maybe? Or an assertion that `last >= first`. Otherwise some strange shit could happen if the sort above was broken

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
tzakian: Doing some more pondering last night -- would it be faster to to find the lowest partition point, and then simply check if `lowest_partition_point + 1 < all_ref_ptrs.len() &&  all_ref_ptrs[lowest_partition_point] == all_ref_ptrs[lowest_partition_point + 1]` to raise an error?

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
```diff
@@ -1128,9 +1127,52 @@ fn push_call_frame(
...
+    if args.len() != parameters.len() {
+        return Err(partial_vm_error!(
+            UNKNOWN_INVARIANT_VIOLATION_ERROR,
+            "argument count ({}) does not match parameter count ({})",
+            args.len(),
+            parameters.len()
+        ));
+    }
+    let mut all_ref_ptrs = SmallVec::<[usize; 256]>::new();
+    let mut mut_ref_ptrs = SmallVec::<[usize; 256]>::new();
```
dariorussi: should those be a much much smaller numbers than 256?

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
tnowacki: Marginally faster, but yeah I like it:
- Find the _first_ occurrence of `mptr`, using partition_point is just faster (and prevents O(n^2) worse case, just O(n lg n) with sorting)
- Ensure that the next ref is not ==.

Could also be `all_ref_ptrs.get(lowest_partition_point + 1).is_some_and(|p| all_ref_ptrs[lowest_partition_point] == p)` or something similar

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
cgswords: No, you can have 255 arguments to a function.

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
cgswords: Makes sense to me.

cgswords: Cleaned it up.

PR #25430, opened 2026-02-12
grpc: implement display

bmwill:
## Description 

This is staked on top of https://github.com/MystenLabs/sui/pull/25360

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14544,7 +14544,7 @@ dependencies = [
 [[package]]
 name = "sui-crypto"
 version = "0.2.0"
-source = "git+https://github.com/MystenLabs/sui-rust-sdk.git?rev=dc3b2bf6f998282e61aeebb803d93499aee2a4a2#dc3b2bf6f998282e61aeebb803d93499aee2a4a2"
+source = "git+https://github.com/MystenLabs/sui-rust-sdk.git?rev=7b020c71d12616c9674b34874f526ac3aa6e5e64#7b020c71d12616c9674b34874f526ac3aa6e5e64"
 dependencies = [
  "ark-bn254",
  "ark-ff 0.4.2",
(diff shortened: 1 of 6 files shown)


crates/sui-display/src/v2/mod.rs
```diff
@@ -242,13 +243,12 @@ impl<'s> Display<'s> {
...
         for ((field, name), value) in self.fields.iter().zip(names).zip(values) {
             use indexmap::map::Entry;
-            use serde_json::Value as JSON;
 
             let src = field.key.src;
 
             let n = match name {
-                Ok(JSON::String(n)) => n,
-                Ok(JSON::Null) => return Err(Error::NameEmpty(src.to_owned())),
+                Ok(v) if v.is_string() => v.as_string().unwrap().to_owned(),
```
bmwill, line 250: Maybe have a way to pull the owned string out?

amnn (approved):
Looks good to me -- could you also stash a TODO somewhere about revisiting `RpcVisitor`'s `Writer` trait to perhaps re-align how things are metered, and replace it with the `Value` abstraction (and/or factor it it out into a `Value` + `Meter` abstraction).

bmwill:
> Looks good to me -- could you also stash a TODO somewhere about revisiting `RpcVisitor`'s `Writer` trait to perhaps re-align how things are metered, and replace it with the `Value` abstraction (and/or factor it it out into a `Value` + `Meter` abstraction).

Yep will do, and then i suppose this will just sit until the stack below it lands.

PR #25431, opened 2026-02-12
[hardening] Replace unwrap/expect/assert with proper error handling in accumulator_root

dongowu:
## Description 

 This PR hardens error handling in `sui-types` accumulator root code while preserving backward compatibility.

  #### What changed

  - Added explicit `try_*` APIs that return `SuiResult`:
    - `AccumulatorValue::try_create_for_testing`
    - `try_update_account_balance_for_testing`
    - `try_build_event_merkle_root`
  - Kept legacy APIs as compatibility wrappers (deprecated):
    - `create_for_testing`
    - `update_account_balance_for_testing`
    - `build_event_merkle_root`
  - Updated internal call sites in `sui-core` and `sui-light-client` to use `try_*` APIs.
  - Added/updated tests in `accumulator_root.rs`, including a compatibility test to ensure legacy signatures still work.

  #### Why

  The previous implementation relied on panic-prone patterns (`unwrap`/`assert!`) in core helper paths.
  This PR moves to explicit error propagation for new usage while keeping old signatures available for smooth migration.

  - `cargo test -p sui-types accumulator_root::tests -- --nocapture`
  - `cargo check -p sui-light-client`
  - `cargo check -p sui-core -j 2`
  - `cargo fmt -p sui-types -p sui-core -p sui-light-client -- --check`
  - `cargo clippy -p sui-types -p sui-core -p sui-light-client -- -D warnings`
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/.claude/skills/hardening/skill.md
+++ b/.claude/skills/hardening/skill.md
@@ -1,0 +1,130 @@
+# Hardening Skill
+
+Security hardening skill - Automatically identify and fix unsafe patterns (unwrap/expect/assert) in code to improve reliability and security.
+
+## Use Cases
+
+Use this skill when you need to:
+- Find panic-prone patterns in code (unwrap, expect, assert)
+- Convert unsafe error handling to proper Result types
+- Generate unit tests for new error paths
+- Improve code safety and reliability
+
+## Commands
+
+```bash
+/hardening                    # Scan current project and generate hardening report
+/hardening <file_path>        # Apply hardening to specified file
+/hardening --auto-fix         # Automatically fix all identified issues
+```
+
+## Workflow
+
+### 1. Scanning Phase
+- Use grep to find all `.unwrap()`, `.expect()`, `assert!()` patterns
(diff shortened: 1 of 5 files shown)


mystenmark:
Thanks for the PR.

We don't use asserts for error handling, we use them to document / enforce invariants. If you see an assert, it is implicitly a claim that the condition it guards is always true.

Of course, we can make mistakes - there may be a way to cause these assertions to fail. If you think there is, then the way to approach this is the following:

- Write a test that causes an assertion failure _using only the public API to do so_. 
- Then write the fix.

Then, importantly: DO NOT submit the fix as a PR (since you would be publicly disclosing a security vulnerability), instead go through our bug bounty program, so that we can fix the vulnerability safely, and you can get paid.

PR #25440, opened 2026-02-12
[bella-ciao] pull out some test-only functionality from `LinkageContext`

tzakian:
## Description 

Gates some functionality over `LinkageContext` to just be for testing.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/shared/linkage_context.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/shared/linkage_context.rs
@@ -7,7 +7,7 @@ use move_binary_format::{
     errors::{PartialVMResult, VMResult},
     partial_vm_error,
 };
-use move_core_types::language_storage::{ModuleId, TypeTag};
+use move_core_types::language_storage::TypeTag;
 
 use crate::shared::types::{OriginalId, VersionId};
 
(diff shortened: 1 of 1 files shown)


external-crates/move/crates/move-vm-runtime/src/shared/linkage_context.rs
```diff
@@ -39,7 +39,48 @@ impl LinkageContext {
...
+            .get(module_id.address())
+            .map(|remapped_address| ModuleId::new(*remapped_address, module_id.name().into()))
+            .ok_or_else(|| {
+                partial_vm_error!(LINKER_ERROR, "Did not find {module_id} in linkage table")
+            })
+    }
+
+    /// Gives the root package plus transitive dependencies (as published package IDs) of the
+    /// linking context. This is computed as the values of the linkage table, which must
+    /// necessarily include the root package address.
```
cgswords: This is dubious because it only incidentally contains the root package. A super-linkage would have no such notion, and this comment is misleading.

external-crates/move/crates/move-vm-runtime/src/shared/linkage_context.rs
```diff
@@ -39,7 +39,48 @@ impl LinkageContext {
...
+    pub fn all_packages(&self) -> VMResult<BTreeSet<VersionId>> {
+        Ok(self
+            .linkage_table
+            .values()
+            .cloned()
+            .collect::<BTreeSet<_>>())
+    }
 
+    /// Gives the transitive dependencies (as published package IDs) of the linking context. This
```
cgswords: Same comment about super-linkages. If you want a set of transitive deps, you should grab the package and resolve all of its dependencies against a linkage context, which cannot be done here and must be done on the virtual tables.

PR #25444, opened 2026-02-12
Fix error message that is reported when building if the active environment isn't defined

mdgeorge4153:
## Description 

This PR fixes a number of package management issues:
 - improves the error messages when an environment is not present (no longer suggests adding it to the toml, and instead suggests adding `--build-env`)
 - makes it so that `sui move build --dump` can be run without an active network
 - removes implicit dependencies from several tests
 - fixes absolute paths in ephemeral pubfiles on windows

## Test plan 

Shell tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [X] CLI:
`sui move build --dump` can now be run without an active network connection if tree shaking is disabled.

fixes for ephemeral publication with local dependencies on windows.

improved error message when building against an undefined environment.

- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-package-alt/src/errors/mod.rs
+++ b/external-crates/move/crates/move-package-alt/src/errors/mod.rs
@@ -105,6 +105,11 @@ pub enum PackageError {
     #[error("Unable to create ephemeral publication file `{file}`: {err:?}")]
     InvalidEphemeralFile { file: PathBuf, err: std::io::Error },
 
+    #[error(
+        "Ephemeral publication file refers to a local package `{path}`, but the directory at `{path}` cannot be accessed."
+    )]
+    InvalidEphemeralPath { path: PathBuf },
+
     #[error("Multiple entries with `source = {{ {dep} }}` exist in the publication file")]
     MultipleEphemeralEntries { dep: String },
 
--- a/external-crates/move/crates/move-package-alt/src/package/root_package.rs
+++ b/external-crates/move/crates/move-package-alt/src/package/root_package.rs
@@ -371,9 +371,8 @@ fn localpubs_to_publications<F: MoveFlavor>(
                 .0
                 .local
                 .canonicalize()
-                .map_err(|err| PackageError::InvalidEphemeralFile {
-                    file: source.0.local.clone(),
-                    err,
+                .map_err(|_| PackageError::InvalidEphemeralPath {
+                    path: source.0.local.clone(),
                 })?;
 
         let old = result.insert(source, new);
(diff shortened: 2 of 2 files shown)


crates/sui/tests/shell_tests/env_error/client.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - alias: mainnet
+    rpc: "https://fullnode.mainnet.sui.io:443"
+    ws: ~
+    basic_auth: ~
+    chain_id: 35834a8a
+  - alias: devnet
+    rpc: "https://fullnode.devnet.sui.io:443"
+    ws: ~
+    basic_auth: ~
+    chain_id: 01accae1
```
stefan-mysten, line 20: Is this file auto-generated?

crates/sui/tests/shell_tests/env_error/no_env_build.snap
```diff
@@ -0,0 +1,20 @@
...
+
+echo "== should fail and ask user to provide -e =="
+sui move --client.config client.yaml build
+
+----- results -----
+success: false
+exit_code: 1
+----- stdout -----
+== should fail and ask user to provide -e ==
+Could not determine the correct dependencies to use for `devnet`; pass one of `--build-env testnet` or `--build-env mainnet`.
```
stefan-mysten, line 18: Do you think it would be good idea to explain more?
```suggestion
Could not determine the correct dependencies to use for `devnet`; pass one of `--build-env testnet` or `--build-env mainnet` if you do not have devnet specific dependencies.

If you wish to have devnet specific dependencies, then please declare them as following.
[...] 
```

.gitignore
```diff
@@ -57,6 +57,7 @@ coverage/
 test-results/
 playwright-report/
 playwright/.cache/
+*.snap.new
```
stefan-mysten, line 61: wouldn't this ignore new snapshots?

stefan-mysten (approved):
Stamping to unblock this fix, but curious what you think about explaining more in details around --build-env and envs. See my comments below.

.gitignore
mdgeorge4153: this ignores snapshots that haven't been accepted - when you run the tests it spits out `foo.snap.new` and then when you run `cargo insta review` and accept the change, it moves it to `foo.snap`.

crates/sui/tests/shell_tests/env_error/client.yaml
mdgeorge4153: good catch, I just copied mine from my home directory, I should clean it up a bit

crates/sui/tests/shell_tests/env_error/no_env_build.snap
mdgeorge4153: I'm wary of providing too much detail here because this is a message likely to be seen by a beginner and I don't want to flood them with information. My instinct is to leave it (or tweak it ever so slightly to indicate that these environments are defined by the package), or maybe to just include a link to the docs like I did with the publication message.

PR #25463, opened 2026-02-13
docs: indexer doc improvements

evan-wall-mysten:
## Description 

1. Adds a diagram to show relationship between indexer stack components.
2. Generates CLI `--help` output text files to reference from docs.

## Test plan 

Viewed locally with `pnpm start`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


evan-wall-mysten:
@jessiemongeon1 is there a way to block the image files from the spell checker? It looks like they are causing failures like https://github.com/MystenLabs/sui/actions/runs/21996488273/job/63557784482?pr=25463.

jessiemongeon1:
@evan-wall-mysten we can honestly remove that CI check, it doesn't catch spelling errors 99.9% of the time, and when it does its cases like this where its an image. I'll open a PR to remove it

jessiemongeon1:
@evan-wall-mysten Ok its removed, you probably need to rebase + rerun the job

docs/content/guides/operator/indexer-stack-setup.mdx
```diff
@@ -8,6 +8,49 @@ keywords: [indexer, consistent store, graphql, setup, dex]
 
 See [GraphQL and General-Purpose Indexer](/concepts/data-access/graphql-indexer.mdx) for more information on the stack.
 
+## Components
+
+<table>
+	<tbody>
+		<tr>
+			<td>
+				![Indexer stack components](./images/indexer_stack.svg "A diagram showing the indexer stack components")
```
wlmyng: could we also provide context on/ link out to more details on external components like what the managed snapshots and archival service are?

docs/content/guides/operator/indexer-stack-setup.mdx
```diff
@@ -8,6 +8,49 @@ keywords: [indexer, consistent store, graphql, setup, dex]
...
+							2c. `--database-path`
+						</td>
+					</tr>
+					<tr>
+						<td>GraphQL connection params</td>
+						<td>
+							3a. `--fullnode-rpc-url`<br/>
+							3b. `--database-url`<br/>
+							3c. `--consistent-store-url`<br/>
+							3d. `--ledger-grpc-url`
```
wlmyng: I'm not the biggest fan of doing something like this, because it's quite easy to introduce drift between what the params really are, and when we updated the docs last

docs/content/guides/operator/indexer-stack-setup.mdx
```diff
@@ -446,6 +489,21 @@ AlloyDB Omni recommends 8GB RAM per vCPU [link](https://docs.cloud.google.com/al
...
 
+## Add new pipeline to pre-existing indexer
+
+Adding a new pipeline to an existing indexer currently requires these manual steps:
+
+1. Determine the most recently indexed checkpoint (`recent-checkpoint`) by doing one of the following (an approximate value is ok to use because the new pipeline will catch up):
+    * Query with GraphQL: `curl -d '{"query":"{ checkpoint { sequenceNumber } }"}' '<graphql-url>'`
+    * Use the `indexer_alt_indexer_latest_processed_checkpoint` metric
+2. Start the indexer with the new pipeline specifying
+    * `--first-checkpoint <recent-checkpoint>`
```
wlmyng: we should call out that other pipelines may "stall" until the new pipeline is ingesting a next checkpoint that other existing pipelines expect to see next

docs/content/guides/operator/indexer-stack-setup.mdx
```diff
@@ -446,6 +489,21 @@ AlloyDB Omni recommends 8GB RAM per vCPU [link](https://docs.cloud.google.com/al
...
+
+   Note: `--first-checkpoint` only needs to be specified the first time the indexer runs with the new pipeline in order to bootstrap the watermark record; it will resume from the watermark record on subsequent runs.
+3. If you also want to index checkpoints before `recent-checkpoint`, start a backfill on a separate indexer instance for the new pipeline specifying
```
wlmyng: why dont we always recommend starting a new pipeline on a separate indexer instance, then when it has caught up, redeploy the existing indexer with the pipeline, which now has an existing watermark entry?

docs/content/guides/operator/indexer-stack-setup.mdx
evan-wall-mysten: @jessiemongeon1 could you help with this?

docs/content/guides/operator/indexer-stack-setup.mdx
evan-wall-mysten: @jessiemongeon1 could you help with the best way to format this information?

docs/content/guides/operator/indexer-stack-setup.mdx
evan-wall-mysten: I think this is ok for now because it will eventually be replaced by being able to index backwards.

Is what you are recommending simpler? I think it is the same level of complexity?

wlmyng: I think it just comes down to whether operators know that other pipelines on the indexer will likely stall while the new pipeline catches up, and whether they are ok with that. Making that explicit is the main goal

.github/workflows/docs-ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Install Nodejs
         uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # pin@v6.2.0
         with:
-          node-version: '24'
+          node-version: '25'
```
jessiemongeon1: this shouldn't change, its breaking the Vercel build.

docs/content/guides/operator/indexer-stack-setup.mdx
jessiemongeon1: @evan-wall-mysten Is there a good place in the code where these are defined? We can import that rather than statically list them here.

docs/content/guides/operator/indexer-stack-setup.mdx
```diff
@@ -8,6 +8,49 @@ keywords: [indexer, consistent store, graphql, setup, dex]
 
 See [GraphQL and General-Purpose Indexer](/concepts/data-access/custom-indexers.mdx) for more information on the stack.
 
+## Components
+
+<table>
+	<tbody>
+		<tr>
+			<td>
+				![Indexer stack components](./images/indexer_stack.svg "A diagram showing the indexer stack components")
```
jessiemongeon1:
```suggestion
				![Indexer stack components](./images/indexer_stack.svg "A diagram showing the indexer stack components")
				
The indexer stack uses the following components:

- [Archival service](/concepts/data-access/archival-store): A service that provides access to historical data beyond what full nodes or indexer databases retain.
- [Consistent store](/concepts/data-access/graphql-rpc#working-with-consistent-store): Combined indexer and RPC service for indexing live data on-chain and serving queries for recent checkpoints.
- [Full node](/guides/operator/sui-full-node): Nodes that track and verify the network's state, and serve read requests from clients.
- [GraphQL](/guides/developer/accessing-data/query-with-graphql): A query language for APIs.
- [Managed snapshots](/guides/operator/snapshots): Point-in-time view of a database's store.
- [Remote store](/concepts/data-access/indexer-data-integration#remote-store): A remote storage location for checkpoint data. 
```

.github/workflows/docs-ci.yml
evan-wall-mysten: Ok I'll just change the version on my machine to 24 then which I noticed is the LTS version anyways.

docs/content/guides/operator/indexer-stack-setup.mdx
evan-wall-mysten: They are defined in code, but not in this format. For example, "1a. `--remote-store-url`" is here: https://github.com/MystenLabs/sui/blob/8b4767a3d3537dd70c69b8f2403ede0e58269674/crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs#L58

The [clap](https://docs.rs/clap/latest/clap/) crate (`#[derive(clap::Args)]` above it) handles converting `remote_store_url` to `--remote-store-url`.

There is a [clap_mangen](https://crates.io/crates/clap_mangen) crate to convert this to a man page which I think we could link to. @amnn do we do anything like this?

docs/content/guides/operator/indexer-stack-setup.mdx
evan-wall-mysten: Updated based on https://github.com/MystenLabs/sui/pull/25463/changes#r2823456237

docs/content/guides/operator/indexer-stack-setup.mdx
evan-wall-mysten: @jessiemongeon1 I generated the output of running the commands with `--help`. 

These files are in this PR  as the `*-help.txt` files. How should I reference the specific lines that contain these configs?

docs/content/guides/operator/indexer-stack-setup.mdx
evan-wall-mysten: Going to switch to what you recommended based on our talk because it prevents the operator from needing to manually modify the new pipeline's watermark record.

crates/sui-indexer-alt-consistent-store/src/lib.rs
amnn, line 1: This should be in its own E2E test (under the `test` directory of the crate), which gives it access to the binaries defined in that crate as environment variables, which prevents the need for a recursive call into `cargo run`. See this [forum post](https://users.rust-lang.org/t/cargo-how-do-i-make-a-integration-test-a-depend-on-binary-b/9821/2) for some details.

crates/sui-indexer-alt-graphql/src/snapshots/sui_indexer_alt_graphql__tests__schema.graphql.snap
amnn, line 1: Spurious?

amnn (commented):
Just a comment about how to generate these `help`

PR #25468, opened 2026-02-13
Add discovery-shared address overrides to consensus

aschran:
- Adds config option to run consensus network on a different address than the on-chain address.
- Shares configured override (if any) with peers via discovery protocol.
- Accepts consensus address overrides from peers and updates local EndpointManager.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-network/src/discovery/mod.rs
+++ b/crates/sui-network/src/discovery/mod.rs
@@ -251,6 +251,81 @@ impl Message for VersionedNodeInfo {
 pub type SignedVersionedNodeInfo = Envelope<VersionedNodeInfo, Ed25519Signature>;
 pub type VerifiedSignedVersionedNodeInfo = VerifiedEnvelope<VersionedNodeInfo, Ed25519Signature>;
 
+/// Verifies the signature and endpoint identity consistency of a
+/// `SignedVersionedNodeInfo`, returning a `VerifiedSignedVersionedNodeInfo`
+/// on success.
+///
+/// Checks:
+/// 1. The signature is valid for the P2P peer_id embedded in the node info.
+/// 2. Any non-P2P endpoint identities (e.g. `Consensus`) match the P2P
+///    peer_id, preventing a peer from advertising endpoints under another
+///    key.
+fn verify_versioned_node_info(
+    peer_info: &SignedVersionedNodeInfo,
+) -> Result<VerifiedSignedVersionedNodeInfo, &'static str> {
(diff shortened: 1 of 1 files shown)


crates/sui-config/src/node.rs
```diff
@@ -973,6 +973,19 @@ pub struct ConsensusConfig {
     pub submit_delay_step_override_millis: Option<u64>,
 
     pub parameters: Option<ConsensusParameters>,
+
+    /// Override for the consensus network listen address.
+    /// When set, Mysticeti binds to this address instead of deriving from the committee.
+    /// Address override is advertised via the discovery protocol.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub listen_address: Option<Multiaddr>,
```
akichidis, line 981: Not sure if the `listen_address` should be advertised to p2p. I mean, if we want to override on what "local" address we want our node to listen to, then having a separate property (the `external_address` you introduced) would be the way to advertise the "public" ip? I know that you did it for convenience and flexibility but it might be a bit more straightforward?

crates/sui-network/src/discovery/mod.rs
```diff
@@ -1031,6 +1072,21 @@ fn update_known_peers_versioned(
             continue;
         }
 
+        // Forward non-P2P addresses from configured peers to EndpointManager.
+        if configured_peers.contains_key(&peer_id)
+            && let VersionedNodeInfo::V2(info_v2) = peer_info.data()
+        {
+            for (endpoint_id, addrs) in &info_v2.addresses {
+                if !matches!(endpoint_id, EndpointId::P2p(_)) && !addrs.is_empty() {
```
akichidis, line 1134: So those will be for now the consensus addresses right? I might have missed this, but how do we authenticate the updated address via this network?

crates/sui-config/src/node.rs
aschran: This config works the same as the identically named parameters in `DiscoveryConfig`. For clarity I think we should keep that behavior the same. If you don't want to advertise the `listen_address` then you can always set the `external_address` back to your on-chain address, though I can't imagine that ever happening for validators who are configured with `AccessType::Trusted`.

crates/sui-network/src/discovery/mod.rs
aschran: This is a great point, will address

PR #25469, opened 2026-02-13
[move-compiler] Fix a crash when public(package) processing involves pre-compiled modules

awelc:
## Description 

In a recent [PR](https://github.com/MystenLabs/sui/pull/25234) we changed how `move-analyzer` uses pre-compiled libraries. In particular, some dependency modules from the same package can be passed to the compiler as pre-compiled libs and some can be fully compiled. 

This exposed a problem fixed in the current PR which is related to updating information related `package(public)` in dependency modules. Prior to the `move-analyzer` change it was never possible to have some modules from the same package in pre-compiled libs and some fully compiled and for a given fully compiled module it was expected to have all its friend modules also fully compiled (or the compiler would crash)

## Test plan 

Added a new `move-analyzer` test that exposed the problem prior to the fix. All tests must pass

### DIFF
--- a/external-crates/move/crates/move-compiler/src/typing/translate.rs
+++ b/external-crates/move/crates/move-compiler/src/typing/translate.rs
@@ -260,11 +260,15 @@ fn modules(
         } else if compilation_env.ide_mode() {
             // if a module is not in the typed modules, it must be in pre-compiled library
             // (info contains both typed and pre-compiled modules)
-            compilation_env
-            .diagnostic_reporter_at_top_level().add_diag(ice!((
-                mident.loc,
-                "Compiler added a friend to module but friend is not in typed modules nor in pre-compiled library (in IDE mode)"
-            )));
+            if !info.modules.contains_key(&mident) {
+                compilation_env
+                    .diagnostic_reporter_at_top_level()
+                    .add_diag(ice!((
+                        mident.loc,
+                        "Compiler added a friend to module but friend is not in typed modules \
+                         nor in pre-compiled library (in IDE mode)"
+                    )));
+            }
             // This can happen if some (dependency) modules from the same package are in typed
             // modules and some are in pre-compiled library. Technically this could lead to
             // incorrect friends list for one of the pre-compiled modules, but in practice


external-crates/move/crates/move-compiler/src/typing/translate.rs
```diff
@@ -250,12 +250,42 @@ fn modules(
...
-        // "current" friends becahse all thew new friends are generated off of
-        // `public(package)` usage, which disallows other friends.
-        mdef.friends = UniqueMap::maybe_from_iter(friends.into_iter())
+        let friends = UniqueMap::maybe_from_iter(friends.into_iter())
             .expect("ICE compiler added duplicate friends to public(package) friend list");
+        if let Some(mdef) = typed_modules.get_mut(&mident) {
+            // point of interest: if we have any new friends, we know there can't be any
+            // "current" friends because all the new friends are generated off of
+            // `public(package)` usage, which disallows other friends.
+            mdef.friends = friends;
```
cgswords, line 259: Enforce this if we believe it to be true. Emit an `ice` in the other case.

external-crates/move/crates/move-compiler/src/typing/translate.rs
```diff
@@ -250,12 +250,42 @@ fn modules(
...
+        } else {
+            // if a module is not in the typed modules, it must be in pre-compiled library
+            // (info contains both typed and pre-compiled modules)
+            assert!(info.modules.contains_key(&mident));
```
cgswords: `ice` here

external-crates/move/crates/move-compiler/src/typing/translate.rs
```diff
@@ -250,12 +250,42 @@ fn modules(
...
+            // This can happen if some (dependency) modules from the same package are in typed
+            // modules and some are in pre-compiled library. Technically this could lead to
+            // incorrect friends list for one of the pre-compiled modules, but in practice
+            // it does not appear to be a problem.
```
cgswords, line 275: This sort of makes sense: if it was precompiled as a downstream dependency, any friends list change is probably un-observable due to it -- not being changed. That said, I think we should gate this branch: ensure we are in `IDE` mode or throw an absolute fit.

external-crates/move/crates/move-compiler/src/typing/translate.rs
awelc: I reverted "regular" (no IDE) use to what it used to be before (panic if this operation cannot be performed, but with more description as opposed to simple `unwrap`)

external-crates/move/crates/move-compiler/src/typing/translate.rs
awelc: The "new" functionality now only triggers in IDE mode (and still panics if a given module cannot be found anywhere), and during "regular" (no-IDE compilation) we throw essentially the same fit as before if we can't update a given module's friends

external-crates/move/crates/move-compiler/src/typing/translate.rs
awelc: Added ICE message (keeping assert as reported is not easily available here to create ICE diagnostic but can massage it further if need be)

external-crates/move/crates/move-compiler/src/typing/translate.rs
cgswords: Instead of panicking, we probably should emit `ice!` . Please make that change.

cgswords (approved):
LGTM, one small request.

external-crates/move/crates/move-compiler/src/typing/translate.rs
awelc: I changed both this `panic` and IDE mode's `assert` into ICE diagnostics

PR #25473, opened 2026-02-15
fix: prevent panic in pattern matching with malformed patterns

### LINKED ISSUE #25460: Compiler panic in match exhaustiveness: `unwrap()` on empty `pats` in `specialize_variant()`
The compiler panics with `unwrap()` on `None` in [`shared/matching.rs:279`](https://github.com/MystenLabs/sui/blob/1a27af4bf7ea29b0058c4d5b19ccc98f963f9b2e/external-crates/move/crates/move-compiler/src/shared/matching.rs#L279) when malformed match patterns produce an empty `pats` list after error recovery.

Root cause: `specialize_variant()` calls `pats.pop_front().unwrap()` without checking for empty `pats`.

Related: #25455.

MRE:
```move
module 0x0::M {
    public enum Inner {
        P(u64),
    }

    public enum Outer {
        X(Inner),
        Y(u64),
    }

    fun f(o: Outer): u64 {
        match (o) {
            Outer::X(Inner::P(v0x0::M)) => v,
            Outer::X Outer::X(Inner::P(v)) => v,
            Outer::X(Inner::Q) => 0,
            Outer::Y(v) => v,
        }
    }
}
```

Reproduce steps:
```
cd external-crates/move
cargo +nightly run --release -p move-compiler --bin move-check -- mre/mre_shared_matching_279.move
```

Output:
```
thread '<unnamed>' panicked at crates/move-compiler/src/shared/matching.rs:279:53:
```

Expected behavior: a parse error diagnostic about the malformed patterns.

Build: release

aviu16:
## Summary
Fixes #25460

This PR prevents a compiler panic when processing malformed match patterns that result in an empty pattern list during error recovery.

## The Problem
The `specialize_variant` function in `matching.rs:279` calls `pats.pop_front().unwrap()` without checking if the pattern list is empty. When error recovery from malformed patterns results in an empty list, this causes a panic with `"called Option::unwrap() on a None value"`.

## The Solution
Replace `.unwrap()` with the `?` operator. This returns `None` if the pattern list is empty, which is the correct behavior—when there are no patterns to specialize, the function should indicate that specialization is not applicable by returning `None`.

## Test Case
Added `invalid_malformed_pattern_panic.move` test case that demonstrates the previously panicking scenario with malformed match patterns.

## Changes
- `external-crates/move/crates/move-compiler/src/shared/matching.rs`: Changed line 279 from `.unwrap()` to `?`
- Added test case for malformed pattern error recovery

### DIFF
--- a/external-crates/move/crates/move-compiler/src/shared/matching.rs
+++ b/external-crates/move/crates/move-compiler/src/shared/matching.rs
@@ -276,7 +276,7 @@ impl PatternArm {
         arg_types: &Vec<&Type>,
     ) -> Option<(Binders, PatternArm)> {
         let mut output = self.clone();
-        let first_pattern = output.pats.pop_front().unwrap();
+        let first_pattern = output.pats.pop_front()?;
         let loc = first_pattern.pat.loc;
         match first_pattern.pat.value {
             TP::Variant(mident, enum_, name, _, fields)
--- a/external-crates/move/crates/move-compiler/tests/move_2024/matching/invalid_malformed_pattern_panic.move
+++ b/external-crates/move/crates/move-compiler/tests/move_2024/matching/invalid_malformed_pattern_panic.move
@@ -1,0 +1,17 @@
+module 0x42::M {
+    public enum Outer {
+        X { y: Inner }
+    }
+
+    public enum Inner {
+        Y(u64),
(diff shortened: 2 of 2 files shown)


aviu16:

PR #25476, opened 2026-02-15
fix: prevent panic when binding dollar-prefixed identifier

### LINKED ISSUE #25456: Compiler panic in naming pass: `debug_assert` fires for `$`-prefixed identifier in `let mut`
The compiler panics with a `debug_assert` failure in [`naming/translate.rs:4028`](https://github.com/MystenLabs/sui/blob/1a27af4bf7ea29b0058c4d5b19ccc98f963f9b2e/external-crates/move/crates/move-compiler/src/naming/translate.rs#L4028) when `let mut $x` (a syntax identifier in bind position) passes through the parser but hits a naming pass assertion.

The parser allows `let mut $<ident>` (dollar-prefixed identifiers) in bind position, but the naming pass assumes this syntax is rejected during parsing and fires `debug_assert!` with `"ICE this should fail during parsing"`.

**Note:** this only reproduces in debug builds.

MRE:
```move
module 0x0::M {
    fun f() {
        let mut $x = 0;
    }
}
```

Reproduce steps:
```
cd external-crates/move
cargo +nightly run -p move-compiler --bin move-check -- mre/mre_naming_translate_4028.move
```

Output:
```
thread 'main' panicked at crates/move-compiler/src/naming/translate.rs:4028:21:
ICE this should fail during parsing
```

Expected behavior: a parse error diagnostic about invalid identifier syntax.

Build: debug

aviu16:
## Summary
Fixes #25456

This PR prevents a debug assertion failure when dollar-prefixed identifiers are used in let bindings.

## The Problem
When code like `let mut $x = 0;` is written, the parser accepts it but the naming pass at `translate.rs:4042` has a `debug_assert!` that assumes this syntax was already rejected during parsing. This causes an ICE (Internal Compiler Error) in debug builds with the message "ICE this should fail during parsing".

## The Solution
Replace the `debug_assert!` with proper error handling for both `Bind` and `Assign` cases. Instead of panicking, the compiler now emits a diagnostic error explaining that syntax identifiers (dollar-prefixed) cannot be bound or assigned—they must be used in value positions.

## Test Case
Added `invalid_dollar_binding.move` test case demonstrating the previously panicking scenario with dollar-prefixed identifier in let binding.

## Changes
- `external-crates/move/crates/move-compiler/src/naming/translate.rs`: Replaced debug_assert with match expression that handles both Bind and Assign cases
- Added test case for dollar-prefixed identifier binding

### DIFF
--- a/external-crates/move/crates/move-compiler/src/naming/translate.rs
+++ b/external-crates/move/crates/move-compiler/src/naming/translate.rs
@@ -4039,15 +4039,18 @@ fn lvalue(
                     context.add_diag(diag!(Declarations::DuplicateItem, primary, secondary));
                 }
                 if v.is_syntax_identifier() {
-                    debug_assert!(
-                        matches!(case, C::Assign),
-                        "ICE this should fail during parsing"
-                    );
-                    let msg = format!(
-                        "Cannot assign to argument for parameter '{}'. \
-                        Arguments must be used in value positions",
-                        v.0
-                    );
+                    let msg = match case {
+                        C::Bind => format!(
+                            "Cannot bind argument for parameter '{}'. \
+                            Arguments must be used in value positions",
+                            v.0
+                        ),
+                        C::Assign => format!(
+                            "Cannot assign to argument for parameter '{}'. \
+                            Arguments must be used in value positions",
+                            v.0
+                        ),
+                    };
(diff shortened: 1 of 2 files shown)


aviu16:

PR #25477, opened 2026-02-16
[Block Streaming] observer client & tests for the observer service

akichidis:
## Description 

* Providing implementation for the Observer client
* Adding end to end test - using the real observer client and observer server
* Small refactorings

## Test plan 

CI

## Next

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


consensus/core/src/network/observer.rs
```diff
@@ -85,34 +93,198 @@ pub(crate) struct FetchCommitsResponse {
...
 /// Tonic-based implementation of ObserverNetworkClient to talk to a peer's observer server.
-#[allow(dead_code)]
 pub(crate) struct TonicObserverClient {
     context: Arc<Context>,
-    _network_keypair: NetworkKeyPair,
+    network_keypair: NetworkKeyPair,
+    channel_pool: Arc<ChannelPool>,
+}
+
+struct ChannelPool {
```
akichidis, line 108: We could reuse the existing `ChannelPool` , or reuse part of the logic. Decided to use a separate now as there were some differences

consensus/core/src/network/observer.rs
```diff
@@ -85,34 +93,198 @@ pub(crate) struct FetchCommitsResponse {
...
 
 #[async_trait]
 impl ObserverNetworkClient for TonicObserverClient {
     async fn stream_blocks(
         &self,
-        _peer: PeerId,
-        _request_stream: super::BlockRequestStream,
-        _timeout: Duration,
+        peer: PeerId,
+        highest_round_per_authority: Vec<u64>,
```
akichidis, line 259: For now I am choosing to go with a simpler approach and not use the commands to start/stop the stream. I'll do that until we have a good working solution and then will do the control flow as an on top optimisation.

consensus/config/src/parameters.rs
```diff
@@ -225,6 +228,40 @@ impl Default for Parameters {
     }
 }
 
+/// Represents a peer observer node with its network key and address.
+#[derive(Clone, Debug, Deserialize, Serialize)]
```
mwtian, line 232: Why does this needs to support ser/de, and why the formatting needs to be human readable (hex)?

consensus/core/src/network/mod.rs
```diff
@@ -46,6 +52,15 @@ pub(crate) enum PeerId {
     Observer(NodeId),
 }
 
+impl Display for PeerId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PeerId::Validator(authority) => write!(f, "[{}]", authority),
+            PeerId::Observer(node_id) => write!(f, "[{:?}]", node_id),
```
mwtian, line 59: What does this look like?

mwtian (approved):

consensus/core/src/network/mod.rs
akichidis: For the observer node it should be the compacted public key. For the authority it should be the index it self (0,1,2 etc)

consensus/config/src/parameters.rs
akichidis: Well I've used hex for convenience when it comes to setting up the node.

PR #25480, opened 2026-02-16
[cli] Fixes pretty printing for tx output

stefan-mysten:
## Description 

Fix pretty printing for transaction results, as well as json output format, to match what used to exist before 1.65.0.

The JSON format has one slight difference, so not sure what's the best approach here:
old:
```
 "confirmedLocalExecution": true
 ```
new:
```
  "timestampMs": "1771265071051",
  "checkpoint": "354"
 ```

## Test plan 

Existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/tests/shell_tests/with_network/publish_with_flags/publish_dry_run.snap
+++ b/crates/sui/tests/shell_tests/with_network/publish_with_flags/publish_dry_run.snap
@@ -18,6 +18,7 @@ success: true
 exit_code: 0
 ----- stdout -----
 Dry run completed, execution status: success
+Dry run completed, execution status: success
 
 ----- stderr -----
 BUILDING test_pkg


crates/sui/src/client_commands.rs
```diff
@@ -2328,6 +2331,197 @@ impl Display for SuiClientCommandResult {
...
+struct NoopModuleCache;
+
+impl GetModule for NoopModuleCache {
+    type Error = ();
+    type Item = CompiledModule;
+
+    fn get_module_by_id(&self, _id: &ModuleId) -> Result<Option<Self::Item>, Self::Error> {
+        Ok(None)
+    }
+}
```
stefan-mysten, line 2347: @bmwill I assume we should have a proper RPC backed module cache for this, but wanted to first check with you.

crates/sui/src/client_commands.rs
```diff
@@ -2328,6 +2331,197 @@ impl Display for SuiClientCommandResult {
...
+            .map(|(event_seq, event)| SuiEvent {
+                id: EventID {
+                    tx_digest: digest,
+                    event_seq: event_seq as u64,
+                },
+                package_id: event.package_id,
+                transaction_module: event.transaction_module.clone(),
+                sender: event.sender,
+                type_: event.type_.clone(),
+                parsed_json: json!({}),
```
bmwill: We can request the json event from the node and provide that if we want to.

crates/sui/src/client_commands.rs
bmwill: Not really, so the only place this  is used is for creating the legacy txn and in that it's used for parsing inputs into something human readable, which is just best effort given how inputs work. So as long as this allows for rendering the rest of the txn then it should be good

crates/sui/src/client_commands.rs
stefan-mysten: Got it, I did add it now.

bmwill (approved):

PR #25482, opened 2026-02-16
Fix EventServiceClient connection hang in authenticated_events tests

mystenmark:
## Summary

- Fix rare test hang in `list_authenticated_events_start_beyond_highest` simtest
- Add `connect_event_service_client()` helper with 1s timeout and 10 retry attempts
- Root cause: msim TCP connection race condition can cause `EventServiceClient::connect()` to hang indefinitely

## Details

gRPC connection establishment can hang indefinitely if the remote peer is unable to complete connection establishment. In the simulator, this manifests as a race condition where TCP connect signals can be lost when the server's accept loop yields to handle another event.

Rather than fixing the simulator, this fix makes the client code resilient to transient connection failures - which is the right behavior for real networks too.

## Test plan

- [x] Verified with seed-search.py: 2000+ seeds pass with 0 failures (previously ~1 in 1000 would hang)

### DIFF
--- a/crates/sui-e2e-tests/tests/authenticated_events_api_tests.rs
+++ b/crates/sui-e2e-tests/tests/authenticated_events_api_tests.rs
@@ -4,6 +4,7 @@
 use itertools::Itertools;
 use move_core_types::language_storage::StructTag;
 use std::str::FromStr;
+use std::time::Duration;
 use sui_keys::keystore::AccountKeystore;
 use sui_light_client::authenticated_events::mmr::apply_stream_updates;
 use sui_light_client::proof::base::{Proof, ProofContents, ProofTarget, ProofVerifier};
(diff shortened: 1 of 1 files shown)


crates/sui-e2e-tests/tests/authenticated_events_api_tests.rs
```diff
@@ -169,6 +170,40 @@ async fn emit_large_test_event(
     test_cluster.sign_and_execute_transaction(&tx_data).await
 }
 
+/// Connect to the EventServiceClient with timeout and retry logic.
+///
+/// gRPC connection establishment can hang indefinitely if the remote peer is unable to complete
+/// connection establishment. This helper ensures we always have bounded connection times and
+/// can retry on transient failures.
+async fn connect_event_service_client(
```
alex-mysten, line 178: It seems like all gRPC clients would be equally susceptible to this bug. Should we generalize this solution?

alex-mysten (approved):

crates/sui-e2e-tests/tests/authenticated_events_api_tests.rs
mystenmark: good question - @bmwill is there an obvious place to do this?

bmwill: all other gRPC clients (that I'm aware of) use the lazy connection mechanism with a keep alive interval which should make it resilient to these types of issues.

mystenmark: sounds good, I guess we can always watch out for more test failures

PR #25485, opened 2026-02-17
[antithesis] default stress to HEAD if unset

alex-mysten:
## Description 

When running experimental antithesis tests, stress container may be behind sui-node, which may result in confusing behavior if the local feature branch contains changes to both stress and sui-node. Notably, if the protocol-config version on sui-node is higher than stress, the stress container will exit, resulting in an unsuccessful run.

This change defaults stress to sui-commit if unset, ensuring that stress and sui-node are running from the same SHA for experimental antithesis runs.

## Test plan 

run locally with dry-run, stress commit set correctly to HEAD when unspecified

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/scripts/run-antithesis-tests.py
+++ b/scripts/run-antithesis-tests.py
@@ -129,7 +129,7 @@ def main():
     parser.add_argument(
         "-S", "--stress-commit",
         type=str,
-        help="Sui repo sha for stress image",
+        help="Sui repo sha for stress image (default: same as sui_commit)",
     )
     parser.add_argument(
         "-p", "--protocol-override",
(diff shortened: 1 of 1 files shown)


scripts/run-antithesis-tests.py
```diff
@@ -189,6 +189,8 @@ def main():
     if not sui_commit:
         sui_commit = get_git_output(["rev-parse", "HEAD"])
 
+    stress_commit = args.stress_commit if args.stress_commit else sui_commit
```
mystenmark, line 192: this should be sui_commit_alt if it is set - we usually want the newest version for stress

alex-mysten: alt is newer? `./run_antithesis_tests.py -u -a mainnet` would test an upgrade from mainnet to HEAD, thereby making sui_commit newer. Is that right?

PR #25490, opened 2026-02-17
[bella-ciao] Cleanup resolution interface in the VM

tzakian:
## Description 

Cleans up the resolution interface for the VM so that external resolution requests get mapped to `EXTERNAL_RESOLUTION_REQUEST_ERROR`s.

It also ensures that external requests cannot cause undue interning of strings during the resolution process.

## Test plan 

CI + amended two tests and added another for internment checking. Updated some snapshots on this as well. 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-transactional-test-runner/tests/vm_test_harness/publish_and_call_invalid.snap
+++ b/external-crates/move/crates/move-transactional-test-runner/tests/vm_test_harness/publish_and_call_invalid.snap
@@ -6,10 +6,10 @@ processed 1 task
 task 0, lines 1-10:
 //# publish-and-call --call 0x42::m::a --call 0x42::m::d
 Error: Function execution failed with VMError: {
-    major_status: VTABLE_KEY_LOOKUP_ERROR,
+    major_status: EXTERNAL_RESOLUTION_REQUEST_ERROR,
     sub_status: None,
-    location: undefined,
+    location: 0x42::m,
     indices: [],
     offsets: [],
-    message: "Could not find function 0000000000000000000000000000000000000000000000000000000000000042::m::d",
+    message: "Failed to find function 0000000000000000000000000000000000000000000000000000000000000042::m::d",
 }
--- a/external-crates/move/crates/move-vm-runtime/src/dev_utils/in_memory_test_adapter.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/dev_utils/in_memory_test_adapter.rs
@@ -71,6 +71,10 @@ impl InMemoryTestAdapter {
         self.runtime.vm_config()
     }
 
+    pub fn runtime(&self) -> &MoveRuntime {
+        &self.runtime
+    }
+
     /// Compute all of the transitive dependencies for a `root_package`, including itself.
     pub fn transitive_dependencies(
         &self,
--- a/external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
@@ -239,15 +239,72 @@ impl VMDispatchTables {
         }
     }
 
-    pub fn to_virtual_table_key(
+    /// This returns an `Option` instead of a `Result` because this is used in the function
+    /// resolution logic for external calls, where failure to resolve should be treated as a "not
+    /// found" instead of an error, and the caller will handle the erroring logic for "not found"
+    /// cases.
+    /// NB: It is important that this function return the same error (`None`) in the case where
+    /// either:
+    /// 1. The underlying identifiers were not found in the interner, or
+    /// 2. the underlying package/module/function is not found.
+    pub(crate) fn try_resolve_function_for_external(
+        &self,
+        original_id: &ModuleId,
+        function_name: &IdentStr,
+    ) -> Option<VMPointer<Function>> {
+        let vtable_key = self.try_to_virtual_table_key(
+            original_id.address(),
+            original_id.name(),
+            function_name,
+        )?;
+        let pkg = self.loaded_packages.get(&vtable_key.0.package_key)?;
+        let function_ = pkg.vtable.functions.get(&vtable_key.0.inner_pkg_key)?;
+        Some(function_.ptr_clone())
+    }
+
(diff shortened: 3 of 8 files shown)


external-crates/move/crates/move-transactional-test-runner/tests/vm_test_harness/publish_and_call_invalid.snap
```diff
@@ -6,10 +6,10 @@ processed 1 task
 task 0, lines 1-10:
 //# publish-and-call --call 0x42::m::a --call 0x42::m::d
 Error: Function execution failed with VMError: {
-    major_status: VTABLE_KEY_LOOKUP_ERROR,
+    major_status: EXTERNAL_RESOLUTION_REQUEST_ERROR,
```
tnowacki, line 9: I think we have a status for function not found? Not sure about types? Either way looks fine to me

tnowacki (approved):

external-crates/move/crates/move-vm-runtime/src/dev_utils/in_memory_test_adapter.rs
```diff
@@ -71,6 +71,10 @@ impl InMemoryTestAdapter {
         self.runtime.vm_config()
     }
 
+    pub fn runtime(&self) -> &MoveRuntime {
+        &self.runtime
+    }
```
cgswords, line 76: Can we make this `pub(crate)`? For sanity?

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -239,15 +239,72 @@ impl VMDispatchTables {
...
-    pub fn to_virtual_table_key(
+    /// This returns an `Option` instead of a `Result` because this is used in the function
+    /// resolution logic for external calls, where failure to resolve should be treated as a "not
+    /// found" instead of an error, and the caller will handle the erroring logic for "not found"
+    /// cases.
+    /// NB: It is important that this function return the same error (`None`) in the case where
+    /// either:
+    /// 1. The underlying identifiers were not found in the interner, or
+    /// 2. the underlying package/module/function is not found.
+    pub(crate) fn try_resolve_function_for_external(
```
cgswords, line 250: Nit: this does not appear called anywhere and should not be visible outside of this module.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -239,15 +239,72 @@ impl VMDispatchTables {
...
+
+    /// This returns an `Option` instead of a `Result` because this is used in the type resolution
+    /// logic for external calls into the VM, where failure to resolve should be treated as a "not
+    /// found"/"failure to resolve" instead of an error, and the caller will handle the erroring
+    /// logic for these cases.
+    /// NB: It is important that this function return the same error (`None`) in the case where
+    /// either:
+    /// 1. The underlying identifiers were not found in the interner, or
+    /// 2. the underlying package/module/type is not found.
+    pub(crate) fn try_resolve_type_for_external(
```
cgswords, line 273: Nit: this does not appear called anywhere and should not be visible outside of this module.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -263,7 +320,7 @@ impl VMDispatchTables {
     // Load a type from a TypeTag into a VM type.
     // NB: the type `TypeTag` _must_ be defining ID based. Otherwise, the type resolution will
     // fail.
-    pub fn load_type(&self, type_tag: &TypeTag) -> VMResult<Type> {
+    pub(crate) fn load_type(&self, type_tag: &TypeTag) -> VMResult<Type> {
```
cgswords, line 323: For both this and `find_function`, it might be nice to make some

```
fn make_vtable_lookup_error(msg: &string) -> PartialVmError {
    partial_vm_error!(
        EXTERNAL_RESOLUTION_REQUEST_ERROR,
        msg,
    )
}
```

external-crates/move/crates/move-vm-runtime/src/execution/vm.rs
```diff
@@ -365,15 +358,16 @@ impl<'extensions> MoveVM<'extensions> {
...
-            original_id.name(),
-            function_name,
-        );
         let function = self
             .virtual_tables
-            .resolve_function(&vtable_key)
-            .map_err(|err| err.finish(Location::Undefined))?;
+            .try_resolve_function_for_external(original_id, function_name)
+            .ok_or_else(|| {
+                partial_vm_error!(
```
cgswords, line 365: Also a call to the error function mentioned above

external-crates/move/crates/move-vm-runtime/src/dev_utils/in_memory_test_adapter.rs
tzakian: I actually can remove this on a second look -- didn't see the `runtime()` function on the trait that I can use for what I needed this for (just for testing the interner size)

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
tzakian: This is called from the `vm.rs` module in `find_function`, so I think this can't be private. That being said we can restrict this to `pub(super)` though.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
tzakian: See above :) And will do the same thing here as well.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
tzakian: Played around with this and I think it reads better/makes errors a bit more greppable to keep them inline, but happy to discuss more.

PR #25494, opened 2026-02-17
Add package data to kv store

nickvikeras:
## Description

Including package data in bigtable.

### DIFF
--- a/crates/sui-kvstore/tests/e2e.rs
+++ b/crates/sui-kvstore/tests/e2e.rs
@@ -458,7 +458,8 @@ async fn test_indexer_e2e() -> Result<()> {
     // since execute_transaction_and_wait_for_checkpoint already waited).
     let mut tx_checkpoints = Vec::new();
 
-    // Include the publish transaction
+    // Look up the publish transaction's checkpoint separately so that
+    // tx_checkpoints stays aligned with tx_digests (transfers only).
     let publish_cp_resp = harness
         .grpc_client
         .ledger_client()
@@ -474,7 +475,6 @@ async fn test_indexer_e2e() -> Result<()> {
         .transaction
         .and_then(|t| t.checkpoint)
         .context("publish tx missing checkpoint")?;
-    tx_checkpoints.push(publish_cp);
 
     for digest in &tx_digests {
         let resp = harness
@@ -495,7 +495,7 @@ async fn test_indexer_e2e() -> Result<()> {
         tx_checkpoints.push(cp);
     }
 
-    let max_checkpoint = *tx_checkpoints.iter().max().unwrap();
+    let max_checkpoint = *tx_checkpoints.iter().max().unwrap().max(&publish_cp);
 
     // Wait for all pipelines to catch up via the same path GraphQL uses.
     harness.wait_for_watermark(max_checkpoint, 0).await?;


crates/sui-kvstore/src/tables/packages_by_checkpoint.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Packages-by-checkpoint table: checkpoint-ordered index of packages.
+//!
+//! Row key: `cp_sequence_number (8B) || original_id (32B) || package_version (8B)`
```
bmwill, line 6: how is this table expected to be queried?

crates/sui-kvstore/src/tables/system_packages.rs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! System packages table: index of system packages (those published by address 0x0).
+//!
+//! Row key: `original_id (32B)`
```
bmwill, line 6: what benefit is this over the packages table that has the column for "is_system_package"

nickvikeras: graphql queries for the latest version of all system packages at some watermark. So this allows you to avoid doing a full table-scan of the whole package table to satisfy that query (by doing a full table scan of this tiny table instead).

crates/sui-kvstore/src/tables/packages_by_checkpoint.rs
nickvikeras: paginated reads of "tell me the packages that were created in this checkpoint range"

crates/sui-kvstore/src/handlers/system_packages.rs
```diff
@@ -0,0 +1,57 @@
...
+                let Some(package) = obj.data.try_as_package() else {
+                    continue;
+                };
+
+                let original_id = package.original_package_id().to_vec();
+
+                let entry = tables::make_entry(
+                    tables::system_packages::encode_key(&original_id),
+                    tables::system_packages::encode(cp_sequence_number),
+                    Some(timestamp_ms),
```
amnn, line 43: IIUC, every time a system package is upgraded, this will write a new version to its record, rather than only inserting a new value once it is published for the first time. Does that have any bearing on read time performance?

Also, regarding timestamp -- thoughts on using the `cp_sequence_number` as the timestamp, rather than the actual timestamp? The queries where we scan over all system packages are typically bounded by a checkpoint, so it seems a natural fit.

nickvikeras: >IIUC, every time a system package is upgraded, this will write a new version to its record, rather than only inserting a new value once it is published for the first time. Does that have any bearing on read time performance?

I don't think so as long as you are only reading the newest version in your query? But let me verify that.  

>Also, regarding timestamp -- thoughts on using the cp_sequence_number as the timestamp, rather than the actual timestamp? The queries where we scan over all system packages are typically bounded by a checkpoint, so it seems a natural fit.

I was sort of wondering why didn't just do that for _every_ record in the database. But yeah maybe it makes sense to default to that for new tables?

amnn: I was also wondering why we don't do that for all tables as well (I think I asked that question at the time of building the last indexer as well) -- any reason not to do that now?

PR #25495, opened 2026-02-17
bigtable connection pool

nickvikeras:
## Description

Supporting connection pools for high throughput bigtable backfills

### DIFF
--- a/crates/sui-kvstore/src/bigtable/client.rs
+++ b/crates/sui-kvstore/src/bigtable/client.rs
@@ -58,6 +58,11 @@ use crate::tables;
 
 const DEFAULT_MAX_DECODING_MESSAGE_SIZE: usize = 32 * 1024 * 1024;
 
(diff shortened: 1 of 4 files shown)


crates/sui-kvstore/src/bigtable/client.rs
```diff
@@ -147,8 +156,21 @@ impl BigTableClient {
...
+                tx.try_send(tonic::transport::channel::Change::Insert(
+                    i,
+                    endpoint.clone(),
+                ))
+                .expect("channel balancer dropped");
+            }
+            channel
+        } else {
+            endpoint.connect_lazy()
+        };
```
bmwill, line 171: This can probably be simplified to:
Channel::balance_list(std::iter::repeat(endpoint).take(pool_size))

bmwill: oh i see that the way Change works internally this may not work as i expected.

PR #25496, opened 2026-02-17
Defer transactions with unpaid consensus amplification

mystenmark:
## Summary

SIP-45 defines paid amplification where submitters pay extra gas (gas_price = N * RGP) to have N+1 validators submit their transaction. However, MEV searchers can bypass this by submitting directly to multiple validators without paying.

This PR defers transactions with "unpaid amplification" - where the observed duplicate count exceeds what the gas price covers:
- A transaction is deferred if: `occurrence_count > gas_price / RGP + 1`
- If amplification persists in subsequent rounds, deferral repeats up to `max_deferral_rounds_for_congestion_control`
- After exceeding the deferral limit, the transaction is allowed through

## Changes

- Add `defer_unpaid_amplification` protocol flag (enabled for devnet in v112)
- Track occurrence counts in `CommitHandlerState` during deduplication (reuses existing `occurrence_counts` HashMap)
- Check for unpaid amplification at the start of `handle_deferral_and_cancellation`
- Add `consensus_handler_unpaid_amplification_deferrals` metric

## Test plan

- [x] `cargo check -p sui-core -p sui-protocol-config`
- [x] `cargo xclippy`
- [x] `cargo nextest run -p sui-core --lib consensus_handler` (14/14 tests pass)

### DIFF
--- a/crates/sui-open-rpc/tests/snapshots/generate_spec__openrpc.snap.json
+++ b/crates/sui-open-rpc/tests/snapshots/generate_spec__openrpc.snap.json
@@ -1337,6 +1337,7 @@
                 "correct_gas_payment_limit_check": false,
                 "create_root_accumulator_object": false,
                 "debug_fatal_on_move_invariant_violation": false,
+                "defer_unpaid_amplification": false,
                 "dependency_linkage_error": false,
                 "deprecate_global_storage_ops": false,
                 "deprecate_global_storage_ops_during_deserialization": false,


crates/sui-benchmark/src/drivers/bench_driver.rs
```diff
@@ -1132,12 +1132,20 @@ async fn run_bench_worker(
...
-                            .execute_transaction_block(tx.clone())
-                        .then(|(client_type, res)| async move {
+
+                        // Occasionally submit to multiple validators to test unpaid amplification deferral.
+                        // With 5% probability, submit to 3 validators to trigger the deferral logic.
+                        let use_amplification = rand::thread_rng().gen_bool(0.05);
+                        let proxy = worker.execution_proxy.clone_new();
+                        let res = async move {
+                            let (client_type, res) = if use_amplification {
+                                proxy.execute_transaction_block_with_amplification(tx.clone(), 3).await
```
akichidis: Shall we make also the number of submissions also random? Something in the range `3...committee_size - 1`? In the spirit that we might increase the chances of testing long deferrals (across multiple commits)

crates/sui-benchmark/src/lib.rs
```diff
@@ -447,6 +458,35 @@ impl LocalValidatorAggregatorProxy {
             response.events.unwrap_or_default(),
         ))
     }
+
+    /// Submit a transaction to multiple validators to cause consensus amplification.
```
akichidis, line 477: Probably specify that the `num_validators` designate additional validators (ex if `num_validators == 1` , then the transaction will eventually be submitted to two validators)

crates/sui-core/src/consensus_handler.rs
```diff
@@ -1969,6 +1972,55 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+        if protocol_config.defer_unpaid_amplification() {
+            let occurrence_count = state
+                .occurrence_counts
+                .get(transaction.tx().digest())
+                .copied()
+                .unwrap_or(1);
+
+            let rgp = self.epoch_store.reference_gas_price();
+            let gas_price = transaction.tx().transaction_data().gas_price();
+            let allowed_count = (gas_price / rgp.max(1)) + 1;
```
akichidis, line 1984: I believe [at transaction driver level](https://github.com/MystenLabs/sui/blob/c19e916de408767804b4cf60f2f0a12abdd9bfd9/crates/sui-core/src/transaction_driver/mod.rs#L170C17-L170C37) we are a bit stricter on the amplification factor (we just do `gas_price/rgp.max(1)`) , whereas it seems that here we default to `allowed_count >= 2` . 

I am now thinking that it might be ok to not be totally aligned so we allow some slack for edge cases on the submission side and avoid end up deferring every single transaction. But equally we have to be very careful now on the submission side as any change on that side will impact this side as well.

crates/sui-core/src/consensus_handler.rs
```diff
@@ -1969,6 +1972,55 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+
+                let deferred_from_round = previously_deferred_tx_digests
+                    .get(transaction.tx().digest())
+                    .map(|k| k.deferred_from_round())
+                    .unwrap_or(commit_info.round);
+
+                let deferral_key = DeferralKey::new_for_consensus_round(
+                    commit_info.round + 1,
+                    deferred_from_round,
+                );
```
akichidis, line 1999: I see the `DeferralReason` enum and I am thinking how much of an issue it would be to introduce a new reason such as `UnpaidAmplification` . Does that reason appear somewhere? Do you think it would provide some value?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2868,9 +2917,14 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
-            let in_set = !processed_set.insert(key.clone());
+            // Increment count and check if this is a duplicate within this commit.
+            // This replaces the separate processed_set HashSet.
+            let count = occurrence_counts.entry(key.clone()).or_insert(0);
+            *count += 1;
+            let in_commit = *count > 1;
+
             let in_cache = self.processed_cache.put(key.clone(), ()).is_some();
-            if in_set || in_cache {
+            if in_commit || in_cache {
```
akichidis, line 2924: Out of curiosity - a deduped transaction will still appear in the checkpoint? I am mostly thinking the requirements around the determinism against the check on the LRU cache `processed_cache`. I am thinking that all honest validators will perform the exact same evictions - if needed - so in the end they'll end up with same dedupe results?

akichidis: Oh actually never mind - I see that below we also do a db check `is_consensus_message_processed`

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2883,11 +2937,31 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+                && count > 1
+            {
+                self.metrics
+                    .consensus_handler_duplicate_tx_count
+                    .observe(count as f64);
+            }
+        }
+
+        // Copy user transaction occurrence counts to state for unpaid amplification detection.
+        state.occurrence_counts.reserve(occurrence_counts.len());
```
lxfind, line 2959: Could we assert that `state.occurrence_counts` is still empty here? Want to make sure we don't accidentally modify it beforehand.

crates/sui-core/src/consensus_handler.rs
```diff
@@ -1969,6 +1972,55 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
     ) {
+        // Check for unpaid amplification before other deferral checks.
+        // SIP-45: Paid amplification allows (gas_price / RGP + 1) submissions.
+        // Transactions with more duplicates than paid for are deferred.
+        if protocol_config.defer_unpaid_amplification() {
+            let occurrence_count = state
+                .occurrence_counts
+                .get(transaction.tx().digest())
+                .copied()
+                .unwrap_or(1);
```
lxfind: Why 1 instead of 0?

crates/sui-core/src/consensus_handler.rs
mystenmark: this is just the formula written in the SIP - I can't figure out why it uses the + 1, but it doesn't seem like a big deal either way

crates/sui-core/src/consensus_handler.rs
mystenmark: I suppose 0 makes more sense - this is the number of occurrences in the current commit. since the tx may have been previously deferred, it may appear 0 times in the current commit.

crates/sui-core/src/consensus_handler.rs
mystenmark: its just used for the return value from the should_defer method, it wouldn't be necessary here unless we moved this logic into that method

PR #25498, opened 2026-02-17
Debottleneck indexing framework

nickvikeras:
## Description

I have noticed that indexing framework seems to have artificial software bottlenecks for high-throughput, small-batch-size indexers like GCS and Big Table. I hit throughput limits well before bandwidth, compute, or the database or object stores were ever saturated and no amount of concurrency-tuning or channel-sizing ever seemed to help.

In my java past life, I have seen this technique where "greedily" draining queues before yielding the thread massively improves throughput by avoid scheduling overhead. I tried applying it to these channels in tokio and it seemed to work here too.

My bigtable throughput doubled and I'm now able to saturate a 6-node cluster if I set the concurrency limits high enough.

image

## Test plan

It 2x'd throughput on my bigtable indexer. I'd like to a do a test where I do a bucket to bucket GCS copy using the object store indexer to see how fast it can go, but I haven't had a chance yet. This screenshot shows the ingestion throughput over the same watermark range. Commit throughput is similarly doubled. I suspect it would be more than doubled if I wasn't hitting database limits.

### DIFF
--- a/crates/sui-futures/src/stream.rs
+++ b/crates/sui-futures/src/stream.rs
@@ -89,10 +89,6 @@ impl<S: Stream + Sized + 'static> TrySpawnStreamExt for S {
                 }
             }
 
-            if draining && permits == limit {
-                break;
-            }
-
             tokio::select! {
                 biased;
 
--- a/crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
@@ -115,28 +115,35 @@ pub(super) fn collector<H: Handler + 'static>(
                 }
             }
 
-            // Eagerly build and send batches while enough data is pending.
-            while pending_rows >= H::MIN_EAGER_ROWS {
-                let (batch, watermark, batch_len) =
-                    build_batch::<H>(&*handler, &mut pending, &checkpoint_lag_reporter, &metrics);
-                pending_rows -= batch_len;
-
-                let batched_rows = BatchedRows {
-                    batch,
-                    batch_len,
-                    watermark,
-                };
-
-                if tx.send(batched_rows).await.is_err() {
-                    info!(
-                        pipeline = H::NAME,
-                        "Committer closed channel, stopping collector"
+            // Eagerly build and send batches if enough data is pending. Once we start
+            // batching, flush everything including any small tail below the threshold.
+            if pending_rows >= H::MIN_EAGER_ROWS {
+                loop {
+                    let (batch, watermark, batch_len) = build_batch::<H>(
+                        &*handler,
+                        &mut pending,
+                        &checkpoint_lag_reporter,
+                        &metrics,
                     );
-                    return Ok(());
-                }
+                    pending_rows -= batch_len;
 
-                if batch_len == 0 {
-                    break;
+                    let batched_rows = BatchedRows {
+                        batch,
+                        batch_len,
+                        watermark,
+                    };
+
+                    if tx.send(batched_rows).await.is_err() {
+                        info!(
+                            pipeline = H::NAME,
+                            "Committer closed channel, stopping collector"
+                        );
+                        return Ok(());
+                    }
+
+                    if batch_len == 0 {
+                        break;
+                    }
                 }
             }
 


crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
```diff
@@ -96,67 +96,78 @@ pub(super) fn collector<H: Handler + 'static>(
...
         loop {
-            tokio::select! {
-                // Time to create another batch and push it to the committer.
-                _ = poll.tick() => {
-                    let guard = metrics
-                        .collector_gather_latency
-                        .with_label_values(&[H::NAME])
-                        .start_timer();
+            // Eager inner loop: drain processor channel while under backpressure limit.
+            while pending_rows < H::MAX_PENDING_ROWS {
```
wlmyng: ok, similarly here, previously we receive one checkpoint from the channel, now we greedily drain

`try_recv()` is the equivalent to `now_or_never()` in the stream code? non-blocking check to grab everything already available

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
```diff
@@ -96,67 +96,78 @@ pub(super) fn collector<H: Handler + 'static>(
...
+            }
 
-                    metrics
-                        .collector_batch_size
-                        .with_label_values(&[H::NAME])
-                        .observe(batch_len as f64);
+            // Fall back to select! for waiting.
+            tokio::select! {
+                _ = poll.tick() => {
+                    let (batch, watermark, batch_len) = build_batch::<H>(
```
wlmyng, line 153: and we still need this to flush whatever we have, i.e under `MIN_EAGER_ROWS`

crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
```diff
@@ -54,167 +53,234 @@ pub(super) fn committer<H: Handler + 'static>(
...
+        let mut fatal_error: Option<anyhow::Error> = None;
+
+        loop {
+            // Eager inner loop: dispatch batches while capacity allows and items are ready,
+            // avoiding select! overhead when items are immediately available.
+            while !draining && active < limit {
+                match rx.try_recv() {
+                    Ok(batch) => {
+                        if batch.batch_len == 0 {
+                            if tx.send(batch.watermark).await.is_err() {
```
wlmyng: should we be awaiting here in the eager loop? I guess the idea is that, even though we're trying to be greedy, we still want to ensure correctness by making sure the watermark channel is receiving what the committer is handling

crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
```diff
@@ -54,167 +53,234 @@ pub(super) fn committer<H: Handler + 'static>(
...
+                        .with_label_values(&[H::NAME])
+                        .inc();
+
+                    Err(BE::transient(e))
+                }
+            }
+        }
+    };
+
+    backoff::future::retry(backoff, commit).await.unwrap();
```
wlmyng: unwrap safe here because infinite retries `max_elapsed_time: None`?

crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
```diff
@@ -54,167 +53,234 @@ pub(super) fn committer<H: Handler + 'static>(
...
-                                    );
+            if draining && active == 0 {
+                break;
+            }
 
-                                    metrics
-                                        .total_committer_batches_failed
-                                        .with_label_values(&[H::NAME])
-                                        .inc();
+            tokio::select! {
```
wlmyng: so there are three exit conditions

- `draining` is true (channel disconnected, watermark channel closed)
- `active >= limit` all concurrency slots full
- `try_recv` channel has no further batches

At the select, for most of the exit conditions `join_next` is the only valid branch, which drains `active` and then re-loops

except for `try_recv` empty, then both branches are valid, and either we finish a commit or a new batch arrives first, either way we loop back ... or if no tasks are in flight, only `rx.recv()` is live, so we wait until collector sends something

crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
```diff
@@ -54,167 +53,234 @@ pub(super) fn committer<H: Handler + 'static>(
...
+                        }
+                        Err(e) => {
+                            fatal_error = Some(anyhow::anyhow!("Committer task panicked: {e}"));
+                            draining = true;
+                        }
+                    }
+                }
 
-                                    Err(BE::transient(Break::Err(e)))
+                // Then read new items
```
wlmyng: This is an Or, not a Then right...?

crates/sui-futures/src/stream.rs
```diff
@@ -71,18 +71,27 @@ impl<S: Stream + Sized + 'static> TrySpawnStreamExt for S {
         let mut self_ = pin!(self);
 
         loop {
-            tokio::select! {
-                next = self_.next(), if !draining && permits > 0 => {
-                    if let Some(item) = next {
+            // Eager inner loop: spawn tasks while permits allow and items are ready,
+            // avoiding select! overhead when items are immediately available.
+            while !draining && permits > 0 {
+                match poll_fn(|cx| self_.as_mut().poll_next(cx)).now_or_never() {
```
bmwill, line 77: How does this compare to using https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html#method.ready_chunks

nickvikeras: That looks like a convenience method to do what im doing here. I can try and use it instead.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
amnn, line 1: Would it be possible to make the same change, functionally, by making the outer `tokio::select!` biased and moving the `recv` call to the top? AFAICT that's the same change, but without hurting readability so much.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
```diff
@@ -54,167 +53,234 @@ pub(super) fn committer<H: Handler + 'static>(
...
-                    let commit = move || {
-                        let batch = batch.clone();
-                        let handler = handler.clone();
-                        let db = db.clone();
-                        let metrics = metrics.clone();
-                        let checkpoint_lag_reporter = checkpoint_lag_reporter.clone();
-                        async move {
-                            if batch_len == 0 {
-                                return Ok(());
+        let limit = config.write_concurrency;
```
amnn: I thought your change to `try_for_each_spawned` would essentially result in the same change in behaviour here, or did I miss something?

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
amnn: Hmm, I suppose that will still `await` the `recv` which will yield back to the runtime... Could you wrap `rx` in a stream and then use `ready_chunks` here too? Then you could go back to roughly the same structure we had before, but we deal with chunks of rows at a time, rather than individual rows.

crates/sui-futures/src/stream.rs
nickvikeras: Actually I think the permit counting makes it kind of awkward to use ready_chunks here because you'd have to buffer the extra tasks you don't have permits for like this:

```
  let mut chunked = pin!(self.ready_chunks(limit));
  let mut buffer: VecDeque<Self::Item> = VecDeque::new();

  loop {
      // Drain buffer first
      while !draining && permits > 0 && !buffer.is_empty() {
          permits -= 1;
          join_set.spawn(f(buffer.pop_front().unwrap()));
      }

      if draining && permits == limit {
          break;
      }

      tokio::select! {
          biased;

          Some(res) = join_set.join_next() => {
              // ... permits += 1, handle errors, etc.
          }

          Some(chunk) = chunked.next(),
              if !draining && permits > 0 && buffer.is_empty() => {
              // chunk could have 10 items but only 3 permits available
              for item in chunk {
                  if permits > 0 {
                      permits -= 1;
                      join_set.spawn(f(item));
                  } else {
                      buffer.push_back(item);
                  }
              }
          }
      }
  }
```

crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
nickvikeras: I think you are correct. In my original branch I had started with this commit, then made the change to try_for_each_spawned but I never reverted this change. Let me test one time without it.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
nickvikeras: I didn't use ready chunks but I did eliminate the code duplication. IMO using ready_chunks isn't really any more readable and it doesn't really provide the exact semantics I want. I think the current loop logic is pretty clear.

1. drain the channel from the processor until you hit MAX_PENDING_ROWS backpressure.
2. Fire off all those batches to the comitter until you get blocked on backpressure from that channel, leaving 'MIN_EAGER_ROWS in pending (so we don't prematurely fire off tons of small batches).
3. Wait for either the timer to fire and send one batch OR the next checkpoint to arrive and receive it into the pending map.
4. Repeat. 

Adding in ready_chunks seems to make this more complicated because now I have to worry about pulling a huge number of rows beyond what MAX_PENDING_ROWS allows in my chunk.

nickvikeras: Actually with pull/25503, now we are guarding against max_pending_rows upstream at the pipeline level before the processor runs, so I think we don't have to check it in the collector's at all... but I still think ready_chunks isn't exactly the correct tool here. I want to drain _everything_, not some fixed number per select tick. 

It seems like ready_chunks would be really useful if you had two streams with this same bottlneck and want to fairly pull work from each one.

crates/sui-futures/src/stream.rs
amnn: Couldn't you limit the chunk size by the currently available `permits`?

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
amnn: > Adding in ready_chunks seems to make this more complicated because now I have to worry about pulling a huge number of rows beyond what MAX_PENDING_ROWS allows in my chunk.

I guess this is also related to my question above -- could you not apply the limit in `ready_chunks` based on `MAX_PENDING_ROWS` to make sure you don't exceed your bound in one shot? I could even imagine an approach where you create a stream in order to get one chunk from it, of at most the bounded size, and then you discard that stream.

Maybe that also suggests an abstraction for this general pattern of eagerly gathering values up to some limit (but not as a stream of chunks, just as a one-off).

nickvikeras: >could you not apply the limit in ready_chunks based on MAX_PENDING_ROWS to make sure you don't exceed your bound in one shot? 

the stream works at the checkpoint granularity and MAX_PENDING_ROWS works at the row granularity so not really since the number of rows/checkpoint is not predictable?

But if you look at pull/25503 again I don't think this is even relevant anymore.

amnn: Ah yes, got it, okay fair enough!

crates/sui-futures/src/stream.rs
nickvikeras: No it's a stream adapter not a 'pull N items from stream' and the number of available permits is dynamic within the lifecycle of the stream.

crates/sui-futures/src/stream.rs
```diff
@@ -71,18 +71,30 @@ impl<S: Stream + Sized + 'static> TrySpawnStreamExt for S {
...
                         // know we can wind down the pool.
                         draining = true;
                     }
+                    None => break,
                 }
+            }
+
+            if draining && permits == limit {
+                break;
+            }
```
amnn: Do you need this check here and below? If this condition is true then:

- `join_set` is empty, so `join_set.join_next()` will return `None`
- the guard on the `next` pattern is false (because it's guarded by `!draining`)

So we will evaluate the `else` condition straight away and exit.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
```diff
@@ -96,67 +96,60 @@ pub(super) fn collector<H: Handler + 'static>(
...
+                if batch_len == 0 {
+                    break;
+                }
+            }
 
-                    metrics
-                        .collector_batch_size
-                        .with_label_values(&[H::NAME])
-                        .observe(batch_len as f64);
+            // Fall back to select! for waiting.
```
amnn, line 150: Factoring out the functions does help a bit, but I'm not really seeing the value of still having this piece at the end here -- why not have a `loop` with `poll.tick()` at the top, followed by a `try_recv` loop, followed by batch construction? It would be a simpler implementation (could probably get rid of the helper functions again).

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
```diff
@@ -96,67 +96,60 @@ pub(super) fn collector<H: Handler + 'static>(
...
+                        );
                     }
-                    pending_rows -= batch_len;
-                    let elapsed = guard.stop_and_record();
-                    debug!(
+                    Err(_) => break,
+                }
+            }
+
+            // Eagerly build and send batches while enough data is pending.
```
amnn: If we have a small amount of data left over, I reckon we should just send it through, in a small batch. The pathology we're trying to avoid is sending a stream of small batches, which is solved by checking before we start generating batches at all whether we have `>= MIN_EAGER_ROWS` in total.

crates/sui-futures/src/stream.rs
nickvikeras: Fixed

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
nickvikeras: Fixed

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
nickvikeras: Let's discuss this one more, I wasn't sure about how removing the rx.recv() branch from the select arm would affect the throughput. I.e, if the processor channel is full it could stall everything until the timer wakes up.

amnn: Hmm, right - thinking...

amnn: What about if you rotate the loop -- start with a `tokio::select!` between the polling interval and fetching the first block, and continue with the `try_recv`, gather batch, etc?

nickvikeras: >What about if you rotate the loop -- start with a tokio::select! between the polling interval and fetching the first block, and continue with the try_recv, gather batch, etc?

I tried this and spent a few hours debugging weird timing issues in the tests and had to time box it. It seemed to slow down the movement of checkpoints out of the processor channel in a way I couldn't figure out.

PR #25503, opened 2026-02-18
row count backpressure

nickvikeras:
## Description

I have found it quite difficult to get predictable, low memory use with the indexing framework because there are occasionally very large checkpoints that create a huge number of rows, and sometimes there are a few of those in succession. It's especially common when indexing testnet.

I have found it to be _much_ easier to set a max number of in-flight rows (1M) and use that as backpressure. With this configuration, I get a predictable ~1-2gb of memory utilization rather than bouncing between 1gb to 20gb to 40gb to OOM.

I'm quite unfamiliar with the pruner/reader watermark stuff, so please check that part carefully to make sure I didn't mess anything up.

## Test plan

Tested in a test kvstore indexer environment. The metrics seem correct.


nickvikeras:
> I like the general direction to use row count as the back-pressure mechanism -- it certainly seems like that is a bigger contributor to the overall memory usage than the checkpoints themselves. I'd like to avoid making that a concern of the ingestion service though, if that's possible, for two reasons:
> 
>     * It's intended to be a standalone service, but the wider its interface becomes the harder it becomes to use in that capacity.
> 
>     * Currently this choice means we're treating all rows the same, which I'm not sure is right. A row from the `tx_digests` table is going to be smaller than a row for `kv_objects`, for example.
> 
> 
> Can we make this a concern for the individual pipelines? E.g. each pipeline tracks its own pending rows, and uses that as a way to regulate whether it accepts more checkpoints to process? That way we can also configure the pending rows per-pipeline.
> 
> I think the main issue that needs to be addressed with an approach like this is that ingestion service needs some way to avoid creating tasks that are just waiting to send checkpoints, but maybe this is less of an issue now that rows are being used for back-pressure? E.g. `checkpoint_buffer_size` now becomes an independent variable that controls how many checkpoints you can have queued up, but it's not the only parameter you have to tune back-pressure, you don't end up having to push it to high values to get more throughput.
> 
> We can also use this change as an opportunity to make this easier to track in prometheus as well (no polling updates to prometheus):
> 
>     * Introduce two new metrics per pipeline: rows in and rows out -- the processor increments rows in, and the committer increment rows out.
> 
>     * Then for the actual limiting behaviour we can either set-up a `Notify` that the committer uses to ping the checkpoint to say that it's relieved backpressure, or you can set-up a `watch` channel per-pipeline between these components, where the `processor` increments the channel, the committer decrements the channel, and the top of the pipeline waits for back-pressure to be relieved.
> 
> 
> There's benefits to both of those approaches, but what I'm keen to avoid is introducing polling to update prometheus, because prometheus is already performing polling in the background, so adding on an extra layer can cause confusion.

This made me realize this can all be massively simpler if we just move the MAX_PENDING_ROWS check to the processor instead of the collector.  

Check the new commit I think it addresses all of your concerns.

Also I didn't understand what new metrics you were asking for? it looks like we already have rows create/comitted ?

* https://github.com/MystenLabs/sui/blob/47562964b8e2ed4899912fc15365eb7d0b3934f9/crates/sui-indexer-alt-framework/src/metrics.rs?plain=1#L84
* https://github.com/MystenLabs/sui/blob/47562964b8e2ed4899912fc15365eb7d0b3934f9/crates/sui-indexer-alt-framework/src/metrics.rs?plain=1#L99

amnn:
Awesome, will take a deeper look tomorrow -- and ah, if we already have those metrics, then it seems like we don't need to track the back pressure metric specially? It should now be the difference between those.

nickvikeras:
> Awesome, will take a deeper look tomorrow -- and ah, if we already have those metrics, then it seems like we don't need to track the back pressure metric specially? It should now be the difference between those.

Yeah I added that metric when I was debugging throughput issues cuz I didn't realize I could already compute it from the existing metrics.

nickvikeras:
I'm now realizing this adds a deadlock mode to sequential pipelines because the processor can stop producing rows before the comitter commits...

I removed row-based backpressure from sequential pipelines and I think we will want to leave the regulator in place for sequential pipelines.

nickvikeras:
updated it to work only between collector and comitter

crates/sui-indexer-alt-framework/src/pipeline/concurrent/pending_rows.rs
amnn, line 1: We talked offline about how this is a [`Semaphore`](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html) in disguise -- but that turns out not to be a great fit, because a semaphore is strict about how many permits it has, while this abstraction will allow a request to go through if there's at least one permit, even if the consequence of that is that we go over the allowed permits (and this is important for liveness to deal with single large checkpoints, etc)

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
```diff
@@ -261,7 +290,14 @@ fn receive_checkpoint<H: Handler>(
         .set(reader_lo as i64);
 
     let len = indexed.len();
-    pending.insert(indexed.checkpoint(), indexed.into());
+
+    let guard = if !skipped && len > 0 {
```
amnn: nit: `skipped` should imply `len == 0`, no? So do you need to keep `skipped` around to add it to this check here?

crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
```diff
@@ -201,6 +202,11 @@ pub(super) fn committer<H: Handler + 'static>(
                     };
 
                     async move {
+                        // Hold the row guard until after the commit so the
+                        // inflight counter stays accurate for collector
+                        // backpressure.
+                        let _guard = guard;
```
amnn: nit: it may be even clearer to add a call to `drop(guard)` at the bottom of the `async move` block, so then we don't even need to think about the scope implications of moving the `guard` in here, or dress it with an `_`, etc.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
```diff
@@ -104,68 +111,48 @@ pub(super) fn collector<H: Handler + 'static>(
...
-                        .total_collector_rows_received
-                        .with_label_values(&[H::NAME])
-                        .inc_by(recv_rows as u64);
-                    metrics
-                        .total_collector_checkpoints_received
-                        .with_label_values(&[H::NAME])
-                        .inc_by(recv_cps as u64);
-
-                    if pending_rows < min_eager_rows {
+                Some(indexed) = rx.recv(), if !backpressured => {
```
amnn: There was a thought in the back of my mind about dealing with the case where the a commit lands while the system is back-pressured, and maybe not immediately jumping out of the loop, which was not a big deal, but I nerdsniped myself. I think you can address that by adding something like this to `InflightRows`:

```rust
impl InflightRows {
    async fn backpressured<T>(&self, max: usize, f: impl Future<T>) {
        while self.counter.load(Ordering::Relaxed) >= max {
            // Wait for the next commit to relieve pressure
            self.notify.notified().await;
        }

        f.await
    }
}
```

And using that here:

```rust
Some(indexed) = inflight.backpressured(rx.recv()) => { /* ... */ }
```

PR #25505, opened 2026-02-18
Decouple cross-pipeline lag and ingestion channel size

nickvikeras:
## Description

I find it kind of hard to reason about the memory consumption and throughput consequences of checkpoint-buffer-size since it controls both the lag between the slowest and fastest pipeline _and_ the size of the ingestion channel. I find it much easier to set it to infinity, then a set a reasonably small ingestion channel size, and cap memory usage via the new row counter.

### DIFF
--- a/crates/sui-checkpoint-blob-indexer/src/config.rs
+++ b/crates/sui-checkpoint-blob-indexer/src/config.rs
@@ -78,7 +78,6 @@ pub struct IngestionConfig {
     pub streaming_backoff_max_batch_size: usize,
     pub streaming_connection_timeout_ms: u64,
     pub streaming_statement_timeout_ms: u64,
-    pub checkpoint_channel_size: usize,
 }
 
 impl Default for IngestionConfig {
@@ -97,7 +96,6 @@ impl From<framework::ingestion::IngestionConfig> for IngestionConfig {
             streaming_backoff_max_batch_size: config.streaming_backoff_max_batch_size,
             streaming_connection_timeout_ms: config.streaming_connection_timeout_ms,
             streaming_statement_timeout_ms: config.streaming_statement_timeout_ms,
-            checkpoint_channel_size: config.checkpoint_channel_size,
         }
     }
 }
@@ -112,7 +110,6 @@ impl From<IngestionConfig> for framework::ingestion::IngestionConfig {
             streaming_backoff_max_batch_size: config.streaming_backoff_max_batch_size,
             streaming_connection_timeout_ms: config.streaming_connection_timeout_ms,
             streaming_statement_timeout_ms: config.streaming_statement_timeout_ms,
-            checkpoint_channel_size: config.checkpoint_channel_size,
         }
     }
 }
--- a/crates/sui-indexer-alt-consistent-store/src/config.rs
+++ b/crates/sui-indexer-alt-consistent-store/src/config.rs
@@ -44,7 +44,6 @@ pub struct IngestionConfig {
     pub streaming_backoff_max_batch_size: usize,
     pub streaming_connection_timeout_ms: u64,
     pub streaming_statement_timeout_ms: u64,
-    pub checkpoint_channel_size: usize,
 }
 
 #[DefaultConfig]
@@ -152,7 +151,6 @@ impl From<framework::ingestion::IngestionConfig> for IngestionConfig {
             streaming_backoff_max_batch_size: config.streaming_backoff_max_batch_size,
             streaming_connection_timeout_ms: config.streaming_connection_timeout_ms,
             streaming_statement_timeout_ms: config.streaming_statement_timeout_ms,
-            checkpoint_channel_size: config.checkpoint_channel_size,
         }
     }
 }
@@ -167,7 +165,6 @@ impl From<IngestionConfig> for framework::ingestion::IngestionConfig {
             streaming_backoff_max_batch_size: config.streaming_backoff_max_batch_size,
             streaming_connection_timeout_ms: config.streaming_connection_timeout_ms,
             streaming_statement_timeout_ms: config.streaming_statement_timeout_ms,
-            checkpoint_channel_size: config.checkpoint_channel_size,
         }
     }
 }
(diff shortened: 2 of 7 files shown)


amnn:
Checkpoint buffer size needs to be coupled with the ingestion channel size to prevent a deadlock in sequential pipelines: If the former is larger than the latter (or unbounded), you can get into a situation where a sequential pipeline is waiting for some checkpoint `C`, and so is refusing to commit anything, but as a result, it is not emptying its ingestion channel which is full, so `C` never makes it through the pipeline to unstick everything.

nickvikeras:
> Checkpoint buffer size needs to be coupled with the ingestion channel size to prevent a deadlock in sequential pipelines: If the former is larger than the latter (or unbounded), you can get into a situation where a sequential pipeline is waiting for some checkpoint `C`, and so is refusing to commit anything, but as a result, it is not emptying its ingestion channel which is full, so `C` never makes it through the pipeline to unstick everything.

I think I see a deadlock risk that already exists today where if `checkpoint_buffer_size <= checkpoint_lag`, then the comitter will get stuck and never be able to satisfy [this check](https://github.com/MystenLabs/sui/blob/f6f49270d47632864f5853ae420bf8fce4e385e4/crates/sui-indexer-alt-framework/src/pipeline/sequential/committer.rs?plain=1#L398), but I don't see how this PR is making the risk of that happening any worse/different?

As far as I can understand, the sequential comitter will always unconditionally drain its channel [here](https://github.com/MystenLabs/sui/blob/f6f49270d47632864f5853ae420bf8fce4e385e4/crates/sui-indexer-alt-framework/src/pipeline/sequential/committer.rs?plain=1#L346-L371) into the unbounded btreemap, which would then unblock the processor from producing its next row(s), so it would pull the next checkpoint from the ingestion channel [here](https://github.com/MystenLabs/sui/blob/f6f49270d47632864f5853ae420bf8fce4e385e4/crates/sui-indexer-alt-framework/src/pipeline/processor.rs?plain=1#L84-L86). IndexedCheckpoints would just keep building up in that committer pending btreemap until the `can_process_pending` check can pass.

Even if you set the `checkpoint_channel_size` size to 1, I don't think it would deadlock. 

If we ship my row-count-based back-pressure change, I think we could even consider just deprecating checkpoint_buffer_size config entirely. It doesn't seem to earn it's weight if the pending row count is a better signal for memory pressure.

amnn:
> As far as I can understand, the sequential comitter will always unconditionally drain its channel [here](https://github.com/MystenLabs/sui/blob/f6f49270d47632864f5853ae420bf8fce4e385e4/crates/sui-indexer-alt-framework/src/pipeline/sequential/committer.rs?plain=1#L346-L371) into the unbounded btreemap, which would then unblock the processor from producing its next row(s), so it would pull the next checkpoint from the ingestion channel [here](https://github.com/MystenLabs/sui/blob/f6f49270d47632864f5853ae420bf8fce4e385e4/crates/sui-indexer-alt-framework/src/pipeline/processor.rs?plain=1#L84-L86). IndexedCheckpoints would just keep building up in that committer pending btreemap until the `can_process_pending` check can pass.

Ah yes, you're right -- I was sure at some point we bounded the sizes of all structures at every stage of these pipelines, but maybe I'm misremembering or things have changed -- this was the whole reason why we introduced the regulator component in the first place (the potential for the deadlock that I described above).

> If we ship my row-count-based back-pressure change, I think we could even consider just deprecating checkpoint_buffer_size config entirely. It doesn't seem to earn it's weight if the pending row count is a better signal for memory pressure.

Is your point here that we may not need to regulate/apply back pressure according to the checkpoint buffer size anymore? If so, I think I agree, but then I'd vote for just doing that (removing the back pressure logic) but keeping the name the same and only using it to control the channel size (which we can also make smaller if we want).

nickvikeras:
> > As far as I can understand, the sequential comitter will always unconditionally drain its channel [here](https://github.com/MystenLabs/sui/blob/f6f49270d47632864f5853ae420bf8fce4e385e4/crates/sui-indexer-alt-framework/src/pipeline/sequential/committer.rs?plain=1#L346-L371) into the unbounded btreemap, which would then unblock the processor from producing its next row(s), so it would pull the next checkpoint from the ingestion channel [here](https://github.com/MystenLabs/sui/blob/f6f49270d47632864f5853ae420bf8fce4e385e4/crates/sui-indexer-alt-framework/src/pipeline/processor.rs?plain=1#L84-L86). IndexedCheckpoints would just keep building up in that committer pending btreemap until the `can_process_pending` check can pass.
> 
> Ah yes, you're right -- I was sure at some point we bounded the sizes of all structures at every stage of these pipelines, but maybe I'm misremembering or things have changed -- this was the whole reason why we introduced the regulator component in the first place (the potential for the deadlock that I described above).
> 
> > If we ship my row-count-based back-pressure change, I think we could even consider just deprecating checkpoint_buffer_size config entirely. It doesn't seem to earn it's weight if the pending row count is a better signal for memory pressure.
> 
> Is your point here that we may not need to regulate/apply back pressure according to the checkpoint buffer size anymore? If so, I think I agree, but then I'd vote for just doing that (removing the back pressure logic) but keeping the name the same and only using it to control the channel size (which we can also make smaller if we want).

yeah okay I think we are on the same page. Lemme try that out and make sure everything seems stable in my test environment...

nickvikeras:
Actually I realized sequential pipelines can deadlock if we add row-based backpressure to them because the processor can stop producing rows before the comitter has a contiguous sequence of rows to commit.

So I think we need to keep the regulator as a backpressure mechanism for sequential pipelines, and hence it makes sense to have different config knobs for the the regulator window size and the and ingestion channel size.

amnn (approved):
Accepting to unblock, but question for you: In the world we talked about where concurrent pipelines are not regulated, does it make sense to have both parameters? IIUC:

 - For an instance of the indexer that is only running concurrent pipelines, the `checkpoint_buffer_size` parameter would basically do nothing, and you would only care about `checkpoint_channel_size` (and this is the case you're mostly dealing with now, right?)
- For an instance of the indexer that is running sequential pipelines, you now have an extra knob to turn, but it can only have one effect: when `channel_size < buffer_size`, you can introduce a bottleneck but this doesn't seem like a desirable thing to do.
- For an instance of the indexer that is running both kinds of pipeline, you could set `channel_size > buffer_size`, and that would reduce bottlenecking when a concurrent pipeline runs slower than the sequential pipeline for some period of time (its processor channel would soak up checkpoints, allowing the sequential pipeline to run ahead. This circumstance seems rare though -- concurrent pipelines are usually faster than sequential pipelines.

To me, it seems like once the unified regulation logic is wound back, `checkpoint_buffer_size` basically becomes `checkpoint_channel_size` in the case you care about.

If we do go down this route though, please add a release notes section detailing the change, thanks!

PR #25507, opened 2026-02-18
Channel utilizaiton metrics

nickvikeras:
## Description

Pretty self-explainatory. These are helpful in debugging throughput bottlenecks.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/monitored_channel.rs
+++ b/crates/sui-indexer-alt-framework/src/monitored_channel.rs
@@ -1,0 +1,223 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use futures::Stream;
+use prometheus::IntGauge;
+use tokio::sync::mpsc;
+
+/// Creates a bounded channel whose in-flight item count is tracked by the provided
+/// [`IntGauge`]. The gauge is incremented on every successful `send` and decremented
+/// on every successful `recv` / `try_recv` / stream poll.
+pub fn channel<T>(size: usize, inflight: IntGauge) -> (Sender<T>, Receiver<T>) {
+    let (tx, rx) = mpsc::channel(size);
+    (
+        Sender {
+            inner: tx,
+            inflight: inflight.clone(),
+        },
+        Receiver {
+            inner: rx,
+            inflight,
+        },
+    )
+}
+
+/// Wraps [`mpsc::Sender`] with inline gauge tracking.
+pub struct Sender<T> {
+    inner: mpsc::Sender<T>,
+    inflight: IntGauge,
+}
+
+impl<T> Sender<T> {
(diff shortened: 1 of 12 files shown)


amnn (commented):
Have a look at [`mysten-metrics::metered_channel`](https://github.com/MystenLabs/sui/blob/aa5dd6a17d6170ede714f35845dc16f81ce5933b/crates/mysten-metrics/src/metered_channel.rs).

I think you're trying to achieve something similar here. We didn't use this crate initially because it's quite specific to the mono-repo, and we were trying to minimise dependencies to avoid having to port them when we pulled the framework out of the mono-repo, but if it's the right tool for the job, maybe we should just take the hit and use it for now?

It also avoids the thing I mentioned on an earlier PR of introducing polling in front of updating prometheus metrics.

nickvikeras:
> Have a look at [`mysten-metrics::metered_channel`](https://github.com/MystenLabs/sui/blob/aa5dd6a17d6170ede714f35845dc16f81ce5933b/crates/mysten-metrics/src/metered_channel.rs).
> 
> I think you're trying to achieve something similar here. We didn't use this crate initially because it's quite specific to the mono-repo, and we were trying to minimise dependencies to avoid having to port them when we pulled the framework out of the mono-repo, but if it's the right tool for the job, maybe we should just take the hit and use it for now?

metered_channel is [marked as deprecated](https://github.com/MystenLabs/sui/blob/aa5dd6a17d6170ede714f35845dc16f81ce5933b/crates/mysten-metrics/src/metered_channel.rs#L319) and it's replacement monitored_mpsc doesn't have a way to add custom labels to the metrics which would make it impossible to add a pipeline label, which would be sad for us.

Rather than refactoring the monitored_mpsc to support the labels we need, it seems easiest to just add our own lightweight wrapper in the framework, especially if you don't want to depend on the monorepo's utilities in the long run. 

I implemented that in the latest revision.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
```diff
@@ -315,19 +315,23 @@ mod tests {
...
     use crate::metrics::tests::test_metrics;
+    use crate::monitored_channel;
     use crate::pipeline::PendingRowsGuard;
     use crate::pipeline::Processor;
     use crate::pipeline::concurrent::BatchStatus;
     use crate::types::full_checkpoint_content::Checkpoint;
+    use async_trait::async_trait;
+    use prometheus::IntGauge;
+    use sui_pg_db::Connection;
+    use sui_pg_db::Db;
```
amnn, line 327: nit: import order (these should go back to where they were before).

crates/sui-indexer-alt-framework/src/monitored_channel.rs
amnn, line 1: Looks good -- just file ordering on this -- types, then impls, then trait impls, then free functions, then tests.

nickvikeras: oh yeah. I will add this to my claude.local.md

nickvikeras:
As discussed dont need this, can compute from channel in/out metrics

PR #25508, opened 2026-02-18
Fix typo

akmysti:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


akmysti:
> Good spot!
> 
> Two things before we can land this:
> 
> 1. This should probably be landed on `main` and not on bella-ciao;
> 2. It's annoying, but I think this change will also need to be protocol-gated since it will change error behavior/data.

Closing in favor of  https://github.com/MystenLabs/sui/pull/25553

PR #25509, opened 2026-02-18
Introduce ExecutionErrorTrait into sui-execution

jordanjennings-mysten:
- create trait
- add error type parameter for execution
- implement across programable transaction and static programmable transaction as Mode::Error
- uses E: ExecutionErrorTrait in areas of execution where mode is less relevant

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-types/src/error.rs
```diff
@@ -1162,6 +1166,41 @@ type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
 pub type ExecutionErrorKind = ExecutionFailureStatus;
 
+/// A trait for execution errors that provides common methods for accessing error information and creating new errors.
+pub trait ExecutionErrorTrait:
```
jordanjennings-mysten, line 1163: Not my favorite name but seems to be a pattern.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/mod.rs
tnowacki, line 25: I'm bit confused why the return type for these functions did not change

crates/sui-types/src/error.rs
```diff
@@ -59,6 +59,8 @@ macro_rules! make_invariant_violation {
         if cfg!(debug_assertions) {
             panic!($($args),*)
         }
+        #[allow(unused_imports)]
+        use $crate::error::ExecutionErrorTrait;
```
amnn: This will need a broader change later -- the call into `ExecutionError` here is going to need to be replaced with a call to this function, but on some specific implementation of `ExecutionErrorTrait` which is a bit tricky to get right in a macro.

crates/sui-types/src/error.rs
amnn: Once we've been able to eliminate `ExecutonError` as a type, we can co-opt its name.

crates/sui-types/src/error.rs
```diff
@@ -1162,6 +1166,41 @@ type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
...
+    }
+
+    fn invariant_violation<E: Into<BoxError>>(source: E) -> Self
+    where
+        Self: Sized,
+    {
+        Self::new_with_source(ExecutionFailureStatus::InvariantViolation, source.into())
+    }
+
+    fn from_kind(kind: ExecutionErrorKind) -> Self;
```
amnn: ...this is just another way of sneaking an implementation of `From<ExecutionErrorkind>` in, which we want to avoid, is it actually used somewhere?

crates/sui-types/src/error.rs
amnn, line 1163: Let's be consistent in the trait definition about `ExecutionErrorKind` vs `ExecutionFailureStatus` -- the fact that we are using both is breaking my brain.

crates/sui-types/src/error.rs
```diff
@@ -1162,6 +1166,41 @@ type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
...
+            Box::new(e),
+        )
+    }
+
+    fn invariant_violation<E: Into<BoxError>>(source: E) -> Self
+    where
+        Self: Sized,
+    {
+        Self::new_with_source(ExecutionFailureStatus::InvariantViolation, source.into())
+    }
```
amnn: I'm not sure you want these yet -- they look like exactly the interface for constructing an `ExecutionError`, ported into this trait. We're trying to replace this API because it's at the wrong level of abstraction -- we want to push it further into the execution layer where we have more access to the details of the source error. By adding these APIs here, you're making that job more difficult for yourself.

Where you use these APIs now, stick with constructing an `ExecutionError` and then converting it into an `ExecutionErrorTrait` impl at the latest possible stage. Then go back and push the `ExecutionErrorTrait` further down into the execution layer to the places where the `ExecutionError` was constructed, or perhaps even further, to where its source was created.

crates/sui-types/src/error.rs
```diff
@@ -1185,49 +1224,43 @@ impl ExecutionError {
...
-    pub fn from_kind(kind: ExecutionErrorKind) -> Self {
-        Self::new(kind, None)
+impl std::error::Error for ExecutionError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        self.inner.source.as_ref().map(|e| &**e as _)
     }
+}
 
-    pub fn kind(&self) -> &ExecutionErrorKind {
+impl ExecutionErrorTrait for ExecutionError {
```
amnn: `ExecutionError` is not supposed to implement `ExecutionErrorTrait`. `ExecutionErrorKind` is.

amnn:
> I wonder if perhaps we should remove `From<ExecutionError>` on the trait.

@tnowacki some context -- the idea was that we would remove `ExecutionError` in increments, and until it is removed, the trait implements also requires `From<ExecutionError>` so you can convert into the trait type on the boundary. That is not what this PR is doing though.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/mod.rs
jordanjennings-mysten: I'll take a closer look might have been unnecessary.

jordanjennings-mysten:
> The fact that you have had to introduce imports for `ExecutionErrorTrait` in a bunch of places so that you can keep constructing an `ExecutionError` is a sign that something has gone wrong.
> 
> Remember, in this PR, you want to introduce the trait right on the interface boundary, keep using `ExecutionError` as long as possible, and then at the last moment, convert it into an instance of `ExecutionErrorTrait`. Then you can push the trait further down into the execution layer, incrementally. That means:
> 
> * This change should only introduce the trait, an impl for `ExecutionFailureKind`, and a conversion from `ExecutionError`.
> * `ExecutionError` should not implement `ExecutionErrorTrait`. You only want to implement that trait for error types that are destined to leave the execution layer. The whole point of this first step is that you are trapping `ExecutionError` inside the execution layer, and then slowly getting rid of it.

I think this may conflict with keeping backwards compatibility since error source would be dropped on Fullnodes. For example `test_arity_mismatch` no longer passes since the source information is dropped if we "narrow" the type down to ExecutionErrorKind.

crates/sui-types/src/error.rs
tzakian: Not sure if I'm being incredibly dense here but this seems to be doing the second part as well (pushing it down to where the error is created -- i.e, where the execution errors are created today), which seems largely fine to me.

My reading of this is that this is just trying to supplant the existing API on `ExecutionError` to be over this so that `ExecutionError::fun_call(...)` can be replaced with `Mode::Error::fun_call(..)` which seems reasonable to me unless you're envisioning this changing the way the errors are constructed completely possibly.

tzakian:
> ExecutionError should not implement ExecutionErrorTrait. You only want to implement that trait for error types that are destined to leave the execution layer. The whole point of this first step is that you are trapping ExecutionError inside the execution layer, and then slowly getting rid of it.

This is the eventual goal, but in the interim/as part of the conversion process it seems fine to me that `ExecutionErrorTrait` would be implemented for `ExecutionError` and then that is slowly removed and then eventually completely removed.

tzakian:
But that being said, I feel as though this PR is occupying a middle ground on both sides of the issues that Ashok mentioned above -- 

1. It pushes the `ExecutionErrorTrait`/`Mode::Error::fun_call` constructors deeply; but
2. It doesn't quite push it far enough, so it winds up that `ExecutionErrorTrait` needs to be implemented for `ExecutionError`.

I think it would probably be good to either 
1. split this up into a very simple "intro trait, and add conversion/implentation for ExecutionError" type of PR (and maybe even plumbing in the `Mode` but not using it yet), with follow-on PRs to scoop out the different parts bit-by-bit; or
2. Go all-in and basically one-shot it in a single PR, which I guess you could do, but please don't.... 🫠

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/interpreter.rs
```diff
@@ -378,7 +378,8 @@ fn execute_command<Mode: ExecutionMode>(
     if Mode::TRACK_EXECUTION {
         let argument_updates = context.argument_updates(args_to_update)?;
         let command_result = context.tracked_results(&result, &result_type)?;
-        Mode::finish_command_v2(mode_results, argument_updates, command_result)?;
+        Mode::finish_command_v2(mode_results, argument_updates, command_result)
```
tzakian: Kinda surprised this is needed as `?` should do the `into()` call on the error. But sometimes Rust surprises you 🤷🏻

sui-execution/latest/sui-adapter/src/execution_engine.rs
tzakian: Mode::Error? That would save the changes needed below I think?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/interpreter.rs
jordanjennings-mysten: Yeah I had the same problem with this line and I'm not sure why that is..

crates/sui-types/src/error.rs
amnn: So the reason for my comment is that this change **is** doing part of the second part (pushing it down to where the error is created), but by following exactly the interface that `ExecutionError` used, we are baking in the exact depth the abstraction gets pushed into, where in some cases we may want to push it further than that.

`invariant_violation` is a great example of that -- in many cases, we may want to push the constructor even further down to capture specific invariant violations and their details, before they get boxed up into a dynamic error.

sui-execution/latest/sui-adapter/src/execution_engine.rs
tnowacki: +1

crates/sui-types/src/error.rs
tnowacki: +1

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/interpreter.rs
tnowacki: I'm very confused because this is what ? does

sui-execution/latest/sui-adapter/src/programmable_transactions/execution.rs
```diff
@@ -139,7 +141,7 @@ mod checked {
         gas_charger: &mut GasCharger,
         pt: ProgrammableTransaction,
         trace_builder_opt: &mut Option<MoveTraceBuilder>,
-    ) -> Result<Mode::ExecutionResults, ExecutionError> {
+    ) -> Result<Mode::ExecutionResults, Mode::Error> {
```
tnowacki: Do we need to change this (here and below) code at all? It is dead

jordanjennings-mysten:
> Let's be consistent in the trait definition about ExecutionErrorKind vs ExecutionFailureStatus -- the fact that we are using both is breaking my brain.

created a PR over here that cleans this up https://github.com/MystenLabs/sui/pull/25710

crates/simulacrum/src/epoch_state.rs
```diff
@@ -143,7 +143,7 @@ impl EpochState {
         let transaction_data = transaction.data().transaction_data();
         let (kind, signer, gas_data) = transaction_data.execution_parts();
         let (inner_temp_store, gas_status, effects, _timings, result) =
-            self.executor.execute_transaction_to_effects(
+            self.executor.execute_transaction_to_effects_for_fullnode(
```
jordanjennings-mysten: using tims suggestion I'm pulling the behavior up to the executor Trait so if you want the previous behavior you will want the `execute_transaction_to_effects_for_fullnode()`

crates/sui-core/src/authority.rs
```diff
@@ -1957,29 +1957,63 @@ impl AuthorityState {
...
+                        signer,
+                        tx_digest,
+                        &mut None,
+                    );
+                (
+                    inner_temp_store,
+                    gas_status,
+                    effects,
+                    timings,
+                    result.map_err(Into::into),
```
jordanjennings-mysten: branching on fullnode vs validator then normalizing the error to ExecutionError

crates/sui-types/src/error.rs
```diff
@@ -1157,7 +1157,29 @@ impl std::fmt::Debug for SuiError {
...
+
+/// A trait for execution errors that provides common methods for accessing error information and creating new errors.
+pub trait ExecutionErrorTrait:
+    From<ExecutionError> + Debug + std::error::Error + Send + Sync + Sized
+{
+    fn new(
+        kind: ExecutionErrorKind,
+        command: Option<CommandIndex>,
+        source: Option<BoxError>,
+    ) -> Self;
```
jordanjennings-mysten: I suppose we could From<ExecutionError> for every instantiation but that didnt quite make sense to me. although `new` is possibly at odds with what was said earlier in this thread about getting around `From<ExecutionErrorKind>`

crates/sui-types/src/error.rs
```diff
@@ -1157,7 +1157,29 @@ impl std::fmt::Debug for SuiError {
...
+    fn with_command_index(self, command: CommandIndex) -> Self;
```
jordanjennings-mysten, line 1166: There are areas in the executor where its not known what command index until higher in the stack so seems very necessary

crates/sui-types/src/error.rs
```diff
@@ -1157,7 +1157,29 @@ impl std::fmt::Debug for SuiError {
...
+    fn kind(&self) -> &ExecutionErrorKind;
```
jordanjennings-mysten, line 1167: there are areas in execution gas charger calculation which need to branch on `kind`

crates/sui-types/src/error.rs
```diff
@@ -1157,7 +1157,29 @@ impl std::fmt::Debug for SuiError {
...
+    fn command(&self) -> Option<CommandIndex>;
+    fn source_ref(&self) -> Option<&(dyn std::error::Error + 'static)>;
+
+    fn to_execution_failure(&self) -> ExecutionFailure {
+        ExecutionFailure {
+            error: self.kind().clone(),
+            command: self.command(),
+        }
+    }
```
jordanjennings-mysten, line 1177: used for generating the onchain ExecutionStatus

crates/sui-types/src/error.rs
jordanjennings-mysten: I think the only option we have here is make a second type `ExecutionErrorWithSource` which mirrors the ExecutionError then we can get away from the `ExecutionError -> ExecutionErrorTrait -> ExecutionError` sandwich. I haven't done that here yet since it would maybe grow the PR a bit more to find all the places in authority where ExecutionError is and convert them over but think it might be okay for now.

crates/sui-types/src/execution_status.rs
```diff
@@ -27,6 +29,52 @@ pub struct ExecutionFailure {
...
+    fn from(value: ExecutionError) -> Self {
+        let (kind, command) = value.to_execution_status();
+        Self {
+            error: kind,
+            command,
+        }
+    }
+}
+
+impl Error for ExecutionFailure {}
```
jordanjennings-mysten, line 41: this is our smallest unit of error (kind + command index)

crates/sui-types/src/error.rs
```diff
@@ -1216,6 +1233,24 @@ impl ExecutionError {
...
+
+    fn command(&self) -> Option<CommandIndex> {
+        self.command()
+    }
+
+    fn source_ref(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        self.inner.source.as_deref().map(|e| e as _)
+    }
+}
```
jordanjennings-mysten, line 1253: Probably could get rid of this and create a cleaner type graph but we would end up creating an error type that mirrors this one (call it ExecutionErrorWithSource?), I'm sort of with tim on this where maybe this exeuction error sandwich is not ideal due to the loops in the type graph but maybe something we will just slowly phase out. If we try and fix this in the PR here it will require a lot more changes.

Another option is make the ExecutionError -> ExecutionErrorWithSource in a separate PR try that cleanup on its own then integrate the change here.

crates/sui-types/src/error.rs
```diff
@@ -1234,6 +1269,18 @@ impl From<ExecutionErrorKind> for ExecutionError {
...
+    fn from(value: ExecutionFailure) -> Self {
+        let ExecutionFailure { error, command } = value;
+        let err = ExecutionError::from_kind(error);
+        if let Some(command) = command {
+            err.with_command_index(command)
+        } else {
+            err
+        }
+    }
+}
```
jordanjennings-mysten, line 1283: Lower error information -> higher error information is safe. Although this makes me think about maybe fixing some of the original interfaces here..

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -341,16 +342,17 @@ mod checked {
...
                     let mut execution_result: ResultWithTimings<
                         Mode::ExecutionResults,
-                        ExecutionError,
+                        Mode::Error,
                     > = match execution_params {
-                        ExecutionOrEarlyError::Err(early_execution_error) => {
-                            Err((ExecutionError::new(early_execution_error, None), vec![]))
-                        }
+                        ExecutionOrEarlyError::Err(early_execution_error) => Err((
+                            ExecutionError::from_kind(early_execution_error).into(),
```
jordanjennings-mysten: there's quite a few `ExecutionError::from_kind().into()`s going on might be better off just having a `new(kind, command_index, source)` on the ExecutionErrorTrait?

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -699,9 +703,10 @@ mod checked {
                 None,
                 pt,
                 trace_builder_opt,
-            ),
+            )
+            .map_err(|(e, timings)| (e.into(), timings)),
             TransactionKind::ProgrammableSystemTransaction(pt) => {
-                SPT::execute::<execution_mode::System>(
+                SPT::execute::<execution_mode::System<Mode::Error>>(
```
jordanjennings-mysten, line 735: Mode::Error doesn't do anything yet with the new VM, later on it will allow us to plumb Error further down. Which I assume we want.

sui-execution/latest/sui-adapter/src/execution_mode.rs
```diff
@@ -190,14 +259,18 @@ impl ExecutionMode for Genesis {
 }
 
 #[derive(Copy, Clone)]
-pub struct System;
+pub struct System<E = ExecutionError>(PhantomData<fn() -> E>);
```
jordanjennings-mysten, line 201: This helps later on, `execution_loop` fn needs to select `E`, in otherwords System mode could either be running on fullnode or validator.

crates/sui-types/src/error.rs
jordanjennings-mysten: I believe I've mostly resolved this and limited it to the basic set of functions necessary in the trait. I do think it will take time to reveal the exact shape as we expand the borders and what we capture.

sui-execution/latest/sui-adapter/src/programmable_transactions/execution.rs
jordanjennings-mysten: fixed

crates/sui-types/src/error.rs
amnn: Why do you need this? This was the direction we wanted to avoid.

crates/sui-types/src/sui_system_state/mod.rs
```diff
@@ -491,6 +492,23 @@ pub mod advance_epoch_result_injection {
...
+    /// This function is used to modify the result of advance_epoch transaction for testing.
+    /// If the override is set, the result will be an execution error, otherwise the original result will be returned.
+    pub fn maybe_modify_result_for<E: ExecutionErrorTrait>(
+        result: ResultWithTimings<(), E>,
+        current_epoch: EpochId,
+    ) -> ResultWithTimings<(), E> {
+        if let Some((start, end)) = OVERRIDE.with(|o| *o.borrow()) {
+            if current_epoch >= start && current_epoch < end {
+                return Err((
+                    ExecutionError::new(ExecutionErrorKind::FunctionNotFound, None).into(),
```
amnn, line 505: Let's add a TODO here -- anywhere you introduce an `into` from `ExecutionError` to your trait is a place you want to revisit.

crates/sui-core/src/authority.rs
amnn: It seems strange that you need to do this, instead of just calling one of these functions, seeing as the only difference between them is the error type, and you follow up by normalizing on error type.

Why does this function return `ExecutionError` instead of `ExecutionFailure`?

crates/sui-types/src/execution_status.rs
```diff
@@ -27,6 +29,42 @@ pub struct ExecutionFailure {
     pub command: Option<CommandIndex>,
 }
 
+impl From<ExecutionError> for ExecutionFailure {
+    fn from(value: ExecutionError) -> Self {
+        let (kind, command) = value.to_execution_status();
+        Self {
+            error: kind,
```
amnn: nit: it looks like there are still some vestigial names here:


```suggestion
        let (kind, command) = value.to_execution_failure();
        Self {
            kind,
```

crates/sui-types/src/error.rs
```diff
@@ -1157,7 +1157,24 @@ impl std::fmt::Debug for SuiError {
...
+pub(crate) type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
+
+/// A trait for execution errors that provides common methods for accessing error information and creating new errors.
+pub trait ExecutionErrorTrait:
+    From<ExecutionError> + Debug + std::error::Error + Send + Sync + Sized
+{
+    fn with_command_index(self, command: CommandIndex) -> Self;
+    fn kind(&self) -> &ExecutionErrorKind;
+    fn command(&self) -> Option<CommandIndex>;
+    fn source_ref(&self) -> Option<&(dyn std::error::Error + 'static)>;
```
amnn, line 1170: Let's add a TODO on here to get rid of it eventually -- everywhere that you need access to the error source needs to be changed so that instead of capturing the source and acting on it later, we act on the source before it gets wrapped in a `dyn Error`.

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -212,15 +213,15 @@ mod checked {
...
                     )
                 }
 
                 _ => (),
             };
 
-            let (status, command) = error.to_execution_status();
-            ExecutionStatus::new_failure(status, command)
+            let ExecutionFailure { error, command } = error.to_execution_failure();
+            ExecutionStatus::new_failure(error, command)
```
amnn, line 248: There seems to be some redundancy here -- we are unwrapping an `ExecutionFailure` presumably to pass to this constructor which just rewraps it.

sui-execution/latest/sui-adapter/src/execution_mode.rs
amnn: Is this an argument for doing the same thing to `Normal`?

sui-execution/latest/sui-adapter/src/execution_engine.rs
amnn: I don't think you want to add that constructor, at least not yet. You still need to assess each location where you added a conversion from `ExecutionError` into `impl ExecutionErrorTrait`, to see what the most appropriate constructor is and what extra context it needs to be passed.

It could be that there are errors that only have the error kind and the command index as context, but the risk of adding a generic constructor to the trait at this stage is that it becomes too easy to just fall back to it.

sui-execution/latest/sui-adapter/src/execution_engine.rs
jordanjennings-mysten: I want to refactor this in a separate PR 
`ExecutionStatus::new_failure(error: ExecutionFailure)`
to fix. After my last cleanup this became a little strange I agree.

crates/sui-types/src/error.rs
jordanjennings-mysten: I think this becomes not necessary after I've talked to brandon on how to plumb the authority.rs. Removed

sui-execution/latest/sui-adapter/src/execution_mode.rs
jordanjennings-mysten: I don't foresee a situation where normal would use more than one ExecutionErrorTrait trait type other than ExecutionFailure. It's very easy to add in later on if we see the need to change error type on Normal and gain that flexibility later. because we tie off the error type argument in sui-execution and make it concrete there, that argument doesn't have as much use. System makes sense to have flexibility because you want to select for different types within sui-execution.

crates/sui-types/src/sui_system_state/mod.rs
jordanjennings-mysten: a few more in execution_loop make sense as well.

sui-execution/src/executor.rs
```diff
@@ -45,6 +46,31 @@ pub trait Executor {
...
         trace_builder_opt: &mut Option<MoveTraceBuilder>,
+    ) -> (
+        InnerTemporaryStore,
+        SuiGasStatus,
+        TransactionEffects,
+        Vec<ExecutionTiming>,
+        Result<(), ExecutionFailure>,
+    );
+
+    /// Execution mode returns greater error information
```
tzakian: Nit: could you expand out the comment here a bit more? E.g., "should usually only be called for fullnode/non-validator execution"?

crates/sui-core/src/authority.rs
```diff
@@ -1957,7 +1957,7 @@ impl AuthorityState {
 
         #[allow(unused_mut)]
         let (inner_temp_store, _, mut effects, timings, execution_error_opt) =
-            epoch_store.executor().execute_transaction_to_effects(
+            epoch_store.executor().execute_transaction_to_effects_and_execution_error(
```
tzakian: I'm a bit confused as to why this call changed. I would have expected this to still be calling `execute_transaction_to_effects` but maybe (I probably am) missing something?

sui-execution/latest/sui-adapter/src/execution_mode.rs
```diff
@@ -69,6 +72,71 @@ pub struct Normal;
 impl ExecutionMode for Normal {
     type ArgumentUpdates = ();
     type ExecutionResults = ();
+    type Error = ExecutionFailure;
```
tzakian: Maybe this is what Ashok was getting at, but since these two are the same except for the error, thoughts on something like what you did for system and doing (ish)

```rust
pub struct Base<E>(PhantomData<fn() -> E>);

pub type Normal = Base<ExecutionFailure>;
pub type FullNode = Base<ExecutionError>;
```

crates/sui-core/src/authority.rs
tzakian: unless this is meant to be changed later on when we go full "VM says NO", but a comment to that effect would be helpful here.

(and also, in that case I'd feel much more comfortable if the different Normal and Fullnode modes were the exact same implementation but a different instantiation for the error just out of paranoia/making sure the implementations can't drift there).

jordanjennings-mysten: For now we need to just return the greater information in these areas, using this function ensures we keep the previous behavior in authority.rs because regular `execute_transaction_to_effects` will now return the lesser information. The validator and fullnode code are a bit intertwined here and I'm going to hold off on finding the right pattern to get this new run mode into

sui-execution/latest/sui-adapter/src/execution_mode.rs
jordanjennings-mysten: this is clean! I'll add it.

crates/sui-core/src/authority.rs
tzakian: That sounds fine to me. But could you enshrine the fact that this should change back to `execute_transaction_to_effects` in the future in a TODO possibly? 

Another option is what if `Normal` had the more full-featured error (and this would then call `execute_transaction_to_effects`) and then the change of the `Normal` error would be happen in the same PR as the other changes needed to support the sparser error type? Not sure how reasonable/possible this approach would be, but might be worth pondering for a second and if that may be cleaner or not.

crates/sui-types/src/error.rs
amnn: > I'm sort of with tim on this where maybe this exeuction error sandwich is not ideal due to the loops in the type graph

I am also in this camp FWIW. I would be okay to leave it as-is for this PR, but would follow-up with addressing this issue first if I could, before going on to eliminating `ExecutionError` inside the execution layer.

crates/sui-types/src/execution_status.rs
```diff
@@ -27,6 +29,38 @@ pub struct ExecutionFailure {
     pub command: Option<CommandIndex>,
 }
 
+impl From<ExecutionError> for ExecutionFailure {
+    fn from(value: ExecutionError) -> Self {
+        value.to_execution_failure()
```
amnn: Best not to depend on this here, because you are introducing a new dependency on the execution error sandwich.

sui-execution/latest/sui-adapter/src/execution_mode.rs
amnn: I think @tzakian clarified what I was asking earlier, but just to be clear:

> I don't foresee a situation where normal would use more than one ExecutionErrorTrait trait type other than ExecutionFailure

Doesn't `Normal` need to support running on a fullnode and a validator just like `System`?

sui-execution/latest/sui-adapter/src/execution_mode.rs
```diff
@@ -126,12 +133,15 @@ impl ExecutionMode for Normal {
     }
 }
 
+pub type FullNode = Normal<ExecutionError>;
```
amnn: Personally I would avoid the `FullNode` nomenclature here and just use `Normal<...>` as it seems a bit of an abstraction leak.

crates/sui-types/src/error.rs
amnn: Could you articulate where the loop comes from as well? I could sense it's there, but I don't think I'm aware of all parts of it.

jordanjennings-mysten: In a way the ExecutionError sandwich is a loop right?
(sui-types)             (execution)                    (authority)
ExecutionError -> ExecutionErrorTrait -> ExecutionError: is a loop
ExecutionError -> ExecutionErrorTrait -> ExecutionErrorWithSource: would be effectively similar in memory but the type graph wouldnt be causing a loop.

sui-execution/latest/sui-adapter/src/execution_mode.rs
jordanjennings-mysten: Take for example `ChangeEpoch` system transaction bumping the epoch up by 1 in the network.
1. Validators detect end of epoch -> ChangeEpoch tx created and broadcast -> normal mode sui-execution -> execution_loop calls into a system transaction -> system execution mode inherts normal mode's error type and executes
2. Fullnodes sync with validators -> ChangeEpoch tx is included in the checkpoint and processed -> execution_loop calls into a system transaction -> uses `System` execution mode but inherits the FullNode error type and executes `ChangeEpoch`

System needs to run on both because system run mode is used either within a `Normal` or a `FullNode` execution.

crates/sui-types/src/execution_status.rs
jordanjennings-mysten: Was a bit confused after your last comment but I agree here yup!

crates/sui-core/src/authority.rs
jordanjennings-mysten: I think a todo makes the most sense here. temporarily changing the return types feels off.

sui-execution/latest/sui-adapter/src/execution_mode.rs
jordanjennings-mysten: that said I think given your other comment a type arg works cleanest here.

sui-execution/src/v2.rs
```diff
@@ -169,6 +176,51 @@ impl executor::Executor for Executor {
...
+        metrics: Arc<LimitsMetrics>,
+        enable_expensive_checks: bool,
+        execution_params: ExecutionOrEarlyError,
+        epoch_id: &EpochId,
+        epoch_timestamp_ms: u64,
+        input_objects: CheckedInputObjects,
+        gas: GasData,
+        gas_status: SuiGasStatus,
+        transaction_kind: TransactionKind,
+        _rewritten_inputs: Option<Vec<bool>>,
```
tnowacki, line 192: I see these ignored in a few spots? Aren't these being used now?

jordanjennings-mysten: same as execute_transaction_to_effects() older versions ignore `rewritten_inputs` from what I can tell.

PR #25513, opened 2026-02-18
allow custom natives in dependents of sui-move and sui-execution

karlwuest:
## Description 

This PR exposes APIs to create an executor with custom native functions and allows to easily construct a tool to run move tests with custom native functions. We want to make use of this for research projects, where we need to use the `sui-execution` crate (but don't want to fork the full codebase) and require additional or different native functions.

I've put the new interface behind a feature flag in `sui-execution` to make it obvious that it is not a standard interface that should be used, but happy to change that.

## Test plan 

Existing tests. The PR does not change any functionality, just exposes additional interfaces.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-move/src/unit_test.rs
+++ b/crates/sui-move/src/unit_test.rs
@@ -1,6 +1,15 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::{
+    cell::RefCell,
+    collections::BTreeMap,
+    ops::{Deref, DerefMut},
+    path::Path,
+    rc::Rc,
+    sync::{Arc, LazyLock},
+};
+
 use clap::Parser;
 use move_cli::base::{
     self,
@@ -9,14 +18,8 @@ use move_cli::base::{
 use move_package_alt_compilation::build_config::BuildConfig;
 use move_unit_test::{UnitTestingConfig, vm_test_setup::VMTestSetup};
 use move_vm_config::runtime::VMConfig;
-use move_vm_runtime::natives::extensions::NativeContextExtensions;
-use std::{
-    cell::RefCell,
-    collections::BTreeMap,
-    ops::{Deref, DerefMut},
-    path::Path,
-    rc::Rc,
-    sync::{Arc, LazyLock},
+use move_vm_runtime::natives::{
+    extensions::NativeContextExtensions, functions::NativeFunctionTable,
 };
 use sui_adapter::gas_meter::SuiGasMeter;
 use sui_move_build::decorate_warnings;
@@ -57,6 +60,7 @@ impl Test {
         mut build_config: BuildConfig,
         wallet: &WalletContext,
         flavor: SuiFlavor,
+        vm_test_setup: SuiVMTestSetup,
     ) -> anyhow::Result<UnitTestResult> {
         let compute_coverage = self.test.compute_coverage;
         if !cfg!(feature = "tracing") && compute_coverage {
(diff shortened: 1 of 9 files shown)


crates/sui-move/src/unit_test.rs
```diff
@@ -101,6 +106,7 @@ pub async fn run_move_unit_tests(
     config: Option<UnitTestingConfig>,
     compute_coverage: bool,
     save_disassembly: bool,
+    vm_test_setup: SuiVMTestSetup,
```
tzakian, line 118: Another option is having `vm_test_setup: Option<SuiVMTestSetup>` and then do a `vm_test_setup.unwrap_or_else(SuiVMTestSetup::new)` just to make it easier to "do the right thing" in the calling context in the majority of cases if that makes sense. 

I _think_ I'd prefer that, but I don't have a super strong opinion (and am fine with it as-is) but it's worth mentioning as another possibility in case other folks have opinions on it.

sui-execution/src/latest.rs
```diff
@@ -1,42 +1,41 @@
...
 
+use std::{cell::RefCell, rc::Rc, sync::Arc};
+
 use move_binary_format::CompiledModule;
+use move_bytecode_verifier_meter::Meter;
 use move_trace_format::format::MoveTraceBuilder;
 use move_vm_config::verifier::{MeterConfig, VerifierConfig};
-use std::{cell::RefCell, rc::Rc, sync::Arc};
+use move_vm_runtime_latest::move_vm::MoveVM;
+#[cfg(feature = "custom-natives")]
```
tzakian, line 11: It would be helpful to understand what scenarios you want to use this in E.g., if you only intend to use this for testing, I'd prefer this to be gated to just non-prod builds (so `cfg(debug_assertions)` as I would like to minimize the amount of cfg-flags that we have if at all possible. 

Otherwise if you do need these in release builds for your purposes, it's not perfect, but I would consider gating it on the `tracing` feature flag as that can be enabled in release builds, _but_ you cannot start a node that has been compiled with that feature flag set (but not sure if this is an issue or not for your purposes).

sui-execution/src/custom_natives.rs
```diff
@@ -0,0 +1,41 @@
...
+            panic!("Custom native functions are not supported for execution versions before 3")
+        }
+        3 => Arc::new(latest::Executor::new_with_custom_natives(
+            protocol_config,
+            silent,
+            native_functions,
+        )?),
+        v => panic!("Unsupported execution version {v}"),
+    })
+}
```
tnowacki, line 41: I'm not sure I understand what this module is for. But if I am understanding things correctly, I would just put the function in the template and add feature gate the function. 

And also drop the "latest" prefix so just `executor_with_custom_natives`

crates/sui-move/src/unit_test.rs
tnowacki: +1 to an override with an Option, like we do for the config. 
I would probably just also make the NativeFunctionTable the thing you pass in here, there is no reason to pass in the whole test setup just to configure the natives

dariorussi: > I would probably just also make the NativeFunctionTable the thing you pass in here

I like that, it feels a NativeFunctionTable configuration thingy if that makes sense

karlwuest: Sounds good 👍

karlwuest:
Turning this into a draft again after offline discussion. The conclusion was that we will for now use this branch as dependency in our research codebase and the Move team will think about how to add a nice interface to specify custom native functions, s.t. we this will eventually be possible on main.

PR #25516, opened 2026-02-18
[move][move-compiler] More VM hardening

cgswords:
## Description 

Hardening fixes for the Move VM runtime identified during a security audit of the bella-ciao branch. These changes fix a gas metering issue for strings, remove potential panics from RefCell::borrow(), swap many `checked_add` calls for `safe_add`, and handle an unsafe length computation.

## Test plan 

Tests all still pass.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/cache/arena.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/cache/arena.rs
@@ -55,8 +55,8 @@ impl ArenaBuilder {
         &self,
         items: impl ExactSizeIterator<Item = T>,
     ) -> PartialVMResult<ArenaVec<T>> {
-        let size = items.len();
         if let Ok(slice) = self.0.try_alloc_slice_fill_iter(items) {
+            let size = slice.len();
             Ok(ArenaVec(unsafe {
                 std::mem::ManuallyDrop::new(Vec::from_raw_parts(slice.as_mut_ptr(), size, size))
             }))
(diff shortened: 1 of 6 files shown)


external-crates/move/crates/move-vm-runtime/src/cache/arena.rs
```diff
@@ -51,8 +51,8 @@ impl Arena {
         &self,
         items: impl ExactSizeIterator<Item = T>,
     ) -> PartialVMResult<ArenaVec<T>> {
-        let size = items.len();
         if let Ok(slice) = self.0.try_alloc_slice_fill_iter(items) {
+            let size = slice.len();
```
cgswords, line 59: Compute size _after_.

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/string.rs
```diff
@@ -198,17 +198,15 @@ fn native_index_of(
     let s_arg = pop_arg!(args, VectorRef);
     let s_ref = s_arg.as_bytes_ref()?;
     let s_str = unsafe { std::str::from_utf8_unchecked(s_ref.as_slice()) };
+    // Charge search fee upfront based on haystack length (worst-case cost)
```
cgswords, line 201: Move to before the find attempt to ensure we don't run out of gas after doing too much work.

PR #25522, opened 2026-02-19
Add protocol-config skill for safe protocol config changes

mystenmark:
## Summary

- Adds a new `/protocol-config` skill that guides safe modification of the protocol config
- The skill can be used to make changes or verify existing changes are correct
- Adds a directive to CLAUDE.md requiring invocation of this skill when modifying protocol config

The skill covers:
- How to determine the latest release branch and its max protocol version
- When a new protocol version must be created
- How to use Chain for network-specific feature rollouts (devnet -> testnet -> mainnet)
- Common patterns for adding features and constants

## Test plan

- [ ] Verify skill file renders correctly
- [ ] Test `/protocol-config` invocation

### DIFF
--- a/.claude/skills/protocol-config/SKILL.md
+++ b/.claude/skills/protocol-config/SKILL.md
@@ -100,11 +100,13 @@ cfg.some_constant = Some(new_value);
 ```rust
 cfg.new_constant = Some(new_value);
 ```
+Note: You must also add the field to the `ProtocolConfig` struct with type `Option<T>`. The `ProtocolConfigAccessors` derive macro automatically generates accessor methods (`new_constant()` and `new_constant_as_option()`).
 
 **Add a new feature flag:**
 ```rust
 cfg.feature_flags.new_feature = true;
 ```
+Note: You must also add the field to the `FeatureFlags` struct with type `bool`. The `ProtocolConfigFeatureFlagsGetters` derive macro automatically generates the getter method.
 
 ### 7. Using Chain for network-specific features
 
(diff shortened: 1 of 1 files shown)


.claude/skills/protocol-config/SKILL.md
```diff
@@ -0,0 +1,165 @@
...
+
+```bash
+cargo insta test -p sui-protocol-config --accept
+```
+
+### 9. Common mistakes to avoid
+
+- **Never modify settings for a version that exists in a release branch** - this will break consensus
+- **Always add new feature flags to FeatureFlags struct** if they don't exist
+- **Always add new config values to ProtocolConfig struct** if they don't exist
```
akichidis, line 155: Haven't tried the skill, but will this also create the corresponding accessor method in `ProtocolConfig`?

.claude/skills/protocol-config/SKILL.md
```diff
@@ -0,0 +1,165 @@
...
+- **Remember to update version history comments** - they serve as documentation
+
+## Quick Reference
+
+| Network | Chain Value | Pattern |
+|---------|-------------|---------|
+| Devnet | `Chain::Unknown` | `chain != Chain::Mainnet && chain != Chain::Testnet` |
+| Testnet | `Chain::Testnet` | `chain != Chain::Mainnet` |
+| Mainnet | `Chain::Mainnet` | `chain == Chain::Mainnet` or no condition |
```
akichidis: `chain == Chain::Mainnet` or no condition --> again not sure how exactly this will work, but we want to make sure that whenever someone wants to enable a flag in mainnet from the beginning, the skill will not do `chain == Chain::Mainnet`, but it will prioritise the no condition.

mystenmark: good call

.claude/skills/protocol-config/SKILL.md
mystenmark: I added some instructions to that effect - but mostly we will have to see how it does and update it as necessary

PR #25523, opened 2026-02-19
runtime configurable pipeline knobs

nickvikeras:
## Description

Making the handler/processor trait consts overridable with runtime-config to make it possible to tune without recompiling.


wlmyng (approved):
makes sense

Issue #25459, opened 2026-02-13
Compiler panic in HLIR pass: `unwrap()` on `assert!` with wrong number of arguments


jubnzv:
The compiler panics with `unwrap()` on `Err` in [`hlir/translate.rs:1134`](https://github.com/MystenLabs/sui/blob/1a27af4bf7ea29b0058c4d5b19ccc98f963f9b2e/external-crates/move/crates/move-compiler/src/hlir/translate.rs#L1134) when `assert!` is called with more than 2 arguments.

MRE:
```move
module 0x0::M {
    fun f() {
        assert!(true, 0, 1);
    }
}
```

Reproduce steps:
```
cd external-crates/move
cargo +nightly run --release -p move-compiler --bin move-check -- mre/mre_hlir_translate_1134.move
```

Output:
```
thread 'main' panicked at crates/move-compiler/src/hlir/translate.rs:1134:61:
called `Result::unwrap()` on an `Err` value: [Single(...), Single(...), Single(...)]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Expected behavior: a diagnostic error about wrong number of arguments to `assert!`.

Git commit: 1a27af4bf7ea29b0058c4d5b19ccc98f963f9b2e
Build: release
cargo 1.91.0-nightly (840b83a10 2025-07-30)
rustc 1.91.0-nightly (3672a55b7 2025-08-13)

lau90eth:
Fixed in #25471 

**Root cause:** `arg_list.try_into().unwrap()` panicked when `assert!` received 3+ arguments.

**Solution:** Replaced `.unwrap()` with proper error handling using `match`.

**Files changed:**
- `external-crates/move/crates/move-compiler/src/hlir/translate.rs`
- `external-crates/move/crates/move-compiler/tests/typing/bad_assert_args/`

The compiler now reports: `"assert! expects 1 or 2 arguments, but got a different number"` instead of panicking.

tnowacki:
Thanks for the report! A fix is up in https://github.com/MystenLabs/sui/pull/25525

tnowacki:
Thanks for all of the reports!

Issue #25453, opened 2026-02-13
Compiler panic in HLIR pass: `unwrap()` on `None` in `PackVariant` field lookup


jubnzv:
The compiler panics with `unwrap()` on `None` in [`hlir/translate.rs:1455`](https://github.com/MystenLabs/sui/blob/1a27af4bf7ea29b0058c4d5b19ccc98f963f9b2e/external-crates/move/crates/move-compiler/src/hlir/translate.rs#L1455) when a garbled enum variant declaration with parser-recovered fields is used in a `PackVariant` expression.

MRE:
```move
module 0x0::T {
    enum E {
        V x{: u8},
    }
    fun f(): E {
        E::V { x: 0 }
    }
}
```

Reproduce steps:
```
cd external-crates/move
cargo +nightly run --release -p move-compiler --bin move-check -- mre/mre_hlir_translate_1455.move
```

Output:
```
thread 'main' panicked at crates/move-compiler/src/hlir/translate.rs:1455:49:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Expected behavior: a parse error diagnostic about the malformed variant declaration.

Git commit: 1a27af4bf7ea29b0058c4d5b19ccc98f963f9b2e
Build: release
cargo 1.91.0-nightly (840b83a10 2025-07-30)
rustc 1.91.0-nightly (3672a55b7 2025-08-13)

lau90eth:
Fixed in #25471

Root cause: `field_map.get(&f).unwrap()` panicked when a malformed enum variant declaration (with parser-recovered fields) was used in a `PackVariant` expression.

Solution: Replaced `.map().unwrap()` with `.filter_map()` + `match` to handle the `None` case gracefully.

Files changed:
- `external-crates/move/crates/move-compiler/src/hlir/translate.rs`

The compiler now reports: "Invalid field '{field}' in enum variant constructor" instead of panicking.

tnowacki:
Thanks for the report! A fix is up in https://github.com/MystenLabs/sui/pull/25525

tnowacki:
Thanks for all of the reports!

PR #25525, opened 2026-02-19
[move-compiler] Fix Pack/PackVariant ICE. Fix `assert` arity ICE

tnowacki:
## Description 

- #25453 PackVariant could be reached in HLIR with invalid arguments if invoked with arguments (even if none were expected). As such, I made several errors non-blocking and fixed the downstream issues.
- #25459 Fixed overly tight checks on assert arity. Like with PackVariant, this could be reached in some cases even with the block errors.

## Test plan 

- Added tests
- Updated tests 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-compiler/src/hlir/translate.rs
+++ b/external-crates/move/crates/move-compiler/src/hlir/translate.rs
@@ -131,7 +131,7 @@ pub(super) struct HLIRDebugFlags {
     pub(super) eval_order: bool,
 }
 
-pub struct Context<'env> {
+pub(super) struct Context<'env> {
     pub env: &'env CompilationEnv,
     pub info: Arc<TypingProgramInfo>,
     #[allow(dead_code)]
--- a/external-crates/move/crates/move-compiler/src/shared/matching.rs
+++ b/external-crates/move/crates/move-compiler/src/shared/matching.rs
@@ -4,7 +4,7 @@
 use crate::{
     diagnostics::DiagnosticReporter,
     expansion::ast::{Fields, ModuleIdent, Mutability, Value},
-    hlir::translate::{Context, NEW_NAME_DELIM},
+    hlir::translate::NEW_NAME_DELIM,
     ice, ice_assert,
     naming::ast::{self as N, Type, Var},
     parser::ast::{BinOp_, ConstantName, Field, VariantName},
@@ -154,13 +154,14 @@ impl PatternArm {
             .all(|pat| matches!(pat.pat.value, TP::Wildcard | TP::Binder(_, _)))
     }
 
-    fn all_wild_arm(
+    fn all_wild_arm<'env>(
         &mut self,
-        context: &mut Context,
+        env: &'env CompilationEnv,
+        reporter: &mut DiagnosticReporter<'env>,
         fringe: &VecDeque<FringeEntry>,
     ) -> Option<ArmResult> {
         if self.is_wild_arm() {
-            let bindings = self.make_arm_bindings(context, fringe);
+            let bindings = self.make_arm_bindings(env, reporter, fringe);
             let PatternArm {
                 pats: _,
                 guard,
@@ -178,16 +179,17 @@ impl PatternArm {
         }
     }
 
-    fn make_arm_bindings(
+    fn make_arm_bindings<'env>(
         &mut self,
-        context: &mut Context,
+        env: &'env CompilationEnv,
+        reporter: &mut DiagnosticReporter<'env>,
         fringe: &VecDeque<FringeEntry>,
     ) -> PatBindings {
         let mut bindings = BTreeMap::new();
         // If the lengths don't match, we have an error from typing
         ice_assert!(
-            context,
-            self.pats.len() == fringe.len() || context.env.has_errors(),
+            reporter,
+            self.pats.len() == fringe.len() || env.has_errors(),
             self.arm.orig_pattern.pat.loc,
             "mismatched length should have caused an error in typing"
         );
@@ -533,21 +535,22 @@ impl PatternMatrix {
             .any(|pat| pat.is_wild_arm() && pat.guard.is_none())
     }
 
-    pub fn wild_tree_opt(
+    pub fn wild_tree_opt<'env>(
         &mut self,
-        context: &mut Context,
+        env: &'env CompilationEnv,
+        reporter: &mut DiagnosticReporter<'env>,
         fringe: &VecDeque<FringeEntry>,
     ) -> Option<Vec<ArmResult>> {
         // NB: If the first row is all wild, we need to collect _all_ wild rows that have guards
         // until we find one that does not. If we do not find one without a guard, then this isn't
         // a wild tree.
-        if let Some(arm) = self.patterns[0].all_wild_arm(context, fringe) {
+        if let Some(arm) = self.patterns[0].all_wild_arm(env, reporter, fringe) {
             if arm.guard.is_none() {
                 return Some(vec![arm]);
             }
             let mut result = vec![arm];
             for pat in self.patterns[1..].iter_mut() {
-                if let Some(arm) = pat.all_wild_arm(context, fringe) {
+                if let Some(arm) = pat.all_wild_arm(env, reporter, fringe) {
                     let has_guard = arm.guard.is_some();
                     result.push(arm);
                     if !has_guard {
--- a/external-crates/move/crates/move-compiler/src/hlir/match_compilation.rs
+++ b/external-crates/move/crates/move-compiler/src/hlir/match_compilation.rs
@@ -157,7 +157,7 @@ fn build_match_tree(
         return MatchTree::Failure;
     }
 
-    if let Some(leaf) = matrix.wild_tree_opt(context, &fringe) {
+    if let Some(leaf) = matrix.wild_tree_opt(context.env, &mut context.reporter, &fringe) {
         debug_print!(context.debug.match_specialization, (msg "wild leaf"), ("matrix" => matrix));
         return MatchTree::Leaf(leaf);
     }


external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -1649,6 +1499,76 @@ fn value(
     maybe_freeze(context, block, expected_type.cloned(), preresult)
 }
 
+// Handles fields for both Pack and PackVariant
+fn value_fields(
```
tnowacki, line 1517: @cgswords thoughts on the naming of this function?

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -1649,6 +1499,76 @@ fn value(
...
+            fields.len() == field_exps.len(),
+            "ICE exp_evaluation_order changed arity"
+        );
+        field_exps
+            .into_iter()
+            .zip(fields)
+            .map(|(e, (f, bt))| (f, bt, e))
+            .collect()
+    } else {
+        let mut fields: BTreeMap<usize, (_, _, _)> = BTreeMap::new();
```
tnowacki, line 1584: I went back and forth on doing a `BTreeMap` with `into_values` vs a `Vec` with a `sort_by_key` and a final `map` to drop the index. 

My main motivation for going with the map was to get the debug assert below
```
            debug_assert!(fields.get(&decl_idx).is_none());
```
but maybe that isn't necessary

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -2391,28 +2285,39 @@ fn assign_variant_fields(
...
+
+fn assign_fields(
+    context: &mut Context,
+    decl_fields: Option<UniqueMap<Field, usize>>,
+    tfields: Fields<(N::Type, T::LValue)>,
+) -> Vec<(Field, H::BaseType, T::LValue)> {
+    match decl_fields {
+        Some(field_map) => {
+            let field_len = field_map.len();
+            let mut fields: BTreeMap<usize, (_, _, _)> = BTreeMap::new();
```
tnowacki, line 2331: same point as above with regards to a BTreeMap vs Vec + sorting

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -1094,60 +1094,48 @@ fn value(
...
-            if matches!(&*bt, sp!(_, T::BuiltinFunction_::Assert(Some(_)))) =>
-        {
-            use T::ExpListItem as TI;
             let [cond_item, code_item]: [TI; 2] = match arguments.exp.value {
-                E::ExpList(arg_list) => arg_list.try_into().unwrap(),
+                E::ExpList(arg_list) => match arg_list.try_into() {
+                    Ok(arr) => arr,
+                    Err(_) => {
+                        // invalid call arity should be caught during typing
+                        debug_assert!(context.env.has_errors());
```
cgswords: Should we ICE in these cases?

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -1094,60 +1094,48 @@ fn value(
...
                     context.add_diag(ice!((eloc, "ICE type checking assert failed")));
                     return error_exp(eloc);
                 }
             };
-            let cond = value(context, block, Some(&tbool(eloc)), econd);
-            let mut else_block = make_block!();
-            let code = value(context, &mut else_block, None, ecode);
             let if_block = make_block!();
+            let mut else_block = make_block!();
+            let (cond, code) = if is_macro.is_some() {
```
cgswords, line 1142:
```suggestion
            // If the abort is marked as a macro, we evaluate the code only after the branch, and
            // so we use the `else_block` as its block for lowering. If it is not a macro, we instead
            // evaluate the condition and code before doing the branch test.
            let (cond, code) = if is_macro.is_some() {
```

external-crates/move/crates/move-compiler/src/hlir/translate.rs
cgswords: It's not ideal but it does match the convention, so I am happy with it as-is.

external-crates/move/crates/move-compiler/src/hlir/translate.rs
cgswords: The map is certainly going to be slower, but the assurance here is worth it IMO.

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -1649,6 +1499,76 @@ fn value(
...
+                fields.push((f, bt.clone()));
+                let t = H::Type_::base(bt);
+                (te, Some(t))
+            })
+            .collect();
+        let field_exps = value_evaluation_order(context, block, field_exps);
+        assert!(
+            fields.len() == field_exps.len(),
+            "ICE exp_evaluation_order changed arity"
+        );
```
cgswords: Maybe actually produce an ICE here instead of an abort? Though this likely has to become an `Option` return type.

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -1649,6 +1499,76 @@ fn value(
...
+                .map(|(f, (exp_idx, (bt, tf)))| {
+                    // If the field is not a valid one, typing will produce an error.
+                    // So keep the field in a consistent order, but after all of the
+                    // valid fields.
+                    let decl_idx = field_map.get(&f).copied().unwrap_or(field_len + exp_idx);
+                    (decl_idx, f, exp_idx, bt, tf)
+                })
+                .collect()
+        } else {
+            // If no field map, compiler error in typing.
```
cgswords, line 1547: Assert / ICE here?

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -2275,8 +2195,7 @@ fn assign(
             let bs = base_types(&context.reporter, &tbs);
 
             let mut fields = vec![];
-            for (decl_idx, f, bt, tfa) in assign_struct_fields(context, &m, &s, tfields) {
-                assert!(fields.len() == decl_idx);
+            for (f, bt, tfa) in assign_struct_fields(context, &m, &s, tfields) {
```
cgswords, line 2228: Letters are actually free to use when naming variables 😂

tnowacki: Not anymore! Gonna cost more for the LLM to process! (I do think the names make sense if you know what the types are :P but I hear ya)

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -2391,28 +2285,39 @@ fn assign_variant_fields(
...
+    tfields: Fields<(N::Type, T::LValue)>,
+) -> Vec<(Field, H::BaseType, T::LValue)> {
+    match decl_fields {
+        Some(field_map) => {
+            let field_len = field_map.len();
+            let mut fields: BTreeMap<usize, (_, _, _)> = BTreeMap::new();
+            for (f, (exp_idx, (tbt, tfa))) in tfields {
+                // If the field is not a valid one, typing will produce an error.
+                // So keep the field in a consistent order, but after all of the
+                // valid fields.
```
cgswords, line 2335: This is odd -- maybe just pitch the field anyway? I would also check there is an error, because this is super hinky.

external-crates/move/crates/move-compiler/src/shared/matching.rs
```diff
@@ -176,7 +176,7 @@ impl PatternArm {
 
     fn make_arm_bindings(&mut self, fringe: &VecDeque<FringeEntry>) -> PatBindings {
         let mut bindings = BTreeMap::new();
-        assert!(self.pats.len() == fringe.len());
+        // If the lengths don't match, we have an error from typing
```
cgswords, line 189: Again, possibly double-check?

external-crates/move/crates/move-compiler/tests/move_2024/naming/clever_errors_line_assert.snap
```diff
@@ -13,11 +13,8 @@ warning[W00001]: DEPRECATED. will be removed
...
-error[E04016]: too few arguments
+error[E13002]: feature is deprecated in specified edition
   ┌─ tests/move_2024/naming/clever_errors_line_assert.move:3:9
   │
 3 │         assert(false);
-  │         ^^^^^^^^^^^^^
-  │         │     │
-  │         │     Found 1 argument(s) here
-  │         Invalid call of 'assert'. The call expected 2 argument(s) but got 1
+  │         ^^^^^^^^^^^^^ 'assert' function syntax has been deprecated and cannot be used with clever assertions
```
cgswords, line 20: This is a weird error change. Maybe add a note?

cgswords (approved):
LGTM modulo some small nits. Thanks for unifying the field logic in HLIR lowering, it's a nice cleanup of the code that has me wondering why I did not do it that way in the first place.

external-crates/move/crates/move-compiler/src/hlir/translate.rs
tnowacki: I don't think there should be an ice here. This is just the case that the datatype was unbound but we keep compiling.

external-crates/move/crates/move-compiler/src/hlir/translate.rs
tnowacki: I think just an ice_assert here is fine, no need to Option--we will just drop some elements of the Vec

external-crates/move/crates/move-compiler/src/shared/matching.rs
tnowacki: I don't have a context here, nor a loc. Do you think it is worth threading that through?

external-crates/move/crates/move-compiler/tests/move_2024/naming/clever_errors_line_assert.snap
tnowacki: It is just that the second arg is now a clever assert. But I'll add the note here as well as the deprecation message

tnowacki:
> LGTM modulo some small nits. Thanks for unifying the field logic in HLIR lowering, it's a nice cleanup of the code that has me wondering why I did not do it that way in the first place.

Oh, we can't blame @tzakian? :P

PR #25527, opened 2026-02-19
[Block Streaming] introduce the Observer service and the client

akichidis:
## Description 

This is the first PR from the many follow up PRs of the block streaming work.

This PR is:
* Adding the `ObserverService`
* Renaming the `NetworkService` to `ValidatorService`
* Adding the ability to the `TonicManager` to boot the `ObserverService` on a separate port
* Introduce configuration in `Properties`

## Test plan 

CI

## Next
- [ ] https://github.com/MystenLabs/sui/pull/25528
- [ ] https://github.com/MystenLabs/sui/pull/25477
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/config/tests/snapshots/parameters_test__parameters.snap
+++ b/consensus/config/tests/snapshots/parameters_test__parameters.snap
@@ -30,5 +30,7 @@ tonic:
   connection_buffer_size: 33554432
   excessive_message_size: 16777216
   message_size_limit: 67108864
+  observer_server_port: ~
+  observer_allowlist: []
 internal:
   skip_equivocation_validation: false


consensus/core/src/network/mod.rs
```diff
@@ -172,22 +181,48 @@ pub(crate) trait NetworkService: Send + Sync + 'static {
...
 }
 
+/// Observer network service for handling requests from observer nodes.
+/// Unlike ValidatorNetworkService which uses AuthorityIndex, this uses NodeId (NetworkPublicKey)
+/// to identify peers since observers are not part of the committee.
+#[async_trait]
+pub(crate) trait ObserverNetworkService: Send + Sync + 'static {
+    /// Handles the request to fetch blocks by references from an observer peer.
+    #[allow(unused)]
+    async fn handle_fetch_blocks(
```
mwtian, line 191: IIUC this is not the method for block subscription right?

consensus/core/src/network/observer.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub(crate) command: Option<block_stream_request::Command>,
+}
+
+pub(crate) mod block_stream_request {
+    #[derive(Clone, PartialEq, prost::Oneof)]
+    pub(crate) enum Command {
+        #[prost(message, tag = "1")]
+        Start(super::StartBlockStream),
+        #[prost(message, tag = "2")]
+        Stop(super::StopBlockStream),
```
mwtian, line 28: Still not sure if `Stop` is needed, over just disconnecting. But we can keep it for now.

consensus/core/src/network/observer.rs
```diff
@@ -0,0 +1,133 @@
...
+
+#[derive(Clone, PartialEq, prost::Message)]
+pub(crate) struct StopBlockStream {}
+
+#[derive(Clone, prost::Message)]
+pub(crate) struct BlockStreamResponse {
+    #[prost(bytes = "bytes", tag = "1")]
+    pub(crate) block: Bytes,
+    #[prost(uint64, tag = "2")]
+    pub(crate) highest_commit_index: u64,
```
mwtian, line 47: Add a comment? Is this highest persisted or known or executed commit?

consensus/core/src/network/observer.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub(crate) highest_round_per_authority: Vec<u64>,
+}
+
+#[derive(Clone, PartialEq, prost::Message)]
+pub(crate) struct StopBlockStream {}
+
+#[derive(Clone, prost::Message)]
+pub(crate) struct BlockStreamResponse {
+    #[prost(bytes = "bytes", tag = "1")]
+    pub(crate) block: Bytes,
```
mwtian, line 44: Should this be Vec<Bytes>? Otherwise attaching block with `highest_commit_index` seems a bit overkill. Or that field can be optional.

consensus/core/src/network/observer.rs
```diff
@@ -0,0 +1,133 @@
...
+    #[prost(uint64, tag = "2")]
+    pub(crate) highest_commit_index: u64,
+}
+
+// Observer fetch messages
+#[derive(Clone, prost::Message)]
+pub(crate) struct FetchBlocksRequest {
```
mwtian, line 52: For FetchBlocks.* and FetchCommits.*, does it make sense to share the definitions between validator and observer services?

consensus/core/src/network/mod.rs
akichidis: exactly, the streaming method is added on the next PR

consensus/core/src/network/observer.rs
akichidis: We can certainly have the `block` as `Vec` - and in fact it will be more versatile. The `highest_commit_index` can be as optional as well, but do you think it would be - performance wise - too much of an issue?

consensus/core/src/network/observer.rs
akichidis: Although convenient yes, I believe it would be cleaner having those as separated.

consensus/core/src/network/observer.rs
akichidis: I would suggest keeping it for now, and we can remove later if needed.

consensus/core/src/network/observer.rs
mwtian: Maybe not performance, but it seems not great to resend the same info many times. You can leave a TODO to revisit this design after having client and server working e2e.

PR #25528, opened 2026-02-19
[Block Streaming] introduce the Synchronizer & CommitSyncer clients

akichidis:
## Description 

After introducing the `ObserverService` we now have multiple combinations of nodes communicating with each other:
* validator <-> validator
* validator <-> observer
* observer <-> observer

introducing the `SynchronizerClient` and the `CommitSyncerClient` to abstract away the communication so the components themselves are not dealing with the networking details and making this more transparent to them. 

## Test plan 

CI

## Next
- [ ] https://github.com/MystenLabs/sui/pull/25477

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/network/clients.rs
+++ b/consensus/core/src/network/clients.rs
@@ -1,0 +1,181 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{sync::Arc, time::Duration};
+
+use bytes::Bytes;
+use consensus_config::AuthorityIndex;
+use consensus_types::block::{BlockRef, Round};
+
+use super::{
+    ObserverNetworkClient, PeerId, ValidatorNetworkClient, observer::TonicObserverClient,
+    tonic_network::TonicValidatorClient,
+};
+use crate::{
+    commit::CommitRange,
+    context::Context,
+    error::{ConsensusError, ConsensusResult},
+};
+
+/// Concrete client implementation for synchronizer operations.
+/// Wraps validator and observer network clients and routes requests based on whether this node is
+/// a validator and the peer is an authority.
+pub(crate) struct SynchronizerClient<
+    V: ValidatorNetworkClient = TonicValidatorClient,
+    O: ObserverNetworkClient = TonicObserverClient,
+> {
+    context: Arc<Context>,
+    validator_client: Option<Arc<V>>,
+    observer_client: Option<Arc<O>>,
+}
+
+impl<V, O> SynchronizerClient<V, O>
+where
+    V: ValidatorNetworkClient,
+    O: ObserverNetworkClient,
+{
+    pub fn new(
+        context: Arc<Context>,
+        validator_client: Option<Arc<V>>,
+        observer_client: Option<Arc<O>>,
+    ) -> Self {
+        Self {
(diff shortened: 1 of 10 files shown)


consensus/core/src/network/clients.rs
```diff
@@ -0,0 +1,181 @@
...
+        &self,
+        peer: PeerId,
+        block_refs: Vec<BlockRef>,
+        highest_accepted_rounds: Vec<Round>,
+        breadth_first: bool,
+        timeout: Duration,
+    ) -> ConsensusResult<Vec<Bytes>> {
+        // A validator node will always talk via the validator interface to another authority.
+        // Otherwise, the only way to communicate with the other peer is via the observer interface.
+        if self.context.is_validator()
```
mwtian, line 59: Is this condition needed? Basically, will PeerId just be used to differentiate between talking to validator vs observer, or will it also decide if contacting the Consensus port or Observer port?

consensus/core/src/network/clients.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    pub async fn fetch_commits(
+        &self,
+        peer: PeerId,
+        commit_range: CommitRange,
+        timeout: Duration,
+    ) -> ConsensusResult<(Vec<Bytes>, Vec<Bytes>)> {
+        // A validator node will always talk via the validator interface to another authority.
+        // Otherwise, the only way to communicate with the other peer is via the observer interface.
+        if self.context.is_validator()
```
mwtian, line 134: Same comment. It seems to be simpler to just map PeerId::Validator to using validator client. Otherwise there needs to be logic in observer_client to handle or reject PeerId::Validator.

consensus/core/src/network/clients.rs
```diff
@@ -0,0 +1,181 @@
...
+        &self,
+        peer: PeerId,
+        block_refs: Vec<BlockRef>,
+        highest_accepted_rounds: Vec<Round>,
+        breadth_first: bool,
+        timeout: Duration,
+    ) -> ConsensusResult<Vec<Bytes>> {
+        // A validator node will always talk via the validator interface to another authority.
+        // Otherwise, the only way to communicate with the other peer is via the observer interface.
+        if self.context.is_validator()
```
mwtian, line 159: Same as the comment in other functions.

consensus/core/src/network/clients.rs
akichidis: Yes the `peer` it self only doesn't suffice. We need to understand to which interface to talk to. And to do that we need to identify "our" node's role. But we need both information:
* our node's type
* the peer's type (validator or observer) 

Initially I also thought that I could achieve the distinction based only on the peer type, but it's not really possible

consensus/core/src/network/clients.rs
akichidis: See my reply above

consensus/core/src/network/clients.rs
mwtian: Sounds good, we can revisit if needed.

PR #25530, opened 2026-02-19
settle early in consensus handler

alex-mysten:
## Description 

This change triggers accumulator settlement in consensus handler when split_checkpoints_in_consensus_handler is enabled, reducing the time to settlement as commits may be merged together into checkpoints with a time delay.

## Test plan 

simtest, 
[antithesis single-version 4h](https://mystenlabs.antithesis.com/report/nNZXEMyii7fjRAydQXY-bVYR/NBgTIdFNffzIVf1vN7-dXl9UeWaulmne7lP6zWyk79w.html?auth=v2.public.eyJzY29wZSI6eyJSZXBvcnRTY29wZVYxIjp7ImFzc2V0IjoiTkJnVElkRk5mZnpJVmYxdk43LWRYbDlVZVdhdWxtbmU3bFA2eld5azc5dy5odG1sIiwicmVwb3J0X2lkIjoibk5aWEVNeWlpN2ZqUkF5ZFFYWS1iVllSIn19LCJuYmYiOiIyMDI2LTAyLTI1VDE4OjU3OjMzLjczMzAzNzA2NVoifWnr5PyBxVFL4gdKXftHHyyWg-pPskx41x_AmNGsj5B0-7V67neY2jXbzWHP2kePfHs3QargiAZZDsNOr4LCMgM) - PASSED
[antithesis testnet:head protocol-config=testnet 2h](https://mystenlabs.antithesis.com/report/oe-b9odkUopZNH-C2xEaH5g2/KZ8lCe6fqNopL6zfuVdV-ig93c77D02I6NOOYNQEOxg.html?auth=v2.public.eyJzY29wZSI6eyJSZXBvcnRTY29wZVYxIjp7ImFzc2V0IjoiS1o4bENlNmZxTm9wTDZ6ZnVWZFYtaWc5M2M3N0QwMkk2Tk9PWU5RRU94Zy5odG1sIiwicmVwb3J0X2lkIjoib2UtYjlvZGtVb3BaTkgtQzJ4RWFINWcyIn19LCJuYmYiOiIyMDI2LTAyLTI1VDE4OjMwOjI2Ljk1NTc0NjY4NloifftLX-lIR1beurUcIwMzOUJhjm59ZvPNXxxQVhtC9AOKjomqfQWaowepAZu2JtgrdpOtYmyNtR5mgo3HrcjTmgo) -  PASSED

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/sui-core/src/authority/authority_per_epoch_store.rs
@@ -331,28 +331,6 @@ impl From<ExecutionIndicesWithStats> for ExecutionIndicesWithStatsV2 {
     }
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, Eq)]
-pub struct ExecutionIndicesWithStatsV2 {
-    pub index: ExecutionIndices,
-    pub height: u64,
-    pub stats: ConsensusStats,
-    pub last_checkpoint_flush_timestamp: u64,
-    // Reserved for future use.
-    pub checkpoint_seq: u64,
-}
-
-impl From<ExecutionIndicesWithStats> for ExecutionIndicesWithStatsV2 {
-    fn from(v1: ExecutionIndicesWithStats) -> Self {
-        Self {
-            index: v1.index,
-            height: v1.height,
-            stats: v1.stats,
-            last_checkpoint_flush_timestamp: 0,
-            checkpoint_seq: 0,
-        }
-    }
-}
-
 type ExecutionModuleCache = SyncModuleCache<ResolverWrapper>;
 
 // Data related to VM and Move execution and type layout
@@ -447,8 +425,10 @@ pub struct AuthorityPerEpochStore {
     pub(crate) metrics: Arc<EpochMetrics>,
     epoch_start_configuration: Arc<EpochStartConfiguration>,
 
-    /// The last checkpoint sequence number from the previous epoch. Used to derive the
-    /// first checkpoint sequence number of this epoch for the consensus handler.
+    /// The last checkpoint sequence number from the previous epoch.
+    /// Used to derive the first checkpoint sequence number of this epoch for the
+    /// consensus handler.
+    /// In epoch 0, this is checkpoint_seq 0 - our first non-genesis checkpoint is seq=1.
     previous_epoch_last_checkpoint: CheckpointSequenceNumber,
 
     /// Execution state that has to restart at each epoch change
--- a/crates/sui-core/src/consensus_handler.rs
+++ b/crates/sui-core/src/consensus_handler.rs
@@ -777,11 +777,9 @@ impl CheckpointQueue {
     }
 
     pub(crate) fn checkpoint_seq(&self) -> CheckpointSequenceNumber {
-        debug_assert!(
-            self.current_checkpoint_seq > 0,
-            "checkpoint_seq called before any checkpoint was assigned"
-        );
-        self.current_checkpoint_seq - 1
+        self.current_checkpoint_seq
+            .checked_sub(1)
+            .expect("checkpoint_seq called before any checkpoint was assigned")
     }
 }
 
--- a/crates/sui-core/src/execution_scheduler/settlement_scheduler.rs
+++ b/crates/sui-core/src/execution_scheduler/settlement_scheduler.rs
@@ -15,7 +15,7 @@ use crate::{
     execution_scheduler::funds_withdraw_scheduler::FundsSettlement,
 };
 use futures::stream::{FuturesUnordered, StreamExt};
-use mysten_common::{assert_reachable, fatal};
+use mysten_common::assert_reachable;
 use mysten_metrics::{monitored_mpsc, spawn_monitored_task};
 use parking_lot::Mutex;
 use std::sync::Arc;
@@ -262,19 +262,10 @@ impl SettlementScheduler {
         tx_index_offset: u64,
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> usize {
-        let digests = match epoch_store
+        let digests = epoch_store
             .notify_read_tx_key_to_digest(&batch_info.tx_keys)
             .await
-        {
-            Ok(digests) => digests,
-            Err(e) => {
-                fatal!(
-                    "Failed to read tx digests for settlement {:?}: {:?}",
-                    settlement_key,
-                    e
-                );
-            }
-        };
+            .expect("db error");
 
         let effects = self
             .transaction_cache_read


crates/sui-core/src/execution_scheduler/execution_scheduler_impl.rs
```diff
@@ -472,6 +547,158 @@ impl ExecutionScheduler {
         }
     }
 
+    async fn construct_and_execute_settlement_early(
```
lxfind: Could there be any code sharing between this and checkpoint builder code?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -1261,11 +1266,14 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
-                self.checkpoint_queue.lock().unwrap().last_built_timestamp();
+            let queue = self.checkpoint_queue.lock().unwrap();
+            stats_to_record.last_checkpoint_flush_timestamp = queue.last_built_timestamp();
+            stats_to_record.checkpoint_seq = queue.checkpoint_seq();
         }
 
-        state.output.record_consensus_commit_stats(stats_to_record);
+        state
+            .output
+            .record_consensus_commit_stats(stats_to_record.clone());
```
lxfind: `clone` probably not needed

crates/sui-core/src/checkpoints/mod.rs
```diff
@@ -1963,50 +1958,143 @@ impl CheckpointBuilder {
...
                 if cfg!(debug_assertions) {
-                    for tx in unsorted.iter() {
+                    for tx in roots_effects.iter() {
                         assert!(tx.transaction_digest() != &ccp_digest);
                     }
                 }
-                sorted.push(ccp_effects);
+                roots_effects.insert(0, ccp_effects);
             }
-            sorted.extend(CausalOrder::causal_sort(unsorted));
```
lxfind, line 1988: How should I reason about this change? Does that mean we no longer need the `CausalOrder` module as well?

alex-mysten: All transactions are included in the consensus commit post-MFP. Therefore the causal completion that included and causally sorted owned objects that were not part of the checkpoint transaction can be removed. I removed more of the causal completion code in af1cd6d

crates/sui-core/src/consensus_handler.rs
alex-mysten: Removed

crates/sui-core/src/execution_scheduler/execution_scheduler_impl.rs
alex-mysten: I assume you mean resolve_settlement_transactions?

I've refactored resolve_settlement_transactions - now, it just reads from transaction cache, builds the txs to compute digests, then waits for effects... construct_and_execute is responsible for a lot more.

crates/sui-core/src/consensus_handler.rs
```diff
@@ -568,146 +593,114 @@ impl CheckpointQueue {
         self.height
     }
 
-    fn push_checkpoint_roots(
+    // Estimates the total number of effects this root will produce in the checkpoint.
+    // Settlement adds a barrier tx (always) plus settlement txns (only when there are
+    // user txns beyond the CCP that produce accumulator events, i.e. tx_count > 1).
```
mystenmark: This doesn't seem right - first, if there are user transactions that don't emit accumulator events, then we won't build any settlement transactions (except the barrier) - second, there can be any number of settlement transactions.

I think the best thing is to just treat the txn limit as a limit on the number of user transactions and ignore settlements here.

crates/sui-core/src/consensus_handler.rs
```diff
@@ -568,146 +593,114 @@ impl CheckpointQueue {
...
+                && let Some(checkpoint) = self.flush_forced()
+            {
+                flushed_checkpoints.push(checkpoint);
+            }
+
+            if let Some(settlement_key) = roots.settlement_root {
+                let tx_keys: Vec<_> = roots
+                    .tx_roots
+                    .iter()
+                    .filter(|k| !matches!(k, TransactionKey::AccumulatorSettlement(..)))
```
mystenmark: I think this + a fair bit of other code would be simpler if we stopped jamming the settlement in with the other schedulables. 

- `build_chunks` can return `Vec<(Vec<Schedulable>, Option<Schedulable> /* settlement */, CheckpointHeight)>`
- `to_checkpoint_roots` can be simpler (no need to check `last()`)
- this code here also gets simpler

in fact we might want to make a `Chunk` struct to hold the tuple in the vector above - it could have a `schedulables_for_version_assignment` that could encapsulate this code:

```
        let schedulables_for_version_assignment: Vec<_> = chunked_schedulables
            .iter()
            .flat_map(|(chunk, _)| chunk.iter())
            .cloned()
            .collect();
```

which we repeat twice in `create_pending_checkpoints_v2`

crates/sui-core/src/checkpoints/mod.rs
mystenmark: Hmm, it definitely should be the case that there is no causal completion necessary. But I hadn't actually thought about whether causal sorting is unnecessary as well. In theory it should be but I don't know how to verify that.

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -466,6 +469,9 @@ pub struct AuthorityPerEpochStore {
     /// Waiters for barrier transactions. Used by execution scheduler to wait for
     /// barrier transaction (keyed by the same AccumulatorSettlement key as settlements).
     barrier_registrations: Arc<Mutex<HashMap<TransactionKey, BarrierRegistration>>>,
+
+    /// Maps settlement key to batch transaction info for early settlement.
+    settlement_batch_info: Arc<Mutex<HashMap<TransactionKey, SettlementBatchInfo>>>,
```
mystenmark: this seems to be duplicating information that is already present in the CheckpointQueue, so I think there's an  opportunity for simplification:

- Every transaction goes into a CheckpointRoots instance
- They need to be scheduled in order for the checkpoint to be built
- So why not just schedule them when we push to the CheckpointQueue, instead of returning them all the way back up?
- This way we can send settlement_scheduler exactly the info it needs to build one settlement: the set of regular roots that it is settling and its own Key + any other info (e.g. height, checkpoint seq, etc)

crates/sui-core/src/consensus_handler.rs
alex-mysten: Good point. 

We also use this for determining tx_index_offset for authenticated event settlement, which would be wrong if we have > 1 accumulator settlement.

I think settlement_scheduler needs to retain some state to be able to compute this - each settlement needs to know how many transaction effects so far are included in the current checkpoint in order to construct the subsequent settlement.

crates/sui-core/src/checkpoints/mod.rs
alex-mysten: Causal sorting re-added.

crates/sui-core/src/consensus_handler.rs
alex-mysten: Chunk struct added with helpers

crates/sui-core/src/authority/authority_per_epoch_store.rs
alex-mysten: We now enqueue all transactions and settlements associated with each Chunk in push_chunk() in CheckpointQueue.

crates/sui-core/src/consensus_handler.rs
alex-mysten: Addressed - the settlement_scheduler is now responsible for computing tx_index_offset based on the actual # of txn + settlement effects we include.

Splitting is orthogonal and just based on user txs

crates/sui-core/src/consensus_handler.rs
```diff
@@ -568,146 +681,114 @@ impl CheckpointQueue {
...
 
-        if let Some(last) = pending_checkpoints.last() {
-            self.last_built_timestamp = last.details.timestamp_ms;
-        }
+        Some(checkpoint)
+    }
 
-        pending_checkpoints
+    pub(crate) fn checkpoint_seq(&self) -> CheckpointSequenceNumber {
+        debug_assert!(
```
mystenmark: should be `assert!` - or just change to used checked_sub()

crates/sui-core/src/execution_scheduler/settlement_scheduler.rs
```diff
@@ -0,0 +1,413 @@
...
+        tx_index_offset: u64,
+        epoch_store: &Arc<AuthorityPerEpochStore>,
+    ) -> usize {
+        let digests = match epoch_store
+            .notify_read_tx_key_to_digest(&batch_info.tx_keys)
+            .await
+        {
+            Ok(digests) => digests,
+            Err(e) => {
+                fatal!(
```
mystenmark: typically we just write `.expect("db error")` in these cases

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -426,6 +425,10 @@ pub struct AuthorityPerEpochStore {
     pub(crate) metrics: Arc<EpochMetrics>,
     epoch_start_configuration: Arc<EpochStartConfiguration>,
 
+    /// The last checkpoint sequence number from the previous epoch. Used to derive the
+    /// first checkpoint sequence number of this epoch for the consensus handler.
```
mystenmark: took me a minute to figure out that this will be okay in epoch 0 - presumably it uses checkpoint seq 0 here, which is incorrect according to the comment, but will work because the first checkpoint that we should build is 1.

alex-mysten: Yes - added a comment to clarify

crates/sui-core/src/execution_scheduler/settlement_scheduler.rs
alex-mysten: Addressed.

crates/sui-core/src/consensus_handler.rs
alex-mysten: Addressed.

PR #25541, opened 2026-02-19
Add gas validation fuzz and edge-case tests for dry-run and dev-inspect

akmysti:
## Description 

Extend the transaction-fuzzer crate to test gas data handling across all
four execution modes: normal (validator), dry-run (fullnode), dev-inspect
with skip_checks=true, and dev-inspect with skip_checks=false. This
exercises the gas validation paths fixed in #25361 where (1) gas price
was accessed unchecked causing panics and (2) non-gas-coin objects as
gas in dev-inspect caused crashes.

Changes to executor.rs:
- Add `new_fullnode()` constructor that uses a keypair not in the
  committee, enabling dry-run and dev-inspect APIs
- Add `dry_run_transaction()` wrapping `AuthorityState::dry_exec_transaction`
- Add `dev_inspect_transaction()` wrapping
  `AuthorityState::dev_inspect_transaction_block`

Changes to lib.rs:
- Add `GasDataGenConfig::sender_owned_only()` which forces all gas coins
  to `AddressOwner(sender)`, needed for dev-inspect with skip_checks=true
  where ownership validation is bypassed and non-sender-owned gas objects
  would panic during execution
- Add `run_proptest_with_fullnode()` helper and refactor existing
  `run_proptest()` to share logic via `run_proptest_with_executor()`

Changes to gas_data_tests.rs:
- Add 3 new proptest fuzz tests: dry-run, dev-inspect with skip_checks,
  and dev-inspect without skip_checks
- Add 6 deterministic edge-case tests covering: valid gas data,
  gas_price=0, gas_price < RGP, gas_budget=0, gas_budget > max, and
  insufficient balance — each verified across all four execution modes
- Add helpers for constructing gas coins, gas data, and running all
  modes in a single test

## Test plan 

The gas tests cover 4 execution modes across 2 testing strategies:

### Proptest Fuzz Tests (5 total)

Randomly generate gas data (price, budget, balance, number of gas
objects, ownership) and verify no panics across execution modes:

- `test_gas_data_owned_or_immut`: sender-owned or immutable gas, normal execution
- `test_gas_data_any_owner`: any owner type, normal execution
- `test_gas_data_dry_run_fuzz`: sender-owned or immutable gas, dry-run
- `test_gas_data_dev_inspect_skip_checks_fuzz`: sender-owned only, dev-inspect (skip_checks=true)
- `test_gas_data_dev_inspect_no_skip_fuzz`: sender-owned or immutable gas, dev-inspect (skip_checks=false)

The skip_checks fuzz uses `sender_owned_only()` because when ownership
validation is bypassed, immutable/shared gas objects panic in
`temporary_store.rs` when the system tries to modify them.

### Deterministic Edge Case Tests (6 total)

Each test runs through all 4 modes (normal, dry-run, dev-inspect with
and without skip_checks) and asserts the expected error:

- `test_valid_gas_data_all_modes`: valid price/budget/balance, all modes succeed
- `test_gas_price_zero`: gas_price=0, expects "Gas price 0 under min"
- `test_gas_price_below_rgp`: gas_price=rgp-1, expects "Gas price ... under min"
- `test_gas_budget_zero`: gas_budget=0, expects "Gas budget 0 is below min"
- `test_gas_budget_exceeds_max`: gas_budget > max_tx_gas, expects "Gas budget ... is above max"
- `test_gas_balance_insufficient`: balance=1, expects "balance of ... is lower than the needed"

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/transaction-fuzzer/src/executor.rs
+++ b/crates/transaction-fuzzer/src/executor.rs
@@ -10,12 +10,14 @@ use sui_core::authority::AuthorityState;
 use sui_core::authority::authority_test_utils::submit_and_execute;
 use sui_core::authority::test_authority_builder::TestAuthorityBuilder;
 use sui_move_build::BuildConfig;
-use sui_types::base_types::ObjectID;
+use sui_types::base_types::{ObjectID, ObjectRef, SuiAddress};
+use sui_types::crypto::get_authority_key_pair;
+use sui_types::digests::TransactionDigest;
 use sui_types::effects::{TransactionEffects, TransactionEffectsAPI};
 use sui_types::error::SuiError;
 use sui_types::execution_status::{ExecutionFailureStatus, ExecutionStatus};
 use sui_types::object::Object;
-use sui_types::transaction::{Transaction, TransactionData};
+use sui_types::transaction::{Transaction, TransactionData, TransactionKind};
 use sui_types::utils::to_sender_signed_transaction;
 use tokio::runtime::Runtime;
 
@@ -71,6 +73,20 @@ impl Executor {
     pub fn new() -> Self {
         let rt = Runtime::new().unwrap();
         let state = rt.block_on(TestAuthorityBuilder::new().build());
+        Self {
+            state,
+            rt: Arc::new(rt),
+        }
+    }
+
+    pub fn new_fullnode() -> Self {
+        let rt = Runtime::new().unwrap();
+        let fullnode_key_pair = get_authority_key_pair().1;
+        let state = rt.block_on(
+            TestAuthorityBuilder::new()
+                .with_keypair(&fullnode_key_pair)
+                .build(),
+        );
         Self {
             state,
             rt: Arc::new(rt),
@@ -148,5 +164,36 @@ impl Executor {
         txn.into_iter()
             .map(|txn| self.execute_transaction(txn))
             .collect()
+    }
+
+    pub fn dry_run_transaction(&self, tx_data: TransactionData) -> Result<(), SuiError> {
+        let digest = TransactionDigest::random();
+        self.rt
+            .block_on(self.state.dry_exec_transaction(tx_data, digest))
+            .map(|_| ())
+    }
+
+    pub fn dev_inspect_transaction(
+        &self,
+        sender: SuiAddress,
+        kind: TransactionKind,
+        gas_price: Option<u64>,
+        gas_budget: Option<u64>,
+        gas_sponsor: Option<SuiAddress>,
+        gas_objects: Option<Vec<ObjectRef>>,
+        skip_checks: Option<bool>,
+    ) -> Result<(), SuiError> {
+        self.rt
+            .block_on(self.state.dev_inspect_transaction_block(
+                sender,
+                kind,
+                gas_price,
+                gas_budget,
+                gas_sponsor,
+                gas_objects,
+                None,
+                skip_checks,
+            ))
+            .map(|_| ())
     }
 }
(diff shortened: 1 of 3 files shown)


crates/transaction-fuzzer/src/executor.rs
```diff
@@ -149,4 +165,35 @@ impl Executor {
...
     }
+
+    pub fn dry_run_transaction(&self, tx_data: TransactionData) -> Result<(), SuiError> {
+        let digest = TransactionDigest::random();
+        self.rt
+            .block_on(self.state.dry_exec_transaction(tx_data, digest))
+            .map(|_| ())
+    }
+
+    pub fn dev_inspect_transaction(
```
tzakian, line 176: This is great. Could you also add a similar method for calling into `simulate_transaction` as well (https://github.com/MystenLabs/sui/blob/main/crates/sui-core/src/authority.rs#L2390).

Not needed for this PR, but would be good to have in a followup PR.

crates/transaction-fuzzer/src/lib.rs
```diff
@@ -51,11 +52,19 @@ fn generate_random_gas_data(
     let num_gas_objects = gas_coin_owners.len();
     let gas_coin_owners = gas_coin_owners
         .iter()
-        .map(|o| match o {
-            Owner::ObjectOwner(_) | Owner::AddressOwner(_) if owned_by_sender => {
+        .map(|o| {
+            if force_all_sender_owned {
```
tzakian, line 56: Not for this PR (but for a followup PR): we should add the ability to generate transactions with address balance gas payments as well.

crates/transaction-fuzzer/tests/gas_data_tests.rs
```diff
@@ -41,6 +156,49 @@ fn test_with_random_gas_data(
...
+) -> Result<(), TestCaseError> {
+    let gas_data = gas_data_test.gas_data;
+    let objects = gas_data_test.objects;
+    let sender = gas_data_test.sender_key.public().into();
+
+    executor.add_objects(&objects);
+    let kind = make_transfer_sui_pt();
+    let tx_data = TransactionData::new_with_gas_data(kind, sender, gas_data);
+
+    let result = executor.dry_run_transaction(tx_data);
```
tzakian, line 171: Question: Is this test purely testing for panic freedom? Or should we be doing something with the result here?

crates/transaction-fuzzer/tests/gas_data_tests.rs
```diff
@@ -41,6 +156,49 @@ fn test_with_random_gas_data(
...
+) -> Result<(), TestCaseError> {
+    let gas_data = gas_data_test.gas_data;
+    let objects = gas_data_test.objects;
+    let sender: SuiAddress = gas_data_test.sender_key.public().into();
+
+    executor.add_objects(&objects);
+    let kind = make_transfer_sui_pt();
+    let gas_refs: Vec<ObjectRef> = gas_data.payment.clone();
+
+    let result = executor.dev_inspect_transaction(
```
tzakian, line 189: Similar question here.

crates/transaction-fuzzer/tests/gas_data_tests.rs
```diff
@@ -58,3 +216,199 @@ fn test_gas_data_any_owner() {
...
+}
+
+// ---------------------------------------------------------------------------
+// Deterministic edge case tests
+// ---------------------------------------------------------------------------
+
+/// Valid gas data succeeds in all execution modes.
+#[test]
+#[cfg_attr(msim, ignore)]
+fn test_valid_gas_data_all_modes() {
```
tzakian, line 257: Not needed for this PR, but extending these to be proptests, and using `prop_oneof!` to make sure the specific edge cases are covered would be great.

crates/transaction-fuzzer/tests/gas_data_tests.rs
akmysti: Yes, for fuzz testing this is what I had in mind. The results may or maynot be valid as long as we dont panic somewhere in between.

PR #25542, opened 2026-02-19
Concurrent pipeline rate limiter

nickvikeras:
## Description

Introducing a rate limiter for concurrent pipelines. For bigtable we need to be able to create small batches really fast, but not _too_ fast. This let's us hit the sweet spot.

Wasn't sure if it really made sense for sequntial pipelines but we can hook it in there too if there is a use case.

## Test plan

I was finally able to hover at a 60% p100 cpu utilization on my test cluster.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
@@ -1,10 +1,13 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::num::NonZeroU32;
 use std::sync::Arc;
 use std::time::Duration;
 
 use backoff::ExponentialBackoff;
+use governor::Quota;
+use governor::RateLimiter;
 use sui_futures::service::Service;
 use sui_futures::stream::Break;
 use sui_futures::stream::TrySpawnStreamExt;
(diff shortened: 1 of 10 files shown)


amnn:
Hey @nickvikeras, please don't forget to add release notes when adding new parameters to the indexing framework. Also cc @jessiemongeon1, because I think if the notes are added after land, the release-notes channel doesn't get pinged.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
```diff
@@ -54,6 +63,15 @@ pub(super) fn committer<H: Handler + 'static>(
             &metrics.latest_partially_committed_checkpoint,
         );
 
+        let rate_limiter: Option<Arc<SharedRateLimiter>> = config.max_rows_per_second.map(|rps| {
+            let burst = NonZeroU32::new(rps as u32).expect("max_rows_per_second must be > 0");
```
amnn, line 67: It seems like it's completely possible for someone to pass a `0` value and cause this line to panic, which is not ideal. Couple of options:

- Use `NonZeroU32` as the type all the way up in the config layer. `serde` and `clap` both support validating for zero when parsing `NonZero*` types (preferred).
- Use `0` as the sentinel value meaning "no rate limit" (less ideal).

crates/sui-indexer-alt-framework/src/pipeline/concurrent/committer.rs
amnn, line 1: I did have a thought about whether the pipeline level is the right granularity for sharing the rate limiter (vs trying to apply a rate limit at the store level). I settled on "yes, it's the best place" given the fact that someone can always split up an indexer to run over multiple instance that each have their own store. We're not preventing anything by pulling things up into the store, and it would introduce complexity too.

PR #25544, opened 2026-02-19
Support object funds in simulate

lxfind:
## Description 

We check object funds balance after execution in normal execution, if not sufficient, we basically abort the execution and relies on scheduler to reschedule the transaction again with insufficient balance early error. This however doesn't work for `_simulate`, because we don't have a chance to re-schedule it there and need to return a result right away.
This PR introduces a new version of dev inspect in the executor for `_simulate`, so that it checks the object balance, and if not sufficient, re-execute right away with the early error.

Let me know if this change makes sense, or if you prefer a different way to support object funds in `_simulate`.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/address_balance_tests.rs
+++ b/crates/sui-e2e-tests/tests/address_balance_tests.rs
@@ -3201,3 +3201,127 @@ async fn test_funds_withdraw_scheduler_type_alternation() {
         );
     }
 }
+
+#[sim_test]
+async fn test_simulate_address_funds_sufficient() {
+    let mut test_env = TestEnvBuilder::new()
+        .with_proto_override_cb(Box::new(|_, mut cfg| {
+            cfg.create_root_accumulator_object_for_testing();
+            cfg.enable_accumulators_for_testing();
+            cfg
+        }))
+        .build()
+        .await;
+
+    let sender = test_env.get_sender(0);
+    test_env.fund_one_address_balance(sender, 1000).await;
+
+    let tx = test_env
+        .tx_builder(sender)
+        .transfer_sui_to_address_balance(FundSource::address_fund(), vec![(500, dbg_addr(2))])
+        .build();
+
+    let result = test_env
+        .cluster
+        .grpc_client()
+        .simulate_transaction(&tx, false)
+        .await
+        .unwrap();
+    assert!(result.transaction.effects.status().is_ok());
+}
+
+#[sim_test]
+async fn test_simulate_address_funds_insufficient() {
+    let mut test_env = TestEnvBuilder::new()
+        .with_proto_override_cb(Box::new(|_, mut cfg| {
+            cfg.create_root_accumulator_object_for_testing();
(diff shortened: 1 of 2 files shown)


crates/sui-types/src/inner_temporary_store.rs
```diff
@@ -48,6 +50,36 @@ pub struct InnerTemporaryStore {
...
+    pub fn has_sufficient_object_funds_for_simulate(
+        &self,
+        address_funds: &BTreeSet<AccumulatorObjId>,
+        resolver: &dyn ChildObjectResolver,
+    ) -> bool {
+        for (account_id, max_withdraw) in &self.accumulator_running_max_withdraws {
+            if address_funds.contains(account_id) {
+                continue;
+            }
+            let balance = AccumulatorValue::load_by_id::<U128>(resolver, None, *account_id)
```
mystenmark: use `AccountFundsRead::get_latest_account_amount` or `AccountFundsRead::check_amounts_available` please

crates/sui-core/src/authority.rs
```diff
@@ -2443,12 +2493,15 @@ impl AuthorityState {
         )
         .expect("Creating an executor should not fail here");
 
+        let address_funds: BTreeSet<_> = transaction
+            .process_funds_withdrawals_for_execution(epoch_store.get_chain_identifier())
+            .into_keys()
+            .collect();
```
mystenmark: while we're at it we should probably check if the address funds are available as well, right?

crates/sui-types/src/gas.rs
```diff
@@ -58,7 +58,7 @@ pub mod checked {
 
     /// Version aware enum for gas status.
     #[enum_dispatch(SuiGasStatusAPI)]
-    #[derive(Debug)]
+    #[derive(Debug, Clone)]
```
tnowacki: This makes me scared. ~~Where are you cloning it?~~ EDIT: Found it

sui-execution/src/executor.rs
```diff
@@ -102,3 +105,77 @@ pub trait Executor {
...
+    address_funds: &BTreeSet<AccumulatorObjId>,
+) -> (
+    InnerTemporaryStore,
+    SuiGasStatus,
+    TransactionEffects,
+    Result<Vec<ExecutionResult>, ExecutionError>,
+) {
+    let cloned_input_objects = input_objects.clone();
+    let cloned_gas = gas.clone();
+    let cloned_gas_status = gas_status.clone();
```
tnowacki: I do not like having the ability to close a gas status conceptually, as 99.9999% (maybe some more 9s) of the time, it is the wrong thing to do. I'm not sure if @tzakian has thoughts on alternatives here. 

Since it is for dev inspect, could you just pass in two different gas statuses?

Issue #25472, opened 2026-02-15
Compiler panic in HLIR pass: `assert!(idx == 0)` in `type_at_index` for macro with mismatched return type


jubnzv:
The compiler panics with `assert!(idx == 0)` in [`hlir/ast.rs:825`](https://github.com/MystenLabs/sui/blob/1a27af4bf7ea29b0058c4d5b19ccc98f963f9b2e/external-crates/move/crates/move-compiler/src/hlir/ast.rs#L825) when a macro with a declared tuple return type has a garbled body that produces a single value.

MRE:
```move
module 0x42::M {
    macro fun f($x: u64): (u64, u64) {
       ($, $x)
    }
    fun test() {
        let (_a, _b) = f!(1);
    }
}
```

Reproduce steps:
```
cd external-crates/move
cargo +nightly run --release -p move-compiler --bin move-check -- mre/mre_hlir_ast_825.move
```

Output:
```
thread 'main' panicked at crates/move-compiler/src/hlir/ast.rs:825:17:
assertion failed: idx == 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Expected behavior: a diagnostic error about the malformed macro body or type mismatch.

Git commit: 1a27af4bf7ea29b0058c4d5b19ccc98f963f9b2e
Build: release
cargo 1.91.0-nightly (840b83a10 2025-07-30)
rustc 1.91.0-nightly (3672a55b7 2025-08-13)

PR #25546, opened 2026-02-20
[move-compiler] Fix two crashes with macros

### LINKED ISSUE #25456: Compiler panic in naming pass: `debug_assert` fires for `$`-prefixed identifier in `let mut`

tnowacki:
## Description 

- Fixes #25456 by simply skipping over the binding
- #25472 is a bit trickier in that it is a bit convoluted. And results from the edge case of this sort of `UnresolvedError` showing up in `make_bindings` for named blocks (and presumably other control flow structures)

## Test plan 

- New tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-compiler/src/hlir/translate.rs
+++ b/external-crates/move/crates/move-compiler/src/hlir/translate.rs
@@ -2234,22 +2234,18 @@ fn make_assignments(
     let mut lvalues = vec![];
     let mut after = Block::new();
     for (idx, a) in assigns.into_iter().enumerate() {
-        let error_ty;
-        let a_ty = match rvalue.ty.value.type_at_index(idx) {
-            Some(ty) => ty,
-            None => {
-                // we can only get here if the rvalue was malformed and somewhere has an unresolved
-                // error
-                ice_assert!(
-                    context,
-                    context.env.has_errors(),
-                    a.loc,
-                    "Unable to find a type for assignment lvalue from the rvalue"
-                );
-                error_ty = error_single_type(a.loc);
-                &error_ty
-            }
-        };
+        let error_ty = error_single_type(a.loc);
+        let a_ty = rvalue.ty.value.type_at_index(idx).unwrap_or_else(|| {
+            // we can only get here if the rvalue was malformed and somewhere has an unresolved
+            // erro
+            ice_assert!(
+                context,
+                context.env.has_errors(),
+                a.loc,
+                "Unable to find a type for assignment lvalue from the rvalue"
+            );
+            &error_ty
+        });
         let (ls, mut af) = assign(context, case, a, a_ty);
 
         lvalues.push(ls);


external-crates/move/crates/move-compiler/src/naming/translate.rs
```diff
@@ -4039,18 +4039,26 @@ fn lvalue(
...
-                    );
-                    let msg = format!(
-                        "Cannot assign to argument for parameter '{}'. \
-                        Arguments must be used in value positions",
-                        v.0
-                    );
-                    let mut diag = diag!(TypeSafety::CannotExpandMacro, (loc, msg));
-                    diag.add_note(ASSIGN_SYNTAX_IDENTIFIER_NOTE);
-                    context.add_diag(diag);
+                    match case {
```
tnowacki, line 4042: This fix seems simple enough

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -2230,7 +2234,17 @@ fn make_assignments(
...
+        let a_ty = match rvalue.ty.value.type_at_index(idx) {
+            Some(ty) => ty,
+            None => {
+                // we can only get here if the rvalue was malformed and somewhere has an unresolved
+                // error
+                debug_assert!(context.env.has_errors());
+                error_ty = error_single_type(a.loc);
+                &error_ty
+            }
+        };
```
tnowacki: This seemed like the easiest fix, and maybe the best fix? 
Otherwise we would have to a lot of massaging for `make_binders` and `bind_value_in_block`. And since we must already be in an error, might as well keep propagating

cgswords: 😢 I see _why_ but damn this is awful. Is this any better?
```
  let error_ty = error_single_type(a.loc);
  let a_ty = rvalue.ty.value.type_at_index(idx).unwrap_or_else(|| {
      debug_assert!(context.env.has_errors());
      &error_ty
  });
```

cgswords: We could also consider just bailing here, putting nothing else on the block (or an error expression) and returning from this function. It might miss some further errors that calling `assign` on the subsequent argument will turn up, but at this point the program already has errors and it's already a mess.

external-crates/move/crates/move-compiler/src/naming/translate.rs
cgswords: Is there a test case to trip this ICE?

external-crates/move/crates/move-compiler/src/hlir/translate.rs
tnowacki: I can do that if you want, sure.

PR #25547, opened 2026-02-20
Clean up collector loop

nickvikeras:
## Description

<https://github.com/MystenLabs/sui/pull/25498#discussion_r2829310537>

### DIFF
--- a/crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
@@ -103,41 +103,62 @@ pub(super) fn collector<H: Handler + 'static>(
         let reader_lo_atomic = main_reader_lo.wait().await;
 
         loop {
-            let mut force_flush = false;
             // === IDLE: block until timer fires or enough data accumulates ===
             tokio::select! {
-                _ = poll.tick() => {
-                    // Timer: always flush (even if empty, for watermark progress)
-                    force_flush = true;
-                }
+                biased;
 
                 // docs::#collector (see docs/content/guides/developer/advanced/custom-indexer.mdx)
-                Some(indexed) = rx.recv(), if pending_rows < max_pending_rows => {
-                    pending_rows += receive_checkpoint::<H>(
-                        indexed, &mut pending, reader_lo_atomic, &metrics,
-                    );
+                Some(mut indexed) = rx.recv(), if pending_rows < max_pending_rows => {
+                    let reader_lo = reader_lo_atomic.load(Ordering::Relaxed);
+
+                    metrics
+                        .collector_reader_lo
+                        .with_label_values(&[H::NAME])
+                        .set(reader_lo as i64);
+
+                    let mut recv_cps = 0usize;
+                    let mut recv_rows = 0usize;
+                    loop {
+                        if indexed.checkpoint() < reader_lo {
+                            indexed.values.clear();
+                            metrics
+                                .total_collector_skipped_checkpoints
+                                .with_label_values(&[H::NAME])
+                                .inc();
+                        }
+
+                        recv_cps += 1;
+                        recv_rows += indexed.len();
+                        pending_rows += indexed.len();
+                        pending.insert(indexed.checkpoint(), indexed.into());
+
+                        if pending_rows >= max_pending_rows {
+                            break;
+                        }
+
+                        match rx.try_recv() {
+                            Ok(next) => indexed = next,
+                            Err(_) => break,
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
```diff
@@ -443,9 +397,6 @@ mod tests {
 
         let batch2 = recv_with_timeout(&mut collector_rx, Duration::from_secs(1)).await;
         assert_eq!(batch2.batch_len, 1);
-
-        let batch3 = recv_with_timeout(&mut collector_rx, Duration::from_secs(1)).await;
-        assert_eq!(batch3.batch_len, 0);
```
nickvikeras, line 448: The collector only sends empty batches that actually have watermarks to update now. Empty batches with no watermarks are never sent, hence this expectation is deleted.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
```diff
@@ -751,6 +751,10 @@ mod tests {
         let config = ConcurrentConfig {
             committer: CommitterConfig {
                 write_concurrency: 1, // Single committer for deterministic blocking
+                // MIN_EAGER_ROWS is 1000 and MAX_PENDING_ROWS is 4, so
+                // this test relies on the collect interval tip to force
+                // batch flushing.
+                collect_interval_ms: 10,
```
nickvikeras, line 757: I have been trying to figure this out for way too long, but I think this test has always had a race condition. MIN_EAGER_ROWS is effectively disabled and the collect interval is 500ms, so it's possible that processor channel fills before the timer ticks and the initial sender below gets blocked for longer than 100ms. 

Reducing this to 10ms means the collector will always run within the tests expected timeouts.

I cannot for the life of me figure out why this started failing more frequently with this specific change, but I chalk it up to "scheduling differences" I guess.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/collector.rs
```diff
@@ -103,105 +103,126 @@ pub(super) fn collector<H: Handler + 'static>(
...
-                    };
-
-                    if tx.send(batched_rows).await.is_err() {
-                        info!(
-                            pipeline = H::NAME,
-                            "Committer closed channel, stopping collector"
-                        );
-                        return Ok(());
+                    if pending_rows < min_eager_rows {
+                        continue;
```
amnn, line 155: I think you want to do ths check after your flushing loop -- there may be more checkpoints immediately available, waiting behind the one that woke the future.

amnn: More extreme: You could make this a biased `select` that favours receiving, and pull the flushing loop into the receive arm -- that way you no longer need `receive_checkpoint` -- you could inline it into the match arm, and also amortize the read of `reader_lo` metric updates etc.:

```rust
tokio::select! {
    biased;

    Some(mut indexed) = rx.recv(), if pending_rows < max_pending_rows => {
        let reader_lo = reader_lo.load(Ordering::Relaxed);

        metrics
            .collector_reader_lo
            .with_label_values(&[H::NAME])
            .set(reader_lo as i64);

        let mut recv_cps = 0;
        let mut recv_rows = 0;
        while pending_rows < max_pending_rows {
            if indexed.checkpoint() < reader_lo {
                indexed.values.clear();
                metrics
                    .total_collector_skipped_checkpoints
                    .with_label_values(&[H::NAME])
                    .inc();
            }

            recv_cps += 1;
            recv_rows += indexed.len();
            pending_rows += indexed.len();
            pending.insert(indexed.checkpoint(), indexed.into());

            if let Ok(next) = rx.try_recv() {
                indexed = next;
            } else {
                break;
            }
        }

        metrics
            .total_collector_rows_received
            .with_label_values(&[H::NAME])
            .inc_by(recv_rows as u64);
        metrics
            .total_collector_checkpoints_received
            .with_label_values(&[H::NAME])
            .inc_by(recv_cps as u64);

        if pending_rows < min_eager_rows {
            continue;
        }
    }

    // Always flush (even if there are no rows, for watermark progress)
    _ = poll.tick() => {}
}
```

amnn: ~~Nick tested this and it's noticeably slower, which is interesting! Worth digging into why but later, no need to block this change going out.~~

Turns out it was faster when the database wasn't further bottlenecked, go figure...

PR #25549, opened 2026-02-20
[Block Streaming] broadcasting the accepted blocks

akichidis:
## Description 

A node (either validator or observer) when accepts a block, it broadcasts it so any Observer subscriber can stream the blocks. Also the PR introduces the `ObserverService` which implements the `ObserverNetworkService` which will be responsible for serving the corresponding endpoints.

## Test plan 

CI

## Next

- https://github.com/MystenLabs/sui/pull/25939

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/core.rs
+++ b/consensus/core/src/core.rs
@@ -1380,10 +1380,8 @@ impl CoreSignals {
     /// need to acquire the dag_state lock on every delivery.
     /// Silently drops the send when there are no active observer subscribers.
     pub(crate) fn new_accepted_block(&self, block: VerifiedBlock, commit_index: CommitIndex) {
-        if self.context.committee.size() > 1 {
-            // Ignoring send errors here: it is normal for there to be no observers subscribed.
-            let _ = self.tx_accepted_block_broadcast.send((block, commit_index));
-        }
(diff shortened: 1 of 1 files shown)


consensus/core/src/core.rs
```diff
@@ -1337,6 +1374,18 @@ impl CoreSignals {
         Ok(())
     }
 
+    /// Broadcasts a block that has been accepted into the local DAG to any active observer
+    /// subscribers. Unlike `new_block()`, this covers blocks from all authorities, not just
+    /// own proposals. The current commit index is bundled with each block so receivers do not
+    /// need to acquire the dag_state lock on every delivery.
+    /// Silently drops the send when there are no active observer subscribers.
+    pub(crate) fn new_accepted_block(&self, block: VerifiedBlock, commit_index: CommitIndex) {
+        if self.context.committee.size() > 1 {
```
mwtian: It seems a 1-validator can still have observers.

akichidis: hhm good point, let me change that

PR #25560, opened 2026-02-20
Coin reservation execution support and gas smashing

mystenmark:
## Summary
- Add core execution infrastructure for coin reservations (fake coins)
- Implement gas smashing when transactions use coin reservation objects as gas payment
- Add validation to reject non-SUI coin reservations as gas
- Allow transactions with no input objects (needed for coin reservation gas)
- Add test infrastructure for multiple gas objects and balance tracking

**Stack:** 1/4

## Test plan
- Backward compatibility layer tests pass
- Gas smashing tests with real/fake coin combinations
- Coin reservation validation tests

### DIFF
--- a/sui-execution/latest/sui-adapter/src/gas_charger.rs
+++ b/sui-execution/latest/sui-adapter/src/gas_charger.rs
@@ -304,7 +304,7 @@ pub mod checked {
             self.smash_gas(tx_ctx, temporary_store);
         }
 
-        fn should_write_gas_coin(&self, gas_coin: Option<&Object>) -> bool {
+        fn should_write_gas_coin(&self, gas_coin: &Object) -> bool {
             if let PaymentMethod::Metered {
                 primary_coin,
                 address_balance_payer,
@@ -312,7 +312,7 @@ pub mod checked {
             } = &self.payment_method
             {
                 primary_coin.is_some()
-                    || gas_coin.unwrap().owner() != &Owner::AddressOwner(*address_balance_payer)
+                    || gas_coin.owner() != &Owner::AddressOwner(*address_balance_payer)
             } else {
                 false
             }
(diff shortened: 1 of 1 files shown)


crates/sui-core/src/accumulators/transaction_rewriting.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use sui_types::base_types::SuiAddress;
+use sui_types::coin_reservation::{CoinReservationResolverTrait, ParsedObjectRefWithdrawal};
+use sui_types::digests::ChainIdentifier;
+use sui_types::transaction::{CallArg, ObjectArg, ProgrammableTransaction, TransactionKind};
+
+pub fn rewrite_transaction_for_coin_reservations(
```
alex-mysten, line 9: Presuming we rewrite a ptb, what digest would we return, the original one or rewritten? would we run into problems with idempotency?

crates/sui-core/src/accumulators/transaction_rewriting.rs
```diff
@@ -0,0 +1,54 @@
...
+    chain_identifier: ChainIdentifier,
+    coin_reservation_resolver: &dyn CoinReservationResolverTrait,
+    sender: SuiAddress,
+    pt: &mut ProgrammableTransaction,
+) -> Option<Vec<bool>> {
+    if pt.coin_reservation_obj_refs().count() == 0 {
+        return None;
+    }
+
+    let mut compat_args = Vec::with_capacity(pt.inputs.len());
```
alex-mysten, line 38: Nit: compat_args naming isn't super clear to me. Maybe rewritten_inputs would work better?

crates/sui-core/src/authority.rs
```diff
@@ -1852,6 +1905,67 @@ impl AuthorityState {
         );
     }
 
+    /// Helper function that handles transaction rewriting for coin reservations and executes
+    /// the transaction to effects. Returns the execution results along with the command offset
+    /// used for rewriting failure command indices.
+    fn execute_transaction_with_rewriting(
```
alex-mysten, line 1896: should we push rewriting logic down into execute_transaction_to_effects?

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -374,22 +393,62 @@ mod checked {
...
-                object_id: obj_ref.0,
-                version: Some(obj_ref.1),
-            })?;
-            gas_objects.push(obj);
+            if let Ok(parsed) = ParsedDigest::try_from(obj_ref.2) {
+                // Verify that the coin reservation is for SUI.
+                // Gas payments must be in SUI, not other coin types.
+                let unmasked_id = mask_or_unmask_id(obj_ref.0, chain_identifier);
+                if unmasked_id != *expected_sui_accumulator_id {
+                    return Err(UserInputError::GasObjectNotOwnedObject {
```
alex-mysten, line 416: Curious about why this error is the best semantic match for a non-sui coin reservation

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -374,22 +393,62 @@ mod checked {
...
         // We reserve gas budget in advance
         if !gas_paid_from_address_balance {
             if gas_ownership_checks {
                 gas_status.check_gas_objects(&gas_objects)?;
             }
-            gas_status.check_gas_data(&gas_objects, gas_budget)?;
+            gas_status.check_gas_balance(
+                &gas_objects,
+                gas_budget,
+                available_address_balance_gas,
```
alex-mysten, line 451: can we just pass in `coin_reservation_total` as this is already gated by `!gas_paid_from_address_balance`

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -398,6 +457,8 @@ mod checked {
     /// that they are all the correct version and number.
     #[instrument(level = "trace", skip_all)]
     fn check_objects(transaction: &TransactionData, objects: &InputObjects) -> UserInputResult<()> {
+        // Note: objects may be empty when address balance gas payments are used.
```
alex-mysten, line 461: should we explicitly gate this such that objects can only be empty when address balance gas payments are used?

crates/sui-types/src/accumulator_event.rs
```diff
@@ -63,7 +63,7 @@ impl AccumulatorEvent {
 
         let accumulator_address = AccumulatorAddress::new(address, balance_type);
 
-        let (operation, amount) = if net_change > 0 {
+        let (operation, amount) = if net_change < 0 {
```
alex-mysten, line 66: this makes more sense

crates/sui-types/src/accumulator_event.rs
```diff
@@ -63,7 +63,7 @@ impl AccumulatorEvent {
...
             (AccumulatorOperation::Split, net_change as u64)
```
alex-mysten: if we have (-100 as u64) - i think we'd have some garbage positive integer. We'd need to negate first - right?

why would this not fail tests?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -50,8 +50,25 @@ pub mod checked {
...
+
+        // Note: in both of the following cases the `coins` vec only contains real gas coins.
+
+        // We are smashing into a real gas coin (the first in the `coins` vec.) Because of the
+        // compatibility layer, there may be coin reservations that need to be smashed into
+        // primary coin - if so the sum of their values is in `sui_from_address_balance`.
+        SmashIntoCoin {
+            gas_coins: NonEmpty<ObjectRef>,
+            address_balance_payer: SuiAddress,
+            available_address_balance_gas: u64,
```
alex-mysten, line 55: we also use this if the payer does not have a balance?

crates/sui-core/src/accumulators/transaction_rewriting.rs
mystenmark: The rewriting happens just before execution - the signed tx is not altered.

crates/sui-core/src/authority.rs
mystenmark: I'd prefer not to, because the safety argument here is that we are only producing a PTB that the user themselves could have constructed in the first place. Once we start pushing things into the execution layer it becomes less clear that we aren't bypassing any safety checks.

crates/sui-types/src/accumulator_event.rs
mystenmark: this is already fixed further down the stack, sorry for the confusion

sui-execution/latest/sui-adapter/src/gas_charger.rs
mystenmark: Yes, the balance will just be 0 in that case

crates/sui-core/src/accumulators/transaction_rewriting.rs
```diff
@@ -0,0 +1,54 @@
...
+    for input in pt.inputs.iter_mut() {
+        if let CallArg::Object(ObjectArg::ImmOrOwnedObject(object_ref)) = input
+            && let Some(parsed) = ParsedObjectRefWithdrawal::parse(object_ref, chain_identifier)
+        {
+            compat_args.push(true);
+
+            let withdraw = coin_reservation_resolver
+                .resolve_funds_withdrawal(sender, parsed)
+                .unwrap();
```
lxfind, line 47: Is this unwrap safe? For instance, is it possible that the coin object exists at signing time, but then got deleted when we started executing this transaction (due to balance reaching 0)?

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -375,22 +394,62 @@ mod checked {
...
+            // When paying from address balance via the legacy API, we reserve whatever is
+            // specified by the coin object refs.
+            coin_reservation_total
+        };
+
         // Skip gas balance check for address balance payments
         // We reserve gas budget in advance
         if !gas_paid_from_address_balance {
             if gas_ownership_checks {
                 gas_status.check_gas_objects(&gas_objects)?;
```
lxfind, line 446: From what I can tell, we would call `check_gas_objects` unconditionally inside `gas_status.check_gas_balance`

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -319,96 +339,84 @@ pub mod checked {
...
+            if let Err(err) = self.gas_status.bucketize_computation(Some(is_move_abort))
+                && execution_result.is_ok()
+            {
+                *execution_result = Err(err);
             }
 
-            self.compute_storage_and_rebate(temporary_store, execution_result);
+            // On error we need to dump writes, deletes, etc before charging storage gas
+            if execution_result.is_err() {
+                self.reset(tx_ctx, temporary_store);
```
lxfind, line 385: Would this change `self.smashed_gas_coin`? If so, would that make `gas_coin_id` stale?

PR #25564, opened 2026-02-21
Relax ValidDuring requirement for transactions with owned inputs

mystenmark:
## Summary
- When `is_gas_paid_from_address_balance()` is true, the code previously required `ValidDuring` expiration unconditionally
- This is overly strict - the requirement is only necessary for "stateless" transactions (those with no owned object inputs), because they need expiration/nonce for replay protection
- Transactions with owned inputs already have replay protection because they consume unique object versions

This change adds a new protocol config flag `relax_valid_during_for_owned_inputs` (enabled in version 113) that allows transactions with owned object inputs to use any expiration type (None, Epoch, or ValidDuring) when paying gas from address balances.

## Test plan
- [x] Added unit tests in `address_balance_gas_tests.rs` verifying:
  - Transactions with owned inputs can use None, Epoch, or ValidDuring expiration
  - Stateless transactions still require ValidDuring
- [x] All e2e simtests pass (291 tests)

### DIFF
--- a/crates/sui-transactional-test-runner/src/test_adapter.rs
+++ b/crates/sui-transactional-test-runner/src/test_adapter.rs
@@ -75,7 +75,7 @@ use sui_types::committee::EpochId;
 use sui_types::crypto::{
     AuthorityKeyPair, AuthorityPublicKeyBytes, RandomnessRound, get_authority_key_pair,
 };
-use sui_types::digests::{ConsensusCommitDigest, TransactionDigest};
+use sui_types::digests::{ChainIdentifier, ConsensusCommitDigest, TransactionDigest};
 use sui_types::effects::{
     AccumulatorOperation, TransactionEffects, TransactionEffectsAPI, TransactionEvents,
 };
@@ -976,6 +976,11 @@ impl MoveTestAdapter<'_> for SuiTestAdapter {
                 };
                 let gas_price: u64 = gas_price.unwrap_or(self.gas_price);
                 let gas_budget = gas_budget.unwrap_or(DEFAULT_GAS_BUDGET);
+                let expiration = if address_balance_gas {
+                    self.get_replay_protected_expiration()
+                } else {
+                    TransactionExpiration::None
+                };
                 let transaction = self.sign_txn(sender, |sender, gas| {
                     let rec_arg = builder.pure(recipient).unwrap();
                     builder.command(sui_types::transaction::Command::TransferObjects(
@@ -984,7 +989,10 @@ impl MoveTestAdapter<'_> for SuiTestAdapter {
                     ));
                     let pt = builder.finish();
                     let gas = if address_balance_gas { vec![] } else { gas };
-                    TransactionData::new_programmable(sender, gas, pt, gas_budget, gas_price)
+                    let mut tx_data =
+                        TransactionData::new_programmable(sender, gas, pt, gas_budget, gas_price);
+                    *tx_data.expiration_mut_for_testing() = expiration;
+                    tx_data
                 });
                 let summary = self.execute_txn(transaction).await?;
                 let output = self.object_summary_output(&summary, /* summarize */ false);
@@ -1076,9 +1084,11 @@ impl MoveTestAdapter<'_> for SuiTestAdapter {
 
                 let summary = if !dev_inspect && !dry_run {
                     let gas_price = gas_price.unwrap_or(self.gas_price);
-                    let expiration = expiration
-                        .map(TransactionExpiration::Epoch)
-                        .unwrap_or(TransactionExpiration::None);
+                    let expiration = match expiration {
+                        Some(epoch) => TransactionExpiration::Epoch(epoch),
+                        None if address_balance_gas => self.get_replay_protected_expiration(),
+                        None => TransactionExpiration::None,
+                    };
                     let transaction = self.sign_sponsor_txn(
                         sender,
                         sponsor,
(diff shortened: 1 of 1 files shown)


crates/sui-protocol-config/src/lib.rs
```diff
@@ -299,6 +299,7 @@ const MAX_PROTOCOL_VERSION: u64 = 114;
 // Version 112: Enable Ristretto255 in devnet.
 // Version 113: Validate gas price >= RGP at signing for address balance gas payments.
 // Version 114: Gate seeded test overrides for checkpoint tx limit behind feature flag.
+//              Relax ValidDuring requirement for transactions with owned inputs.
```
alex-mysten: Should this now be 115 now as 1.67 was cut?

crates/sui-types/src/unit_tests/address_balance_gas_tests.rs
```diff
@@ -467,3 +496,116 @@ fn test_regular_gas_payment_with_epoch_expiration() {
         "Regular gas payment with Epoch expiration should be allowed"
     );
 }
+
+// Tests for relax_valid_during_for_owned_inputs feature
```
alex-mysten, line 658: Should we test with only SharedObjects and Coin Reservations (if possible now?)

mystenmark: Good idea - coin reservations contain a replay prevention mechanism so if present, ValidDuring should not be needed.

crates/sui-core/src/unit_tests/transaction_tests.rs
```diff
@@ -147,9 +147,21 @@ async fn test_empty_gas_data() {
     do_transaction_test_skip_cert_checks(
         |tx| {
             tx.gas_data_mut().payment = vec![];
+            // Clear inputs so this is a stateless transaction (no owned inputs).
+            // Address balance gas without ValidDuring requires replay protection from
+            // owned inputs, so this should fail with MissingGasPayment.
```
alex-mysten, line 158: I think as a user intending to use address balances I would be confused by a MissingGasPayment error for this scenario

crates/sui-types/src/transaction.rs
```diff
@@ -2009,6 +2009,20 @@ pub enum TransactionExpiration {
...
+    /// epoch. Therefore, ValidDuring with a one or two epoch range provides replay protection.
+    /// Either the transaction is statically invalid (current epoch not within range) or the
+    /// validator will remember if the transaction was already executed.
+    pub fn is_replay_protected(&self) -> bool {
+        matches!(self, TransactionExpiration::ValidDuring {
+                min_epoch: Some(min_epoch),
+                max_epoch: Some(max_epoch),
+                ..
+            } if *max_epoch == *min_epoch || *max_epoch == min_epoch.saturating_add(1))
+    }
```
alex-mysten, line 2023: Why would it be semantically incorrect to match TransactionExpiration::Epoch? But doing so would break ln 2995

crates/sui-types/src/transaction.rs
```diff
@@ -2818,9 +2832,12 @@ impl TransactionDataAPI for TransactionDataV1 {
                     .into());
                 }
 
-                // TODO: these checks can be loosened in the case where the transaction is not stateless,
-                // i.e. contains AddressOwned inputs.
+                // Legacy behavior: If ValidDuring is present, it must have either one- or two-epoch
+                // validity, even if the transaction is has other replay-protection.
+                // New behavior: ValidDuring can specify any epoch range. Replay protection is enforced
```
alex-mysten, line 2837: So... we relax the requirement for coin transactions, but enforce it for address-balance txns? There's a lot of conditionality that I think makes the API more difficult to reason about.

crates/sui-core/src/unit_tests/transaction_tests.rs
mystenmark: There is no clear correct answer imo:

- User who doesn't know about AB, but forgot to include a gas payment: `MissingGasPayment` is most helpful
- User who was trying to pay gas from AB, but didn't provide replay protection: `MissingGasPayment` is confusing

I think for now I'd prefer to err on the side of not changing behavior unnecessarily, hence the first choice

crates/sui-types/src/transaction.rs
mystenmark: I'm not sure i understand the question. `TransactionExpiration::ValidDuring` is the only form of expiration that currently provides replay protection, so it is the only case we can match against here.

crates/sui-types/src/transaction.rs
mystenmark: Perhaps, but it is the correct behavior: ValidDuring should not have artificial restrictions on what it can express. If I want to make a transaction that is valid from epochs 1045 to 1234, why shouldn't I be able to do that? (such a transaction must have some other form of replay protection, since a multi-epoch ValidDuring does not provide any)

I think the way to reason about it is:
- Transactions must have replay protection
- A one- or two-epoch ValidDuring is one form of replay protection, but not the only one.

mystenmark: or, maybe put another way: there is no conditionality here in the "new" path. The new path is only checking whether the transaction is expired. Replay protection is an entirely different concept that is checked elsewhere.

alex-mysten (approved):

PR #25568, opened 2026-02-23
feat: SIP-70 PTB Structural Digest

92GC:
## Summary

Adds one `sui::tx_context` native that lets governance and smart-account contracts verify PTB structure at execution time:

1. `structural_digest(&TxContext): vector<u8>` — a versioned, deterministic structural digest of the current PTB

This lets proposals approve an exact PTB shape and verify at execution time that the submitted PTB matches that approval.

## Main Changes

- SIP-70 structural digest over PTB commands, arguments, type arguments, and result flow
- Exact structural commitment only: no wildcarding, no second masked/normalized mode, no PTB introspection
- Length framing for variable-length fields and list counts
- Shared-object mutability included in the digest
- Owned and receiving object versions dropped from the digest for approval/execution stability
- Protocol-config gating in v113 and updated SIP documentation

## Follow-up Fixes

- Per-byte gas charging so digest work scales with PTB size
- Lazy digest computation plus transient `StructuralDigestData` so programmable transactions are not eagerly rehashed or fully cloned
- Explicit abort when structural digest data is unavailable
- Regression coverage for shared-object mutability and digest stability

## Validation

- [x] `cargo check -p sui-adapter-latest -p sui-move-natives-latest -p sui-protocol-config -p sui-types`
- [x] `cargo test -p sui-types structural_digest_tests --lib`

### DIFF
--- a/crates/sui-framework/packages/sui-framework/sources/tx_context.move
+++ b/crates/sui-framework/packages/sui-framework/sources/tx_context.move
@@ -81,6 +81,16 @@ public fun gas_price(_self: &TxContext): u64 {
 }
 native fun native_gas_price(): u64;
 
+/// Return the normalized structural digest of the current PTB (SIP-70).
+/// Deterministic: same logical PTB structure -> same digest,
+/// even if coin object IDs differ due to splits/merges.
+/// Enables governance contracts to vote on a PTB template hash
+/// and verify at execution time that the executor's PTB matches.
+public fun structural_digest(_self: &TxContext): vector<u8> {
+    native_structural_digest()
+}
+native fun native_structural_digest(): vector<u8>;
+
 // ==== test-only functions ====
 #[test_only]
 /// Return the number of id's created by the current transaction.
--- a/crates/sui-framework/packages/sui-framework/sources/tx_context.move
+++ b/crates/sui-framework/packages/sui-framework/sources/tx_context.move
@@ -86,10 +86,24 @@ native fun native_gas_price(): u64;
 /// even if coin object IDs differ due to splits/merges.
 /// Enables governance contracts to vote on a PTB template hash
 /// and verify at execution time that the executor's PTB matches.
+/// Output: [version_byte | blake2b256_hash] (33 bytes, version 0x01).
 public fun structural_digest(_self: &TxContext): vector<u8> {
     native_structural_digest()
 }
 native fun native_structural_digest(): vector<u8>;
+
+/// Return the structural digest with specified Pure inputs treated as wildcards (SIP-70 v2).
+/// Wildcarded inputs are hashed as a marker (0xFF) instead of their actual value,
+/// allowing governance contracts to verify PTB structure while letting the executor
+/// vary certain parameters (e.g. slippage tolerance, deadline timestamp).
+/// `wildcard_pure_indices` contains the input indices to wildcard.
+public fun structural_digest_masked(
+    _self: &TxContext,
+    wildcard_pure_indices: vector<u64>,
+): vector<u8> {
+    native_structural_digest_masked(wildcard_pure_indices)
+}
+native fun native_structural_digest_masked(wildcard_pure_indices: vector<u64>): vector<u8>;
 
 // ==== test-only functions ====
 #[test_only]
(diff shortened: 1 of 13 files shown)


tnowacki (commented):
Thanks for the PR! 

Internally, we have discussed this sort of pattern _a lot_ over the past two years. The main point we reached is one of unease. There are a lot of different considerations in the space, and the problems around them don't have a clear answer to us at this time. As a highlight, some of those problem spots are

- It is a difficult problem to carve up and to find what information should be included or excluded from the digests. As an example, in this PR the entire object ref is included. But does that make this more difficult to use? Should it just be object ID? How do we make it clear one way or another?
- Following that do we have a version that hashes the commands? Just the inputs? What about the value of the inputs?
- How do we ensure that this design feels future proof? What happens when we inevitably change/amend the PTB transaction type? What about new, future transaction types?
  - Particularly here we would be worried about contracts/modules/objects that have hard coded the results of these digests.

Outside of the difficulties in the design space for these sorts of hashes/digests, we wonder if this is the best approach for solving some of these problems. That is to say, is there a different feature that we should be focusing on instead. Or can this problem already be solved with the tools at hand? Some examples here:
- Can a shared object/transaction digest be used?
- Can hot-potato patterns help?
  - Changes to `entry` functions recently might change the calculus here
- Can dynamic fields or similar patterns help?
  - We have plans to add a sort of "temporary store" or "scratch pad" of dynamic fields limited to the transaction. This might also help in some cases

In summary, there might be some version of this feature that will land at some point, but the current PR would need some changes. The difficulty is we are not sure what exactly a version of this PR that could land would look like. @damirka is going to follow up (and talk to some other developers) to try to collect requirements in that regard.
Thanks again for the contribution and we look forward to discussing further 😄

92GC:
Thanks for the detailed review @tnowacki, appreciate the candor on the design space.

On the alternatives, from someone who's built the workarounds:

I've been building futarchy and multisig governance on Sui for the past year. Because there's no way to commit to PTB structure, I built a full 3-layer execution framework: BCS-serialized ActionSpecs stored in on-chain Intents, 65 typed action handlers (do_* functions), a witness-gated executable_resources bag for secure inter-action object passing, phantom-typed increment_action_idx for sequential type-safe execution, and position-based ordering that prevents skipping or reordering. This is ~35k LOC of Move.

Every new protocol integration means a new action module, a new add_*_spec, a new do_* function, new BCS serialization/deserialization, new phantom marker types, and new event emission. This is the direct cost of not having a structural commitment primitive.

Hot potatoes: I use them extensively (Executable is a hot potato). They solve atomicity but not commitment. The executor still controls which do_* functions to call and in what order. I had to build the entire ActionSpec + typed increment system on top to constrain that. Recent entry function changes help composability but don't eliminate the per-protocol wrapper burden.

Transaction-scoped dynamic fields ("scratch pad"): Would help with inter-action state passing (I built executable_resources for this). But it doesn't answer "is this the PTB the DAO voted on?" That requires comparing a pre-voted hash to the runtime structure.

tx_context::digest(): Changes every execution by design (includes signatures, gas, exact object refs). Can't be used for template matching.

On the design concerns, two concrete changes:

1. Drop version from owned objects. You're right that ObjectID + version is fragile. Version increments between vote and execution. Hash owned objects by ObjectID only, same treatment as shared objects.
2. Version prefix on the digest. First byte = digest scheme version (starting at 0x01). When the PTB type changes or new commands land, bump the version. Contracts storing digests can check compatibility. Cheap future-proofing.

On "what to include": For governance, the full structure is necessary: commands, order, arguments, flow. The protocol primitive should expose the complete structural hash.

Happy to work with @damirka on requirements. The concrete use case: proposal stores a digest at proposal time, voting happens, then executor builds a matching PTB, on-chain assert!(structural_digest(ctx) == proposal.digest()). Coin fungibility (type + balance normalization) is the key future extension.

92GC:
Update: this implementation is stale.

I no longer think the right primitive is a full `structural_digest(ctx)`. The SIP text in this PR has been rewritten around a smaller primitive:

`current_command_range_hash(ctx, n)`

That hashes the current command plus the next `n - 1` commands, without hashing sender/caller address, gas coin identity, or commands outside the selected range.

Closing this code PR so review does not focus on stale implementation details. If there is interest in the API shape, I can open a fresh implementation PR for the range-hash design.

PR #25591, opened 2026-02-23
remove max pending rows overrides

nickvikeras:
## Description

Pretty sure these dont matter


henryachen (approved):
Should this be higher than 5000 so it can ingest as much possible before back pressure?

PR #25596, opened 2026-02-24
Skip alias testing in composite workload if disabled in protocol config

aschran:
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/tests/simtest.rs
+++ b/crates/sui-benchmark/tests/simtest.rs
@@ -1552,15 +1552,18 @@ mod test {
         sui_protocol_config::ProtocolConfig::poison_get_for_min_version();
         let test_cluster = build_test_cluster(4, 10000, 1).await;
 
-        let protocol_config = sui_protocol_config::ProtocolConfig::get_for_version(
+        let mut protocol_config = sui_protocol_config::ProtocolConfig::get_for_version(
             sui_protocol_config::ProtocolVersion::max(),
             test_cluster.get_chain_identifier().chain(),
         );
         let address_balance_enabled = protocol_config.enable_address_balance_gas_payments();
-        assert!(
-            protocol_config.address_aliases(),
-            "address aliases must be enabled for this test"
-        );
+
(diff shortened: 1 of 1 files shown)


crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -331,7 +340,7 @@ impl CompositeWorkloadConfig {
             requirements.insert(InitRequirement::SeedAddressBalance);
         }
 
-        if self.alias_tx_probability > 0.0 {
+        if self.alias_tx_probability > 0.0 && !address_alias_disabled(protocol_config) {
```
alex-mysten, line 343: this will disable the init requirement... but the operation is still included, wouldn't it just fail running the test?

aschran: alias tx are a special case because they can't be combined in ptb with other tx and need custom handling for the state machine. so they are not generated via the "operation" abstraction. if the init is disabled, `alias_state` will not be set, and `advance_alias_state` will never indicate that an alias tx should be generated.

crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -57,6 +57,12 @@ fn address_balance_disabled(protocol_config: Option<&ProtocolConfig>) -> bool {
         .unwrap_or(false)
 }
 
+fn address_alias_disabled(protocol_config: Option<&ProtocolConfig>) -> bool {
```
alex-mysten, line 60: Do we still need the gating in composite/mod? Either we've overridden the protocol config feature flag, or the flag is enabled in mainnet

aschran: I think we need it for some other tests where I we want it to be allowed to be disabled. It just shouldn't be possible to disable for this one test

PR #25600, opened 2026-02-24
[cli] Add a ptb split coins test that preserves formatting

stefan-mysten:
## Description 

Follow-up to #25587 PR with a test to ensure we're not breaking the basic format of transaction response for `sui client ptb`.

## Test plan 

```
cargo nextest run -- ptb_split_coins
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/tests/shell_tests/with_network/ptb_split_coins/split_gas.sh
+++ b/crates/sui/tests/shell_tests/with_network/ptb_split_coins/split_gas.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+# Copyright (c) Mysten Labs, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+run_ptb() {
+  local output=$1
+  shift
+  if ! sui client --client.config "$CONFIG" ptb \
+    --move-call sui::tx_context::sender \
+    --assign sender \
+    --split-coins gas "[1000]" \
+    --assign split_coin \
+    --transfer-objects "[split_coin]" sender \
+    --gas-budget 50000000 \
+    "$@" \
+    > "$output" 2>&1; then
+    cat "$output"
+    exit 1
+  fi
+}
+
+redact_output() {
+  perl -pe '
+    s/0x[0-9a-fA-F]{64}/"0x" . ("X" x 64)/eg;
+
+    s/\b([1-9A-HJ-NP-Za-km-z]{32,})\b/"D" x length($1)/eg;
+    s/\b([A-Za-z0-9+\/]{80,}={0,2})\b/"S" x length($1)/eg;
+
+    s/"D{32,}"/"<DIGEST>"/g;
+    s/"S{80,}"/"<SIGNATURE>"/g;
+    s/(Transaction Digest:\s+)D+/$1<DIGEST>/g;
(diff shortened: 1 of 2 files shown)


crates/sui/tests/shell_tests/with_network/ptb_split_coins/split_gas.snap
```diff
@@ -0,0 +1,497 @@
...
+  | .effects.modifiedAtVersions |= sort_by(.objectId)
+' json.log | redact_output
+
+----- results -----
+success: true
+exit_code: 0
+----- stdout -----
+=== dry-run ===
+Dry run completed, execution status: success
+╭──────────────────────────────────────────────────────────────────────────────────────╮
```
mdgeorge4153, line 90: I'm not sure snapshotting the formatted output is important (and it seems like a lot of work) - I think that the key thing we care about is that automation doesn't break, and the json output should be sufficient for that.

crates/sui/tests/shell_tests/with_network/ptb_split_coins/split_gas.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+# Copyright (c) Mysten Labs, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+run_ptb() {
```
mdgeorge4153, line 5: A short comment here would be helpful

crates/sui/tests/shell_tests/with_network/ptb_split_coins/split_gas.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+# Copyright (c) Mysten Labs, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
```
mdgeorge4153, line 4: A top-level comment explaining what the goal of the test is accomplishing would be super helpful

stefan-mysten:
@mdgeorge4153 friendly ping!

crates/sui/tests/shell_tests/with_network/ptb_split_coins/split_gas.snap
mdgeorge4153: if there are specific things that we want to check for, we should just grep for them I think. The `redact_output` command is overkill

stefan-mysten: > - I think that the key thing we care about is that automation doesn't break, 

I am not so sure. Most users care at least about pretty printing. CI / automations care about json stuff. If we don't care that we break the pretty printing output, then sure, we don't need to invest in this and this test would be obsolete.

I can only do the `--json` stuff, but we will not catch if there's an issue with the pretty printing.

stefan-mysten: So this PR came after this PR https://github.com/MystenLabs/sui/pull/25587. Not sure what kind of test would make sense here, as the main PR was to bring back the right pretty printing and map new json to old json, thus snapshotting both.

PR #25614, opened 2026-02-25
Take the accumulator root as read-only in composite workload

mystenmark:
- **Add AccumulatorBalanceRead operation to composite workload**
- **Increase AccumulatorBalanceRead probability to 0.3**

### DIFF
--- a/crates/sui-benchmark/tests/simtest.rs
+++ b/crates/sui-benchmark/tests/simtest.rs
@@ -1642,6 +1642,18 @@ mod test {
                 "expected at least one alias remove"
             );
         }
+
+        let accum_read_stats = metrics
+            .get_stats(OperationSet::new().with(AccumulatorBalanceRead::FLAG))
+            .expect("expected accumulator balance read stats");
+        info!(
+            "accumulator balance read metrics: success={}",
+            accum_read_stats.success_count
+        );
+        assert!(
+            accum_read_stats.success_count > 0,
+            "expected at least one accumulator balance read"
+        );
     }
 
     #[sim_test(config = "test_config()")]


crates/sui-benchmark/src/workloads/composite/operations.rs
```diff
@@ -928,3 +934,61 @@ impl Operation for AddressBalanceOverdraw {
         );
     }
 }
+
+pub struct AccumulatorBalanceRead;
+
+impl AccumulatorBalanceRead {
+    pub const FLAG: u32 = 1 << 12;
```
alex-mysten: flag collision with Address Aliases operations

crates/sui-benchmark/tests/simtest.rs
```diff
@@ -1594,7 +1594,8 @@ mod test {
         .with_probability(TestCoinAddressDeposit::FLAG, 0.1)
         .with_probability(TestCoinAddressWithdraw::FLAG, 0.05)
         .with_probability(TestCoinObjectWithdraw::FLAG, 0.05)
-        .with_probability(AddressBalanceOverdraw::FLAG, 0.3);
+        .with_probability(AddressBalanceOverdraw::FLAG, 0.3)
+        .with_probability(AccumulatorBalanceRead::FLAG, 0.3);
```
alex-mysten, line 1592: should we assert success?

crates/sui-benchmark/src/workloads/data/composite/sources/accumulator_read.move
```diff
@@ -0,0 +1,18 @@
...
+use sui::event;
+use sui::sui::SUI;
+
+public struct SettledBalanceEvent has copy, drop {
+    value: u64,
+}
+
+entry fun read_settled_balance(root: &AccumulatorRoot, addr: address) {
+    let value = balance::settled_funds_value<SUI>(root, addr);
+    event::emit(SettledBalanceEvent { value });
```
alex-mysten, line 17: unused?

mystenmark: this commits to the value that was read in the effects, so if there is any discrepancy it will cause a fork

crates/sui-benchmark/tests/simtest.rs
mystenmark: I don't understand?

alex-mysten: With other operations, we capture metrics and assert that we have a certain proportion of successes (later in this test file). I was wondering if we should do the same here, effectively an assert_sometimes() that this operation succeeds.

PR #24523, opened 2025-12-04
[indexer-alt-framework] Set checkpoint_hi_inclusive, reader_lo, pruner_hi if watermark does not exist

evan-wall-mysten:
## Description 

Without the change in this PR, the indexer logs these errors when started using a non-zero `--first-checkpoint` for the first time (no watermark record exists):
```
2025-12-03T20:23:45.594924Z ERROR sui_indexer_alt_framework::pipeline::concurrent::pruner: Failed to prune data for range: 121926000 to 121928000: No checkpoint mapping found for checkpoint 121926000 pipeline="kv_epoch_starts"
```

This error is caused by the pruner attempting to prune data that was never indexed because `pruner_hi` is initialized to `0`.

This PR sets `checkpoint_hi_inclusive`, `reader_lo`, `pruner_hi` based on the `default_next_checkpoint` which is either the value from `--first-checkpoint` (defaulting to `0` if not set). When the pruner runs, it uses this value of `pruner_hi` to avoid trying to prune data that was never indexed.

## Test plan 

1. Added unit tests.
2. Tested with sui-indexer-alt-benchmark:
  a. Deploy sui-indexer-alt-benchmark with `pulumi up` (this line causes the benchmark to use the branch from this PR https://github.com/MystenLabs/sui-operations/blob/c3011e3ae777c8b58019c4a83211aefa54f8d372/pulumi/gcp/sui-indexer-alt-benchmark/Pulumi.dev.yaml#L8)
  b. Verify error is not in logs 
    ```
    kubectl logs sui-indexer-alt-benchmark-indexer-89f9c6965-s6hx9 -n sui-indexer-alt-benchmark`
    ```
---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Fix pruning for concurrent pipelines when indexer is initialized with `--first-checkpoint`.

PR #25616, opened 2026-02-25
sui-indexer-alt: remove init_watermark

evan-wall-mysten:
## Description 

1. Removes the `init_watermark` function added in #24523.

   This is a prereq to be able to index backwards because a checkpoint_hi_inclusive value cannot be written until that checkpoint has been indexed. 

   For example: Using `--first-checkpoint 10` would have previously resulted in a watermark record with `checkpoint_hi_inclusive = 9` even though checkpoint 9 was not actually indexed. If the indexer tried to index backwards from 10, it would have incorrectly skipped over checkpoint 9.

2. Delay the delay `pruner`, `reader_watermark` tasks until after a watermark exists to prevent WARN logs during initialization when those tasks attempt to read a watermark before it exists.

## Test plan 

1. Added assertion to unit tests.
2. Verified no warning/error logs in `sui-indexer-alt-benchmark`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-analytics-indexer/src/store/migration.rs
+++ b/crates/sui-analytics-indexer/src/store/migration.rs
@@ -251,23 +251,6 @@ impl MigrationStore {
         }
     }
 
-    /// Initialize migration mode for a pipeline.
-    ///
-    /// Reads existing watermark file if present.
-    ///
-    /// Returns the last processed checkpoint if a watermark exists, or None if starting fresh.
-    /// The watermark file is created/updated by `update_watermark` after file uploads.
-    pub(crate) async fn init_watermark(
-        &self,
-        pipeline: &str,
-        _default_next_checkpoint: u64,
-    ) -> anyhow::Result<Option<u64>> {
-        Ok(self
-            .committer_watermark(pipeline)
-            .await?
-            .map(|w| w.checkpoint_hi_inclusive))
-    }
-
     /// Update watermark for a single pipeline after successful file upload.
     ///
     /// Called by the upload worker after each file is successfully uploaded.
--- a/crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
@@ -424,6 +424,15 @@ mod tests {
         // Verify watermark progression
         let watermark = setup.store.watermark(DataPipeline::NAME).unwrap();
         assert_eq!(watermark.checkpoint_hi_inclusive, 3);
+
+        // pruner_hi must be initialized to checkpoint_hi_inclusive of the first write
+        // instead of 0. Otherwise, starting an indexer with --first-checkpoint > 0 causes the
+        // pruner to attempt to prune the checkpoints less than checkpoint_hi_inclusive that were
+        // never indexed.
+        assert_eq!(watermark.pruner_hi, 3);
+
+        // reader_lo cannot be below pruner_hi so it must be initialized to at least the same value.
+        assert_eq!(watermark.reader_lo, 3);
     }
 
     #[tokio::test]
--- a/crates/sui-pg-db/src/store.rs
+++ b/crates/sui-pg-db/src/store.rs
@@ -5,7 +5,6 @@ use std::time::Duration;
 
 use async_trait::async_trait;
 use chrono::DateTime;
-use chrono::Utc;
 use diesel::ExpressionMethods;
 use diesel::OptionalExtension;
 use diesel::prelude::*;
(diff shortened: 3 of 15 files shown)


crates/sui-analytics-indexer/tests/e2e_tests.rs
amnn, line 1: Could you remove as part of prep for landing?

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -473,10 +474,10 @@ async fn watermarks_from_pg(
         remaining_pipelines.remove(&row.pipeline);
     }
 
-    ensure!(
-        remaining_pipelines.is_empty(),
-        "Missing watermarks for {remaining_pipelines:?}",
-    );
+    if !remaining_pipelines.is_empty() {
```
amnn: This should not be relaxed -- the purpose of this check is to make sure we don't initialise the watermark at a higher value than it should be because we haven't accounted for all the pipelines, and then later have to roll it back -- the watermark is supposed to monotonic.

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -504,7 +505,11 @@ async fn watermark_from_consistent(
             tx_lo: 0,
         })),
 
-        Ok(_) => bail!("Missing data in consistent watermark"),
+        Ok(resp) => {
```
amnn: ditto

crates/sui-analytics-indexer/src/store/migration.rs
```diff
@@ -188,6 +188,16 @@ impl MigrationStore {
...
+
+            // Migration mode uses watermarks to keep track of what files have been migrated.
+            let epoch_hi = self
+                .committer_watermark(output_prefix)
+                .await?
+                .map(|w| w.epoch_hi_inclusive);
+            let index =
+                FileRangeIndex::load_from_store(&self.object_store, output_prefix, epoch_hi)
+                    .await?;
+            file_ranges.insert(output_prefix.to_string(), index);
```
nickvikeras: It looks like here you are inserting the file ranges for each pipeline twice, once with the pipeline name as the key, and once with the obejct store prefix? We shouldn't be storing it twice, and I think we should probably continue to use the pipeline name if that is what it was doing before.

nickvikeras: Also on line 144 we are already computing the backfill ranges and storing them on line 190 so this extra code seems redundant. 

Now I'm trying to remember I ever was doing this check in init_watermark in the first place. Lemme look into a bit more.

evan-wall-mysten: Removing this because it is no longer necessary after #25662.

crates/sui-indexer-alt-graphql/src/task/watermark.rs
evan-wall-mysten: Logging a warning for an expected initialization state is confusing for an operator because there is nothing actually wrong and no action to take.

Can we delay starting tasks that depend on the watermark (tasks I changed the log level for) until after we know a watermark record exists either through the init logic or after the committer task runs?

My goal isn't to relax validation logic, but to avoid having to distinguish between "good" and "bad" warnings/errors.

amnn: There is a potential action to take here though, which is that they may have forgotten to start these pipelines or mistakenly included the pipeline in GraphQL's config when it is not in use, and if we didn't warn about it, there would be no way of knowing there was an issue -- the system would just hang, waiting for the first watermark.

evan-wall-mysten: I added another commit that I think solves the issue I was talking about without switching the warns to infos.

evan-wall-mysten: Talked about this in our meeting - removed channels commit, removed the graphql changes, and changed the pruner/reader tasks' related warns to infos.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
```diff
@@ -424,6 +424,11 @@ mod tests {
         // Verify watermark progression
         let watermark = setup.store.watermark(DataPipeline::NAME).unwrap();
         assert_eq!(watermark.checkpoint_hi_inclusive, 3);
+
+        // pruner_hi must be initialized to checkpoint_hi_inclusive of the first write instead of 0.
+        // Otherwise, starting an indexer with --first-checkpoint > 0 causes the pruner to
+        // attempt pruning the checkpoints less than checkpoint_hi_inclusive that were never indexed.
+        assert_eq!(watermark.pruner_hi, 3);
```
amnn, line 432: Also `reader_lo`?

crates/sui-pg-db/src/store.rs
```diff
@@ -160,9 +118,14 @@ impl store::Connection for Connection<'_> {
             checkpoint_hi_inclusive: watermark.checkpoint_hi_inclusive as i64,
             tx_hi: watermark.tx_hi as i64,
             timestamp_ms_hi_inclusive: watermark.timestamp_ms_hi_inclusive as i64,
-            reader_lo: 0,
+            reader_lo: watermark.checkpoint_hi_inclusive as i64,
             pruner_timestamp: DateTime::UNIX_EPOCH.naive_utc(),
-            pruner_hi: 0,
+            // Initialize pruner_hi to checkpoint_hi_inclusive so the pruner does not attempt to
```
amnn: nit: let's move this comment above `reader_lo` and adapt it to reference both fields.

evan-wall-mysten:
The commit watermark task has a problem that it might commit a watermark that is not the first. This watermark's checkpoint_hi_inclusive should not be used as pruner_hi and reader_lo. Work is continued in https://github.com/MystenLabs/sui/pull/25737/

PR #25624, opened 2026-02-26
Consolidate tx checking in signing and simulate

lxfind:
## Description 

`simulate_transaction` duplicated validation logic from the signing path but had diverged: it called `process_funds_withdrawals_for_execution` (which uses `.unwrap()`) instead of the safe `process_funds_withdrawals_for_signing` (which returns `Result`), meaning a malformed transaction could panic a fullnode during simulation. This extracts the shared pre-object-load validation — deny checks, funds withdrawal parsing, and balance availability checks — into `pre_object_load_checks`, called by both `handle_transaction_deny_checks` and `simulate_transaction`, fixing the panic and ensuring the two paths stay in sync.

## Test plan 

- `cargo check -p sui-core` — compiles cleanly
- `SUI_SKIP_SIMTESTS=1 cargo nextest run -p sui-core` — all tests pass
- `cargo xclippy` — no warnings

---

## Release notes

- [ ] Protocol: 
- [x] Nodes (Validators and Full nodes): Fixes a potential fullnode panic when simulating a malformed transaction with invalid funds withdrawals.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/address_balance_tests.rs
+++ b/crates/sui-e2e-tests/tests/address_balance_tests.rs
@@ -3368,3 +3368,35 @@ async fn test_address_balance_gas_pay_all_sui() {
 
     test_env.verify_accumulator_exists(sender, 10_000_000);
 }
+
+/// Simulating a transaction with overflowing funds withdrawals must return an error.
+#[sim_test]
+async fn test_simulate_overflowing_funds_withdrawal_returns_error() {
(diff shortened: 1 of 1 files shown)


mystenmark:
should we add a test?

lxfind:
> should we add a test?

done

PR #25627, opened 2026-02-26
Make system object input checks default-deny

mystenmark:
Just hardening the system object checks.

### DIFF
--- a/crates/sui-transaction-checks/src/lib.rs
+++ b/crates/sui-transaction-checks/src/lib.rs
@@ -21,8 +21,9 @@ mod checked {
         TransactionDataAPI, TransactionKind,
     };
     use sui_types::{
-        SUI_AUTHENTICATOR_STATE_OBJECT_ID, SUI_CLOCK_OBJECT_ID, SUI_CLOCK_OBJECT_SHARED_VERSION,
-        SUI_RANDOMNESS_STATE_OBJECT_ID,
+        SUI_ACCUMULATOR_ROOT_OBJECT_ID, SUI_ADDRESS_ALIAS_STATE_OBJECT_ID, SUI_BRIDGE_OBJECT_ID,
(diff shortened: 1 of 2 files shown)


lxfind:
Is there a reason to not allow immutable access to system shared objects by default?
Asking because it seems error-prone to not to. e.g. are we sure we don't allow immutable access to randomness objects by users?

mystenmark:
> Is there a reason to not allow immutable access to system shared objects by default? Asking because it seems error-prone to not to. e.g. are we sure we don't allow immutable access to randomness objects by users?

We do allow immutable random access, that is handled in an earlier case. I could rewrite all the existing cases using the new pattern but it seems possibly error prone to do that

But also, yes, I suppose we could allow immutable access by default, that can't hurt anything

lxfind:
> We do allow immutable random access, that is handled in an earlier case. I could rewrite all the existing cases using the new pattern but it seems possibly error prone to do that

Oh, shall we merge them? hard to reason when there are separate cases like these

mystenmark:
> > We do allow immutable random access, that is handled in an earlier case. I could rewrite all the existing cases using the new pattern but it seems possibly error prone to do that
> 
> Oh, shall we merge them? hard to reason when there are separate cases like these

ok - i was trying to make the reasoning "this change doesn't widen the scope of allowed accesses" as easy as possible

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -522,58 +523,39 @@ mod checked {
...
-                }
-            }
-            InputObjectKind::SharedMoveObject {
-                id: SUI_AUTHENTICATOR_STATE_OBJECT_ID,
-                ..
-            } => {
-                if system_transaction {
-                    return Ok(());
-                } else {
-                    return Err(UserInputError::InaccessibleSystemObject {
```
lxfind, line 547: `InaccessibleSystemObject` is probably no longer used?

lxfind (approved):

PR #25644, opened 2026-02-27
jsonrpc: compute gas from accumulator events for failed ab transactions

alex-mysten:
## Description 

If we have a failed ab tx, gas costs are now attributed to the correct gas_owner 

## Test plan 

unit tested

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-json-rpc/src/balance_changes.rs
+++ b/crates/sui-json-rpc/src/balance_changes.rs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::collections::{BTreeMap, HashMap, HashSet};
-use std::ops::Neg;
 
 use async_trait::async_trait;
 use sui_types::balance_change::derive_balance_changes;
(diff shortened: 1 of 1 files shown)


crates/sui-json-rpc/src/balance_changes.rs
```diff
@@ -26,14 +28,29 @@ pub async fn get_balance_changes_from_effect<P: ObjectProvider<Error = E>, E>(
...
+        let accumulator_changes: Vec<_> = address_balance_changes_from_accumulator_events(effects)
+            .map(|(address, coin_type, amount)| BalanceChange {
+                owner: Owner::AddressOwner(address),
+                coin_type,
+                amount,
+            })
+            .collect();
+
+        if !accumulator_changes.is_empty() {
+            return Ok(accumulator_changes);
```
mystenmark: wouldn't this omit gas balance changes if accumulator changes are present and the gas is paid from a coin?

alex-mysten: I had assumed that accumulator changes would always be empty except for gas changes if the tx failed. Good point though, I will fix this

bmwill: Why can't we just unconditionally call the derive_balance_change function no matter if the txn fails or not?

bmwill: That's what we do for the gRPC path so it needs to work there and assuming it does, it should work here too.

alex-mysten: Much better idea. Updated

PR #25647, opened 2026-02-28
[address-balances] Additional tests for gas usage

tnowacki:
## Description 

- Refactor the gas `gas-budget-from-address-balance` into a boolean `address-balance-gas` to remove edge cases were both gas budgets were set
- Added some hardening around the flags and their usages
- Added additional tests

## Test plan 

- This is tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas.snap
+++ b/crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas.snap
@@ -1,7 +1,7 @@
 ---
 source: external-crates/move/crates/move-transactional-test-runner/src/framework.rs
 ---
-processed 6 tasks
+processed 5 tasks
 
 init:
 A: object(0,0)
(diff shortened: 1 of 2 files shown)


crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas.snap
```diff
@@ -0,0 +1,62 @@
...
+    },
+    balance: sui::balance::Balance<sui::sui::SUI> {
+        value: 300000000000000u64,
+    },
+}
+
+task 2, lines 13-15:
+//# programmable --sender A --address-balance-gas
+// View gas coin after -- balance should be unchanged
+Error: Error checking transaction input objects: Wrong number of parameters for the transaction
```
tzakian, line 27: Just to double-check -- this error is expected?

crates/sui-transactional-test-runner/src/test_adapter.rs
```diff
@@ -955,28 +960,30 @@ impl MoveTestAdapter<'_> for SuiTestAdapter {
                 recipient,
                 sender,
                 gas_budget,
-                gas_budget_from_address_balance,
+                address_balance_gas,
                 gas_price,
             }) => {
+                // address_balance_gas ==> is simulator
+                ensure!(
```
tzakian, line 967: Why not `assert!`?

crates/sui-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1001,14 +1008,19 @@ impl MoveTestAdapter<'_> for SuiTestAdapter {
...
+                address_balance_gas,
                 gas_price,
                 gas_payment,
                 dev_inspect,
                 dry_run,
                 expiration,
                 inputs,
             }) => {
+                // address_balance_gas ==> is simulator
+                ensure!(
```
tzakian, line 1020: Similar question. (super nitty, but I'm now just curious lol)

tzakian (approved):
a couple small questions, but nothing blocking I think.

crates/sui-transactional-test-runner/src/test_adapter.rs
tnowacki: Because the ones below arent 🤷

crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas.snap
tnowacki: Let me change these... I should duplicate them so we have one version that does something interesting

crates/sui-transactional-test-runner/src/test_adapter.rs
tzakian: sounds like a reasonable reason.

tnowacki: We should do a pass over these someday... but not right now

PR #25656, opened 2026-03-01
[core] tests/logs - clean up reachability assertions from v1 path, clarify log messages

alex-mysten:
## Description 

the asserts in settlement_scheduler.rs are unreachable in antithesis with `split_checkpoints_in_consensus_handler` and cannot be usefully replicated on that path.
`transaction executed in last epoch` is explicitly validated by simtest and impractical to hit with antithesis

## Test plan 

tests/logs only

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -1152,7 +1152,6 @@ impl AuthorityState {
             .get_transaction_cache_reader()
             .transaction_executed_in_last_epoch(transaction.digest(), epoch_store.epoch())
         {
-            assert_reachable!("transaction executed in last epoch");
             return Err(SuiErrorKind::TransactionAlreadyExecuted {
                 digest: (*transaction.digest()),
             }
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -1574,9 +1574,6 @@ impl AuthorityState {
             }
         }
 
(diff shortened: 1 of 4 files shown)


crates/sui-core/src/authority.rs
```diff
@@ -1152,7 +1152,6 @@ impl AuthorityState {
             .get_transaction_cache_reader()
             .transaction_executed_in_last_epoch(transaction.digest(), epoch_store.epoch())
         {
-            assert_reachable!("transaction executed in last epoch");
```
mystenmark, line 1155: doesn't seem like this should be removed. (we have never hit it, doing so is a testing TODO)

alex-mysten: `test_reject_transaction_executed_in_previous_epoch` directly exercises this scenario - do we also need antithesis coverage?

to repro, we need to:
1. submit tx
2. epoch change
3. aggressive checkpoint pruning (retention = 0)
4. validator restart to clear fx cache
5. resubmit tx

simtest seems like the right tool for this job

mystenmark: ok fair enough

mystenmark (approved):

PR #25657, opened 2026-03-01
[graphql-alt] Add SignatureScheme union with simple signature types [1/n]

tpham-mysten:
## Description 

- Convert `user_signature.rs` into a `user_signature/` module directory to support upcoming submodules for complex signature types
- Add a `SignatureScheme` GraphQL union type with `Ed25519Signature`, `Secp256k1Signature`, and `Secp256r1Signature` variants, each exposing parsed `signature` and `publicKey` fields

## Test plan 

How did you test the new or updated feature?

- Unit test

---

## Stack
- #25657 <-
- #25658
- #25659 

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Introduce a new union `SignatureScheme` for `UserSignature`
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #25658, opened 2026-03-01
[graphql-alt] Add MultisigSignature to SignatureScheme union [2/n]

tpham-mysten:
## Description 
- Add `MultisigSignature` variant to the `SignatureScheme` GraphQL union
- Introduce `MultisigMemberSignature` (flat struct with `MultisigMemberSignatureScheme` enum), `MultisigCommittee`, `MultisigMember`, and `MultisigMemberPublicKey` (union) types
- `MultisigMemberPublicKey` uses a union with `ZkLoginPublicIdentifier` (exposing `iss` + `addressSeed`) since it's structurally different from other key types
- `MultiSigLegacy` is converted to `MultiSig` via `TryFrom` to reuse a single code path
- ZkLogin public identifier conversion follows the same approach as gRPC (`rpc_proto_conversions.rs:1620`) — going through `sui_sdk_types`

## Test plan 

- Unit test

---

## Stack
- #25657 
- #25658 <-
- #25659 

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/user_signature/multisig.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/user_signature/multisig.rs
@@ -1,0 +1,190 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use async_graphql::{Enum, SimpleObject, Union};
+use sui_types::crypto::{CompressedSignature, PublicKey};
+use sui_types::multisig::MultiSig;
+
+use crate::api::scalars::base64::Base64;
+
+/// An aggregated multisig signature.
+#[derive(SimpleObject, Clone)]
+pub(crate) struct MultisigSignature {
+    /// The individual member signatures, one per signer who participated.
+    signatures: Option<Vec<MultisigMemberSignature>>,
+    /// A bitmap indicating which members of the committee signed.
+    bitmap: Option<u16>,
+    /// The multisig committee (public keys + weights + threshold).
+    committee: Option<MultisigCommittee>,
+}
+
+/// A single member's signature within a multisig.
+#[derive(SimpleObject, Clone)]
+pub(crate) struct MultisigMemberSignature {
+    /// The signature scheme used by this member.
+    scheme: Option<MultisigMemberSignatureScheme>,
+    /// The raw signature bytes (without public key).
+    signature: Option<Base64>,
+}
+
+/// The signature scheme of a multisig member's signature.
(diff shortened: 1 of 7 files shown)


crates/sui-indexer-alt-graphql/src/api/types/user_signature/multisig.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// The individual member signatures, one per signer who participated.
+    signatures: Option<Vec<MultisigMemberSignature>>,
+    /// A bitmap indicating which members of the committee signed.
+    bitmap: Option<u16>,
+    /// The multisig committee (public keys + weights + threshold).
+    committee: Option<MultisigCommittee>,
+}
+
+/// A single member's signature within a multisig.
+#[derive(SimpleObject, Clone)]
```
amnn, line 22: Could we re-use `UserSignature` for this? Then you wouldn't need to introduce a couple of these multi-sig specific types, and the schema would offer more detail on the individual member signatures (the same detail you added to `UserSignature` in the previous PR in the stack).

The only detail is that `UserSignature` includes `MultisigSignature` but I think that's okay, and easily explained away in the docs.

tpham-mysten: The issue is `UserSignature` currently wraps `GenericSignature`, but multisig member sigs are `CompressedSignature` (which contains signature bytes only, no public key). There's no conversion
  from `CompressedSignature` -> `GenericSignature`.

To reuse UserSignature, we'd need to make its internals more flexible — e.g. change from:

```
  pub(crate) struct UserSignature {
      pub(crate) native: GenericSignature,
  }
```

to something like:

```
  enum SignatureData {
      Full(GenericSignature),
      Compressed(CompressedSignature),
  }
```
.  Hence, I would prefer keeping thing as-is. Please let me know if I missed anything and I will fix in another PR

amnn: Understood, and agreed -- let's keep them separate. In that case, should we make `MultisigMemberSignature` match the shape of `UserSignature`? i.e. It's a union that exposes the structured data, rather than an `enum` + bytes?

tpham-mysten: Sounds good. I have added it together with ZkLogin logic in #25696

PR #25671, opened 2026-03-02
[Draft] Add GCP Confidential Spaces attestation native function

chargarlic:
## Description

Adds on-chain verification of GCP Confidential Spaces attestation JWTs, enabling Move contracts to cryptographically verify that code is running inside a GCP Confidential Space (AMD SEV-SNP) and inspect the workload's claims (image digest, hardware model, secure boot status, etc.).

### Design

- **Single native function** (`verify_gcp_attestation_internal`) mirrors the Nitro attestation pattern
- **Trust root via AuthenticatorState**: the Move entry point `verify_gcp_attestation(token, auth_state, kid, clock)` looks up the GCP RSA public key from `AuthenticatorState` using the GCP issuer + `kid`, so callers cannot inject arbitrary keys
- **JWK pipeline**: validators fetch GCP keys from Google's JWKS endpoint on the same cadence as zkLogin keys; keys flow through consensus into `AuthenticatorState`
- **Protocol-gated**: `enable_gcp_attestation()` in protocol config version 115+; disabled on Mainnet/Testnet pending review

### What's verified
JWT RS256 signature, expiry, and issuer. Returns a `GcpAttestationDocument` with: `iss`, `sub`, `aud`, `exp`, `iat`, `eat_nonce`, `secboot`, `hwmodel`, `swname`, `dbgstat`, `swversion`, container `image_digest`, `image_reference`, `restart_policy`.

### Security hardening
- Minimum RSA exponent size enforced (e ≥ 65537)
- Overflow guard on `exp`/`iat` before ms conversion
- 10s HTTP timeout + 64 KiB body cap on JWKS fetch
- Issuer not reflected in error messages

## Test plan

- 13 Rust unit tests (`sui-types`) covering signature verification, expiry, ISS validation, oversized tokens, etc.
- 4 Move unit tests covering parse errors, invalid signature, and unknown `kid`
- 2 e2e simtests: valid JWT verified on-chain after JWK injected into AuthenticatorState; unknown `kid` causes transaction abort

```
SUI_SKIP_SIMTESTS=1 cargo test -p sui-types --lib -- gcp_attestation
cargo simtest -p sui-e2e-tests --test gcp_attestation_tests
UPDATE=1 cargo test -p sui-framework --test build-system-packages
```

---

## Release notes

- [x] Protocol: new native function `verify_gcp_attestation_internal` in `sui::gcp_attestation`; new `enable_gcp_attestation` flag in protocol config (version 115+, non-Mainnet/Testnet only); `AuthenticatorState` (0x7) now allows immutable reads in user transactions
- [x] Nodes (Validators and Full nodes): new background task polls GCP JWKS endpoint and submits keys to consensus (same pipeline as zkLogin JWKs)


sui-execution/v1/sui-move-natives/src/crypto/gcp_attestation.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Stub implementation for old execution layers. The gcp_attestation Move module
+// is included in the compiled framework so the native must be registered to pass
+// bytecode verification. The enable_gcp_attestation feature flag is never set
+// for this execution layer, so this native always returns NOT_SUPPORTED_ERROR.
```
tzakian: I don't think this is needed for any version other than `latest`? We should be selecting older (stored/pre-compiled) frameworks when starting a network at an old protocol version, so it shouldn't be looking for this native I believe. Let me know if this is running into issues though...

crates/sui-framework-snapshot/manifest.json
```diff
@@ -2608,5 +2608,65 @@
         "id": "0x000000000000000000000000000000000000000000000000000000000000000b"
       }
     ]
+  },
+  "120": {
```
tzakian: Shouldn't have this snapshot. This should only be created after a release branch is cut and a commit is fixed for it..

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4750,6 +4773,8 @@ impl ProtocolConfig {
                     // Disabled while debugging
                     cfg.feature_flags.defer_unpaid_amplification = false;
                     cfg.feature_flags.enable_display_registry = true;
+                    // Bump package size limit to accommodate new framework modules (GCP attestation).
+                    cfg.max_move_package_size = Some(110 * 1024);
```
tzakian: Needed? If so we should chat...

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4742,6 +4757,14 @@ impl ProtocolConfig {
                 }
                 115 => {
                     cfg.feature_flags.normalize_depth_formula = true;
+                    // Enable GCP Confidential Spaces attestation on devnet only initially.
```
tzakian: Will need to shift these protocol gates into a more recent protocol version.

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -629,7 +630,9 @@ mod checked {
                         // System objects that can only be taken immutably
                         | (SUI_CLOCK_OBJECT_ID, SharedObjectMutability::Immutable)
                         | (SUI_RANDOMNESS_STATE_OBJECT_ID, SharedObjectMutability::Immutable)
-                        | (SUI_ACCUMULATOR_ROOT_OBJECT_ID, SharedObjectMutability::Immutable) => (),
+                        | (SUI_ACCUMULATOR_ROOT_OBJECT_ID, SharedObjectMutability::Immutable)
+                        // AuthenticatorState: immutable reads allowed for JWK lookup (e.g. GCP attestation)
+                        | (SUI_AUTHENTICATOR_STATE_OBJECT_ID, SharedObjectMutability::Immutable) => (),
```
tzakian: I'm not sure how I feel about this... probably should chat with some core folks and get their thoughts on this

crates/sui-framework/packages/sui-framework/sources/authenticator_state.move
```diff
@@ -355,11 +352,92 @@ fun get_active_jwks(self: &AuthenticatorState, ctx: &TxContext): vector<ActiveJw
     self.load_inner().active_jwks
 }
 
+/// Look up a JWK by issuer and key ID. Returns `none` if not found.
+public fun get_jwk_by_kid(self: &AuthenticatorState, iss: String, kid: String): Option<JWK> {
```
tzakian: Can this be `public(package)` or is the plan to allow for external callers of this?

crates/sui-json-rpc-types/src/sui_object.rs
```diff
@@ -655,7 +655,9 @@ impl TryInto<Object> for SuiObjectData {
     type Error = anyhow::Error;
 
     fn try_into(self) -> Result<Object, Self::Error> {
-        let protocol_config = ProtocolConfig::get_for_min_version();
+        // Use max version so that package size limits reflect the current protocol,
+        // not the historical minimum (which may be too restrictive for the current framework).
+        let protocol_config = ProtocolConfig::get_for_max_version_UNSAFE();
```
tzakian: This looks fine to me, but someone from the indexers/rpc-side of things should double-check this.

crates/sui-network/src/discovery/tests.rs
```diff
@@ -1645,13 +1645,12 @@ async fn test_discovery_only_peer_not_in_network_known_peers() -> Result<()> {
...
     assert!(
-        event_loop
+        !event_loop
             .state
             .read()
             .unwrap()
             .peer_addresses
-            .get(&gossip_peer_id)
-            .is_none(),
+            .contains_key(&gossip_peer_id),
```
tzakian: This change looks to be saying the same thing as what it was previously?

crates/sui-e2e-tests/tests/gcp_attestation_tests.rs
```diff
@@ -0,0 +1,267 @@
...
+        10_000_000,
+        rgp,
+    );
+    let tx = test_cluster.wallet.sign_transaction(&tx_data).await;
+    test_cluster.execute_transaction(tx).await;
+}
+
+/// A JWT whose kid is not present in AuthenticatorState must abort.
+#[sim_test]
+async fn test_gcp_attestation_rejects_unknown_kid() {
```
tzakian, line 100: Can you also add tests for calling `verify_gcp_attestation` where it isn't enabled in the protocol config and verify that it fails as expected?

crates/sui-e2e-tests/Cargo.toml
```diff
@@ -76,3 +76,6 @@ passkey-authenticator.workspace = true
 coset.workspace = true
 url.workspace = true
 p256.workspace = true
+base64.workspace = true
+rsa = { version = "0.9", default-features = false, features = ["std"] }
```
tzakian: Should these be workspace dependencies (and/or added/using e.g., fastcrypto or something for this?). Could you sync with the crypto folks to get their eyes on this as well -- I'm always a bit careful about pulling in crypto deps that we don't already have...

crates/sui-framework-snapshot/manifest.json
chargarlic: Yep, already removed the v120 snapshot. Left v121 since MAX_PROTOCOL_VERSION requires it for the compatibility test, but happy to remove that too if we should wait for a release cut.

crates/sui-protocol-config/src/lib.rs
chargarlic: Yep, will shift once we know which version to target. Just lmk.

crates/sui-transaction-checks/src/lib.rs
chargarlic: Totally understand the hesitation. The reason: `verify_gcp_attestation` needs to look up a trusted JWK rather than accepting a caller-supplied key (which would be insecure). Same pattern as Clock and RandomnessState — immutable only, mutable access stays blocked. The alternative would be a separate system object just for GCP JWKs but that felt like unnecessary duplication. Definitely worth getting core team eyes on this though.

crates/sui-framework/packages/sui-framework/sources/authenticator_state.move
chargarlic: Good catch — `public(package)` works since both modules are in sui-framework. I'll change `get_jwk_by_kid`. The JWK accessors (`jwk_n`, `jwk_e` etc.) I'd keep `public` since users need them to read the returned key, but open to locking those down too.

crates/sui-json-rpc-types/src/sui_object.rs
chargarlic: Will check with the indexer team.

crates/sui-network/src/discovery/tests.rs
chargarlic: Ugh, no-op from a rebase snafu. Will revert.

crates/sui-e2e-tests/tests/gcp_attestation_tests.rs
chargarlic: Good call, will add one.

crates/sui-e2e-tests/Cargo.toml
chargarlic: Switched to hardcoded test vectors for e2e tests (same pattern as zkLogin's `zklogin_test_vectors.json`), so `rsa` and `sha2` are no longer needed in `sui-e2e-tests`. They stay as dev-deps in `sui-types` for the benchmark which needs dynamic key generation.

sui-execution/v1/sui-move-natives/src/crypto/gcp_attestation.rs
chargarlic: You're right, removed them. They were a workaround for our v118 snapshot accidentally including gcp_attestation.move (dirty tree when I regenerated it).

crates/sui-protocol-config/src/lib.rs
chargarlic: Shifted to v121 (max version). Was at 115 which was way too old.

crates/sui-json-rpc-types/src/sui_object.rs
chargarlic: Reverted. The workaround was only needed during some broken-snapshot iterations. Main's `get_for_min_version()` works fine now.

crates/sui-protocol-config/src/lib.rs
chargarlic: Nope, removed. Thought there were package size limits earlier, but system packages use `u64::MAX`.

crates/sui-transaction-checks/src/lib.rs
chargarlic: Redesigned — no longer need immutable AuthenticatorState access. Moved the JWK lookup into the native via a `JwkMap` extension (populated from AuthenticatorState at execution setup time, same pattern as `NativesCostTable`). Move API is now just `verify_gcp_attestation(token, kid, clock)`. Aligns with how zkLogin handles JWKs

crates/sui-protocol-config/src/lib.rs
chargarlic: Actually had to add back a package size bump (100KB → 110KB) at v121 — turns out the genesis builder publishes framework via `Command::Publish` which enforces `max_move_package_size`, unlike `MovePackage::new_system`. Our framework with gcp_attestation is 102987 bytes.

chargarlic:
Superseded by https://github.com/MystenLabs/sui/pull/27398 — rebased onto current \`main\` at protocol **v133** (JwkMap design, no AuthState user-tx input). This branch is stale/conflicting; please close in favor of #27398.

chargarlic:
Closing in favor of rebased PR https://github.com/MystenLabs/sui/pull/27398 (aligned to main @ protocol v133).

PR #25679, opened 2026-03-03
Delete consistent pipelines code from sui-indexer-alt

wlmyng:
## Description 

Graphql and jsonrpc-alt now read object-related queries from consistent store instead of `obj_info` and `coin_balance_buckets`, so we can remove these pipelines from the postgres indexer.

The sui-indexer-alt-restorer crate will be removed because we don't have any postgres pipelines that would make use of it anymore. There is also a separate implementation in consistent-store's `restore` command that would be a better reference if we need restore from snapshot logic on the sui-indexer-alt indexer.

I'm not sure if it's preferable to delete the migrations related to  `obj_info` and `coin_balance_buckets` outright, or take the path in the PR, which is a new set of migrations that drop the tables in `up.sql`.

The `graphql-rpc.mdx` is also updated, now that the only object pipeline remaining is `obj_versions`

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-schema/src/objects.rs
+++ b/crates/sui-indexer-alt-schema/src/objects.rs
@@ -1,13 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use diesel::FromSqlRow;
-use diesel::backend::Backend;
-use diesel::deserialize;
-use diesel::expression::AsExpression;
 use diesel::prelude::*;
-use diesel::serialize;
-use diesel::sql_types::SmallInt;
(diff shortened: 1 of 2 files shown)


crates/sui-indexer-alt-schema/src/objects.rs
amnn: Do we need this anymore?

amnn: ditto

PR #25689, opened 2026-03-03
[bella-ciao][sui] Super minor cleanups and making sure things match `main` that should

tzakian:
## Description 

Diff the `sui` side directly against `main` and revert any drift. 

This also removes a new error for duplicate module names that was added in development, and reverts it to the existing error that you would get today.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-types/src/execution_status.rs
+++ b/crates/sui-types/src/execution_status.rs
@@ -247,9 +247,6 @@ pub enum ExecutionFailureStatus {
 
     #[error("Non-exclusive write input object {id} has been modified")]
     NonExclusiveWriteInputObjectModified { id: ObjectID },
-
-    #[error("Duplicate module name {duplicate_module_name} in package")]
-    DuplicateModuleName { duplicate_module_name: String },
     // NOTE: if you want to add a new enum,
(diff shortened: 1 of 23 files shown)


crates/sui-types/src/execution_status.rs
```diff
@@ -247,9 +247,6 @@ pub enum ExecutionFailureStatus {
 
     #[error("Non-exclusive write input object {id} has been modified")]
     NonExclusiveWriteInputObjectModified { id: ObjectID },
-
-    #[error("Duplicate module name {duplicate_module_name} in package")]
-    DuplicateModuleName { duplicate_module_name: String },
```
cgswords, line 252: 🎉

cgswords (approved):
LGTM. I assume removing that error is okay -- and I am happy it is gone.

tzakian:
It is -- only ever appeared on this branch/never made it onto `main` so no issue removing it.

PR #25659, opened 2026-03-01
[graphql-alt] Add ZkLogin and Passkey to SignatureScheme union [3/n]

tpham-mysten:
## Description 

- Add `ZkLoginSignature` variant exposing `maxEpoch` and the inner `userSignature`
- Add `PasskeySignature` variant exposing `authenticatorData`, `clientDataJson`, and the inner `userSignature`

## Test plan 

How did you test the new or updated feature?

- Unit test

---

## Stack
- #25657 
- #25658 
- #25659 <-

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #25696, opened 2026-03-04
[graphql-alt] Add missing ZkLogin fields to match gRPC interface [4/n]

tpham-mysten:
## Description 

- Add `inputs`, `publicIdentifier`, and `jwkId` fields to `ZkLoginSignature` to match the gRPC interface

## Test plan 

`cargo check -p sui-indexer-alt-graphql`

---

## Stack
- #25657 
- #25658 
- #25659
- #25696 <-

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/schema.graphql
+++ b/crates/sui-indexer-alt-graphql/schema.graphql
@@ -2830,31 +2830,6 @@ A multisig member's public key, varying by scheme.
 union MultisigMemberPublicKey = Ed25519PublicKey | Secp256K1PublicKey | Secp256R1PublicKey | PasskeyPublicKey | ZkLoginPublicIdentifier
 
 """
-A single member's signature within a multisig.
-"""
-type MultisigMemberSignature {
-	"""
-	The signature scheme used by this member.
-	"""
-	scheme: MultisigMemberSignatureScheme
-	"""
-	The raw signature bytes (without public key).
-	"""
-	signature: Base64
-}
-
-"""
-The signature scheme of a multisig member's signature.
-"""
-enum MultisigMemberSignatureScheme {
-	ED25519
-	SECP256K1
-	SECP256R1
-	ZKLOGIN
-	PASSKEY
-}
-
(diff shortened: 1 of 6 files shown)


tpham-mysten:
I agree `inputs` is complex and JSON would be simpler. If gRPC didn't already [expose](https://github.com/MystenLabs/sui-rust-sdk/blob/master/crates/sui-rpc/vendored/proto/sui/rpc/v2/signature.proto#L134) the same structured fields (`ZkLoginInputs`, `ZkLoginProof`, `ZkLoginClaim`, etc.), I would definitely go with JSON instead. But since it does, I'm leaning towards keeping it typed for consistency between the two interfaces.

Separately, my feeling is this field won't be commonly queried, mostly by advanced clients who would benefit from the full type safety and selective field querying that the structured approach provides. The logic is already implemented and not hard to maintain.

amnn:
> I agree `inputs` is complex and JSON would be simpler. If gRPC didn't already [expose](https://github.com/MystenLabs/sui-rust-sdk/blob/master/crates/sui-rpc/vendored/proto/sui/rpc/v2/signature.proto#L134) the same structured fields (`ZkLoginInputs`, `ZkLoginProof`, `ZkLoginClaim`, etc.), I would definitely go with JSON instead. But since it does, I'm leaning towards keeping it typed for consistency between the two interfaces.
> 
> Separately, my feeling is this field won't be commonly queried, mostly by advanced clients who would benefit from the full type safety and selective field querying that the structured approach provides. The logic is already implemented and not hard to maintain.

Yeah, that's fair -- I don't have a strong feeling on it. My main reason to do this was to avoid making people fetch every field they cared about from these types, because it's likely they would want all of them (and that would also if these fields are nested inside pages, it would have a multiplicative effect on the output nodes).

PR #25703, opened 2026-03-04
[bella-cia] Further small cleanups to make things matchup with `main` better

tzakian:
## Description 

What it says in the title

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-move-natives/src/crypto/groth16.rs
+++ b/sui-execution/latest/sui-move-natives/src/crypto/groth16.rs
@@ -70,8 +70,7 @@ pub fn prepare_verifying_key_internal(
             groth16_prepare_verifying_key_cost_params.groth16_prepare_verifying_key_bn254_cost_base
         }
         _ => {
-            // Charge for failure but dont fail if we run out of gas otherwise the actual error is masked by OUT_OF_GAS error
-            let _ = context.charge_gas(crypto_invalid_arguments_cost);
+            context.charge_gas(crypto_invalid_arguments_cost)?;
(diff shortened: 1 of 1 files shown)


sui-execution/latest/sui-move-natives/src/crypto/groth16.rs
```diff
@@ -71,7 +71,7 @@ pub fn prepare_verifying_key_internal(
         }
         _ => {
             // Charge for failure but dont fail if we run out of gas otherwise the actual error is masked by OUT_OF_GAS error
-            context.charge_gas(crypto_invalid_arguments_cost)?;
+            let _ = context.charge_gas(crypto_invalid_arguments_cost);
```
tzakian: Note this change in particular. Happy to revert it, but this matches `main`.

tnowacki: jfc. Can we fix this on main?

tzakian: I'd prefer to not fix (but fix forward here) in that case, as we'd need to protocol gate it (etc etc.)

dariorussi (approved):
I am fine with this.
So to be clear, we have not changed anything in the function and that comment was and is effectively a TODO?

PR #25708, opened 2026-03-04
docs: Add "Read this first" section to address balances migration guide

mystenmark:
## Summary
- Adds a "Read this first" section at the top of the address balances migration guide
- Clarifies that coins are not being deprecated and nothing is breaking
- Highlights the specific impact for wallet/exchange/custody providers who may receive funds via `send_funds()`

## Test plan
- [ ] Preview the docs build to verify formatting

### DIFF
--- a/docs/content/guides/developer/address-balances-migration.mdx
+++ b/docs/content/guides/developer/address-balances-migration.mdx
@@ -41,7 +41,6 @@ But, since the possibility exists that funds can arrive at any wallet as an addr
 Additionally, user funds may become **split across both coins and address balances**. Balance queries (via JSON-RPC, gRPC, or GraphQL) will show the combined total. However, to send funds that include the address balance portion, you must either:
 
 - Use `coinWithBalance` from a recent version of the TypeScript SDK (v2+), which automatically draws from both sources.
-- Use the JSON-RPC with an older SDK version, which provides backward compatibility through "fake coins" that represent address balance reservations.
 - Implement manual withdrawal logic as described in this guide.
 
 ## Overview


docs/content/guides/developer/address-balances-migration.mdx
```diff
@@ -18,6 +18,26 @@ Address balances introduce a canonical balance system for fungible assets tied t
...
+### Impact for wallets, exchanges, and custody providers
+
+Anyone operating a wallet (custodial, non-custodial, exchange wallets, etc) may be impacted by this change.
+The reason is: **users can now send funds to your address via `send_funds()` instead of transferring a coin object**.
+If a wallet receives funds this way, but cannot "see" the funds or access them, the user may believe that the funds have been lost.
+The funds cannot actually be lost, but this may cause confusion and concern for affected users.
+
+Initially, this will not happen often, as most of the ecosystem will continue to transfer coin objects, just as they do today.
+But, since the possibility exists that funds can arrive at any wallet as an address balance transfer, wallet implementors should be prepared for this.
+
```
hayes-mysten, line 40: should we add something about user funds being potentially split across address balances and coins, so transaction construction should use coinWithBalance from a recent sdk version?

hayes-mysten: or maybe for exchanges frame it as balances will show the combination of coins + balances and users won't be able to send the address balance parts unless they upgrade their sdk (or use an old version of the sdk with json rpc which has backwards compatibility through fake coins?)

mystenmark:
Addressed @hayes-mysten's feedback - added a note about funds potentially being split across coins and address balances, and the options for handling this (coinWithBalance, JSON-RPC backward compat, or manual withdrawals).

PR #25710, opened 2026-03-05
Rename ExecutionFailureStatus to ExecutionErrorKind

jordanjennings-mysten:
## Description 

- Uses ExecutionErrorKind in favor of ExecutionFailureStatus
- All uses of ExecutionFailureStatus are changed to ExecutionErrorKind
- Type alias ExecutionFailureStatus -> ExecutionErrorKind to keep CI happy and give some time to  to fully switch over 

## Test plan 

Existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-types/src/execution_status.rs
+++ b/crates/sui-types/src/execution_status.rs
@@ -39,6 +39,8 @@ impl fmt::Display for CongestedObjects {
     }
 }
 
+pub type ExecutionFailureStatus = ExecutionErrorKind;
+
 #[derive(Eq, PartialEq, Clone, Debug, Serialize, Deserialize, Error, EnumVariantOrder)]
 pub enum ExecutionErrorKind {
     //


crates/sui-types/src/error.rs
```diff
@@ -37,7 +37,7 @@ macro_rules! fp_ensure {
         }
     };
 }
-use crate::execution_status::{CommandIndex, ExecutionFailureStatus};
+pub use crate::execution_status::{CommandIndex, ExecutionFailureStatus};
```
jordanjennings-mysten: note the re-export, `sui_types::error::ExecutionFailureStatus` feels nicer + less work.

amnn: I'm generally not a fan of re-exporting a type when it is also accessible from the other location (because `rust-analyzer` is not very good about being consistent on where to import it from), but it's also maybe a bit late to be nitpicking about the structure of `sui-types`, so 🤷

crates/sui-types/src/error.rs
```diff
@@ -916,8 +916,8 @@ impl From<SuiError> for Status {
     }
 }
 
-impl From<ExecutionErrorKind> for SuiError {
-    fn from(kind: ExecutionErrorKind) -> Self {
+impl From<ExecutionFailureStatus> for SuiError {
+    fn from(kind: ExecutionFailureStatus) -> Self {
         ExecutionError::from_kind(kind).into()
```
amnn, line 920: Seems like the name of this function also needs to change, to `from_status`?

crates/sui-types/src/error.rs
```diff
@@ -1198,11 +1196,11 @@ impl ExecutionError {
         self
     }
 
-    pub fn from_kind(kind: ExecutionErrorKind) -> Self {
+    pub fn from_kind(kind: ExecutionFailureStatus) -> Self {
         Self::new(kind, None)
     }
 
-    pub fn kind(&self) -> &ExecutionErrorKind {
+    pub fn kind(&self) -> &ExecutionFailureStatus {
```
amnn: ditto, `status`.

sui-execution/v0/sui-adapter/src/execution_engine.rs
```diff
@@ -110,7 +110,7 @@ mod checked {
 
         let status = if let Err(error) = &execution_result {
             // Elaborate errors in logs if they are unexpected or their status is terse.
-            use ExecutionErrorKind as K;
+            use ExecutionFailureStatus as K;
```
amnn: `K` no longer makes sense, `E` or `S` would work.

jordanjennings-mysten:
After talking with @tzakian decided to go the other way here as you suggested @amnn: ExecutionErrorKind as the lasting name and removing ExecutionFailureStatus. Also appears protos/grpc is using `ExecutionErrorKind` not `ExecutionFailureStatus`. typescript SDK does reference "ExecutionFailureStatus" but it's only an private variable name, there was some small docs stuff but that seems to be it.  Error kind is definitely more idiomatic, status makes sense for success/fail enums but I think "kind" makes more sense in the context of an error enum.

PR #25715, opened 2026-03-05
Remove error from ZkLoginVerifyResult

tpham-mysten:
## Description 

Just return a regular error message if verification fails. GraphQL supports partial responses, so the fact that we produce an error in one sub-tree will not prevent other sub-trees from finishing resolving.

## Test plan 

- Unit test

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: ZkLoginVerifyResult will not contain error anymore. Error will not be part of GraphQL response error
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/query.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/query.rs
@@ -868,16 +868,21 @@ impl Query {
         signature: Base64,
         intent_scope: ZkLoginIntentScope,
         author: SuiAddress,
-    ) -> Result<ZkLoginVerifyResult, RpcError<zklogin::Error>> {
-        zklogin::verify_signature(
-            ctx,
-            self.scope(ctx)?,
-            bytes,
-            signature,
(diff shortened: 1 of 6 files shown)


crates/sui-indexer-alt-graphql/src/api/query.rs
henryachen: Do you want `Option<Result<ZkLoginVerifyResult, RpcError<zklogin::Error>>>` here? sibling resolvers will still run to completion but without this change won't you still get `data: null`?

tpham-mysten: Ah right! Thanks for the catch. I have applied Option

tpham-mysten:
Thanks @henryachen, I have fixed the tests

henryachen (approved):

PR #25717, opened 2026-03-05
graphql-alt: scan apis

henryachen:
## Description 

Scanning queries that finds transactions and events matching filters over a range of checkpoints

  Changes

  GraphQL (sui-indexer-alt-graphql)
  - New scanTransactions query endpoint
  - New scenEvents query endpoint
  - maxScanLimit service config to limit checkpoints scanned per query
  - Transaction and Event filter and scan logic in `bloom/mod.rs`

  Reader (sui-indexer-alt-reader)
  - `cp_blooms` loader for batch-loading bloom filter data

## Test plan 

```
cargo nextest run -p sui-indexer-alt-graphql graphql_scan_limit_tests
cargo nextest run -p sui-indexer-alt-graphql
cargo nextest run -p sui-indexer-alt
cargo nextest run -p sui-indexer-alt-schema
```

Query Plan:
```
Limit  (cost=1911.69..1358704.33 rows=62 width=8) (actual time=14.454..14.836 rows=62 loops=1)
  Buffers: shared hit=7287
  CTE block_bit_probes
    ->  ProjectSet  (cost=0.00..8.20 rows=1635 width=18) (actual time=0.004..0.431 rows=1635 loops=1)
          ->  Result  (cost=0.00..0.01 rows=1 width=0) (actual time=0.001..0.001 rows=1 loops=1)
  ->  Nested Loop  (cost=1903.49..1358696.13 rows=62 width=8) (actual time=14.453..14.830 rows=62 loops=1)
        Buffers: shared hit=7287
        ->  Limit  (cost=1903.06..1903.22 rows=62 width=24) (actual time=14.411..14.414 rows=1 loops=1)
              Buffers: shared hit=6976
              CTE block_lookup
                ->  Nested Loop Left Join  (cost=41.30..1723.38 rows=200 width=131) (actual time=0.654..2.899 rows=327 loops=1)
                      Buffers: shared hit=1348
                      ->  HashAggregate  (cost=40.88..42.88 rows=200 width=10) (actual time=0.620..0.682 rows=327 loops=1)
                            Group Key: block_bit_probes_1.cp_block_index, block_bit_probes_1.bloom_idx
                            Batches: 1  Memory Usage: 61kB
                            ->  CTE Scan on block_bit_probes block_bit_probes_1  (cost=0.00..32.70 rows=1635 width=10) (actual time=0.000..0.159 rows=1635 loops=1)
                      ->  Index Scan using cp_bloom_blocks_pkey on cp_bloom_blocks bb  (cost=0.42..8.40 rows=1 width=131) (actual time=0.006..0.006 rows=1 loops=327)
                            Index Cond: ((cp_block_index = block_bit_probes_1.cp_block_index) AND (bloom_block_index = block_bit_probes_1.bloom_idx))
                            Buffers: shared hit=1343
              ->  Sort  (cost=179.69..180.17 rows=192 width=24) (actual time=14.410..14.412 rows=1 loops=1)
                    Sort Key: ((block_bit_probes.cp_block_index * '1000'::bigint))
                    Sort Method: quicksort  Memory: 25kB
                    Buffers: shared hit=6976
                    ->  Hash Right Anti Join  (cost=48.29..173.01 rows=192 width=24) (actual time=14.392..14.400 rows=5 loops=1)
                          Hash Cond: (p.cp_block_index = block_bit_probes.cp_block_index)
                          Buffers: shared hit=6976
                          ->  Hash Join  (cost=7.00..129.78 rows=8 width=8) (actual time=3.224..12.355 rows=1011 loops=1)
                                Hash Cond: ((p.cp_block_index = bl.cp_block_index) AND (p.bloom_idx = bl.bloom_idx))
                                Join Filter: ((bl.bloom_filter IS NULL) OR (p.bit_mask <> get_byte(bl.bloom_filter, (p.byte_pos % length(bl.bloom_filter)))))
                                Rows Removed by Join Filter: 624
                                Buffers: shared hit=6976
                                ->  CTE Scan on block_bit_probes p  (cost=0.00..32.70 rows=1635 width=18) (actual time=0.000..0.211 rows=1635 loops=1)
                                ->  Hash  (cost=4.00..4.00 rows=200 width=42) (actual time=3.142..3.142 rows=327 loops=1)
                                      Buckets: 1024  Batches: 1  Memory Usage: 93kB
                                      Buffers: shared hit=1348
                                      ->  CTE Scan on block_lookup bl  (cost=0.00..4.00 rows=200 width=42) (actual time=0.656..3.060 rows=327 loops=1)
                                            Buffers: shared hit=1348
                          ->  Hash  (cost=38.79..38.79 rows=200 width=8) (actual time=1.773..1.774 rows=327 loops=1)
                                Buckets: 1024  Batches: 1  Memory Usage: 21kB
                                ->  HashAggregate  (cost=36.79..38.79 rows=200 width=8) (actual time=1.668..1.717 rows=327 loops=1)
                                      Group Key: block_bit_probes.cp_block_index
                                      Batches: 1  Memory Usage: 61kB
                                      ->  CTE Scan on block_bit_probes  (cost=0.00..32.70 rows=1635 width=8) (actual time=0.007..1.162 rows=1635 loops=1)
        ->  Index Scan using cp_blooms_pkey on cp_blooms cb  (cost=0.43..21883.74 rows=1 width=8) (actual time=0.040..0.408 rows=62 loops=1)
              Index Cond: ((cp_sequence_number >= ((block_bit_probes.cp_block_index * '1000'::bigint))) AND (cp_sequence_number <= ((((block_bit_probes.cp_block_index * '1000'::bigint) + '1000'::bigint) - 1))))
              Filter: (((get_byte(bloom_filter, (208 % length(bloom_filter))) & 8) = 8) AND ((get_byte(bloom_filter, (5988 % length(bloom_filter))) & 32) = 32) AND ((get_byte(bloom_filter, (8084 % length(bloom_filter))) & 32) = 32) AND ((get_byte(bloom_filter, (15059 % length(bloom_filter))) & 32) = 32) AND ((get_byte(bloom_filter, (8330 % length(bloom_filter))) & 64) = 64) AND ((get_byte(bloom_filter, (10313 % length(bloom_filter))) & 4) = 4))
              Rows Removed by Filter: 318
              Buffers: shared hit=311
Planning:
  Buffers: shared hit=78
Planning Time: 2.181 ms
Execution Time: 15.060 ms
```

## Stack:
https://github.com/MystenLabs/sui/pull/24788 
https://github.com/MystenLabs/sui/pull/24900 👈 
https://github.com/MystenLabs/sui/pull/25097
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/events/scan.move
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/events/scan.move
@@ -1,0 +1,169 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//# init --protocol-version 70 --addresses P1=0x0 P2=0x0 --accounts A B C --simulator
+
+//# publish
+module P1::M1 {
+    use sui::event;
+    use std::ascii;
+
+    public struct EventA has copy, drop {
+        message: ascii::String,
+        value: u64,
+    }
+
+    public struct EventB has copy, drop {
+        message: ascii::String,
+        value: u64,
+    }
+
+    public entry fun emit_a(value: u64) {
+        event::emit(EventA {
+            message: ascii::string(b"Event A"),
+            value,
+        });
+    }
+
+    public entry fun emit_b(value: u64) {
+        event::emit(EventB {
+            message: ascii::string(b"Event B"),
+            value,
+        });
+    }
+
+    public entry fun emit_multiple(count: u64) {
+        let mut i = 0;
+        while (i < count) {
+            event::emit(EventA {
+                message: ascii::string(b"Event from loop"),
+                value: i + 1,
+            });
+            i = i + 1;
+        };
+    }
+}
+
+//# publish
+module P2::M2 {
+    use sui::event;
+    use std::ascii;
+
+    public struct EventC has copy, drop {
+        message: ascii::String,
+        value: u64,
+    }
+
+    public entry fun emit_c(value: u64) {
+        event::emit(EventC {
+            message: ascii::string(b"Event C from P2"),
+            value,
+        });
+    }
+}
+
+// Checkpoint 1: A emits EventA from P1::M1
+//# run P1::M1::emit_a --sender A --args 1
+
+//# create-checkpoint
+
+// Checkpoint 2: B emits EventB from P1::M1
+//# run P1::M1::emit_b --sender B --args 2
+
+//# create-checkpoint
+
+// Checkpoint 3: A emits EventC from P2::M2
+//# run P2::M2::emit_c --sender A --args 3
+
+//# create-checkpoint
+
+// Checkpoint 4: C emits EventA from P1::M1
+//# run P1::M1::emit_a --sender C --args 4
+
+//# create-checkpoint
+
+// Checkpoint 5: Multiple events in one checkpoint (A emits 3 EventA)
+//# run P1::M1::emit_multiple --sender A --args 3
+
+//# create-checkpoint
+
+//# advance-epoch
+
+// Generate 10,501 checkpoints to create 10 complete blocks + 500 extra for incomplete block
+//# create-checkpoint 10501
+
+// Checkpoint after blocks: B emits EventC from P2
+//# run P2::M2::emit_c --sender B --args 100
+
+//# create-checkpoint
+
+//# run-graphql
+# Scan vs indexed cross-check, multi-filter combinations, empty-result edge cases,
+# checkpoint-bounded scan, and scan across bloom blocks.
+{
+  scanEventsA: scanEvents(filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+  paginateEventsA: events(filter: { sender: "@{A}" }) { ...EV }
+
+  eventsAFromP1: scanEvents(filter: {
+    sender: "@{A}",
+    module: "@{P1}",
+    beforeCheckpoint: 10510
+  }) { ...EV }
+  scanTripleFilter: scanEvents(filter: {
+    sender: "@{A}",
+    module: "@{P1}",
+    type: "@{P1}::M1::EventA",
+    beforeCheckpoint: 10510
+  }) { ...EV }
+
+  emptyNonExistent: scanEvents(filter: {
+    sender: "0x0000000000000000000000000000000000000000000000000000000000000001",
+    beforeCheckpoint: 10510
+  }) { ...EV }
+
+  earlyEvents: scanEvents(filter: {
+    sender: "@{A}",
+    beforeCheckpoint: 4
+  }) { ...EV }
+  atCheckpoint5: scanEvents(filter: {
+    sender: "@{A}",
+    atCheckpoint: 5
+  }) { ...EV }
+
+  noFilter: scanEvents(first: 5, filter: {
+    afterCheckpoint: 0,
+    beforeCheckpoint: 6
+  }) { ...EV }
+
+  scanAcrossBlocks: scanEvents(filter: {
+    sender: "@{A}",
+    afterCheckpoint: 0,
+    beforeCheckpoint: 10510
+  }) { ...EV }
+}
+
+fragment EV on EventConnection {
+  pageInfo { startCursor endCursor hasPreviousPage hasNextPage }
+  edges { cursor node { sequenceNumber sender { address } contents { type { repr } } } }
+}
+
+//# run-graphql --cursors {"c":1,"t":3,"e":0} {"c":5,"t":7,"e":0} {"c":5,"t":7,"e":2} {"c":20,"t":20,"e":0}
+# Cursor pagination: cursor_0 is cp1 event (t=3,e=0),
+# cursor_1 is cp5 first event (t=7,e=0), cursor_2 is cp5 last event (t=7,e=2).
+{
+  first2: scanEvents(first: 2, filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+  last2: scanEvents(last: 2, filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+
+  firstAfter: scanEvents(first: 10, after: "@{cursor_0}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+  firstBefore: scanEvents(first: 10, before: "@{cursor_2}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+
+  windowFirst: scanEvents(first: 10, after: "@{cursor_0}", before: "@{cursor_2}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+
+  nonexistentCursor: scanEvents(last: 10, after: "@{cursor_3}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+  invalidOrder: scanEvents(first: 10, after: "@{cursor_2}", before: "@{cursor_0}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+}
+
+fragment EV on EventConnection {
+  pageInfo { startCursor endCursor hasPreviousPage hasNextPage }
+  edges { cursor node { sequenceNumber sender { address } contents { type { repr } } } }
+}
--- a/crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
@@ -4,7 +4,10 @@
 use std::ops::Range;
 
 use anyhow::Context as _;
+use async_graphql::CustomValidator;
 use async_graphql::InputObject;
+use async_graphql::InputValueError;
+use sui_indexer_alt_schema::blooms::should_skip_for_bloom;
 use sui_pg_db::query::Query;
 use sui_sql_macro::query;
 use sui_types::event::Event as NativeEvent;
@@ -16,7 +19,6 @@ use crate::api::scalars::uint53::UInt53;
 use crate::api::types::event::CEvent;
 use crate::api::types::lookups::CheckpointBounds;
 use crate::error::RpcError;
-use crate::error::feature_unavailable;
 use crate::pagination::Page;
 
 #[derive(InputObject, Debug, Default, Clone)]
@@ -35,7 +37,7 @@ pub(crate) struct EventFilter {
 
     /// Events emitted by a particular module. An event is emitted by a particular module if some function in the module is called by a PTB and emits an event.
     ///
-    /// Modules can be filtered by their package, or package::module. We currently do not support filtering by emitting module and event type at the same time so if both are provided in one filter, the query will error.
+    /// Modules can be filtered by their package, or package::module.
     pub module: Option<ModuleFilter>,
 
     /// This field is used to specify the type of event emitted.
@@ -51,12 +53,7 @@ impl EventFilter {
     /// Uses the provided transaction bounds subquery to limit results to a specific transaction range
     pub(crate) fn query<'q>(&self) -> Result<Query<'q>, RpcError> {
         let table = match (&self.module, &self.type_) {
-            (Some(_), Some(_)) => {
-                return Err(feature_unavailable(
-                    "Filtering by both emitting module and event type is not supported",
-                ));
-            }
-            (Some(_), None) => query!("ev_emit_mod"),
+            (Some(_), _) => query!("ev_emit_mod"),
             (None, _) => query!("ev_struct_inst"),
         };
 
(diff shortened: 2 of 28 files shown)


amnn (approved):

crates/sui-indexer-alt-schema/src/blooms/bloom.rs
```diff
@@ -1,9 +1,17 @@
...
+use std::collections::BTreeMap;
+
 use crate::blooms::hash::DoubleHasher;
 use crate::blooms::hash::set_bit;
 
+/// Probe for checking membership in a bloom filter.
+/// Each entry in `bit_probes` is a (byte_offset, bit_mask) pair where the bit must be set.
+pub struct BloomProbe {
+    pub bit_probes: Vec<(usize, u8)>,
+}
```
wlmyng, line 13: If I understand correctly, the algorithm is the following

`BloomFilter.bytes[BloomProbe.bit_probes[0][0]] & BloomProbe.bit_probes[0][1] == BloomProbe.bit_probes[0][1]`

and we do that for every entry in `bit_probes`
- if anyone fails, the thing we desire is definitely not in the filter
- otherwise, if all pass, the value is probably in the filter

I just had a hard time wrapping my head around how the `byte_offset` and `bit_mask` interacted with the bloom filter. We're indexing into the `bytes` and then comparing the entire byte against the mask

crates/sui-indexer-alt-schema/src/blooms/bloom.rs
```diff
@@ -64,6 +72,20 @@ impl<const BYTES: usize, const HASHES: u32, const SEED: u128> BloomFilter<BYTES,
...
             .map(move |h| (h as usize) % num_bits)
     }
+
+    /// Byte offsets and bit masks of values, used for SQL membership checks.
+    /// Probes sharing the same byte offset are grouped by ORing their bit masks together.
+    pub fn probe(values: impl IntoIterator<Item = impl AsRef<[u8]>>) -> BloomProbe {
+        let mut by_offset: BTreeMap<usize, u8> = BTreeMap::new();
+        for value in values {
+            for b in Self::hash(value.as_ref()) {
+                *by_offset.entry(b / 8).or_default() |= 1u8 << (b % 8);
```
wlmyng, line 82: when creating the bloom filter, we also fold it

On the read/ probe side, we compute the bit position against the full-size filter, so using the probe as-is might produce offsets that are too large for a folded filter.

This means that somewhere, we need to adjust for the fold

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -0,0 +1,413 @@
...
+/// Check if all filter values are present in a checkpoint's bloom filter.
+fn cp_bloom_check_sql(probe: &BloomProbe) -> Query<'static> {
+    if probe.bit_probes.is_empty() {
+        return query!("TRUE");
+    }
+
+    let mut condition = query!("TRUE");
+    for &(offset, mask) in &probe.bit_probes {
+        condition += query!(
+            " AND (get_byte(cp_blooms.bloom_filter, {Integer} % length(cp_blooms.bloom_filter)) & {Integer}) = {Integer}",
```
wlmyng, line 393: so the fold adjustment happens here .. on one hand it's nice that the reader doesn't need to know the exact adjustment, sinc eit's based on the final size of the column itself, on the other hand, the reader needs to know to remember to do `% length(bloom_filter)`

crates/sui-indexer-alt-schema/src/blooms/bloom.rs
wlmyng: I think it's worth calling out as a doc comment that this produces probes assuming full-sized filter, and that users need to adjust for the stored folded filter

wlmyng:
```

  /// Probe for checking membership in a bloom filter.
  /// Each entry in `bit_probes` is a (byte_offset, bit_mask) pair.
  /// Byte offsets are computed against the full-size filter. If the filter has been
  /// folded, offsets must be taken modulo the folded size (e.g. `offset % filter.len()`).
```

crates/sui-indexer-alt-schema/src/blooms/blocked.rs
```diff
@@ -100,6 +102,31 @@ impl<const BYTES: usize, const BLOCKS: usize, const HASHES: u32>
...
+    /// Probes for use bloom filter membership checks, probes from different values that hash to the
+    /// same block are merged together.
+    pub fn probe(
+        seed: u128,
+        values: impl IntoIterator<Item = impl AsRef<[u8]>>,
+    ) -> Vec<BlockedBloomProbe> {
+        let mut by_block: BTreeMap<usize, BTreeMap<usize, u8>> = BTreeMap::new();
+        for value in values {
+            let (block_idx, bits) = Self::hash(value.as_ref(), seed);
+            let block_entry = by_block.entry(block_idx).or_default();
```
wlmyng, line 115: seems this logic is the same whether in `bloom` or here ... what if `BloomProbe` has a constructor `from_bit_positions` that calculates this offset + mask?

crates/sui-indexer-alt-schema/src/blooms/bloom.rs
```diff
@@ -64,6 +72,20 @@ impl<const BYTES: usize, const HASHES: u32, const SEED: u128> BloomFilter<BYTES,
             .take(HASHES as usize)
             .map(move |h| (h as usize) % num_bits)
     }
+
+    /// Byte offsets and bit masks of values, used for SQL membership checks.
+    /// Probes sharing the same byte offset are grouped by ORing their bit masks together.
```
wlmyng, line 77: When different hashes have bit positions that land in the same byte (offset), their masks get OR'd so we don't have two entries for the same byte offset. But the overall query is an `AND` not an `OR`

crates/sui-indexer-alt-schema/src/blooms/blocked.rs
```diff
@@ -100,6 +102,31 @@ impl<const BYTES: usize, const BLOCKS: usize, const HASHES: u32>
             .map(move |h| (h as usize) % bits_per_block);
         (block_idx, bit_iter)
     }
+
+    /// Probes for use bloom filter membership checks, probes from different values that hash to the
+    /// same block are merged together.
+    pub fn probe(
+        seed: u128,
+        values: impl IntoIterator<Item = impl AsRef<[u8]>>,
+    ) -> Vec<BlockedBloomProbe> {
```
wlmyng, line 111: I think this doc comment only made sense to me once I realized that each value may map to a different block, and we can't merge probes across blocks, hence we produce a `Vec` of probes. I was thrown off since in `bloom.rs`, given a vec of values we produce a single `BloomProbe`. Could we add to/ rephrase the doc comments that each value potentially hashes to a specific block and only probes within the same block can be merged

crates/sui-indexer-alt-graphql/src/api/types/transaction/filter.rs
```diff
@@ -42,38 +65,10 @@ pub(crate) struct TransactionFilter {
...
-
-        Ok(())
-    }
-}
-
-#[derive(thiserror::Error, Debug)]
-pub(crate) enum Error {
-    #[error("Invalid filter, expected: {0}")]
-    InvalidFormat(&'static str),
+pub(crate) struct ScanFilterValidator {
```
wlmyng, line 70: would be nice to have some doc comments to explain what it enforces and why (this and TFValidator)

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -0,0 +1,413 @@
...
+    // The NOT EXISTS short-circuits on the first failing probe per pair.
+    // The GROUP BY / HAVING ensures ALL bloom_block_indices pass per cp_block_index.
+    let matched_blocks = query!(
+        r#"
+        SELECT
+            cp_bloom_blocks.cp_block_index,
+            cp_bloom_blocks.cp_block_index * {BigInt} AS cp_lo,
+            cp_bloom_blocks.cp_block_index * {BigInt} + {BigInt} - 1 AS cp_hi_inclusive
+        FROM
+            (SELECT DISTINCT cp_block_index, bloom_block_index, bloom_count
```
wlmyng, line 249: we cant dedupe earlier because the CTE needs the full granularity for the NOT EXISTS check, but the dedupe is so each bloom filter gets read once, then probed N times (where N is num probes)

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -0,0 +1,413 @@
...
+
+    let (cp_lo, cp_hi) = clamped_cp_bounds(page, &cp_bounds);
+    let filter_values = filter.bloom_probe_values();
+    let candidate_cps = candidate_cps(ctx, &filter_values, cp_lo, cp_hi, page).await?;
+
+    if candidate_cps.is_empty() {
+        return Ok(BTreeMap::new());
+    }
+
+    let checkpoints = kv_loader
```
wlmyng, line 139: clamp bounds -> bloom block scan -> bloom checkpoint scan -> load checkpoints -> extract digests is shared between events and txns (and likely other readers of bloom filters), we could extract this into a shared fn

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -298,6 +314,54 @@ impl Transaction {
...
+            filter.after_checkpoint().map(u64::from),
+            filter.at_checkpoint().map(u64::from),
+            filter.before_checkpoint().map(u64::from),
+            reader_lo,
+            checkpoint_viewed_at,
+        ) else {
+            return Ok(Connection::new(false, false));
+        };
+
+        let transactions = bloom::transactions(ctx, &page, &filter, cp_bounds).await?;
```
wlmyng, line 349: beware, only madness lies ahead

crates/sui-indexer-alt-e2e-tests/tests/graphql/events/scan.move
```diff
@@ -0,0 +1,169 @@
...
+}
+
+fragment EV on EventConnection {
+  pageInfo { startCursor endCursor hasPreviousPage hasNextPage }
+  edges { cursor node { sequenceNumber sender { address } contents { type { repr } } } }
+}
+
+//# run-graphql --cursors {"c":1,"t":3,"e":0} {"c":5,"t":7,"e":0} {"c":5,"t":7,"e":2} {"c":20,"t":20,"e":0}
+# Cursor pagination: cursor_0 is cp1 event (t=3,e=0),
+# cursor_1 is cp5 first event (t=7,e=0), cursor_2 is cp5 last event (t=7,e=2).
```
wlmyng, line 152: having written cursor pagination tests, i recommend moving these into rust, otherwise it's quite brittle

crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
```diff
@@ -51,12 +53,7 @@ impl EventFilter {
     /// Uses the provided transaction bounds subquery to limit results to a specific transaction range
     pub(crate) fn query<'q>(&self) -> Result<Query<'q>, RpcError> {
         let table = match (&self.module, &self.type_) {
-            (Some(_), Some(_)) => {
```
amnn, line 54: IIUC, this has gone away because we now have custom validation on the code path used by `Query.events` pagination, but we probably still want to introduce an internal error here, so that we can catch cases where an `EventFilter` that is constructed "by hand" gets used to generate a postgres query.

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -41,6 +43,8 @@ use crate::task::watermark::Watermarks;
 pub(crate) mod filter;
 mod lookups;
 
+pub(crate) type CScanEvent = JsonCursor<ScanEventCursor>;
```
amnn, line 46: nit: the ordering is a bit weird here: alias for scan cursor, regular cursor, alias for regular cursor, scan cursor.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+}
+
+impl CpBoundsCursor for CScanEvent {
+    fn cp_sequence_number(&self) -> u64 {
+        self.cp_sequence_number
+    }
+}
+
+pub(super) type TransactionsBySequenceNumbers =
```
amnn, line 61: nit: file order

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -0,0 +1,413 @@
...
+    let sequenced_tx_digests: Vec<_> = checkpoints
+        .into_values()
+        .flat_map(|(summary, content, _)| {
+            let cp_seq = summary.sequence_number;
+            content
+                .enumerate_transactions(&summary)
+                .map(move |(tx_seq, &ExecutionDigests { transaction, .. })| {
+                    (tx_seq, cp_seq, transaction)
+                })
+                .collect::<Vec<_>>()
```
amnn, line 93: nit: This collect should not be required (`flat_map` can return an iterator)

amnn (commented):
Was mostly just refreshing my understanding of this PR, but spotted some nits -- (still) looks good!

amnn:
Probably safe to close these now -- we're well into exploring the bitmap-based strategy.

PR #25718, opened 2026-03-05
graphql-alt: ensure full walk of scanning pagination

henryachen:
## Problem
Bloom filters were built by tracking addresses only, not the full filter criteria (module, function, event type, etc.). A bloom filter match told you "this checkpoint contains a transaction or event from package X," but the actual query might filter on a specific module or function within that package. Combined with a one-shot candidate fetch, this caused pagination to terminate early for queries targeting rare events or functions on high-activity packages.

### Iterative bloom scanning with adaptive limits

Replace the one-shot bloom candidate fetch with `BloomScan`, an iterative scanner that keeps fetching batches until the page is filled or the checkpoint range is exhausted. The old approach used a fixed 3x overfetch multiplier and returned after a single batch — when the real hit rate was low this caused premature pagination termination.

The scanner uses adaptive batch sizing:
- **Cold start**: assumes ~10% bloom FPR and doubles the request size each round to handle unknown sparsity
- **Warm**: switches to the minimum observed per-batch hit rate as a pessimistic estimate, clamped to `[remaining, 500]` candidates per round

### More selective bloom values

Add values so queries can reject checkpoints at the module and function level, not just the package level. Single-byte prefix tags prevent collisions across dimensions that would otherwise share the same space.

**Bloom values inserted per checkpoint:**

| Dimension | Values | Tag |
|---|---|---|
| Sender / affected address | raw address bytes | (none) |
| Affected object | address bytes | `O` |
| Move call package | address bytes | `P` |
| Move call module | `{module}` | `M` |
| Event type package | address bytes | `A` |
| Event emitting module | `{module}` | `E` |
| Event type module | `{module}` | `T` |
| Function / type name | `{name}` | (none) |
| Event type parameter | canonical string (e.g. `0x2::sui::SUI`) | (none) |

Previously a query like `function: "0x2::coin::TYPO"` would only probe the bloom for the `0x2` address, matching nearly every checkpoint. With compound keys, the bloom also probes for the module and function name, rejecting checkpoints that don't have matching move calls.

### Bloom filter saturation impact

CP range 10M–13M (~3M checkpoints). Most bloom filters still fold down to 1KB.

**1 KB filters (8192 total bits, 99.9%+ of all filters):**

| Percentile | Old (bits) | New (bits) | Δ bits |
|---|---|---|---|
| p25 | 689 | 779 | +90 |
| p50 | 924 | 1064 | +140 |
| p75 | 1227 | 1403 | +176 |
| p90 | 1644 | 1834 | +190 |
| p95 | 2400 | 2470 | +70 |
| p99 | 3521 | 3631 | +110 |
| max | 5288 | 5274 | -14 |
| avg | 1065 | 1182 | +117 |

Average density: 13.0% → 14.4% (+1.4%). Well below the 40% fold threshold. 2KB+ filters (<0.1%) unchanged.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/events/scan.move
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/events/scan.move
@@ -18,6 +18,10 @@ module P1::M1 {
         value: u64,
     }
 
+    public struct GenericEvent<phantom T> has copy, drop {
+        value: u64,
+    }
+
     public entry fun emit_a(value: u64) {
         event::emit(EventA {
             message: ascii::string(b"Event A"),
@@ -30,6 +34,14 @@ module P1::M1 {
             message: ascii::string(b"Event B"),
             value,
         });
+    }
+
+    public entry fun emit_generic_bool(value: u64) {
+        event::emit(GenericEvent<bool> { value });
+    }
+
+    public entry fun emit_generic_u64(value: u64) {
+        event::emit(GenericEvent<u64> { value });
     }
 
     public entry fun emit_multiple(count: u64) {
@@ -87,6 +99,16 @@ module P2::M2 {
 
 //# create-checkpoint
 
+// Checkpoint 6: A emits GenericEvent<bool>
+//# run P1::M1::emit_generic_bool --sender A --args 10
+
+//# create-checkpoint
+
+// Checkpoint 7: B emits GenericEvent<u64>
+//# run P1::M1::emit_generic_u64 --sender B --args 20
+
+//# create-checkpoint
+
 //# advance-epoch
 
 // Generate 10,501 checkpoints to create 10 complete blocks + 500 extra for incomplete block
@@ -101,24 +123,24 @@ module P2::M2 {
 # Scan vs indexed cross-check, multi-filter combinations, empty-result edge cases,
 # checkpoint-bounded scan, and scan across bloom blocks.
 {
-  scanEventsA: scanEvents(filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+  scanEventsA: scanEvents(filter: { sender: "@{A}", beforeCheckpoint: 10512 }) { ...EV }
   paginateEventsA: events(filter: { sender: "@{A}" }) { ...EV }
 
   eventsAFromP1: scanEvents(filter: {
     sender: "@{A}",
     module: "@{P1}",
-    beforeCheckpoint: 10510
+    beforeCheckpoint: 10512
   }) { ...EV }
   scanTripleFilter: scanEvents(filter: {
     sender: "@{A}",
     module: "@{P1}",
     type: "@{P1}::M1::EventA",
-    beforeCheckpoint: 10510
+    beforeCheckpoint: 10512
   }) { ...EV }
 
   emptyNonExistent: scanEvents(filter: {
     sender: "0x0000000000000000000000000000000000000000000000000000000000000001",
-    beforeCheckpoint: 10510
+    beforeCheckpoint: 10512
   }) { ...EV }
 
   earlyEvents: scanEvents(filter: {
@@ -135,10 +157,30 @@ module P2::M2 {
     beforeCheckpoint: 6
   }) { ...EV }
 
+  scanByTypeOnly: scanEvents(filter: {
+    type: "@{P2}::M2::EventC",
+    beforeCheckpoint: 6
+  }) { ...EV }
+
   scanAcrossBlocks: scanEvents(filter: {
     sender: "@{A}",
     afterCheckpoint: 0,
-    beforeCheckpoint: 10510
+    beforeCheckpoint: 10512
+  }) { ...EV }
+
+  # Primitive type param filtering: GenericEvent<bool> vs GenericEvent<u64>
+  scanGenericBool: scanEvents(filter: {
+    type: "@{P1}::M1::GenericEvent<bool>",
+    beforeCheckpoint: 10512
+  }) { ...EV }
+  scanGenericU64: scanEvents(filter: {
+    type: "@{P1}::M1::GenericEvent<u64>",
+    beforeCheckpoint: 10512
+  }) { ...EV }
+  # Without type params, should return both GenericEvent<bool> and GenericEvent<u64>
+  scanGenericAny: scanEvents(filter: {
+    type: "@{P1}::M1::GenericEvent",
+    beforeCheckpoint: 10512
   }) { ...EV }
 }
 
@@ -151,16 +193,16 @@ fragment EV on EventConnection {
 # Cursor pagination: cursor_0 is cp1 event (t=3,e=0),
 # cursor_1 is cp5 first event (t=7,e=0), cursor_2 is cp5 last event (t=7,e=2).
 {
-  first2: scanEvents(first: 2, filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
-  last2: scanEvents(last: 2, filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+  first2: scanEvents(first: 2, filter: { sender: "@{A}", beforeCheckpoint: 10512 }) { ...EV }
+  last2: scanEvents(last: 2, filter: { sender: "@{A}", beforeCheckpoint: 10512 }) { ...EV }
 
-  firstAfter: scanEvents(first: 10, after: "@{cursor_0}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
-  firstBefore: scanEvents(first: 10, before: "@{cursor_2}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+  firstAfter: scanEvents(first: 10, after: "@{cursor_0}", filter: { sender: "@{A}", beforeCheckpoint: 10512 }) { ...EV }
+  firstBefore: scanEvents(first: 10, before: "@{cursor_2}", filter: { sender: "@{A}", beforeCheckpoint: 10512 }) { ...EV }
 
-  windowFirst: scanEvents(first: 10, after: "@{cursor_0}", before: "@{cursor_2}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+  windowFirst: scanEvents(first: 10, after: "@{cursor_0}", before: "@{cursor_2}", filter: { sender: "@{A}", beforeCheckpoint: 10512 }) { ...EV }
 
-  nonexistentCursor: scanEvents(last: 10, after: "@{cursor_3}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
-  invalidOrder: scanEvents(first: 10, after: "@{cursor_2}", before: "@{cursor_0}", filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
+  nonexistentCursor: scanEvents(last: 10, after: "@{cursor_3}", filter: { sender: "@{A}", beforeCheckpoint: 10512 }) { ...EV }
+  invalidOrder: scanEvents(first: 10, after: "@{cursor_2}", before: "@{cursor_0}", filter: { sender: "@{A}", beforeCheckpoint: 10512 }) { ...EV }
 }
 
 fragment EV on EventConnection {
--- a/crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
@@ -177,17 +177,17 @@ impl EventFilter {
 
         if let Some(m) = &self.module {
             let pkg: AccountAddress = m.package().into_bytes().into();
-            values.push(BloomValue::EvAddress(pkg));
+            values.push(BloomValue::EventAddress(pkg));
             if let Some(mod_name) = m.module() {
-                values.push(BloomValue::EvEmitModule(mod_name.to_owned()));
+                values.push(BloomValue::EventEmitModule(mod_name.to_owned()));
             }
         }
 
         if let Some(t) = &self.type_ {
             let pkg: AccountAddress = t.package().into_bytes().into();
-            values.push(BloomValue::EvAddress(pkg));
+            values.push(BloomValue::EventAddress(pkg));
             if let Some(mod_name) = t.module() {
-                values.push(BloomValue::EvTypeModule(mod_name.to_owned()));
+                values.push(BloomValue::EventTypeModule(mod_name.to_owned()));
                 if let Some(name) = t.type_name() {
                     values.push(BloomValue::Name(name.to_owned()));
                 }
@@ -201,7 +201,7 @@ impl EventFilter {
 
         values
             .into_iter()
-            .filter(|v| !v.should_skip())
+            .filter(|v| !v.exclude())
             .map(|v| v.to_bytes())
             .collect()
     }
--- a/crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
@@ -7,7 +7,8 @@ use anyhow::Context as _;
 use async_graphql::CustomValidator;
 use async_graphql::InputObject;
 use async_graphql::InputValueError;
-use sui_indexer_alt_schema::blooms::should_skip_for_bloom;
+use move_core_types::account_address::AccountAddress;
+use sui_indexer_alt_schema::blooms::BloomValue;
 use sui_pg_db::query::Query;
 use sui_sql_macro::query;
 use sui_types::event::Event as NativeEvent;
@@ -167,16 +168,42 @@ impl EventFilter {
     }
 
     /// Values to probe in bloom filters.
-    pub(crate) fn bloom_probe_values(&self) -> Vec<[u8; 32]> {
-        [
-            self.sender.map(|s| s.into_bytes()),
-            self.module.as_ref().map(|m| m.package().into_bytes()),
-            self.type_.as_ref().map(|t| t.package().into_bytes()),
-        ]
-        .into_iter()
-        .flatten()
-        .filter(|v| !should_skip_for_bloom(v))
-        .collect()
+    pub(crate) fn bloom_probe_values(&self) -> Vec<Vec<u8>> {
+        let mut values: Vec<BloomValue> = Vec::new();
+
+        if let Some(sender) = self.sender {
+            values.push(BloomValue::SenderOrRecipient(AccountAddress::from(sender)));
+        }
+
+        if let Some(m) = &self.module {
+            let pkg: AccountAddress = m.package().into_bytes().into();
+            values.push(BloomValue::EvAddress(pkg));
+            if let Some(mod_name) = m.module() {
+                values.push(BloomValue::EvEmitModule(mod_name.to_owned()));
+            }
+        }
+
+        if let Some(t) = &self.type_ {
+            let pkg: AccountAddress = t.package().into_bytes().into();
+            values.push(BloomValue::EvAddress(pkg));
+            if let Some(mod_name) = t.module() {
+                values.push(BloomValue::EvTypeModule(mod_name.to_owned()));
+                if let Some(name) = t.type_name() {
+                    values.push(BloomValue::Name(name.to_owned()));
+                }
+            }
+            if let Some(type_params) = t.type_params() {
+                for tp in type_params {
+                    values.push(BloomValue::TypeParam(tp.to_canonical_string(false)));
+                }
+            }
+        }
+
+        values
+            .into_iter()
+            .filter(|v| !v.should_skip())
+            .map(|v| v.to_bytes())
+            .collect()
     }
 }
 
--- a/crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
@@ -39,7 +39,7 @@ use crate::scope::Scope;
 
 mod scan;
 
-struct CandidateTxn {
+struct SequencedTxnDigest {
     cp_sequence_number: u64,
     tx_sequence_number: u64,
     digest: TransactionDigest,
@@ -76,7 +76,7 @@ pub(crate) async fn transactions(
     let mut result = BTreeMap::new();
 
     while let Some(candidates) = scan.next(pg_reader, &filter_values).await? {
-        let txns = candidate_txns(kv_loader, candidates.candidates()).await?;
+        let txns = sequenced_txn_digests(kv_loader, candidates.candidates()).await?;
         let digests = txns.iter().map(|t| t.digest).collect();
         let mut transactions_by_digest = kv_loader
             .load_many_transactions(digests)
@@ -122,7 +122,7 @@ pub(crate) async fn events(
     let mut result = BTreeMap::new();
 
     while let Some(candidates) = scan.next(pg_reader, &filter_values).await? {
-        let txns = candidate_txns(kv_loader, candidates.candidates()).await?;
+        let txns = sequenced_txn_digests(kv_loader, candidates.candidates()).await?;
         let digests = txns.iter().map(|t| t.digest).collect();
         let events_by_digest = kv_loader
             .load_many_transaction_events(digests)
@@ -163,10 +163,10 @@ pub(crate) async fn events(
 }
 
 /// Load checkpoints for the given candidate CPs to get transaction digests with checkpoint and transaction sequence numbers.
-async fn candidate_txns(
+async fn sequenced_txn_digests(
     kv_loader: &KvLoader,
     candidate_cps: &[u64],
-) -> Result<Vec<CandidateTxn>, RpcError> {
+) -> Result<Vec<SequencedTxnDigest>, RpcError> {
     let checkpoints = kv_loader
         .load_many_checkpoints(candidate_cps.to_vec())
         .await
@@ -178,7 +178,7 @@ async fn candidate_txns(
             content
                 .enumerate_transactions(&summary)
                 .map(
-                    move |(tx_seq, &ExecutionDigests { transaction, .. })| CandidateTxn {
+                    move |(tx_seq, &ExecutionDigests { transaction, .. })| SequencedTxnDigest {
                         tx_sequence_number: tx_seq,
                         cp_sequence_number: cp_seq,
                         digest: transaction,
@@ -276,8 +276,7 @@ pub(super) async fn candidate_cps(
         limit as i64,
     );
 
-    // For each matched block, scan cp_blooms until we have adjusted_limit checkpoints that
-    // match the probe.
+    // Scan cp_blooms within matched blocks to find up to `limit` checkpoints matching the probe.
     let query = query!(
         r#"
         WITH
--- a/crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
@@ -195,7 +195,7 @@ async fn candidate_txns(
 /// then a finer filter over those ranges for checkpoint matches using cp_blooms.
 pub(super) async fn candidate_cps(
     pg_reader: &PgReader,
-    filter_values: &[[u8; 32]],
+    filter_values: &[Vec<u8>],
     cp_lo: u64,
     cp_hi_inclusive: u64,
     is_from_front: bool,
--- a/crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
@@ -13,7 +13,6 @@ use diesel::sql_types::Integer;
 use diesel::sql_types::SmallInt;
 use sui_indexer_alt_reader::kv_loader::KvLoader;
 use sui_indexer_alt_reader::kv_loader::TransactionContents;
-use sui_indexer_alt_reader::kv_loader::TransactionEventsContents;
 use sui_indexer_alt_reader::pg_reader::PgReader;
 use sui_indexer_alt_schema::blooms::blocked::BlockedBloomProbe;
 use sui_indexer_alt_schema::blooms::bloom::BloomProbe;
@@ -32,17 +31,21 @@ use crate::api::types::event::ScanEventCursor;
 use crate::api::types::event::filter::EventFilter;
 use crate::api::types::transaction::CScanTransaction;
 use crate::api::types::transaction::ScanTransactionCursor;
+use crate::api::types::transaction::bloom::scan::BloomScan;
 use crate::api::types::transaction::filter::TransactionFilter;
 use crate::error::RpcError;
 use crate::pagination::Page;
 use crate::scope::Scope;
 
-/// Multiplier to page limit to adjust for bloom filter false positives.
-const OVERFETCH_MULTIPLIER: f64 = 3.0;
+mod scan;
 
-pub(super) type EventsBySequenceNumbers = BTreeMap<ScanEventCursor, Event>;
+struct CandidateTxn {
+    cp_sequence_number: u64,
+    tx_sequence_number: u64,
+    digest: TransactionDigest,
+}
 
-pub(crate) trait CpBoundsCursor {
+pub(super) trait CpBoundsCursor {
     fn cp_sequence_number(&self) -> u64;
 }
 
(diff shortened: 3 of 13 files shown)


crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,140 @@
...
+const OVERFETCH_MULTIPLIER: f64 = 5.0;
+
+/// Approximate true positive rate of the bloom filter (1 - FPR). With a ~70% FPR,
+/// ~30% of bloom-matched candidates actually contain the filter value.
+const BLOOM_TRUE_POSITIVE_RATE: f64 = 0.30;
+
+/// Maximum number of SQL+KV iterations before returning partial results. Prevents unbounded
+/// scanning when filters match no or very few results (e.g. nonexistent event types).
+/// Estimated to cap at ~5s assuming ~50ms per iteration at max page size.
+const MAX_SCAN_ITERATIONS: usize = 100;
```
henryachen: This is effectively another scan limit. I added this to prevent runaway queries when users filter for non existent types or functions using existing packages or modules. 

I think this can be removed since I added a package resolver check to the scan to only scan when the module or event type referenced by an event filter exists on-chain.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -399,15 +411,77 @@ fn cp_bloom_check_sql(probe: &BloomProbe) -> Query<'static> {
...
-        c.cp_sequence_number().max(*cp_bounds.start())
-    });
-    let cp_hi_inclusive = page.before().map_or(*cp_bounds.end(), |c| {
-        c.cp_sequence_number().min(*cp_bounds.end())
-    });
-    (cp_lo, cp_hi_inclusive)
+/// Validates that the module/function referenced by a transaction filter exists on-chain.
+/// Returns false if the module or function doesn't exist within a successfully loaded package,
+/// allowing callers to short-circuit and return empty results. Returns true if the package
+/// cannot be fetched (e.g. kv_packages not populated), so the scan proceeds normally.
```
wlmyng: then, operators are protected from runaway queries only if they also index `kv_packages`?

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,140 @@
...
+
+use super::CpBoundsCursor;
+
+/// Safety margin multiplier applied to hit-rate estimates to account for variance
+/// in the observed rate.
+const OVERFETCH_MULTIPLIER: f64 = 5.0;
+
+/// Approximate true positive rate of the bloom filter (1 - FPR). With a ~70% FPR,
+/// ~30% of bloom-matched candidates actually contain the filter value.
+const BLOOM_TRUE_POSITIVE_RATE: f64 = 0.30;
```
wlmyng: bloom filters have a 100% true positive rate though, right? I think precision is more accurate, we could just use ESTIMATED_HIT_RATE. is this from your observations specific to `0xdee9` or just a really low precision that gets updated as we get live hit rates from scanning

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// When no hits have been observed, accounts for the bloom FPR and doubles the limit
+    /// for the unknown real hit density:
+    ///   Round 0: 52 / 0.30 = 173
+    ///   Round 1: 52 / 0.30 * 2 = 347
+    ///   Round 2: 52 / 0.30 * 4 = 693 → clamped to MAX_LIMIT (500)
+    ///
+    /// Once hits are observed, uses the end-to-end hit rate (bloom TPR * hit density)
+    /// with a safety margin:
+    ///   5 hits from 260 seen → hit_rate ≈ 0.02, remaining = 47
+    ///   estimate = 47 / 0.02 * 5 = 11,750 → clamped to 500
```
wlmyng:
```
   /// Estimates how many bloom-matched candidate checkpoints to request based on the current precision.
      ///                                                                                                                                               
      /// Before any hits are observed, assumes the STARTING_HIT_RATE and                                                                            
      /// doubles the request size each round to handle unknown sparsity.               
      ///  
      /// Once hits are observed, switches to the empirical hit rate with
      /// an overfetch multiplier to overfetch and reduce the chance of needing another round.
      ///
      /// The result is always clamped to `[remaining, MAX_LIMIT]` — at least
      /// enough to fill the page at 100% hit rate, and at most MAX_LIMIT to
      /// bound per-round work.
```

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
henryachen: True, i'm currently testing adding more precision to the bloom filter with `pkg::mod` and `pkg::mod::name`.  This will handle the runaway cases like typos in module names.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
henryachen: Yeah, this is misleading. I think it makes sense to change this to  `BLOOM_FPR` (1%) to separate the effects of bloom filter FPR and FPR from precision from the values inserted into the bloom filter. The initial intention of this number was to use the bloom filter FPR to overfetch while `limit` below accounts for the precision issues from the values inserted into the filter.

crates/sui-indexer-alt-schema/src/blooms/mod.rs
```diff
@@ -2,34 +2,165 @@
...
-/// - Cause queries to match nearly all blocks and checkpoints
-/// - Require fetching and probing more bloom filter rows at both levels
+/// checkpoints, so including them would cause queries to match nearly all blocks.
 const BLOOM_SKIP_ADDRESSES: &[AccountAddress] = &[AccountAddress::ZERO, SUI_CLOCK_ADDRESS];
 
-/// Returns true if the bytes should be excluded from bloom filter operations.
-pub fn should_skip_for_bloom(bytes: &[u8]) -> bool {
-    BLOOM_SKIP_ADDRESSES.iter().any(|id| id.as_ref() == bytes)
+/// Single-byte prefix tags for bloom key dimensions that would otherwise collide.
+/// Not every `BloomValue` variant is tagged — untagged variants are unlikely to
```
tpham-mysten, line 17: Is this due to disk space concern? Having tag for every variant might still be beneficial for debugging (i.e. we instantly know the type by looking at the prefix)

crates/sui-indexer-alt-schema/src/blooms/mod.rs
```diff
@@ -2,34 +2,165 @@
...
+    /// Transaction sender or affected address (recipients, gas owner, etc.).
+    SenderOrRecipient(AccountAddress),
+    /// Object ID mutated, created, or deleted by a transaction.
+    AffectedObject(AccountAddress),
+    /// Move call package address.
+    MoveCallPackage(AccountAddress),
+    /// Move call module name.
+    MoveCallModule(String),
+    /// Event emitting or type package address.
+    EvAddress(AccountAddress),
```
tpham-mysten: nit: I personally prefer using the full "Event" prefix. It's fine here with documentation but could be not quite readable when using alone.

crates/sui-indexer-alt-schema/src/blooms/mod.rs
```diff
@@ -2,34 +2,165 @@
...
+            Self::EvEmitModule(module) => BloomTag::EvEmitModule.prefix(module.as_bytes()),
+            Self::EvTypeModule(module) => BloomTag::EvTypeModule.prefix(module.as_bytes()),
+            Self::Name(name) => name.into_bytes(),
+            Self::TypeParam(s) => s.into_bytes(),
+        }
+    }
+
+    /// Returns true if this value should be excluded from bloom filter operations
+    /// because it matches a high-frequency address.
+    pub fn should_skip(&self) -> bool {
```
tpham-mysten: optional nit: I personally found should_skip is a bit vague (i.e. skip which one), but maybe it's just me.

Maybe something like `should_exclude_from_bloom` is more explicit but can be verbose.

crates/sui-indexer-alt-schema/src/blooms/mod.rs
```diff
@@ -2,34 +2,165 @@
...
+    ///
+    /// The top-level struct gets three keys at increasing specificity:
+    ///   - `EvAddress(pkg)`        — matches queries filtering by package only
+    ///   - `EvTypeModule(mod)`       — matches queries filtering by module
+    ///   - `Name(name)`             — matches queries filtering by type name
+    ///
+    /// Each type parameter is inserted as its canonical string representation
+    /// (e.g. "bool", "u64", "0x2::sui::SUI"). This handles all TypeTag variants
+    /// including primitives and nested generics.
+    pub fn from_struct_tag(tag: &StructTag) -> Vec<BloomValue> {
```
tpham-mysten: nit: from_event_struct_tag?

crates/sui-indexer-alt/src/handlers/cp_blooms.rs
```diff
@@ -169,56 +156,6 @@ mod tests {
         );
     }
 
-    #[tokio::test]
-    async fn test_cp_blooms_with_function_calls() {
```
tpham-mysten, line 173: Why do we remove these 2 tests?

crates/sui-indexer-alt-schema/src/blooms/mod.rs
```diff
@@ -2,34 +2,165 @@
...
+}
+
+/// Bloom filter values, each variant represents a distinct dimension
+/// that can be inserted into or probed against a checkpoint bloom filter.
+/// Some bloom filter values are prepended with tags to prevent collisions across
+/// dimensions, e.g. a Move call package address colliding with an event type module address.
+#[derive(Debug)]
+pub enum BloomValue {
+    /// Transaction sender or affected address (recipients, gas owner, etc.).
+    SenderOrRecipient(AccountAddress),
```
tpham-mysten, line 44: Why don't we have one enum for Sender and one enum for Recipient? 

With this, is it true that probing for sent_address: "0xA" will match blooms where 0xA appears as a recipient too?

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,169 @@
...
+
+use crate::api::types::transaction::bloom::CpBoundsCursor;
+use crate::api::types::transaction::bloom::candidate_cps;
+
+/// Approximate false positive rate of the bloom filter. Used to inflate the initial
+/// candidate fetch size before any real hit-rate data is available.
+const BLOOM_FPR: f64 = 0.1;
+
+/// Maximum number of bloom-matched candidate checkpoints to fetch in a single batch.
+const MAX_LIMIT: usize = 500;
```
tpham-mysten: nit:maybe MAX_CHECKPOINT_FETCH is clearer?

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,169 @@
...
+
+use crate::error::RpcError;
+use crate::pagination::Page;
+
+use crate::api::types::transaction::bloom::CpBoundsCursor;
+use crate::api::types::transaction::bloom::candidate_cps;
+
+/// Approximate false positive rate of the bloom filter. Used to inflate the initial
+/// candidate fetch size before any real hit-rate data is available.
+const BLOOM_FPR: f64 = 0.1;
```
tpham-mysten, line 16: Just curious how do you come up with 10%. It seems to be quite a conservative estimate. 

Also, It would be great if you can add some tuning tips doc on this variable for future maintainence.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.hits += batch_hits;
+
+        let Some(&last_cp) = candidates.last() else {
+            return;
+        };
+
+        if self.is_from_front {
+            self.cp_lo = last_cp.saturating_add(1);
+        } else {
+            self.cp_hi_inclusive = last_cp.saturating_sub(1);
```
tpham-mysten: For edge case scenario like: backward scan, cp_lo = 0, cp_hi_inclusive = 0, and candidate_cps returns 0.

IIUC,  0u64.saturating_sub(1) will keep returning 0 and potentially cause an infinite loop.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,169 @@
...
+/// each batch of candidates returned by [`next`](Self::next).
+///
+/// ```ignore
+/// while let Some(batch) = scan.next(&pg_reader, &filter_values).await? {
+///     let hits = process(batch.candidates());
+///     batch.record(hits);
+/// }
+/// ```
+pub(super) struct BloomScan {
+    cp_lo: u64,
```
tpham-mysten, line 33: nit: Maybe I missed some convention, but there is an inconsistency between cp_lo and cp_hi_inclusive, while they are both inclusive. Is it intended?

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -58,161 +61,153 @@ impl CpBoundsCursor for CScanEvent {
...
-    if candidate_cps.is_empty() {
-        return Ok(BTreeMap::new());
+        candidates.record(batch_hits);
     }
 
+    Ok(result)
+}
+
+/// Load checkpoints for the given candidate CPs to get transaction digests with checkpoint and transaction sequence numbers.
+async fn candidate_txns(
```
tpham-mysten: nit: load_candidate_txns_digest is more clear imo

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -271,8 +264,12 @@ async fn candidate_cps<C>(
...
-        adjusted_limit,
+        if is_from_front {
+            query!("ASC")
+        } else {
+            query!("DESC")
+        },
+        limit as i64,
     );
 
     // For each matched block, scan cp_blooms until we have adjusted_limit checkpoints that
```
tpham-mysten: nit: Stale comment on adjusted_limit

tpham-mysten (commented):
The Adaptive approach is very nice and looking great overall! Most comments are just nits and for clarification

crates/sui-indexer-alt-e2e-tests/tests/graphql/events/scan.move
```diff
@@ -101,24 +123,24 @@ module P2::M2 {
 # Scan vs indexed cross-check, multi-filter combinations, empty-result edge cases,
 # checkpoint-bounded scan, and scan across bloom blocks.
 {
-  scanEventsA: scanEvents(filter: { sender: "@{A}", beforeCheckpoint: 10510 }) { ...EV }
```
evan-wall-mysten, line 104: Does changing `10510` -> `10512` test that "premature pagination termination" is not happening?

evan-wall-mysten:
> Replace the one-shot bloom candidate fetch with BloomScan, an iterative scanner that keeps fetching batches until the page is filled or the checkpoint range is exhausted. The old approach used a fixed 3x overfetch multiplier and returned after a single batch — when the real hit rate was low this caused premature pagination termination.
>
> The scanner uses adaptive batch sizing:
> ...

I had a few high level questions since I am not that familiar with the bloom filter code:

1. Does "premature pagination termination" result in results that can never be returned after the termination?
2. Is fix for (1) that `BloomScan` is now iterative instead of one-shot, regardless of "adaptive batch sizing"?
3. Is "adaptive batch sizing" an optimization on top of the fix in (2)?

> More selective bloom values
> ...

4. Are the "more selective bloom values" a fix unrelated to the above to address irrelevant results being returned by a query like `0x2::coin::TYPO`? Meaning we want `0x2::coin::TYPO` to return nothing instead of unrelated coins at that address? 
5. Were the results in (4) returned to the user or just returned by the bloom filter and filtered out by the GraphQL server?

crates/sui-indexer-alt/src/handlers/cp_blooms.rs
```diff
@@ -72,55 +73,41 @@ impl Handler for CpBlooms {
...
+
+    for addr in affected_addresses(&tx.effects) {
+        values.push(BloomValue::SenderOrRecipient(addr.into()));
+    }
+
+    for change in tx.effects.object_changes() {
+        values.push(BloomValue::AffectedObject(change.id.into()));
+    }
+
+    for (_, package_id, module, function) in tx.transaction.move_calls() {
```
wlmyng, line 106: it would be nice to capture some of the reasoning that led us to choosing to capture the individual components of a type tag, I can imagine someone would be curious why we didn't index `pkg` + `pkg::module` + `pkg::module::fn`

crates/sui-indexer-alt/src/handlers/cp_blooms.rs
```diff
@@ -6,14 +6,15 @@ use std::sync::Arc;
 use anyhow::Result;
 use async_trait::async_trait;
 use diesel_async::RunQueryDsl;
+use move_core_types::account_address::AccountAddress;
```
wlmyng, line 9: do we have to use the `AccountAddress` from `move_core_types`? is there some other address related type already pulled in as a dependency

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -58,161 +61,153 @@ impl CpBoundsCursor for CScanEvent {
     }
 }
 
-pub(super) type TransactionsBySequenceNumbers =
-    BTreeMap<ScanTransactionCursor, (TransactionDigest, TransactionContents)>;
-
+/// Scans a checkpoint range for transactions matching a filter using bloom-based
+/// candidate selection until it has filled a page of results that match the filter.
```
wlmyng, line 65:
```suggestion
/// candidate selection until it has filled a page of results.
```

to circle back: there is a maximum number of times we try to fill up a page?

crates/sui-indexer-alt-schema/src/blooms/mod.rs
wlmyng: I suppose that should be possible already:
- when looking at a 32 byte entry, absence of any tag indicates `SenderOrRecipient`

Actually, I think it's just the function/ type name and event type params that don't have a bloom tag? Which the last sentence touches on hmm

crates/sui-indexer-alt-schema/src/blooms/mod.rs
wlmyng: Or just `exclude`, but I think the main thing is that the doc comment is clear on the method is supposed to do

Personally though, I think callers should not need to remember to filter their bloom with `should_skip`, when instantiating through a constructor, the constructor should check whether to `exclude` itself and return an `Option<Self>`

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,169 @@
...
+            min_batch_hit_rate: None,
+        }
+    }
+
+    /// Fetches the next batch of bloom-matched checkpoints, sized based on the hit rate
+    /// observed so far. Bloom filters operate on the values inserted and values inserted
+    /// from busy packages may cause a high false positive rate for other items,
+    /// so candidates include false positives — low hit rates cause larger fetches to compensate.
+    ///
+    /// Returns `None` when the page is filled or the checkpoint window is exhausted.
```
wlmyng:
```suggestion
    /// Fetches the next batch of bloom-matched checkpoints. The current hit rate influences the number of checkpoints fetched in a batch.
    ///
    /// Returns `None` when the page is filled or the checkpoint window is exhausted.
```

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
wlmyng: `next` checks `cp_lo > cp_hi_inclusive`, so it seems the loop is only terminated if `candidate_cps(0, 0)` returns nothing, otherwise it does seem we end up in an infinite loop. `cp_lo == cp_hi_inclusive` is valid (such as dealing with `atCheckpoint`), so one way to avoid this is to make `cp_hi` exclusive?

Would mean adjustments to the SQL query and initialization of the struct... 

Or add a flag + use it in `next`

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,169 @@
...
+        &'a mut self,
+        pg_reader: &PgReader,
+        filter_values: &[Vec<u8>],
+    ) -> Result<Option<Candidates<'a>>, RpcError> {
+        if self.hits >= self.page_limit || self.cp_lo > self.cp_hi_inclusive {
+            return Ok(None);
+        }
+
+        let limit = self.limit();
+        let candidates = candidate_cps(
```
wlmyng, line 97: or, it seems we can avoid the underflow/ infinite loop error if we check `cp_lo == cp_hi_inclusive` after calling `candidate_cps`. This is the last batch regardless of direction?

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,169 @@
...
+const MAX_LIMIT: usize = 500;
+
+/// Scan state for paginating through bloom-filtered checkpoint candidates.
+///
+/// Tracks the current checkpoint window, hit rate, and iteration count to size
+/// each batch of candidates returned by [`next`](Self::next).
+///
+/// ```ignore
+/// while let Some(batch) = scan.next(&pg_reader, &filter_values).await? {
+///     let hits = process(batch.candidates());
```
wlmyng, line 28: this feels a bit hot potatoey... `next` already has `candidates` before wrapping it in `Candidates`. could we not stash `candidates.len()` and call `candidates.last()` to determine the next checkpoint range?

Then `record` just needs to take `hits`. I don't think we need to pass in `candidates` since we just need its length and `last_cp`

wlmyng (commented):
Some suggestions in the bloom implementation!

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
henryachen: Yes, in graphql and db schema the convention is that `cp_lo` is inclusive and `cp_hi` is exclusive so whenever we have an inclusive `cp_hi`.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
henryachen: I wanted `BloomScan` to represent the settled state of the scan at any point, so I wanted to keep intermidate state out of the struct. But I guess if we added something like `last_batch_size/last_cp` the `&mut borrow` enforces that these values cannot be used in a stale state 🤔

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
henryachen: Ah thank you - I think I was going back and forth between `cp_lo` and `cp_hi` but forgot why I ended up on `cp_hi_inclusive`. Will use `cp_lo` and `cp_hi`.

crates/sui-indexer-alt-schema/src/blooms/mod.rs
henryachen: Yes, this affects the size of blooms - trading off precision for smaller blooms. I've been very careful to not add to much bloat to bloom if we can avoid it, `SenderOrRecipient` is one of those - technically we can add it but it didn't add much value during testing.

crates/sui-indexer-alt-schema/src/blooms/mod.rs
henryachen: I like the renaming to `exclude`. I want to leave this as a follow up to improve devx around exclusions. I think there is an option to fold exclusion into `to_bytes()` -> `Option<Vec<u8>>` — forces callers to handle `None`. But that couples serialization with filtering. For now I want to keep this as a follow up instead of changing to much in this PR.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
henryachen: How about `sequenced_txn_digests`? Usually we avoid verbs like load / get / compute in the codebase

crates/sui-indexer-alt-e2e-tests/tests/graphql/events/scan.move
henryachen: No, this was just adjusting the number of checkpoints to create, i'll add a separate snapshot test to check the premature pagination termination case.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
henryachen: No, that would effectively be the new scan limit if we did that - the ideal case is we add bloom values to handle cases that cause runaway queries if we encounter them.

crates/sui-indexer-alt/src/handlers/cp_blooms.rs
henryachen: `AccountAddress` is the common base type for `SuiAddress` `ObjectID` `StructTag.address` so it felt more natural here or else it would be messy to handle `SuiAddress`, `ObjectID` and `StructTag` separately.

crates/sui-indexer-alt/src/handlers/cp_blooms.rs
henryachen: The existing tests covered these more specifically (`test_cp_blooms_merge_on_conflict`, `scan.move`) so I removed as clean up!

henryachen:
> 1. Does "premature pagination termination" result in results that can never be returned after the termination?
> 2. Is fix for (1) that `BloomScan` is now iterative instead of one-shot, regardless of "adaptive batch sizing"?
> 3. Is "adaptive batch sizing" an optimization on top of the fix in (2)?

Yes. The old one-shot fetch would get N bloom candidates, apply the real filter, and if too few matched, return a short page with `hasNextPage: false` — even though real matches existed in later checkpoints, those results were silently lost to the user. Right adaptive batch sizing is to help get to a full page faster based on the number of matches so far.

>
> 4. Are the "more selective bloom values" a fix unrelated to the above to address irrelevant results being returned by a query like `0x2::coin::TYPO`? Meaning we want `0x2::coin::TYPO` to return nothing instead of unrelated coins at that address?
> 5. Were the results in (4) returned to the user or just returned by the bloom filter and filtered out by the GraphQL server?

The more selective bloom values allow us to return a smaller set of "matches": without them, querying `0x2::coin::TYPO` would bloom-match every checkpoint that has any `0x2` activity (very common), then the real filter rejects all of them. With compound keys (package + module + name), the bloom itself rejects most of those checkpoints, so the blooms return 0 matches and we don't have to enter the loop. In the case of a hot package with rare activity this also help reduce the set of candidate checkpoints we need to check / reject. As the bloom values get more selective the set of candidates get smaller but we still need the iterative scan for any edge cases.

crates/sui-indexer-alt/src/handlers/cp_blooms.rs
henryachen: Added a comment to `BloomValue`!

```
/// Bloom filter values indexed as individual components (package, module, name) rather than
/// compound keys (pkg::module::name). AND-probing N separate components requires N*k bloom
/// bits to match, giving better selectivity than a single compound probe (k bits).
///
```

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -58,168 +61,165 @@ impl CpBoundsCursor for CScanEvent {
...
+    Ok(result)
+}
+
+/// Load checkpoints for the given candidate CPs to get transaction digests with checkpoint and transaction sequence numbers.
+async fn sequenced_txn_digests(
+    kv_loader: &KvLoader,
+    candidate_cps: &[u64],
+) -> Result<Vec<SequencedTxnDigest>, RpcError> {
     let checkpoints = kv_loader
         .load_many_checkpoints(candidate_cps.to_vec())
```
nickvikeras, line 171: Assuming this calls into the grpc loader, each checkpoint is being loaded sequentially internally which seems likely to cause latency problems. 

https://github.com/MystenLabs/sui/blob/bc2c02c9f621ce810b98e991028524f4cf46eba0/crates/sui-indexer-alt-reader/src/checkpoints.rs?plain=1#L114-L146

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
```diff
@@ -58,168 +61,165 @@ impl CpBoundsCursor for CScanEvent {
...
+    let mut scan = BloomScan::new(page, &cp_bounds);
+    let mut result = BTreeMap::new();
 
-    if candidate_cps.is_empty() {
-        return Ok(BTreeMap::new());
+    while let Some(candidates) = scan.next(pg_reader, &filter_values).await? {
+        let txns = sequenced_txn_digests(kv_loader, candidates.candidates()).await?;
+        let digests = txns.iter().map(|t| t.digest).collect();
+        let events_by_digest = kv_loader
+            .load_many_transaction_events(digests)
```
nickvikeras, line 128: Same deal, these are loaded sequentially. 

https://github.com/MystenLabs/sui/blob/0a1e37d9f99f6ebe30a527fb3ffd95b0a1f0dd9b/crates/sui-indexer-alt-reader/src/events.rs?plain=1#L118-L122

Here I think you can just use BatchGetTransactions.

Obvs not a blocking comment since this isn't even the code you are writing/changing. Just calling it out so you are aware.

henryachen: Thanks! Just batched this up here https://github.com/MystenLabs/sui/pull/26118

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/mod.rs
henryachen: Thanks for the call out! The change was merged upstream https://github.com/MystenLabs/sui/pull/26118

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
henryachen: I had claude gather up all the testing and tuning scripts into `graphql-benchmark` https://github.com/MystenLabs/graphql-benchmark/pull/39. It's a ts cli for now but I'm thinking it would be better off as a rust crate for better reuse but this is still useful for initial tuning and exploration.

crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
```diff
@@ -167,16 +168,42 @@ impl EventFilter {
...
-            self.sender.map(|s| s.into_bytes()),
-            self.module.as_ref().map(|m| m.package().into_bytes()),
-            self.type_.as_ref().map(|t| t.package().into_bytes()),
-        ]
-        .into_iter()
-        .flatten()
-        .filter(|v| !should_skip_for_bloom(v))
-        .collect()
+    pub(crate) fn bloom_probe_values(&self) -> Vec<Vec<u8>> {
+        let mut values: Vec<BloomValue> = Vec::new();
```
amnn, line 172: This function could probably be materialising less:

- You don't need to create a vector of `BloomValue`s to then create a `Vec<Vec<u8>>`. Instead you could construct each `BloomValue` and convert it into bytes to put directly into a `Vec<Vec<u8>>`
- The variants that deal with identifiers could probably hold a `&str` instead of an owned type.
- You maybe also don't even need to create the `Vec<Vec<u8>>`, because you could implement this as:

```rust
pub(crate) fn add_bloom_probes(&self, probes: &mut impl Extend) {
    use BloomValue as BV;

    if let Some(sender) = self.sender {
        let value = BV::SenderOrRecipient(sender.into());
        probes.extend(value.include().then_some(value));
    }

    // ... and so on
}
```

(This suggestion also implies replacing `BloomValue::exclude` with a `BloomValue::include` to avoid a double negation in the filter).

crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
```diff
@@ -167,16 +168,42 @@ impl EventFilter {
...
+
+        if let Some(sender) = self.sender {
+            values.push(BloomValue::SenderOrRecipient(AccountAddress::from(sender)));
+        }
+
+        if let Some(m) = &self.module {
+            let pkg: AccountAddress = m.package().into_bytes().into();
```
amnn, line 179: nit: it would be good to avoid having to go from SuiAddress -> bytes -> AccountAddress -- could we define a `From<SuiAddress> for AccountAddress` for the GraphQL `SuiAddress` to avoid that? (I half expected us to already have one).

crates/sui-indexer-alt-graphql/src/api/types/event/filter.rs
```diff
@@ -167,16 +168,42 @@ impl EventFilter {
...
+            values.push(BloomValue::EventAddress(pkg));
+            if let Some(mod_name) = t.module() {
+                values.push(BloomValue::EventTypeModule(mod_name.to_owned()));
+                if let Some(name) = t.type_name() {
+                    values.push(BloomValue::Name(name.to_owned()));
+                }
+            }
+            if let Some(type_params) = t.type_params() {
+                for tp in type_params {
+                    values.push(BloomValue::TypeParam(tp.to_canonical_string(false)));
```
amnn, line 197: This is a bit of a departure from how we're encoding types elsewhere -- it would be more consistent (and byte efficient) to recursively traverse the types and add their addresses, modules, and names separately.

crates/sui-indexer-alt-schema/src/blooms/mod.rs
```diff
@@ -2,34 +2,167 @@
...
+}
+
+/// Bloom filter values indexed as individual components (package, module, name) rather than
+/// compound keys (pkg::module::name). AND-probing N separate components requires N*k bloom
+/// bits to match, giving better selectivity than a single compound probe (k bits).
+///
+/// Tags prevent collisions across dimensions (e.g. a Move call package address vs an event
+/// type module address).
+#[derive(Debug)]
+pub enum BloomValue {
```
amnn, line 42: I would recommend a more general encoding than this, something like:

```rust
pub enum Tag {
    Account,
    Object,
    Function,
    Module,
    Type,
}

pub enum BloomValue<'s> {
    Address(Tag, AccountAddress),
    Ident(&'s str),
}
```

This provides a more reusable set of primitives, and it shouldn't lose fidelity:

- We don't need to differentiate modules and types, because they don't look the same (because of casing, etc).
- We don't need to differentiate a module name used for an object type, vs for an event type, vs for a function, because the package address' annotation is already differentiated.
- Type parameters can re-use the same encoding as a "top-level" type -- if this doesn't work well, then you can add a special value variant for a `Type`'s address with a saturating depth:

```
Type(u8, AccountAddress),
```

The reusability should mean that future filters can be added without messing with the encoding.

crates/sui-indexer-alt-schema/src/blooms/mod.rs
```diff
@@ -2,34 +2,167 @@
...
+    EventTypeModule(String),
+    /// Function or type name (without package or module).
+    Name(String),
+    /// Event type parameter canonical string (e.g. "bool", "u64", "0x2::sui::SUI").
+    TypeParam(String),
+}
+
+impl BloomValue {
+    pub fn to_bytes(self) -> Vec<u8> {
+        match self {
```
amnn, line 65: Any reason why some of these get prefixes and some don't? I was thinking it could be because `SenderOrRecipient` is the most common case, but that may not still be true if we later add affected object type filtering.

crates/sui-indexer-alt-schema/src/blooms/mod.rs
```diff
@@ -2,34 +2,167 @@
...
+    ///
+    /// The top-level struct gets three keys at increasing specificity:
+    ///   - `EventAddress(pkg)`        — matches queries filtering by package only
+    ///   - `EventTypeModule(mod)`       — matches queries filtering by module
+    ///   - `Name(name)`             — matches queries filtering by type name
+    ///
+    /// Each type parameter is inserted as its canonical string representation
+    /// (e.g. "bool", "u64", "0x2::sui::SUI"). This handles all TypeTag variants
+    /// including primitives and nested generics.
+    pub fn from_event_struct_tag(tag: &StructTag) -> Vec<BloomValue> {
```
amnn, line 106: The suggestion above would allow you to generalise this function to generate a bloom filter probe from any struct tag (not specifically for an event).

amnn: Similar to the comment about generating values from filters, I'd recommend a pattern where this function accepts a structure to accumulate into, which would allow you to accumulate into the bloom filter directly, rather than create an intermediate array of values.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,200 @@
...
+            limit,
+        )
+        .await?;
+
+        Ok(if candidates.is_empty() {
+            None
+        } else {
+            self.last_batch_size = candidates.len();
+            self.last_cp = candidates.last().copied();
+            Some(Candidates {
```
amnn, line 112: This pattern of returning `Candidates` and then expecting the caller to use it to update the record is a little bit convoluted -- I think you might want to have this function accept a closure instead, to filter the checkpoints into a list of values.

That way this function can count the number of candidates you got from the closure, and not rely on the calling code "following up" via `Candidates`.

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
amnn: IIRC, the target error rate is something that was set as part of the bloom filter schema -- is there a value in the schema crate that we could refer to for this so that there's a single source of truth?

crates/sui-indexer-alt-graphql/src/api/types/transaction/bloom/scan.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Estimates how many bloom-matched candidate checkpoints to request based on the current precision.
+    ///                                                                                                                                               
+    /// Before any hits are observed, assumes the STARTING_HIT_RATE and                                                                            
+    /// doubles the request size each round to handle unknown sparsity.               
+    ///  
+    /// Once hits are observed, switches to the empirical hit rate with
+    /// an overfetch multiplier to overfetch and reduce the chance of needing another round.
+    ///
+    /// The result is always clamped to `[remaining, MAX_CHECKPOINT_FETCH]` — at least
+    /// enough to fill the page at 100% hit rate, and at most MAX_CHECKPOINT_FETCH to
```
amnn, line 155: Doesn't this min bound assume that you get only one result from each checkpoint? You may get much more than that in practice, so this lowerbound may be too generous.

amnn: The algorithm I'd aim for is one that estimates "number of candidates per checkpoint" and then works back from that to the number of checkpoints to try and fetch next, always with the aim of fetching enough checkpoints in the next iteration to satisfy all the remaining candidates required.

To calculate that metric, I would keep a running count of the number of checkpoints I had seen so far, and the number of hits I had seen so far.

The use of the bloom false positive rate and the remaining count are throwing me off here, because we're mixing up the number of candidates we need with the number of checkpoints we're fetching.

amnn:
Closing in favour of bitmap-based strategy.

PR #25720, opened 2026-03-05
Move TxContext into the execution layer 

dariorussi:
## Description 

We talked in past about moving the `TxContext` from `sui-types` in execution. That was after changing `TxContext` in Move to rely on native functions.
This PR attempts to do that at the best I can. And I truly dislike it.
There is a usage of `TxContext` in `authority_test.rs` that could likely be removed however I do not have a good home for `TxContext`.
This PR is the best I can do and I propose to leave everything as is and forget about it.
However open to suggestions if anybody cared.

This is what we end up with in terms of dependencies
```
sui-execution ──►  sui-adapter-*
     │                    │                         
     ▼                    │                     
sui-execution-types  ◄────┘
     │
     ▼
  sui-types
```
the issue is that we cannot put `TxContext` in `sui-execution` because we would create a circular dependency between adapter and execution.
So we need to create a new crate and at that point it feels just fine to leave `TxContext` where it is now

## Test plan 

Existing tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13887,6 +13887,7 @@ dependencies = [
  "parking_lot 0.12.3",
  "serde",
  "serde_json",
+ "sui-execution-types",
  "sui-macros",
  "sui-move-natives-latest",
  "sui-protocol-config",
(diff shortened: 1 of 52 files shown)


dariorussi:
gone

PR #25723, opened 2026-03-05
[move][move-compiler] Add draft signed integer implementation

cgswords:
## Description 

Signed integers through the majority of the compiler, removed around CFG/bytecode generation.

## Test plan 

New tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -978,6 +980,19 @@ fn base_type(context: &mut Context, sp!(bt_loc, bt_): H::BaseType) -> IR::Type {
...
+        | B::Apply(_, sp!(_, TN::Builtin(sp!(_, BT::I16))), _)
+        | B::Apply(_, sp!(_, TN::Builtin(sp!(_, BT::I32))), _)
+        | B::Apply(_, sp!(_, TN::Builtin(sp!(_, BT::I64))), _)
+        | B::Apply(_, sp!(_, TN::Builtin(sp!(_, BT::I128))), _)
+        | B::Apply(_, sp!(_, TN::Builtin(sp!(_, BT::I256))), _) => {
+            context
+                .env
+                .diagnostic_reporter_at_top_level()
+                .add_diag(dev_feature!(FeatureGate::SignedIntegers, bt_loc));
+            IRT::U64
```
cgswords, line 1056: I am not thrilled about manufacturing these dummy values, but we already emitted errors -- the only other thing to do is make every call here return an `Option` and that seems worse.

external-crates/move/crates/move-compiler/src/cfgir/optimize/constant_fold.rs
```diff
@@ -265,177 +398,104 @@ fn fold_binary_op(
...
+            V::I64(v) => V::$target_v(<$target_ty>::try_from(v).ok()?),
+            V::I128(v) => V::$target_v(<$target_ty>::try_from(v).ok()?),
+            V::I256(v) => V::$target_v(<$target_ty>::try_from(v).ok()?),
+            _ => return None,
+        }
     };
-    Some(evalue_(loc, v))
 }
 
 fn fold_cast(loc: Loc, sp!(_, bt_): &BuiltinTypeName, v: Value_) -> Option<UnannotatedExp_> {
```
akmysti: don't see any U8 -> i8 casts but the signed_int_cast_cross_signedness are passing. Is it implicitly handled somewhere else?

external-crates/move/crates/move-compiler/src/typing/core.rs
```diff
@@ -3574,7 +3681,25 @@ fn check_num_tvar_(subst: &Subst, ty: &Type) -> bool {
                 assert!(!ty.value.is_var());
                 check_num_tvar_(subst, ty)
             } else {
-                subst.is_num_var(&last_tvar)
+                subst.is_num_var(&last_tvar) || subst.is_signed_num_var(&last_tvar)
```
akmysti, line 3825: Here we accept both `num_var` and `signed_num_var` but below `check_signed_num_tvar_` function does not allow it.

external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -1186,6 +1201,13 @@ fn exp(context: &mut Context, code: &mut IR::BytecodeBlock, e: H::Exp) {
                 V::U64(u) => B::LdU64(u),
                 V::U128(u) => B::LdU128(u),
                 V::U256(u) => B::LdU256(u),
+                V::I8(_) | V::I16(_) | V::I32(_) | V::I64(_) | V::I128(_) | V::I256(_) => {
+                    context
+                        .env
+                        .diagnostic_reporter_at_top_level()
+                        .add_diag(dev_feature!(FeatureGate::SignedIntegers, loc));
```
akmysti: So there is `add_diag(dev_feature!(` in multiple locations. Are these diagnostics exclusive of each other?

external-crates/move/crates/move-core-types/src/i256.rs
```diff
@@ -0,0 +1,1179 @@
...
+        let err = "bad".parse::<I256>().unwrap_err();
+        assert!(!format!("{}", err).is_empty());
+    }
+
+    // Down-cast lossy
+
+    #[test]
+    fn down_cast_lossy_basic() {
+        let val = I256::from(42i32);
+        assert_eq!(val.down_cast_lossy::<i8>(), 42i8);
```
akmysti, line 988: nit: we should probably add a test for corner cases like `-128 : i8` etc.

external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
cgswords: They are not exclusive; when a span is used repeatedly, the diagnostic is collapsed. But a little noisiness here is fine either way, since it is only for internal use and I am the only consumer.

external-crates/move/crates/move-compiler/src/typing/core.rs
cgswords: Right, promotion is one-way: a signed number can be used anywhere we expect a number, but a number needs to be a signed number to be valid.

external-crates/move/crates/move-compiler/src/cfgir/optimize/constant_fold.rs
cgswords: My original inclination is that to/from signedness are not currently folded away, because that would change the underlying type's signedness. That said, it's easy enough to fix so I did.

external-crates/move/crates/move-compiler/src/parser/lexer.rs
```diff
@@ -923,17 +923,19 @@ fn get_hex_number(text: &str) -> (Tok, usize) {
...
-        (Tok::NumTypedValue, num_text_len + 4)
-    } else {
-        // No typed suffix
-        (Tok::NumValue, num_text_len)
+    for suffix in UNSIGNED_INT_SUFFIXES {
+        if rest.starts_with(suffix) {
+            return (Tok::NumTypedValue, num_text_len + suffix.len());
+        }
+    }
+    for suffix in SIGNED_INT_SUFFIXES {
```
tnowacki: Should we just chain these? or are the cases subtly different and I am missing it?
Also maybe we just do an iter+find+or_else kinda thing? (fine with the loop though)

external-crates/move/crates/move-compiler/src/shared/builtin_types.rs
```diff
@@ -0,0 +1,114 @@
...
+// Suffix helpers
+//**************************************************************************************************
+
+/// Checks whether a numeric literal token string (e.g., `"123i64"`) ends with a signed integer
+/// type suffix. Intended to be called on lexer token content, not on arbitrary strings.
+pub fn has_signed_suffix(s: &str) -> bool {
+    SIGNED_INT_SUFFIXES.iter().any(|sfx| s.ends_with(sfx))
+}
+
+pub fn has_unsigned_suffix(s: &str) -> bool {
```
tnowacki, line 44: nit, duplicate the doc comment above for this one

external-crates/move/crates/move-core-types/src/i256.rs
```diff
@@ -0,0 +1,1179 @@
+// Copyright (c) Mysten Labs, Inc.
```
tnowacki: While probably mechanical, it would be easier for me if this was a separate PR

external-crates/move/crates/move-compiler/src/expansion/translate.rs
```diff
@@ -3095,6 +3096,48 @@ fn exp(context: &mut Context, pe: Box<P::Exp>) -> Box<E::Exp> {
...
+                    }
+                }
+            }
+            sp!(vloc, PE::Value(sp!(_, v_ @ P::Value_::Num(_)))) => EE::UnaryExp(
+                op,
+                exp(context, Box::new(sp(vloc, PE::Value(sp(vloc, v_))))),
+            ),
+            // We could just allow these through because the type error will be caught later, but
+            // we are already looking at the literal here, so we might as well catch the error and
+            // provide a better error message to the user.
```
tnowacki: Would the error be much different in typing?

tnowacki: Though I guess might as well do this since we have to match on the num above anyway? 😵‍💫

external-crates/move/crates/move-compiler/src/unit_test/plan_builder.rs
```diff
@@ -382,6 +382,13 @@ fn convert_builtin_type_to_typetag(s_type: &HA::SingleType_) -> Option<TypeTag>
                         TypeTag::Vector(Box::new(tts.remove(0)))
                     }
                     NA::BuiltinTypeName_::Signer => TypeTag::Signer,
+                    // TODO (signed-ints): Fix this when we have signed ints in the bytecode
+                    NA::BuiltinTypeName_::I8
+                    | NA::BuiltinTypeName_::I16
+                    | NA::BuiltinTypeName_::I32
+                    | NA::BuiltinTypeName_::I64
+                    | NA::BuiltinTypeName_::I128
+                    | NA::BuiltinTypeName_::I256 => return None,
```
tnowacki: Maybe debug assert false here?

external-crates/move/crates/move-compiler/src/typing/expand.rs
```diff
@@ -360,39 +429,14 @@ fn inferred_numerical_value(
...
-    let max = match bt {
-        BT::U8 => u8_max,
-        BT::U16 => u16_max,
-        BT::U32 => u32_max,
-        BT::U64 => u64_max,
-        BT::U128 => u128_max,
-        BT::U256 => u256_max,
-        BT::Address | BT::Signer | BT::Vector | BT::Bool => unreachable!(),
-    };
+    let max = max_value_for(*bt);
```
tnowacki: Do we not need to replicate this logic for min?

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -628,8 +633,19 @@ pub(crate) fn move_value_from_value_(v_: Value_) -> MoveValue {
...
+        V::Vector(_, vs) => {
+            let mvs: Option<Vec<_>> = vs
+                .into_iter()
+                .map(|v| move_value_from_value(env, v))
+                .collect();
+            MV::Vector(mvs?)
+        }
+        V::I8(_) | V::I16(_) | V::I32(_) | V::I64(_) | V::I128(_) | V::I256(_) => {
+            env.diagnostic_reporter_at_top_level()
+                .add_diag(dev_feature!(FeatureGate::SignedIntegers, vloc));
```
tnowacki: not sure if necessary but maybe a debug assert here too so we catch those... though I suppose this one will be obvious either way

external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
tnowacki: It seems fine... we have the diag at least

external-crates/move/crates/move-core-types/src/i256.rs
tnowacki: Actually... its probably fine

external-crates/move/crates/move-core-types/src/i256.rs
```diff
@@ -0,0 +1,1179 @@
...
+    fn rem_assign(&mut self, rhs: Self) {
+        *self = Self(self.0 % rhs.0);
+    }
+}
+
+//**************************************************************************************************
+// Inherent methods
+//**************************************************************************************************
+
+impl I256 {
```
tnowacki, line 291: Should we have neg?

tnowacki: I see the checked/wrapping variants but am not sure how the normal one is handled

external-crates/move/crates/move-core-types/src/i256.rs
```diff
@@ -0,0 +1,1179 @@
...
+    /// Wrapping negation. Computes -self, wrapping MIN to MIN.
+    pub fn wrapping_neg(self) -> Self {
+        Self(self.0.wrapping_neg())
+    }
+
+    /// Downcast to a smaller signed type. The value is truncated (low 128 bits reinterpreted).
+    /// T must be at most i128.
+    pub fn down_cast_lossy<T: TryFrom<i128>>(self) -> T {
+        let low = self.0.as_i128();
+        match T::try_from(low) {
```
tnowacki: Try from and `as` and all that has confused me a lot lately. Can you do some test cases? Like for example doing `(from(i128::MAX) + 1).down_cast_lossy()` does this come back positive or negative?

tnowacki: Might be more like that idk

external-crates/move/crates/move-core-types/src/i256.rs
```diff
@@ -0,0 +1,1179 @@
...
+}
+
+impl TryFrom<I256> for i8 {
+    type Error = I256CastError;
+
+    fn try_from(n: I256) -> Result<Self, Self::Error> {
+        if fits_in_range(n, i8::MIN as i128, i8::MAX as i128) {
+            Ok(n.0.as_i128() as i8)
+        } else {
+            Err(I256CastError::new(n, I256CastErrorKind::OutOfRangeForI8))
```
tnowacki, line 487: The error kinds down't seem tested anywhere... if we care

external-crates/move/crates/move-core-types/src/i256.rs
```diff
@@ -0,0 +1,1179 @@
...
+
+        let min_bytes = I256::min_value().to_le_bytes();
+        assert_eq!(I256::from_le_bytes(&min_bytes), I256::min_value());
+
+        let max_bytes = I256::max_value().to_le_bytes();
+        assert_eq!(I256::from_le_bytes(&max_bytes), I256::max_value());
+    }
+
+    #[test]
+    fn u256_bits_roundtrip() {
```
tnowacki, line 738: we probably want some larger tests? Like `1 << 255` and similar strange bit layouts?

external-crates/move/crates/move-core-types/src/i256.rs
```diff
@@ -0,0 +1,1179 @@
...
+    }
+
+    /// I256 to 32 little-endian bytes (two's complement)
+    pub fn to_le_bytes(self) -> [u8; U256_NUM_BYTES] {
+        self.0.to_le_bytes()
+    }
+
+    /// I256 to 32 big-endian bytes (two's complement)
+    pub fn to_be_bytes(self) -> [u8; U256_NUM_BYTES] {
+        self.0.to_be_bytes()
```
tnowacki, line 331: A lot of these delegates seem to be untested... though its just a delegate so who cares really, just thought I would mention

external-crates/move/crates/move-core-types/src/i256.rs
tnowacki: Similarly with ! and maybe a few other small traits... again not a big deal to add these over time, just thought it would be worth mentioning

external-crates/move/crates/move-compiler/src/parser/lexer.rs
cgswords: Chained!

external-crates/move/crates/move-compiler/src/typing/expand.rs
cgswords: My initial thinking was we could do this later in the compiler (so as not to refuse `128` if it was `-128`), but being stricter here is probably the right call. I added a `negated: bool` parameter to `inferred_numerical_value` and `find_fit_signed`, an `abs_min_value_for` helper, and a new arm`exp()` that folds `UnaryExp(Neg, InferredNum)` into a signed number. New tests `signed_int_inferred_negative.move` and `signed_int_inferred_negative_overflow.move` cover both sides.

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
cgswords: Done, plus a few other places (such as below)

external-crates/move/crates/move-core-types/src/i256.rs
cgswords: Added `Neg` and `Not` impls; `neg_operator` and `not_operator` tests exercise them, including the `-MIN` two's-complement wrap. I assume these will grow naturally in the coming PRs.

external-crates/move/crates/move-core-types/src/i256.rs
cgswords: Added `down_cast_lossy_i128_wrap` tests to handle the case you asked about (equals min due to twos-complement wrap), plus `down_cast_lossy_full_width_wrap` pins MAX/MIN behavior at the I256 boundary. Should we change these behaviors?

external-crates/move/crates/move-core-types/src/i256.rs
cgswords: Added `cast_error_kinds_per_target` to verify each.

external-crates/move/crates/move-core-types/src/i256.rs
cgswords: Added `shift_to_sign_bit`, `bitwise_extreme_patterns`, `extreme_byte_roundtrip`. Any more I might have missed?

external-crates/move/crates/move-core-types/src/i256.rs
cgswords: Added several tests, plus some proptesting for this stuff.

external-crates/move/crates/move-compiler/src/shared/builtin_types.rs
```diff
@@ -0,0 +1,116 @@
...
+        pub fn $fname(s: &str, negated: bool) -> Result<($signed, NumberFormat), ParseIntError> {
+            let (magnitude, fmt) = $parse_unsigned(s)?;
+            let value = if negated {
+                if magnitude > <$signed>::MIN.unsigned_abs() {
+                    return Err(signed_overflow_parse_error());
+                }
+                // We can wrap here safely because the only way to get an overflow is if the
+                // magnitude is exactly abs(MIN), in which case the correct negated value is MIN.
+                // In that case, the wrapping negation will also produce the correct value.
+                (magnitude as $signed).wrapping_neg()
```
tnowacki: This is a bit funky I think right? Because IIRC if `magnitude == <$signed>::MIN.unsigned_abs()` then `magnitude as $signed == <$signed>::MIN` and then `wrapping_neg` is a no-op. Should we handle that case separately for clarity? Or maybe just update the comment?

external-crates/move/crates/move-compiler/src/typing/expand.rs
```diff
@@ -240,12 +240,36 @@ pub fn exp(context: &mut Context, e: &mut T::Exp) {
...
+        // Neg unifies operand and result types, so `e.ty` stands in for the inner's type.
+        E::UnaryExp(sp!(_, UnaryOp_::Neg), inner)
+            if matches!(&inner.exp.value, E::Value(sp!(_, Value_::InferredNum(_))))
+                && matches!(
+                    e.ty.value.builtin_name(),
+                    Some(sp!(_, bt)) if bt.is_signed_numeric()
+                ) =>
+        {
+            let inner_loc = inner.exp.loc;
+            let E::Value(sp!(_, Value_::InferredNum(v))) = &inner.exp.value else {
```
tnowacki, line 261: Some days I feel like I understand Rust patterns... some days I don't. But does this work to remove `let v = * v` below?
```suggestion
            let E::Value(sp!(_, Value_::InferredNum(&v))) = &inner.exp.value else {
```

external-crates/move/crates/move-core-types/src/i256.rs
```diff
@@ -0,0 +1,1319 @@
+// Copyright (c) Mysten Labs, Inc.
```
tnowacki: The copyright here needs to be core contributors, not mysten

tnowacki (changes requested):
Overall looks good, and should be good to go (with some small nits and questions).

Though I'll request changes just in case the copyright checker on CI isn't functioning properly for external-crates:
We historically haven't been putting copyrights on `*.move` test files (at one point we didn't have comments and that just kind of stuck).
But for other files, be sure that the copyright is for the Move core contributors and not Mysten.

external-crates/move/crates/move-core-types/src/i256.rs
cgswords: Whoops

external-crates/move/crates/move-compiler/src/typing/expand.rs
cgswords: Sadly, no, but we can use `*v` directly inline.

external-crates/move/crates/move-compiler/src/shared/builtin_types.rs
cgswords: I pulled it out separately for clarify, since it did feel a little magical.

tnowacki (approved):

external-crates/move/crates/move-core-types/src/i256.rs
cgswords: Closing the loop on my own question: the latest push makes `down_cast_lossy` truncate uniformly (pure bit-truncation) for every target width, matching `U256::down_cast_lossy`'s contract, with the behavior pinned per width in tests. `(from(i128::MAX) + 1).down_cast_lossy()` comes back as `i128::MIN`.

PR #25728, opened 2026-03-06
[sui-exploration] Unify address payment and gas smashing logic

tnowacki:
## Description 

- Gas payment handling was modified so that smashed gas also supports a new AddressBalance payment in a unified model (see `PaymentKind`, `PaymentMethod`, and `SmashMetadata`).
- All transactions can now use `GasCoin` (except dev-inspect if one isn't specified). If the payment method (the first item in the smashed list) is an address balance. A new coin object is created. 
  - If that coin remains with the original address, it is converted back into an address balance. In other words, the change in the coin's value is deposited to/withdrawn from the address balance.
  - If that coin was transferred to a new address, it remains transferred. This also "overrides" where to take the final gas payment from (it switches from the address balance to the coin)
- Changed AccumulatorEvent::from_balance_change to be more predictable in terms of +/- sign (+ is merge, - is split). 

  
## Test plan 

- Small tests, but waiting on @mystenmark's larger PR

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-open-rpc/tests/snapshots/generate_spec__openrpc.snap.json
+++ b/crates/sui-open-rpc/tests/snapshots/generate_spec__openrpc.snap.json
@@ -1400,7 +1400,6 @@
                 "narwhal_new_leader_election_schedule": false,
                 "narwhal_versioned_metadata": false,
                 "native_charging_v2": false,
-                "new_vm_enabled": false,
                 "no_extraneous_module_bytes": false,
                 "normalize_depth_formula": false,
                 "normalize_ptb_arguments": false,


sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -246,35 +248,69 @@ impl<'env, 'pc, 'vm, 'state, 'linkage, 'gas> Context<'env, 'pc, 'vm, 'state, 'li
...
-                    &mut input_object_map,
-                    gas_coin,
-                    ObjectMutability::Mutable,
-                    ty,
-                )?;
+                let (gas_metadata, gas_value) = match payment_location {
+                    PaymentLocation::AddressBalance(sui_address) => {
+                        // TODO we need a way to:
+                        // - get the "in memory" value for this address balance
+                        // - deduct that amount from the "in memory" balance
```
mystenmark: we will need to pass it as part of PaymentLocation. The value will be the gas budget (if its a pure AB payment) or the sum of the coin reservation amounts (if compat layer is being used).

To deduct we just emit a withdraw event

crates/sui-types/src/accumulator_event.rs
```diff
@@ -63,10 +63,10 @@ impl AccumulatorEvent {
 
         let accumulator_address = AccumulatorAddress::new(address, balance_type);
 
-        let (operation, amount) = if net_change > 0 {
-            (AccumulatorOperation::Split, net_change as u64)
+        let (operation, amount) = if net_change < 0 {
+            (AccumulatorOperation::Split, net_change.unsigned_abs())
         } else {
-            (AccumulatorOperation::Merge, (-net_change) as u64)
```
tnowacki, line 69: This confused me as to why negative meant deposit/merge and positive meant withdrawal/split

mystenmark: yeah i think i also fixed this in my stack

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -407,6 +412,15 @@ impl<'backing> TemporaryStore<'backing> {
         self.execution_results.written_objects.insert(id, object);
     }
 
+    pub fn mutate_new_or_input_object(&mut self, object: Object) {
```
tnowacki, line 415: I don't love this but... it seemed better than changing the behavior of the existing function

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -87,27 +123,47 @@ pub mod checked {
...
-                }
+        pub(crate) fn used_coins(&self) -> impl Iterator<Item = &'_ ObjectRef> {
+            match &self.payment {
+                PaymentMetadata::Unmetered => Either::Left(std::iter::empty()),
+                PaymentMetadata::Smash(metadata) => Either::Right(metadata.used_coins()),
+            }
+        }
+
+        // Override the gas payment location for smashing
+        pub fn override_gas_charge_location(
```
tnowacki, line 170: This seems a bit brittle, but allows for the ephemeral gas coin to be transferred to a new address

mystenmark:
Looks good except:
- The broken simtest indicates that the protocol gating isn't quite right
- Did you see the indexer transactional test? concerning

crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas_transfer.move
```diff
@@ -3,16 +3,24 @@
...
 
 //# create-checkpoint
 
-// Transfer gas coin to 0x0 via TransferObjects while paying with address balance
 //# programmable --sender A --inputs @0x0 object(0,0) --address-balance-gas
+// Transfer gas coin to 0x0 via TransferObjects while paying with address balance
+//> TransferObjects([Gas], Input(0))
+
+//# programmable --sender B --inputs @B object(0,1) --address-balance-gas
+// Transfer gas coin to the sender to show no object was created
```
tzakian: Larger semantic question: Should a self transfer result in the same result as a transfer to any other address? It feels kinda odd to me that `TransferObjects` results in different actions depending on the receiving address.

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -86,6 +87,31 @@ mod checked {
...
 
+    fn payment_kind(gas_data: &GasData, transaction_kind: &TransactionKind) -> PaymentKind {
+        if gas_data.is_unmetered() || transaction_kind.is_system_tx() {
+            PaymentKind::unmetered()
+        } else if gas_data.payment.is_empty() {
+            PaymentKind::smash(vec![PaymentMethod::AddressBalance(
+                gas_data.owner,
+                gas_data.budget,
+            )])
+            .unwrap()
```
tzakian: nit: maybe consider `expect`s here with a reason for why this is safe in them?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -39,46 +40,81 @@ pub mod checked {
...
+        total_smashed: u64,
+        methods: Vec<PaymentMethod>,
+    }
+
+    #[derive(Debug)]
+    pub struct PaymentKind(PaymentKind_);
+
+    #[derive(Debug)]
+    enum PaymentKind_ {
+        Unmetered,
```
tzakian, line 84: Private variants on Rust enums would be nice..

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -39,46 +40,81 @@ pub mod checked {
...
+    }
+
+    #[derive(Debug)]
+    pub enum PaymentMethod {
+        Coin(ObjectRef),
+        AddressBalance(SuiAddress, /* withdrawal reservation */ u64),
+    }
+
+    #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
+    pub enum PaymentLocation {
```
tzakian, line 101: Why the separation between `PaymentLocation` and `PaymentMethod`? One just seems to be a less precise version of the other.

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -39,46 +40,81 @@ pub mod checked {
...
             protocol_config: &ProtocolConfig,
         ) -> Self {
             let gas_model_version = protocol_config.gas_model_version();
+            let payment = match payment_kind.0 {
+                PaymentKind_::Unmetered => PaymentMetadata::Unmetered,
+                PaymentKind_::Smash(payment_methods) => {
+                    let mut metadata = SmashMetadata {
+                        // dummy value set below in smash_gas
+                        total_smashed: 0,
+                        gas_charge_location: payment_methods[0].location(),
```
tzakian: Can you add a comment on the `Smash` variant (or on `PaymentKind_`) about the invariant that the `payment_methods` must always be non-empty?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -39,46 +40,81 @@ pub mod checked {
...
+
+    #[derive(Debug)]
+    pub struct PaymentKind(PaymentKind_);
+
+    #[derive(Debug)]
+    enum PaymentKind_ {
+        Unmetered,
+        /// A non-empty vector of gas coins or address balance withdrawals. The payments are
+        /// "smashed" into the first location in the vector.
+        Smash(Vec<PaymentMethod>),
```
tzakian: Thoughts on something like

```
Smash {
   primary: PaymentMethod,
   smashed_away: Vec<PaymentMethod>,
}
```

Or something similar? All of the indexing/slicing logic around this is making me a smidge worried.

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -478,4 +467,189 @@ pub mod checked {
...
+            self.gas_charge_location = self.smash_target().location();
+
+            // sum the value of all gas coins
+            let total_smashed = self
+                .methods
+                .iter()
+                .map(|payment| match payment {
+                    PaymentMethod::AddressBalance(_, reservation) => Ok(*reservation),
+                    PaymentMethod::Coin(obj_ref) => {
+                        let obj = temporary_store.objects().get(&obj_ref.0).unwrap();
```
tzakian: Why not invariant violation here?

IK IK it'll just panic at the unwrap_or_else below but feels maybe a smidge nicer?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -478,4 +467,189 @@ pub mod checked {
...
+                // are valid
+                .unwrap_or_else(|_| {
+                    panic!(
+                        "Unable to process gas payments for transaction {}",
+                        tx_digest
+                    )
+                })
+                .iter()
+                .sum();
+            self.total_smashed = total_smashed;
```
tzakian, line 555: nit: debug_assert `total_smashed = 0` before this?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -478,4 +467,189 @@ pub mod checked {
...
+
+            let smash_target = self.smash_target();
+            let smash_location = smash_target.location();
+            // delete all gas objects except the primary_gas_object
+            for payment_method in &self.methods[1..] {
+                let location = payment_method.location();
+                assert_ne!(location, smash_location, "Payment methods must be unique");
```
tzakian, line 561: Do we care about checking uniqueness between non-primary smash methods here as well? (i.e., that positions 1 and 2 can't be the same). I believe the answer is no, but since you wrote this check I gotta ask ;)

tzakian: also to double check here, this means that having payment methods like `vec![AddressBalance(0x1, 10), AddressBalance(0x1, 11)]` is invalid/already checked against? (/but `[Coin(0x...), AddressBalance(0x1, 10), AddressBalance(0x1, 11)]` I believe is valid?)

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -478,4 +467,189 @@ pub mod checked {
...
+                            sui_types::gas_coin::GAS::type_tag(),
+                        );
+                        let event = AccumulatorEvent::from_balance_change(
+                            *sui_address,
+                            balance_type,
+                            i64::try_from(deposit).unwrap(),
+                        )
+                        .expect("Failed to create accumulator event for gas smashing");
+                        temporary_store.add_accumulator_event(event);
+                    }
```
tzakian, line 597: what if `deposit == 0`? is that possible e.g., with zero coins or address balance methods with zero reservations in them?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -478,4 +467,189 @@ pub mod checked {
...
+                assert_ne!(location, smash_location, "Payment methods must be unique");
+                match payment_method {
+                    PaymentMethod::AddressBalance(sui_address, reservation) => {
+                        let balance_type = sui_types::balance::Balance::type_tag(
+                            sui_types::gas_coin::GAS::type_tag(),
+                        );
+                        let event = AccumulatorEvent::from_balance_change(
+                            *sui_address,
+                            balance_type,
+                            i64::try_from(*reservation).unwrap(),
```
alex-mysten, line 570: we want to negate reservation, correct?

crates/sui-types/src/transaction.rs
```diff
@@ -2958,6 +2958,7 @@ impl TransactionDataAPI for TransactionDataV1 {
             && self.is_gas_paid_from_address_balance()
         {
             if config.address_balance_gas_reject_gas_coin_arg()
+                && !config.gasless_transaction_drop_safety()
```
dariorussi: I am having difficulty understanding the reason for the added check

sui-execution/latest/sui-adapter/src/gas_charger.rs
tnowacki: In object terms, think about it as sometimes needing the ObjectID instead of the full ObjectRef.

sui-execution/latest/sui-adapter/src/gas_charger.rs
tnowacki: Both @mystenmark and I found that confusing and changed the logic inside of `AccumulatorEvent::from_balance_change`

crates/sui-types/src/transaction.rs
tnowacki: Ah shit that actually needs to be a different check. But basically we banned this and are now unbanning it

crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas_transfer.move
tnowacki: The difficulty is in that the transfer-to-self is implicit at the end of PTBs. So currently I can't differentiate easily between the two scenarios. But maybe I should

sui-execution/latest/sui-adapter/src/gas_charger.rs
tnowacki: We cannot invariant violation in this code. If it it doesn't succeed, there is no way to handle the transaction.

sui-execution/latest/sui-adapter/src/gas_charger.rs
tnowacki: It might not be 0 if the we have run this before, and are now re-running it after a reset (if the transaction failed)

sui-execution/latest/sui-adapter/src/gas_charger.rs
tnowacki: We make things unique in PaymentKind::Smash. Really this assert is from the old code and maybe should just be removed?

sui-execution/latest/sui-adapter/src/gas_charger.rs
tnowacki: That happens often. If there is just a single payment method of an address balance. In that case, we don't need to pre-withdraw anything, and we simply charge the amount used at the end of the transaction.

In a way, this is happening too in the Coin case. If the total_smashed-original_balance is 0, the coin doesn't actually change in value (but we overwrite it anyway)

sui-execution/latest/sui-adapter/src/gas_charger.rs
tzakian: Agreed that the larger function can't error, but it's just in the code below this we're creating errors which just get unwrapped at the end of this `map`. So it seems nice that this should be a similar error if it fails? (Or perhaps those errors below this should become expects/unwraps)?

sui-execution/latest/sui-adapter/src/gas_charger.rs
tzakian: ah I see. In that case the invariant would be something like `self.total_smashed == 0 || self.total_smashed == total_smashed)`? 

I don't care particularly if we add it just more making sure I understand things.

sui-execution/latest/sui-adapter/src/gas_charger.rs
tzakian: I'd be in favor of keeping it probably (or moving it to a `debug_assert` possibly).

crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas_transfer.move
tzakian: Yea, I could see this as going either way, respecting existing semantics around `TransferObjects` seems to be better, but I also can see how that would be "fun" to implement, and not sure if it's worth that level of effort or not 🤷🏻‍♂️

sui-execution/latest/sui-adapter/src/gas_charger.rs
tnowacki: I kept it but also moved everything into an indexset

sui-execution/latest/sui-adapter/src/gas_charger.rs
alex-mysten: I totally agree that from_balance_change was confusing - sorry about that. 

Shouldn't we withdraw rather than deposit from the accumulator for each one of these reservations, as they will be smashed into the smash_target?

It's possible I am misunderstanding this

crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas_transfer.move
tnowacki: I kind of hate the code, but done. We can revisit the infra when we inevitably support "send_funds"

crates/sui-protocol-config/src/lib.rs
```diff
@@ -5022,6 +5022,12 @@ impl ProtocolConfig {
         self.feature_flags.enable_address_balance_gas_payments = true;
         self.feature_flags.address_balance_gas_check_rgp_at_signing = true;
         self.feature_flags.address_balance_gas_reject_gas_coin_arg = true;
+        self.feature_flags.new_vm_enabled = true;
+        if let Some(cur) = self.execution_version
+            && cur < 4
+        {
+            self.execution_version = Some(4);
+        };
```
tnowacki: This was the main test fix

crates/sui-types/src/transaction.rs
```diff
@@ -2958,6 +2958,7 @@ impl TransactionDataAPI for TransactionDataV1 {
             && self.is_gas_paid_from_address_balance()
         {
             if config.address_balance_gas_reject_gas_coin_arg()
+                && !config.new_vm_enabled()
```
tnowacki: @tzakian this is your fault for making me not need a protocol version. Should I instead rely on the execution version?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -399,11 +442,19 @@ impl<'env, 'pc, 'vm, 'state, 'linkage, 'gas, 'extension>
         let object_input_metadata = std::mem::take(&mut self.locations.input_object_metadata);
         let mut object_inputs =
             std::mem::replace(&mut self.locations.object_inputs, Locals::new_invalid(0)?);
+        let mut created_input_object_ids = BTreeSet::new();
         let mut loaded_runtime_objects = BTreeMap::new();
         let mut by_value_shared_objects = BTreeSet::new();
         let mut consensus_owner_objects = BTreeMap::new();
+        let mut gas_payment_location = None;
+        let mut gas_coin_moved = None;
```
tnowacki, line 450: @tzakian I am now tracking this and destroying the ephemeral coin only when it wasn't moved.

crates/sui-e2e-tests/tests/address_balance_tests.rs
```diff
@@ -3336,48 +3336,49 @@ async fn test_simulate_object_funds_insufficient() {
...
-        .await
-        .unwrap_err();
-    assert!(
-        err.to_string()
-            .contains("Argument::GasCoin is not supported with address balance gas payments"),
-    );
-
-    test_env.verify_accumulator_exists(sender, 10_000_000);
-}
+// TODO(address-balances): Re-enable this once input checks land
```
tnowacki, line 3339: @mystenmark I will need your fixes to re-enable this

mystenmark: can you use `#[ignore = "reason"]` instead of commenting it out?

mystenmark: actually i doesn't matter, I will have a PR up to re-enable it pretty much immediately

sui-execution/latest/sui-adapter/src/gas_charger.rs
mystenmark: I did have to fix something like this in my branch that is stacked on top of this. We don't need to fix it in this PR imo, my branch has a fix already which is covered by tests.

crates/sui-types/src/transaction.rs
tzakian: They're the same -- I needed it for one place in sui-types and it felt odd to do `config.execution_version() > 3` type stuff so I added that in. So relying on this is fine.

crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas_transfer.move
tzakian: nitty nit: comment on the test is now wrong ;) (I wouldn't bother fixing unless you need to update this PR anyways for other reasons though/feel free to roll into other PRs).

crates/sui-e2e-tests/tests/address_balance_tests.rs
tnowacki: Ah fancy. Was not aware that this worked

crates/sui-types/src/transaction.rs
mystenmark: shouldn't this be `||`?

tnowacki: TLDR: No? I think? We don't want this to run when the new VM is enabled. Therefore, `&& !config.new_vm_enabled()`

---

But your question had me spiraling so to double check

If we have the old check as `old_check`. We would want 
```
assert!(old_check)
```
to become
```
assert!(!config.new_vm_enabled() ==> old_check)
```
which is 
```
!!config.new_vm_enabled() || old_check
```
which is 
```
config.new_vm_enabled() || old_check
```

But here 
```
old_check := (address_balance_gas_reject_gas_coin_arg && is_ptb) ==> !exists cmd. cmd.is_gas_coin_used
```
so 
```
config.new_vm_enabled()  || ((address_balance_gas_reject_gas_coin_arg && is_ptb) ==> !exists cmd. cmd.is_gas_coin_used)
```
entails
```
config.new_vm_enabled()  || (!(address_balance_gas_reject_gas_coin_arg && is_ptb) || !exists cmd. cmd.is_gas_coin_used)
```
entails
```
(config.new_vm_enabled() || !(address_balance_gas_reject_gas_coin_arg && is_ptb)) || !exists cmd. cmd.is_gas_coin_used
```
entails
```
!!(config.new_vm_enabled() || !(address_balance_gas_reject_gas_coin_arg && is_ptb)) || !exists cmd. cmd.is_gas_coin_used
```
entails
```
!(!config.new_vm_enabled() && !!(address_balance_gas_reject_gas_coin_arg && is_ptb)) || !exists cmd. cmd.is_gas_coin_used
```
entails
```
!(!config.new_vm_enabled() && address_balance_gas_reject_gas_coin_arg && is_ptb) || !exists cmd. cmd.is_gas_coin_used
```
entails
```
(!config.new_vm_enabled() && address_balance_gas_reject_gas_coin_arg && is_ptb) ==> !exists cmd. cmd.is_gas_coin_used
```
entails
```
(address_balance_gas_reject_gas_coin_arg && !config.new_vm_enabled() && is_ptb) ==> !exists cmd. cmd.is_gas_coin_used
```
(citations of de morgan's law, double negation elimination/introduction, associativity, and commutativity are left as an exercise for the user)

crates/sui-protocol-config/src/lib.rs
```diff
@@ -1025,10 +1025,6 @@ struct FeatureFlags {
     // If true, mark the gas coin as uninitialized in drop safety when there is no gas coin.
     #[serde(skip_serializing_if = "is_false")]
     gasless_transaction_drop_safety: bool,
-
-    // Set to true if new VM (bella ciao) is enabled in this protocol version.
-    #[serde(skip_serializing_if = "is_false")]
-    new_vm_enabled: bool,
```
tnowacki, line 1031: Discussed this with @tzakian offline

PR #25737, opened 2026-03-06
sui-indexer-alt: change checkpoint_hi_inclusive to checkpoint_hi

evan-wall-mysten:
## Description 

This PR changes the watermark `checkpoint_hi_inclusive` field to `checkpoint_hi` (exclusive) in order to support the initial `[reader_lo, checkpoint_hi)` empty state where no checkpoints have been indexed.

This replaces https://github.com/MystenLabs/sui/pull/25616.

## Test plan 

Unit tests passing.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-consistent-store/src/db/map.rs
+++ b/crates/sui-indexer-alt-consistent-store/src/db/map.rs
@@ -206,7 +206,7 @@ mod tests {
 
         let db = Arc::new(Db::open(d.path().join("db"), opts, 4, cfs).unwrap());
         let map: DbMap<u64, u64> = DbMap::new(db.clone(), "test");
-        db.take_snapshot(wm(0));
+        db.take_snapshot(wm(1));
 
         // Access succeeds at first, but will start to fail after the column family is dropped.
         assert!(map.get(0, 42).unwrap().is_none());
@@ -234,14 +234,14 @@ mod tests {
         let mut batch = rocksdb::WriteBatch::default();
         map.insert(42, 43, &mut batch).unwrap();
         map.insert(44, 45, &mut batch).unwrap();
-        db.write("batch", wm(0), batch).unwrap();
-        db.take_snapshot(wm(0));
+        db.write("batch", wm(1), batch).unwrap();
+        db.take_snapshot(wm(1));
 
         let mut batch = rocksdb::WriteBatch::default();
         map.remove(42, &mut batch).unwrap();
         map.insert(43, 42, &mut batch).unwrap();
-        db.write("batch", wm(1), batch).unwrap();
-        db.take_snapshot(wm(1));
+        db.write("batch", wm(2), batch).unwrap();
+        db.take_snapshot(wm(2));
 
         // Point look-ups
         assert_eq!(map.get(0, 42).unwrap(), Some(43));
@@ -333,21 +333,21 @@ mod tests {
         // Successfully perform a merge to counts.
         let mut batch = rocksdb::WriteBatch::default();
         counts.merge(42, 1, &mut batch).unwrap();
-        db.write("counts", wm(0), batch).unwrap();
-        db.take_snapshot(wm(0));
+        db.write("counts", wm(1), batch).unwrap();
+        db.take_snapshot(wm(1));
 
         // Merges allow for the accumulation of values.
         let mut batch = rocksdb::WriteBatch::default();
         counts.merge(42, 2, &mut batch).unwrap();
-        db.write("counts", wm(1), batch).unwrap();
-        db.take_snapshot(wm(1));
+        db.write("counts", wm(2), batch).unwrap();
+        db.take_snapshot(wm(2));
 
         // A single batch can include multiple merges for the same key.
         let mut batch = rocksdb::WriteBatch::default();
         counts.merge(42, 3, &mut batch).unwrap();
         counts.merge(42, 4, &mut batch).unwrap();
-        db.write("counts", wm(2), batch).unwrap();
-        db.take_snapshot(wm(2));
+        db.write("counts", wm(3), batch).unwrap();
+        db.take_snapshot(wm(3));
 
         assert_eq!(counts.get(0, 42).unwrap(), Some(1));
         assert_eq!(counts.get(1, 42).unwrap(), Some(3));
@@ -399,8 +399,8 @@ mod tests {
         map.insert(0x0000_0003, 30, &mut batch).unwrap();
         map.insert(0xffff_0004, 40, &mut batch).unwrap();
         map.insert(0x0000_0005, 50, &mut batch).unwrap();
-        db.write("batch", wm(0), batch).unwrap();
-        db.take_snapshot(wm(0));
+        db.write("batch", wm(1), batch).unwrap();
+        db.take_snapshot(wm(1));
 
         assert_eq!(
             map.prefix(0, &0x0000u16)
--- a/crates/sui-indexer-alt-consistent-store/src/db/mod.rs
+++ b/crates/sui-indexer-alt-consistent-store/src/db/mod.rs
@@ -80,7 +80,7 @@ pub(crate) struct Db(RwLock<Inner>);
 #[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) struct Watermark {
     pub epoch_hi_inclusive: u64,
-    pub checkpoint_hi_inclusive: u64,
+    pub checkpoint_hi: u64,
     pub tx_hi: u64,
     pub timestamp_ms_hi_inclusive: u64,
 }
@@ -319,7 +319,10 @@ impl Db {
     pub(crate) fn take_snapshot(&self, watermark: Watermark) {
         self.0.write().expect("poisoned").with_mut(|f| {
             f.snapshots.insert(
-                watermark.checkpoint_hi_inclusive,
+                watermark
+                    .checkpoint_hi
+                    .checked_sub(1)
+                    .unwrap_or_else(|| panic!("Watermark checkpoint_hi underflow {watermark:?}")),
                 (Arc::new(f.db.snapshot()), watermark),
             );
 
@@ -753,8 +756,7 @@ unsafe impl marker::Send for Db {}
 /// Watermarks are identified by their checkpoint, so comparison can be limited to them.
 impl Ord for Watermark {
     fn cmp(&self, other: &Self) -> cmp::Ordering {
-        self.checkpoint_hi_inclusive
-            .cmp(&other.checkpoint_hi_inclusive)
+        self.checkpoint_hi.cmp(&other.checkpoint_hi)
     }
 }
 
@@ -768,7 +770,7 @@ impl From<Watermark> for CommitterWatermark {
     fn from(w: Watermark) -> Self {
         Self {
             epoch_hi_inclusive: w.epoch_hi_inclusive,
-            checkpoint_hi_inclusive: w.checkpoint_hi_inclusive,
+            checkpoint_hi: w.checkpoint_hi,
             tx_hi: w.tx_hi,
             timestamp_ms_hi_inclusive: w.timestamp_ms_hi_inclusive,
         }
@@ -779,7 +781,7 @@ impl From<CommitterWatermark> for Watermark {
     fn from(w: CommitterWatermark) -> Self {
         Self {
             epoch_hi_inclusive: w.epoch_hi_inclusive,
-            checkpoint_hi_inclusive: w.checkpoint_hi_inclusive,
+            checkpoint_hi: w.checkpoint_hi,
             tx_hi: w.tx_hi,
             timestamp_ms_hi_inclusive: w.timestamp_ms_hi_inclusive,
         }
@@ -841,10 +843,10 @@ pub(crate) mod tests {
         opts
     }
 
-    pub(crate) fn wm(cp: u64) -> Watermark {
+    pub(crate) fn wm(checkpoint_hi: u64) -> Watermark {
         Watermark {
             epoch_hi_inclusive: 0,
-            checkpoint_hi_inclusive: cp,
+            checkpoint_hi,
             tx_hi: 0,
             timestamp_ms_hi_inclusive: 0,
         }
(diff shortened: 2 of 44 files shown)


crates/sui-indexer-alt-consistent-store/src/db/map.rs
```diff
@@ -206,7 +206,7 @@ mod tests {
 
         let db = Arc::new(Db::open(d.path().join("db"), opts, 4, cfs).unwrap());
         let map: DbMap<u64, u64> = DbMap::new(db.clone(), "test");
-        db.take_snapshot(wm(0));
+        db.take_snapshot(wm(1));
```
wlmyng, line 209: Since `take_snapshot` subtracts 1 from the incoming checkpoint_hi, reads still work i.e someone can query `db.get(0, ...)`

crates/sui-indexer-alt-consistent-store/src/db/mod.rs
```diff
@@ -319,7 +319,10 @@ impl Db {
     pub(crate) fn take_snapshot(&self, watermark: Watermark) {
```
wlmyng, line 319: we should update the doc comments since checkpoint_hi is exclusive

crates/sui-indexer-alt-consistent-store/src/restore/formal_snapshot.rs
```diff
@@ -178,7 +178,7 @@ impl FormalSnapshot {
 
         let watermark = Watermark {
             epoch_hi_inclusive: epoch,
-            checkpoint_hi_inclusive: summary.sequence_number,
+            checkpoint_hi: summary.sequence_number + 1,
```
wlmyng, line 181: if I recall, this checkpoint summary is from the end-of-epoch checkpoint, so if we want to bump `checkpoint_hi` to be exclusive, shouldn't the other fields also be bumped?

crates/sui-indexer-alt-consistent-store/src/rpc/consistent_service/available_range.rs
```diff
@@ -12,9 +12,19 @@ pub(super) fn available_range(
     grpc::AvailableRangeRequest {}: grpc::AvailableRangeRequest,
 ) -> Result<grpc::AvailableRangeResponse, RpcError> {
     let range = state.store.db().snapshot_range(checkpoint);
+    let min_checkpoint = range.as_ref().map(|r| {
+        r.start().checkpoint_hi.checked_sub(1).unwrap_or_else(|| {
+            panic!("Snapshot range start checkpoint_hi underflow checkpoint={checkpoint}")
+        })
```
wlmyng, line 18: here and in `rpc/state.rs`, has the same logic, why don't we just have `snapshot_range` return the inclusive checkpoint range?

crates/sui-indexer-alt-consistent-store/src/rpc/state.rs
```diff
@@ -71,8 +75,8 @@ impl State {
             .parse()
             .map_err(|_| Error::BadCheckpoint(checkpoint.clone()))?;
 
-        if checkpoint < snapshot_range.start().checkpoint_hi_inclusive
-            || checkpoint > snapshot_range.end().checkpoint_hi_inclusive
+        if checkpoint + 1 < snapshot_range.start().checkpoint_hi
+            || checkpoint >= snapshot_range.end().checkpoint_hi
```
wlmyng, line 79: if the snapshot_range can convert to `checkpoint_hi_inclusive` we can avoid having this logic strewn everywhere in the consistent-store

crates/sui-indexer-alt-consistent-store/src/store/mod.rs
```diff
@@ -660,7 +663,7 @@ mod tests {
 
         // With the fix, no snapshot is taken until after the first checkpoint is written.
         // The first snapshot will be at checkpoint 100, not 99.
-        wait_until(|| async { has_range(&store, Some(100), Some(100)) })
+        wait_until(|| async { has_range(&store, Some(101), Some(101)) })
```
wlmyng, line 666: if snapshot_range accounted for checkpoint_hi -> inclusive then we also don't need to do this kind of bumping

crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
```diff
@@ -279,10 +279,8 @@ async fn write_watermark<H: Handler>(
 
             logger.log::<H>(watermark, elapsed);
 
-            checkpoint_lag_reporter.report_lag(
-                watermark.checkpoint_hi_inclusive,
-                watermark.timestamp_ms_hi_inclusive,
-            );
+            checkpoint_lag_reporter
+                .report_lag(watermark.checkpoint_hi, watermark.timestamp_ms_hi_inclusive);
```
wlmyng, line 283: could it throw our alerts off, that we're reporting `checkpoint_hi - 1`'s `timestamp_ms_hi_inclusive` for some `checkpoint_hi`?

crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
```diff
@@ -399,7 +397,7 @@ mod tests {
     fn create_watermark_part_for_checkpoint(checkpoint: u64) -> WatermarkPart {
         WatermarkPart {
             watermark: CommitterWatermark {
-                checkpoint_hi_inclusive: checkpoint,
+                checkpoint_hi: checkpoint + 1,
```
wlmyng, line 400: it'd be great if we could use some kind of constructor, so we don't need to keep remembering to do a `+1`. There are places where we do saturating sub / add and other places where we don't, too

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -265,7 +265,10 @@ impl Watermarks {
         let pipeline = Pipeline {
             hi: Watermark {
                 epoch: row.epoch_hi_inclusive,
-                checkpoint: row.checkpoint_hi_inclusive,
+                checkpoint: row
+                    .checkpoint_hi
+                    .checked_sub(1)
+                    .unwrap_or_else(|| panic!("WatermarkRow checkpoint_hi underflow {row:?}")),
```
wlmyng, line 271: I think either downstream services also reflect the exclusive nature of checkpoint_hi, or the framework should expose a way for the inclusive checkpoint to be retrieved, so that people don't have to add this check everywhere if they want the inclusive hi

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -442,13 +445,15 @@ async fn watermarks_from_pg(
         .await
         .context("Failed to connect to database")?;
 
+    // Filtering on `reader_lo < checkpoint_hi` prevents returning an empty
+    // [reader_lo, checkpoint_hi) range.
```
wlmyng, line 449: but can this happen where we have `reader_lo = checkpoint_hi`?

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -499,7 +505,7 @@ async fn watermark_from_consistent(
         }) => Ok(Some(WatermarkRow {
             pipeline: "consistent".to_owned(),
             epoch_hi_inclusive: max_epoch as i64,
-            checkpoint_hi_inclusive: max_checkpoint as i64,
+            checkpoint_hi: max_checkpoint as i64 + 1,
```
wlmyng, line 508: i think the user of the reader should not need to know whether the checkpoint_hi is stored inclusively or exclusively on the store they're trying to read from, that's something for the reader client itself to manage

crates/sui-indexer-alt-object-store/src/lib.rs
```diff
@@ -216,7 +216,7 @@ mod tests {
         // Set initial watermark
         let initial_watermark = framework_traits::CommitterWatermark {
             epoch_hi_inclusive: 1,
-            checkpoint_hi_inclusive: 100,
+            checkpoint_hi: 100,
```
wlmyng, line 219: shouldnt this be 101?

crates/sui-indexer-alt/src/handlers/obj_info.rs
```diff
@@ -158,9 +158,9 @@ impl Handler for ObjInfo {
 
     /// To prune `obj_info`, entries between `[from, to_exclusive)` are read from the reference
     /// table, and the previous versions of the objects are deleted from the main table. Finally,
-    /// the reference entries themselves are deleted. The framework guarantees that `to_exclusive <=
-    /// checkpoint_hi_inclusive - retention`, so we only prune data that has been committed but is
-    /// beyond the retention window.
+    /// the reference entries themselves are deleted. The framework guarantees that
+    /// `to_exclusive < checkpoint_hi - retention`, so we only prune data that has been committed
```
wlmyng, line 162: minor but i think `to_exclusive <= checkpoint_hi - retention` is correct? Thinking about an earlier test where checkpoint_hi = 11 and retention = 5, to_exclusive could at most be 6, since `to_exclusive` reflects the checkpoint that the pruner can prune up to, exclusive of that checkpoint

crates/sui-pg-db/migrations/2026-03-06-132800_watermarks_checkpoint_hi/down.sql
```diff
@@ -0,0 +1,12 @@
+-- replace watermarks.checkpoint_hi with watermarks.checkpoint_hi_inclusive
+
+ALTER TABLE watermarks ADD COLUMN checkpoint_hi_inclusive BIGINT;
+
+UPDATE watermarks SET checkpoint_hi_inclusive = checkpoint_hi - 1;
+
+-- delete these records because they would not have been written by the old version
+DELETE FROM watermarks WHERE checkpoint_hi_inclusive = 0;
```
wlmyng, line 8: I think we track all watermarks including at checkpoint 0?

wlmyng:
@evan-wall-mysten did you close this accidentally?

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -117,12 +117,12 @@ pub trait TransactionalStore: Store {
...
 
 /// Represents the highest checkpoint for some pipeline that has been processed by the indexer
-/// framework. When read from the `Store`, this represents the inclusive upper bound checkpoint of
+/// framework. When read from the `Store`, this represents the exclusive upper bound checkpoint of
 /// data that has been written to the Store for a pipeline.
 #[derive(Default, Debug, Clone, Copy, PartialEq)]
 pub struct CommitterWatermark {
     pub epoch_hi_inclusive: u64,
-    pub checkpoint_hi_inclusive: u64,
+    pub checkpoint_hi: u64,
```
wlmyng, line 125: additionally or alternatively, we could think of `checkpoint_hi` as the checkpoint for the indexer to resume from for a particular pipeline

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -14,7 +14,7 @@ use scoped_futures::ScopedBoxFuture;
 #[async_trait]
 pub trait Connection: Send {
     /// If no existing watermark record exists, initializes it with `default_next_checkpoint`.
-    /// Returns the committer watermark `checkpoint_hi_inclusive`.
+    /// Returns the committer watermark `checkpoint_hi`.
```
wlmyng, line 17:
```suggestion
    /// Returns the committer watermark `checkpoint_hi` representing the next checkpoint the pipeline expects to commit.
```

or something like that

crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -1706,25 +1705,31 @@ mod tests {
...
         let (committer_watermark, pruner_watermark) = test_init_watermark(Some(0), true).await;
-        // Indexer will not init the watermark, pipeline tasks will write commit watermarks as normal.
-        assert_eq!(committer_watermark, None);
-        assert_eq!(pruner_watermark, None);
+
+        let committer_watermark = committer_watermark.unwrap();
+        assert_eq!(committer_watermark.checkpoint_hi, 0);
+
+        let pruner_watermark = pruner_watermark.unwrap();
+        assert_eq!(pruner_watermark.pruner_hi, 0);
```
wlmyng, line 1724: I do think shifting to an exclusive checkpoint_hi makes this a bit easier to think about. The checkpoint_hi represents the next checkpoint a pipeline expects to commit, so when using default, or explicitly setting to 0, checkpoint_hi=0 makes sense because that's the next cp to index

crates/sui-indexer-alt-framework/src/postgres/mod.rs
```diff
@@ -167,8 +167,11 @@ pub mod tests {
     async fn test_add_existing_pipeline() {
         let (mut indexer, _temp_db) = Indexer::new_for_testing(&MIGRATIONS).await;
         {
-            let watermark = CommitterWatermark::new_for_testing(10);
             let mut conn = indexer.store().connect().await.unwrap();
+            conn.init_watermark(ConcurrentPipeline1::NAME, 0)
+                .await
+                .unwrap();
+            let watermark = CommitterWatermark::new_for_testing(11);
```
wlmyng, line 174: why not just set `conn.init_watermark(..., 11)`?

crates/sui-indexer-alt-framework/src/pipeline/mod.rs
```diff
@@ -88,7 +88,7 @@ impl<P: Processor> IndexedCheckpoint<P> {
         Self {
             watermark: CommitterWatermark {
                 epoch_hi_inclusive: epoch,
-                checkpoint_hi_inclusive: cp_sequence_number,
+                checkpoint_hi: cp_sequence_number + 1,
```
wlmyng, line 91: how come we do `.checked_sub(1)` and panic but not the equivalent add

PR #25745, opened 2026-03-07
[move-compiler] Relax asserts in borrows/mod.rs 

### LINKED ISSUE #25607: Compiler panic in `cfgir/borrows/mod.rs`: reference value in `BinopExp` operand
The compiler panics with an ICE in [`cfgir/borrows/mod.rs:346`](https://github.com/MystenLabs/sui/blob/67710b8a2b/external-crates/move/crates/move-compiler/src/cfgir/borrows/mod.rs#L346) when a dot-access expression has a dereference (`*`) as the "field name". The parser interprets `id(x).*x` as a binary expression where the right-hand operand is a reference, violating the `!v2.is_ref()` assertion in the borrow checker's `BinopExp` handler.

MRE:
```move
module 0x0::MRE {
    fun id<T>(x: &T): &T { x }

    fun t(x: &u64) {
        id(x).*x;
    }
}
```

Reproduce steps:
```
cd external-crates/move
cargo +nightly run --release -p move-compiler --bin move-check -- mre.move
```

Output:
```
thread 'main' panicked at crates/move-compiler/src/cfgir/borrows/mod.rs:346:13:
assertion failed: !v2.is_ref()
```

Expected behavior: a diagnostic error about the malformed expression, not an ICE.

Git commit: 67710b8a2b
Build: release

### LINKED ISSUE #25608: Compiler panic in `cfgir/borrows/state.rs`: `assert_single_value` fails on zero-field positional struct with unit argument
The compiler panics with an ICE in [`cfgir/borrows/state.rs:85`](https://github.com/MystenLabs/sui/blob/67710b8a2b/external-crates/move/crates/move-compiler/src/cfgir/borrows/state.rs#L85) when a zero-field positional struct is constructed with a unit expression `()` as argument. The `assert_single_value` function expects exactly one value, but the unit expression produces zero values. The arity mismatch should be caught as a type error before reaching the CFGIR borrow checker.

MRE:
```move
module 0x42::MRE {
    public struct A() has drop;

    fun f() {
        A(());
    }
}
```

Reproduce steps:
```
cd external-crates/move
cargo +nightly run --release -p move-compiler --bin move-check -- mre.move
```

Output:
```
thread 'main' panicked at crates/move-compiler/src/cfgir/borrows/state.rs:85:5:
assertion failed: values.len() == 1
```

Expected behavior: a diagnostic error about wrong number of arguments to struct constructor, not an ICE.

Git commit: 67710b8a2b
Build: release

tnowacki:
## Description 

- Relax the asserts into ice_asserts conditional on there being no errors from before borrow checking
- Fixes #25608 
- Fixes #25607 

## Test plan 

- Added tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-compiler/src/cfgir/borrows/state.rs
+++ b/external-crates/move/crates/move-compiler/src/cfgir/borrows/state.rs
@@ -81,11 +81,6 @@ pub struct BorrowState {
 // impls
 //**************************************************************************************************
 
-pub fn assert_single_value(mut values: Values) -> Value {
-    assert!(values.len() == 1);
-    values.pop().unwrap()
-}
-
 impl Value {
     pub fn is_ref(&self) -> bool {
         match self {


external-crates/move/crates/move-compiler/src/cfgir/borrows/mod.rs
```diff
@@ -145,6 +155,47 @@ fn unused_mut_borrows(context: &super::CFGContext, mutably_used: RefExpInfoMap)
...
+        }
+        v
+    }};
+}
+
+macro_rules! single_value_exp {
+    ($context:ident, $e:expr) => {{
+        let e = $e;
+        let values = exp($context, e);
+        assert_single_value!($context, values, e.exp.loc)
```
cgswords: I would consider inlining this macro for readability, as it is only called in this macro.

PR #25760, opened 2026-03-09
[rpc-alt] add address balance backward compat layer

emmazzz:
## Description 

Add a backward-compatibility layer to `sui-indexer-alt-jsonrpc` that synthesizes fake `Coin` objects from address balances, so clients that don't yet understand address balances can still use `suix_getCoins`,`sui_getObject`, and `sui_multiGetObjects` to discover and spend them. 

More specifically:                                                                                                                                                   
  - Extend `suix_getCoins` pagination to merge the synthetic AB coin into the real coin stream in correct sorted order (descending balance), handling pagination when the AB coin exists.                                                                                                                                                                                                               
  - Add `sui_getObject` / `sui_multiGetObjects` fallback: when an object ID isn't found, unmask it (XOR with chain identifier) and check if it resolves to a balance accumulator. If so, synthesize a `Coin` object.
  - Expose `ObjectByOwnerKey::encode_key` from `sui-indexer-alt-consistent-store` so the JSON-RPC layer can construct keys for AB coin cursor

## Test plan 

Added e2e tests covering: AB coin via getObject, AB coin via multiGetObjects (mixed with real and nonexistent objects), pagination with AB coin on first page, AB coin trimmed to next page, AB coin naturally on second page, AB coin as cursor boundary, AB coin as transaction input (coin reservation), and AB coin as gas payment

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_address_balance_coin_tests.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_address_balance_coin_tests.rs
@@ -323,7 +323,7 @@ async fn test_pagination_ab_coin_on_first_page() {
             .await
             .unwrap();
         assert!(fx.status().is_ok());
-        gas = fx.gas_object().0;
+        gas = fx.gas_object().unwrap().0;
         cluster
             .test_env
             .cluster
@@ -414,7 +414,7 @@ async fn test_pagination_ab_coin_trimmed_to_next_page() {
             .await
             .unwrap();
         assert!(fx.status().is_ok());
-        gas = fx.gas_object().0;
+        gas = fx.gas_object().unwrap().0;
         cluster
             .test_env
             .cluster
@@ -501,7 +501,7 @@ async fn test_pagination_ab_coin_on_second_page() {
             .await
             .unwrap();
         assert!(fx.status().is_ok());
-        gas = fx.gas_object().0;
+        gas = fx.gas_object().unwrap().0;
         cluster
             .test_env
             .cluster
@@ -587,7 +587,7 @@ async fn test_pagination_ab_coin_as_cursor() {
             .await
             .unwrap();
         assert!(fx.status().is_ok());
-        gas = fx.gas_object().0;
+        gas = fx.gas_object().unwrap().0;
         cluster
             .test_env
             .cluster


crates/sui-indexer-alt-jsonrpc/src/data/address_balance_coins.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    Ok(Some(coin))
+}
+
+/// Try to resolve an object ID as a masked address balance coin.
+///
+/// When `getObject` is called with an ID that doesn't exist, this function unmaskes the ID
+/// (XOR with chain identifier) and checks if it points to a balance accumulator object.
+/// If so, it synthesizes a Coin object to return to the client.
+pub(crate) async fn try_resolve_address_balance_object(
```
wlmyng, line 75: ooh pretty neat

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/address_balance_coins/get_coins.move
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Fund an address balance for B and verify that suix_getCoins returns the address balance coin.
+
+//# init --protocol-version 108 --addresses Test=0x0 --accounts A B --simulator --enable-accumulators --enable-address-balance-gas-payments
```
wlmyng, line 6: do we still need these flags since address balances are in mainnet?

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/address_balance_coins/get_coins_merged.snap
```diff
@@ -0,0 +1,81 @@
...
+        "previousTransaction": "ANBqdWhAdyjzhkFDhoFnkLmWfDQxooM2ZZh84AQR4Qy7"
+      },
+      {
+        "coinType": "0x0000000000000000000000000000000000000000000000000000000000000002::sui::SUI",
+        "coinObjectId": "0x087806e7e384e4d23ad866962f397e7d90a14fa12ef9a7402509142926095465",
+        "version": "2",
+        "digest": "3xmpjvy9GNH5CSg6nZsKgcfL2MQfzh8DgVhNANZaZT1u",
+        "balance": "300",
+        "previousTransaction": "9gD9hJtLRrmobfCYJrmJNFntkYTZ4tVjaKXxNtTSZWSw"
+      },
```
wlmyng, line 68: this is the fake address balance coin?

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/address_balance_coins/get_coins_pagination.snap
```diff
@@ -0,0 +1,110 @@
...
+        "version": "2",
+        "digest": "7wfBzVaoDL5a9xz8ViU2vvUjkkQKAA4Utgc9QC1b8Yax",
+        "balance": "500",
+        "previousTransaction": "ANBqdWhAdyjzhkFDhoFnkLmWfDQxooM2ZZh84AQR4Qy7"
+      }
+    ],
+    "nextCursor": "zAEAp7AycDh4qnTDEmk1eJ/R1NfhEdWRGwkkfWljBhwxK1oAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgAAAAAAAAAEY29pbgAAAAAAAAAEQ29pbgAAAAAAAAABAAAABwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAANzdWkAAAAAAAAAA1NVSQAAAAAAAAAAAf////////4LVu2TDkfRtAjZJ9xdNyJ7ViA+obJK2PSnia9OhFl81UI=",
+    "hasNextPage": true
+  }
+}
```
wlmyng, line 65: ah, cursor pagination tests ... but can we guarantee that we know there is a next page because of the address balance coin, and not that there's also the real 100 balance coin?

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/address_balance_coins/get_object.move
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Fund an address balance, get the address balance coin's object ID from getCoins, then call
+// sui_getObject with it to verify the synthesized Coin object is returned.
+// Note: The dynamic getObject flow (extracting coin ID from getCoins response) is tested in the
+// Rust e2e tests. Here we verify getCoins returns the address balance coin correctly.
```
wlmyng: i think we're missing this part of the test?

crates/sui-indexer-alt-jsonrpc/Cargo.toml
```diff
@@ -53,6 +53,7 @@ move-core-types.workspace = true
 sui-default-config.workspace = true
 sui-display.workspace = true
 sui-futures.workspace = true
+sui-indexer-alt-consistent-store.workspace = true
```
wlmyng, line 56: does this significantly impact build time/ size of the sui-indexer-alt-jsonrpc crate build?

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
```diff
@@ -159,27 +160,102 @@ impl CoinsApiServer for Coins {
...
+        // type, presented as a Coin object for clients that don't understand address balances.
+        let address_balance_coin = address_balance_coin_for_owner(ctx, owner, inner)
+            .await
+            .context("Failed to get address balance coin")
+            .map_err(RpcError::<Error>::from)?;
+
+        let mut cursor_override: Option<Vec<u8>> = None;
+
+        if let Some(ab_coin) = address_balance_coin {
+            use sui_indexer_alt_consistent_store::{ObjectByOwnerKey, OwnerKind};
```
wlmyng: does it make a diff whether we conditionally pull it in per jsonrpc request, or just pull it in at the top?

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
```diff
@@ -159,27 +160,102 @@ impl CoinsApiServer for Coins {
...
+
+                    let last_idx = page.limit as usize - 1;
+                    if pos == last_idx {
+                        // AB coin is the last item on the page — use its key as cursor.
+                        cursor_override = Some(ab_key_bytes);
+                    } else if pos < last_idx {
+                        // Last item is a store coin. AB took one slot, so the store coin
+                        // at page position `last_idx` was at index `last_idx - 1` before
+                        // insertion.
+                        cursor_override = Some(results.results[last_idx - 1].token.clone());
```
emmazzz: This can potentially be simplified and merged with the logic on line 238.

wlmyng: the if branch only exists because the token is created differently if ab or cb? But yeah, seems like at the end when we create the cursor, we can just check whether it was ab or cb and go accordingly

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_address_balance_coin_tests.rs
```diff
@@ -0,0 +1,837 @@
...
+        sui_types::coin::Coin::type_(GAS::type_tag()),
+        Some(!ab_coin.balance),
+        ab_coin.coin_object_id,
+    );
+    let ab_cursor = Base64::encode(bcs::to_bytes(&ab_key).unwrap());
+    assert_eq!(
+        ab_cursor,
+        cursor1.clone().unwrap(),
+        "Returned cursor should equal AB coin's encoded key"
+    );
```
wlmyng, line 650: cool cool exhaustive

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_address_balance_coin_tests.rs
```diff
@@ -0,0 +1,837 @@
...
+
+    // Verify recipient received funds as a regular coin
+    let CoinsResponse {
+        result: Page {
+            data: recipient_coins,
+            ..
+        },
+    } = cluster.get_coins(recipient, &gas_type, None, 10).await;
+    let recipient_balance: u64 = recipient_coins.iter().map(|c| c.balance).sum();
+    assert_eq!(recipient_balance, transfer_amount);
```
wlmyng, line 842: Oh ... the recipient receives a coni, instead of their address balance being updated?

crates/sui-indexer-alt-consistent-store/src/schema/object_by_owner.rs
```diff
@@ -21,7 +21,7 @@ use sui_indexer_alt_framework::types::object::Owner;
 /// Key for the index that supports fetching an owner's objects, optionally filtering by object
 /// type.
 #[derive(Encode, Decode, PartialEq, Eq)]
-pub(crate) struct Key {
+pub struct Key {
```
amnn, line 24: Is it possible to increase the visibility of these types purely by re-exporting them, or does Rust complain if you publicly export a non-`pub` type?

crates/sui-indexer-alt-consistent-store/src/schema/object_by_owner.rs
```diff
@@ -56,6 +56,23 @@ impl Key {
             object_id: obj.id(),
         })
     }
+
+    /// Encode a key for comparison with cursors. This is used by the JSON-RPC layer
```
amnn: nit: let's not include the explanation of why we're exposing this function here, that kind of documentation gets stale very quickly, because by nature of being public anyone is free to use the function for whatever they like.

crates/sui-indexer-alt-consistent-store/src/schema/object_by_owner.rs
```diff
@@ -56,6 +56,23 @@ impl Key {
...
+    /// to determine if an address balance coin should be included in pagination results.
+    pub fn encode_key(
+        kind: OwnerKind,
```
amnn: Maybe have this accept an `Owner` instead of `OwnerKind` to avoid having to export an extra type?

crates/sui-indexer-alt-consistent-store/src/lib.rs
```diff
@@ -64,6 +64,9 @@ mod metrics;
 pub mod restore;
 mod rpc;
 pub(crate) mod schema;
+
+// Re-export specific types for JSON-RPC pagination
+pub use schema::object_by_owner::{Key as ObjectByOwnerKey, OwnerKind};
 mod store;
```
amnn: File order, individual import per line, doc comment based on a specific usage site -- this `pub use` is in the middle of all the module declarations.


```suggestion
mod store;

pub use schema::object_by_owner::Key as ObjectByOwnerKey;
pub use schema::object_by_owner::OwnerKind;
```

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/address_balance_coins/get_coins.move
amnn: Isn't this flag more about having PTBs built by the test runner use address balances for gas selection?

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_address_balance_coin_tests.rs
```diff
@@ -0,0 +1,837 @@
...
+}
+
+/// Test pagination where the AB coin naturally falls on the second page because its balance is
+/// lower than all coins on the first page.
+///
+/// Setup: 4 real coins (5000, 4000, 3000, 2000) + AB coin (500). With limit=3, the first page
+/// contains only the top 3 real coins. The AB coin appears on the second page after the remaining
+/// real coin.
+#[tokio::test]
+async fn test_pagination_ab_coin_on_second_page() {
```
amnn, line 484: I'm not following the difference between these two tests -- is it just that the first test is a boundary case?

crates/sui-indexer-alt-jsonrpc/src/api/objects/response.rs
amnn: This code would now be better off as a cascade of `if let`'s:

```rust
let object = if let Some(object) = load_live(ctx, object_id)
    .await
    .context("Failed to load latest object")?
{
    object
} else if let Some(object) = try_resolve_address_balance_object(ctx, object_id)
    .await
    .context("Failed to resolve address balance object")?
{
    object
} else {
    return Ok(SuiObjectResponse::new_with_error(
        SuiObjectResponseError::NotExists { object_id },
    ));
};

Ok(SuiObjectResponse::new_with_data(
    object_data_with_options(ctx, object, options).await?
))
```

crates/sui-indexer-alt-jsonrpc/Cargo.toml
amnn: It would indeed be better to move whatever logic you need from here into `sui-indexer-alt-consistent-api`, rather than depend on the whole consistent store implementation for two types.

crates/sui-indexer-alt-jsonrpc/src/lib.rs
```diff
@@ -44,6 +44,7 @@ use crate::metrics::RpcMetrics;
 use crate::metrics::middleware::MetricsLayer;
 use crate::timeout::TimeoutLayer;
 
+mod address_balance_coins;
```
amnn: This does not belong at the top-level -- nested inside `data` might be a better fit? It would involve turning that into a nested module, and maybe taking the functions that exist there already and putting them under `data::object` or something.

crates/sui-indexer-alt-jsonrpc/src/context.rs
```diff
@@ -137,4 +170,27 @@ impl Context {
     pub(crate) fn config(&self) -> &RpcConfig {
         self.config.as_ref()
     }
+
```
amnn, line 174: I can somewhat get on board with having `chain_identifier` here, but `current_epoch` is very out of place. `Context` is not supposed to include arbitrary read functions, because if it did, it would become unmanageable.

In this repo, this accessor belongs as a free function inside the `data` module, maybe `data::epoch::current(ctx: &Context)`

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
amnn: Let's just pull it in at the top.

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
```diff
@@ -159,27 +160,102 @@ impl CoinsApiServer for Coins {
...
+            .zip(&coin_ids)
             .map(|(r, id)| r.with_internal_context(|| format!("Failed to get object {id}")))
             .collect::<Result<Vec<_>, _>>()?;
 
+        let mut has_next_page = results.has_next_page;
+
+        // Backward compatibility: synthesize an address balance coin and merge it with real
+        // coins. The address balance coin represents the owner's address balance for this coin
+        // type, presented as a Coin object for clients that don't understand address balances.
+        let address_balance_coin = address_balance_coin_for_owner(ctx, owner, inner)
```
amnn: Can you fetch this concurrently with the other coins?

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
```diff
@@ -159,27 +160,102 @@ impl CoinsApiServer for Coins {
...
+            .await
+            .context("Failed to get address balance coin")
+            .map_err(RpcError::<Error>::from)?;
```
amnn, line 175: Is the `map_err` necessary? `?` should do something like that for you, automatically.

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
```diff
@@ -159,27 +160,102 @@ impl CoinsApiServer for Coins {
...
+
+        let mut cursor_override: Option<Vec<u8>> = None;
+
+        if let Some(ab_coin) = address_balance_coin {
+            use sui_indexer_alt_consistent_store::{ObjectByOwnerKey, OwnerKind};
+
+            let ab_key_bytes = ObjectByOwnerKey::encode_key(
+                OwnerKind::AddressOwner(owner),
+                object_type.clone(),
+                Some(!ab_coin.balance), // Balance is stored as bitwise negation for descending order
```
amnn: This is an implementation detail of the encoding -- maybe make this part of the implementation for `encode_key`? Perhaps introduce a function that explicitly creates a key for an address balance coin or something?

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
```diff
@@ -159,27 +160,102 @@ impl CoinsApiServer for Coins {
...
+                ab_coin.coin_object_id,
+            );
+
+            // Determine if the address balance coin should be included on this page.
+            let include = if let Some(cursor) = &page.cursor {
```
amnn: This is all very complicated. Can we do the following?

- Associate each coin with its token from `results.results` in `coins`.
- Unconditionally add `address_balance_coin` into `coins` (along with its token), in sorted order, using `Vec::partition_point` to find the right place to insert it.
- Calculate `has_next_page` as whether the the response from the consistent store has a next page OR `coins > page.limit` now.
- Truncate the result to the page size.
- `cursor` is the cursor from the last element after truncation.
- `data` is the vector with the tokens stripped away.

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
```diff
@@ -393,6 +469,27 @@ async fn coin_metadata_response(
     Ok(Some(coin_metadata.into()))
 }
 
+/// Query the address balance for this owner/coin_type and synthesize an address balance coin if
+/// the balance is non-zero.
+async fn address_balance_coin_for_owner(
```
amnn: This seems spiritually similar to `coin_response`, should it be called `address_balance_coin_response`?

crates/sui-indexer-alt-jsonrpc/src/address_balance_coins.rs
```diff
@@ -0,0 +1,110 @@
...
+use sui_types::object::Owner;
+
+use crate::context::Context;
+use crate::data::load_live;
+
+/// Synthesize an address balance coin for the given owner and coin type.
+///
+/// Returns `None` if the accumulator object doesn't exist (e.g. the address has never received
+/// this coin type as an address balance).
+pub(crate) async fn get_address_balance_coin_info(
```
amnn: To accompany the suggestion to move this under `data`, maybe call this something like `load_address_balance_coin` to make it clear that it's fetching objects etc.

crates/sui-indexer-alt-jsonrpc/src/context.rs
```diff
@@ -47,6 +55,10 @@ pub(crate) struct Context {
 
     /// Access to the RPC's configuration.
     config: Arc<RpcConfig>,
+
+    /// The chain identifier (derived from the genesis checkpoint digest), used for masking
+    /// address balance coin object IDs.
```
amnn: nit: Do not explain the one specific use you have for introducing an otherwise general thing, it works against the point of an abstraction.

crates/sui-indexer-alt-jsonrpc/src/context.rs
```diff
@@ -91,6 +103,26 @@ impl Context {
...
+                .context("Failed to connect to the database")?;
+
+            let genesis_digest_bytes: Vec<u8> = conn
+                .first(g::kv_genesis.select(g::genesis_digest))
+                .await
+                .context("Failed to fetch genesis digest")?;
+
+            let bytes: [u8; 32] = genesis_digest_bytes
+                .try_into()
+                .map_err(|_| anyhow::anyhow!("Invalid genesis digest length"))?;
```
amnn:
```suggestion
                .ok()
                .context("Invalid genesis digest length")?;
```

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/address_balance_coins/get_object.move
emmazzz: True. Now this is just a duplicate of the other tests. Will remove this one.

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/address_balance_coins/get_coins_pagination.snap
emmazzz: We are not testing that. Here we are just testing pagination works (to the extent allowed by functional test runner) when a fake coin is returned.

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/address_balance_coins/get_coins_merged.snap
emmazzz: yes this is the fake coin.

crates/sui-indexer-alt-consistent-store/src/schema/object_by_owner.rs
emmazzz: Right rust doesn't allow me to publicly export a pub crate type.

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_address_balance_coin_tests.rs
emmazzz: Yes this is expected, aiming to test how address balance works with existing coin system.

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_address_balance_coin_tests.rs
emmazzz: Right the first test case tests out when the AB coin is at the boundary.

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
emmazzz: The `map_err` is required since `?` only does one `From` conversion, and the return type is `RpcResult (Result<_, ErrorObject>)`. Without it, Rust can't go `anyhow::Error → RpcError → ErrorObject` in one step.

crates/sui-indexer-alt-jsonrpc/Cargo.toml
emmazzz: Agreed that depending on the full consistent store just for the key encoding isn't ideal. However, `sui-indexer-alt-consistent-api` is currently purely protobuf/gRPC definitions and doesn't include any of the consistent store's key schemas or their bincode encoding. Doing this properly would mean moving all the consistent store schemas and the key encoding logic into a shared types crate (possibly renaming `sui-indexer-alt-consistent-api` to something like `sui-indexer-alt-consistent-types`, or creating a new crate). That feels like a bigger refactor that's orthogonal to this PR that we could consider doing as a follow-up.

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4874,6 +4884,18 @@ impl ProtocolConfig {
             OverrideGuard
         })
     }
+
+    /// Process-wide variant of [`apply_overrides_for_testing`]. Use this when the test spawns
```
wlmyng: what is this for?

crates/sui-indexer-alt-consistent-store/src/schema/object_by_owner.rs
```diff
@@ -56,6 +56,21 @@ impl Key {
...
+        object_id: ObjectID,
+    ) -> Vec<u8> {
+        let key = Key {
+            kind: OwnerKind::from_owner(owner),
+            type_,
+            balance: Some(!balance),
+            object_id,
+        };
+        crate::db::key::encode(&key)
+    }
```
amnn: It would be better to offer the more generic thing here -- it also justifies its existence as a constructor/associated function on this type, but I acknowledge that this means you would also need to expose the encoding function which maybe is not desirable?

```suggestion
    pub fn from_coin_parts(
        owner: &Owner,
        type_: StructTag,
        balance: u64,
        object_id: ObjectID,
    ) -> Key {
        Key {
            kind: OwnerKind::from_owner(owner),
            type_,
            balance: Some(!balance),
            object_id,
        }
    }
```

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
```diff
@@ -159,27 +162,70 @@ impl CoinsApiServer for Coins {
...
+
+        // Pair each coin with its cursor token so we can derive the final cursor from
+        // whichever coin ends up last on the page.
+        let mut coins: Vec<(Coin, Vec<u8>)> = coin_results
+            .into_iter()
+            .zip(&coin_ids)
+            .map(|(r, id)| r.with_internal_context(|| format!("Failed to get object {id}")))
+            .collect::<Result<Vec<_>, _>>()?
+            .into_iter()
+            .zip(results.results.iter().map(|e| e.token.clone()))
```
amnn: nit: seeing as you don't refer to `results.results` after this it should be okay to move out of it:

```suggestion
            .zip(results.results.into_iter().map(|e| e.token))
```

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
```diff
@@ -159,27 +162,70 @@ impl CoinsApiServer for Coins {
...
+                .is_none_or(|cursor| ab_token > cursor.0);
+
+            if include_ab_coin {
+                let pos = coins.partition_point(|(c, _)| {
+                    c.balance > ab_coin.balance
+                        || (c.balance == ab_coin.balance
+                            && c.coin_object_id < ab_coin.coin_object_id)
+                });
+                coins.insert(pos, (ab_coin, ab_token));
+            }
```
amnn: You should be able to compare the token of the coin with the `ab_token`, to avoid having to duplicate the key logic, and it should be possible to do this unconditionally:

```suggestion
            let ab_token = ObjectByOwnerKey::encode_coin_key(
                &Owner::AddressOwner(owner),
                object_type.clone(),
                ab_coin.balance,
                ab_coin.coin_object_id,
            );

            let pos = coins.partition_point(|(_, t)| t < ab_token);
            coins.insert(pos, (ab_coin, ab_token));
```

crates/sui-indexer-alt-jsonrpc/src/data/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+
+pub(crate) use object::{load_live, load_live_deserialized};
+
+use anyhow::Context as _;
+use diesel::ExpressionMethods;
+use diesel::QueryDsl;
+use sui_indexer_alt_schema::schema::kv_epoch_starts;
+use sui_types::committee::EpochId;
+
+use crate::context::Context;
```
amnn: nit:

```suggestion
mod address_balance_coins;
mod object;

use anyhow::Context as _;
use diesel::ExpressionMethods;
use diesel::QueryDsl;
use sui_indexer_alt_schema::schema::kv_epoch_starts;
use sui_types::committee::EpochId;

use crate::context::Context;

pub(crate) use address_balance_coins::load_address_balance_coin;
pub(crate) use object::load_live;
pub(crate) use object::load_live_deserialized;
```

(and then update the imports for `load_address_balance_coin` accordingly).

crates/sui-protocol-config/src/lib.rs
amnn: I'm guessing this might be a rebase artifact as @emmazzz's work is sitting on top of Mark's branch.

crates/sui-indexer-alt-jsonrpc/Cargo.toml
amnn: Agreed -- could you task that up under the Better Engineering project?

crates/sui-protocol-config/src/lib.rs
emmazzz: Surprise! I actually added this because using fake coin in transaction requires `enable_coin_reservation_for_testing` override for protocol config. Mark's tests  are single threaded sim tests where the thread local override suffices but our test cases can't use sim test (sim test does not work with pg db connections) and afaict need multi thread override, hence this change. I've asked @mystenmark  to take a look here.

emmazzz:
cc @mystenmark to review the protocol config override change

crates/sui-protocol-config/src/lib.rs
mystenmark: typically we do this sort of thing via a single API for both msim/non-msim, by using #[cfg]

see `POISON_VERSION_METHODS` in this file for an example

crates/sui-indexer-alt-jsonrpc/src/api/coin.rs
emmazzz: Comparing the token directly makes sense. But I don't think this can be done unconditionally. `include_ab_coin` is true when there is a cursor and the ab coin was already shown on a previous page. Inserting unconditional would cause the ab coin to be returned again as the first item of the current page.

crates/sui-indexer-alt-jsonrpc/Cargo.toml
emmazzz: Filed DVX-2053.

PR #25765, opened 2026-03-09
Store trusted peer addresses received via discovery

aschran:
## Description 

Saves received addresses for preconfigured peers to a local YAML file, loaded on startup.

Also adds buffering for consensus address updates received before consensus is ready yet.

## Test plan 

CI, added new unit tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-network/src/discovery/mod.rs
+++ b/crates/sui-network/src/discovery/mod.rs
@@ -120,7 +120,7 @@ struct State {
     connected_peers: HashMap<PeerId, ()>,
     known_peers: HashMap<PeerId, VerifiedSignedNodeInfo>,
     known_peers_v2: HashMap<PeerId, VerifiedSignedVersionedNodeInfo>,
-    peer_address_overrides: HashMap<PeerId, BTreeMap<AddressSource, Vec<anemo::types::Address>>>,
+    peer_addresses: HashMap<PeerId, BTreeMap<AddressSource, Vec<anemo::types::Address>>>,
 }
 
 /// The information necessary to dial another peer.
@@ -402,7 +402,7 @@ impl DiscoveryEventLoop {
             }
         }
 
-        self.save_stored_peers_if_changed();
+        self.save_stored_peers();
         info!("Discovery ended");
     }
 
@@ -424,11 +424,11 @@ impl DiscoveryEventLoop {
                     &self.endpoint_manager,
                 );
                 if changed {
-                    self.save_stored_peers_if_changed();
+                    self.save_stored_peers();
                 }
             }
             DiscoveryMessage::ConfiguredPeersUpdated => {
-                self.save_stored_peers_if_changed();
+                self.save_stored_peers();
             }
         }
     }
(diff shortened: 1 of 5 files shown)


crates/sui-node/src/lib.rs
```diff
@@ -1095,7 +1095,15 @@ impl SuiNode {
             .with_metrics(prometheus_registry)
             .build();
 
-        let mut discovery_builder = discovery::Builder::new().config(config.p2p_config.clone());
+        let mut p2p_config = config.p2p_config.clone();
+        {
+            let disc = p2p_config.discovery.get_or_insert_with(Default::default);
+            if disc.peer_addr_store_path.is_none() {
+                disc.peer_addr_store_path =
+                    Some(config.db_path().join("discovery_peer_cache.yaml"));
```
alex-mysten, line 1103: why doesn't this defaulting live in p2p.rs?

crates/sui-network/src/discovery/mod.rs
```diff
@@ -554,6 +572,71 @@ impl DiscoveryEventLoop {
...
+            if !self.configured_peers.contains_key(&peer_id) {
+                continue;
+            }
+            let anemo_addrs: Vec<_> = entry
+                .p2p_addresses()
+                .iter()
+                .filter_map(|addr| addr.to_anemo_address().ok())
+                .collect();
+            if !anemo_addrs.is_empty() {
+                p2p_overrides.push((peer_id, anemo_addrs));
```
alex-mysten: What happens if a configured node changes IP (say, back to original) when this node is offline?

crates/sui-network/src/discovery/mod.rs
```diff
@@ -554,6 +572,71 @@ impl DiscoveryEventLoop {
         }
     }
 
+    fn load_stored_peers_on_startup(&mut self) {
```
alex-mysten, line 599: What happens if all seed nodes change IPs while this node is offline?

crates/sui-network/src/discovery/mod.rs
alex-mysten, line 1214: If a validator is down for > 24 hours, we'd filter all stored peer addrs, correct?

crates/sui-node/src/lib.rs
aschran: I wanted it to reuse the db_path which is in the NodeConfig, not directly available in the P2P configs (unless I copy it somehow which could do if you think this is too ugly)

crates/sui-network/src/discovery/mod.rs
aschran: We will hopefully receive it via gossip from other connected nodes, however I can shore this up so that it will fall back after the 1-day expiration in case gossip fails

crates/sui-network/src/discovery/mod.rs
aschran: ALL nodes changing while this node is offline seems quite unlikely unless it's offline for a very long time, but if so the operator could just delete the stored address file

crates/sui-network/src/discovery/mod.rs
aschran: yes but I think the part that actually clears this from the anemo network configuration itself is missing, will update

crates/sui-network/src/discovery/mod.rs
```diff
@@ -542,18 +580,66 @@ impl DiscoveryEventLoop {
...
+        );
+
+        // update_known_peers_versioned forwards addresses through the mailbox.
+        // Drain it now so addresses are applied before the event loop starts.
+        while let Ok(msg) = self.mailbox.try_recv() {
+            self.handle_message(msg);
+        }
+    }
+
+    fn save_stored_peers_if_changed(&self) {
```
alex-mysten: Nit: doesn't this save unconditionally?

alex-mysten (approved):
LGTM! One nit

crates/sui-network/src/discovery/mod.rs
aschran: good point, fixed name

PR #25770, opened 2026-03-09
Remove unnecessary --simulator usage in transactional tests

mystenmark:
## Summary
- Remove the `--simulator` requirement for several protocol config options: `--enable-accumulators`, `--enable-address-balance-gas-payments`, and `--reference-gas-price`
- Pass `reference_gas_price` through to the non-simulator executor chain so tests can specify custom gas prices without requiring simulator mode
- Update test files that don't need simulator functionality (no `create-checkpoint` usage) to run without `--simulator`

### Tests that can now run without `--simulator`:
- `address_balance_gas.move` - tests address balance gas payments
- `address_balance_gas_transfer.move` - tests transferring GasCoin with address balance gas
- `coin_deny_and_undeny_address_balance_receiver.move` - tests deny list with accumulators
- `gas_api.move` - tests gas price APIs

### Tests that still need `--simulator` (use `create-checkpoint` which requires simulator):
- `balance_read.move`
- `large_balance_transfer.move`
- `withdraw_and_send_account.move`
- `emit_authenticated.move`

## Test plan
- [ ] Run the updated transactional tests to verify they pass without simulator mode

### DIFF
--- a/crates/sui-adapter-transactional-tests/tests/address_balances/large_balance_transfer.snap
+++ b/crates/sui-adapter-transactional-tests/tests/address_balances/large_balance_transfer.snap
@@ -87,5 +87,4 @@ task 13, lines 73-77:
 //> 1: sui::balance::redeem_funds<test::large_balance::MARKER>(Input(1));
 //> 2: sui::balance::join<test::large_balance::MARKER>(Result(0), Result(1));
 //> 3: sui::balance::send_funds<test::large_balance::MARKER>(Result(0), Input(2));
-Error: Transaction Effects Status: Move Primitive Runtime Error. Location: sui::funds_accumulator::withdraw_from_accumulator_address (function index 9) at offset 0. Arithmetic error, stack overflow, max value depth, etc.
-Debug of error: MovePrimitiveRuntimeError(MoveLocationOpt(Some(MoveLocation { module: ModuleId { address: sui, name: Identifier("funds_accumulator") }, function: 9, instruction: 0, function_name: Some("withdraw_from_accumulator_address") }))) at command Some(1)
+Error: Error checking transaction input objects: Invalid withdraw reservation: Balance withdraw reservation overflow
--- a/crates/sui-adapter-transactional-tests/tests/address_balances/balance_read.move
+++ b/crates/sui-adapter-transactional-tests/tests/address_balances/balance_read.move
@@ -3,7 +3,7 @@
 
 // Test settled_funds_value from sui::balance
 
-//# init --addresses test=0x0 --accounts A B C D --enable-accumulators --simulator
+//# init --addresses test=0x0 --accounts A B C D --enable-accumulators
 
 //# publish
 module test::balance_read;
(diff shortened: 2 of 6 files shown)


tnowacki (commented):
It looks fine to me, but could you add some additional tests for paying for gas? Or maybe there are these tests already? 

The tests you removed --simulator from aren't really inspecting the balance after the transaction is executed, which is the bit I want to make sure is working properly before enabling address balance gas payments outside of the simulator

mystenmark:
> It looks fine to me, but could you add some additional tests for paying for gas? Or maybe there are these tests already?
> 
> The tests you removed --simulator from aren't really inspecting the balance after the transaction is executed, which is the bit I want to make sure is working properly before enabling address balance gas payments outside of the simulator

There are such tests which aren't included in this PR, but will be handled in the next one

crates/sui-adapter-transactional-tests/tests/address_balances/large_balance_transfer.snap
```diff
@@ -87,5 +87,4 @@ task 13, lines 73-77:
 //> 1: sui::balance::redeem_funds<test::large_balance::MARKER>(Input(1));
 //> 2: sui::balance::join<test::large_balance::MARKER>(Result(0), Result(1));
 //> 3: sui::balance::send_funds<test::large_balance::MARKER>(Result(0), Input(2));
-Error: Transaction Effects Status: Move Primitive Runtime Error. Location: sui::funds_accumulator::withdraw_from_accumulator_address (function index 9) at offset 0. Arithmetic error, stack overflow, max value depth, etc.
-Debug of error: MovePrimitiveRuntimeError(MoveLocationOpt(Some(MoveLocation { module: ModuleId { address: sui, name: Identifier("funds_accumulator") }, function: 9, instruction: 0, function_name: Some("withdraw_from_accumulator_address") }))) at command Some(1)
+Error: Error checking transaction input objects: Invalid withdraw reservation: Balance withdraw reservation overflow
```
tnowacki, line 90: This is good?

crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas_transfer.snap
```diff
@@ -1,13 +1,26 @@
...
 A: object(0,0)
 
-task 1, lines 9-10:
+task 1, lines 9-12:
+//# programmable --sender A --inputs 10000000000 @A
+//> 0: SplitCoins(Gas, [Input(0)]);
+//> 1: sui::coin::into_balance<sui::sui::SUI>(Result(0));
+//> 2: sui::balance::send_funds<sui::sui::SUI>(Result(1), Input(1));
+mutated: object(0,0)
+accumulators_written: (object(1,0), A, sui::balance::Balance<sui::sui::SUI>, Merge)
```
tnowacki, line 16: 🎉

crates/sui-adapter-transactional-tests/tests/address_balances/large_balance_transfer.snap
mystenmark: yes, simulator didn't do validity checks before running

mystenmark: and, we still test the execution error path above, since it can be caused by deposits that are not detectable via validity checks

PR #25774, opened 2026-03-09
sui-indexer-alt: init pipelines to checkpoint_hi_inclusive = reader_lo - 1

evan-wall-mysten:
## Description 

This PR initializes pipelines in the `checkpoint_hi_inclusive = reader_lo - 1` state. This is a temporary state that indicates that no checkpoints have been indexed yet. After a checkpoint has been indexed and the committer watermark has been updated, `reader_lo <= checkpoint_hi_inclusive` will always be true. When in the `checkpoint_hi_inclusive = reader_lo - 1` state, the watermark record is ignored.

## Test plan 

Updated existing unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-analytics-indexer/src/store/migration.rs
+++ b/crates/sui-analytics-indexer/src/store/migration.rs
@@ -24,6 +24,7 @@ use object_store::PutPayload;
 use object_store::UpdateVersion;
 use object_store::path::Path as ObjectPath;
 use sui_indexer_alt_framework_store_traits::CommitterWatermark;
+use sui_indexer_alt_framework_store_traits::InitWatermark;
 use sui_storage::object_store::util::find_all_dirs_with_epoch_prefix;
 use thiserror::Error;
 use tracing::debug;
@@ -260,12 +261,16 @@ impl MigrationStore {
     pub(crate) async fn init_watermark(
         &self,
         pipeline: &str,
-        _default_next_checkpoint: u64,
-    ) -> anyhow::Result<Option<u64>> {
-        Ok(self
+        init_watermark: InitWatermark,
+    ) -> anyhow::Result<InitWatermark> {
+        let checkpoint_hi_inclusive = self
             .committer_watermark(pipeline)
             .await?
-            .map(|w| w.checkpoint_hi_inclusive))
+            .map(|w| w.checkpoint_hi_inclusive);
+        Ok(InitWatermark {
+            checkpoint_hi_inclusive,
+            ..init_watermark
+        })
     }
 
     /// Update watermark for a single pipeline after successful file upload.
--- a/crates/sui-analytics-indexer/src/store/mod.rs
+++ b/crates/sui-analytics-indexer/src/store/mod.rs
@@ -34,6 +34,7 @@ use sui_indexer_alt_framework::store::Connection;
 use sui_indexer_alt_framework::store::Store;
 use sui_indexer_alt_framework::store::TransactionalStore;
 use sui_indexer_alt_framework_store_traits::CommitterWatermark;
+use sui_indexer_alt_framework_store_traits::InitWatermark;
 use sui_indexer_alt_framework_store_traits::PrunerWatermark;
 use sui_indexer_alt_framework_store_traits::ReaderWatermark;
 use sui_types::base_types::EpochId;
@@ -123,7 +124,6 @@ pub use migration::FileRangeEntry;
 pub use migration::FileRangeIndex;
 pub use migration::MigrationStore;
 pub use migration::WatermarkUpdateError;
-
 use uploader::PendingFileUpload;
 
 /// The operational mode of the analytics store.
@@ -628,27 +628,27 @@ impl Connection for AnalyticsConnection<'_> {
     /// Initialize watermark.
     ///
     /// In live mode: Watermarks are derived from file names, so just delegates to `committer_watermark`.
-    /// In migration mode: If no watermark exists and `default_next_checkpoint > 0`, initializes
-    /// the watermark to `default_next_checkpoint - 1` so migration starts from the configured
-    /// `first_checkpoint`.
+    /// In migration mode: Delegates to `MigrationStore::init_watermark`.
     async fn init_watermark(
         &mut self,
         pipeline_task: &str,
-        default_next_checkpoint: u64,
-    ) -> anyhow::Result<Option<u64>> {
+        init_watermark: InitWatermark,
+    ) -> anyhow::Result<InitWatermark> {
         match &self.store.mode {
             StoreMode::Live(_) => {
                 // Live mode: derive from file names
-                Ok(self
+                let checkpoint_hi_inclusive = self
                     .committer_watermark(pipeline_task)
                     .await?
-                    .map(|w| w.checkpoint_hi_inclusive))
+                    .map(|w| w.checkpoint_hi_inclusive);
+                Ok(InitWatermark {
+                    checkpoint_hi_inclusive,
+                    ..init_watermark
+                })
             }
             StoreMode::Migration(store) => {
                 let output_prefix = self.pipeline_config(pipeline_task).output_prefix();
-                store
-                    .init_watermark(output_prefix, default_next_checkpoint)
-                    .await
+                store.init_watermark(output_prefix, init_watermark).await
             }
         }
     }
--- a/crates/sui-indexer-alt-consistent-store/src/store/mod.rs
+++ b/crates/sui-indexer-alt-consistent-store/src/store/mod.rs
@@ -13,6 +13,7 @@ use prometheus::Registry;
 use scoped_futures::ScopedBoxFuture;
 use sui_indexer_alt_framework::service::Service;
 use sui_indexer_alt_framework::store::CommitterWatermark;
+use sui_indexer_alt_framework::store::InitWatermark;
 use sui_indexer_alt_framework::store::Store as _;
 use sui_indexer_alt_framework::store::{self};
 
@@ -165,11 +166,19 @@ impl<S: Send + Sync + 'static> store::TransactionalStore for Store<S> {
 
 #[async_trait::async_trait]
 impl<S: Send + Sync> store::Connection for Connection<'_, S> {
-    async fn init_watermark(&mut self, pipeline_task: &str, _: u64) -> anyhow::Result<Option<u64>> {
-        Ok(self
+    async fn init_watermark(
+        &mut self,
+        pipeline_task: &str,
+        init_watermark: InitWatermark,
+    ) -> anyhow::Result<InitWatermark> {
+        let checkpoint_hi_inclusive = self
             .committer_watermark(pipeline_task)
             .await?
-            .map(|w| w.checkpoint_hi_inclusive))
+            .map(|w| w.checkpoint_hi_inclusive);
+        Ok(InitWatermark {
+            checkpoint_hi_inclusive,
+            ..init_watermark
+        })
     }
 
     async fn committer_watermark(
--- a/crates/sui-indexer-alt-e2e-tests/src/lib.rs
+++ b/crates/sui-indexer-alt-e2e-tests/src/lib.rs
@@ -463,6 +463,7 @@ impl OffchainCluster {
         let latest: HashMap<String, i64> = w::watermarks
             .select((w::pipeline, w::checkpoint_hi_inclusive))
             .filter(w::pipeline.eq_any(&self.pipelines))
+            .filter(w::reader_lo.le(w::checkpoint_hi_inclusive))
             .load(&mut conn)
             .await?
             .into_iter()
--- a/crates/sui-indexer-alt-framework-store-traits/src/lib.rs
+++ b/crates/sui-indexer-alt-framework-store-traits/src/lib.rs
@@ -13,13 +13,14 @@ use scoped_futures::ScopedBoxFuture;
 /// operations, agnostic of the underlying store implementation.
 #[async_trait]
 pub trait Connection: Send {
-    /// If no existing watermark record exists, initializes it with `default_next_checkpoint`.
-    /// Returns the committer watermark `checkpoint_hi_inclusive`.
+    /// Returns the `InitWatermark` based on the existing watermark if it exists.
+    /// Otherwise, initializes a new watermark record with `InitWatermark` and returns
+    /// the value passed in.
     async fn init_watermark(
         &mut self,
         pipeline_task: &str,
-        default_next_checkpoint: u64,
-    ) -> anyhow::Result<Option<u64>>;
+        init_watermark: InitWatermark,
+    ) -> anyhow::Result<InitWatermark>;
 
     /// Given a `pipeline_task` representing either a pipeline name or a pipeline with an associated
     /// task (formatted as `{pipeline}{Store::DELIMITER}{task}`), return the committer watermark
@@ -114,6 +115,13 @@ pub trait TransactionalStore: Store {
         F: for<'r> FnOnce(
             &'r mut Self::Connection<'_>,
         ) -> ScopedBoxFuture<'a, 'r, anyhow::Result<R>>;
+}
+
+/// Used during watermark initialization to set and return state.
+#[derive(Default, Debug, Clone, Copy, PartialEq)]
+pub struct InitWatermark {
+    pub checkpoint_hi_inclusive: Option<u64>,
+    pub reader_lo: u64,
 }
 
 /// Represents the highest checkpoint for some pipeline that has been processed by the indexer
--- a/crates/sui-indexer-alt-framework/src/lib.rs
+++ b/crates/sui-indexer-alt-framework/src/lib.rs
@@ -15,6 +15,7 @@ use ingestion::ingestion_client::IngestionClient;
 use metrics::IndexerMetrics;
 use prometheus::Registry;
 use sui_indexer_alt_framework_store_traits::Connection;
+use sui_indexer_alt_framework_store_traits::InitWatermark;
 use sui_indexer_alt_framework_store_traits::Store;
 use sui_indexer_alt_framework_store_traits::TransactionalStore;
 use sui_indexer_alt_framework_store_traits::pipeline_task;
@@ -375,13 +376,21 @@ impl<S: Store> Indexer<S> {
         let pipeline_task =
             pipeline_task::<S>(P::NAME, self.task.as_ref().map(|t| t.task.as_str()))?;
 
-        let checkpoint_hi_inclusive = conn
-            .init_watermark(&pipeline_task, self.default_next_checkpoint)
+        let InitWatermark {
+            checkpoint_hi_inclusive,
+            reader_lo,
+        } = conn
+            .init_watermark(
+                &pipeline_task,
+                InitWatermark {
+                    checkpoint_hi_inclusive: self.default_next_checkpoint.checked_sub(1),
+                    reader_lo: self.default_next_checkpoint,
+                },
+            )
             .await
             .with_context(|| format!("Failed to init watermark for {pipeline_task}"))?;
 
-        let next_checkpoint =
-            checkpoint_hi_inclusive.map_or(self.default_next_checkpoint, |c| c + 1);
+        let next_checkpoint = checkpoint_hi_inclusive.map_or(reader_lo, |c| c + 1);
 
         self.first_ingestion_checkpoint = next_checkpoint.min(self.first_ingestion_checkpoint);
 
@@ -698,6 +707,10 @@ mod tests {
         test_pipeline!(D, "sequential_d");
 
         let mut conn = store.connect().await.unwrap();
+
+        conn.init_watermark(A::NAME, InitWatermark::default())
+            .await
+            .unwrap();
         conn.set_committer_watermark(
             A::NAME,
             CommitterWatermark {
@@ -707,6 +720,10 @@ mod tests {
         )
         .await
         .unwrap();
+
+        conn.init_watermark(B::NAME, InitWatermark::default())
+            .await
+            .unwrap();
         conn.set_committer_watermark(
             B::NAME,
             CommitterWatermark {
@@ -716,6 +733,10 @@ mod tests {
         )
         .await
         .unwrap();
+
+        conn.init_watermark(C::NAME, InitWatermark::default())
+            .await
+            .unwrap();
         conn.set_committer_watermark(
             C::NAME,
             CommitterWatermark {
@@ -725,6 +746,10 @@ mod tests {
         )
         .await
         .unwrap();
+
+        conn.init_watermark(D::NAME, InitWatermark::default())
+            .await
+            .unwrap();
         conn.set_committer_watermark(
             D::NAME,
             CommitterWatermark {
@@ -2053,10 +2078,10 @@ mod tests {
         }
         let main_pipeline_watermark = store.watermark("test").unwrap();
         // assert that the main pipeline's watermarks are not updated
-        assert_eq!(main_pipeline_watermark.checkpoint_hi_inclusive, 10);
+        assert_eq!(main_pipeline_watermark.checkpoint_hi_inclusive, Some(10));
         assert_eq!(main_pipeline_watermark.reader_lo, 5);
         let tasked_pipeline_watermark = store.watermark("test@task").unwrap();
-        assert_eq!(tasked_pipeline_watermark.checkpoint_hi_inclusive, 25);
+        assert_eq!(tasked_pipeline_watermark.checkpoint_hi_inclusive, Some(25));
         assert_eq!(tasked_pipeline_watermark.reader_lo, 9);
     }
 
--- a/crates/sui-indexer-alt-framework/src/mocks/store.rs
+++ b/crates/sui-indexer-alt-framework/src/mocks/store.rs
@@ -1,7 +1,6 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::ops::Deref;
 use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::atomic::AtomicUsize;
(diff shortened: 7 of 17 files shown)


crates/sui-pg-db/src/store.rs
```diff
@@ -63,31 +55,37 @@ impl store::Connection for Connection<'_> {
...
                 watermarks::checkpoint_hi_inclusive,
                 watermarks::tx_hi,
                 watermarks::timestamp_ms_hi_inclusive,
+                watermarks::reader_lo,
             ))
             .filter(watermarks::pipeline.eq(pipeline_task))
             .first(self)
-            .await
-            .optional()?;
+            .await?;
```
evan-wall-mysten, line 88: Now that `init_watermark` always creates a watermark record, removing `.optional()` here helps distinguish between a missing watermark (error) and `checkpoint_hi_inclusive = reader_lo - 1` (will be fixed when committer watermark is updated).

crates/sui-pg-db/src/store.rs
```diff
@@ -153,33 +155,17 @@ impl store::Connection for Connection<'_> {
...
-            pruner_hi: 0,
-        };
-
-        use diesel::query_dsl::methods::FilterDsl;
-        Ok(diesel::insert_into(watermarks::table)
-            .values(&stored_watermark)
-            // There is an existing entry, so only write the new `hi` values
-            .on_conflict(watermarks::pipeline)
-            .do_update()
+        Ok(diesel::update(watermarks::table)
```
evan-wall-mysten, line 165: Because the watermark record is always created in `init_watermark`, this is always an update.

crates/sui-analytics-indexer/src/store/migration.rs
```diff
@@ -260,12 +260,12 @@ impl MigrationStore {
...
         pipeline: &str,
-        _default_next_checkpoint: u64,
-    ) -> anyhow::Result<Option<u64>> {
+        default_next_checkpoint: u64,
+    ) -> anyhow::Result<u64> {
         Ok(self
             .committer_watermark(pipeline)
             .await?
-            .map(|w| w.checkpoint_hi_inclusive))
+            .map_or(default_next_checkpoint, |w| w.checkpoint_hi_inclusive + 1))
```
wlmyng: I still think it would be great if `init_watermark` didn't do double duty in 1) trying to initialize the watermark 2) return the committer watermark. Store operators should just define how to populate the full watermark given the `default_next_checkpoint`. The indexer framework is responsible for passing in `default_next_checkpoint = checkpoint_hi_inclusive = reader_lo - 1`, and is also responsible for determining the next checkpoint. This next checkpoint is always `checkpoint_hi_inclusive + 1`.

crates/sui-indexer-alt-e2e-tests/src/lib.rs
```diff
@@ -472,7 +472,11 @@ impl OffchainCluster {
             return Ok(None);
         }
 
-        Ok(latest.into_values().min().map(|l| l as u64))
+        let min = latest
+            .into_values()
+            .min()
+            .and_then(|c| u64::try_from(c).ok());
+        Ok(min)
```
wlmyng: does the current code not compile anymore or something? clippy?

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -13,13 +13,14 @@ use scoped_futures::ScopedBoxFuture;
 /// operations, agnostic of the underlying store implementation.
 #[async_trait]
 pub trait Connection: Send {
-    /// If no existing watermark record exists, initializes it with `default_next_checkpoint`.
-    /// Returns the committer watermark `checkpoint_hi_inclusive`.
+    /// Returns the `default_next_checkpoint` based on the existing watermark if it exists.
+    /// Otherwise, initializes a new watermark record with `default_next_checkpoint` and returns
+    /// the value passed in.
```
wlmyng: Or, `init_watermark` can continue to return a value, but just the existing watermark, not the `default_next_checkpoint`. The `default_next_checkpoint = checkpoint_hi_inclusive + 1` is indexer framework specific logic. Otherwise, every store will need to remember that they need to do a `checkpoint_hi_inclusive + 1` - in which case we should specify it in the trait comments - but then why not just consolidate it in one place, in the framework logic

crates/sui-indexer-alt-framework/src/mocks/store.rs
```diff
@@ -25,7 +25,9 @@ use crate::store::TransactionalStore;
 #[derive(Default, Clone)]
 pub struct MockWatermark {
     pub epoch_hi_inclusive: u64,
-    pub checkpoint_hi_inclusive: u64,
+    // Because checkpoint_hi_inclusive is a u64 here, None is similar to a -1 in postgres where the
+    // pipeline has been initialized, but no checkpoints have been indexed.
+    pub checkpoint_hi_inclusive: Option<u64>,
```
wlmyng:
```suggestion
    // checkpoint_hi_inclusive is a u64 or None to indicate when a pipeline has been initialized but no checkpoints have been indexed.
    pub checkpoint_hi_inclusive: Option<u64>,
```

crates/sui-indexer-alt-framework/src/mocks/store.rs
```diff
@@ -84,85 +86,99 @@ pub struct MockStore {
...
-        } = self
+    ) -> anyhow::Result<u64> {
+        let watermark = self
             .0
             .watermarks
             .entry(pipeline_task.to_string())
             .or_insert(MockWatermark {
                 epoch_hi_inclusive: 0,
-                checkpoint_hi_inclusive,
+                checkpoint_hi_inclusive: default_next_checkpoint.checked_sub(1),
```
wlmyng: in sui-analytics-indexer, consistent-store, we don't seem to do the `checked_sub`, instead accepting and writing `default_next_checkpoint` as is. This indicates to me that the framework should be responsible for determining what values the store should write, and the store just accepts the value and defines how it actually writes to the underlying db.

wlmyng: btw, shouldn't we set `checkpoint_hi_inclusive` to `None` if `default_next_checkpoint - 1` is < 0?

crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
```diff
@@ -635,6 +635,6 @@ mod tests {
 
         // Verify watermark has progressed
         let watermark = setup.store.watermark(DataPipeline::NAME).unwrap();
-        assert_eq!(watermark.checkpoint_hi_inclusive, 1);
+        assert_eq!(watermark.checkpoint_hi_inclusive, Some(1));
```
wlmyng, line 638: i think we should have a test for when someone specifies `--first-checkpoint 0`, or even when no `--first-checkpoint` specified, if not present already

crates/sui-indexer-alt-framework/src/postgres/mod.rs
```diff
@@ -169,6 +169,9 @@ pub mod tests {
         {
             let watermark = CommitterWatermark::new_for_testing(10);
             let mut conn = indexer.store().connect().await.unwrap();
+            conn.init_watermark(ConcurrentPipeline1::NAME, 0)
+                .await
+                .unwrap();
```
wlmyng: how come we're `init_watermark`ing before the `assert!` on `set_committer_watermark`?

crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -375,14 +375,11 @@ impl<S: Store> Indexer<S> {
...
             pipeline_task::<S>(P::NAME, self.task.as_ref().map(|t| t.task.as_str()))?;
 
-        let checkpoint_hi_inclusive = conn
+        let next_checkpoint = conn
             .init_watermark(&pipeline_task, self.default_next_checkpoint)
             .await
             .with_context(|| format!("Failed to init watermark for {pipeline_task}"))?;
 
-        let next_checkpoint =
-            checkpoint_hi_inclusive.map_or(self.default_next_checkpoint, |c| c + 1);
```
wlmyng, line 384: yeah i think the framework needs to be responsible for the +1 -1, this shouldn't be something we push down to the store implementations.

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -464,7 +466,8 @@ async fn watermarks_from_pg(
                 cp_sequence_numbers c
             ON (w.reader_lo = c.cp_sequence_number)
             WHERE
-                pipeline = ANY({Array<Text>})
+                w.pipeline = ANY({Array<Text>})
+            AND w.reader_lo <= w.checkpoint_hi_inclusive
```
wlmyng, line 470: and it's not just postgres right? we need to do this filtering for every store?

crates/sui-pg-db/src/store.rs
```diff
@@ -63,31 +55,37 @@ impl store::Connection for Connection<'_> {
...
+                watermarks::reader_lo,
             ))
             .filter(watermarks::pipeline.eq(pipeline_task))
             .first(self)
-            .await
-            .optional()?;
+            .await?;
 
-        if let Some(watermark) = watermark {
+        if reader_lo <= checkpoint_hi_inclusive {
```
wlmyng, line 90: I don't think the sui-pg-db store implementation should need to know any special cases around `reader_lo` and `checkpoint_hi_inclusive`. Watermark-related data is something only the framework needs to understand - I think the check should be done in the framework, whether a watermark row is valid or not.

Otherwise, if the framework defers to the store implementation, an operator might forget to account for this, which means the framework would need to defensively check anyways

crates/sui-indexer-alt-framework/src/mocks/store.rs
evan-wall-mysten: > in sui-analytics-indexer, consistent-store, we don't seem to do the checked_sub, instead accepting and writing default_next_checkpoint as is. This indicates to me that the framework should be responsible for determining what values the store should write, and the store just accepts the value and defines how it actually writes to the underlying db.

`sui-analytics-indexer` and `consistent-store` ignore `default_next_checkpoint` and return the highest indexed checkpoint based on the underlying data - there is no separate watermark table to be checked.

> btw, shouldn't we set checkpoint_hi_inclusive to None if default_next_checkpoint - 1 is < 0?

`default_next_checkpoint.checked_sub(1)` does this.

crates/sui-indexer-alt-framework/src/postgres/mod.rs
evan-wall-mysten: `set_committer_watermark` does an update now instead of an upsert.
 
Only tests were relying on the insert part of the upsert so I made the tests consistent with real behavior by calling `init_watermark` first.

crates/sui-analytics-indexer/src/store/migration.rs
evan-wall-mysten: > I still think it would be great if init_watermark didn't do double duty in 1) trying to initialize the watermark 2) return the committer watermark.

I originally made it do both 1 and 2 so that it is atomic.

> Store operators should just define how to populate the full watermark given the default_next_checkpoint. The indexer framework is responsible for passing in default_next_checkpoint = checkpoint_hi_inclusive = reader_lo - 1, and is also responsible for determining the next checkpoint. This next checkpoint is always checkpoint_hi_inclusive + 1.

I'll change the `init_watermark` inputs to the watermark fields. `checkpoint_hi_inclusive = default_next_checkpoint - 1`, `reader_lo = default_next_checkpoint`.

crates/sui-indexer-alt-e2e-tests/src/lib.rs
evan-wall-mysten: This was to handle converting -1 to u64, but I moved it into the query above.

crates/sui-indexer-alt-framework/src/mocks/store.rs
evan-wall-mysten: I changed it to something in between since I think it was wordy either way.

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
evan-wall-mysten: I changed it to accept and return watermark fields. The -1/+1 logic is now in the framework.

crates/sui-indexer-alt-graphql/src/task/watermark.rs
evan-wall-mysten: The other stores don't have a `reader_lo` AFAIK.

crates/sui-pg-db/src/store.rs
evan-wall-mysten: `reader_lo` is specific to the pg impl so this check can't happen in the framework because other stores don't know about it.

crates/sui-indexer-alt-framework/src/mocks/store.rs
evan-wall-mysten: This no longer applies since I changed `init_watermark` to accept and return `InitWatermark`.

crates/sui-indexer-alt-framework/src/lib.rs
evan-wall-mysten: Moved this into the framework.

crates/sui-indexer-alt-framework/src/mocks/store.rs
```diff
@@ -84,85 +87,102 @@ pub struct MockStore {
...
+        if let Some(watermark) = self.0.watermarks.get(pipeline) {
+            if let Some(checkpoint_hi_inclusive) = watermark.checkpoint_hi_inclusive {
+                Ok(Some((watermark, checkpoint_hi_inclusive)))
+            } else {
+                Ok(None)
+            }
+        } else {
+            Err(anyhow::anyhow!("Pipeline {} not found", pipeline))
+        }
+    }
```
wlmyng: How about

```
  let Some(watermark) = self.0.watermarks.get(pipeline) else {
      return Err(anyhow::anyhow!("Pipeline {} not found", pipeline));
  };

  Ok(watermark
      .checkpoint_hi_inclusive
      .map(|cp| (watermark, cp)))

```

so we can handle the error early, and then it's a matter of returning the existing watermark or None

crates/sui-indexer-alt-framework/src/pipeline/concurrent/commit_watermark.rs
evan-wall-mysten: I have a follow up PR that changes some indexer fields to make testing easier and adds more assertions.

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -116,6 +117,13 @@ pub trait TransactionalStore: Store {
         ) -> ScopedBoxFuture<'a, 'r, anyhow::Result<R>>;
 }
 
+/// Used during watermark initialization to set and return state.
+#[derive(Default, Debug, Clone, Copy, PartialEq)]
+pub struct InitWatermark {
+    pub checkpoint_hi_inclusive: Option<u64>,
```
wlmyng, line 123: would be good to have a comment about how to think about `checkpoint_hi_inclusive` here ... the framework always calculates `checkpoint_hi_inclusive` as the `default_next_checkpoint - 1`, and sets the `reader_lo` to `default_next_checkpoint`.

Pipelines with watermark entries will ignore this entirely, so this watermark is used to initialize new pipelines without watermark entries

crates/sui-analytics-indexer/src/store/mod.rs
```diff
@@ -628,27 +628,27 @@ impl Connection for AnalyticsConnection<'_> {
...
                 // Live mode: derive from file names
-                Ok(self
+                let checkpoint_hi_inclusive = self
                     .committer_watermark(pipeline_task)
                     .await?
-                    .map(|w| w.checkpoint_hi_inclusive))
+                    .map(|w| w.checkpoint_hi_inclusive);
+                Ok(InitWatermark {
+                    checkpoint_hi_inclusive,
+                    ..init_watermark
```
wlmyng, line 646: I suppose `reader_lo` is not a concern right now on the non-pg store implementations because none of the indexers have prunable pipelines, but we'd need to account for that if we did want to add a prunable pipeline

crates/sui-pg-db/src/store.rs
wlmyng: No I don't think that's correct, `reader_lo` is defined on the framework's `Connection` trait:
                                                                                                                                                                                                                                
  - set_reader_watermark(pipeline, reader_lo) — framework calls this                                                                                                                                                            
  - reader_watermark() returns ReaderWatermark { checkpoint_hi_inclusive, reader_lo }                                                                                                                                           
  - pruner_watermark() returns PrunerWatermark { pruner_hi, reader_lo, .. }                                                                                                                                                     
                                                                                                                                                                                                                              
It's a framework concept that any store must support if it wants pruning (specifically to concurrent pipelines). BigTable and ObjectStore currently bail on set_reader_watermark / set_pruner_watermark, but that's because they haven't implemented pruning, not  because `reader_lo` is Postgres-specific.

It also is used by readers of the store to understand what the available data range is for a particular pipeline. That's trivial for non-pg impls today since none of them can prune, but would be important if pruning is possible

crates/sui-analytics-indexer/src/store/mod.rs
wlmyng: Actually ... if we for some reason init at a non-genesis checkpoint, shouldn't we also need to account for the `reader_lo`?

We'll need to set `reader_lo`, if we don't, then we'll erroneously report a data range of `[0, checkpoint_hi_inclusive]` even though we only started indexing from `checkpoint_hi_inclusive`

crates/sui-indexer-alt-framework/src/mocks/store.rs
evan-wall-mysten: Yeah this is more clear. I also updated the call sites to follow this pattern.

crates/sui-analytics-indexer/src/store/mod.rs
evan-wall-mysten: > Actually ... if we for some reason init at a non-genesis checkpoint, shouldn't we also need to account for the reader_lo?
>
> We'll need to set reader_lo, if we don't, then we'll erroneously report a data range of [0, checkpoint_hi_inclusive] even though we only started indexing from checkpoint_hi_inclusive

The `..init_watermark` line handles this by passing back what was passed in (`checkpoint_hi_inclusive + 1`).

I will add a util function for this code in a follow up because the pattern appears so often.

evan-wall-mysten:
> i think the reader_lo as of now feels a bit redundant because we know reader_lo = next_checkpoint and so checkpoint_hi_inclusive = next_checkpoint or reader_lo - 1

It is redundant in that one can be derived from the other, but the framework is doing the deriving and passing in the derived result in order to avoid having the impls do -1/+1.

crates/sui-indexer-alt-framework/src/mocks/store.rs
```diff
@@ -84,63 +87,74 @@ pub struct MockStore {
 #[derive(Clone)]
 pub struct MockConnection<'c>(pub &'c MockStore);
 
+impl MockConnection<'_> {
+    async fn get_watermark(
+        &self,
+        pipeline: &str,
+    ) -> anyhow::Result<Option<(Ref<String, MockWatermark>, u64)>> {
+        let Some(watermark) = self.0.watermarks.get(pipeline) else {
+            return Err(anyhow::anyhow!("Pipeline {} not found", pipeline));
```
amnn, line 96: nit: use `bail!`


```suggestion
            bail!("Pipeline {pipeline} not found");
```

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -116,6 +117,13 @@ pub trait TransactionalStore: Store {
         ) -> ScopedBoxFuture<'a, 'r, anyhow::Result<R>>;
 }
 
+/// Used during watermark initialization to set and return state.
+#[derive(Default, Debug, Clone, Copy, PartialEq)]
```
amnn, line 121: Should this be deriving `Default`? Is `0` always the correct default for `reader_lo`?

crates/sui-pg-db/src/store.rs
```diff
@@ -28,66 +26,73 @@ impl store::Connection for Connection<'_> {
...
+            reader_lo,
+        }: store::InitWatermark,
+    ) -> anyhow::Result<store::InitWatermark> {
         let stored_watermark = StoredWatermark {
             pipeline: pipeline_task.to_string(),
             epoch_hi_inclusive: 0,
-            checkpoint_hi_inclusive: checkpoint_hi_inclusive as i64,
+            // Initially, `checkpoint_hi_inclusive` is less than `reader_lo` meaning that no
+            // checkpoints have been indexed yet.
+            checkpoint_hi_inclusive: checkpoint_hi_inclusive.map_or(-1, |c| c as i64),
```
amnn, line 39: This should not be `-1`, it should be `reader_lo as i64 - 1`, otherwise when you start indexing backwards, you cannot just update/decrement `reader_lo`/`pruner_hi`, you would also need to update `checkpoint_hi_inclusive`.

crates/sui-pg-db/src/store.rs
amnn: It's true that `reader_lo` is a framework concept, not a Postgres concept, but at the same time, the way in which the watermarks are encoded in the store is a Postgres-specific concept (or specific to whichever store we're talking about).

crates/sui-pg-db/src/store.rs
amnn, line 99: nit

```suggestion
        Ok((reader_lo <= checkpoint_hi_inclusive).then_some(store::CommitterWatermark {
            epoch_hi_inclusive: epoch_hi_inclusive as u64,
            checkpoint_hi_inclusive: checkpoint_hi_inclusive as u64,
            tx_hi: tx_hi as u64,
            timestamp_ms_hi_inclusive: timestamp_hi_inclusive as u64,
        }))
```

crates/sui-pg-db/src/store.rs
```diff
@@ -98,17 +103,16 @@ impl store::Connection for Connection<'_> {
...
+        let (checkpoint_hi_inclusive, reader_lo): (i64, i64) = watermarks::table
             .select((watermarks::checkpoint_hi_inclusive, watermarks::reader_lo))
             .filter(watermarks::pipeline.eq(pipeline))
             .first(self)
-            .await
-            .optional()?;
+            .await?;
 
-        if let Some(watermark) = watermark {
+        if reader_lo <= checkpoint_hi_inclusive {
```
amnn, line 112: nit: Ditto, on use of `then_some`

crates/sui-pg-db/src/store.rs
amnn, line 122: Why should `pruner_watermark` care about whether `checkpoint_hi_inclusive` is set? This seems to introduce an undesirable/unnecessary coupling.

crates/sui-indexer-alt-consistent-store/src/store/mod.rs
amnn, line 1: I'm not sure I feel great about this pattern where stores that don't care about `reader_lo` pass the `reader_lo` that is given as input to `init_watermark` back to the caller. This gives a false impression that the store has actually heeded that `reader_watermark` entry, which does not seem correct.

It seems like the return value from the store should reflect what the store actually did. I notice that the framework uses this return value to seed the next checkpoint. I can't quite put my finger on it, but the set-up as a whole seems not right/too complex.

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
evan-wall-mysten: I copied these from the other `*Watermark` models for consistency, it looks like `*Watermark::Default()` is mainly called from tests (including this for `InitWatermark`).

crates/sui-pg-db/src/store.rs
evan-wall-mysten: These are equivalent because `checkpoint_hi_inclusive = None` iff `reader_lo = 0` at this point:
```
            .init_watermark(
                &pipeline_task,
                InitWatermark {
                    checkpoint_hi_inclusive: self.default_next_checkpoint.checked_sub(1),
                    reader_lo: self.default_next_checkpoint,
                },
            )
```

The reason I made `checkpoint_hi_inclusive` an `Option<u64>` is because its range is `[-1, u64::MAX - 1]` so  `None = -1`.

crates/sui-indexer-alt-consistent-store/src/store/mod.rs
evan-wall-mysten: Yeah I agree it is weird, but I think it is consistent with the other `Connection` methods in that even the `Connections` that do not support pruning have methods that use `reader_lo`/`pruner`. 

The way I think about it is that `init_watermark` accepts the proposed `checkpoint_hi_inclusive` and `reader_lo`, but has the chance to override what is passed in. In the case of `reader_lo` it just gets passed back as-is if the `Connection` impl doesn't change it.

I think it can be further cleaned up by having different traits for stores that do not use reader_lo/pruner_hi to avoid them needing to ignore these fields.

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
evan-wall-mysten: Adding this in #25814

crates/sui-pg-db/src/store.rs
evan-wall-mysten: I thought it was useful to be consistent with the other `*_watermark` methods, but the check is not necessary in this case so I'll remove it.

crates/sui-pg-db/src/store.rs
amnn: I agree that these are equivalent today, but in the presence of backwards ingestion, I'm not sure that remains true. That's what I was trying to say here:

> when you start indexing backwards, you cannot just update/decrement `reader_lo`/`pruner_hi`, you would also need to update `checkpoint_hi_inclusive`.

If you set this value to `reader_lo - 1` then backwards ingestion only needs to worry about decrementing `reader_lo` as it commits lower checkpoints. If you set it to `-1` this is not the case.

PR #25776, opened 2026-03-10
indexer: log and track processor retries

amnn:
## Description

Track the number of processor retries for each pipeline in the indexer, and log a warning when the processor fails. We recently faced an issue where the `kv_feature_flags` and `kv_protocol_configs` tables stalled because of a processor failure and this logging would have helped us figure out what was happening much faster.

## Test plan

New unit tests for the framework:

```
$ cargo nextest run -p sui-indexer-alt-framework
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/metrics.rs
+++ b/crates/sui-indexer-alt-framework/src/metrics.rs
@@ -3,6 +3,7 @@
 
 use std::sync::Arc;
 use std::sync::atomic::AtomicU64;
+use std::time::Duration;
 
 use prometheus::Histogram;
 use prometheus::HistogramVec;
@@ -83,6 +84,7 @@ pub struct IngestionMetrics {
 pub struct IndexerMetrics {
     // Statistics related to individual ingestion pipelines' handlers.
     pub total_handler_checkpoints_received: IntCounterVec,
+    pub total_handler_processor_retries: IntCounterVec,
     pub total_handler_checkpoints_processed: IntCounterVec,
     pub total_handler_rows_created: IntCounterVec,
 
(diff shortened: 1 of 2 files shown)


crates/sui-indexer-alt-framework/src/metrics.rs
```diff
@@ -705,6 +714,24 @@ impl IndexerMetrics {
...
+        &self,
+        checkpoint: u64,
+        error: &anyhow::Error,
+        delay: Duration,
+    ) {
+        warn!(
+            pipeline = P::NAME,
+            checkpoint,
+            retry_delay_ms = delay.as_millis(),
+            "Retrying handler after error: {error}",
```
tpham-mysten: nit: Should we use {error:#} or {error:?} here for more context?

crates/sui-indexer-alt-framework/src/metrics.rs
```diff
@@ -705,6 +714,24 @@ impl IndexerMetrics {
             .unwrap(),
         })
     }
+
+    pub(crate) fn inc_retry<P: Processor>(
```
tpham-mysten: nit: The name `inc_retry` is fairly general but the implementation is specific to handler/processor retries (logs "Retrying handler", bumps `total_handler_checkpoint_retries`). Maybe something like `inc_handler_retry` would make the scope clearer; and less chance of someone mistakenly calling this?

PR #25777, opened 2026-03-10
chore: remove stale TODO for settle_early (already implemented)

alex-mysten:
The TODO comment in `authority.rs` noted that address funds should be settled during execution rather than from the checkpoint builder. This has since been implemented via the `settle_early` workstream and is live on testnet. Removing the stale comment.


alex-mysten:
Cherry-picked into #25656

PR #25778, opened 2026-03-10
graphql: uninitialized watermark is considered unhealthy

amnn:
## Description

GraphQL's health check looks for the sigil that indicates that a watermark is uninitialized and treats that as an error, which pushes the service into an unhealthy state.

## Test plan

New unit test:

```
$ cargo nextest run -p sui-indexer-alt-graphql -- uninitialized_watermark_is_unhealthy
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/health.rs
+++ b/crates/sui-indexer-alt-graphql/src/health.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use anyhow::anyhow;
+use anyhow::ensure;
 use async_graphql::indexmap::IndexMap;
 use axum::Extension;
 use axum::Json;
@@ -31,7 +32,7 @@ pub(crate) struct Params {
 }
 
 /// Response body for the health check endpoint. This is output as JSON.
-#[derive(Serialize)]
+#[derive(Serialize, Default)]
 pub(crate) struct Response {
     #[serde(skip_serializing_if = "Option::is_none")]
     checkpoint_lag_ms: Option<u64>,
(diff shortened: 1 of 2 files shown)


crates/sui-indexer-alt-graphql/src/health.rs
```diff
@@ -52,45 +53,54 @@ pub(crate) async fn check(
...
+        }
+    };
+
     let max_lag = params
         .max_checkpoint_lag_ms
         .unwrap_or(config.max_checkpoint_lag.as_millis() as u64);
 
-    if lag > max_lag {
-        errors.push("Watermark lag is too high".to_owned());
+    if checkpoint_lag_ms > max_lag {
```
evan-wall-mysten, line 74: Can the durations be passed around as `std::time::Duration` instead of `u64` to avoid accidentally mixing different time units?

crates/sui-indexer-alt-graphql/src/health.rs
```diff
@@ -52,45 +53,54 @@ pub(crate) async fn check(
...
 
 /// Returns the lag between the latest checkpoint the indexer is aware of and the current time.
-async fn check_watermarks(watermarks: &WatermarksLock) -> (u64, IndexMap<String, u64>) {
+async fn check_watermarks(
+    watermarks: &WatermarksLock,
+) -> anyhow::Result<(u64, IndexMap<String, u64>)> {
     let now = Utc::now();
     let watermarks = watermarks.read().await;
 
+    ensure!(watermarks.initialized(), "Watermark has not initialized");
```
tpham-mysten: tiny nit: Not sure if it's just me but "Watermark has not been initialized" or "Watermark is not initialized" sounds more natural

crates/sui-indexer-alt-graphql/src/health.rs
amnn: In practice, all this would mean is that on the line above, we would wrap `params.max_checkpoint_lag_ms` in `Duration::from_millis`, and then we would need logic below to unwrap all those things. The `lag_ms` function exists exactly because we're storing durations as `Duration`s and need to convert it into a `u64` for output in this endpoint.

PR #25783, opened 2026-03-10
display: avoid holding hash map lock while fetching objects

amnn:
## Description

The previous implementation of dynamic object loading created an entry in the object loading cache, and then tried to fetch the object. The cache is a concurrent hash-map, which means that we are implicitly taking a lock on that entry while waiting for the object to be fetched, which is not ideal.

This change splits the initial cache fetch, object load, and cache update into three updates without having to hold a lock between all three steps.

## Test plan

New test covers this case:

```
$ cargo nextest run -p sui-display -- concurrent_dynamic_field
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-display/src/v2/interpreter.rs
+++ b/crates/sui-display/src/v2/interpreter.rs
@@ -6,7 +6,6 @@ use std::mem;
 use std::sync::Arc;
 
 use dashmap::DashMap;
-use dashmap::mapref::entry::Entry;
 use futures::future::OptionFuture;
 use futures::future::join_all;
 use futures::join;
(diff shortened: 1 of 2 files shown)


crates/sui-display/src/v2/interpreter.rs
```diff
@@ -395,18 +394,20 @@ impl<S: V::Store> Interpreter<S> {
...
     async fn object<'s>(&'s self, id: AccountAddress) -> Result<Option<V::Slice<'s>>, FormatError> {
-        let entry = match self.cache.entry(id) {
-            Entry::Occupied(entry) => entry,
-            Entry::Vacant(entry) => entry.insert_entry(
-                self.store
-                    .object(id)
-                    .await
-                    .map_err(|e| FormatError::Store(Arc::new(e)))?
-                    .map(Arc::new),
-            ),
```
bmwill, line 406: Theres a hidden lock guard in this? how was it able to be held over an await point? This actually would push me to not use dashmap :/

crates/sui-display/src/v2/interpreter.rs
```diff
@@ -395,18 +394,20 @@ impl<S: V::Store> Interpreter<S> {
...
+        let owned = if let Some(cached) = self.cache.get(&id) {
+            cached.clone()
```
wlmyng, line 398: ah this drops the reference tied to the shared lock? I think the locking makes sense - if we go the `entry` route we want to keep some lock held between check and insert for atomicity, which understandably blocks other tasks trying to access the same data, but if we don't care for that, then we'd want to release the locks asap

PR #25792, opened 2026-03-11
[sui system] stricter metadata validation, cleanup

damirka:
## Description 

- make metadata checks stricter
- code reorganisation and cleanup
- drastically improves testing of metadata

## Test plan 

- massive tests

---

## Release notes

- [x] Protocol: metadata hardening in Sui System
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority/snapshots/sui_core__authority__execution_time_estimator__tests__execution_time_observer_v117.snap
+++ b/crates/sui-core/src/authority/snapshots/sui_core__authority__execution_time_estimator__tests__execution_time_observer_v117.snap
@@ -1,0 +1,156 @@
+---
+source: crates/sui-core/src/authority/execution_time_estimator.rs
+expression: snapshot_data
+---
+protocol_version: 117
+consensus_observations:
+  - - MakeMoveVec
+    - observations:
+        - - 6
+          - secs: 0
+            nanos: 8000000
+        - - 6
+          - secs: 0
+            nanos: 7000000
+        - - 5
+          - secs: 0
+            nanos: 19000000
+        - - 1
+          - secs: 0
+            nanos: 25000000
+      stake_weighted_median:
+        secs: 0
+        nanos: 19000000
+  - - MergeCoins
+    - observations:
+        - - 6
+          - secs: 0
+            nanos: 63000000
+        - - 7
+          - secs: 0
+            nanos: 52000000
+        - - 8
+          - secs: 0
+            nanos: 0
+        - - 7
+          - secs: 0
+            nanos: 0
+      stake_weighted_median:
+        secs: 0
+        nanos: 52000000
+  - - SplitCoins
+    - observations:
+        - - 10
+          - secs: 0
+            nanos: 38000000
+        - - 3
+          - secs: 0
+            nanos: 45000000
+        - - 5
+          - secs: 0
+            nanos: 54000000
+        - - 3
+          - secs: 0
+            nanos: 61000000
+      stake_weighted_median:
+        secs: 0
+        nanos: 54000000
+  - - TransferObjects
+    - observations:
+        - - 10
+          - secs: 0
+            nanos: 0
+        - - 0
+          - ~
+        - - 0
+          - ~
+        - - 4
+          - secs: 0
+            nanos: 40000000
+      stake_weighted_median:
+        secs: 0
+        nanos: 40000000
+  - - Upgrade
+    - observations:
+        - - 0
+          - ~
+        - - 0
+          - ~
+        - - 6
+          - secs: 0
+            nanos: 981000000
+        - - 4
+          - secs: 0
+            nanos: 587000000
+      stake_weighted_median:
+        secs: 0
+        nanos: 981000000
+  - - MoveEntryPoint:
+        package: "0x0000000000000000000000000000000000000000000000000000000000000001"
+        module: coin
+        function: transfer
+        type_arguments: []
+    - observations:
+        - - 7
+          - secs: 0
+            nanos: 404000000
+        - - 6
+          - secs: 0
+            nanos: 162000000
+        - - 9
+          - secs: 0
+            nanos: 268000000
+        - - 0
+          - ~
+      stake_weighted_median:
+        secs: 0
+        nanos: 268000000
+  - - MoveEntryPoint:
+        package: "0x0000000000000000000000000000000000000000000000000000000000000002"
+        module: nft
+        function: mint
+        type_arguments: []
+    - observations:
+        - - 8
+          - secs: 0
+            nanos: 116000000
+        - - 1
+          - secs: 0
+            nanos: 172000000
+        - - 10
+          - secs: 0
+            nanos: 432000000
+        - - 8
+          - secs: 0
+            nanos: 499000000
+      stake_weighted_median:
+        secs: 0
+        nanos: 432000000
+transaction_estimates:
+  - - coin_transfer_call
+    - secs: 0
+      nanos: 268000000
+  - - mixed_move_calls
+    - secs: 0
+      nanos: 700000000
+  - - native_commands_with_observations
+    - secs: 0
+      nanos: 330000000
+  - - transfer_objects_5_items
+    - secs: 0
+      nanos: 240000000
+  - - split_coins_1_amounts
+    - secs: 0
+      nanos: 108000000
+  - - merge_coins_3_sources
+    - secs: 0
+      nanos: 208000000
+  - - make_move_vec_5_elements
+    - secs: 0
+      nanos: 114000000
+  - - mixed_commands
+    - secs: 0
+      nanos: 229000000
+  - - upgrade_package
+    - secs: 0
+      nanos: 981000000
--- a/crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/objects/display/display_v2.snap
+++ b/crates/sui-indexer-alt-e2e-tests/tests/jsonrpc/objects/display/display_v2.snap
@@ -57,9 +57,9 @@ Response: {
   "result": [
     {
       "data": {
-        "objectId": "0x406a6a87303b52b32ee7168e60401f0b8db85750e4e29fda65a77d556c051a82",
+        "objectId": "0x4f4f8a1ebebaa50008c85f2d4aadc5fc9473a9b9b4c83a701670482ea76f22b2",
         "version": "4",
-        "digest": "C2GfzaAYdUuqZs7QWS2aYjsiThc3KR3L8Q4XwaRcKitd",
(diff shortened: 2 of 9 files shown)


crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
```diff
@@ -429,11 +429,10 @@ public(package) fun request_set_gas_price(
 ) {
     // Verify the represented address is an active or pending validator, and the capability is still valid.
     let verified_cap = self.validators.verify_cap(cap, ACTIVE_OR_PENDING_VALIDATOR);
-    let validator = self
+    self
         .validators
-        .get_validator_mut_with_verified_cap(&verified_cap, false /* include_candidate */);
```
manolisliolios, line 434: The old one was not including candidates here but I think now we do?

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
```diff
@@ -593,214 +590,265 @@ fun undo_report_validator_impl(
     }
 }
 
-// ==== validator metadata management functions ====
+// ==== Candidate Metadata Management ====
```
damirka: These guys are not new, I changed the order of definitions. So it's not the best review experience, but in the source it makes life 100x better.

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
damirka: This is not very explicit, but it will fail on `verify_cap` call if validator is not active or pending.

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
```diff
@@ -1179,16 +1225,16 @@ public(package) fun pending_validator_by_address(
...
 
 #[test_only]
 /// Return the currently candidate validator by address
 public(package) fun candidate_validator_by_address(
-    self: &SuiSystemStateInnerV2,
+    self: &mut SuiSystemStateInnerV2,
     validator_address: address,
 ): &Validator {
-    self.validators().get_candidate_validator_ref(validator_address)
+    self.validators_mut().candidate_validator(validator_address)
```
manolisliolios, line 1240: Why is mut needed here?

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -35,12 +35,13 @@ const EMinJoiningStakeNotReached: u64 = 5;
 const EAlreadyValidatorCandidate: u64 = 6;
 const EValidatorNotCandidate: u64 = 7;
 const ENotValidatorCandidate: u64 = 8;
-const ENotActiveOrPendingValidator: u64 = 9;
+// const ENotActiveOrPendingValidator: u64 = 9;
```
manolisliolios, line 38: remove?

damirka: Nit, but kept for consistency in indexes.

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
damirka: ValidatorWrapper is "load_or_upgrade". Unavoidable `&mut` for candidates.

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
```diff
@@ -593,212 +590,261 @@ fun undo_report_validator_impl(
...
 public(package) fun update_validator_name(
     self: &mut SuiSystemStateInnerV2,
     name: vector<u8>,
     ctx: &TxContext,
 ) {
-    let validator = self.validators.get_validator_mut_with_ctx_including_candidates(ctx);
+    let validator = self.validators.any_validator_mut(ctx.sender());
     validator.update_name(name);
+    let validator: &Validator = validator; // Avoid parallel mutable borrow.
+    self.validators.assert_no_pending_or_active_duplicates(validator);
```
lxfind: What happens here if the validator is a candidate?

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -810,85 +797,99 @@ fun get_validator_indices(
...
-    self.get_active_or_pending_or_candidate_validator_mut(validator_address, false)
+    let idx = find_validator_from_table_vec(
+        &self.pending_active_validators,
+        validator,
+    ).destroy_or!(abort ENotAPendingValidator);
+    &mut self.pending_active_validators[idx]
 }
 
-public(package) fun get_validator_mut_with_ctx_including_candidates(
+/// Get mutable reference to a pending validator by address.
```
lxfind: candidate validator?

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
damirka: This validator's data is compared against all active and pending validators. Hence, aborts if the name is duplicate with an existing validator.

crates/sui-framework/packages/sui-system/sources/validator_set.move
damirka: Yup! Fixed.

crates/sui-framework/packages/sui-system/sources/validator.move
```diff
@@ -808,6 +819,7 @@ public(package) fun update_candidate_protocol_pubkey(
     proof_of_possession: vector<u8>,
 ) {
     assert!(self.is_preactive(), ENotValidatorCandidate);
+    assert!(protocol_pubkey.length() == 96, EInvalidProtocolPubKeyLength);
```
jonas-lj, line 822: Perhaps add a comment why this check is needed for future reference? Something like "To ensure uniqueness of the binary representation of a public key, we enforce compressed form of the public key (BLS12-381-G2 element since we're using minsig)".

crates/sui-framework/packages/sui-system/tests/builders/test_runner.move
```diff
@@ -68,8 +68,16 @@ public fun build(builder: TestRunnerBuilder): TestRunner {
     } = builder;
 
     let validators = validators.destroy_or!({
-        vector::tabulate!(validators_count.destroy_or!(4), |_| {
-            validator_builder::new().initial_stake(validators_initial_stake.destroy_or!(100))
+        let validators_count = validators_count.destroy_or!(4);
+        let initial_stake = validators_initial_stake.destroy_or!(100);
+        vector::tabulate!(validators_count, |idx| {
+            if (idx < 4) {
```
emmazzz, line 74: To make sure I understand, we are doing this magic-looking `idx < 4` check because we have preset info for 4 validators. These are 4 sets of validator metadata generated manually to test things like `proof_of_possession`. We are making things cleaner in this PR by moving the manually generated metadata into the `validator_preset` file.

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
emmazzz: To be very explicit, you meant the transaction will fail on the previous line (line 431) where we check the cap belongs to an active or pending validator right? So a candidate won't even get to line 432.

crates/sui-framework/packages/sui-system/tests/builders/test_runner.move
emmazzz, line 74: To confirm my understanding: in `validator_preset` we have included four manually (auto) generated sets of validator metadata for convenience of usage inside tests. Previously we just had one preset defined at the top of `validator_builder.move`. We need four sets of metadata now in order to add more thorough and meaningful metadata tests. Is this understanding correct?

crates/sui-framework/packages/sui-system/tests/delegation_tests.move
```diff
@@ -330,7 +330,9 @@ fun add_stake_post_active_flow() {
 // 4. Unstake from the preactive validator. There should be no rewards earned.
 fun add_preactive_remove_preactive() {
     let mut runner = test_runner::new().validators_initial_stake(100).validators_count(2).build();
-    let validator = validator_builder::preset().sui_address(NEW_VALIDATOR_ADDR).build(runner.ctx());
+    let validator = validator_builder::preset(2)
```
emmazzz, line 333: We are just using `preset(2)` arbitrarily right? It doesn't matter which set of metadata we use here. Same for the changes below as well.

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
```diff
@@ -593,212 +590,261 @@ fun undo_report_validator_impl(
...
+        .validators
+        .any_validator_mut(ctx.sender())
+        .new_unverified_validator_operation_cap_and_transfer(ctx)
 }
 
 /// Update a validator's name.
+///
+/// Unlike `description` and `image_url`, `name` is checked against both active
+/// and pending validators and must be unique in the system, even for validator
+/// candidates.
```
emmazzz: This comment could be written more clearly. Maybe say:
"When any validator, including candidate, pending and active validators, updates their name, check the new name against the names of existing pending and active validators."
The important thing to note here is that we DO NOT check against existing candidate names. Otherwise a candidate can DOS a potential future validator by taking their name first.

crates/sui-framework/packages/sui-system/tests/builders/test_runner.move
damirka: Yeah, that's right. There are 4 presets at the moment.

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
damirka: Just updated it!

crates/sui-framework/packages/sui-system/tests/delegation_tests.move
damirka: More or less, yes, any preset will do in this case.

crates/sui-framework/packages/sui-system/tests/builders/test_runner.move
damirka: Absolutely!

crates/sui-framework/packages/sui-system/sources/sui_system_state_inner.move
damirka: @emmazzz precisely, see the selector on the `verify_cap` method, it will abort if not found.

PR #25800, opened 2026-03-11
[unit-test] Support custom gas price for unit tests and fix statistics gas estimation

domenicodev:
## Description 

Add support for a custom **gas price** in unit tests by using `--gas-price` flag, and fix gas estimations in statistics.

## Test plan 

Currently, by debugging move tests after compiling the binary.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-move/src/unit_test.rs
+++ b/crates/sui-move/src/unit_test.rs
@@ -97,12 +97,9 @@ impl Test {
         // }
 
         // cap `gas_limit` to the effective max gas budget.
-        if unit_test_config
+        unit_test_config.gas_limit = unit_test_config
             .gas_limit
-            .is_some_and(|gas_limit| gas_limit > max_computation_budget)
-        {
-            unit_test_config.gas_limit = Some(max_computation_budget);
-        }
+            .map(|lim| lim.min(max_computation_budget));
 
         // set the environment (this is a little janky: we get it from the manifest here, then pass
         // it as the optional argument in the build-config, which then looks it up again, but it
--- a/crates/sui-move/src/unit_test.rs
+++ b/crates/sui-move/src/unit_test.rs
@@ -6,7 +6,6 @@ use move_cli::base::{
     self,
     test::{self, UnitTestResult},
 };
-use move_core_types::gas_algebra::InternalGas;
 use move_package_alt_compilation::build_config::BuildConfig;
 use move_unit_test::{UnitTestingConfig, vm_test_setup::VMTestSetup};
 use move_vm_config::runtime::VMConfig;
--- a/crates/sui-move/src/unit_test.rs
+++ b/crates/sui-move/src/unit_test.rs
@@ -37,12 +37,13 @@ use sui_types::{
     metrics::LimitsMetrics,
 };
 
-// The maximum gas budget(limit) used for tests, in MIST. Execution bound.
-pub static MAX_GAS_BUDGET_TESTS: LazyLock<u64> =
-    LazyLock::new(|| ProtocolConfig::get_for_max_version_UNSAFE().max_tx_gas());
+/// The actual max. computation units for unit tests.
+/// Max. gas limit/budget in tests will effectively be calculated as: `BUCKET * GAS_PRICE`.
+pub static MAX_GAS_COMPUTATION_BUCKET: LazyLock<u64> =
+    LazyLock::new(|| ProtocolConfig::get_for_max_version_UNSAFE().max_gas_computation_bucket());
 
 /// Gas price used for the meter during Move unit tests.
-const TEST_GAS_PRICE: u64 = 500;
+pub const TEST_GAS_PRICE: u64 = 500;
 
 #[derive(Parser)]
 #[group(id = "sui-move-test")]
@@ -80,17 +81,25 @@ impl Test {
         let rerooted_path = base::reroot_path(path)?;
         let mut unit_test_config = self.test.unit_test_config();
 
+        // Use custom gas price if set, otherwise use default.
+        let gas_price = self.gas_price.unwrap_or(TEST_GAS_PRICE);
+
+        // the gas meter silently caps the max. gas budget to `max_gas_computation_bucket * gas_price`, we reflect it here.
+        // if we pass a higher value the budget in used_gas will produce inconsistent results.
+        let max_computation_budget = *MAX_GAS_COMPUTATION_BUCKET * gas_price;
+
         // TODO: 
-        // MAX_GAS_BUDGET_TESTS as a default value is always bypassed, at this point gas_limit would be none
-        // it will be overridden in move_unit_tests::run_and_report_unit_tests()->test_runner with DEFAULT_EXECUTION_BOUND
+        // (MAX_GAS_COMPUTATION_BUCKET * effective_gas_price) as a default value is always bypassed in this module's scope.
+        // at this point `gas_limit` would be none. it will be overridden in:
+        // move_unit_tests::run_and_report_unit_tests()->test_runner with DEFAULT_EXECUTION_BOUND
+        // to enable it here, use:
         // if unit_test_config.gas_limit.is_none() {
-        //     unit_test_config.gas_limit = Some(*MAX_GAS_BUDGET_TESTS);
+        //     unit_test_config.gas_limit = Some(max_computation_budget);
         // }
 
-        // cap the gas limit to MAX_GAS_BUDGET_TESTS
-        let max_gas_budget = *MAX_GAS_BUDGET_TESTS;
-        if unit_test_config.gas_limit.is_some_and(|gas_limit| gas_limit > max_gas_budget) {
-            unit_test_config.gas_limit = Some(max_gas_budget);
+        // cap `gas_limit` to the effective max gas budget.
+        if unit_test_config.gas_limit.is_some_and(|gas_limit| gas_limit > max_computation_budget) {
+            unit_test_config.gas_limit = Some(max_computation_budget);
         }
 
         // set the environment (this is a little janky: we get it from the manifest here, then pass
@@ -122,8 +131,9 @@ pub async fn run_move_unit_tests(
     save_disassembly: bool,
     gas_price: Option<u64>,
 ) -> anyhow::Result<UnitTestResult> {
+    let effective_gas_price = gas_price.unwrap_or(TEST_GAS_PRICE);
     let config = config.unwrap_or_else(|| {
-        UnitTestingConfig::default_with_bound(Some(*MAX_GAS_BUDGET_TESTS))
+        UnitTestingConfig::default_with_bound(Some(*MAX_GAS_COMPUTATION_BUCKET * effective_gas_price))
     });
 
     let result = move_cli::base::test::run_move_unit_tests::<sui_package_alt::SuiFlavor, _, _>(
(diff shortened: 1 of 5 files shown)


crates/sui-move/src/unit_test.rs
```diff
@@ -36,8 +37,8 @@ use sui_types::{
     metrics::LimitsMetrics,
 };
 
-// Move unit tests will halt after executing this many steps. This is a protection to avoid divergence
-pub static MAX_UNIT_TEST_INSTRUCTIONS: LazyLock<u64> =
+// The maximum gas budget(limit) used for tests, in MIST. Execution bound.
+pub static DEFAULT_GAS_BUDGET_TESTS: LazyLock<u64> =
```
domenicodev: Renamed for better understanding/scope

crates/sui-move/src/unit_test.rs
```diff
@@ -75,6 +77,14 @@ impl Test {
         let rerooted_path = base::reroot_path(path)?;
         let unit_test_config = self.test.unit_test_config();
 
+        // TODO:
```
domenicodev: Enabling this would actually make the default(max) gas budget work as intended.

domenicodev:
Note:
With this new implementation, **gas price** on `statistics` could be avoided at all.
Gas Used now reports the total `Gas Units (internal)`.

Would make sense if an additional field in the output table will be added, e.g.: `Total Gas Cost (MIST)`

crates/sui-move/src/unit_test.rs
tnowacki: I wouldn't rename this, since I'm not sure if we should allow the user defined limit to go over this. We have discussed having a gasless setting for tests, and I think that would be used in such a case

external-crates/move/crates/move-cli/src/base/test.rs
```diff
@@ -79,6 +79,10 @@ pub struct Test {
     /// Enable tracing for tests.
     #[clap(long = "trace")]
     pub trace: bool,
+
+    /// Set custom gas price for tests (in gas units).
+    #[clap(name = "custom-gas-price", long = "gas-price")]
```
tnowacki:
```suggestion
    #[clap(name = "gas-price", long = "gas-price")]
```
With some comment that it defaults to the reference gas price

tnowacki: All flags are inherently custom ;)

external-crates/move/crates/move-unit-test/src/lib.rs
```diff
@@ -125,6 +126,10 @@ pub struct UnitTestingConfig {
     // Enable tracing for tests
     #[clap(long = TRACE_FLAG)]
     pub trace: bool,
+
+    /// Set custom gas price for tests (in gas units).
+    #[clap(long = CUSTOM_GAS_PRICE_FLAG)]
+    pub custom_gas_price: Option<u64>,
```
tnowacki: I don't think this is a concept in the external crates side of things?

external-crates/move/crates/move-cli/src/base/test.rs
tnowacki: Also, this should be on the sui side of things? There isn't a notion of "price" in external-crates

external-crates/move/crates/move-unit-test/src/lib.rs
```diff
@@ -272,6 +278,11 @@ impl UnitTestingConfig {
         } else {
             None
         };
+
+        if let Some(gas_price) = self.custom_gas_price {
+            writeln!(shared_writer.lock().unwrap(), "Using Custom Gas Price: {}", gas_price)?;
+        }
+
```
tnowacki: Is this necessary?

external-crates/move/crates/move-unit-test/src/test_runner.rs
```diff
@@ -549,6 +549,7 @@ impl<V: VMTestSetup> SharedTestingConfig<V> {
                         )
                     }
                     None if err.major_status() == StatusCode::OUT_OF_GAS => {
+                        // TODO should we just implementa new error: OUT_OF_GAS, suggesting gas limit can be increased?
```
tnowacki, line 552: Yeah this error does suck--I hit it recently it it really gives no help

tnowacki (commented):
I am trying to give some quick feedback today, and otherwise won't be able to look at this again until later next week. But I imagine @tzakian and @dariorussi will have some ideas

crates/sui-move/src/unit_test.rs
domenicodev: The thing is, currently, it's indeed used both as a gas budget, and as execution bound; the actual definition would be default_execution_bound to match external crate's def, but not sure. Comment was also misleading before.
Open to ideas here

external-crates/move/crates/move-cli/src/base/test.rs
domenicodev: CLI args for tests are entirely managed in this file so I did it directly here to not add other duplications but yep, if it's an issue I can make it work on Sui crates

crates/sui-move/src/unit_test.rs
```diff
@@ -36,18 +37,23 @@ use sui_types::{
     metrics::LimitsMetrics,
 };
 
-// Move unit tests will halt after executing this many steps. This is a protection to avoid divergence
-pub static MAX_UNIT_TEST_INSTRUCTIONS: LazyLock<u64> =
-    LazyLock::new(|| ProtocolConfig::get_for_max_version_UNSAFE().max_tx_gas());
+/// The actual max. computation units for unit tests.
+/// Max. gas limit/budget in tests will effectively be calculated as: `BUCKET * GAS_PRICE`.
+pub static MAX_GAS_COMPUTATION_BUCKET: LazyLock<u64> =
```
domenicodev, line 41: Should be the intended naming and behavior (actual execution bound is this * gas_price); open to suggestions

crates/sui-move/src/unit_test.rs
```diff
@@ -73,7 +79,28 @@ impl Test {
...
+        // TODO: 
+        // (MAX_GAS_COMPUTATION_BUCKET * effective_gas_price) as a default value is always bypassed in this module's scope.
+        // at this point `gas_limit` would be none. it will be overridden in:
+        // move_unit_tests::run_and_report_unit_tests()->test_runner with DEFAULT_EXECUTION_BOUND
+        // to enable it here, use:
+        // if unit_test_config.gas_limit.is_none() {
+        //     unit_test_config.gas_limit = Some(max_computation_budget);
+        // }
+
+        // cap `gas_limit` to the effective max gas budget.
```
domenicodev, line 99: Added the Cap we were mentioning earlier (max. gas budget for tests)

domenicodev (commented):
To support MIST conversions along with custom outputs, we should rewrite(clone and customize) the test external crate into the framework; for the current purpose, this PR is enough.

crates/sui-move/src/unit_test.rs
```diff
@@ -178,9 +213,15 @@ impl VMTestSetup for SuiVMTestSetup {
     }
 
     fn used_gas<'a>(&'a self, execution_bound: u64, meter: Self::Meter<'a>) -> u64 {
```
cgswords, line 215: How about a doc comment explaining what these input units are, or better yet a type alias that makes it clear?

domenicodev: Improved, please check

crates/sui-move/src/unit_test.rs
```diff
@@ -212,15 +212,16 @@ impl VMTestSetup for SuiVMTestSetup {
...
 
+    /// Returns gas consumed so far in the VM, in internal units, by computing `initial gas_budget - gas_left`
+    ///
+    /// `execution_bound` is the total gas budget passed by the test runner, in external units and not scaled by gas price.
+    /// It is the same as the initial gas budget of the meter.
     fn used_gas<'a>(&'a self, execution_bound: u64, meter: Self::Meter<'a>) -> u64 {
-        // execution_bound (gas budget) is in external units, create a Gas instance from it to convert to internal units.
-        let budget_external_units = Gas::new(execution_bound / self.gas_price);
+        // `gas_left` is in internal units and is scaled by the gas price. Normalize the gas budget to match its scale
+        let gas_budget_normalized = Gas::new(execution_bound / self.gas_price).to_unit();
```
domenicodev, line 217: May lose precision here but it's intended, it matches the the mainnet math the closest possible.
Fix would be multiply-then-divide.

ref: `sui-types/src/gas_model/tables.rs:77-78`
```
let budget_in_unit = budget / gas_price;
let gas_left = Self::to_internal_units(budget_in_unit);
```

Not sure if there's any considerations to add here or I am missing something.

crates/sui-move/src/unit_test.rs
```diff
@@ -73,7 +78,31 @@ impl Test {
...
+        //     unit_test_config.gas_limit = Some(max_computation_budget);
+        // }
+
+        // cap `gas_limit` to the effective max gas budget.
+        if unit_test_config
+            .gas_limit
+            .is_some_and(|gas_limit| gas_limit > max_computation_budget)
+        {
+            unit_test_config.gas_limit = Some(max_computation_budget);
+        }
```
tzakian: Nit: may be a smidge cleaner as
```rust
unit_test_config.gas_limit = unit_test_config.gas_limit.map(|lim| lim.min(max_computation_budget));
```

crates/sui-move/src/unit_test.rs
tzakian: This makes sense to me. But good question.

external-crates/move/crates/move-unit-test/src/test_runner.rs
tzakian: We can definitely have the discussion on exposing this in a better way :) 

When I was initially writing the unit testing framework the idea was to hide away the concept of gas as much as possible as the idea was (1) unit test are for behavior only, and adding gas to the mix adds fragility to them; and (2) places where amount of gas matters is a place for integration tests/transactional tests.

But if we're now exposing more of the gas logic in tests (in terms of setting gas price, not using a tick-only meter, etc.) then exposing this a bit more in our error messages doesn't seem to be too bad either.

tzakian:
FYI: you probably want to base this on `main` and not `bella-ciao` :)

domenicodev:
> FYI: you probably want to base this on `main` and not `bella-ciao` :)

I just tried, and figured out that unfortunately I started from bella-ciao as a base branch(as I thought it was our current kind of "development" branch), so It'd need a rebase. 

I can go and rebase from main if needed, let me know if you'd like me to

PS: apologies to the new "review requests" people, it's been triggered automatically by CI/CD when I changed base branch because it included bella-ciao commits affecting multiple stacks; you can skip this

external-crates/move/crates/move-unit-test/src/test_runner.rs
domenicodev: It's not very necessary in my opinion, but it stings a bit when you arbitrarily set gas price and/or gas limit, and then issue a "timeout" out of nowhere 😆  though it's easy to understand it would be caused by that.

It'd be a very small fix but it touches the external crate, not sure if in the end we want to make any changes to it at the moment. Happy to hear your opinion

PR #25814, opened 2026-03-12
sui-indexer-alt-framework: add init_with_committer_watermark util function

evan-wall-mysten:
## Description 

The `Connection` impls that do not support pruning have the same `init_watermark` impl. This PR adds a util function for them to call to ensure they stay in sync.

Long term I think the `Connection` trait can be split into a version
that supports trailing edge operations and version that does not support them which may clean
this up further.

## Test plan 

Existing unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework-store-traits/src/lib.rs
+++ b/crates/sui-indexer-alt-framework-store-traits/src/lib.rs
@@ -119,7 +119,9 @@ pub trait TransactionalStore: Store {
 /// Used during watermark initialization to set and return state.
 #[derive(Default, Debug, Clone, Copy, PartialEq)]
 pub struct InitWatermark {
+    /// Calculated by the framework as `default_next_checkpoint.checked_sub(1)`.
     pub checkpoint_hi_inclusive: Option<u64>,
+    /// Calculated by the framework as `default_next_checkpoint`.
     pub reader_lo: u64,
 }
 
--- a/crates/sui-pg-db/src/store.rs
+++ b/crates/sui-pg-db/src/store.rs
@@ -87,16 +87,14 @@ impl store::Connection for Connection<'_> {
             .first(self)
             .await?;
 
-        if reader_lo <= checkpoint_hi_inclusive {
-            Ok(Some(store::CommitterWatermark {
+        Ok(
+            (reader_lo <= checkpoint_hi_inclusive).then_some(store::CommitterWatermark {
                 epoch_hi_inclusive: epoch_hi_inclusive as u64,
                 checkpoint_hi_inclusive: checkpoint_hi_inclusive as u64,
                 tx_hi: tx_hi as u64,
                 timestamp_ms_hi_inclusive: timestamp_ms_hi_inclusive as u64,
-            }))
-        } else {
-            Ok(None)
-        }
+            }),
+        )
     }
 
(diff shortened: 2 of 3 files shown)


crates/sui-analytics-indexer/src/store/mod.rs
```diff
@@ -634,23 +635,7 @@ impl Connection for AnalyticsConnection<'_> {
...
-                    checkpoint_hi_inclusive,
-                    ..init_watermark
-                })
-            }
-            StoreMode::Migration(store) => {
-                let output_prefix = self.pipeline_config(pipeline_task).output_prefix();
-                store.init_watermark(output_prefix, init_watermark).await
-            }
-        }
+        init_with_committer_watermark(self, pipeline_task, init_watermark).await
```
evan-wall-mysten, line 638: The original behavior always ended up calling `store.committer_watermark(pipeline_task)`, but arrived there through different paths:
```
init_watermark + Live -> call self.committer_watermark which internally calls output_prefix() then store.committer_watermark
    self.committer_watermark + Live -> call output_prefix() then store.committer_watermark
    self.committer_watermark + Stored (not applicable) -> call output_prefix() then store.committer_watermark
init_watermark + Stored -> call output_prefix() then store.committer_watermark
```

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -87,6 +86,23 @@ pub trait Connection: Send {
     ) -> anyhow::Result<bool>;
 }
 
+/// A utility function for connections that do not have special initialization logic. These
```
amnn: nit: file order (free functions after traits, types, impl blocks).

Also, see my comment on the earlier PR about whether it is correct for a store that does not honour `reader_lo` returning the `reader_lo` that was sent to it, back to the caller.

evan-wall-mysten: I will look at a more comprehensive way of eliminating `reader_lo` completely from the API of `Connection` impls that do not care about it in a follow up PR.

crates/sui-pg-db/src/store.rs
```diff
@@ -134,27 +130,17 @@ impl store::Connection for Connection<'_> {
...
-                .select((
-                    wait_for,
-                    watermarks::pruner_hi,
-                    watermarks::reader_lo,
-                    watermarks::checkpoint_hi_inclusive,
-                ))
-                .filter(watermarks::pipeline.eq(pipeline))
-                .first(self)
-                .await?;
+        let (wait_for_ms, pruner_hi, reader_lo): (i64, i64, i64) = watermarks::table
```
tpham-mysten, line 133: Is this related to the PR?

evan-wall-mysten: This PR is a follow up to to https://github.com/MystenLabs/sui/pull/25774 which was already merged. I got comments on it after I merged so I addressed them in this PR.

https://github.com/MystenLabs/sui/pull/25814#discussion_r2933011203

PR #25824, opened 2026-03-13
refactor(indexer-alt): use query! macro in prune handlers

dongowu:
## Description 

Replace the remaining `format!`-based prune queries in `sui-indexer-alt`
  with `sui_sql_macro::query!` in `obj_info` and `coin_balance_buckets`.

  This keeps the pruning logic unchanged while switching checkpoint range
  parameters to typed SQL binds and removing raw SQL string interpolation.

  Also add the crate dependencies required by the macro expansion in
  `sui-indexer-alt` (`sui-sql-macro` and `sui-pg-db`).

## Test plan 

- `rustfmt --check crates/sui-indexer-alt/src/handlers/coin_balance_buckets.rs crates/sui-indexer-alt/src/handlers/obj_info.rs`
- `cargo check -p sui-indexer-alt`

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt/Cargo.toml
```diff
@@ -41,7 +41,9 @@ sui-default-config.workspace = true
 sui-indexer-alt-framework = { workspace = true, features = ["postgres"] }
 sui-indexer-alt-metrics.workspace = true
 sui-indexer-alt-schema.workspace = true
+sui-pg-db.workspace = true
```
amnn, line 44: Is this dependency necessary? It should be pulled in with the framework, and I don't see a reference to this crate directly in the changed code.

wlmyng: I think we need one or the other for the macro, but not both. To get the `query!` macro to work for `obj_versions` pruning I pulled in `sui-pg-db` - seems like we have to manually import either crate because the path is hardcoded in the macro itself.

dongowu: This is still needed. `sui-indexer-alt` uses `sui_sql_macro::query!`, which expands to `::sui_pg_db::...`. In Rust, that path is resolved in the consuming
  crate, so `sui-pg-db` must be a direct dependency here; the framework’s transitive dependency/re-export is not sufficient. @amnn

dongowu: I think we need both here, not one or the other. `sui-indexer-alt-framework` is still used directly by this crate, and `sui-pg-db` is additionally required
because `query!` expands to `::sui_pg_db::...`, which must resolve in the consuming crate. @wlmyng

amnn: Got it, thanks for clarifying

PR #25834, opened 2026-03-13
sui-indexer-alt-framework: rename fetch to checkpoint

evan-wall-mysten:
## Description 

This PR renames `fetch` to `checkpoint` in anticipation of future PRs that will add additional methods like `chain_id` and `latest_checkpoint_number`.

## Test plan 

No functional changes.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Renamed `IngestionClientTrait::fetch` to `IngestionClientTrait::checkpoint`.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
@@ -439,30 +439,31 @@ mod tests {
     use std::fmt::Debug;
     use std::sync::Arc;
     use std::time::Duration;
+
+    use async_trait::async_trait;
     use tokio::time::error::Elapsed;
     use tokio::time::timeout;
 
-    use super::*;
     use crate::ingestion::IngestionConfig;
-    use crate::ingestion::ingestion_client::FetchData;
+    use crate::ingestion::ingestion_client::CheckpointData;
+    use crate::ingestion::ingestion_client::CheckpointResult;
+    use crate::ingestion::ingestion_client::IngestionClientTrait;
     use crate::ingestion::streaming_client::test_utils::MockStreamingClient;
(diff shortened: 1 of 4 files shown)


amnn:
@evan-wall-mysten, please can you explain what those other functions would be used for, why they belong on `IngestionClient`, and how they would be implemented on our existing ingestion clients? It's not clear to me that they would be easy to implement on the object-store-based ingestion clients.

evan-wall-mysten:
> @evan-wall-mysten, please can you explain what those other functions would be used for, why they belong on `IngestionClient`, and how they would be implemented on our existing ingestion clients? It's not clear to me that they would be easy to implement on the object-store-based ingestion clients.

I looked through the `StoreIngestionClient` API and think the 2 functions can be implemented like this, but I haven't actually tested them out:

* `fetch_chain_id()` to store and compare `chain_id` to make sure it does not change
   *  `object-store` impl
      1. `StoreIngestionClient::checkpoint(0).summary.content_digest`
   * `rpc-client` impl
      1. `Client::ledger_client().get_service_info().chain_id()`
* `fetch_latest_checkpoint_number()` to seed a starting checkpoint to index backwards from
   *  `object-store` impl:
      1. `StoreIngestionClient::end_of_epoch_checkpoints.pop()` to get last checkpoint of the most recently ended epoch
      2. exponential binary search to find a range between an existing and missing checkpoint (`self.store.head` to check if file exists)
      3. binary search to find largest existing checkpoint in that range (`self.store.head` to check if file exists)
   * `rpc-client` impl
      1. `Client::ledger_client().get_service_info().checkpoint_height()`

PR #25837, opened 2026-03-13
Extend --gas-payment to support object() and withdraw() syntax

mystenmark:
## Summary
- Extends `--gas-payment` flag to support `ParsedValue<SuiExtraValueArgs>` syntax (same parser as `--inputs`)
- Supports `object(x,y)` for coin objects and `withdraw<sui::balance::Balance<sui::sui::SUI>>(amount)` for address balance withdrawals
- Keeps `--address-balance-gas` flag unchanged for backward compatibility

## Example usage
```
//# programmable --sender A --gas-payment object(0,0)
//# programmable --sender A --gas-payment 'withdraw<sui::balance::Balance<sui::sui::SUI>>(1000000)'
```

## Changes
- Changed `gas_payment` field type in `ProgrammableTransactionCommand` to use `ParsedValue<SuiExtraValueArgs>`
- Added `resolve_gas_payments()` helper to convert parsed values to `ObjectRef`
- Added `sign_sponsor_txn_with_refs()` variant that accepts pre-resolved object refs

## Test plan
- [x] Existing address balance gas tests pass
- [x] Object-based gas payment syntax still works
- [x] New withdraw syntax supported for address balance gas payments

### DIFF
--- a/crates/sui-transactional-test-runner/src/test_adapter.rs
+++ b/crates/sui-transactional-test-runner/src/test_adapter.rs
@@ -567,7 +567,11 @@ impl MoveTestAdapter<'_> for SuiTestAdapter {
                 builder.publish_immutable(modules_bytes, dependencies);
             };
             let pt = builder.finish();
-            let payments = self.get_payments(sender_acc, vec![]);
+            let payments = vec![
+                self.get_object(&sender_acc.gas, None)
+                    .unwrap()
+                    .compute_object_reference(),
+            ];
 
             let transaction = TransactionData::new_programmable(
                 sender_acc.address,
@@ -1115,7 +1119,7 @@ impl MoveTestAdapter<'_> for SuiTestAdapter {
                         .clone()
                         .map_or(sender_acc, |a| self.get_sender(Some(a)));
                     let payment_refs = self.resolve_gas_payments(sponsor_acc, gas_payment)?;
-                    let transaction = self.sign_sponsor_txn_with_refs(
+                    let transaction = self.sign_sponsor_txn(
                         sender,
                         sponsor,
                         payment_refs,
@@ -1793,32 +1797,14 @@ impl SuiTestAdapter {
             /* gas */ Vec<ObjectRef>,
         ) -> TransactionData,
     ) -> Transaction {
-        self.sign_sponsor_txn(sender, None, vec![], move |sender, _, gas| {
+        let sender_acc = self.get_sender(sender.clone());
+        let gas_ref = self
+            .get_object(&sender_acc.gas, None)
+            .unwrap()
+            .compute_object_reference();
+        self.sign_sponsor_txn(sender, None, vec![gas_ref], move |sender, _, gas| {
             txn_data(sender, gas)
         })
-    }
-
-    fn get_payments(&self, sponsor: &TestAccount, payments: Vec<FakeID>) -> Vec<ObjectRef> {
-        let payments = if payments.is_empty() {
-            vec![sponsor.gas]
-        } else {
-            payments
-                .into_iter()
-                .map(|payment| {
-                    self.fake_to_real_object_id(payment)
-                        .expect("Could not find specified payment object")
-                })
-                .collect::<Vec<ObjectID>>()
-        };
-
-        payments
-            .into_iter()
-            .map(|payment| {
-                self.get_object(&payment, None)
-                    .unwrap()
-                    .compute_object_reference()
-            })
-            .collect()
     }
 
     fn resolve_gas_payments(
@@ -1874,34 +1860,6 @@ impl SuiTestAdapter {
     }
 
     fn sign_sponsor_txn(
-        &self,
-        sender: Option<String>,
-        sponsor: Option<String>,
-        payment: Vec<FakeID>,
-        txn_data: impl FnOnce(
-            /* sender */ SuiAddress,
-            /* sponsor */ SuiAddress,
-            /* gas */ Vec<ObjectRef>,
-        ) -> TransactionData,
-    ) -> Transaction {
-        let sender = self.get_sender(sender);
-        let sponsor = sponsor.map_or(sender, |a| self.get_sender(Some(a)));
-
-        let payment_refs = self.get_payments(sponsor, payment);
-
-        let data = txn_data(sender.address, sponsor.address, payment_refs);
-
-        if sender.address == sponsor.address {
-            to_sender_signed_transaction(data, &sender.key_pair)
-        } else {
-            to_sender_signed_transaction_with_multi_signers(
-                data,
-                vec![&sender.key_pair, &sponsor.key_pair],
-            )
-        }
-    }
-
-    fn sign_sponsor_txn_with_refs(
         &self,
         sender: Option<String>,
         sponsor: Option<String>,


amnn:
@mlogan, could we not keep the old command style, but generalise the input format so that it could accept `object(X,Y)` or `coin_reservation(Z)`? Maintaining backwards compatibility is not really required here, because we are the only users of this syntax, so I'd prefer introducing fewer divergences in the syntax.

mystenmark:
> @mlogan, could we not keep the old command style, but generalise the input format so that it could accept `object(X,Y)` or `coin_reservation(Z)`? Maintaining backwards compatibility is not really required here, because we are the only users of this syntax, so I'd prefer introducing fewer divergences in the syntax.

Sure, if you feel strongly about it. The only reason I didn't want to do that is that it gets really verbose when specifying more than one gas object.

mystenmark:
Another option is I can just convert all tests to use the new syntax and get rid of the old one. I didn't do that previously in order to keep the PR smaller

mystenmark:
Ok, here's my proposal:

- Merge this PR
- Send a second PR which removes the existing `--gas-payment` and renames `--gas-payment-vec` => `--gas-payment` and converts all tests.

The reason I'd like to switch to the vec notation (with custom parsing) is because this gives us the following:
- No `--gas-payment` => use default gas coin
- `--gas-payment []` => address balance payment (instead of `--address-balance-gas`)
- explicit gas specification becomes less verbose

amnn:
> * Merge this PR

(No issues with this if this is blocking you now)

> * No `--gas-payment` => use default gas coin

Is there an explicit desire to keep this behaviour? Would it be fine to have the default behaviour be to use address balance gas? It's always possible to request the previous behaviour using something along the lines of `--gas-payment object(0,0)`.

Conversely, it's not really clear that `--gas-payment []` should behave differently from not supplying the flag (just from reading it), and omitting the flag entirely is historically an indication that you don't care how gas is paid (i.e. it's not relevant to the test) -- it would be fair to say that the default behaviour becomes using an address balance.

That would mean that

- `--gas obj(0,0)` mens use the default gas coin.
- No `--gas` means use address balances.
- This is actually even less verbose (although I'm not sure this matters much because it's very rare that we set gas coins explicitly in tests, and as I mentioned earlier, clarity and convention are way more helpful here).

I also have a vested interest in this particular design point to inform how we might expose similar functionality in the new CLI.

mystenmark:
> Is there an explicit desire to keep this behaviour?

I mainly worry that changing this would produce an almost unreviewably large PR and I'd be worried about breaking tests (breaking in the sense that they still pass but no longer test what they were supposed to).

amnn:
If that's your main fear, I'm happy to take that on -- I'm fairly familiar with most/all the transactional tests.

crates/sui-adapter-transactional-tests/tests/address_balances/address_balance_gas.move
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-// Tests that --address-balance-gas pays for gas from the address balance,
+// Tests that --gas-payment-vec [] (address balance gas) pays for gas from the address balance,
```
tnowacki: We should have tests that use the new syntax used here?

crates/sui-transactional-test-runner/src/args.rs
```diff
@@ -878,3 +882,110 @@ fn parse_policy(x: &str) -> anyhow::Result<u8> {
...
+    matches!(payment, Some(list) if list.is_address_balance())
+}
+
+/// Parse a gas payment list in the format: [(1,0), coin_reservation(1000)]
+/// An empty list [] indicates address balance gas payment.
+pub fn parse_gas_payment_list(s: &str) -> anyhow::Result<GasPaymentList> {
+    let s = s.trim();
+    if !s.starts_with('[') || !s.ends_with(']') {
+        bail!(
+            "Gas payment list must be enclosed in brackets, e.g. [(1,0), coin_reservation(1000)]"
```
tnowacki: You won't be able to have spaces here, since clap will remove them

crates/sui-transactional-test-runner/src/args.rs
```diff
@@ -878,3 +882,110 @@ fn parse_policy(x: &str) -> anyhow::Result<u8> {
...
+}
+
+/// Returns true if the payment option indicates address balance gas (empty list).
+pub fn is_address_balance_gas(payment: &Option<GasPaymentList>) -> bool {
+    matches!(payment, Some(list) if list.is_address_balance())
+}
+
+/// Parse a gas payment list in the format: [(1,0), coin_reservation(1000)]
+/// An empty list [] indicates address balance gas payment.
+pub fn parse_gas_payment_list(s: &str) -> anyhow::Result<GasPaymentList> {
```
tnowacki: I really don't like having what is basically yet another parser, especially one that is operating over the string directly this way instead of using the token stream like the value/type parsers

crates/sui-transactional-test-runner/src/args.rs
```diff
@@ -172,8 +174,10 @@ pub struct UpgradePackageCommand {
     pub sender: String,
     #[clap(long = "gas-budget")]
     pub gas_budget: Option<u64>,
-    #[clap(long = "address-balance-gas")]
-    pub address_balance_gas: bool,
+    #[clap(long = "gas-payment", value_parser = parse_fake_id)]
+    pub gas_payment: Option<Vec<FakeID>>,
+    #[clap(long = "gas-payment-vec", value_parser = parse_gas_payment_list)]
+    pub gas_payment_vec: Option<GasPaymentList>,
```
tnowacki: I think it might be better to parse this into a `ParsedValue::<SuiExtraValueArgs>`. That way we aren't inventing a completely new syntax?

crates/sui-transactional-test-runner/src/args.rs
```diff
@@ -172,8 +174,10 @@ pub struct UpgradePackageCommand {
     pub sender: String,
     #[clap(long = "gas-budget")]
     pub gas_budget: Option<u64>,
-    #[clap(long = "address-balance-gas")]
-    pub address_balance_gas: bool,
+    #[clap(long = "gas-payment", value_parser = parse_fake_id)]
+    pub gas_payment: Option<Vec<FakeID>>,
```
tnowacki: There are two tests that use this arg, I would just replace it since the new one supersedes it

crates/sui-transactional-test-runner/src/args.rs
tnowacki: It is also worth considering doing an `Option<Vec<ParsedValue::<SuiExtraValueArgs>>` with `num_args(0..), action = clap::ArgAction::Append`. Though I'm not sure if it works with Option? Should though... maybe?

crates/sui-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1812,6 +1821,58 @@ impl SuiTestAdapter {
...
+                    )
+                }
+                _ => bail!(
+                    "Invalid gas payment: only object(...) and withdraw<...>(...) are allowed"
+                ),
+            })
+            .collect()
+    }
+
     fn sign_sponsor_txn(
```
tnowacki, line 1862: Is this used still? And if so, can we kill it and use the other one?

crates/sui-transactional-test-runner/src/test_adapter.rs
```diff
@@ -1784,39 +1797,73 @@ impl SuiTestAdapter {
...
+                }
+                _ => bail!(
+                    "Invalid gas payment: only object(...) and withdraw<...>(...) are allowed"
+                ),
             })
             .collect()
     }
 
-    fn sign_sponsor_txn(
+    fn sign_sponsor_txn_with_refs(
```
tnowacki: Do we need to rename it?

PR #25841, opened 2026-03-13
[forking-stack] add APIs to simulacrum for forking 1/5

stefan-mysten:
## Description 

This PR adds the simulacrum/checkpoint-builder foundation for forking mode:

In simulacrum:
- adds `execute_transaction_impersonating` to execute txns for senders without private keys.
- adds `new_from_custom_state` so simulacrum can start from a custom checkpoint & system state.

## Test plan 

Existing CI + forking tests.

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-transactional-test-runner/src/test_adapter.rs
+++ b/crates/sui-transactional-test-runner/src/test_adapter.rs
@@ -2800,7 +2800,7 @@ async fn init_sim_executor(
 
     // Get the actual object values from the simulator
     for (name, (addr, kp)) in account_kps {
-        let o = sim.store().owned_objects(addr).next().unwrap();
+        let o = sim.store_dyn().owned_objects(addr).next().unwrap();
         objects.push(o.clone());
         account_objects.insert(name.clone(), o.id());
 
@@ -2814,7 +2814,7 @@ async fn init_sim_executor(
         );
     }
     let o = sim
-        .store()
+        .store_dyn()
         .owned_objects(default_account_kp.0)
         .next()
         .unwrap();
@@ -2826,7 +2826,7 @@ async fn init_sim_executor(
     objects.push(o.clone());
 
     for (i, (addr, key_pair_wrapper)) in addr_keys.into_iter().enumerate() {
-        let o = sim.store().owned_objects(addr).next().unwrap();
+        let o = sim.store_dyn().owned_objects(addr).next().unwrap();
         let validator_account = TestAccount {
             address: addr,
             key_pair: key_pair_wrapper.account_key_pair,


crates/simulacrum/src/lib.rs
```diff
@@ -420,6 +503,10 @@ impl<R, S: store::SimulatorStore> Simulacrum<R, S> {
             epoch_commitments: vec![],
         };
         let committee = CommitteeWithKeys::new(&self.keystore, self.epoch_state.committee());
+        println!(
+            "Creating checkpoint for end of epoch {} with committee {:?}",
+            next_epoch, committee.committee
+        );
```
stefan-mysten: Oops, forgot to remove this. We should probably just have tracing here.

crates/simulacrum/src/lib.rs
```diff
@@ -437,6 +505,14 @@ impl<R, S: store::SimulatorStore> Simulacrum<R, S> {
         &self.store
     }
 
+    pub fn store_typed(&self) -> &S {
```
dariorussi: not clear about this name, can you explain? `store_typed`

crates/simulacrum/src/lib.rs
```diff
@@ -445,6 +521,14 @@ impl<R, S: store::SimulatorStore> Simulacrum<R, S> {
         self.epoch_state.epoch_start_state()
     }
 
+    pub fn system_state(&self) -> SuiSystemState {
+        self.store.get_system_state()
+    }
+
+    pub fn protocol_config(&self) -> ProtocolConfig {
+        self.epoch_state.protocol_config().clone()
```
dariorussi: any reason to have the clone here instead of returning a ref and making the caller decide?

crates/simulacrum/src/lib.rs
```diff
@@ -205,6 +210,69 @@ impl<R, S: store::SimulatorStore> Simulacrum<R, S> {
...
+    /// be useful for testing specific scenarios or starting from a non-genesis state.
+    ///
+    /// Note: The provided `checkpoint` and `system_state` should be consistent with each other, i.e.
+    /// the checkpoint's epoch and sequence number should align with the system state's epoch and
+    /// the objects in the checkpoint should be reflected in the system state. Inconsistencies may
+    /// lead to unexpected behavior.
+    pub fn new_from_custom_state(
+        keystore: KeyStore,
+        checkpoint: VerifiedCheckpoint,
+        system_state: SuiSystemState,
```
dariorussi, line 225: cannot see the call site yet but I wonder if the SuiSystemState should come from the checkpoint and not being passed in. In fact, I wonder if the system state is needed at all or if we should build everything from the checkpoint

crates/simulacrum/src/lib.rs
```diff
@@ -205,6 +210,69 @@ impl<R, S: store::SimulatorStore> Simulacrum<R, S> {
...
+        config: &NetworkConfig,
+        store: S,
+        rng: R,
+        acc_initial_shared_obj_version: Option<SequenceNumber>,
```
dariorussi: I am not clear what this is used for though it is my lack of understanding here not lack of clarity

crates/simulacrum/src/lib.rs
```diff
@@ -205,6 +210,69 @@ impl<R, S: store::SimulatorStore> Simulacrum<R, S> {
...
+    /// The transaction effects and optional execution error
+    pub fn execute_transaction_impersonating(
+        &mut self,
+        transaction_data: TransactionData,
+    ) -> anyhow::Result<(TransactionEffects, Option<ExecutionError>)> {
+        // Create dummy signatures for each required signer
+        let pk = SuiKeyPair::Ed25519(get_key_pair().1);
+        let sig = pk.sign(transaction_data.sender().as_ref());
+        // Create sender signed data with dummy signatures
+        let sender_signed_data = SenderSignedData::new(transaction_data, vec![sig.into()]);
```
dariorussi, line 265: this is fine though it would be great to understand what is that we are verifying in simulacrum. I looked at the code and it's not too clear but I'll play with it more.
Also wonder if the sender signed data is what should come in input and then we could let the caller decide what that is. 
Not sure it is possible

crates/simulacrum/src/lib.rs
stefan-mysten: Good point, I will fix!

crates/simulacrum/src/lib.rs
stefan-mysten: The `Simulacrum` type has a `epoch_start_state` field with type `EpochState`. 
It might be an issue here, but the way I found to initialize this EpochState is by passing a `SuiSystemState`.
https://github.com/MystenLabs/sui/blob/7ea77f3aeb6db60337857014357d1a3bc93a94f3/crates/simulacrum/src/epoch_state.rs?plain=1#L28

stefan-mysten: Regarding the part with building the state from checkpoint, I do not see how to do that:
https://github.com/MystenLabs/sui/blob/7ea77f3aeb6db60337857014357d1a3bc93a94f3/crates/simulacrum/src/epoch_state.rs?plain=1#L47

crates/simulacrum/src/lib.rs
stefan-mysten: At some point (not sure when) we started having an accumulator object. Simulacrum uses `MockCheckpointBuilder` to create checkpoints. This type has a function called `get_settlement_txns`, which uses the accumulator root obj initial shared version. 

This field `acc_initial_shared_obj_version` basically passes in the version of this accumulator object at the forked checkpoint, when the forked network starts. Otherwise, by default, `MockCheckpointBuilder` would set this version to 1, which breaks executing advancements of checkpoints / epochs (iirc).

crates/simulacrum/src/lib.rs
stefan-mysten: If you want to "impersonate" an address, a.k.a. you set the `sender` to be that address for that transaction, but you do not have the keys, one still needs to sign that transaction. There is a check that at least one signature exists, which happens here:
first call: https://github.com/MystenLabs/sui/blob/7ea77f3aeb6db60337857014357d1a3bc93a94f3/crates/simulacrum/src/lib.rs?plain=1#L292
then the one above calls here at some point: 
https://github.com/MystenLabs/sui/blob/7ea77f3aeb6db60337857014357d1a3bc93a94f3/crates/sui-types/src/signature_verification.rs?plain=1#L140-L149

crates/simulacrum/src/lib.rs
stefan-mysten: There's already one `pub fn store(&self) -> &dyn SimulatorStore`. In some cases, this wasn't sufficient, and this was the best name I could find for returning a `&S`.

crates/simulacrum/src/store/mod.rs
amnn, line 1: Spurious change

crates/simulacrum/src/lib.rs
amnn: I think the `SuiSystemState` that you provide needs to match the state of the checkpoint after `checkpoint` (if that is the last checkpoint in an epoch, then `system_state` should be compatible with the new epoch coming next), because `checkpoint` is used to initialise the previous checkpoint in the `MockCheckpointBuilder`, while `system_state` is used to initialise the current system state.

crates/simulacrum/src/lib.rs
amnn: - The signature is just bytes right? Do you need to go to the trouble of actually creating a valid signature, or could you just provide some empty bytes?
- What about the case where the transaction has a sponsor as well as a sender? Should you be paying attention to `required_signers` like the check is and producing that many dummy values?

crates/simulacrum/src/lib.rs
amnn: nit: I would personally call this one `store` and call that one `store_dyn`.

crates/sui-core/src/mock_checkpoint_builder.rs
```diff
@@ -33,10 +33,14 @@ pub struct MockCheckpointBuilder {
...
     epoch_rolling_gas_cost_summary: GasCostSummary,
     epoch: u64,
+    object_start_version: SequenceNumber,
 }
 
 impl MockCheckpointBuilder {
-    pub fn new(previous_checkpoint: VerifiedCheckpoint) -> Self {
+    pub fn new(
+        previous_checkpoint: VerifiedCheckpoint,
+        object_start_version: Option<SequenceNumber>,
```
amnn: This is too generic a name for what this is -- this is exclusively used for `0xacc` right?

crates/simulacrum/src/lib.rs
dariorussi: I assumed all of that can be determined by the checkpoint you start from. Is that not the case?

amnn: I don't think so. If you're starting from a checkpoint in the middle of the epoch, you need information from the start of the epoch to complete `SuiSystemState` (but in the scenario I described where you are starting from the last checkpoint in the previous epoch, you may be able to extract the system state for the next epoch from the end of epoch data in the checkpoint)

crates/sui-core/src/mock_checkpoint_builder.rs
stefan-mysten: Yes. I'll rework it, thanks!

crates/simulacrum/src/lib.rs
stefan-mysten: > from a checkpoint in the middle of the epoch, you need information from the start of the epoch to complete SuiSystemState

Yes, that's correct. We need to know information about the epoch starting point (timestamp, committee, etc).

I see what you mean regarding providing the state for after that checkpoint you forked from. I will make sure this is correct in `forking` side of things.

For the part where you fork from the last checkpoint in an epoch, that's interesting one. I am not sure how to handle it, because in the forked network you can still advance the checkpoint after you forked from that last checkpoint, and then of course you'd have a different system state for the next epoch.

crates/simulacrum/src/lib.rs
stefan-mysten: - The signatures required by SenderSignedData are of type `GenericSignature`, which is an enum, and it does not have `From<Vec<u8>>`. Also, I think there's a need to actually specify at least of flag representing the signature type.
- good catch, I have not considered the case for having a sponsor. I assume that's the right way to do, but I would need to test it out. I made an issue on Linear to keep track of this.

crates/sui-core/src/mock_checkpoint_builder.rs
stefan-mysten: Renamed to what it's called everywhere else where this is passed: `accumulator_root_obj_initial_shared_version`. It's long but it's consistent everywhere.

stefan-mysten:
> I'd be curious to understand what goes wrong if that's always set to OBJECT_START_VERSION though.

I went ahead to try again without the checkpoint builder changes, and all works. I am confused; there were some transactions where I wasn't able to execute because of this, though I can't recall which ones. That's why I added it.

crates/simulacrum/src/lib.rs
stefan-mysten: I removed it and will add back if needed.

crates/simulacrum/src/lib.rs
amnn: My comment was that the doc comment for this function is wrong and could be a source of bugs because of the end-of-epoch checkpoint edge case:

```
    /// Note: The provided `checkpoint` and `system_state` should be consistent with each other, i.e.
    /// the checkpoint's epoch and sequence number should align with the system state's epoch and
    /// the objects in the checkpoint should be reflected in the system state. Inconsistencies may
    /// lead to unexpected behavior.
```

The first line is correct, but the details are not:

- the system state's epoch should align with the epoch for the checkpoint after `checkpoint`.
- I'm not really sure what "the objects in the checkpoint should be reflected in the system sate" means, but it sounds like a property of perhaps only the first or last checkpoint in the epoch.

crates/simulacrum/src/lib.rs
```diff
@@ -516,7 +598,7 @@ impl<R, S: store::SimulatorStore> Simulacrum<R, S> {
         // explicitly cordon off the faucet account from the rest of the accounts though.
         let (sender, key) = self.keystore().accounts().next().unwrap();
         let object = self
-            .store()
+            .store_dyn()
```
amnn, line 601: Is there a reason these functions need to use the `dyn` variant of `store`?

amnn: ...in fact do you need the dyn version of `Store` at all, I wonder?

crates/simulacrum/src/lib.rs
dariorussi: so for my understanding you cannot go from the current checkpoint to the latest epoch change and go from there?
I would be shocked you cannot

crates/simulacrum/src/lib.rs
dariorussi: my question was if we could move the signature "generation" out of here and pass it in.
That was a user can provide whatever they want.
Maybe an API to create a signature like that? or an api to create the specific "empty" sig?

crates/simulacrum/src/lib.rs
amnn: It depends what you mean by "go from" the checkpoint does not contain all the details of the epoch it is in, but if you had a way of fetching epochs you could fetch the corresponding epoch.

crates/simulacrum/src/lib.rs
stefan-mysten: I don't know - I usually try not to mess with what exists, unless I really understand it. When I start putting up PRs for forking, I will play more with this.

crates/simulacrum/src/lib.rs
stefan-mysten: > the system state's epoch should align with the epoch for the checkpoint after checkpoint.

I will fix that and rework the comment.

> if you had a way of fetching epochs you could fetch the corresponding epoch.

We can fetch epoch info with `sui-data-store`, so will keep that in mind when I rework the forking PRs.

crates/simulacrum/src/lib.rs
stefan-mysten: I think we can do that. I'll probably have to think more on it.

My brain is a bit too coupled to the CLI, so in order for using the CLI and impersonating an address, we'd have to either create and sign the tx manually, or add some new flags. With what exists in this PR, this is not an issue for the CLI, but of course it's bad for other reasons (what if users actually care about their sigs when they're not impersonating an address and they own the keys).

crates/simulacrum/src/lib.rs
stefan-mysten: @amnn
I had a look and we can indeed switch - there are advantages and disadvantages, but I think in this case it's perfectly fine. For `forking`, I needed `store -> &S` to take advantage of some functions on `S` that were not defined in the `SimulatorStore` trait.

Put up a PR here: https://github.com/MystenLabs/sui/pull/25933

PR #25842, opened 2026-03-13
[forking-stack] add APIs to sui-data-store 2/5

stefan-mysten:
## Description 

This PR adds checkpoint API for the various stores, and one writing object API. For the latter, we need it when executing transactions via the simulacrum as we need to write the new objects to the cache.

## Test plan 
Tested via the forking tool tests + manually running the forking tool.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14703,12 +14703,16 @@ dependencies = [
  "cynic-codegen",
  "fastcrypto",
  "lru 0.10.0",
+ "prost 0.14.1",
  "reqwest",
  "serde",
- "sui-config",
+ "sui-rpc",
+ "sui-rpc-api",
  "sui-types",
  "tokio",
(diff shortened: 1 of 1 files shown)


stefan-mysten:
> I do not know how possible it is and we'll talk more later about it, but it would be great to re-think the `read_through.rs` file and maybe create more specific version of the API? Maybe same for the write? Bottom line I think we will have a bit to rework in here but it's difficult to make a call without seeing how it is hooked up in the forking tool. So I'd say check it in and we'll talk more when we hook up forking

Thanks for having a look! Agreed it will need some refactoring.

PR #25854, opened 2026-03-16
[gasless] Add gasless rate limiter

alex-mysten:
## Description 

add per-validator rate limit for gasless txns

## Test plan 

simtests + unit tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/gasless_rate_limiter.rs
+++ b/crates/sui-core/src/gasless_rate_limiter.rs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::sync::Arc;
-use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Instant;
 
 use parking_lot::Mutex;
@@ -12,15 +11,21 @@ use sui_protocol_config::ProtocolConfig;
 /// within the current 1-second window. Updated by the consensus handler on
 /// each commit, and read by the rate limiter to make admission decisions.
 pub struct ConsensusGaslessCounter {
-    window_second: AtomicU64,
-    count: AtomicU64,
+    inner: Mutex<ConsensusWindowInner>,
+}
+
+struct ConsensusWindowInner {
+    window_second: u64,
+    count: u64,
 }
 
 impl Default for ConsensusGaslessCounter {
     fn default() -> Self {
         Self {
-            window_second: AtomicU64::new(0),
-            count: AtomicU64::new(0),
+            inner: Mutex::new(ConsensusWindowInner {
+                window_second: 0,
+                count: 0,
+            }),
         }
     }
 }
@@ -28,19 +33,17 @@ impl Default for ConsensusGaslessCounter {
 impl ConsensusGaslessCounter {
     pub fn record_commit(&self, commit_timestamp_ms: u64, gasless_count: u64) {
         let second = commit_timestamp_ms / 1000;
-        let current = self.window_second.load(Ordering::Relaxed);
-        if second > current {
-            self.window_second.store(second, Ordering::Relaxed);
-            self.count.store(gasless_count, Ordering::Relaxed);
-        } else if second == current {
-            self.count.fetch_add(gasless_count, Ordering::Relaxed);
+        let mut inner = self.inner.lock();
+        if second > inner.window_second {
+            inner.window_second = second;
+            inner.count = gasless_count;
+        } else if second == inner.window_second {
+            inner.count += gasless_count;
         }
     }
 
-    pub fn current_count(&self) -> (u64, u64) {
-        let window = self.window_second.load(Ordering::Relaxed);
-        let count = self.count.load(Ordering::Relaxed);
-        (window, count)
+    pub fn current_count(&self) -> u64 {
+        self.inner.lock().count
     }
 }
 
@@ -93,8 +96,7 @@ impl GaslessRateLimiter {
         let Some(max_tps) = config.gasless_max_tps_as_option() else {
             return true;
         };
-        let (_, consensus_count) = self.consensus_counter.current_count();
-        if consensus_count >= max_tps {
+        if self.consensus_counter.current_count() >= max_tps {
             return false;
         }
         // no single validator can admit more than max_tps burst
@@ -150,10 +152,10 @@ mod tests {
     fn test_record_commit_new_window_resets() {
         let counter = ConsensusGaslessCounter::default();
         counter.record_commit(1000, 10);
-        assert_eq!(counter.current_count(), (1, 10));
+        assert_eq!(counter.current_count(), 10);
 
         counter.record_commit(2000, 5);
-        assert_eq!(counter.current_count(), (2, 5));
+        assert_eq!(counter.current_count(), 5);
     }
 
     #[test]
@@ -161,7 +163,7 @@ mod tests {
         let counter = ConsensusGaslessCounter::default();
         counter.record_commit(1000, 10);
         counter.record_commit(1500, 7);
-        assert_eq!(counter.current_count(), (1, 17));
+        assert_eq!(counter.current_count(), 17);
     }
 
     #[test]
@@ -169,7 +171,7 @@ mod tests {
         let counter = ConsensusGaslessCounter::default();
         counter.record_commit(2000, 10);
         counter.record_commit(1000, 99);
-        assert_eq!(counter.current_count(), (2, 10));
+        assert_eq!(counter.current_count(), 10);
     }
 
     // -- Local admission counter tests --


crates/sui-core/src/authority_server.rs
```diff
@@ -402,6 +404,12 @@ impl ValidatorServiceMetrics {
                 registry,
             )
             .unwrap(),
+            gasless_rate_limited_count: register_int_counter_with_registry!(
+                "validator_service_gasless_rate_limited_count",
+                "Number of free tier transactions rejected by rate limiter",
```
aschran: nit, "gasless"

crates/sui-core/src/authority_server.rs
```diff
@@ -751,6 +765,24 @@ impl ValidatorService {
...
+                .transaction_data()
+                .is_gasless_transaction()
+                && !self
+                    .gasless_limiter
+                    .try_acquire(epoch_store.protocol_config())
+            {
+                metrics.gasless_rate_limited_count.inc();
+                results[idx] = Some(SubmitTxResult::Rejected {
+                    error: SuiErrorKind::ValidatorOverloadedRetryAfter {
+                        retry_after_secs: 0,
```
aschran: is it possible to set this to something more useful? fine if not

crates/sui-core/src/consensus_handler.rs
```diff
@@ -1208,6 +1220,13 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
             new_jwks,
         } = self.build_commit_handler_input(transactions);
 
+        let gasless_count = user_transactions
```
aschran: i know it's only two lines, but for parallelism would it make sense to put this inside a `process_gasless_transactions` fn? feel free to ignore me here if it seems like overkill

crates/sui-core/src/gasless_rate_limiter.rs
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use sui_protocol_config::ProtocolConfig;
+
+pub struct ConsensusGaslessCounter {
```
aschran, line 13: document what this is for?

crates/sui-core/src/gasless_rate_limiter.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    pub fn current_count(&self) -> (u64, u64) {
+        let window = self.window_second.load(Ordering::Relaxed);
+        let count = self.count.load(Ordering::Relaxed);
+        (window, count)
+    }
+}
+
+#[derive(Clone)]
+pub struct GaslessRateLimiter {
```
aschran, line 79: document?

crates/sui-protocol-config/src/lib.rs
```diff
@@ -1913,6 +1913,9 @@ pub struct ProtocolConfig {
 
     /// Allowed token types for gasless transactions, with minimum transfer sizes per token.
     gasless_allowed_token_types: Option<Vec<(String, u64)>>,
+
+    /// Max tps for gasless transactions. Zero when unset.
```
aschran: should this be unlimited when unset, zero when set to zero?

crates/sui-core/src/authority_server.rs
alex-mysten: Addressed

crates/sui-core/src/authority_server.rs
alex-mysten: Addressed - 1 makes sense as it's a per-second limit

crates/sui-core/src/consensus_handler.rs
alex-mysten: Addressed

crates/sui-core/src/gasless_rate_limiter.rs
alex-mysten: Addressed

crates/sui-core/src/gasless_rate_limiter.rs
alex-mysten: Addressed

crates/sui-protocol-config/src/lib.rs
alex-mysten: Addressed

crates/sui-core/src/gasless_rate_limiter.rs
```diff
@@ -0,0 +1,242 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::time::Instant;
+
+use parking_lot::Mutex;
```
aschran, line 7: curious why this mutex choice?

crates/sui-core/src/gasless_rate_limiter.rs
```diff
@@ -0,0 +1,242 @@
...
+            window_second: AtomicU64::new(0),
+            count: AtomicU64::new(0),
+        }
+    }
+}
+
+impl ConsensusGaslessCounter {
+    pub fn record_commit(&self, commit_timestamp_ms: u64, gasless_count: u64) {
+        let second = commit_timestamp_ms / 1000;
+        let current = self.window_second.load(Ordering::Relaxed);
```
aschran: i personally find these orderings so subtle and complicated, and thus err on the side of using mutexes unless there's a good reason not to. all of this to say - just trusting that "Relaxed" is correct and what you want in all of these cases?

aschran: fwiw I asked Claude to double check and it said this

Threading model                                                                             
                        
  - Single writer: record_commit() is called only from the consensus handler                  
  (process_gasless_transactions in consensus_handler.rs), which processes commits
  sequentially.                                                                               
  - Multiple readers: current_count() is called from validator service threads handling
  incoming transactions (via try_acquire).                                                    
   
  Issue 1: Store ordering on window transition (real bug on ARM)                              
                        
  In record_commit, when second > current:                                                    
                        
  self.window_second.store(second, Ordering::Relaxed);   // Store A                           
  self.count.store(gasless_count, Ordering::Relaxed);     // Store B
                                                                                              
  With Relaxed, there is no guarantee that a reader on another thread observes these stores in
   program order. On weakly-ordered architectures (ARM/Apple Silicon — which is the target    
  platform here), a reader could see:                                                         
                        
  - New window_second, stale count: Reader sees the window advanced but count still holds the 
  accumulated value from the previous window. If the old count was high (e.g., 500) and the
  new count is 3, the reader falsely rejects transactions until the count store propagates.   
                        
  This is the most concerning case. On x86 (TSO), stores from the same thread are always      
  observed in order, so this wouldn't manifest — but Sui validators run on ARM (AWS Graviton, 
  Apple Silicon) where this is a real hazard.

    Fix for Issue 1

  Use Release/Acquire to establish a happens-before relationship. Store count first, then     
  publish window_second with Release. Reader loads window_second with Acquire:
                                                                                              
  // record_commit — writer
  if second > current {                                                                       
      self.count.store(gasless_count, Ordering::Relaxed);       // must be visible...
      self.window_second.store(second, Ordering::Release);      // ...before this becomes     
  visible               
  } else if second == current {                                                               
      self.count.fetch_add(gasless_count, Ordering::Relaxed);   // fine as-is
  }                                                                                           
   
  // current_count — reader                                                                   
  pub fn current_count(&self) -> (u64, u64) {
      let window = self.window_second.load(Ordering::Acquire);  // synchronizes-with Release  
  store                                                                                       
      let count = self.count.load(Ordering::Relaxed);                                         
      (window, count)                                                                         
  }                     

  This guarantees: if the reader sees the new window_second, it also sees the corresponding   
  new count.
                                                                                              
  Issue 2: Non-atomic compound read in current_count (minor, acceptable)                      
   
  The two loads in current_count aren't atomic as a pair. Between reading window and count,   
  the writer could advance the window and reset the count. This means the reader could see:
                                                                                              
  - window = old, count = new (reset) → sees artificially low count for old window → slightly 
  over-admits
                                                                                              
  For a best-effort rate limiter this is acceptable — it's a transient one-read inconsistency 
  and the local counter provides a hard backstop.

    Issue 3: fetch_add in same-window branch (correct)                                          
   
  self.count.fetch_add(gasless_count, Ordering::Relaxed);                                     
                                                                                              
  This is a single atomic RMW operation. Since there's only one writer, there's no lost-update
   concern. The reader just sees whatever value is current. Relaxed is fine here.             
                                                                                              
  Issue 4: TOCTOU in record_commit (correct given single writer)                              
   
  The load-compare-store on window_second is not atomic as a compound operation. If multiple  
  threads called record_commit concurrently, two threads could both see second > current and
  race to reset count, losing one's gasless_count. But since the consensus handler is         
  single-threaded, this is safe. Worth a comment though.

crates/sui-core/src/gasless_rate_limiter.rs
alex-mysten: It's the most commonly used one in Sui, apparently locks & unlocks faster than std when contention is low. Is there a reason to avoid using it here?

crates/sui-core/src/gasless_rate_limiter.rs
alex-mysten: Makes sense - and good find. Switched to use a Mutex.

crates/sui-core/src/gasless_rate_limiter.rs
aschran: Interesting that it's more common, my thinking was that std mutex would be default unless there's a reason otherwise but maybe not in this case!

PR #25859, opened 2026-03-16
Allow consumption of input coins for gasless tx

aschran:
## Description 

- All input coins must be destroyed.
- Any storage rebate is consumed as gas.

## Test plan 

Wrote simtests and transactional tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-adapter-transactional-tests/tests/gasless/gasless_coin.move
+++ b/crates/sui-adapter-transactional-tests/tests/gasless/gasless_coin.move
@@ -3,12 +3,12 @@
 
 // Tests gasless transactions with coin object inputs, MergeCoins, and SplitCoins.
 
-//# init --addresses test=0x0 --accounts A B C --enable-gasless --enable-accumulators
+//# init --addresses test=0x0 --accounts A B C --enable-gasless --enable-accumulators --shared-object-deletion true
 
 //# publish --sender A
 #[allow(deprecated_usage)]
 module test::usdc {
-    use sui::coin;
+    use sui::coin::{Self, Coin};
 
     public struct USDC has drop {}
 
@@ -18,6 +18,14 @@ module test::usdc {
         );
         transfer::public_freeze_object(metadata);
         transfer::public_transfer(treasury_cap, ctx.sender());
+    }
+
+    public fun freeze_coin(coin: Coin<USDC>) {
+        transfer::public_freeze_object(coin);
+    }
+
+    public fun share_coin(coin: Coin<USDC>) {
+        transfer::public_share_object(coin);
     }
 }
 
@@ -73,3 +81,53 @@ module test::usdc {
 //# view-funds sui::balance::Balance<test::usdc::USDC> B
 
 //# view-funds sui::balance::Balance<test::usdc::USDC> C
+
+// --- Test 4: Immutable coin input rejected ---
+
+//# programmable --sender A --inputs object(1,1) 1000
+// Mint and freeze a coin
+//> 0: sui::coin::mint<test::usdc::USDC>(Input(0), Input(1));
+//> test::usdc::freeze_coin(Result(0));
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs object(17,0) @B
+// Reject: immutable coin
+//> 0: sui::coin::send_funds<test::usdc::USDC>(Input(0), Input(1));
+
+// --- Test 5: Shared coin input rejected ---
+
+//# programmable --sender A --inputs object(1,1) 1000
+// Mint and share a coin
+//> 0: sui::coin::mint<test::usdc::USDC>(Input(0), Input(1));
+//> test::usdc::share_coin(Result(0));
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs object(19,0) @B
+// Reject: shared coin (used)
+//> 0: sui::coin::send_funds<test::usdc::USDC>(Input(0), Input(1));
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs object(19,0) @B
+// Reject: shared coin (unused)
+//> 0: sui::balance::zero<test::usdc::USDC>();
+//> 1: sui::balance::send_funds<test::usdc::USDC>(Result(0), Input(1));
+
+// --- Test 6: Party coin input succeeds (used and unused) ---
+
+//# programmable --sender A --inputs object(1,1) 2000 @A
+// Mint two coins and party-transfer them
+//> 0: sui::coin::mint<test::usdc::USDC>(Input(0), Input(1));
+//> 1: sui::coin::mint<test::usdc::USDC>(Input(0), Input(1));
+//> 2: sui::party::single_owner(Input(2));
+//> sui::transfer::public_party_transfer<sui::coin::Coin<test::usdc::USDC>>(Result(0), Result(2));
+//> sui::transfer::public_party_transfer<sui::coin::Coin<test::usdc::USDC>>(Result(1), Result(2));
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs object(22,0) @B
+// Success: party coin input used
+//> 0: sui::coin::send_funds<test::usdc::USDC>(Input(0), Input(1));
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs object(22,1) @B
+// Reject: party coin input unused, fails post-execution check
+//> 0: sui::balance::zero<test::usdc::USDC>();
+//> 1: sui::balance::send_funds<test::usdc::USDC>(Result(0), Input(1));
+
+//# create-checkpoint
+
+//# view-funds sui::balance::Balance<test::usdc::USDC> B
--- a/crates/sui-adapter-transactional-tests/tests/gasless/gasless_coin.snap
+++ b/crates/sui-adapter-transactional-tests/tests/gasless/gasless_coin.snap
@@ -1,18 +1,18 @@
 ---
 source: external-crates/move/crates/move-transactional-test-runner/src/framework.rs
 ---
-processed 17 tasks
+processed 27 tasks
 
 init:
 A: object(0,0), B: object(0,1), C: object(0,2)
 
-task 1, lines 8-22:
+task 1, lines 8-30:
 //# publish --sender A
 created: object(1,0), object(1,1), object(1,2)
 mutated: object(0,0)
-gas summary: computation_cost: 1000000, storage_cost: 10776800,  storage_rebate: 0, non_refundable_storage_fee: 0
+gas summary: computation_cost: 1000000, storage_cost: 11567200,  storage_rebate: 0, non_refundable_storage_fee: 0
 
-task 3, lines 26-31:
+task 3, lines 34-39:
 //# programmable --sender A --inputs 5000 object(1,1) @A
 // Mint three USDC coins for A
 //> 0: sui::coin::mint<test::usdc::USDC>(Input(1), Input(0));
@@ -24,11 +24,11 @@ mutated: object(0,0), object(1,1)
 unchanged_shared: 0x0000000000000000000000000000000000000000000000000000000000000403
 gas summary: computation_cost: 1000000, storage_cost: 6657600,  storage_rebate: 2663496, non_refundable_storage_fee: 26904
 
-task 4, lines 33-35:
+task 4, lines 41-43:
 //# create-checkpoint
 Checkpoint created: 1
 
-task 5, lines 37-39:
+task 5, lines 45-47:
 //# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs object(3,0) @B
 // Gasless: send a coin object to B via coin::send_funds
 //> 0: sui::coin::send_funds<test::usdc::USDC>(Input(0), Input(1));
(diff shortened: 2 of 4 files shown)


sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -396,12 +396,12 @@ mod checked {
...
             && result.is_ok()
-            && temporary_store.has_non_accumulator_writes()
+            && let Err(msg) = temporary_store.check_gasless_execution_requirements()
         {
-            debug_fatal!(
-                "gasless transaction produced non-accumulator writes — command validation should have prevented this"
-            );
-            result = Err(ExecutionErrorKind::InvariantViolation.into());
+            result = Err(ExecutionError::new_with_source(
+                ExecutionErrorKind::InsufficientGas,
```
alex-mysten: I switched this to `InvariantViolation` as if we reach this point our validity checks are incorrect. Is that still a correct assumption?

crates/test-cluster/src/addr_balance_test_env.rs
```diff
@@ -239,6 +240,125 @@ impl TestEnv {
         (publisher, coin_a_type)
     }
 
+    /// Publish the mintable_coin package and return (publisher, package_id, coin_type, treasury_cap_ref).
+    /// The TreasuryCap is unfrozen so new Coin objects can be minted.
+    pub async fn setup_mintable_coin(&mut self) -> (SuiAddress, ObjectID, TypeTag, ObjectRef) {
```
alex-mysten, line 246: Can we extend / reuse the coin minting helpers we already have in test_env?

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -635,6 +640,48 @@ mod checked {
...
+        protocol_config: &ProtocolConfig,
+    ) -> UserInputResult<()> {
+        use sui_types::transaction::parse_gasless_allowed_token_types;
+        use sui_types::type_input::TypeInput;
+
+        #[allow(unused_mut)]
+        let mut allowed_token_types = parse_gasless_allowed_token_types(protocol_config);
+        fail_point_arg!("gasless_extra_token_types", |extra: HashSet<TypeInput>| {
+            allowed_token_types.extend(extra);
+        });
```
alex-mysten: should we push the fail_point down into `parse_gasless...`

crates/sui-types/src/gas_model/gas_v2.rs
```diff
@@ -524,11 +524,28 @@ mod checked {
...
+                let storage_cost = self.storage_cost();
+                assert!(
+                    storage_cost == 0,
+                    "Gasless transaction must not incur storage cost, got {storage_cost}"
+                );
+                GasCostSummary {
+                    computation_cost: sender_rebate,
+                    storage_cost: 0,
+                    storage_rebate: sender_rebate,
+                    non_refundable_storage_fee,
```
alex-mysten: A comment would help explain why a non-zero value is acceptable

crates/sui-types/src/transaction.rs
```diff
@@ -1047,7 +1068,7 @@ fn is_move_call(
     call.package == package && call.module.as_str() == module && call.function.as_str() == function
 }
 
-fn parse_gasless_allowed_token_types(config: &ProtocolConfig) -> Vec<TypeInput> {
+pub fn parse_gasless_allowed_token_types(config: &ProtocolConfig) -> HashSet<TypeInput> {
```
alex-mysten: Changed to vec as per your suggestion :)

sui-execution/latest/sui-adapter/src/execution_engine.rs
aschran: merge issue, I put it back, ty

crates/sui-types/src/gas_model/gas_v2.rs
aschran: is the comment on line 529 not adequate? what aspect do you think could be explained more clearly?

crates/sui-types/src/transaction.rs
aschran: whoops and now i'm undoing it! lol. sorry.

crates/sui-types/src/gas_model/gas_v2.rs
alex-mysten: I meant specifically with regard to the non_refundable_storage_fee - the other costs net to 0, this one doesn't, but it's not used for gas charging

aschran:
@dariorussi @tzakian @tnowacki This is the PR that adds support for consuming coins as inputs in gasless tx. For a gasless tx, we don't want to rebate any sui to the sender. My plan for this is to consume the amount of any storage rebate as compute cost, so that it flows back to the network as end of epoch rewards. Adding you all in order to solicit your opinion/approval of the plan.

This PR is stacked on an earlier version of @alex-mysten 's base impl of gasless tx, so it'll need a lot of work to rebase once Alex's PR is merged. Please don't look too much at the details of the implementation until that is finished.

crates/sui-protocol-config/src/lib.rs
tzakian: Panicking in here seems a bit scary (or at least unnecessary). Thoughts on `unwrap_or_else(Vec::new)` here instead along with a `debug_assert!(self.gasless_allowed_token_types.is_some())`?

tzakian, line 5094: should this also unset all the other fields related to gasless?

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -673,6 +677,41 @@ mod checked {
...
+    fn check_gasless_object_inputs(
+        input_objects: &InputObjects,
+        protocol_config: &ProtocolConfig,
+    ) -> UserInputResult<()> {
+        let allowed_token_types =
+            sui_types::transaction::parse_gasless_allowed_token_types(protocol_config);
+
+        for obj_read in input_objects.iter() {
+            let Some(object) = obj_read.as_object() else {
+                continue;
```
tzakian, line 691: Should this be an error instead of a `continue`? IIRC gasless txs shouldn't be taking shared objects (so we don't need to worry about consensus stream ended etc) and so if it gets into this scenario we should just not sign the tx?

crates/sui-types/src/transaction.rs
tzakian, line 979: Nit: if these are only for testing, can these be `#[cfg(debug_assertions)]` gated?

tzakian: For extra paranoia, thoughts on something like
```rust
if config.enable_gassless() && self.is_gasless_transaction() {
  let TransactionKinde::ProgrammableTransaction(pt) = &self.kind else {
     debug_fatal!(...);
     return UserInputError::...
  };
}
```

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -408,7 +408,25 @@ pub mod checked {
...
 
             let Some(gas_payment_location) = gas_payment_location else {
-                return GasCostSummary::default(); // gasless
+                // Gasless: sender pays nothing.
+                if execution_result.is_err() {
+                    return GasCostSummary::default();
+                }
+                // Any storage rebate from destroyed input coins is absorbed as
+                // network fees, not returned to sender.
+                let cost_summary = self.gas_status.summary();
```
tzakian: Nit: can you reuse the `cost_summary` from just above this?

crates/sui-transaction-checks/src/lib.rs
aschran: what about party object coins?

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -673,6 +677,41 @@ mod checked {
...
     }
 
+    /// Verify that all Move object inputs in a gasless transaction are Coin<T>
+    /// where T is in the allowlist.
+    fn check_gasless_object_inputs(
+        input_objects: &InputObjects,
+        protocol_config: &ProtocolConfig,
+    ) -> UserInputResult<()> {
+        let allowed_token_types =
+            sui_types::transaction::parse_gasless_allowed_token_types(protocol_config);
```
tnowacki, line 687: We are already parsing this for the other checks. It could be nice to share that somehow... though not critical given how short the list is

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -673,6 +677,41 @@ mod checked {
         Ok(())
     }
 
+    /// Verify that all Move object inputs in a gasless transaction are Coin<T>
+    /// where T is in the allowlist.
+    fn check_gasless_object_inputs(
```
tnowacki, line 682: I think we should also add some logic for input checking in the check over the PTB itself too, just to make sure things like Receiving are excluded

crates/sui-transaction-checks/src/lib.rs
tnowacki: But neither party objects nor shared objects are checked at this spot? This is just seeing if it is an object?

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -673,6 +677,41 @@ mod checked {
...
+        let allowed_token_types =
+            sui_types::transaction::parse_gasless_allowed_token_types(protocol_config);
+
+        for obj_read in input_objects.iter() {
+            let Some(object) = obj_read.as_object() else {
+                continue;
+            };
+            if object.is_package() {
+                continue;
+            }
```
tnowacki, line 695: We should probably be checking the owner type here? We should disallow immutable objects, and leave a space for party objects (you will need the `Delete` permission)

crates/sui-protocol-config/src/lib.rs
aschran: cc @alex-mysten I think this comment is meant for the base PR?

aschran: cc @alex-mysten I think this comment is meant for the base PR?

crates/sui-types/src/transaction.rs
aschran: cc @alex-mysten I think this comment is meant for the base PR?

aschran: cc @alex-mysten I think this comment is meant for the base PR?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -408,7 +408,25 @@ pub mod checked {
             let cost_summary = self.gas_status.summary();
 
             let Some(gas_payment_location) = gas_payment_location else {
-                return GasCostSummary::default(); // gasless
+                // Gasless: sender pays nothing.
```
tnowacki, line 411: We should assert here that the payment method is gasless. I know that we return above for Unmetered, but it would be very bad if those somehow get merged (since the payment location is also none for unmetered.)

crates/sui-types/src/transaction.rs
```diff
@@ -1159,16 +1163,49 @@ impl ProgrammableMoveCall {
...
+        const SUI_COIN_INTO_BALANCE: FunctionIdent = (
+            SUI_FRAMEWORK_ADDRESS,
+            COIN_MODULE_NAME,
+            INTO_BALANCE_FUNC_NAME,
+        );
+        const SUI_COIN_REDEEM_FUNDS: FunctionIdent = (
+            SUI_FRAMEWORK_ADDRESS,
+            COIN_MODULE_NAME,
+            REDEEM_FUNDS_FUNC_NAME,
+        );
```
tnowacki, line 1221: This makes a coin, yeah? why do we need this one?

crates/sui-transaction-checks/src/lib.rs
aschran: Tim was saying that we don't need to worry about consensus stream ended. I was trying to point out that we maybe do, if we allow party coins as gasless inputs, and so we wouldn't want to refuse to sign the tx if we saw one of those? 

or @tnowacki what are you suggesting needs to change here?

crates/sui-transaction-checks/src/lib.rs
aschran: yea an earlier version of my PR added code to cache the parsed config, but I guess that was lost in one of the rebases

aschran: or nevermind, actually I think it's in https://github.com/MystenLabs/sui/pull/25882 - but now that is quite out of date since the base PR changed so much

crates/sui-types/src/transaction.rs
aschran: @hayes asked for it
edit: I assume it helps avoid the need for special-casing gasless tx in the sdk

crates/sui-transaction-checks/src/lib.rs
aschran: should this go here or the base PR? (in case we merge base PR but not input coin support today)

aschran: cc @alex-mysten

crates/sui-protocol-config/src/lib.rs
alex-mysten: Addressing in base pr

alex-mysten: Addressing in base pr

crates/sui-types/src/transaction.rs
alex-mysten: Addressing in base pr

alex-mysten: Addressing in base pr

crates/sui-adapter-transactional-tests/tests/gasless/gasless_coin.move
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) Mysten Labs, Inc.
```
tnowacki, line 1: Heads up, these tests have moved

aschran: this is also moved - it's now in the gasless/ dir

crates/sui-types/src/transaction.rs
```diff
@@ -999,47 +1003,6 @@ impl ProgrammableTransaction {
...
-            match input {
-                CallArg::Pure(_) | CallArg::FundsWithdrawal(_) => {}
-                _ => {
-                    return Err(UserInputError::Unsupported(format!(
-                        "Gasless transactions only support pure and withdrawal inputs (input {idx})"
-                    )));
-                }
-            }
-        }
-
```
tnowacki, line 1012: We should bring this back to exclude Receiving

crates/sui-adapter-transactional-tests/tests/gasless/gasless_coin.move
```diff
@@ -0,0 +1,75 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// Tests gasless transactions with coin object inputs, MergeCoins, and SplitCoins.
```
tnowacki, line 4: Maybe another file but it would be good to show:
- An immutable coin input
- A shared coin input (used/unused)
- A party coin input (used/unused)
- Receiving (maybe coin and noncoin)

tnowacki (approved):

PR #25860, opened 2026-03-17
[rpc-alt] Use gRPC instead of JSON-RPC for Write API in sui-indexer-alt-jsonrpc

emmazzz:
## Description 

Replace the JSON-RPC HTTP client delegation with gRPC calls for executeTransactionBlock and dryRunTransactionBlock. The Write module now uses FullnodeClient from sui-indexer-alt-reader, which communicates with the fullnode via TransactionExecutionService gRPC.

- Add configurable read_mask parameter to FullnodeClient's execute_transaction and simulate_transaction methods
- Add --fullnode-grpc-url flag (separate from existing --fullnode-rpc-url which remains for DelegationGovernance JSON-RPC delegation)
- Implement proto-to-JSON-RPC response conversion for effects, events, balance changes, and object changes
- Build dynamic FieldMask based on SuiTransactionBlockResponseOptions

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-jsonrpc/src/api/transactions/response.rs
+++ b/crates/sui-indexer-alt-jsonrpc/src/api/transactions/response.rs
@@ -263,16 +263,14 @@ async fn object_changes(
         let input = fetch_object(object_id, input_version, input_digest)?;
         let output = fetch_object(object_id, output_version, output_digest)?;
 
-        if let Some(sui_change) = to_sui_object_change(
+        changes.extend(to_sui_object_change(
             tx_data.sender(),
             object_id,
             id_operation,
             input,
             output,
             effects.lamport_version(),
-        )? {
-            changes.push(sui_change);
-        }
+        )?);
     }
 
     Ok(changes)
--- a/crates/sui-indexer-alt-jsonrpc/src/api/write/mod.rs
+++ b/crates/sui-indexer-alt-jsonrpc/src/api/write/mod.rs
@@ -3,16 +3,13 @@
 
 mod response;
 
+use anyhow::Context as _;
 use fastcrypto::encoding::Base64;
 use jsonrpsee::core::RpcResult;
 use jsonrpsee::proc_macros::rpc;
-use jsonrpsee::types::ErrorObject;
-use jsonrpsee::types::error::INTERNAL_ERROR_CODE;
-use jsonrpsee::types::error::INVALID_PARAMS_CODE;
 use prost_types::FieldMask;
 use sui_indexer_alt_reader::fullnode_client::FullnodeClient;
 use sui_json_rpc_types::DryRunTransactionBlockResponse;
-use sui_json_rpc_types::SuiTransactionBlockData;
 use sui_json_rpc_types::SuiTransactionBlockResponse;
 use sui_json_rpc_types::SuiTransactionBlockResponseOptions;
 use sui_open_rpc::Module;
@@ -22,11 +19,11 @@ use sui_rpc::proto::sui::rpc::v2 as proto;
 use sui_types::crypto::ToFromBytes;
 use sui_types::signature::GenericSignature;
 use sui_types::transaction::TransactionData;
-use sui_types::transaction::TransactionDataAPI;
 use sui_types::transaction_driver_types::ExecuteTransactionRequestType;
 
 use crate::api::rpc_module::RpcModule;
 use crate::context::Context;
+use crate::error::RpcError;
 use crate::error::invalid_params;
 
 #[open_rpc(namespace = "sui", tag = "Write API")]
@@ -66,6 +63,15 @@ pub(crate) struct Write {
 pub enum Error {
     #[error("WaitForLocalExecution mode is deprecated")]
     DeprecatedWaitForLocalExecution,
+
+    #[error("Invalid transaction bytes: {0}")]
+    InvalidTransactionBytes(String),
+
+    #[error("Invalid signature: {0}")]
+    InvalidSignature(String),
+
+    #[error("Transaction execution failed: {0}")]
+    ExecutionFailed(String),
 }
 
 impl Write {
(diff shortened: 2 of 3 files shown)


crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_write_api_tests.rs
```diff
@@ -0,0 +1,466 @@
...
+    telemetry_subscribers::init_for_testing();
+    let cluster = WriteTestCluster::new().await.unwrap();
+
+    let response = cluster
+        .execute_jsonrpc(
+            "sui_dryRunTransactionBlock",
+            json!({ "tx_bytes": "invalid_tx_bytes" }),
+        )
+        .await
+        .unwrap();
```
wlmyng, line 606: this is a nicer error than, "Invalid value was given to the function". makes me think though, are these technically breaking changes since the error string is different?

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_write_api_tests.rs
```diff
@@ -0,0 +1,466 @@
...
+
+    // The transaction either fails at execution (effects with failure status)
+    // or is rejected by the validator during simulation (error response).
+    let has_failure_effects = response["result"]["effects"]["status"]["status"] == "failure";
+    let has_error = response["error"].is_object();
+    assert!(
+        has_failure_effects || has_error,
+        "expected failure effects or error response, got: {response}"
+    );
+}
```
wlmyng: is the response["error"] path just to handle transient errors? wouldn't we expect this to always have failure effects - it's a valid txn but gets aborted in the vm because not run by validator?

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_write_api_tests.rs
```diff
@@ -0,0 +1,466 @@
...
+            "sui_executeTransactionBlock",
+            json!({
+                "tx_bytes": tx_bytes,
+                "signatures": [],
+            }),
+        )
+        .await
+        .unwrap();
+
+    assert!(response["error"]["code"].is_number());
```
wlmyng: any reason we aren't asserting on the exact error cod enumber?

crates/sui-indexer-alt-jsonrpc/src/api/write.rs
```diff
@@ -70,20 +107,187 @@ impl WriteApiServer for Write {
...
+                .effects
+                .as_ref()
+                .and_then(|e| e.bcs.as_ref())
+                .ok_or_else(|| internal_err("Missing effects.bcs in gRPC response"))?;
+
+            if options.show_raw_effects {
+                result.raw_effects = effects_bcs.value().to_vec();
+            }
+
+            let effects: TransactionEffects = effects_bcs
```
wlmyng: i think we nly need to deserialize effects_bcs if show_effects or show_object_changes

crates/sui-indexer-alt-jsonrpc/src/api/write.rs
```diff
@@ -96,3 +300,279 @@ impl RpcModule for Write {
...
+        sui_events.push(SuiEvent::try_from(
+            event, tx_digest, ix as u64, None, layout,
+        )?);
+    }
+
+    Ok(SuiTransactionBlockEvents { data: sui_events })
+}
+
+fn build_object_changes(
+    tx_data: &TransactionData,
```
wlmyng: cant we just pass the sender SuiAddress in instead of the full TransactionData?

crates/sui-indexer-alt-jsonrpc/src/api/write.rs
```diff
@@ -96,3 +300,279 @@ impl RpcModule for Write {
...
+    effects: &TransactionEffects,
+    executed_tx: &proto::ExecutedTransaction,
+) -> anyhow::Result<Vec<SuiObjectChange>> {
+    let native_changes = effects.object_changes();
+
+    // Build a map of (ObjectID, version) → Object from the proto objects
+    let mut objects: HashMap<(ObjectID, u64), Object> = HashMap::new();
+    if let Some(object_set) = &executed_tx.objects {
+        for proto_obj in &object_set.objects {
+            if let Some(bcs) = &proto_obj.bcs {
```
wlmyng: when might an obj not have a bcs? when it's Wrapped or Deleted only? Given its a proto field could we be masking when the bcs should be present but isn't due to malformed data?

crates/sui-indexer-alt-jsonrpc/src/api/write.rs
```diff
@@ -96,3 +300,279 @@ impl RpcModule for Write {
...
+                "Unexpected input version {} for object {object_id} created by transaction",
+                i.version().value(),
+            ),
+
+            (ID::Deleted, _, Some((o, _))) => anyhow::bail!(
+                "Unexpected output version {} for object {object_id} deleted by transaction",
+                o.version().value(),
+            ),
+
+            (ID::Created, _, None) => continue,
```
wlmyng: obj created and immediately wrapped?

crates/sui-indexer-alt-jsonrpc/src/api/write.rs
```diff
@@ -96,3 +300,279 @@ impl RpcModule for Write {
...
+            (ID::None, None, _) => continue,
```
wlmyng: what scenario might result in this? If we have an unwrap and an output object it seems we'd skip it?

crates/sui-indexer-alt-jsonrpc/src/api/write.rs
```diff
@@ -96,3 +300,279 @@ impl RpcModule for Write {
...
+            (ID::None, _, Some((o, _))) if o.is_package() => continue,
+            (ID::Deleted, None, _) => continue,
```
wlmyng: this is a package upgrade?

crates/sui-indexer-alt-jsonrpc/src/api/write.rs
```diff
@@ -96,3 +300,279 @@ impl RpcModule for Write {
...
+        };
+        sui_events.push(SuiEvent::try_from(
+            event, tx_digest, ix as u64, None, layout,
+        )?);
+    }
+
+    Ok(SuiTransactionBlockEvents { data: sui_events })
+}
+
+fn build_object_changes(
```
wlmyng: wait a moment, is it just me or does this look really similar to the `object_Changes` in `transactions/response.rs`? could we abstract and share the logic mapping the existence of an object in `input`, `output`, and the `IDOperation`?

crates/sui-indexer-alt-graphql/src/api/mutation.rs
amnn, line 1: If there's only one real use case of `execute_transaction` and `simulate_transaction` that rely on field masks being set how they are, rather than using the current field mask values as defaults, can we make it a non-optional parameter and have these call-sites be explicit about what they expect?

Maybe there's a potential to consolidate the field masks that we use in GraphQL between execution and simulation as a result.

crates/sui-indexer-alt-jsonrpc/src/api/write.rs
amnn, line 1: Please take a look at the structure used for producing a transaction response for `sui_getTransactionBlock` (where we have a separate `response` module and each response option has its own associated helper function), and replicate that here.

crates/sui-indexer-alt-jsonrpc/src/api/transactions/response.rs
```diff
@@ -263,94 +263,16 @@ async fn object_changes(
...
+        if let Some(sui_change) = to_sui_object_change(
+            tx_data.sender(),
+            object_id,
+            id_operation,
+            input,
+            output,
+            effects.lamport_version(),
+        )? {
+            changes.push(sui_change);
+        }
```
amnn: nit (optional)

```suggestion
        changes.extend(to_sui_object_change(
            tx_data.sender(),
            object_id,
            id_operation,
            input,
            output,
            effects.lamport_version(),
        )?)
```

crates/sui-indexer-alt-jsonrpc/src/api/write/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+        request_type: Option<ExecuteTransactionRequestType>,
+    ) -> RpcResult<SuiTransactionBlockResponse> {
+        if let Some(ExecuteTransactionRequestType::WaitForLocalExecution) = request_type {
+            return Err(invalid_params(Error::DeprecatedWaitForLocalExecution).into());
+        }
+
+        let options = options.unwrap_or_default();
+        let tx_data: TransactionData =
+            bcs::from_bytes(&tx_bytes.to_vec().map_err(invalid_params_err)?).map_err(|e| {
+                invalid_params_err(anyhow::anyhow!(
```
amnn: This is bypassing how errors are set-up in this codebase. an invalid params error  should be represented by adding a variant to the `Error` type, and calling `invalid_params` here (similar to the `DeprecatedWaitForLocalExecution` case above).

crates/sui-indexer-alt-jsonrpc/src/api/write/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+            .execute_transaction(tx_data.clone(), parsed_sigs.clone(), read_mask)
+            .await
+            .map_err(grpc_error_to_error_object)?;
+
+        let executed_tx = grpc_response
+            .transaction
+            .as_ref()
+            .ok_or_else(|| internal_err("Missing transaction in gRPC response"))?;
+
+        let mut result = SuiTransactionBlockResponse::new(tx_digest);
```
amnn: The stuff here and below belongs in a `transaction` function in the `response` module (or `response::executed` if it's different from what dry-run needs).

crates/sui-indexer-alt-jsonrpc/src/api/write/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+        result.timestamp_ms = executed_tx
+            .timestamp
+            .and_then(|ts| sui_rpc::proto::proto_to_timestamp_ms(ts).ok());
+
+        if options.show_input {
+            result.transaction = Some(
+                response::input(&self.context, tx_data.clone(), parsed_sigs.clone())
+                    .await
+                    .map_err(|e| {
+                        internal_err(format!("Failed to convert transaction data: {e}"))
```
amnn: The `internal_err` helper should be unnecessary. Just use `context` and have it return an `anyhow::Result`, which should automatically get converted into an `RpcResult` with the right error code.

crates/sui-indexer-alt-jsonrpc/src/api/write/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    if options.show_balance_changes {
+        paths.push("balance_changes");
+    }
+
+    FieldMask::from_paths(paths)
+}
+
+fn grpc_error_to_error_object(
```
amnn: Similar comment here as for `internal_err` and `invalid_params_err` -- this code should not be dealing with `ErrorObject`, it should be dealing with `RpcError<Error>` which is already set-up to handle JSON-RPC's error-codes.

The previous implementation did not use this because the underlying request was also JSONRPC and the codes all matched, but now that the underlying protocol differs, we should go through the proper translation step.

crates/sui-indexer-alt-jsonrpc/src/api/write/response.rs
```diff
@@ -0,0 +1,213 @@
...
+    tx_data: TransactionData,
+    signatures: Vec<GenericSignature>,
+) -> anyhow::Result<SuiTransactionBlock> {
+    let data =
+        SuiTransactionBlockData::try_from_with_package_resolver(tx_data, ctx.package_resolver())
+            .await?;
+    Ok(SuiTransactionBlock {
+        data,
+        tx_signatures: signatures,
+    })
```
amnn: nit

```suggestion
    tx_signatures: Vec<GenericSignature>,
) -> anyhow::Result<SuiTransactionBlock> {
    let data =
        SuiTransactionBlockData::try_from_with_package_resolver(tx_data, ctx.package_resolver())
            .await?;
    Ok(SuiTransactionBlock {
        data,
        tx_signatures,
    })
```

crates/sui-indexer-alt-jsonrpc/src/api/write/response.rs
```diff
@@ -0,0 +1,213 @@
...
+    Ok(bcs::to_bytes(tx_data)?)
+}
+
+/// Extract the raw effects BCS bytes from the gRPC response.
+pub(super) fn raw_effects(executed_tx: &proto::ExecutedTransaction) -> anyhow::Result<Vec<u8>> {
+    let effects_bcs = executed_tx
+        .effects
+        .as_ref()
+        .and_then(|e| e.bcs.as_ref())
+        .ok_or_else(|| anyhow::anyhow!("Missing effects.bcs in gRPC response"))?;
```
amnn: context

crates/sui-indexer-alt-jsonrpc/src/api/write/response.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use crate::api::object_change::to_sui_object_change;
+use crate::context::Context;
+
+/// Build a representation of the transaction's input data for the response.
+pub(super) async fn input(
+    ctx: &Context,
+    tx_data: TransactionData,
+    signatures: Vec<GenericSignature>,
+) -> anyhow::Result<SuiTransactionBlock> {
```
amnn: Return `Result<..., RpcError<Error>>` here and elsewhere in this file.

emmazzz:
Thanks @amnn for the review!
I've addressed the comments in the latest commit. The main change here is splitting the method into a thin wrapper plus an internal `*_impl` function. That gives us a cleaner boundary where the outer layer deals with `RpcResult`, and the inner implementation can just work with `RpcError` everywhere.

amnn (approved):
Perfect! Thanks @emmazzz

PR #25863, opened 2026-03-17
Implement verifySignature GraphQL query

tpham-mysten:
## Description 

- Add a new `verifySignature` query that supports all signature types (`Ed25519`, `Secp256k1`, `Secp256r1`, `MultiSig`, `ZkLogin`, `Passkey`) using local verification via `GenericSignature::verify_authenticator` from `sui-types`
- Fetch JWKs from the indexed authenticator state (not the fullnode's live object set) to maintain checkpoint consistency with other GraphQL queries
- Deprecate `verifyZkLoginSignature` to avoid breaking existing SDKs before GraphQL GA

## Test plan 

- `cargo nextest run -p sui-indexer-alt-graphql`
- Manually tested with Ed25519 PersonalMessage against local GraphQL server + testnet fullnode gRPC

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL:  Add new verifySignature
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/snapshots/sui_indexer_alt_graphql__api__types__available_range__field_piplines_tests__registry_collect_pipelines_snapshot_staging.snap
+++ b/crates/sui-indexer-alt-graphql/src/api/types/snapshots/sui_indexer_alt_graphql__api__types__available_range__field_piplines_tests__registry_collect_pipelines_snapshot_staging.snap
@@ -1490,6 +1490,9 @@ Query.type
 Query.simulateTransaction
   => {}
 
+Query.verifySignature
+  => {}
+
 Query.verifyZkLoginSignature
   => {}
 
@@ -1617,6 +1620,9 @@ SharedInput.initialSharedVersion
   => {}
 
 SharedInput.mutable
+  => {}
+
+SignatureVerifyResult.success
   => {}
 
 SimulationResult.effects


tpham-mysten:
Yeah sounds good and sorry for getting late back to this PR.

I have implemented GraphQL's logic for `verifySignature` using `GenericSignature::verify_authenticator`.

crates/sui-indexer-alt-graphql/src/api/query.rs
```diff
@@ -796,6 +799,43 @@ impl Query {
...
+                    ctx,
+                    self.scope(ctx)?,
+                    message,
+                    signature,
+                    intent_scope,
+                    author,
+                )
+                .await
+            }
+            .await,
```
amnn, line 848: nit, do you need the outer `async { ... }.await`?

crates/sui-indexer-alt-graphql/src/api/types/signature_verify.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    let sig = GenericSignature::from_bytes(&signature)
+        .map_err(|_| bad_user_input(Error::BadSignature))?;
+
+    let epoch = Epoch::fetch(ctx, scope.clone(), None)
+        .await
+        .map_err(upcast)?
+        .context("Failed to fetch current Epoch")?;
+
+    let jwks = fetch_jwks(ctx, scope).await.map_err(upcast)?;
```
amnn, line 131: It's interesting that we need to fetch this information unconditionally (not just if we are dealing with zklogin signatures).

crates/sui-indexer-alt-e2e-tests/tests/graphql_verify_signature_tests.rs
amnn, line 1: Could we also add a test for the deprecated path, just to understand how that deprecation shows up for clients?

crates/sui-indexer-alt-graphql/src/api/query.rs
tpham-mysten: yes, it's needed to allow ? on self.scope(ctx).

We can handle that by

  let scope = match self.scope(ctx) {
      Ok(s) => s,
      Err(e) => return Some(Err(e)),
  };

but I think this async {...} await is not bad tho

crates/sui-indexer-alt-graphql/src/api/types/signature_verify.rs
tpham-mysten: Yeah. JWKs are only needed for ZkLogin and Passkey signatures (and potentially MultiSig containing them). However, since this is a single indexed DB read on a per-query path, I think the simplicity of fetching unconditionally outweighs the optimization. We could improve this logic if there was a performance concern.

crates/sui-indexer-alt-e2e-tests/tests/graphql_verify_signature_tests.rs
tpham-mysten: We already have that in crates/sui-indexer-alt-e2e-tests/tests/graphql_zklogin_tests.rs

PR #25865, opened 2026-03-17
[move][treesitter] Refactor and update Move treesitter implementation

tzakian:
## Description 

Updated and cleaned up the tree-sitter implementation for Move and added additional testing to catch any parse drift in the testing. 

I verified it against the prettier-move tests, but not sure if there are others I should test against before we call this "OK"?

## Test plan 

CI + updated `run-tests.sh` script.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


damirka (approved):
LGTM! Can you please build wasm and place it under `../prettier-move`, so we make sure the tests pass as well?

damirka:
Also just realised we never added `prettier-move` tests to CI, so I'll do it in a follow up!

PR #25866, opened 2026-03-17
Count gasless tx for DoS protection

aschran:
## Description 

Gasless tx should be counted in traffic controller since they have no cost to sender.

## Test plan 

Added test

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/traffic_control_tests.rs
+++ b/crates/sui-e2e-tests/tests/traffic_control_tests.rs
@@ -9,6 +9,7 @@
 use core::panic;
 use fastcrypto::encoding::Base64;
 use jsonrpsee::{core::client::ClientT, rpc_params};
+use move_core_types::identifier::Identifier;
 use std::fs::File;
 use std::num::NonZeroUsize;
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


crates/sui-e2e-tests/tests/traffic_control_tests.rs
```diff
@@ -976,3 +985,95 @@ async fn assert_validator_traffic_control_dry_run(
     }
     Ok(())
 }
+
+#[tokio::test]
```
alex-mysten, line 989: not a simtest?

crates/sui-e2e-tests/tests/traffic_control_tests.rs
```diff
@@ -976,3 +985,95 @@ async fn assert_validator_traffic_control_dry_run(
...
+        vec![withdraw_arg],
+    );
+    let recipient_arg = builder.pure(recipient).unwrap();
+    builder.programmable_move_call(
+        SUI_FRAMEWORK_PACKAGE_ID,
+        Identifier::new("balance").unwrap(),
+        Identifier::new("send_funds").unwrap(),
+        vec![GAS::type_tag()],
+        vec![balance, recipient_arg],
+    );
```
alex-mysten, line 1046: Do we have helpers for constructing transactions we could reuse here?

aschran: None that I see, can't easily use `TestEnv::create_gasless_transaction` without setting up a full `TestEnv`

crates/sui-e2e-tests/tests/traffic_control_tests.rs
aschran: top-of-file comment -
```
//! NB: Most tests in this module expect real network connections and interactions, thus
//! they should nearly all be tokio::test rather than simtest.
```

PR #25867, opened 2026-03-17
[core] enable defer unpaid amplification 

johnjmartin:
## Description 

  - Fix bench_driver amplification to use fire-and-forget submissions
  - Enable `defer_unpaid_amplification` in protocol version 120 (for devnet and testnet only)
 
## Test plan 

  - [x] Validated on private-testnet with 5% amplification rate
  - [x] TPS stable at ~6000, p99 latency ~1s
  - [x] Deferrals metric (`consensus_handler_unpaid_amplification_deferrals`) confirmed active
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/src/lib.rs
+++ b/crates/sui-benchmark/src/lib.rs
@@ -489,8 +489,8 @@ impl LocalValidatorAggregatorProxy {
         let validators: Vec<_> = self
             .clients
             .values()
-            .cloned()
             .take(num_validators)
+            .cloned()
             .collect();
         let request = SubmitTxRequest::new_transaction(tx.clone());
 
@@ -1182,4 +1182,3 @@ pub fn convert_move_call_args(
         })
         .collect()
 }
-
--- a/crates/sui-protocol-config/src/lib.rs
+++ b/crates/sui-protocol-config/src/lib.rs
@@ -310,7 +310,7 @@ const TESTNET_USDC: &str =
 // Version 117: Update Sui System metadata handling.
 // Version 118: Adds `transfer_migration_cap` to display registry
 // Version 119: Enable the new VM.
-//              Re-enable defer_unpaid_amplification (devnet + testnet).
+// Version 120: Re-enable defer_unpaid_amplification (devnet + testnet).
 
 #[derive(Copy, Clone, Debug, Hash, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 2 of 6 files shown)


crates/sui-benchmark/src/lib.rs
```diff
@@ -486,19 +486,22 @@ impl LocalValidatorAggregatorProxy {
...
+        let request = SubmitTxRequest::new_transaction(tx.clone());
 
-        // Fire off all submissions but don't wait for responses
-        let _ = futures::future::join_all(futures).await;
+        // Fire off all submissions as spawned tasks - don't wait for responses.
+        // We just need to ensure the submissions are sent to trigger amplification.
+        for client in validators {
+            let req = request.clone();
+            tokio::spawn(async move {
+                let _ = client.submit_transaction(req, None).await;
```
mystenmark: does submit_transaction have a built-in timeout? if not we might want to add one to make sure that we can't pile up an unbounded number of tasks

crates/sui-benchmark/src/lib.rs
```diff
@@ -1174,3 +1177,32 @@ pub fn convert_move_call_args(
...
+        for i in 0..5 {
+            tokio::spawn(async move {
+                tokio::time::sleep(Duration::from_secs(10)).await;
+                i
+            });
+        }
+
+        let elapsed = start.elapsed();
+
+        // Spawning should complete almost instantly, not wait for the tasks
```
mystenmark: i'm not sure this test is necessary, its just testing basic tokio functionality that can't possibly break (without breaking literally everything in the world)

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4688,6 +4689,10 @@ impl ProtocolConfig {
                     // Enable new VM.
                     cfg.execution_version = Some(4);
                     cfg.feature_flags.new_vm_enabled = true;
+                    // Re-enable unpaid amplification deferral protection (devnet first)
+                    if chain != Chain::Mainnet && chain != Chain::Testnet {
+                        cfg.feature_flags.defer_unpaid_amplification = true;
```
mystenmark, line 4775: personally i think we could go to testnet right now, we're not going to get any signal on this from devnet

PR #25868, opened 2026-03-17
[sui-execution][chore] Simplify after execution cut

tnowacki:
## Description 

- Inlined several `protocol-config` feature flags

## Test plan 

- 👀 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-move-natives/src/validator.rs
+++ b/sui-execution/latest/sui-move-natives/src/validator.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{NativesCostTable, get_extension, object_runtime::ObjectRuntime};
+use crate::{NativesCostTable, get_extension};
 use move_binary_format::errors::{PartialVMError, PartialVMResult};
 use move_core_types::{gas_algebra::InternalGas, vm_status::StatusCode};
 use move_vm_runtime::pop_arg;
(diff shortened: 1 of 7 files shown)


sui-execution/latest/sui-move-natives/src/validator.rs
```diff
@@ -59,9 +59,7 @@ pub fn validate_metadata_bcs(
 
     let cost = context.gas_used();
 
-    let obj_runtime: &ObjectRuntime = context.extensions().get().unwrap();
-    let metadata_v2 = obj_runtime.protocol_config.validator_metadata_verify_v2();
-    if let Result::Err(err_code) = validator_metadata.verify(metadata_v2) {
+    if let Result::Err(err_code) = validator_metadata.verify(true) {
```
tnowacki, line 62: Not sure how we feel about this one

tzakian: I think we can now just remove the flag completely in an unprotocol-gated manner across the codebase I believe as this was only needed during roll-out of the flag change. But we should do that in a separate PR probably... (/happy to discuss offline)

tzakian (approved):
Nice! 🎉

sui-execution/latest/sui-move-natives/src/validator.rs
tnowacki: Yeah let's worry about any deeper changes in another PR. Just trying to keep this as a simple "inline if true" kinda feature flag change

PR #25875, opened 2026-03-18
sui-indexer-alt-framework: verify indexer uses a single chain_id

evan-wall-mysten:
## Description 

This PR is a follow up to #25895 and #25905 that 
1. stores the `chain_id` per pipeline in the watermark
2. compares the stored `chain_id` to the `chain_id` returned by the streaming and ingestion clients
3. implements the `chain_id` storage for the Postgres and mock store implementations (other impls return the passed in `chain_id` and are no-ops)

## Test plan 

Adds new `test_chain_id` test cases.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Add `Connection::init_chain_id` method to store and retrieve `chain_id`.

### DIFF
--- a/crates/sui-indexer-alt-framework-store-traits/src/lib.rs
+++ b/crates/sui-indexer-alt-framework-store-traits/src/lib.rs
@@ -21,6 +21,17 @@ pub trait Connection: Send {
         init_watermark: InitWatermark,
     ) -> anyhow::Result<InitWatermark>;
 
+    /// Checks if the store can accept a `chain_id`.
+    /// Returns `Ok(true)` if the store accepts this `chain_id` thereby allowing processing to continue.
+    /// Returns `Ok(false)` if the store does not accept the `chain_id` thereby halting processing with an error.
+    /// Returns `Err(_)` if the store encountered an error while trying to determine if it could accept
+    /// the `chain_id` which will cause `accepts_chain_id` to be retried.
+    async fn accepts_chain_id(
+        &mut self,
+        pipeline_task: &str,
+        chain_id: [u8; 32],
+    ) -> anyhow::Result<bool>;
+
     /// Given a `pipeline_task` representing either a pipeline name or a pipeline with an associated
     /// task (formatted as `{pipeline}{Store::DELIMITER}{task}`), return the committer watermark
     /// from the `Store`. The indexer fetches this value for each pipeline added to determine which
--- a/crates/sui-indexer-alt-framework/src/mocks/store.rs
+++ b/crates/sui-indexer-alt-framework/src/mocks/store.rs
@@ -35,6 +35,7 @@ pub struct MockWatermark {
     pub reader_lo: u64,
     pub pruner_timestamp: u64,
     pub pruner_hi: u64,
+    pub chain_id: Option<[u8; 32]>,
 }
 
 /// Configuration for simulating connection failures in tests
@@ -120,12 +121,32 @@ impl Connection for MockConnection<'_> {
                 reader_lo: init_watermark.reader_lo,
                 pruner_timestamp: 0,
                 pruner_hi: init_watermark.reader_lo,
+                chain_id: None,
             });
 
         Ok(InitWatermark {
             checkpoint_hi_inclusive: watermark.checkpoint_hi_inclusive,
             reader_lo: watermark.reader_lo,
         })
+    }
+
+    async fn accepts_chain_id(
+        &mut self,
+        pipeline_task: &str,
+        chain_id: [u8; 32],
+    ) -> anyhow::Result<bool> {
+        let mut wm = self
+            .0
+            .watermarks
+            .entry(pipeline_task.to_string())
+            .or_default();
+
+        if let Some(stored_chain_id) = wm.chain_id {
+            Ok(stored_chain_id == chain_id)
+        } else {
+            wm.chain_id = Some(chain_id);
+            Ok(true)
+        }
     }
 
     async fn committer_watermark(
--- a/crates/sui-indexer-alt-framework/src/pipeline/processor.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/processor.rs
@@ -4,18 +4,21 @@
 use std::sync::Arc;
 use std::time::Duration;
 
+use async_trait::async_trait;
 use backoff::ExponentialBackoff;
 use sui_futures::service::Service;
 use sui_futures::stream::Break;
 use sui_futures::stream::TrySpawnStreamExt;
+use sui_indexer_alt_framework_store_traits::Connection;
+use sui_indexer_alt_framework_store_traits::Store;
+use sui_types::digests::ChainIdentifier;
 use sui_types::full_checkpoint_content::Checkpoint;
+use tokio::sync::OnceCell;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tracing::debug;
 use tracing::error;
 use tracing::info;
-
-use async_trait::async_trait;
 
 use crate::config::ConcurrencyConfig;
 use crate::ingestion::ingestion_client::CheckpointEnvelope;
@@ -60,12 +63,13 @@ pub trait Processor: Send + Sync + 'static {
 ///
 /// Each worker processes a checkpoint into rows and sends them on to the committer using the `tx`
 /// channel.
-pub(super) fn processor<P: Processor>(
+pub(super) fn processor<P: Processor, S: Store>(
     processor: Arc<P>,
     rx: mpsc::Receiver<Arc<CheckpointEnvelope>>,
     tx: mpsc::Sender<IndexedCheckpoint<P>>,
     metrics: Arc<IndexerMetrics>,
     concurrency: ConcurrencyConfig,
+    store: S,
 ) -> Service {
     Service::new().spawn_aborting(async move {
         info!(pipeline = P::NAME, "Starting processor");
@@ -76,6 +80,8 @@ pub(super) fn processor<P: Processor>(
         );
 
         let report_metrics = metrics.clone();
+        // Caches this pipeline's chain_id after the first time it has been read from the database.
+        let chain_id_accepted: Arc<OnceCell<(bool, ChainIdentifier)>> = Arc::default();
         match ReceiverStream::new(rx)
             .try_for_each_send_spawned(
                 concurrency.into(),
@@ -83,8 +89,11 @@ pub(super) fn processor<P: Processor>(
                     let metrics = metrics.clone();
                     let checkpoint_lag_reporter = checkpoint_lag_reporter.clone();
                     let processor = processor.clone();
+                    let store = store.clone();
+                    let chain_id_accepted = chain_id_accepted.clone();
 
                     async move {
+
                         metrics
                             .total_handler_checkpoints_received
                             .with_label_values(&[P::NAME])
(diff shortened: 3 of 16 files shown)


crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -21,6 +21,12 @@ pub trait Connection: Send {
         init_watermark: InitWatermark,
     ) -> anyhow::Result<InitWatermark>;
 
+    async fn init_chain_id(
```
amnn: Docs please

crates/sui-indexer-alt-framework/src/mocks/store.rs
```diff
@@ -128,6 +130,25 @@ impl Connection for MockConnection<'_> {
...
+        pipeline_task: &str,
+        chain_id: [u8; 32],
+    ) -> anyhow::Result<[u8; 32]> {
+        let mut wm = self
+            .0
+            .watermarks
+            .entry(pipeline_task.to_string())
+            .or_default();
+
+        if let Some(stored) = wm.chain_id {
```
amnn: If there is already a stored chain ID and it doesn't match -- should this produce an error? Ignoring a mismatch seems like a mistake.

crates/sui-indexer-alt-framework/src/pipeline/sequential/mod.rs
```diff
@@ -162,6 +162,7 @@ pub(crate) fn pipeline<H: Handler + Send + Sync + 'static>(
         processor_tx,
         metrics.clone(),
         concurrency,
+        db.clone(),
```
amnn: This is not from your change, but we should rename `db` to `store`.

crates/sui-pg-db/src/store.rs
```diff
@@ -65,6 +70,23 @@ impl store::Connection for Connection<'_> {
...
+        let chain_id: Option<Vec<u8>> = diesel::update(watermarks::table)
+            .filter(watermarks::pipeline.eq(pipeline_task))
+            // "coalesce" only updates the value if it is null in the existing row
+            .set(watermarks::chain_id.eq(coalesce(watermarks::chain_id, chain_id)))
+            .returning(watermarks::chain_id)
+            .get_result(self)
+            .await?;
+
+        let chain_id: Vec<u8> = chain_id.expect("missing chain id after update");
+        Ok(chain_id.try_into().expect("chain id has wrong length"))
```
amnn: `expect` needs to be accompanied by a `SAFETY` comment. The first one is safe because of the semantics of the SQL statement, but the second `expect` is not safe because it can depends on the contents of the database which we don't have full control over.

(I would personally just use `context` for both of these).

```suggestion
        // SAFETY: ... 
        let chain_id: Vec<u8> = chain_id.expect("missing chain id after update");
        chain_id.try_into().context("invalid chain id")
```

crates/sui-indexer-alt-framework/src/pipeline/processor.rs
```diff
@@ -76,15 +79,21 @@ pub(super) fn processor<P: Processor>(
...
                 concurrency.into(),
                 |checkpoint_envelope| {
                     let metrics = metrics.clone();
                     let checkpoint_lag_reporter = checkpoint_lag_reporter.clone();
                     let processor = processor.clone();
+                    let store = store.clone();
+                    let stored_chain_id = stored_chain_id.clone();
 
                     async move {
+                        let chain_id = *checkpoint_envelope.chain_id.as_bytes();
```
amnn: nit: This assignment is kind of hanging out here at the top -- why not put it next to `checkpoint`, `checkpoint_sequence_number`, etc?

crates/sui-indexer-alt-framework/src/mocks/store.rs
amnn: From looking at the usage site, the interface I'd prefer is:

```rust
/// Ensure `pipeline_task` is associated with checkpoints from the network with 
/// chain ID `chain_id`. Returns `Ok(true)` if the check succeeded and the
/// pipeline accepts data from the specified network, `Ok(false)` if the check
/// succeeded but the pipeline is already associated with another network and 
/// `Err(_)` if there was an issue completing the check.
async fn accepts_chain_id(&mut self, pipeline_task: &str, chain_id: [u8; 32]) -> anyhow::Result<bool>;
```

This avoids implementations ignoring a mismatch, and this contract is something that we can implement for stores that don't yet support chain ID verification without leaking the abstraction that the chain ID that gets returned is going to be compared with the chain ID that was passed in for a consistency check (they would just return `Ok(true)` without checking the provided `chain_id`).

evan-wall-mysten: IMO, the issue with that is that it requires that the `Store::Connection` contains business logic to validate the value being passed in which is inconsistent with the other trait methods that require the impl to set/get what is passed in. Additionally, `anyhow::Result` would need to be a custom type that can differentiate between transient errors like being unable to connect to the DB and permanent errors like a `chain_id` mismatch.

I think a compromise between the two is to change the return type to `anyhow::Result<Option<[u8; 32]>>` where
* `Ok(None)` -> framework does not validate result
* `Ok(Some(value))` -> framework validates `value` - possibly failing permanently
* `Err(e)` -> framework retries - assuming transient failure

crates/sui-pg-db/src/store.rs
evan-wall-mysten: Good point, switched both to `context`.

crates/sui-indexer-alt-framework/src/mocks/store.rs
amnn: > it requires that the `Store::Connection` contains business logic to validate the value being passed in which is inconsistent with the other trait methods that require the impl to set/get what is passed in.

The fact that existing APIs do not validate the value being passed to them is incidental, because in all cases, the value can be amended by the store to become valid, and the amended value is returned.

The issue with `set_chain_id` is similar to the issue I raised with `init_watermark`'s interface forcing stores that do not support `reader_lo` returning back the value that they was passed in to them -- it gives the impression that this was the value that was set when in fact it was a value that was ignored.

> Additionally, `anyhow::Result` would need to be a custom type that can differentiate between transient errors like being unable to connect to the DB and permanent errors like a `chain_id` mismatch.

This is not the case -- please take another look at the suggestion, in particular the doc comment states that `accepts_chain_id` returns:

- `Ok(true)` if the check succeeded and the pipeline accepts data from the specified network, 
- `Ok(false)` if the check succeeded but the pipeline is already associated with another network and 
- `Err(_)` if there was an issue completing the check (transient error returned using `anyhow::Error`)

> I think a compromise between the two is to change the return type to `anyhow::Result<Option<[u8; 32]>>` where

The issue I'm trying to address with this suggestion is that `set_chain_id` offers too many degrees of freedom. The compromise suggestion introduces ...another degree of freedom. Maybe I am not articulating the issues I am seeing with `set_chain_id` well enough, let me try again:

- The name `set_chain_id` implies an `Ok(_)` response indicates the chain ID supplied was set, but this is not true because if the chain ID was already set to something else, the convention is that the parameter is ignored and the already set value is returned.
- If I'm implementing a store that supports chain ID verification I'm passed an id `I` (the one I'm supposed to set) and I load another id `J`. I have all the data I need to perform the check, and the name of the function suggests that I should perform the check (to know whether `I` was actually set), but the convention implied by usage is that I pass `J` up to be checked by the caller. This is a leaky abstraction.
- If I'm implementing a store that does not support chain ID verification, then I similarly need to know to pass the value I did not set back up to the caller because the caller will check that this matches what it was given, and I don't want to break processing.

If instead the interface was "the store/connection will tell me if this pipeline supports data from this chain", then the contract is clear in both cases.

evan-wall-mysten: I changed this to your suggestion and am mostly happy with it. I think the downsides are minor, but maybe you have an idea how to address them
1. `OnceCell::get_or_try_init` changed to a more complex `RWLock<HashMap>` read lock then optional write lock operation. In the happy path, it is only calls the database once and only acquires a write lock on the initial attempts that occur before the DB results are stored.
2. Error message is missing the original stored chain_id - not that important, but it would make debugging easier since you could do it through the logs instead of needing to connect to the db or look for multiple log statements.
3. The logic to compare the `chain_id` with the `stored_chain_id` is duplicated across store impls - not that big of a deal since it is only a few lines.

crates/sui-indexer-alt-framework/src/pipeline/processor.rs
```diff
@@ -104,14 +114,42 @@ pub(super) fn processor<P: Processor>(
...
 
-                        let checkpoint = checkpoint_envelope.checkpoint.clone();
+                        let chain_id = *checkpoint_envelope.chain_id.as_bytes();
+                        let checkpoint = &checkpoint_envelope.checkpoint;
                         let checkpoint_sequence_number = checkpoint.summary.sequence_number;
                         let retry_metrics = metrics.clone();
                         let values = backoff::future::retry_notify(
                             backoff,
                             || async {
+                                let chain_identifier = checkpoint_envelope.chain_id;
```
amnn: We should be able to use `OnceCell` here, still, by having the `OnceCell` hold a `(bool, ChainIdentifier)`, so that in initialisation we return the output from `accepts_chain_id` and the chain identifier that was passed in initially.

If the return from `accepts_chain_id` is `false` we would fail immediately, but if not, we need to perform a second check that the returned cached chain ID matches the chain ID from the current envelope.

amnn: Also it may be worth factoring this out into a helper, in case it's still large even after switching back to `OnceCell`.

crates/sui-indexer-alt-framework/src/mocks/store.rs
evan-wall-mysten: Talked about this in meeting will address this https://github.com/MystenLabs/sui/pull/25875#discussion_r2989961971.

crates/sui-indexer-alt-framework/src/pipeline/processor.rs
```diff
@@ -104,14 +113,28 @@ pub(super) fn processor<P: Processor>(
...
                             backoff,
                             || async {
+                                let (accepted, accepted_chain_id) = chain_id_accepted.get_or_try_init(async || {
+                                    let mut conn = store.connect().await
+                                        .map_err(backoff::Error::transient)?;
+                                    let accepted = conn.accepts_chain_id(P::NAME, *chain_id.as_bytes()).await
+                                        .map_err(backoff::Error::transient)?;
+                                    Ok::<_, backoff::Error<anyhow::Error>>((accepted, chain_id))
+                                }).await?;
+                                if *accepted && *accepted_chain_id == chain_id {
```
amnn: This seems the wrong way around, no?

```suggestion
                                if !*accepted || *accepted_chain_id != chain_id {
```

evan-wall-mysten: Yeah it was causing the tests to fail.

crates/sui-indexer-alt-framework/src/pipeline/processor.rs
evan-wall-mysten: I switched to `OnceCell<(bool, ChainIdentifier)>`, but didn't end up making a helper because it had 4 parameters which when combined with the async/result wrapping/unwrapping, ended up taking about as much space as the helper itself.

PR #25881, opened 2026-03-18
sui-indexer-alt-framework: split Connection and Store traits

evan-wall-mysten:
## Description 

This PR splits `Connection` into 
* `Connection`
   * `init_watermark` - accepts `checkpoint_hi_inclusive`, returns `reader_lo` only if relevant to pipeline
   * `accepts_chain_id` - same as before
   * `committer_watermark` - same as before
   * `set_committer_watermark` - same as before
* `ConcurrentConnection: Connection`
   * `delegate_to_reader_watermark` - helper to call from impl if it does not write
   * `reader_watermark` - same as before
   * `pruner_watermark` - same as before
   * `set_reader_watermark` - same as before
   * `set_pruner_watermark` - same as before
* `SequentialConnection: Connection`
   * `delegate_to_committer_watermark` - helper to call from impl if it does not write

---

This PR splits `Store` into
* `Store`
   * `connect` - same as before
* `ConcurrentStore: Store`
   * `DELIMITER` - same as before
* `SequentialStore: Store`
   * `transaction` - same as before

---

`ConcurrentConnection` impls:
* `BigTableConnection`
* `ObjectStoreConnection`
* Postgres `Connection`

`SequentialConnection` impls:
* `AnalyticsConnection`
* Consistent store `Connection`
* Postgres `Connection`

## Test plan 

No new functionality - existing unit tests and strengthened compiler checks.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Refactors `Connection` trait into `ConcurrentConnection`/`SequentialConnection` subtraits and `Store` trait into `ConcurrentStore`/`SequentialStore` subtraits.

### DIFF
--- a/crates/sui-indexer-alt-framework-store-traits/src/lib.rs
+++ b/crates/sui-indexer-alt-framework-store-traits/src/lib.rs
@@ -13,6 +13,16 @@ use scoped_futures::ScopedBoxFuture;
 /// operations, agnostic of the underlying store implementation.
 #[async_trait]
 pub trait Connection: Send {
+    /// Initializes a watermark by either returning the existing watermark or, if the impl supports it, attempting to create it.
+    /// Returns `Ok(Some(_))` if a watermark existed or was created by this impl.
+    /// Returns `Ok(None)` if a watermark does not exist and this impl does not attempt to create one.
+    /// Returns `Err(_)` if the store encountered an error while trying to read or create the watermark.
+    async fn init_watermark(
+        &mut self,
+        pipeline_task: &str,
+        checkpoint_hi_inclusive: Option<u64>,
+    ) -> anyhow::Result<Option<InitWatermark>>;
+
     /// Checks if the store can accept a `chain_id`.
     /// Returns `Ok(true)` if the store accepts this `chain_id` thereby allowing processing to continue.
     /// Returns `Ok(false)` if the store does not accept the `chain_id` thereby halting processing with an error.
@@ -47,31 +57,25 @@ pub trait Connection: Send {
 /// Extension of [`Connection`] for concurrent pipeline watermark operations.
 #[async_trait]
 pub trait ConcurrentConnection: Connection {
-    /// Returns the `InitWatermark` based on the existing watermark if it exists.
-    /// Otherwise, initializes a new watermark record and returns the default based on `reader_lo`.
-    ///
-    /// The default implementation delegates to `Connection::committer_watermark`.
-    async fn init_concurrent_watermark(
+    /// Helper to call in [`Connection::init_watermark`] impl that delegates to
+    /// [`ConcurrentConnection::reader_watermark`] if impl does not attempt to write data.
+    async fn delegate_to_reader_watermark(
         &mut self,
         pipeline_task: &str,
-        reader_lo: u64,
-    ) -> anyhow::Result<InitWatermark> {
-        let checkpoint_hi_inclusive = self
-            .committer_watermark(pipeline_task)
+    ) -> anyhow::Result<Option<InitWatermark>> {
+        Ok(self
+            .reader_watermark(pipeline_task)
             .await?
-            .map(|w| w.checkpoint_hi_inclusive);
-        Ok(InitWatermark {
-            checkpoint_hi_inclusive,
-            reader_lo,
-        })
+            .map(|w| InitWatermark {
+                checkpoint_hi_inclusive: Some(w.checkpoint_hi_inclusive),
+                reader_lo: Some(w.reader_lo),
+            }))
     }
 
     /// Given a pipeline, return the reader watermark from the database. This is used by the indexer
     /// to determine the new `reader_lo` or inclusive lower bound of available data.
-    async fn reader_watermark(
-        &mut self,
-        pipeline: &'static str,
-    ) -> anyhow::Result<Option<ReaderWatermark>>;
+    async fn reader_watermark(&mut self, pipeline: &str)
+    -> anyhow::Result<Option<ReaderWatermark>>;
 
     /// Get the bounds for the region that the pruner is allowed to prune, and the time in
     /// milliseconds the pruner must wait before it can begin pruning data for the given `pipeline`.
@@ -115,19 +119,20 @@ pub trait ConcurrentConnection: Connection {
 /// Extension of [`Connection`] for sequential pipeline operations.
 #[async_trait]
 pub trait SequentialConnection: Connection {
-    /// Returns the `checkpoint_hi_inclusive` based on the existing watermark if it exists.
-    /// Otherwise, initializes a new watermark record and returns the value passed in.
-    ///
-    /// The default implementation delegates to `Connection::committer_watermark`.
-    async fn init_sequential_watermark(
+    /// Helper to call in [`Connection::init_watermark`] impl that delegates to
+    /// [`Connection::committer_watermark`] if impl does not attempt to write data.
+    async fn delegate_to_committer_watermark(
         &mut self,
         pipeline_task: &str,
         _checkpoint_hi_inclusive: Option<u64>,
-    ) -> anyhow::Result<Option<u64>> {
+    ) -> anyhow::Result<Option<InitWatermark>> {
         Ok(self
             .committer_watermark(pipeline_task)
             .await?
-            .map(|w| w.checkpoint_hi_inclusive))
+            .map(|w| InitWatermark {
+                checkpoint_hi_inclusive: Some(w.checkpoint_hi_inclusive),
+                reader_lo: None,
+            }))
     }
 }
 
@@ -155,13 +160,6 @@ pub trait ConcurrentStore: for<'c> Store<Connection<'c> = Self::ConcurrentConnec
     /// Delimiter used to separate pipeline names from task identifiers when reading or writing the
     /// committer watermark.
     const DELIMITER: &'static str = "@";
-
-    async fn concurrent_connect<'c>(&'c self) -> anyhow::Result<Self::ConcurrentConnection<'c>> {
-        Ok(self
-            .connect()
-            .await
-            .context("Failed to establish concurrent connection to store")?)
-    }
 }
 
 /// Extension of [`Store`] for stores that support sequential pipeline operations, including
@@ -188,13 +186,16 @@ pub trait SequentialStore: for<'c> Store<Connection<'c> = Self::SequentialConnec
         ) -> ScopedBoxFuture<'a, 'r, anyhow::Result<R>>;
 }
 
-/// Used during watermark initialization to set and return state.
+/// Ingested checkpoint range during pipeline initialization.
 #[derive(Default, Debug, Clone, Copy, PartialEq)]
 pub struct InitWatermark {
-    /// Calculated by the framework as `default_next_checkpoint.checked_sub(1)`.
+    /// Some -> checkpoint_hi_inclusive can be represented as u64
+    /// None -> checkpoint_hi_inclusive cannot be represented as u64. This occurs during pipeline
+    /// initialization if `reader_lo == 0 && checkpoint_hi_inclusive < reader_lo`
     pub checkpoint_hi_inclusive: Option<u64>,
-    /// Calculated by the framework as `default_next_checkpoint`.
-    pub reader_lo: u64,
+    /// Some -> pipeline has a trailing edge checkpoint (concurrent pipeline)
+    /// None -> pipeline does not have a trailing edge checkpoint (sequential pipeline)
+    pub reader_lo: Option<u64>,
 }
 
 /// Represents the highest checkpoint for some pipeline that has been processed by the indexer
--- a/crates/sui-indexer-alt-framework/src/lib.rs
+++ b/crates/sui-indexer-alt-framework/src/lib.rs
@@ -15,10 +15,8 @@ use ingestion::IngestionService;
 use ingestion::ingestion_client::IngestionClient;
 use metrics::IndexerMetrics;
 use prometheus::Registry;
-use sui_indexer_alt_framework_store_traits::ConcurrentConnection;
 use sui_indexer_alt_framework_store_traits::ConcurrentStore;
-use sui_indexer_alt_framework_store_traits::InitWatermark;
-use sui_indexer_alt_framework_store_traits::SequentialConnection;
+use sui_indexer_alt_framework_store_traits::Connection;
 use sui_indexer_alt_framework_store_traits::SequentialStore;
 use sui_indexer_alt_framework_store_traits::Store;
 use sui_indexer_alt_framework_store_traits::pipeline_task;
@@ -322,12 +320,10 @@ impl<S: Store> Indexer<S> {
     /// calculated above.
     ///
     /// Returns `Ok(None)` if the pipeline is disabled.
-    async fn add_pipeline<'a, P, F>(&mut self, init: F) -> Result<Option<u64>>
-    where
-        P: Processor + 'static,
-        F: Send + 'a,
-        F: AsyncFnOnce() -> Result<u64>,
-    {
+    async fn add_pipeline<P: Processor + 'static>(
+        &mut self,
+        pipeline_task: String,
+    ) -> Result<Option<u64>> {
         ensure!(
             self.added_pipelines.insert(P::NAME),
             "Pipeline {:?} already added",
@@ -341,7 +337,25 @@ impl<S: Store> Indexer<S> {
             return Ok(None);
         }
 
-        let next_checkpoint = init().await?;
+        // Create a new record based on `proposed_next_checkpoint` if one does not exist.
+        // Otherwise, use the existing record and disregard the proposed value.
+        let proposed_next_checkpoint = self.first_checkpoint.unwrap_or(0);
+        let mut conn = self.store.connect().await?;
+        let init_watermark = conn
+            .init_watermark(&pipeline_task, proposed_next_checkpoint.checked_sub(1))
+            .await
+            .with_context(|| format!("Failed to init watermark for {pipeline_task}"))?;
+
+        let next_checkpoint = if let Some(init_watermark) = init_watermark {
+            if let Some(checkpoint_hi_inclusive) = init_watermark.checkpoint_hi_inclusive {
+                checkpoint_hi_inclusive + 1
+            } else {
+                0
+            }
+        } else {
+            proposed_next_checkpoint
+        };
+
         self.next_checkpoint = next_checkpoint.min(self.next_checkpoint);
 
         Ok(Some(next_checkpoint))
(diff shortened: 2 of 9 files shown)


crates/sui-indexer-alt-framework-store-traits/src/lib.rs
wlmyng, line 12: We could reword the doc comment since `Connection` is basically just responsible for reporting the committer watermark to the framework and setting it (however the store desires)

crates/sui-indexer-alt/src/lib.rs
```diff
@@ -143,16 +143,18 @@ pub async fn setup_indexer(
...
+                let config = layer.finish(ConcurrentConfig {
+                    committer: committer.clone(),
+                    pruner: Some(pruner.clone()),
+                    ..Default::default()
+                })?;
+                if config.pruner.is_some() {
+                    indexer
+                        .trailing_concurrent_pipeline($handler, config)
+                        .await?
+                } else {
```
wlmyng: The framework knows at compile time whether a store implements `TrailingConnection`, so this logic should live in the framework rather than pushed to the caller. An indexer operator just calls `concurrent_pipeline(handler, config)` and the framework decides internally whether to spawn a regular or trailing pipeline based on `config.pruner` is `Some` (and later, other logic i.e. whether we need to backfill backwards)

crates/sui-kvstore/src/bigtable/store.rs
```diff
@@ -152,38 +139,4 @@ impl Connection for BigTableConnection<'_> {
 
         Ok(true)
     }
-
-    // Phase 1: Return Ok(None) - reader/pruner watermarks not needed for concurrent
```
wlmyng, line 104: it is neat to have the two `Connection` types though. This way, if someone had previously implemented as no-op, and wanted pruning, they might forget to update the trailing watermarks here and create an unstable state in their db

crates/sui-pg-db/src/store.rs
```diff
@@ -26,42 +24,17 @@ impl store::Connection for Connection<'_> {
...
-                .set(watermarks::pipeline.eq(excluded(watermarks::pipeline)))
-                .returning((watermarks::checkpoint_hi_inclusive, watermarks::reader_lo))
-                .get_result(self)
-                .await?;
-
+        let (checkpoint_hi_inclusive, _reader_lo) = self
+            .init_watermark_internal(
+                pipeline_task,
+                init_watermark.checkpoint_hi_inclusive,
+                init_watermark.checkpoint_hi_inclusive.map_or(0, |c| c + 1),
```
wlmyng: would be good to document why `reader_lo > checkpoint_hi_inclusive` and what that means

crates/sui-indexer-alt-framework/src/pipeline/concurrent/pruner.rs
```diff
@@ -100,7 +100,10 @@ pub(super) fn pruner<H: Handler + Send + Sync + 'static>(
...
     store: H::Store,
     metrics: Arc<IndexerMetrics>,
-) -> Service {
+) -> Service
+where
+    for<'c> <H::Store as Store>::Connection<'c>: TrailingConnection,
+{
     Service::new().spawn_aborting(async move {
         let Some(config) = config else {
             info!(pipeline = H::NAME, "Skipping pruner task");
```
wlmyng, line 106: we could consider an internal helper trait `TrailingStore` where the connection is a `TrailingConnection`
Then,

```   
  fn pruner<H: Handler>(...)                                                                                                                                                            
  where H::Store: TrailingStore      
```

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
```diff
@@ -314,9 +297,85 @@ pub(crate) fn pipeline<H: Handler + Send + Sync + 'static>(
...
+        task.as_ref(),
+        checkpoint_rx,
+        metrics,
+        main_reader_lo,
+    )
+}
+
+/// Like [pipeline], but extends it with trailing (reader/pruner) watermark tracking. Requires
+/// the store's connection to implement [TrailingConnection].
+pub(crate) fn trailing_pipeline<H: Handler + Send + Sync + 'static>(
```
wlmyng: framework/src/lib.rs should decide whether to call `pipeline` or `trailing_pipeline`

crates/sui-indexer-alt/src/lib.rs
evan-wall-mysten: Removed this as it no longer applies.

crates/sui-pg-db/src/store.rs
evan-wall-mysten: Added a comment to the postgres `Connection` for this.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/pruner.rs
evan-wall-mysten: I split this differently into `ConcurrentConnection`/`ConcurrentStore` vs `SequentialConnection`/`SequentialStore`. Let me know if this still applies.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
evan-wall-mysten: `trailing_pipeline` no longer exists

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
```diff
@@ -232,15 +234,19 @@ impl Default for PrunerConfig {
...
     config: ConcurrentConfig,
     store: H::Store,
     task: Option<Task>,
     checkpoint_rx: mpsc::Receiver<Arc<CheckpointEnvelope>>,
     metrics: Arc<IndexerMetrics>,
-) -> Service {
+) -> Service
+where
+    H: Handler + Send + Sync + 'static,
+    for<'c> <H::Store as Store>::Connection<'c>: ConcurrentConnection,
```
wlmyng: could we/ is it worth having `Concurrent` and `Sequential` stores have a `fn connect` that returns the respective connection, so as to avoid the additional typing?

crates/sui-pg-db/src/lib.rs
```diff
@@ -81,6 +81,8 @@ pub struct Db {
 struct CancelGuard<'m>(Option<&'m PoolMetrics>);
 
 /// Wrapper struct over the remote `PooledConnection` type for dealing with the `Store` trait.
+/// Internally, `checkpoint_high_inclusive = reader_lo - 1` is a temporary initial state that
```
wlmyng: this seems like a weird place to put this doc comment, I feel like on one of the watermark structs itself makes more sense?

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
wlmyng: if we have the `ConcurrentStore` define a `fn connect` that returns `ConcurrentConnection`, then would that help avoid the additional typing? We already have `Handler`'s store as `ConcurrentStore`

crates/sui-indexer-alt-framework/src/pipeline/concurrent/main_reader_lo.rs
```diff
@@ -14,16 +14,20 @@ use tracing::info;
...
-pub(super) fn track_main_reader_lo<H: Handler + 'static>(
+pub(super) fn track_main_reader_lo<H>(
     reader_lo: Arc<SetOnce<AtomicU64>>,
     reader_interval: Option<Duration>,
     store: H::Store,
-) -> Service {
+) -> Service
+where
+    H: Handler + 'static,
+    for<'c> <H::Store as Store>::Connection<'c>: ConcurrentConnection,
```
amnn: This change should not be necessary. We know that `H: concurrent::Handler` which implies that its `Store` is a `ConcurrentStore`, which implies that its connection is a `ConcurrentConnection`.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
amnn: Yes please, let's do that. They can be defaulted impls.

crates/sui-indexer-alt-framework/src/pipeline/sequential/committer.rs
```diff
@@ -53,7 +53,7 @@ pub(super) fn committer<H>(
 ) -> Service
 where
     H: Handler + Send + Sync + 'static,
-    H::Store: TransactionalStore + 'static,
+    H::Store: SequentialStore + 'static,
```
amnn: Is this constraint necessary in fact? Doesn't `H: Handler` imply this? I know that this constraint exists from before, but I'm curious what happens when it goes away.

crates/sui-indexer-alt-object-store/src/lib.rs
```diff
@@ -15,12 +15,11 @@ use object_store::PutPayload;
...
 use serde::Deserialize;
 use serde::Serialize;
+use sui_indexer_alt_framework_store_traits::ConcurrentConnection;
 use sui_indexer_alt_framework_store_traits::Connection;
-use sui_indexer_alt_framework_store_traits::InitWatermark;
 use sui_indexer_alt_framework_store_traits::PrunerWatermark;
 use sui_indexer_alt_framework_store_traits::ReaderWatermark;
 use sui_indexer_alt_framework_store_traits::Store;
-use sui_indexer_alt_framework_store_traits::init_with_committer_watermark;
 use sui_indexer_alt_framework_store_traits::{self as framework_traits};
```
amnn:
```suggestion
use sui_indexer_alt_framework_store_traits as framework_traits;
```

crates/sui-pg-db/src/lib.rs
amnn: Also `checkpoint_high_inclusive` -> `checkpoint_hi_inclusive`

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
evan-wall-mysten: Rust doesn't allow a subtrait to narrow the return type of a supertrait so I named the subtrait methods differently to avoid needing to disambiguate which one is being called.

crates/sui-pg-db/src/lib.rs
evan-wall-mysten: Moved to `InitWatermark` and fixed typo

crates/sui-indexer-alt-framework/src/pipeline/sequential/committer.rs
evan-wall-mysten: I went through and removed a bunch of redundant constraints like this. Most functions only need to specify `<H: Handler>` now.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/main_reader_lo.rs
evan-wall-mysten: This is no longer needed after changing `ConcurrentStore`'s definition to `ConcurrentStore: for<'c> Store<Connection<'c> = Self::ConcurrentConnection<'c>>`

crates/sui-indexer-alt-object-store/src/lib.rs
evan-wall-mysten: Fix this and a few other places that were doing this.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/main_reader_lo.rs
evan-wall-mysten: Also removed a bunch of other similar redundant constraints.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
amnn: Is there a way to hide the need to disambugate names within the trait hierarchy rather than externalise it? E.g. call the inner `Connection::connect` something else, so that each of `SequentialConnection` and `ConcurrentConnection` could have a `connect` function that gets used externally?

evan-wall-mysten:
> With the latest change, init_concurrent_watermark and init_sequential_watermark are looking remarkably similar, and it makes me wonder whether there is a point in having two different initialization flows -- the difference there is also introducing some undesirable complexity in the flows where we add pipelines.

I'll look at simplifying it further in a follow up PR, but I think it is difficult because more info is passed back for concurrent connections.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
evan-wall-mysten: This is removed now.

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
evan-wall-mysten: I think it is better to leave the trait-level comments as a summary of what the trait does without listing the functionality of its methods because it is easy for this function to become out of sync with them (e.g. `accepts_chain_id` was just added and it is not related to committer watermark).

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -41,6 +33,39 @@ pub trait Connection: Send {
...
+    ) -> anyhow::Result<bool>;
+}
+
+/// Extension of [`Connection`] for concurrent pipeline watermark operations.
+#[async_trait]
+pub trait ConcurrentConnection: Connection {
+    /// Returns the `InitWatermark` based on the existing watermark if it exists.
+    /// Otherwise, initializes a new watermark record and returns the default based on `reader_lo`.
+    ///
+    /// The default implementation delegates to `Connection::committer_watermark`.
```
amnn: Let's add an explanation here about why you would want to override the default implementation.

crates/sui-pg-db/src/store.rs
```diff
@@ -224,6 +213,43 @@ impl store::Connection for Connection<'_> {
...
+            .set(watermarks::pipeline.eq(excluded(watermarks::pipeline)))
+            .returning(watermarks::checkpoint_hi_inclusive)
+            .get_result(self)
+            .await?;
+
+        Ok(u64::try_from(checkpoint_hi_inclusive).ok())
+    }
+}
+
+#[async_trait]
```
amnn: nit: should this be after the `impl store::Store`?

crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -367,38 +342,63 @@ impl<S: Store> Indexer<S> {
...
-                    checkpoint_hi_inclusive: self.default_next_checkpoint.checked_sub(1),
-                    reader_lo: self.default_next_checkpoint,
-                },
-            )
-            .await
-            .with_context(|| format!("Failed to init watermark for {pipeline_task}"))?;
-
-        let next_checkpoint = checkpoint_hi_inclusive.map_or(reader_lo, |c| c + 1);
-
+        let next_checkpoint = init().await?;
```
amnn: The boilerplate associated with setting up this `init` function in the caller is not really justified by avoiding duplicating the two lines below it.

Can we switch to a setup where `add_pipeline` is just responsible for the steps above this line, and `concurrent_pipeline` and `sequential_pipeline` respectively call it, and then calculate `next_checkpoint` within themselves and use it to update `first_ingestion_checkpoint`?

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -107,17 +141,44 @@ pub trait Store: Send + Sync + 'static + Clone {
     where
         Self: 'c;
 
+    async fn connect<'c>(&'c self) -> anyhow::Result<Self::Connection<'c>>;
```
amnn, line 149: Call this one:


```suggestion
    async fn connect_inner<'c>(&'c self) -> anyhow::Result<Self::Connection<'c>>;
```

And then you are free to call `concurrent_connect` and `sequential_connect` both just `connect`, because at the call-sites you always know which kind of store you care about.

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
```diff
@@ -107,17 +141,44 @@ pub trait Store: Send + Sync + 'static + Clone {
...
     /// Delimiter used to separate pipeline names from task identifiers when reading or writing the
     /// committer watermark.
     const DELIMITER: &'static str = "@";
 
-    async fn connect<'c>(&'c self) -> anyhow::Result<Self::Connection<'c>>;
+    async fn concurrent_connect<'c>(&'c self) -> anyhow::Result<Self::ConcurrentConnection<'c>> {
+        Ok(self
+            .connect()
+            .await
+            .context("Failed to establish concurrent connection to store")?)
```
amnn: nit: no need for a context here -- this is a thin wrapper, so it's adding more noise than value.

amnn:
> I'll look at simplifying it further in a follow up PR, but I think it is difficult because more info is passed back for concurrent connections.

This actually touches on a point that I'm still not 100% happy with, which is that the default implementation of `init_concurrent_watermark` still does the thing where it passes through the `reader_lo` that was sent to it, which makes it seem like that is the watermark that was set when in fact the store ignored that watermark.

That is to say, when you look at a concurrent store that (today) doesn't care about pruning, and the sequential store, their needs during initialization seem similar. Does that mean the better fit is to have a single initialisation flow which allow the store to return an optional reader_lo, and if that returns `None` we know that pruning (or generally, the trailing edge) is not supported on the store?

For sequential pipelines it's incidental whether the store supports trailing edge operations or not, but for the concurrent pipelines this is a worthwhile distinction.

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
evan-wall-mysten: This function no longer exists.

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
evan-wall-mysten: `concurrent_connect` and `sequential_connect` no longer exist because the `add_pipeline`/`init_watermark` logic has been unified for both types of pipelines.

crates/sui-indexer-alt-framework/src/lib.rs
evan-wall-mysten: I added a new commit to this PR with an attempt to simplify this, but the end result is slightly different from your suggestion. Let me know your thoughts.

crates/sui-indexer-alt-framework-store-traits/src/lib.rs
evan-wall-mysten: Done

crates/sui-pg-db/src/store.rs
evan-wall-mysten: Moved for more consistent organization.

crates/sui-indexer-alt-framework/src/lib.rs
amnn: Seems good -- I did notice an unused function though, fixed in https://github.com/MystenLabs/sui/pull/26092

PR #25882, opened 2026-03-18
Enforce minimum deposit sizes for gasless tx

aschran:
## Description 

Changes token allow list flag to be a map from token to minimum deposit size.

## Test plan 

Added tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/traffic_control_tests.rs
+++ b/crates/sui-e2e-tests/tests/traffic_control_tests.rs
@@ -1023,7 +1023,7 @@ async fn test_validator_traffic_control_gasless_spam_blocked() -> Result<(), any
     let recipient = test_cluster.wallet.get_addresses()[1];
 
     // Register SUI as an allowed gasless token type for this test.
-    add_gasless_token_for_testing(GAS::type_tag().to_canonical_string(true));
+    add_gasless_token_for_testing(GAS::type_tag().to_canonical_string(true), 0);
 
     // Build a gasless transaction with a valid balance::send_funds MoveCall.
     let mut builder = ProgrammableTransactionBuilder::new();


crates/sui-e2e-tests/tests/gasless_tests.rs
```diff
@@ -911,3 +912,282 @@ async fn test_gasless_coin_split_and_keep_change() {
...
     test_env.trigger_reconfiguration().await;
 }
+
+// ============================================================
+// Minimum Transfer Amount Tests
+// ============================================================
+
+#[cfg_attr(not(msim), ignore)]
+#[sim_test]
+async fn test_gasless_transfer_below_minimum_fails() {
```
mystenmark: i wonder if these tests would be better written as transactional tests?

mystenmark:
Couple thoughts:

- should we allow <minimum transfers from A => A? This would allow people to convert small coins to AB for free, which we would want them to do anyway.
- It is technically less permissive, but should we just statically check the `amount` arg to each `send_funds` call? Does that open a hole that I'm not seeing?

aschran:
> * should we allow <minimum transfers from A => A? This would allow people to convert small coins to AB for free, which we would want them to do anyway.

No strong opinion on this

> * It is technically less permissive, but should we just statically check the `amount` arg to each `send_funds` call? Does that open a hole that I'm not seeing?

It seems better to be more permissive here, so we can avoid SDKs' having to special-case how they generate tx to be gasless compatible.

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -586,6 +587,52 @@ impl<'backing> TemporaryStore<'backing> {
             ));
         }
 
+        let allowed_types =
+            sui_types::transaction::get_gasless_allowed_token_types(self.protocol_config);
+
+        // Compute net amounts per (address, token_type): Merge adds, Split subtracts.
+        // Positive nets are recipient deposits that must meet the minimum transfer amount.
+        // Negative nets are sender withdrawals (expected). Zero nets are no-ops.
```
mystenmark: couldn't we just check the amounts in the send_fund calls in validate_gasless_transaction? is there some loophole that allows that i'm not seeing?

aschran: see my reply above

aschran: (wait, also, what amounts in the send_fund calls are you referring to? I don't think `send_funds` takes an amount as an argument, just the Balance or Coin object?)

mystenmark: lol sorry I forgot I commented earlier. ok, yes, its a bit harder than I thought. possible but probably not important to handle right now.

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -586,6 +587,52 @@ impl<'backing> TemporaryStore<'backing> {
...
+                    )
+                    .ok_or_else(|| "Accumulator event type is not Balance<T>".to_string())?;
+                    let delta = match event.write.operation {
+                        AccumulatorOperation::Merge => amount as i128,
+                        AccumulatorOperation::Split => -(amount as i128),
+                    };
+                    *totals
+                        .entry((event.write.address.address, inner_type))
+                        .or_default() += delta;
+                    Ok::<_, String>(totals)
```
alex-mysten: This netting logic lives several places - should/can we consolidate?

sui-execution/latest/sui-adapter/src/temporary_store.rs
alex-mysten: Are there trivial static ptb validations we should add as a way to short-circuit execution that is guaranteed to fail? I would imagine we would need a withdraw or coin input with at least min_transfer in order to complete a transfer

aschran: there probably are, but given that the whole thing is in theory statically checkable and yet we are not doing that, I wonder if it makes sense to implement piecemeal static checks vs just waiting until such time as we deem it worthwhile to fully implement static verification here? doing partial checks won't stop someone determined to make it fail after execution, and it's also hard for me to imagine accidental tiny transfers being much of a problem in practice?

sui-execution/latest/sui-adapter/src/temporary_store.rs
aschran: addressed

PR #25885, opened 2026-03-18
[move] Introduce TraceType-based filtering for Move VM test traces

tzakian:
## Description 

Add configurable trace filtering via `--trace
[full|instruction-only|function-only]` (default = full which matches existing functionality) to allow callers to capture only the trace events they need. This improves performance for coverage gathering (which only needs instruction and frame events) and other use cases where full traces are unnecessary.

- Unify UnitTestingConfig's trace (bool) and trace_type fields into a
  single trace: Option<TraceType>
- Add InstructionOnlyTracer and FunctionOnlyTracer implementations of the Tracer trait in move-trace-format
- Organize all Tracer implementations under a new tracers/ module
- Wire TraceType through to MoveTraceBuilder construction in the test runner so filtered tracers are used at execution time (this is mainly to improve coverage gathering perf, but has other benefits as mentioned above).

## Test plan 

Existing tests + added additional tests for the different sparse tracing types.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-trace-format/src/tracers/function_only.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::format::TraceEvent;
+use crate::interface::{Tracer, Writer};
+
+/// A tracer that only keeps `OpenFrame`, and `CloseFrame` events.
+pub struct FunctionOnlyTracer;
```
dariorussi, line 8: yes, thank you, yes and
yes

PR #25887, opened 2026-03-18
Better bigtable connection pool.

nickvikeras:
## Description

When I was working on backfill performance, I had noticed there was only a single grpc connection. This is fine under moderate load because of http2 multiplexing, but under very high load (like a backfill, or even a heavily loaded gRPC server) we can start queuing when we hit the [MAX_CONCURRENT_STREAMS](https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#ga5bb9256d0b457cab219f4d9d765148b6) limit. From Google's official java client -

> For google services, gRPC channels will start locally queuing RPC when there are 100 concurrent RPCs.

I tried adding a basic connection pool to avoid bottlenecks during backfills, but that caused problems on the read side where load is low. With a connection pool of size 10, I saw frequently was paged for error spikes due to stale connections. I dug into this a bit and found that GFE (Google Front End) [reaps active connections every hour](https://docs.cloud.google.com/java/docs/reference/gax/latest/com.google.api.gax.grpc.ChannelPoolSettings). I am assuming that with one connection, we reconnect immediately and only a single request fails (or maybe it's just hidden by retries somewhere), but with 10 connections, several requests can fail over a longer period of time when they all go stale. To resolve this I just temporarily set the number of connections back to 1 on the reader side.

So I wanted to solve 2 problems -

1. Zero-config adaptive scaling of the number of connections for RPC providers.
2. Reconnect automatically in the background _before_ GFE reaps the connection to avoid (even small) error spikes.

Both the official Go and Java clients from Google implement this, so I was able to mostly just copy logic over from the official clients (linked below for posterity).

The Java client has some additional features the go client doesn't have that periodically check connection health in the background and proactively replace stale connections _before_ the GFE reaping window kicks in, but I felt that was probably not necessary for us, especially because the official Go client doesn't even do that.

We might eventually want to rip this out of the monorepo and move it to sui-rust-sdk and use it in the Sui grpc client as well, but for now I think that is less important.

### Reference implementations

**Java** ([googleapis/java-bigtable@2933ef64](https://github.com/googleapis/java-bigtable/commit/2933ef64)):

- [BigtableChannelPool.java](https://github.com/googleapis/java-bigtable/blob/2933ef64/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/grpc/BigtableChannelPool.java)
- [BigtableChannelPoolSettings.java](https://github.com/googleapis/java-bigtable/blob/2933ef64/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/grpc/BigtableChannelPoolSettings.java)
- [ChannelPoolHealthChecker.java](https://github.com/googleapis/java-bigtable/blob/2933ef64/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/grpc/ChannelPoolHealthChecker.java)

**Go** ([googleapis/google-cloud-go@e57b013eda](https://github.com/googleapis/google-cloud-go/commit/e57b013eda)):

- [connpool.go](https://github.com/googleapis/google-cloud-go/blob/e57b013eda/bigtable/internal/transport/connpool.go)
- [conn_recycler.go](https://github.com/googleapis/google-cloud-go/blob/e57b013eda/bigtable/internal/transport/conn_recycler.go)
- [dynamic_scale_monitor.go](https://github.com/googleapis/google-cloud-go/blob/e57b013eda/bigtable/internal/transport/dynamic_scale_monitor.go)

## Test plan

So far unit tests. Before merging I'll also do a manual local smoke test against the bigtable emulator before merging. Beyond that I feel comfortable just letting it soak in testnet.

### DIFF
--- a/crates/sui-kvstore/src/bigtable/client/channel_pool.rs
+++ b/crates/sui-kvstore/src/bigtable/client/channel_pool.rs
@@ -396,6 +396,9 @@ impl Service<Request<Body>> for ChannelPool {
         }
 
         let entries = self.inner.entries.load();
+        if entries.is_empty() {
+            return Poll::Ready(Err("no channels available".into()));
+        }
         let idx = self.inner.ticker.fetch_add(1, Ordering::Relaxed) % entries.len();
         let entry = &entries[idx];
         let mut channel = entry.channel.clone();


crates/sui-kv-rpc/src/main.rs
```diff
@@ -43,6 +43,15 @@ struct App {
     /// Channel-level timeout in milliseconds for BigTable gRPC calls (default: 60000)
     #[clap(long = "bigtable-channel-timeout-ms")]
     bigtable_channel_timeout_ms: Option<u64>,
+    /// Number of gRPC channels to create at startup (default: 10)
```
amnn: nit: it can be neater to factor this out into its own struct, with a `Default` impl that sets the defaults, and define `clap` directives on that like so:

```rust
struct App {
  // ...

  #[clap(flatten)]
  connection_pool: PoolArgs,
}

struct PoolArgs {
    #[clap(long, default_value_t = Default::default().bigtable_initial_pool_size)]
    bigtable_initial_pool_size: usize,

    #[clap(long, default_value_t = Default::default().bigtable_min_pool_size)]
    bigtable_min_pool_size: usize,

    #[clap(long, default_value_t = Default::default().bigtable_max_pool_size)]
    bigtable_max_pool_size: usize,
    // ... maybe other values belong here also
}

impl Default for PoolArgs {
    fn default() -> Self {
        Self {
            bigtable_initial_pool_size: 10,
            bigtable_min_pool_size: 1,
            bigtable_max_pool_size: 200,
        }
    }
}
```

With this pattern there's a single source of truth for defaults, it works for clap and for tests where you may be manually constructing the `PoolArgs`, and the documentation produced by `clap` is consistent in its presentation of the default values too (and doesn't require duplicating the defaults which may go stale with time).

crates/sui-kvstore/src/bigtable/client/mod.rs
amnn, line 1: Imports are all over the place in this file.

crates/sui-kv-rpc/src/main.rs
amnn: You may be able to reuse your existing `PoolConfig` for this (although it would be good to call it `PoolArgs` for consistency -- we tend to use `Config` for files and `Args` for command-line args).

crates/sui-kvstore/src/bigtable/client/mod.rs
```diff
@@ -188,25 +212,19 @@ impl BigTableClient {
...
-                ))
-                .expect("channel balancer dropped");
-            }
-            channel
-        } else {
-            endpoint.connect_lazy()
-        };
-        let auth_channel = AuthChannel {
-            channel,
+        // Instance name for PingAndWarm: strip trailing "tables/"
```
amnn: nit: seems strange to create the larger thing, and then strip something off it for the smaller thing, rather than construct `instance_name` first and then append `tables/` to get `table_prefix`.

crates/sui-kvstore/src/bigtable/metrics.rs
```diff
@@ -137,3 +143,46 @@ impl KvMetrics {
...
+                "bt_pool_rpcs_completed",
+                "Total RPCs completed through the pool",
+                registry,
+            )
+            .unwrap(),
+        }
+    }
+}
+
+impl PoolMetrics for PoolPrometheusMetrics {
```
amnn: Does anything else implement this trait (i.e. does it need to be a trait right now?)

crates/sui-kvstore/src/bigtable/client/mod.rs
```diff
@@ -81,41 +78,69 @@ pub struct MutationError {
...
         instance_id: String,
         client_name: &str,
     ) -> Result<Self> {
-        let auth_channel = AuthChannel {
-            channel: Channel::from_shared(format!("http://{host}"))?.connect_lazy(),
-            policy: "https://www.googleapis.com/auth/bigtable.data".to_string(),
-            token_provider: None,
-            token: Arc::new(RwLock::new(None)),
+        let endpoint = Channel::from_shared(format!("http://{host}"))?;
+        let config = PoolConfig {
```
amnn: Maybe `PoolConfig` deserves a `singleton` constructor or something?

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+        if let Some(m) = &self.inner.metrics {
+            m.set_pool_size(entries.len());
+        }
+        self.inner.entries.store(Arc::new(entries));
+        Ok(())
+    }
+
+    /// Spawn the background maintenance loop.
+    pub(crate) fn spawn_background_tasks(&self) {
+        self.spawn_maintenance();
```
amnn: Is there a reason for this double indirection?

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+    }
+
+    /// Convenience: create, connect, and spawn background tasks.
+    pub(crate) async fn new_connected(
```
amnn: nit: constructors to the top

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+        pool.spawn_background_tasks();
+        Ok(pool)
+    }
+
+    fn spawn_maintenance(&self) {
+        let weak = Arc::downgrade(&self.inner);
+        let interval = self.inner.config.maintenance_interval;
+        tokio::spawn(async move {
+            loop {
+                tokio::time::sleep(interval).await;
```
amnn: Use `tokio::time::interval`?

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+                    return;
+                };
+                inner.run_maintenance().await;
+            }
+        });
+    }
+}
+
+impl ChannelPoolInner {
+    async fn run_maintenance(&self) {
```
amnn: Also seems kind of like a redundant helper function.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+            reserved: None,
+        }
+    }
+
+    /// Connect `initial_pool_size` channels, priming each one.
+    pub(crate) async fn connect(&self) -> Result<()> {
+        let mut entries = Vec::with_capacity(self.inner.config.initial_pool_size);
+        for _ in 0..self.inner.config.initial_pool_size {
+            entries.push(Arc::new(
+                create_primed_entry(
```
amnn: Could this be calling `self.inner.create_primed_entry` and in that case do you need the free function at all?

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+    }
+}
+
+impl ChannelPoolInner {
+    async fn run_maintenance(&self) {
+        self.refresh().await;
+        self.resize().await;
+    }
+
+    async fn refresh(&self) {
```
amnn, line 174: Doc comments on these functions would be helpful to understand what maintenance involves.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+        if load <= self.config.min_rpcs_per_channel as f64 {
+            // Scale down: load is too low. Require consecutive low-load observations
+            // before actually shrinking, to avoid flapping (mirrors official Google Go
+            // client continuousDownscaleRuns logic).
+            let runs = self.consecutive_low_load.fetch_add(1, Ordering::Relaxed) + 1;
+            if runs < self.config.downscale_threshold {
+                return;
+            }
+
+            // Ideal pool size to bring per-channel load back to target_load.
```
amnn, line 239: Should this check happen before you determine that you need to increment `consecutive_low_load`? (swap this block with the one above)?

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+            }
+
+            let mut current = (**self.entries.load()).clone();
+            let old_size = current.len();
+            let new_len = current
+                .len()
+                .saturating_sub(delta)
+                .max(self.config.min_pool_size);
+            current.truncate(new_len);
+            let new_size = current.len();
```
amnn: This section seems to be redoing work -- do we need to recompute `current_size` as `old_size`, and `new_len` as `new_size`? It could be that we do for concurrency reasons, in which case a comment would be helpful. It would also be useful to keep the names consistent (even shadowed) to indicate that these really are the same values, re-computed for freshness or something.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
amnn, line 1: nit: file order, import style etc.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+            }
+        }
+    }
+
+    fn call(&mut self, request: Request<Body>) -> Self::Future {
+        let (entry, mut channel) = self.reserved.take().expect("called before poll_ready");
+        let metrics = self.inner.metrics.clone();
+
+        Box::pin(async move {
+            let result = channel.call(request).await;
```
amnn, line 431: Is it possible for the future related to this call to be dropped while it's waiting inside this await? If so your `in_flight` metric will leak and be artificially high, because your `Drop` impl for `ChannelPool` looks for `self.reserved` to be `Some(_)` in order to decrement `in_flight`, but at L372 this has already happened.

This could be quite common if this call is part of handling an RPC request and the client hangs up before Bigtable responds. Axum would then drop the future at that point.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,692 @@
...
+    }
+
+    async fn resize(&self) {
+        let entries = self.entries.load();
+        let current_size = entries.len();
+        if current_size == 0 {
+            return;
+        }
+
+        let total_in_flight: usize = entries
```
amnn, line 228: It seems like this control system is going to be sensitive to whatever the load is at the moment the maintenance cycle executes, which is relatively rare (every minute). Have you measured how this performs against various kinds of workloads (e.g. spiky ones)? Did you consider other control mechanisms?

crates/sui-kv-rpc/src/main.rs
amnn: Although maybe not because now I see that it has a bunch more values in it.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
nickvikeras: This is what the official gcloud Go client does. To avoid flapping due to load it only scales down if multiple measurements in a row show low utilization. 

The Java client on the other hand tracks the peak inflight for every maintenance window and uses that to decide if it should move the pool size.

I just kind of arbitrarily decided to copy the go client because it was a bit simpler and more recently written.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
nickvikeras: It's just vestigial. There used to be multiple background tasks in an earlier iteration of this.

crates/sui-kvstore/src/bigtable/metrics.rs
nickvikeras: I made it a trait because I was looking forward to using this in the Sui grpc client, and I wasn't sure if we had a prometheus dependency in the SDK. But for now it can just be a struct.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
nickvikeras: Also vestigial.

crates/sui-kv-rpc/src/main.rs
nickvikeras: Fixed. I didn't expose all of the knobs in this CLI because I doubt they really will need to be tweaked and eventually I'd like to migrate this service to a toml config file.

crates/sui-kvstore/src/bigtable/client/mod.rs
nickvikeras: Fixed

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
nickvikeras: I think this is fixed now

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
nickvikeras: Good catch. Fixed with an inflight guard that decrements the counter on a drop.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
nickvikeras: Currently the logic is exactly matching the logic in the [reference implementation](
https://github.com/googleapis/google-cloud-go/blob/7ac72cc8c8/bigtable/internal/transport/dynamic_scale_monitor.go#L92-L141), but I am pretty sure what you are describing would be functionally equivalent and is a bit simpler, so I made the change.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
```diff
@@ -0,0 +1,702 @@
...
+        if let Some(m) = &self.inner.metrics {
+            m.pool_size.set(entries.len() as i64);
+        }
+        self.inner.entries.store(Arc::new(entries));
+        Ok(())
+    }
+
+    /// Spawn the background maintenance loop that periodically refreshes aged-out
+    /// channels and resizes the pool based on load.
+    fn spawn_background_tasks(&self) {
```
amnn, line 152: Is it possible to have this return a `Service` from `sui-futures`, so that we can tie its shutdown into the graceful shutdown of the service as a whole? I see that it handles shutdown through the weak pointer, but seeing as most if not all the places we use this have a `Service` that can look after graceful shutdown, it seems low effort to pull this is in as a secondary task, so that it can be aborted when the larger service no longer needs it.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
amnn: Got it -- I guess we can monitor and see!

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
amnn: It's not quite functionally equivalent, because it means that if our target size is larger than the current size, we don't treat that as a low load signal, meaning it could take longer to actually reduce the size, but that seems desirable to me?

There's even an argument to be made that if you return when `delta == 0` you should reset the `consecutive_low_load` count.

amnn (approved):
LGTM!

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
nickvikeras: Cleaned this up. This was also vestigial. There _used_ to be a concurrency issue because I had a separate connection health check task that ran more frequently (ported from the java client) that I ripped out. So now this is the only thing that mutates the state.

crates/sui-kvstore/src/bigtable/client/channel_pool.rs
nickvikeras: I considered doing that initially but it requires more plumbing through all the API layers and managing the service lifecycle adds a little more complexity to every call site. The weak ref ties the maintenance task's lifetime to the pool's lifetime which seemed like a better way to manage the lifecyle to me.

PR #25895, opened 2026-03-19
sui-indexer-alt-framework: add chain_id to IngestionClient

evan-wall-mysten:
## Description 

This PR adds a new `chain_id()` method to `IngestionClientTrait`. The `IngestionClient` calls this method the first time it is used and caches it to return in the `CheckpiontEnvelope`. 

`IngestionClient` callers currently ignore the `chain_id`.

## Test plan 

Updated existing tests with chain_id assertions and mock the chain_id reponse.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Adds new `IngestionClientTrait::chain_id` to retrieve the `chain_id` from that ingestion source.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
@@ -21,6 +21,8 @@ use object_store::local::LocalFileSystem;
 use sui_futures::future::with_slow_future_monitor;
 use sui_rpc::Client;
 use sui_rpc::client::HeadersInterceptor;
+use sui_types::digests::ChainIdentifier;
+use tokio::sync::OnceCell;
 use tracing::debug;
 use tracing::error;
 use tracing::warn;
@@ -30,6 +32,7 @@ use crate::ingestion::Error as IngestionError;
 use crate::ingestion::MAX_GRPC_MESSAGE_SIZE_BYTES;
 use crate::ingestion::Result as IngestionResult;
 use crate::ingestion::decode;
+use crate::ingestion::error::Error::FetchError;
 use crate::ingestion::store_client::StoreIngestionClient;
 use crate::metrics::CheckpointLagMetricReporter;
 use crate::metrics::IngestionMetrics;
@@ -47,6 +50,8 @@ const SLOW_OPERATION_WARNING_THRESHOLD: Duration = Duration::from_secs(60);
 
 #[async_trait]
 pub(crate) trait IngestionClientTrait: Send + Sync {
+    async fn chain_id(&self) -> anyhow::Result<ChainIdentifier>;
+
     async fn checkpoint(&self, checkpoint: u64) -> CheckpointResult;
 }
 
@@ -168,6 +173,13 @@ pub struct IngestionClient {
     /// Wrap the metrics in an `Arc` to keep copies of the client cheap.
     metrics: Arc<IngestionMetrics>,
     checkpoint_lag_reporter: Arc<CheckpointLagMetricReporter>,
+    chain_id: OnceCell<ChainIdentifier>,
+}
+
+#[derive(Clone, Debug)]
+pub struct CheckpointEnvelope {
+    pub checkpoint: Arc<Checkpoint>,
+    pub chain_id: ChainIdentifier,
 }
 
 impl IngestionClient {
@@ -270,6 +282,7 @@ impl IngestionClient {
             client,
             metrics,
             checkpoint_lag_reporter,
+            chain_id: OnceCell::new(),
         }
     }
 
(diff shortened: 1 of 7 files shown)


crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
```diff
@@ -402,7 +415,51 @@ impl IngestionClient {
             .total_ingested_objects
             .inc_by(data.object_set.len() as u64);
 
-        Ok(Arc::new(data))
+        let chain_id_client = self.client.clone();
```
amnn, line 427: Let's factor this into a private helper function `IngestionClient::get_or_init_chain_id`.

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
```diff
@@ -435,8 +493,18 @@ mod tests {
         permanent_failures: DashMap<u64, usize>,
     }
 
+    impl MockIngestionClient {
+        fn mock_chain_id_value() -> ChainIdentifier {
```
amnn: nit: redundant word

```suggestion
        fn mock_chain_id() -> ChainIdentifier {
```

crates/sui-indexer-alt/src/bootstrap.rs
amnn, line 1: It should be possible to simplify this code further, because part of it is about calculating the genesis checkpoint digest, which we already have in the response now.

crates/sui-indexer-alt-framework/src/ingestion/store_client.rs
```diff
@@ -64,6 +65,11 @@ impl StoreIngestionClient {
 
 #[async_trait::async_trait]
 impl IngestionClientTrait for StoreIngestionClient {
+    async fn chain_id(&self) -> anyhow::Result<ChainIdentifier> {
+        let checkpoint = self.checkpoint(0).await?;
+        Ok((*checkpoint.summary.digest()).into())
+    }
```
wlmyng, line 71: so there's an expectation that the store client - this is the remote store/ bigtable/ archival? - will have an unpruned checkpoints table, or at least the genesis checkpoint 0 must exist

crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -1232,8 +1232,8 @@ mod tests {
         let temp_dir = tempfile::tempdir().unwrap();
         synthetic_ingestion::generate_ingestion(synthetic_ingestion::Config {
             ingestion_dir: temp_dir.path().to_owned(),
-            starting_checkpoint: 5,
-            num_checkpoints: 25,
+            starting_checkpoint: 0,
```
wlmyng, line 1235: changing starting checkpoint to 0 to account for required genesis checkpoint -> chain_id?

evan-wall-mysten: Yeah, I think it would be more obvious if this API took a `[start, end]` since the change sets the `start` to `0` and leaves `end` the same.

crates/sui-indexer-alt-framework/src/ingestion/store_client.rs
evan-wall-mysten: Yeah that is currently the expectation. If we want to change that we either have to 
  * Make this method return `Result<Option<ChainIdentifier>>` if we want to support stores that do not know their chain ID
  * Support storing the chain ID in some other way like a "chain_id.txt" file

I think this is ok for now since we know that the genesis checkpoint is present in our object store and we can relax it (make it `Option`) later if we decide we need to.

crates/sui-indexer-alt/src/bootstrap.rs
evan-wall-mysten: I think it is slightly simpler now as
```
genesis_checkpoint_envelope.chain_id.as_bytes().to_vec()
```
because it removed the `digest()` call.

evan-wall-mysten:
> Looks good -- not sure if `Envelope` already has a meaning beyond "wrapper type" (I have only seen it used in the context of cryptographic primitives, to associate a signature with some underlying data)

That is all I meant by it, but I have no problem calling it something else if we already have a name for this pattern.

amnn:
@evan-wall-mysten note that this is a breaking change for the Indexing Framework. You need to update the release notes section of your PR description with a summary of the breaking change. Ideally this needs to be done before the PR lands, so that Jessie is notified and can review for the changelog, but in this case, just update the PR description and ping her on the PR.

PR #25900, opened 2026-03-19
[move][tracing] Update tracing to support not generating unneeded values internally for sparse traces

tzakian:
## Description 

This adds `wants_effects()` to the `Tracer` trait so the VM tracer can skip expensive value resolution when the tracer will discard effect events and those values are not needed.

When `InstructionOnlyTracer` or `FunctionOnlyTracer` is used, the VM tracer was eagerly converting every runtime value into a `SerializableMoveValue` via `into_annotated_move_value` which incurred a recursive tree traversal with StructTag/Identifier cloning) for Push/Pop/Read/Write effects — only for those effects to be immediately filtered out by `push_event`. This is the dominant per-instruction cost in the tracer.

Changes:
- Add `Tracer::wants_effects() -> bool` to the trait, and this is set to false by `InstructionOnlyTracer` and `FunctionOnlyTracer`
- Cache the result as `wants_effects: bool` in VMTracer at construction.
- In `start_instruction_impl`: early-return path that skips all value resolution except `StLoc` side effects (`store_global`/`insert_local`) and index capture for `VecImmBorrow`/`VecMutBorrow`.
- In `end_instruction_impl`: use `push_effect!` / `emit!` macros to gate value resolution behind `wants_effects`, while always performing type-stack management. Write-effect arms (`StLoc`, `WriteRef`, `VecPushBack`, `VecSwap`) use inline conditionals since their resolution logic is treated in a more special manner.
- Full tracing is unchanged — other than the bool check per instruction.

The instruction-only trace baselines shrink (no effect events emitted), and full trace baselines see minor changes from the `FunctionOnlyTracer`'s notify fix (which was incorrectly keeping `Instruction` events). The changes in the instruction only traces is because the trace IDs used for the frame IDs change (because the trace is smaller).

## Test plan 

Existing tracing tests. This reduces generation of sparse traces 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
@@ -269,12 +269,6 @@ impl VMTracer<'_> {
         std::mem::take(&mut self.effects)
     }
 
-    /// Skip building effects: clear any captured pre-effects and return empty.
-    fn skip_effects(&mut self) -> Vec<EF> {
-        self.effects.clear();
-        vec![]
-    }
-
     /// Insert a local with a specifice runtime location into the current frame.
     fn insert_local(&mut self, local_index: usize, local: StackType) -> Option<()> {
         *self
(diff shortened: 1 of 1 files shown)


external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
```diff
@@ -1084,17 +1119,53 @@ impl VMTracer<'_> {
         // executed the instruction and we now need to manage the type transition of the
         // instruction along with snapshoting the effects of the instruction's execution.
         let instruction = &machine.call_stack.current_frame.function.to_ref().code()[pc as usize];
+
+        // Macro: resolve stack values as Push effects, or skip if !wants_effects.
+        // - push_effect!()  — resolve the top-of-stack (index 0) as a single Push effect.
+        // - push_effect!(n) — resolve n stack values as Push effects (in reverse order).
+        macro_rules! push_effect {
```
cgswords: Please write a macro `emit_effects!` that takes a body and reuse it everywhere. 

Also, this macro is weird to see littered in the code below. I think preserving the intent of that code by using some `emit_effects! { ... }` will help a lot.

external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
```diff
@@ -268,6 +269,12 @@ impl VMTracer<'_> {
         std::mem::take(&mut self.effects)
     }
 
+    /// Skip building effects: clear any captured pre-effects and return empty.
+    fn skip_effects(&mut self) -> Vec<EF> {
```
cgswords: Why do we clear them here? I appreciate we are trying to save some work, but generating the pre-effects is work too. We should never run afoul of this in the limit.

external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
tzakian: Updated. Much nicer!

external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
tzakian: Was able to remove without much issue so good call.

PR #25904, opened 2026-03-20
Parallelize bulk_insert_live_objects during snapshot restore

andll:
## Summary
- Refactors `AuthorityStore::bulk_insert_live_objects` to accept `Arc<AuthorityPerpetualTables>` and `Vec<LiveObject>` (collected at the call site)
- Verifies SHA3 digest over the full object set upfront, then splits into 8 chunks and inserts each chunk concurrently via `tokio::task::spawn_blocking`
- Updates `StateSnapshotReaderV1::read` / `sync_live_objects` to propagate the `Arc` and use an async `try_for_each` closure

## Test plan
- [ ] `cargo check -p sui-core -p sui-snapshot -p sui-tool`
- [ ] `SUI_SKIP_SIMTESTS=1 cargo nextest run -p sui-snapshot`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8456,7 +8456,7 @@ dependencies = [
 [[package]]
 name = "minibytes"
 version = "0.1.0"
-source = "git+https://github.com/mystenlabs/tidehunter.git?rev=72a31b4d5c6d4a77c89b778ae528c6d0679f3247#72a31b4d5c6d4a77c89b778ae528c6d0679f3247"
+source = "git+https://github.com/mystenlabs/tidehunter.git?rev=c7aecba15ca5d119f03c148fecc54cf845e98947#c7aecba15ca5d119f03c148fecc54cf845e98947"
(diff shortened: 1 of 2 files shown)


crates/sui-core/src/authority/authority_store.rs
```diff
@@ -636,17 +636,51 @@ impl AuthorityStore {
...
+        let sha3_digest = hasher.finalize().digest;
+        if *expected_sha3_digest != sha3_digest {
+            error!(
+                "Sha does not match! expected: {:?}, actual: {:?}",
+                expected_sha3_digest, sha3_digest
+            );
+            return Err(SuiError::from("Sha does not match"));
+        }
+
+        const NUM_PARALLEL_CHUNKS: usize = 8;
```
johnjmartin: Should this be configurable?

andll: Yes(done)

PR #25905, opened 2026-03-20
sui-indexer-alt-framework: add chain_id to CheckpointStreamingClient

evan-wall-mysten:
## Description 

This continues the work started in #25895 by adding a `chain_id` response header that the `CheckpointStreamingClient` reads. The `chain_id` is plumbed through to the `Processor` alongside `checkpoint`.

The `processor`'s `Service` currently ignores the `chain_id`.

This change depends on https://github.com/MystenLabs/sui/pull/25908.

## Test plan 

Updated existing unit tests to assert the `chain_id` is passed along with the `checkpoint`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Changed `Processor`'s receiver to accept new `CheckpointEnvelope` type containing the `chain_id`.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
@@ -5,11 +5,15 @@ use std::pin::Pin;
 use std::time::Duration;
 
 use anyhow::Context;
+use anyhow::anyhow;
 use async_trait::async_trait;
 use futures::Stream;
 use futures::StreamExt;
+use sui_rpc::headers::X_SUI_CHAIN_ID;
 use sui_rpc::proto::sui::rpc::v2::SubscribeCheckpointsRequest;
 use sui_rpc::proto::sui::rpc::v2::subscription_service_client::SubscriptionServiceClient;
+use sui_types::digests::ChainIdentifier;
+use sui_types::messages_checkpoint::CheckpointDigest;
 use tonic::Status;
 use tonic::transport::Endpoint;
 use tonic::transport::Uri;
@@ -19,12 +23,15 @@ use crate::ingestion::error::Error;
 use crate::ingestion::error::Result;
 use crate::types::full_checkpoint_content::Checkpoint;
 
-/// Type alias for a stream of checkpoint data.
-pub type CheckpointStream = Pin<Box<dyn Stream<Item = Result<Checkpoint>> + Send>>;
+pub struct CheckpointStream {
+    pub stream: Pin<Box<dyn Stream<Item = Result<Checkpoint>> + Send>>,
+    pub chain_id: ChainIdentifier,
+}
 
 /// Trait representing a client for streaming checkpoint data.
 #[async_trait]
 pub trait CheckpointStreamingClient {
+    /// Returns the CheckpointStream and chain id.
     async fn connect(&mut self) -> Result<CheckpointStream>;
 }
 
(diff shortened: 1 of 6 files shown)


crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
```diff
@@ -19,13 +23,19 @@ use crate::ingestion::error::Error;
 use crate::ingestion::error::Result;
 use crate::types::full_checkpoint_content::Checkpoint;
 
+pub struct CheckpointStreamEnvelope {
+    pub checkpoint_stream: CheckpointStream,
+    pub chain_id: ChainIdentifier,
+}
+
 /// Type alias for a stream of checkpoint data.
 pub type CheckpointStream = Pin<Box<dyn Stream<Item = Result<Checkpoint>> + Send>>;
```
amnn: Envelope is probably not an appropriate name for this because the thing being wrapped is not simple data, but you also don't need to name both levels of thing, so I would suggest something like this:

```suggestion
pub struct CheckpointStream {
    pub stream: Pin<Box<dyn Stream<Item = Result<Checkpoint>> + Send>>,
    pub chain_id: ChainIdentifier,
}
```

crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
```diff
@@ -63,13 +73,23 @@ impl CheckpointStreamingClient for GrpcStreamingClient {
...
+        let chain_id_value = response.metadata().get(X_SUI_CHAIN_ID).ok_or_else(|| {
+            Error::StreamingError(anyhow!("Chain ID not found in response metadata"))
+        })?;
+        let chain_id_str = chain_id_value
+            .to_str()
+            .map_err(|e| Error::StreamingError(anyhow!("Chain ID is not valid ASCII: {e}")))?;
+        let chain_id: ChainIdentifier = chain_id_str
+            .parse::<CheckpointDigest>()
+            .map_err(|e| Error::StreamingError(anyhow!("Chain ID parse error: {e}")))?
+            .into();
```
amnn:
```suggestion

        let chain_id: ChainIdentifier = chain_id_value
            .to_str()
            .map_err(|e| Error::StreamingError(anyhow!("Chain ID is not valid ASCII: {e}")))?;
            .parse::<CheckpointDigest>()
            .map_err(|e| Error::StreamingError(anyhow!("Chain ID parse error: {e}")))?
            .into();
```

crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
evan-wall-mysten: Yeah the long `Pin<Box<dyn Stream<Item = Result<Checkpoint>> + Send>>` type is not referenced anywhere else so this works.

evan-wall-mysten:
> Looks good modulo minor comments -- remember to update the release notes!

I'll update the release notes in the change that stores the `chain_id` since it doesn't have an effect until then.

amnn:
> I'll update the release notes in the change that stores the `chain_id` since it doesn't have an effect until then.

This is not correct. The ingestion and streaming clients and services are part of the indexing framework's public API, i.e. they can be used as standalone components, so changes to their signatures are breaking changes for the indexing framework.

Update the release notes section on **this PR** to indicate that breaking change, and migration guidance, and coordinate with @jessiemongeon1 on review, as this PR has now landed without the necessary update to its description, so it is going to trigger notifications/workflows internally.

In future, consider any change to a `pub` type or function a potential breaking change. If in doubt, ask before landing, so that we can avoid the side channel to coordinate review on the release notes, thanks.

PR #25914, opened 2026-03-20
test-cluster: deflake wait_for_authenticator_state_update

evan-wall-mysten:
## Description 

When running `cargo nextest run -p sui-indexer-alt-e2e-tests` locally, I was getting intermittent `Timed out waiting for authenticator state update` errors. These were not happening when I ran the failing tests individually locally and I think do not happen enough to fail each retry in `ci`.

This PR does a one-time authenticator state check after `subscribe_transactions` is called in case the `AuthenticatorStateUpdate` transaction occurred before `subscribe_transactions` was called.

Additonally, the timeout waiting for GraphQL is increased to account for the possibility that `wait_for_authenticator_state_update` completes faster.

## Test plan 

Ran `cargo nextest run -p sui-indexer-alt-e2e-tests` locally and all tests pass consistently.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql_zklogin_tests.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql_zklogin_tests.rs
@@ -102,7 +102,7 @@ impl FullCluster {
         // Trigger an epoch change and wait until GraphQL sees Epoch 1
         onchain.trigger_reconfiguration().await;
         onchain.wait_for_authenticator_state_update().await;
-        tokio::time::timeout(Duration::from_secs(5), async {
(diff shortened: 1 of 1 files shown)


mystenmark:
i'm okay with this, but there is some chance this could introduce new flakiness, e.g. by returning early such that some, but not all, JWKs have been processed. But, we can give it a try and see if any flakiness emerges

PR #25917, opened 2026-03-20
forking-store: land store architecture as a skeleton

stefan-mysten:
## Description 

New traits and APIs for `forking-store` to enable forking tool. We went with having another store to not break main just yet, and we will merge this one with `sui-data-store` soon.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/forking-data-store/src/lib.rs
+++ b/crates/forking-data-store/src/lib.rs
@@ -16,7 +16,7 @@
 //!
 //! ## Store Implementations
 //!
-//! - [`stores::GraphQlStore`] - Remote GraphQL-backed store (mainnet/testnet/devnet/custom GraphQL
+//! - [`stores::GraphQLStore`] - Remote GraphQL-backed store (mainnet/testnet/devnet/custom GraphQL
 //!   endpoint)
 //! - [`stores::FileSystemStore`] - Persistent local disk cache
 //! - [`stores::InMemoryStore`] - Unbounded in-memory cache


crates/forking-data-store/src/lib.rs
```diff
@@ -0,0 +1,365 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Multi-tier caching data store for Sui blockchain data.
+//!
+//! This crate provides a flexible data store abstraction for retrieving and caching
```
dariorussi, line 6: I am not sure this module comment represents what is there, for instance I do not see anything mentioned for checkpoints.
Can we regenerate this comment with a good description of the module?

crates/forking-data-store/src/stores/forking.rs
```diff
@@ -0,0 +1,184 @@
...
+};
+
+/// A router that delegates each capability to a dedicated store.
+#[derive(Debug)]
+pub struct ForkingStore<Tx, Epoch, Obj, Ckpt> {
+    transactions: Tx,
+    epochs: Epoch,
+    objects: Obj,
+    checkpoints: Ckpt,
+}
```
dariorussi, line 28: do we need this generic declaration?
I am a bit confused as this feels like a very "unbounded" declaration but we know exactly what a `ForkingStore` is. One could argue that we may want to change the specific implementation of one of the store (e.g. add/remove an intermediate layer) and so we may want to switch easily the concrete implementation, but I am not sure we need this much abstraction?
We may need it for other reason?

crates/forking-data-store/src/stores/filesystem.rs
```diff
@@ -0,0 +1,222 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! File-system backed store skeleton.
+//!
+//! This module intentionally keeps the public shape of the intended store while
+//! deferring the actual persistence logic to a later commit.
+
```
dariorussi, line 71: I see a lot of todos (as we talked about) so it's good. Let's make sure to have a good module comment that reflects the structure of the file system for the different stores.
This is likely the only reusable module between replay and forking

crates/forking-store/src/stores/in_memory.rs
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! In-memory store skeleton.
```
dariorussi: here overall module comment is less important in my opinion and the shape of the implementation will tell us what is there (which maps), very self contained

crates/forking-data-store/src/stores/in_memory_lru.rs
```diff
@@ -0,0 +1,185 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
dariorussi, line 2: same as previous module comment idea, very private

PR #25919, opened 2026-03-20
refactor(display): RpcVisitor Writer -> Format + Meter

amnn:
## Description

Factor `RpcVisitor`'s `Writer` trait into separate `Format` and `Meter` traits. This allows us to consolidate the logic for formatting Move Values in JSON and Protobuf into single `Format` impls, and combine them with different meters based on need (unmetered, single-threaded, and shared meters are implemented as before). The previously coupled `Writer` trait caused the JSON implementation in particular to be duplicated many times, and made it so that the gRPC Display v2 integration ended up using the JSON cost model which was not ideal.

## Test plan

```
$ cargo nextest run \
  -p sui-display \
  -p sui-indexer-alt-graphql \
  -p sui-rpc-api \
  -p sui-rpc-resolver \
  -p sui-types \
  -j3
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-display/src/v2/error.rs
+++ b/crates/sui-display/src/v2/error.rs
@@ -299,8 +299,24 @@ impl fmt::Display for ExpectedSet {
 }
 
 impl From<RV::Error> for FormatError {
-    fn from(RV::Error: RV::Error) -> Self {
-        FormatError::Bcs(bcs::Error::Custom("unexpected type".to_string()))
+    fn from(error: RV::Error) -> Self {
+        match error {
+            RV::Error::Visitor(err) => err.into(),
+            RV::Error::Option(err) => err.into(),
+            RV::Error::Meter(err) => err.into(),
+            RV::Error::UnexpectedType => {
+                FormatError::Bcs(bcs::Error::Custom("unexpected type".to_string()))
+            }
+        }
+    }
+}
+
+impl From<RV::MeterError> for FormatError {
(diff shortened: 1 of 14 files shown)


crates/sui-types/src/proto_value.rs
```diff
@@ -270,7 +243,6 @@ pub(crate) mod tests {
     fn proto_value_to_json_value(proto: Value) -> serde_json::Value {
         match proto.kind {
             Some(Kind::NullValue(_)) | None => serde_json::Value::Null,
-            // Move doesn't support floats so for these tests can do a convert to u32
```
bmwill, line 273: why remove these comments?

amnn: Wasn't on purpose, I'll add it back, thanks for flagging. Codex really likes to strip comments for some reason, and I didn't spot it had gotten to this one.

bmwill: claude does too i've found

amnn: Before the issue used to be the opposite -- too many redundant comments -- and in response my understanding is that system prompts had phrasing that steered the agent away from doing that, so maybe that's what we're seeing.

PR #25923, opened 2026-03-20
[grpc] Fix LINKER_ERROR when rendering JSON for newly published packages

mystenmark:
## Summary

- Fix event/object JSON rendering for transactions that publish packages with types referenced in events or objects
- When a transaction publishes a package and emits events (or creates objects) using types from that package, the gRPC response rendering failed because the package wasn't yet in the backing store
- Introduces `OverlayPackageStore` that checks output_objects before falling back to the backing store

## Test plan

- [x] Added regression test `execute_transaction_publish_with_event_json` that:
  - Publishes a package that emits an event during init
  - Verifies `event.json.is_some()` for the emitted event
  - Test fails without fix, passes with fix
- [x] Added regression test `execute_transaction_publish_with_object_json` that:
  - Publishes a package that creates an object during init
  - Verifies `object.json.is_some()` for the created object
  - Test fails without fix, passes with fix
- [x] All RPC tests pass: `cargo simtest -p sui-e2e-tests --test rpc`

### DIFF
--- a/crates/sui-rpc-api/src/grpc/v2/ledger_service/get_object.rs
+++ b/crates/sui-rpc-api/src/grpc/v2/ledger_service/get_object.rs
@@ -16,6 +16,7 @@ use sui_rpc::proto::sui::rpc::v2::GetObjectResponse;
 use sui_rpc::proto::sui::rpc::v2::GetObjectResult;
 use sui_rpc::proto::sui::rpc::v2::Object;
 use sui_sdk_types::Address;
+use sui_types::full_checkpoint_content::ObjectSet;
 
 pub const MAX_BATCH_REQUESTS: usize = 1000;
 pub const READ_MASK_DEFAULT: &str = "object_id,version,digest";
@@ -128,5 +129,5 @@ fn get_object_impl(
             .ok_or_else(|| ObjectNotFoundError::new(object_id))?
     };
 
-    Ok(service.render_object_to_proto(&object, read_mask))
+    Ok(service.render_object_to_proto(&object, read_mask, &ObjectSet::default()))
 }
--- a/crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/mod.rs
+++ b/crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/mod.rs
@@ -127,10 +127,8 @@ pub async fn execute_transaction(
     } = executor.execute_transaction(request, None).await?;
 
     let executed_transaction = {
-        let events = read_mask
-            .subtree(ExecutedTransaction::EVENTS_FIELD)
-            .and_then(|mask| events.map(|events| service.render_events_to_proto(&events, &mask)));
-
+        // Build the objects set first so we can use it for event JSON rendering.
+        // This allows resolving types from packages that were just published in this transaction.
         let objects = {
             let mut objects = sui_types::full_checkpoint_content::ObjectSet::default();
             for o in input_objects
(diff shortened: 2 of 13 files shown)


crates/sui-types/src/storage/mod.rs
```diff
@@ -298,6 +298,33 @@ impl<S: ?Sized + BackingPackageStore> BackingPackageStore for &mut S {
...
+    pub fn new(overlay: &'a ObjectSet, backing: S) -> Self {
+        Self { overlay, backing }
+    }
+}
+
+impl<S: BackingPackageStore> BackingPackageStore for OverlayBackingPackageStore<'_, S> {
+    fn get_package_object(&self, package_id: &ObjectID) -> SuiResult<Option<PackageObject>> {
+        // First check the overlay for the package
+        for obj in self.overlay.iter() {
+            if &obj.id() == package_id && obj.is_package() {
```
mystenmark: if the object is not a package we should return whatever error the backing store would return in such a case rather than falling back

crates/sui-rpc-api/src/grpc/v2/ledger_service/get_object.rs
```diff
@@ -128,5 +128,9 @@ fn get_object_impl(
             .ok_or_else(|| ObjectNotFoundError::new(object_id))?
     };
 
-    Ok(service.render_object_to_proto(&object, read_mask))
+    Ok(service.render_object_to_proto(
+        &object,
+        read_mask,
+        &sui_types::full_checkpoint_content::ObjectSet::default(),
```
mystenmark: use imports instead of long qualifications (throughout the PR)

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/mod.rs
```diff
@@ -140,9 +138,15 @@ pub async fn execute_transaction(
             {
                 objects.insert(o);
             }
-            objects
+            std::sync::Arc::new(objects)
```
bmwill: does this need to be in an Arc?

crates/sui-types/src/storage/mod.rs
```diff
@@ -298,6 +298,41 @@ impl<S: ?Sized + BackingPackageStore> BackingPackageStore for &mut S {
     }
 }
 
+/// A BackingPackageStore that overlays objects on top of a backing store.
+/// This allows resolving packages from a set of objects (e.g., output objects from a transaction)
+/// before falling back to the backing store.
+pub struct OverlayBackingPackageStore<'a, S> {
+    overlay: &'a ObjectSet,
+    backing: S,
+}
```
bmwill, line 307: I feel like we must have some pre-existing implementation like this elsewhere in the code base...buy maybe not?

mystenmark: There's a similar pattern in `TemporaryStore::get_package_object` (sui-execution/v0/sui-adapter/src/temporary_store.rs:1042) which checks `self.written` first before falling back to the backing store. However, that implementation is tied to the `TemporaryStore` struct and has side effects (tracking `runtime_packages_loaded_from_db`), so it's not reusable as a standalone overlay. This new implementation is a simple, generic overlay that can wrap any `BackingPackageStore`.

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/mod.rs
mystenmark: Good catch - removed the Arc. It's only used by reference so no need for shared ownership.

PR #25925, opened 2026-03-20
[state sync] migrate to new bucket format

phoenix-o:
## Description 

migrates state sync to the new bucket format

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


phoenix-o:
@amnn: State sync currently uses the old ingestion framework. This PR migrates it(last callsite) and utilizes the new checkpoint format.
Only `IngestionClient` is used to download the checkpoints.
There are not many changes to the framework itself.
A new optional parameter is added `remote_store_extra_settings`. It's empty by default, so it does not change the behavior of any existing clients.
This allows dynamically passing extra settings to the underlying object store library for more precise configuration.
For instance here StateSync needs to pass down credentials for requester-pays behavior.
I only need this for AWS at the moment, but I updated GCP and Azure as well (can revert if needed)

amnn:
> Only `IngestionClient` is used to download the checkpoints.

If only `IngestionClient` is used and even within that you only need to be able to fetch checkpoints from an object store, then I'm still leaning towards not using the indexing framework for this, but just implementing a thin wrapper around `object_store`, and there's a lot hanging off that dependency that you do not need/use. I am also thinking of the fact that we want to move the indexing framework out of the mono-repo at some point and publish it to crates.io.

> For instance here StateSync needs to pass down credentials for requester-pays behavior.

Separately, I think this issue has come up before, so maybe it is worth solving in the framework, even if state sync doesn't use it:

- @nickvikeras I recall you needed to solve this problem for the analytics indexer, and you were able to do this without updating the framework -- could you explain what you did (and whether you have a path towards generalising it in the framework?)
- @rushrs what does the set-up look like for our indexers -- do they use requester pays buckets and if so how do they get configured? I believe we expose configuration via environment variables for this kind of thing, but I'm not sure we use it.

phoenix-o:
@amnn: sounds good, I’ll revert the changes to `IngestionClientArgs`. Just thought it might be helpful for more precise future configuration of underlying object store clients

crates/sui-network/Cargo.toml
```diff
@@ -35,7 +35,9 @@ sui-config.workspace = true
 sui-swarm-config.workspace = true
 sui-simulator.workspace = true
 sui-protocol-config.workspace = true
-sui-data-ingestion-core.workspace = true
+sui-indexer-alt-framework.workspace = true
```
amnn: Should be able to get rid of this now, right?

nickvikeras:
> > Only `IngestionClient` is used to download the checkpoints.
> 
> If only `IngestionClient` is used and even within that you only need to be able to fetch checkpoints from an object store, then I'm still leaning towards not using the indexing framework for this, but just implementing a thin wrapper around `object_store`, and there's a lot hanging off that dependency that you do not need/use. I am also thinking of the fact that we want to move the indexing framework out of the mono-repo at some point and publish it to crates.io.
> 
> > For instance here StateSync needs to pass down credentials for requester-pays behavior.
> 
> Separately, I think this issue has come up before, so maybe it is worth solving in the framework, even if state sync doesn't use it:
> 
>     * @nickvikeras I recall you needed to solve this problem for the analytics indexer, and you were able to do this without updating the framework -- could you explain what you did (and whether you have a path towards generalising it in the framework?)
> 
>     * @rushrs what does the set-up look like for our indexers -- do they use requester pays buckets and if so how do they get configured? I believe we expose configuration via environment variables for this kind of thing, but I'm not sure we use it.

This was for reading out of the parquet bucket, not the checkpoint bucket, but the code is here - 

https://github.com/MystenLabs/sui/blob/67116b973a27681a51282d05774888d27e9127f9/crates/sui-analytics-indexer/src/indexer.rs?plain=1#L151-L169

I think we want to do something similar in the indexing framework so we can enable requester pays on our public checkpoint buckets.

crates/sui-network/Cargo.toml
nickvikeras: I am a little confused at the state of the PR. It looks like Ashok's preference is to move off of the indexing framework's ingestion client and use object_store directly, but we are still using the ingest client in this revision.

phoenix-o: Yes, I’m still using the client. Only reverted the changes to the client itself (i.e. I construct object store instance outside)

amnn: At that point, why use the ingestion client? You are pulling in all the code for building indexers (and which will one day not even be in the mono-repo) but all you need is the naming scheme for fetching a checkpoint, right?

phoenix-o: the point is to minimize code duplication. We’ll need the same functionality in a couple of other places (i.e. sui-tool). Maybe it should live in some other common place. But ok, removing the `sui-indexer-alt-framework` dependency from this PR

crates/sui-network/src/state_sync/worker.rs
```diff
@@ -2,42 +2,144 @@
...
 use sui_types::storage::WriteStore;
+use sui_types::transaction::Transaction;
 
-pub(crate) struct StateSyncWorker<S>(pub(crate) S, pub(crate) Metrics);
+pub(crate) fn build_object_store(
+    ingestion_url: &str,
+    remote_store_options: Vec<(String, String)>,
+) -> Arc<dyn ObjectStore> {
+    let timeout_secs = 5;
+    let client_options = ClientOptions::new()
```
nickvikeras, line 36: Do we need to add .with_default_headers() to set the requester pays headers?

phoenix-o: Don’t think so. The current sui-data-ingestion-core client doesn’t set it and works with requester-pays buckets. But once accepted, I’ll state-sync the mainnet FN with the new binary and verify that it works

phoenix-o: yeah, spun up FN locally. It syncs from the S3 requester-pays bucket

PR #25933, opened 2026-03-21
[simulacrum] cleanup

stefan-mysten:
## Description 

This PR does two things:
- removes `store_dyn` function and switches to use `store -> &S` instead
- updates the comment around `system_state` in custom initialization function

## Test plan 

Existing CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/simulacrum/src/lib.rs
+++ b/crates/simulacrum/src/lib.rs
@@ -999,7 +999,7 @@ mod tests {
         assert_eq!(
             (transfer_amount as i64 - gas_paid) as u64,
             store::SimulatorStore::get_object(sim.store(), &gas_id)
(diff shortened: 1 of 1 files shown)


stefan-mysten:
> LGTM, thanks @stefan-mysten.
> 
> You mentioned in the previous PR that there were disadvantages to switching from returning a dyn trait to a reference to the underlying impl. Were you speaking in the abstract, or did you hit up against potential downsides in this specific case? I could not spot a potential downside in this case.

On the abstract. As of now, it all seems fine.

PR #25938, opened 2026-03-21
[consensus]  Decouple consensus config from core sui crates

omersadika:
## Description 

- ConsensusProtocolConfig was introduced to decouple consensus from ProtocolConfig (which pulls in sui-protocol-config and the Move VM). Made its API consistent with ProtocolConfig - private fields, getters, set_*_for_testing() setters, and a new() constructor for the Sui code.
- Also replaced AuthorityPublicKey with AuthorityIdentifier. It was wrapping a BLS12381PublicKey but only used for identity checks, never verification. Now it's just 96 raw bytes, with Sui passing the key bytes instead.

## Test plan 

- SUI_SKIP_SIMTESTS=1 cargo test -p consensus-config - 2 passed (including snapshot)
- SUI_SKIP_SIMTESTS=1 cargo test -p consensus-core  - 194 passed
- SUI_SKIP_SIMTESTS=1 cargo test -p sui-core -- scoring_decision - 1 passed
- cargo xclippy - clean

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/config/src/consensus_protocol_config.rs
+++ b/consensus/config/src/consensus_protocol_config.rs
@@ -11,7 +11,7 @@ pub struct ConsensusProtocolConfig {
     max_transactions_in_block_bytes: u64,
     max_num_transactions_in_block: u64,
     gc_depth: u32,
-    fastpath_enabled: bool,
+    transaction_voting_enabled: bool,
     num_leaders_per_round: Option<usize>,
     bad_nodes_stake_threshold: u64,
     always_accept_system_transactions: bool,
@@ -25,7 +25,7 @@ impl Default for ConsensusProtocolConfig {
             max_transactions_in_block_bytes: if cfg!(msim) { 256 * 1024 } else { 512 * 1024 },
             max_num_transactions_in_block: if cfg!(msim) { 8 } else { 512 },
             gc_depth: 0,
-            fastpath_enabled: false,
+            transaction_voting_enabled: false,
             num_leaders_per_round: None,
             bad_nodes_stake_threshold: 0,
             always_accept_system_transactions: false,
@@ -40,7 +40,7 @@ impl ConsensusProtocolConfig {
         max_transactions_in_block_bytes: u64,
         max_num_transactions_in_block: u64,
         gc_depth: u32,
-        fastpath_enabled: bool,
+        transaction_voting_enabled: bool,
         num_leaders_per_round: Option<usize>,
         bad_nodes_stake_threshold: u64,
         always_accept_system_transactions: bool,
@@ -51,7 +51,7 @@ impl ConsensusProtocolConfig {
             max_transactions_in_block_bytes,
             max_num_transactions_in_block,
             gc_depth,
-            fastpath_enabled,
+            transaction_voting_enabled,
             num_leaders_per_round,
             bad_nodes_stake_threshold,
             always_accept_system_transactions,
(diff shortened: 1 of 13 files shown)


consensus/config/src/consensus_protocol_config.rs
```diff
@@ -0,0 +1,144 @@
...
+            max_num_transactions_in_block,
+            gc_depth,
+            fastpath_enabled,
+            num_leaders_per_round,
+            bad_nodes_stake_threshold,
+            always_accept_system_transactions,
+        }
+    }
+
+    /// Returns a config matching `ProtocolConfig::get_for_max_version_UNSAFE()`
```
akichidis: This method is a bit different compared to the `get_for_max_version_UNSAFE` . With the later we always take the latest values as set for the "unknown" chain. This is now returning a static set of values as they are set here. I am ok with it, but the comparison with the `ProtocolConfig::get_for_max_version_UNSAFE()` might be a bit misleading.

omersadika: Make sense. I updated it!

consensus/config/src/crypto.rs
```diff
@@ -132,37 +133,62 @@ impl ProtocolKeySignature {
...
+/// Authority identifier is a raw bytes identity for an authority.
+/// It is only used for identity sanity checks and not for cryptographic verification.
+/// The bytes originate from the authority's BLS12381 public key on the Sui side.
+pub const AUTHORITY_IDENTIFIER_LENGTH: usize = 96;
 
-impl AuthorityPublicKey {
-    pub fn new(key: bls12381::min_sig::BLS12381PublicKey) -> Self {
-        Self(key)
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AuthorityIdentifier([u8; AUTHORITY_IDENTIFIER_LENGTH]);
```
mwtian: Let's call this AuthorityName to match the naming on Sui side.

consensus/config/src/consensus_protocol_config.rs
```diff
@@ -0,0 +1,144 @@
...
+/// struct so that `consensus-core` does not depend on `sui-protocol-config`
+/// (and transitively on the Move VM).
+#[derive(Clone, Debug)]
+pub struct ConsensusProtocolConfig {
+    protocol_version: u64,
+    max_transaction_size_bytes: u64,
+    max_transactions_in_block_bytes: u64,
+    max_num_transactions_in_block: u64,
+    gc_depth: u32,
+    fastpath_enabled: bool,
```
mwtian: Since we are renaming, let's rename this to `transaction_voting_enabled`.

omersadika: To rename fastpath_enabled to transaction_voting_enabled only on ConsensusProtocolConfig, right? or also in Sui's ProtocolConfig?

mwtian: Just in `ConsensusProtocolConfig` is fine.

consensus/config/src/crypto.rs
omersadika: Done

consensus/config/src/consensus_protocol_config.rs
omersadika: Done

PR #25939, opened 2026-03-23
[Block Steaming] Introduce the ObserverSubscriber

akichidis:
## Description 

This PR is introducing the `ObserverSubscriber`, the core component that allows an Observer node to connect to the Observer interface/server of another peer to stream blocks from.

The `ObserverSubscriber` is similar-ish to the `Subscriber` but it handles exclusively the streaming from observer server peers. 

One of the key differentiators is the mult-worker architecture to process the streamed blocks. From previous experimentation processing the blocks sequentially was creating a head of line blocking - mostly due to the block verification.

## Test plan 

CI

## Next
- https://github.com/MystenLabs/sui/pull/25940

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/observer_subscriber.rs
+++ b/consensus/core/src/observer_subscriber.rs
@@ -1,10 +1,16 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{sync::Arc, time::Duration};
+use std::{
+    sync::{
+        Arc,
+        atomic::{AtomicUsize, Ordering},
+    },
+    time::Duration,
+};
 
 use futures::StreamExt;
-use mysten_metrics::{monitored_mpsc, monitored_scope, spawn_monitored_task};
+use mysten_metrics::{monitored_scope, spawn_monitored_task};
 use parking_lot::Mutex;
 use tokio::{
     task::{JoinHandle, JoinSet},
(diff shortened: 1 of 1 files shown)


consensus/core/src/observer_subscriber.rs
```diff
@@ -0,0 +1,457 @@
...
+        network_client: Arc<C>,
+        observer_service: Arc<S>,
+        dag_state: Arc<parking_lot::RwLock<DagState>>,
+    ) -> Self {
+        Self {
+            context,
+            network_client,
+            observer_service,
+            dag_state,
+            subscriptions: Arc::new(Mutex::new(BTreeMap::new())),
```
mwtian: If subscribing to multiple peers are allowed, I think there should be a way to drop subscription to specific peers. Otherwise, Subscription should only allow 1 peer.

consensus/core/src/observer_subscriber.rs
```diff
@@ -0,0 +1,457 @@
...
+            // sequential processing would be too slow.
+            let num_workers = context.committee.size();
+            const WORKER_CHANNEL_CAPACITY: usize = 100;
+
+            // Spawn worker pool - each worker processes blocks independently
+            // Each worker gets its own channel to eliminate mutex contention
+            let mut workers = JoinSet::new();
+            let mut senders = Vec::new();
+
+            for worker_id in 0..num_workers {
```
mwtian: Instead of pre-spawn this number of workers, why not spawn a task to process each block received, which is similar to validator? You can keep an inflight counter for backpressure, which is more robust too.

akichidis: Hhm unless I am missing something isn't what I am doing here similar to validator? We pre-spawn a task for each validator to process their blocks https://github.com/MystenLabs/sui/blob/2a6083bafec7a791806866cea60f03fd2e1beb65/consensus/core/src/subscriber.rs#L82

As blocks will be very frequently/constantly processed, I've considered avoiding the cost of keep spawning tasks and instead have dedicated ones. Back-pressure will eventually kick in here as well once we don't have available workers to allocate blocks too. But you raise a good point on simplicity here.

consensus/core/src/observer_subscriber.rs
akichidis: I am thinking that we can keep the code more generic now, but on application level (the calling point) we'll always subscribe to one peer.

consensus/core/src/observer_subscriber.rs
mwtian: On validators, it is natural to run a connect and process loop against each peer validator, which has a fixed set. Here, using a fixed number of processing loops doesn't seem like the most suitable approach, when there can be high variance in block arrival rate. I don't think the cost of spawning tasks is that much higher than enqueueing items to a mpsc channel and having it polled, but the resulting logic is simpler.

consensus/core/src/observer_subscriber.rs
mwtian: Which PR has the callsite? I want to understand how the API is used right now.

akichidis: Yes , you can have a look here https://github.com/MystenLabs/sui/pull/25940/changes#diff-de28c5601b2995b3d78ba8c91e3c09cb99a57dfc88f67bca9f0c8d89821a04e3R479

In the future, once we leverage state sync to publish observer server availability, this can be call from another component. Again for now I've assumed a simple setup.

PR #25940, opened 2026-03-23
[Block Streaming] Extract proposing logic to separate component & refactor node type bootstrap

akichidis:
## Description 

Observer nodes do not propose blocks. `Core` currently uses lots of components that will be either unused in this case or we'll need to set them as `Option`. Although that would be ok, separating the `Proposer` logic and initialising it only for `Validator` nodes deemed more appropriate.

Also the PR is doing some lightweight refactoring on the way the `AuthorityNode` is initialised to support the case where an `Observer` should initialised.

## Test plan 

CI

---

## Next

- https://github.com/MystenLabs/sui/pull/25975

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/authority_service.rs
+++ b/consensus/core/src/authority_service.rs
@@ -363,7 +363,9 @@ impl<C: CoreThreadDispatcher> ValidatorNetworkService for AuthorityService<C> {
             let mut proposed_blocks =
                 dag_state.get_cached_blocks(self.context.own_index, last_received + 1);
             if proposed_blocks.is_empty() {
-                let last_proposed_block = dag_state.get_last_proposed_block();
+                let last_proposed_block = dag_state
+                    .get_last_proposed_block()
+                    .expect("A block should have been returned");
                 proposed_blocks = if last_proposed_block.round() > GENESIS_ROUND {
                     vec![last_proposed_block]
                 } else {
--- a/consensus/core/src/authority_service.rs
+++ b/consensus/core/src/authority_service.rs
@@ -365,7 +365,7 @@ impl<C: CoreThreadDispatcher> ValidatorNetworkService for AuthorityService<C> {
             if proposed_blocks.is_empty() {
                 let last_proposed_block = dag_state
                     .get_last_proposed_block()
-                    .expect("A block should have been returned");
+                    .expect("Last proposed block should be returned on validators");
                 proposed_blocks = if last_proposed_block.round() > GENESIS_ROUND {
                     vec![last_proposed_block]
                 } else {
--- a/consensus/core/src/context.rs
+++ b/consensus/core/src/context.rs
@@ -35,13 +35,20 @@ pub struct Context {
 impl Context {
     pub(crate) fn new(
         epoch_start_timestamp_ms: u64,
-        own_index: AuthorityIndex,
+        own_index: Option<AuthorityIndex>,
         committee: Committee,
         parameters: Parameters,
         protocol_config: ConsensusProtocolConfig,
         metrics: Arc<Metrics>,
         clock: Arc<Clock>,
     ) -> Self {
+        let own_index = if let Some(own_index) = own_index {
+            own_index
+        } else {
+            // If no index is provided, then this is an observer node. We assign a max index to it as a special value.
+            AuthorityIndex::MAX
+        };
+
         Self {
             epoch_start_timestamp_ms,
             own_index,
@@ -74,7 +81,7 @@ impl Context {
 
         let context = Context::new(
             0,
-            AuthorityIndex::new_for_test(0),
+            Some(AuthorityIndex::new_for_test(0)),
             committee,
             Parameters {
                 db_path: temp_dir.keep(),
@@ -115,6 +122,11 @@ impl Context {
     /// Returns true if this node is a validator (i.e., part of the committee).
     pub fn is_validator(&self) -> bool {
         self.committee.is_valid_index(self.own_index)
+    }
+
+    /// Returns true if this node is an observer (i.e., not part of the committee).
+    pub fn is_observer(&self) -> bool {
+        !self.is_validator()
     }
 }
 
--- a/consensus/core/src/core.rs
+++ b/consensus/core/src/core.rs
@@ -3,26 +3,21 @@
 
 use std::{
     collections::{BTreeMap, BTreeSet},
-    iter,
     sync::Arc,
-    time::Duration,
     vec,
 };
 
-use consensus_config::{AuthorityIndex, ProtocolKeyPair};
+use consensus_config::ProtocolKeyPair;
 #[cfg(test)]
-use consensus_config::{Stake, local_committee_and_keys};
-use consensus_types::block::{BlockRef, BlockTimestampMs, Round};
+use consensus_config::{AuthorityIndex, Stake, local_committee_and_keys};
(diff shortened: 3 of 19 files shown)


consensus/core/src/proposer.rs
```diff
@@ -0,0 +1,720 @@
+// Copyright (c) Mysten Labs, Inc.
```
akichidis, line 1: Most of the logic in this file has been transferred as-is from `Core`.

consensus/core/src/core.rs
```diff
@@ -96,28 +69,16 @@ pub(crate) struct Core {
...
-    // proposal or not.
-    ancestor_state_manager: AncestorStateManager,
-    // The round tracker will keep track of the highest received and accepted rounds
-    // from all authorities. It will use this information to then calculate the
-    // quorum rounds periodically which is used across other components to make
-    // decisions about block proposals.
-    round_tracker: Arc<RwLock<RoundTracker>>,
+    /// Block proposal engine for Validator nodes only.
+    /// Validators have a proposer to create blocks, Observers have None (they only receive blocks).
+    proposer: Option<Box<dyn Proposer>>,
```
akichidis, line 76: I was in between using an `Option` for the Proposer or creating a no-op `ObserverProposer` that would be passed here, to avoid the branching on checking the presence of the proposer. I've actually done both, but I've went with making the `Proposer` and option as I think in the end it represented semantically the intention better.

consensus/core/src/network/tonic_network.rs
```diff
@@ -764,6 +764,9 @@ impl TonicManager {
         // Calculate own address
         let own_address = if let Some(listen_addr) = &context.parameters.listen_address_override {
             listen_addr
+        } else if !context.is_validator() {
+            // Observer node - use a default local address since we're not in the committee. The port will be set separately during the server setup.
+            &Multiaddr::try_from("/ip4/0.0.0.0/tcp/0").unwrap()
```
mwtian: Why not use `observer_server_port`?

consensus/core/src/context.rs
```diff
@@ -35,13 +35,20 @@ pub struct Context {
...
         parameters: Parameters,
         protocol_config: ConsensusProtocolConfig,
         metrics: Arc<Metrics>,
         clock: Arc<Clock>,
     ) -> Self {
+        let own_index = if let Some(own_index) = own_index {
+            own_index
+        } else {
+            // If no index is provided, then this is an observer node. We assign a max index to it as a special value.
+            AuthorityIndex::MAX
```
mwtian, line 49: Why not pass AuthorityIndex::MAX or a const by the caller?

consensus/core/src/synchronizer.rs
```diff
@@ -810,7 +810,7 @@ where
                 let mut retry_delay_step = Duration::from_millis(500);
                 'main:loop {
                     if context.committee.size() == 1 {
-                        highest_round = dag_state.read().get_last_proposed_block().round();
+                        highest_round = dag_state.read().get_last_proposed_block().expect("A block should have been returned").round();
```
mwtian: Maybe "Last proposed block should be returned on validators"?

consensus/core/src/authority_service.rs
```diff
@@ -359,7 +359,9 @@ impl<C: CoreThreadDispatcher> ValidatorNetworkService for AuthorityService<C> {
             let mut proposed_blocks =
                 dag_state.get_cached_blocks(self.context.own_index, last_received + 1);
             if proposed_blocks.is_empty() {
-                let last_proposed_block = dag_state.get_last_proposed_block();
+                let last_proposed_block = dag_state
+                    .get_last_proposed_block()
+                    .expect("A block should have been returned");
```
mwtian: Maybe "Last proposed block should be returned on validators"?

mwtian (commented):
Still reviewing but here are some comments.

consensus/core/src/observer_subscriber.rs
mwtian, line 66: Claude pointed out that this should be re-computed each time a new subscription attempt is made.

consensus/core/src/observer_subscriber.rs
```diff
@@ -162,56 +209,39 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
...
                             }
-
-                            // Decrement counter when done
-                            counter.fetch_sub(1, Ordering::AcqRel);
-                        });
+                            next_worker = (next_worker + 1) % num_workers;
+                        }
 
                         // Reset retries when a block is received.
                         retries = 0;
```
mwtian, line 217: backoff needs to reset as well.

mwtian (approved):

consensus/core/src/observer_subscriber.rs
```diff
@@ -146,11 +150,54 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
...
+                let (block_tx, mut block_rx) = monitored_mpsc::channel(
+                    &format!("subscriber_blocks_peer_worker_{}", worker_id),
+                    WORKER_CHANNEL_CAPACITY,
+                );
+                senders.push(block_tx);
+
+                let observer_service = observer_service.clone();
+
+                let peer_cloned = peer.clone();
+                workers.spawn(async move {
```
mwtian: IIUC this logic will be merged from a previous PR and removed from this PR.

akichidis: Exactly, already updated

consensus/core/src/proposer.rs
```diff
@@ -0,0 +1,722 @@
...
+                        verified_block
+                            .timestamp_ms()
+                            .saturating_sub(last_proposed_block.timestamp_ms()),
+                    )
+                    .as_secs_f64(),
+                );
+        }
+
+        // Add the block directly to DagState (skipping BlockManager since our own blocks cannot be suspended)
+        self.dag_state.write().accept_block(verified_block.clone());
```
akichidis, line 601: @mwtian just to make you aware that I had to use the `dag_state` directly here instead of the `BlockManager` , as the BlockManager is wired in the Core. Since for newly created blocks we don't expect to have any dependency already , I don't see reasons for this to create any functional changes. The only side effect I see is that for the metrics that the BlockManager exposes, for this case to not get updated.

consensus/core/src/context.rs
akichidis: My intention was to let the Context handle the special case rather than the caller providing it - so making this more of a central decision.

consensus/core/src/observer_subscriber.rs
akichidis: actually we probably need to fix that for the`subscriber.rs` as well

consensus/core/src/proposer.rs
mwtian: Sounds good. In general, I think BlockManager and CommitObserver should communication their output to Core instead of writing / sending their output downstream directly.

akichidis: If we respect the single threaded processing then yes, `BlockManager` (at least) could communicate its output back to `Core` , and then `Core` command the actual writing etc.

PR #25950, opened 2026-03-23
Treat on-chain validators as trusted peers in discovery

aschran:
## Summary

- **Chain peers are now trusted peers.** When a validator receives another validator's on-chain P2P address (`AddressSource::Chain`), that peer is treated equivalently to a configured seed/allowlisted peer for discovery purposes. This means Discovery-gossiped addresses can override Chain addresses, validator NodeInfo is persisted across restarts, and validators share Trusted-access-type peer info with each other.

- **Chain addresses restricted to validators only.** Fullnodes no longer receive `AddressSource::Chain` addresses for validators. Previously every node attempted High-affinity connections to all validators, which were rejected by validators with `max_concurrent_connections: 0`. Fullnodes continue to reach validators via seed_peers and discovery gossip.

- **Cached Discovery addresses applied at startup.** On restart, previously-cached NodeInfo (with Discovery-sourced addresses) is loaded before Chain addresses arrive. A new code path injects cached Discovery addresses when a Chain peer is first registered, ensuring the correct address priority is applied from the start rather than briefly connecting at the Chain address before the override.

- **Tick-based dialing skips EndpointManager-managed peers.** Peers whose addresses are tracked in `peer_addresses` (via EndpointManager) are excluded from `handle_tick`'s dial loop, preventing stale gossipped addresses from overriding the address priority system.

## Test plan

- Unit tests for chain peer registration, Discovery override of Chain addresses, chain peer removal on empty address, chain peer immunity from failure reports, and allowlisted peer dialing behavior
- Simtest (`test_network_resilience_with_incorrect_discovery_addresses`): 4-validator cluster where 3 validators advertise bad Discovery addresses, verifying star topology after restart and functional resilience (randomness transaction still works)

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-node/src/lib.rs
```diff
@@ -2340,12 +2344,15 @@ impl SpawnOnce {
...
 fn update_peer_addresses(
     config: &NodeConfig,
     endpoint_manager: &EndpointManager,
     epoch_start_state: &EpochStartSystemState,
 ) {
+    if config.consensus_config().is_none() {
+        return;
+    }
     for (peer_id, address) in
         epoch_start_state.get_validator_as_p2p_peers(config.protocol_public_key())
```
alex-mysten: Should we remove validators that are no longer part of the committee from chain_peers?

crates/sui-network/src/discovery/mod.rs
```diff
@@ -68,9 +68,9 @@ pub enum DiscoveryMessage {
     ReceivedNodeInfo {
         peer_info: Box<SignedVersionedNodeInfo>,
     },
-    /// A spawned peer-query task discovered updated info for a configured peer,
+    /// A spawned peer-query task discovered updated info for a trusted peer,
     /// signaling the event loop to update the stored peer addresses.
-    ConfiguredPeersUpdated,
+    TrustedPeersUpdated,
```
alex-mysten, line 73: We are changing the nomenclature because this now refers to any committee validator, not just seeds? Just confirming

crates/sui-network/src/discovery/mod.rs
```diff
@@ -446,7 +448,7 @@ impl DiscoveryEventLoop {
     }
 
     fn handle_peer_failure_report(&mut self, peer_id: PeerId) {
-        if self.configured_peers.contains_key(&peer_id) {
+        if self.is_trusted_peer(&peer_id) {
             info!(?peer_id, "ignoring failure report for configured peer");
```
alex-mysten: Nit: log message string change to match is_trusted

crates/sui-network/src/discovery/mod.rs
```diff
@@ -583,6 +594,37 @@ impl DiscoveryEventLoop {
             } else {
                 source_map.insert(source, addresses);
             }
+
+            // When a Chain address arrives, check if there are cached Discovery
+            // P2P addresses in known_peers_v2 that should take priority. This
+            // handles the startup race where cached NodeInfo is loaded before
+            // chain_peers is populated (so update_known_peers_versioned couldn't
```
alex-mysten, line 601: Is there an opportunity to sequence this such that we avoid the special-case handling for the race?

crates/sui-node/src/lib.rs
```diff
@@ -1619,6 +1619,10 @@ impl SuiNode {
         self.state.clone()
     }
 
+    pub fn connection_monitor_status(&self) -> &Arc<ConnectionMonitorStatus> {
```
alex-mysten: test-only?

crates/sui-network/src/discovery/mod.rs
aschran: yes, I renamed configured->trusted since the validators from the committee are not explicitly configured in the node configs but still considered trusted

crates/sui-node/src/lib.rs
aschran: addressed

crates/sui-network/src/discovery/mod.rs
aschran: not easily, unfortunately, bc of circular dependencies between components

bmwill:
>Chain addresses restricted to validators only. Fullnodes no longer receive AddressSource::Chain addresses for validators. Previously every node attempted High-affinity connections to all validators, which were rejected by validators with max_concurrent_connections: 0. Fullnodes continue to reach validators via seed_peers and discovery gossip.

This is a functional change that will reduce some load on Validators in testnet and mainnet but can cause some issues with devnet and other test networks as it requires a bit more explicit config setup to make sure things communicate properly. What would your thoughts be on restricting this behavior change to testnet and mainnet while retaining it on devnet/localnet so that those environments don't require any additional config tweaking?

aschran:
> > Chain addresses restricted to validators only. Fullnodes no longer receive AddressSource::Chain addresses for validators. Previously every node attempted High-affinity connections to all validators, which were rejected by validators with max_concurrent_connections: 0. Fullnodes continue to reach validators via seed_peers and discovery gossip.
> 
> This is a functional change that will reduce some load on Validators in testnet and mainnet but can cause some issues with devnet and other test networks as it requires a bit more explicit config setup to make sure things communicate properly. What would your thoughts be on restricting this behavior change to testnet and mainnet while retaining it on devnet/localnet so that those environments don't require any additional config tweaking?

I followed up with John and he said the required tweaks should be minimal. In general I think we should minimize the behavior differences between mainnet and test envs wherever possible.

bmwill:
yes but this auto breaks everyone in the ecosystem who was depending on this behavior for test environments so we need to take that into consideration.

aschran:
> yes but this auto breaks everyone in the ecosystem who was depending on this behavior for test environments so we need to take that into consideration.

IMO we can't be hamstrung into never changing this forever, only because some people might need to add one validator as a seed peer in their test network. Happy to start a broader discussion if you feel strongly the other way, though?

mystenmark (approved):
This is all good, given the rollout plan we discussed earlier:
- First rollout to mysten nodes while config is disabled (to test that system works as expected while disabled, i.e. we did not regress the current behavior), followed by wider release to avoid running different binaries.
- Then enable configs on mysten nodes and evaluate change for 1-2 cycles before enabling everywhere.

PR #25954, opened 2026-03-23
[benchmark] Merge gas coins on startup to prevent fragmentation exhaustion

ebmifa:
## Description

The stress benchmark binary splits gas coins for batch payment transactions but never merges them back. Over hours of continuous operation at 50 QPS with `batch_payment=1`, individual coin fragments drop below the per-transaction gas budget threshold (~0.1 SUI), causing 20%+ transaction failure rates.

This was observed on devnet where the stress pod runs continuously. The existing code had a TODO at `benchmark_setup.rs:116`: *"Merge all owned gas objects into one and use that as the primary gas object"*.

### Fix

On startup (or restart), merge all owned gas coins into a single primary coin using `MergeCoins` PTB with `Argument::GasCoin` before initializing the bank. This handles:
- **Fresh start after genesis**: multiple genesis coins → one large coin
- **Restart after fragmentation**: many small coins → one large coin

Key implementation details:
- Uses `Argument::GasCoin` as the merge target (not the gas object as a PTB input) to avoid `DuplicateObjectRefInput` validation errors
- Batches merges in groups of 500 coins to stay within the `max_arguments` protocol limit (512)
- Each batch updates the gas object reference from transaction effects for correctness
- Matches the established merge pattern used in `sui-rosetta` and other crates

The stress pod's restart-on-failure behavior naturally self-heals: when gas exhaustion causes persistent failures → pod restarts → merge runs → all coins consolidated → operations resume.

### Evidence

Devnet heartbeat failures over Mar 20-23:
```
TPS = 40, num_error_tx = 100, no_gas = 0
Balance of gas object 90210020 is lower than the needed amount: 100788000
```

CI/staging (which don't use `batch_payment`) are unaffected.

## Test plan

- [x] `cargo check -p sui-benchmark` passes
- [x] `cargo clippy -p sui-benchmark` clean
- [x] `cargo fmt -p sui-benchmark` clean
- [x] Verified `Argument::GasCoin` pattern matches established usage in `sui-rosetta` (e.g. `stake.rs`, `pay_sui.rs`)
- [x] Verified `get_owned_objects` only returns `GasCoin`-typed objects (filters with `GasCoin::type_()`) — no risk of merging non-coin objects
- [x] Verified batching stays within `max_arguments` (512) and `max_input_objects` (2048) protocol limits
- [x] Confirmed no impact on downstream `BenchmarkBank` — it only consumes a single `(ObjectRef, addr, keypair)` tuple
- [x] Change is fully isolated to `sui-benchmark` crate — no protocol, validator, RPC, or framework code affected
- [ ] Verify on devnet after next deploy: stress pod should merge coins on startup and run without gas exhaustion

### DIFF
--- a/crates/sui-benchmark/src/benchmark_setup.rs
+++ b/crates/sui-benchmark/src/benchmark_setup.rs
@@ -157,11 +157,7 @@ impl BenchmarkSetup {
                     let mut builder = ProgrammableTransactionBuilder::new();
                     let coin_args: Vec<_> = batch
                         .iter()
-                        .map(|coin| {
-                            builder
-                                .obj(ObjectArg::ImmOrOwnedObject(*coin))
-                                .unwrap()
-                        })
+                        .map(|coin| builder.obj(ObjectArg::ImmOrOwnedObject(*coin)).unwrap())
                         .collect();
                     builder.command(Command::MergeCoins(Argument::GasCoin, coin_args));
                     let pt = builder.finish();


stefan-mysten (approved):
LGTM

stefan-mysten:
We probably want to move to address balances, but lgtm for now.

PR #25957, opened 2026-03-23
GRPC simulate: prepend coin reservation when GasCoin used with address balance

mystenmark:
## Summary

When simulating a transaction via the GRPC API that:
1. Has gas payment already provided
2. Uses `Argument::GasCoin` in its PTB
3. Sender has positive address balance

We now prepend a coin reservation to the gas payment array. This gives the execution layer access to the combined value of coins + address balance through the GasCoin argument, enabling scenarios where neither source alone is sufficient.

**Key changes:**
- Added `maybe_prepend_coin_reservation()` function in simulate module
- Creates coin reservation for the sender's AB amount
- Prepends it to existing gas payment
- Sets `ValidDuring` expiration for AB usage

## Test plan

- [x] `test_has_ab_has_coins_uses_gas_coin` - Verifies coin reservation is prepended when GasCoin used with AB
- [x] `test_has_ab_has_coins_no_gas_coin` - Verifies no coin reservation when GasCoin not used
- [x] `test_has_ab_no_coins` - Basic AB test
- [x] `test_no_ab_has_coins` - Traditional coin payment (no coin reservation)
- [x] `test_insufficient_funds` - Error handling for insufficient funds
- [x] `test_protocol_config_disabled` - Fallback to traditional behavior
- [x] `test_combined_ab_and_coins_needed` - Demonstrates combining AB + coins when neither alone is sufficient

All tests simulate AND execute the transaction to verify validity.

### DIFF
--- a/crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
+++ b/crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
@@ -135,12 +135,16 @@ async fn test_has_ab_has_coins_uses_gas_coin() {
     );
 
     // Verify the entire address balance is reserved, not just the gas budget
+    // Note: The actual balance may be slightly less than ab_amount due to gas
+    // consumed during the funding transaction
     let parsed_digest = ParsedDigest::try_from(first_payment.2)
         .expect("Should be able to parse coin reservation digest");
-    assert_eq!(
-        parsed_digest.reservation_amount(),
-        ab_amount,
-        "Coin reservation should reserve the entire address balance"
+    let reservation_amount = parsed_digest.reservation_amount();
+    assert!(
+        reservation_amount >= ab_amount - 100_000_000, // Allow up to 0.1 SUI for gas
+        "Coin reservation should reserve nearly the entire address balance, got {} (expected ~{})",
+        reservation_amount,
+        ab_amount
     );
 
     // Execute the simulated transaction to verify it's valid


crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
```diff
@@ -0,0 +1,589 @@
...
+use sui_types::{
+    base_types::SuiAddress,
+    coin_reservation::ParsedDigest,
+    effects::TransactionEffectsAPI,
+    programmable_transaction_builder::ProgrammableTransactionBuilder,
+    transaction::{Argument, TransactionData, TransactionDataAPI},
+};
+use test_cluster::addr_balance_test_env::TestEnvBuilder;
+
+const MIST_PER_SUI: u64 = 1_000_000_000;
```
mystenmark: can this constant should be imported from gas_coin.rs?

crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
```diff
@@ -0,0 +1,589 @@
...
+#[sim_test]
+async fn test_has_ab_has_coins_uses_gas_coin() {
+    let test_env = TestEnvBuilder::new()
+        .with_proto_override_cb(Box::new(|_, mut cfg| {
+            cfg.create_root_accumulator_object_for_testing();
+            cfg.enable_accumulators_for_testing();
+            cfg.enable_coin_reservation_for_testing();
+            cfg.enable_address_balance_gas_payments_for_testing();
+            cfg.enable_coin_reservation_for_testing();
+            cfg.enable_address_balance_gas_payments_for_testing();
```
mystenmark: duplicate enable calls

crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
```diff
@@ -0,0 +1,589 @@
...
+        "Gas payment should not be empty when coins exist"
+    );
+
+    // First element should be a coin reservation (identified by magic in digest)
+    let first_payment = &gas_payment[0];
+    assert!(
+        ParsedDigest::is_coin_reservation_digest(&first_payment.2),
+        "First gas payment should be a coin reservation, got digest: {:?}",
+        first_payment.2
+    );
```
mystenmark, line 135: we also need to verify that the entire address balance is reserved by the coin reservation, not just the gas budget

crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
```diff
@@ -0,0 +1,589 @@
...
+
+    let response = result.unwrap();
+    assert!(
+        response.transaction.effects.status().is_ok(),
+        "Expected successful execution, got: {:?}",
+        response.transaction.effects.status()
+    );
+
+    // When GasCoin is not used, prefer AB if sufficient, else coins.
+    // No coin reservation needed since user doesn't need access to combined balance.
```
mystenmark: This comment describes two cases but only tests one. In this test, we should build two transactions. One of which can be satisfied by address balance alone, the other of which has a higher budget and requires both AB and coins

crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
```diff
@@ -0,0 +1,589 @@
...
+
+// =============================================================================
+// Test 3: Has AB + NO coins
+// Expected: Pure AB payment (empty gas_data.payment + expiration)
+// =============================================================================
+
+#[sim_test]
+async fn test_has_ab_no_coins() {
+    // This test is tricky: we need a sender with AB but no coins.
+    // We'll use address balance gas to achieve this.
```
mystenmark: this test case is not tricky. just make a new address and fund it with an AB transfer, then do the remainder of the test using that address

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -153,6 +153,13 @@ pub fn simulate_transaction(
         }
     }
 
+    // When gas payment is provided but the sender has address balance and uses GasCoin,
+    // we need to prepend a coin reservation to give access to the combined balance.
+    // This is done regardless of do_gas_selection since it affects transaction semantics.
```
bmwill: Can you elaborate on this a bit? We've historically tried to avoid making changes (adding inputs, etc) to the provided transaction unless explicitly asked to do so. For example if gas objects are provided we skip doing gas selection and opt to not mess with what has already been provided.

I think its fine to adjust and change semantics I just want to better understand the motivation here.

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -361,6 +368,101 @@ fn round_up_to_nearest(value: u64, step: u64) -> u64 {
...
+        )?;
+
+    let reservation =
+        ParsedObjectRefWithdrawal::new(*accumulator_obj_id.inner(), current_epoch, ab_value);
+    let coin_reservation = reservation.encode(SequenceNumber::new(), service.chain_id);
+
+    // Prepend coin reservation to existing payment
+    let mut new_payment = vec![coin_reservation];
+    new_payment.extend(transaction.gas_data().payment.clone());
+    transaction.gas_data_mut().payment = new_payment;
```
bmwill:
```suggestion
    transaction.gas_data_mut().payment.insert(0, coin_reservation);
```

This is maybe slightly more simple

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -361,6 +368,101 @@ fn round_up_to_nearest(value: u64, step: u64) -> u64 {
...
+
+    // Set expiration for AB usage
+    *transaction.expiration_mut() = TransactionExpiration::ValidDuring {
+        min_epoch: Some(current_epoch),
+        max_epoch: Some(current_epoch.saturating_add(1)),
+        min_timestamp: None,
+        max_timestamp: None,
+        chain: service.chain_id,
+        nonce: rand::random(),
+    };
```
bmwill: Do we want to check if expiration is already set to a valid `ValidDuring` value?

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -361,6 +368,101 @@ fn round_up_to_nearest(value: u64, step: u64) -> u64 {
...
+                RpcError::new(
+                    tonic::Code::Internal,
+                    format!("Failed to get accumulator object ID: {e}"),
+                )
+            },
+        )?;
+
+    let reservation =
+        ParsedObjectRefWithdrawal::new(*accumulator_obj_id.inner(), current_epoch, ab_value);
+    let coin_reservation = reservation.encode(SequenceNumber::new(), service.chain_id);
```
bmwill: Do we also want to check if we already have provided a reservation gas object ref?

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -361,6 +368,101 @@ fn round_up_to_nearest(value: u64, step: u64) -> u64 {
...
+    // Only proceed if GasCoin is used
+    if !gas_coin_used {
+        return Ok(());
+    }
+
+    let address_balance = reader
+        .lookup_address_balance(owner, GAS::type_())
+        .map(|balance| {
+            let reserved_sui = transaction
+                .get_funds_withdrawals()
```
bmwill: This doesn't take into account gas object fund reservation right?

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -361,6 +368,101 @@ fn round_up_to_nearest(value: u64, step: u64) -> u64 {
     }
 }
 
+/// When gas payment is already provided, check if we need to prepend a coin reservation.
+/// This is needed when the sender has address balance AND the transaction uses GasCoin.
+fn maybe_prepend_coin_reservation(
```
bmwill: What are the semantics of when this succeeds? are all coins smashed and inserted into the user's address balance at the end of the txn?

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
bmwill: Or we could do some cleverness with [swap_remove](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.swap_remove) if we want to golf this ;) 

This is a nit though so feel free to ignore.

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
mystenmark: ah, this wasn't actually the intent, will fix.

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
mystenmark: Yes, it causes gas smashing, and if the first gas coin listed is actually a coin reservation, then the real coins are destroyed and smashed into the address balance.

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
mystenmark: wow good catch, it does not but it should. sending another PR for fixes + tests to this

PR #25959, opened 2026-03-24
[Move] Bulletproofs range proof as native function

jonas-lj:
## Description 

Add bulletproofs range proof as a native function.

## Test plan 

Unit tests

---

## Release notes

- [x] Protocol: Add a new native function, **verify_bulletproof_ristretto255_internal** and enable on devnet.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-framework/packages/sui-framework/sources/crypto/bulletproofs.move
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+module sui::bulletproofs;
```
benr-ml: rangeproofs?

crates/sui-framework/packages/sui-framework/sources/crypto/bulletproofs.move
```diff
@@ -0,0 +1,30 @@
...
+const EInvalidProof: u64 = 1;
+const EInvalidCommitment: u64 = 2;
+const EInvalidRange: u64 = 3;
+const EInvalidBatchSize: u64 = 4;
+const EUnsupportedVersion: u64 = 5;
+
+/// Verify a range proof over the Ristretto255 curve that all committed values are in the range [0, 2^bits).
+/// Currently, the only supported version is 0 which corresponds to the original Bulletproofs construction (https://eprint.iacr.org/2017/1066.pdf).
+/// In the future, we may add support for newer versions of Bulletproofs, such as Bulletproofs+ or Bulletproofs++.
+public fun verify_range_proof_ristretto255(proof: &vector<u8>, bits: u8, commitments: &vector<Element<ristretto255::Point>>, version: u8): bool {
```
benr-ml:
```suggestion
public fun verify_bulletproofs_ristretto255(proof: &vector<u8>, bits: u8, commitments: &vector<Element<ristretto255::Point>>, version: u8): bool {
```

benr-ml: anything to say about the format of `proof`?

crates/sui-protocol-config/src/lib.rs
```diff
@@ -1704,6 +1708,9 @@ pub struct ProtocolConfig {
     group_ops_ristretto_scalar_div_cost: Option<u64>,
     group_ops_ristretto_point_div_cost: Option<u64>,
 
+    bulletproofs_ristretto255_verify_base_cost: Option<u64>,
+    bulletproofs_ristretto255_verify_cost_per_bit: Option<u64>,
```
benr-ml: we also need the cost per commitment, no?

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    let Ok(proof) = bcs::from_bytes::<RangeProof>(&proof.as_bytes_ref()?) else {
+        return Ok(NativeResult::err(context.gas_used(), INVALID_PROOF));
+    };
+
+    let Ok(range) = range_from_bits(range_bits).map_err(|_| InvalidInput) else {
+        return Ok(NativeResult::err(context.gas_used(), INVALID_RANGE));
+    };
+
+    let length = commitments
```
benr-ml: should we abort on 0?

crates/sui-protocol-config/src/lib.rs
jonas-lj: The performance depends on the number of commitments times the range size, so I think this way is simpler

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: Yea. I think that'll happen already since it's not a power of two, but we can definitely be specific

crates/sui-protocol-config/src/lib.rs
benr-ml: so bulletproofs_ristretto255_verify_base_cost = cost per commitment?

jonas-lj: No, the base cost is just a small base fee. The cost_per_bit is per bit per commitment. So 4 proofs of < 2^16 will cost base_cost + 4x16xcost_per_bit. I'll add an example to the docs to make this clea.

crates/sui-framework/packages/sui-framework/sources/crypto/bulletproofs.move
jonas-lj: Added a reference to the dalek implementation.

crates/sui-framework/packages/sui-framework/sources/crypto/rangeproofs.move
```diff
@@ -0,0 +1,41 @@
...
+const EInvalidRange: u64 = 3;
+#[allow(unused_const)]
+const EInvalidBatchSize: u64 = 4;
+const EUnsupportedVersion: u64 = 5;
+
+/// Verify a range proof over the Ristretto255 curve that all committed values are in the range [0, 2^bits).
+/// Currently, the only supported version is 0 which corresponds to the original Bulletproofs construction (https://eprint.iacr.org/2017/1066.pdf).
+/// In the future, we may add support for newer versions of Bulletproofs, such as Bulletproofs+ or Bulletproofs++.
+/// 
+/// The format of the proof follows the specifications from https://github.com/dalek-cryptography/bulletproofs/blob/be67b6d5f5ad1c1f54d5511b52e6d645a1313d07/src/range_proof/mod.rs#L59-L76 serialized using BCS.
```
benr-ml: let's make sure they use our format for scalar and points

crates/sui-protocol-config/src/lib.rs
```diff
@@ -1704,6 +1708,9 @@ pub struct ProtocolConfig {
     group_ops_ristretto_scalar_div_cost: Option<u64>,
     group_ops_ristretto_point_div_cost: Option<u64>,
 
+    verify_bulletproofs_ristretto255_base_cost: Option<u64>,
+    verify_bulletproofs_ristretto255_cost_per_bit: Option<u64>,
```
benr-ml:
```suggestion
    verify_bulletproofs_ristretto255_cost_per_bit_and_commitment: Option<u64>,
```

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4705,6 +4719,13 @@ impl ProtocolConfig {
                     cfg.execution_version = Some(4);
                     cfg.feature_flags.address_balance_gas_reject_gas_coin_arg = false;
                     cfg.feature_flags.merge_randomness_into_checkpoint = true;
+
+                    // Enable bulletproofs range proofs on devnet
+                    cfg.verify_bulletproofs_ristretto255_base_cost = Some(100);
```
benr-ml: TODO to update those

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -0,0 +1,147 @@
...
+                    "verify_bulletproofs_ristretto255_base_cost not available".to_string()
+                )
+            )?
+    );
+
+    let commitments = pop_arg!(args, VectorRef);
+    let range_bits = pop_arg!(args, u8);
+    let proof = pop_arg!(args, VectorRef);
+
+    let Ok(proof) = bcs::from_bytes::<RangeProof>(&proof.as_bytes_ref()?) else {
```
benr-ml: we don't charge here based on the proof length, so one can call us with a very long string and not pay for the deserialization costs. can we limit the proof length somehow? maybe based on the number of commitments and bits?

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -0,0 +1,147 @@
...
+        return Ok(NativeResult::err(context.gas_used(), INVALID_RANGE));
+    };
+
+    let length = commitments
+        .len(&Type::Vector(Box::new(Type::U8)))?
+        .value_as::<u64>()?;
+
+    // The performance is linear in the product of length and range bits, so it is computed as base_cost + cost_per_bit * length * range_bits. 
+    let total_bits = length * range_bits as u64;
+    if length == 0 || !length.is_power_of_two() || total_bits > MAX_TOTAL_BITS {
```
benr-ml, line 97: hmm, can we verify a batch of 3?

jonas-lj: No, the batch size has to be a power of two.

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: Yes, but we know what the size of the proof should be, so we could just fail before even trying to deserialize. Perhaps that already happens.

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: We could in that case use a batch of 4 and just use trivial commitment for the last one?

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: A proof can at most be 800 bytes, so let's set a hard limit on that before the deserialization.

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: I've added a comment in the docs about padding with trivial commitments in that case.

crates/sui-framework/packages/sui-framework/sources/crypto/rangeproofs.move
```diff
@@ -0,0 +1,51 @@
...
+/// Verify a range proof over the Ristretto255 curve that all committed values are in the range [0, 2^bits).
+/// Currently, the only supported version is 0 which corresponds to the original Bulletproofs construction (https://eprint.iacr.org/2017/1066.pdf).
+/// In the future, we may add support for newer versions of Bulletproofs, such as Bulletproofs+ or Bulletproofs++.
+///
+/// The format of the proof follows the specifications from https://github.com/dalek-cryptography/bulletproofs/blob/be67b6d5f5ad1c1f54d5511b52e6d645a1313d07/src/range_proof/mod.rs#L59-L76.
+///
+/// The commitments are Pedersen commitments to the values used in the proof. 
+/// The number of commitments must be a power of two, but if needed, the input to the prover can be padded with trivial commitments to zero.
+///
+/// The number of commitments times bits can be at most 256.
```
benr-ml: why is 256 the limit? for 16 bit chunks, we get 256/16/4 = 4 encrypted balances, which is not a lot for batch transfers, no?

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    let result = proof.verify_batch(&commitments, &range, &mut thread_rng());
+
+    Ok(NativeResult::ok(
+        context.gas_used(),
+        smallvec![Value::bool(result.is_ok())],
+    ))
+}
+
+fn range_from_bits(bits: u8) -> FastCryptoResult<Range> {
```
benr-ml, line 139: we should document in Move that the number of bits should be a power of 2 as well

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -0,0 +1,156 @@
...
+    let commitment_bytes = (0..length)
+        .map(|i| {
+            commitments
+                .borrow_elem(i as usize, &Type::Vector(Box::new(Type::U8)))
+                .and_then(|reference| reference.value_as::<VectorRef>())
+                .and_then(|v| Ok(v.as_bytes_ref()?.to_vec()))
+        })
+        .collect::<PartialVMResult<Vec<Vec<u8>>>>()?;
+
+    let Ok(commitments) = commitment_bytes
```
benr-ml: aren't all the error cases here are not user errors but a bug in our impl (and thus should return PartialVMResult)?

crates/sui-framework/packages/sui-framework/sources/crypto/rangeproofs.move
jonas-lj: The 256 limit is to keep the verification time low. With that limit, verifying a proof takes at most ~2ms and if we set the limit to 512 instead it'll be ~4ms, which is also fine.

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: You mean the errors on line 130 and 132? An error on line 130 happens if one of the user provided commitments are not 32 bytes, an on line 132 it's if the bytes is not a valid RistrettoPoint, so I don't think this should be a PartialVMResult.

crates/sui-framework/packages/sui-framework/sources/crypto/rangeproofs.move
```diff
@@ -0,0 +1,52 @@
...
+const EInvalidProof: u64 = 1;
+#[allow(unused_const)]
+const EInvalidCommitment: u64 = 2;
+#[allow(unused_const)]
+const EInvalidRange: u64 = 3;
+#[allow(unused_const)]
+const EInvalidBatchSize: u64 = 4;
+const EUnsupportedVersion: u64 = 5;
+
+/// Verify a range proof over the Ristretto255 curve that all committed values are in the range [0, 2^bits).
```
benr-ml, line 19: "Enabled only on devnet"

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
benr-ml: the user does not call the Move function with bytes but with `commitments: &vector<Element<ristretto255::G>>,`

Cargo.toml
```diff
@@ -638,10 +638,10 @@ move-abstract-stack = { path = "external-crates/move/crates/move-abstract-stack"
 move-analyzer = { path = "external-crates/move/crates/move-analyzer" }
 move-trace-format = { path = "external-crates/move/crates/move-trace-format" }
 
-fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "7ba456d008a3a657e01589de64cb0eeb8aca532b" }
-fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", rev = "7ba456d008a3a657e01589de64cb0eeb8aca532b" }
-fastcrypto-zkp = { git = "https://github.com/MystenLabs/fastcrypto", rev = "7ba456d008a3a657e01589de64cb0eeb8aca532b", package = "fastcrypto-zkp" }
-fastcrypto-vdf = { git = "https://github.com/MystenLabs/fastcrypto", rev = "7ba456d008a3a657e01589de64cb0eeb8aca532b", features = [
+fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "4d2550a9833b3e686a94b823271e38063d309243" }
```
benr-ml: let's move it to a different pr and merge it first

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: Oh yes, you're right. In that case we could even skip the verification of the points.

benr-ml (commented):
LGTM, let's add snapshots, etc.
Also, @tzakian could you please take a look as well to confirm we didn't miss anything with respect to gas costs?

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    // Charge the base cost for this operation
+    native_charge_gas_early_exit!(
+        context,
+        cost_parameters
+            .verify_bulletproofs_ristretto255_base_cost
+            .ok_or_else(
+                || PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR).with_message(
+                    "verify_bulletproofs_ristretto255_base_cost not available".to_string()
+                )
```
tzakian: Nit: we have a fancy `partial_vm_error!` macro now that you can use here if you want


```suggestion
                || partial_vm_error!(UNKNOWN_INVARIANT_VIOLATION_ERROR,
                    "verify_bulletproofs_ristretto255_base_cost not available"
                )
```

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -0,0 +1,151 @@
...
+    let Ok(proof) = RangeProof::from_bytes(&proof_bytes) else {
+        return Ok(NativeResult::err(context.gas_used(), INVALID_PROOF));
+    };
+
+    let Ok(range) = range_from_bits(range_bits).map_err(|_| InvalidInput) else {
+        return Ok(NativeResult::err(context.gas_used(), INVALID_RANGE));
+    };
+
+    let length = commitments
+        .len(&Type::Vector(Box::new(Type::U8)))?
```
tzakian: Nit: in a couple places you're needing to construct this, maybe binding this and then reusing it would be nice.

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -0,0 +1,151 @@
...
+                })
+                .and_then(|b| {
+                    RistrettoPoint::from_trusted_byte_array(&b)
+                        .map_err(|_| PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR))
+                })
+                .map(PedersenCommitment)
+        })
+        .collect::<PartialVMResult<Vec<PedersenCommitment>>>()?;
+
+    let result = proof.verify_batch(&commitments, &range, &mut thread_rng());
```
tzakian, line 131: whoa...is this safe? using a thread rng in validator code feels really scary to me.

Are we 100% absolutely certain that this could not lead to non-deterministic results and/or errors?

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -0,0 +1,151 @@
...
+            commitments
+                .borrow_elem(i as usize, &Type::Vector(Box::new(Type::U8)))
+                .and_then(|reference| reference.value_as::<VectorRef>())
+                .and_then(|v| Ok(v.as_bytes_ref()?.to_vec()))
+                .and_then(|v| {
+                    v.try_into()
+                        .map_err(|_| PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR))
+                })
+                .and_then(|b| {
+                    RistrettoPoint::from_trusted_byte_array(&b)
```
tzakian, line 124: When can these fail? Could they ever fail? I'm guessing not/they should never ever fail, but figured best to ask :)

jonas-lj: They should not, no. The `RistrettoPoint` type is converted to bytes in Move and used here. So if they are not the right format, something is seriously broken in our implementation :).

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: We're 99.99999...% certain about this :). Assuming our implementation is correct, the risk that two honest validators get different results are negligible, so this is safe.

tzakian (approved):
Looks good to me!

PR #25960, opened 2026-03-24
[system] more cleanup

damirka:
## Description 

- inline some private functions for better visibility
- set next epoch value in advance for candidate
- extended tests to check removal sorting
- use macros in a few more places
- nits and bits

## Test plan 

Existing tests pass, adds more tests for existing behaviors

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -1020,42 +992,13 @@ fun clean_report_records_leaving_validator(
     });
 }
 
-/// Sort all the pending removal indexes.
```
damirka, line 1023: Replaced with `insertion_sort_by`.

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -1020,42 +992,13 @@ fun clean_report_records_leaving_validator(
...
-            } else {
-                break
-            };
-        };
-        i = i + 1;
-    };
-}
-
-/// Process all active validators' pending stake deposits and withdraws.
-fun process_pending_stakes_and_withdraws(validators: &mut vector<Validator>, ctx: &TxContext) {
```
damirka, line 1043: Inlined.

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -1020,42 +992,13 @@ fun clean_report_records_leaving_validator(
...
-
 /// Calculate the total active validator stake.
 public(package) fun calculate_total_stakes(validators: &vector<Validator>): u64 {
     let mut stake = 0;
     validators.do_ref!(|v| stake = stake + v.total_stake());
     stake
 }
 
-/// Process the pending stake changes for each validator.
-fun adjust_stake_and_gas_price(validators: &mut vector<Validator>) {
```
damirka, line 1055: Inlined.

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -715,34 +703,23 @@ public(package) fun is_active_validator_by_sui_address(
...
+        .active_validators
+        .any!(|v| v.sui_address() != search.sui_address() && v.is_duplicate(search))
 }
 
 /// Checks whether `new_validator` is duplicate with any currently pending validators.
-fun is_duplicate_with_pending_validator(self: &ValidatorSet, new_validator: &Validator): bool {
-    count_duplicates_tablevec(&self.pending_active_validators, new_validator) > 0
-}
-
-fun count_duplicates_tablevec(validators: &TableVec<Validator>, validator: &Validator): u64 {
```
damirka, line 738: Instead of counting we stop when we found 1 duplicate. The counting logic is not used anywhere.

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -605,11 +598,6 @@ fun update_validator_positions_and_calculate_total_stake(
     initial_total_stake - total_removed_stake
 }
 
-/// Effectuate pending next epoch metadata if they are staged.
-fun effectuate_staged_metadata(self: &mut ValidatorSet) {
```
damirka, line 609: Inlined.

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -260,11 +259,8 @@ public(package) fun assert_no_pending_or_active_duplicates(
     validator: &Validator,
 ) {
     // Validator here must be active or pending, and thus must be identified as duplicate exactly once.
-    assert!(
-        count_duplicates_vec(&self.active_validators, validator) +
-            count_duplicates_tablevec(&self.pending_active_validators, validator) == 1,
-        EDuplicateValidator,
-    );
+    assert!(!self.is_duplicate_with_active_validator(validator), EDuplicateValidator);
+    assert!(!self.is_duplicate_with_pending_validator(validator), EDuplicateValidator);
```
tzakian, line 266: Should we add an additional assertion that the validator is active or pending here to keep it in-line with the previous check/checking in depth?

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -169,15 +169,14 @@ public(package) fun request_add_validator_candidate(
...
     ctx: &mut TxContext,
 ) {
-    // The next assertions are not critical for the protocol, but they are here to catch problematic configs earlier.
-    assert!(
-        !self.is_duplicate_with_active_validator(&validator)
-            && !self.is_duplicate_with_pending_validator(&validator),
-        EDuplicateValidator,
-    );
     let validator_address = validator.sui_address();
     assert!(!self.validator_candidates.contains(validator_address), EAlreadyValidatorCandidate);
```
manolisliolios, line 174: order of assertions was changed here, it probably doesn't matter as long as tests pass (but you could get diff error kind)

manolisliolios (approved):
lgtm!

crates/sui-framework/packages/sui-system/sources/validator_set.move
damirka: They do not overlap, if you look at checks performed.

crates/sui-framework/packages/sui-system/sources/validator_set.move
damirka: This is already enforced in each call to this function. The function is used to compare metadata against active/pending set on an update. And we restrict these updates (except for 1 special case) to active validators.

The one special case is `update_validator_name` which can be called by candidates, but we avoid assertion there.

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -169,14 +170,21 @@ public(package) fun request_add_validator_candidate(
...
+    );
+    assert!(
+        find_validator_from_table_vec(&self.pending_active_validators, validator_address).is_none(),
+        EAlreadyValidator,
+    );
+
+    // The next assertions are not critical for the protocol, but they are here
+    // to catch problematic configs earlier.
+    assert!(!self.is_duplicate_with_active_validator(&validator), EDuplicateValidator);
+    assert!(!self.is_duplicate_with_pending_validator(&validator), EDuplicateValidator);
```
emmazzz: We could also just call `assert_no_pending_or_active_duplicates` here? Also cool if you think inlining would be more clear.

crates/sui-framework/packages/sui-system/sources/validator_set.move
```diff
@@ -260,11 +268,8 @@ public(package) fun assert_no_pending_or_active_duplicates(
     validator: &Validator,
 ) {
     // Validator here must be active or pending, and thus must be identified as duplicate exactly once.
```
emmazzz: This comment needs updating to reflect the new behavior.

emmazzz (approved):
Left some nit comments. Otherwise lgtm!

PR #25963, opened 2026-03-24
graphql-alt: add tls config to FN client and enforce URL validation

henryachen:
## Description 
- Configure TLS on the gRPC channel when --fullnode-rpc-url uses an https:// scheme, using the native root certificates
- Change FullnodeArgs::fullnode_rpc_url from `Option<String>` to `Option<Url>` so clap rejects invalid or schemeless URLs at parse time
  - Add unit tests for the fullnode client construction (no URL, http, https)

## Test plan 

- `cargo nextest run -p sui-indexer-alt-reader -- fullnode_client`
- Verify `--fullnode-rpc-url https://...` connects with TLS to a full node

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/src/sui_commands.rs
+++ b/crates/sui/src/sui_commands.rs
@@ -1164,7 +1164,7 @@ async fn start(
         };
 
         let fullnode_args = FullnodeArgs {
-            fullnode_rpc_url: Some(fullnode_rpc_url.parse()?),
+            fullnode_rpc_url: Some(socket_addr_to_url(fullnode_rpc_address)?),
         };
 
         let mut graphql_config = GraphQlConfig::default();


crates/sui/src/sui_commands.rs
```diff
@@ -1164,7 +1164,7 @@ async fn start(
         };
 
         let fullnode_args = FullnodeArgs {
-            fullnode_rpc_url: Some(fullnode_rpc_url.clone()),
+            fullnode_rpc_url: Some(fullnode_rpc_url.parse()?),
```
tpham-mysten: nit: I think the logic on line 1042 already returns a `Url`

PR #25970, opened 2026-03-25
feat(tidehunter): EpochMarkerKey with cfg-gated fixed-length serialization for object_per_epoch_marker_table_v2

andll:
## Motivation

`object_per_epoch_marker_table_v2` stores `(EpochId, FullObjectKey)` keys. The standard bincode encoding of this type is **variable-length**: `FullObjectKey::Fastpath` serialises to 52 bytes and `FullObjectKey::Consensus` to 60 bytes. This causes two performance problems with the TideHunter storage backend:

1. **Slower iteration / range-scan.** TideHunter's flat-buffer path for variable-length keys maintains a per-cell offset table, making scans more expensive than the fixed-length path which can index directly.
2. **No `drop_cells_in_range`.** TideHunter's O(1) bulk-delete primitive requires fixed-length, cell-boundary-aligned keys. Without it, epoch cleanup falls back to a full scan + range-delete tombstone.

## Approach

Rather than a per-field codec attached to `DBMap` (which requires derive macro changes), this PR introduces a named **`EpochMarkerKey`** type that owns its serialization strategy via `#[cfg]`-gated `Serialize`/`Deserialize` impls.

### `EpochMarkerKey` — two compile-time serialization formats

`EpochMarkerKey(EpochId, FullObjectKey)` has two serde implementations:

- **`#[cfg(not(tidehunter))]`**: delegates to the `(EpochId, FullObjectKey)` tuple, producing identical bincode bytes — preserving on-disk compatibility with existing RocksDB data.

- **`#[cfg(tidehunter)]`**: uniform **56-byte** fixed-length encoding via a custom `serialize_tuple` / `SeqAccess` visitor. The MSB of the EpochId word carries a Fastpath/Consensus discriminant bit, eliminating all variable-length fields.

```
[0..8]   EpochId (u64 big-endian) with bit 63 = discriminant: 0=Fastpath, 1=Consensus
[8..40]  ObjectID (32 bytes)
[40..48] Consensus start version, or 0 padding for Fastpath (u64 big-endian)
[48..56] Object version (u64 big-endian)
```

The discriminant-in-MSB trick also preserves the natural sort order: all Fastpath keys for epoch N sort before all Consensus keys for epoch N, which sort before epoch N+1.

Because `be_fix_int_ser` calls `Serialize`, all typed-store read/write paths (get, insert, bounds computation, iteration) automatically use the right encoding for each backend — **no changes to `typed-store-derive` or the codec machinery in `DBMap`**.

### Epoch cleanup: `drop_cells_in_range_raw`

With fixed-length keys, the TH path can use `drop_cells_in_range_raw` with `EpochMarkerKey::MIN_KEY` / `MAX_KEY` (all-zeros / all-ones byte arrays) instead of a full-table scan + range-delete tombstone.

## Files changed

| File | Change |
|------|--------|
| `typed-store/src/rocks/mod.rs` | `drop_cells_in_range_raw` — raw-byte O(1) bulk-delete |
| `sui-core/src/authority/epoch_marker_key.rs` | New: `EpochMarkerKey` with cfg-gated serde impls + unit tests |
| `sui-core/src/authority/authority_store_tables.rs` | Table type → `DBMap<EpochMarkerKey, MarkerValue>`; `KeyIndexing::fixed(56)`; pruner closure updated |
| `sui-core/src/authority/authority_store.rs` | Call sites wrap with `EpochMarkerKey(...)`; TH epoch cleanup uses `drop_cells_in_range_raw` |
| `sui-core/src/authority.rs` | Module declaration for `epoch_marker_key` |

### DIFF
--- a/crates/sui-core/src/authority/epoch_marker_key.rs
+++ b/crates/sui-core/src/authority/epoch_marker_key.rs
@@ -1,14 +1,14 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use sui_types::committee::EpochId;
 #[cfg(tidehunter)]
 use sui_types::base_types::{ObjectID, SequenceNumber};
+use sui_types::committee::EpochId;
 #[cfg(tidehunter)]
 use sui_types::storage::ConsensusObjectKey;
+use sui_types::storage::FullObjectKey;
 #[cfg(tidehunter)]
 use sui_types::storage::ObjectKey;
-use sui_types::storage::FullObjectKey;
 
 /// A key type for `object_per_epoch_marker_table_v2`.
 ///


crates/sui-core/src/authority/epoch_marker_key.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use sui_types::base_types::{ObjectID, SequenceNumber};
+    use sui_types::storage::ObjectKey;
+    use typed_store::be_fix_int_ser;
+
+    #[cfg(tidehunter)]
```
johnjmartin, line 152: is it worth adding any tests for the rockdb path (i realize this part is quite small)?

andll: Rocksdb implementation essentially forwards serialization to a tuple, i feel like unit test for that would be a bit redundant. (e.g. we would be just checking that the code that does (a, b).serialize is indeed equal to (a,b).serialize). Unless you are talking about some other tests

PR #25974, opened 2026-03-25
[effects] switch gas_object method to optional

phoenix-o:
## Description 

switches effects' gas_object method to optional.
Analytics and JSON RPC still convert None results internally to a dummy response

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/src/lib.rs
+++ b/crates/sui-benchmark/src/lib.rs
@@ -152,9 +152,9 @@ impl ExecutionEffects {
     pub fn gas_object(&self) -> (ObjectRef, Owner) {
         match self {
             ExecutionEffects::FinalizedTransactionEffects(effects, ..) => {
-                effects.data().gas_object()
+                effects.data().gas_object().unwrap()
             }
-            ExecutionEffects::ExecutedTransaction(txn) => txn.effects.gas_object(),
+            ExecutionEffects::ExecutedTransaction(txn) => txn.effects.gas_object().unwrap(),
         }
     }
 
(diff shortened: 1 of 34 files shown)


crates/sui-indexer-alt-e2e-tests/tests/graphql/transaction_effects/gas_effects.snap
```diff
@@ -83,11 +83,7 @@ Response: {
   "data": {
     "functionCallTransaction": {
       "gasEffects": {
-        "gasObject": {
-          "address": "0x0000000000000000000000000000000000000000000000000000000000000000",
-          "version": 0,
-          "digest": "11111111111111111111111111111111"
-        },
+        "gasObject": null,
```
phoenix-o, line 86: do we want to keep the old behavior here for GraphQL and continue converting `null` to a dummy object?

amnn: All good, in fact I made the same change as part of this related PR: https://github.com/MystenLabs/sui/pull/25994

crates/sui-benchmark/src/lib.rs
```diff
@@ -152,9 +152,9 @@ impl ExecutionEffects {
     pub fn gas_object(&self) -> (ObjectRef, Owner) {
         match self {
             ExecutionEffects::FinalizedTransactionEffects(effects, ..) => {
-                effects.data().gas_object()
+                effects.data().gas_object().unwrap()
```
amnn, line 155: If `sui-benchmark` is used to benchmark transactions that use address balance withdrawals for gas this could cause a problem right? Not sure if that is a possibility at the moment.

PR #25975, opened 2026-03-25
[Block Streaming] Process received blocks

akichidis:
## Description 

This PR is implementing the ObserverService to process received blocks from the peer that the node is streaming the blocks from.

## Test plan 

CI

## Next

- https://github.com/MystenLabs/sui/pull/25979

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/metrics.rs
+++ b/consensus/core/src/metrics.rs
@@ -493,7 +493,7 @@ impl NodeMetrics {
             invalid_blocks: register_int_counter_vec_with_registry!(
                 "invalid_blocks",
                 "Number of invalid blocks per peer authority",
-                &["authority", "source", "error", "node_type"],
+                &["authority", "source", "error"],
                 registry,
             ).unwrap(),
             rejected_blocks: register_int_counter_vec_with_registry!(
--- a/consensus/core/src/metrics.rs
+++ b/consensus/core/src/metrics.rs
@@ -493,7 +493,7 @@ impl NodeMetrics {
             invalid_blocks: register_int_counter_vec_with_registry!(
                 "invalid_blocks",
                 "Number of invalid blocks per peer authority",
-                &["authority", "source", "error"],
+                &["authority", "source", "error", "node_type"],
                 registry,
             ).unwrap(),
             rejected_blocks: register_int_counter_vec_with_registry!(
--- a/consensus/core/src/observer_service.rs
+++ b/consensus/core/src/observer_service.rs
@@ -318,11 +318,8 @@ mod tests {
         let core_dispatcher = Arc::new(MockCoreThreadDispatcher::default());
         let block_verifier = Arc::new(NoopBlockVerifier);
         let commit_vote_monitor = Arc::new(CommitVoteMonitor::new(context.clone()));
-        let transaction_vote_tracker = TransactionVoteTracker::new(
-            context.clone(),
-            block_verifier.clone(),
-            dag_state.clone(),
-        );
+        let transaction_vote_tracker =
+            TransactionVoteTracker::new(context.clone(), block_verifier.clone(), dag_state.clone());
 
         let observer_service = ObserverService::new(
             context.clone(),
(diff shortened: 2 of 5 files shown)


consensus/core/src/observer_service.rs
```diff
@@ -51,29 +70,144 @@ impl ObserverService {
...
+                    .authority(*authority)
+                    .hostname
+                    .as_str(),
+                "validator",
+            ),
+            PeerId::Observer(node_id) => {
+                observer_name = format!("{:?}", node_id);
+                (observer_name.as_str(), "observer")
+            }
+        };
```
akichidis, line 97: Occurrences like this will get replaced by the `hostname` method which I've introduced on the next PR https://github.com/MystenLabs/sui/pull/25979/changes#diff-2650b17582a47fcd8c4a264d69a952f517821bfc3324c8ebbda4aceb7b01d707R65

consensus/core/src/metrics.rs
```diff
@@ -493,7 +493,7 @@ impl NodeMetrics {
             invalid_blocks: register_int_counter_vec_with_registry!(
                 "invalid_blocks",
                 "Number of invalid blocks per peer authority",
-                &["authority", "source", "error"],
+                &["authority", "source", "error", "node_type"],
```
mwtian: Why do we need this label here, but not other metrics? Can we figure out node type with higher level queries?

consensus/core/src/observer_service.rs
```diff
@@ -1,48 +1,67 @@
...
     context::Context,
+    core_thread::CoreThreadDispatcher,
     dag_state::DagState,
     error::{ConsensusError, ConsensusResult},
     network::{
         NodeId, ObserverBlockStream, ObserverBlockStreamItem, ObserverNetworkService, PeerId,
     },
 };
 
+pub(crate) const COMMIT_LAG_MULTIPLIER: u32 = 5;
```
mwtian, line 32: Add a comment?

consensus/core/src/observer_service.rs
mwtian: Since peer name is going to metrics, we probably want to limit its cardinality.

consensus/core/src/observer_service.rs
```diff
@@ -51,29 +70,144 @@ impl ObserverService {
...
+                    .as_str(),
+                "validator",
+            ),
+            PeerId::Observer(node_id) => {
+                observer_name = format!("{:?}", node_id);
+                (observer_name.as_str(), "observer")
+            }
+        };
+
+        // Reject blocks failing parsing and validations.
```
mwtian, line 99: After this PR, we should think of reusing logic between observer and validator handlers.

mwtian (approved):

consensus/core/src/observer_service.rs
akichidis: So on the upcoming PR I've introduced the `BlockSyncService` that hosts the common application logic for `fetch_blocks` etc to tackle the duplication issue https://github.com/MystenLabs/sui/pull/25997/changes#diff-e419a2c434171dce71c787bfff7e22cfec84ac0563d1999394289d16ec03e765R31

I believe we could move part of the code there

akichidis:
> Since peer name is going to metrics, we probably want to limit its cardinality.

let me fix that to the next PR where I have moved the label calculation

consensus/core/src/metrics.rs
akichidis: In theory we can yes, but it should be more convenient this way

mwtian: I don't think here is the best approach. The "role" / "type" of a node can be applied on the metrics infra layer.

akichidis: Agreed, I'll remove that.

PR #25979, opened 2026-03-25
[Block Streaming] Introduce the PeersPool & Refactor synchornizer

akichidis:
## Description 

This PR is introducing the `PeersPool` which from now on will act as a billboard to keep the peers available to communicate with and their available interfaces/servers that they support (Validator, Observer). On the longer term the `PeersPool` will be updated dynamically from peers (other Observers) that join the network and make their selfs available.

The `synchronizer` has been refactored to support the `PeersPool` so the Observer nodes can also synchronize missing blocks from the peer.

## Test plan 

CI

## Next

- https://github.com/MystenLabs/sui/pull/25997

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/authority_node.rs
+++ b/consensus/core/src/authority_node.rs
@@ -795,7 +795,7 @@ mod tests {
             ..Default::default()
         };
 
-        let (observer_commit_consumer, observer_commit_receiver) = CommitConsumerArgs::new(0, 0);
+        let (observer_commit_consumer, _observer_commit_receiver) = CommitConsumerArgs::new(0, 0);
         let observer = ConsensusAuthority::start(
             network_type,
             0,
@@ -811,8 +811,6 @@ mod tests {
             0,
         )
         .await;
-
-        commit_receivers.push(observer_commit_receiver);
 
         // Give Observer more time to connect and sync
         sleep(Duration::from_secs(5)).await;
--- a/consensus/core/src/authority_node.rs
+++ b/consensus/core/src/authority_node.rs
@@ -486,7 +486,7 @@ where
                 {
                     PeerId::Validator(index)
                 } else {
-                    PeerId::Observer(peer_record.public_key.clone())
+                    PeerId::Observer(Box::new(peer_record.public_key.clone()))
                 };
 
                 info!("Observer subscribing to peer: {:?}", peer_id);
--- a/consensus/core/src/authority_node.rs
+++ b/consensus/core/src/authority_node.rs
@@ -437,6 +437,7 @@ where
                     block_verifier,
                     commit_vote_monitor.clone(),
                     transaction_vote_tracker.clone(),
+                    synchronizer.clone(),
                 ));
                 network_manager
                     .start_observer_server(observer_service)
@@ -455,6 +456,7 @@ where
                 block_verifier,
                 commit_vote_monitor.clone(),
                 transaction_vote_tracker.clone(),
+                synchronizer.clone(),
             ));
 
             let observer_subscriber = ObserverSubscriber::new(
--- a/consensus/core/src/authority_node.rs
+++ b/consensus/core/src/authority_node.rs
@@ -34,7 +34,7 @@ use crate::{
     },
     observer_service::ObserverService,
     observer_subscriber::ObserverSubscriber,
-    peers_pool::{PeerServer, PeersPool},
+    peers_pool::PeersPool,
     round_prober::{RoundProber, RoundProberHandle},
     round_tracker::RoundTracker,
     storage::rocksdb_store::RocksDBStore,
@@ -451,27 +451,6 @@ where
 
             (SubscriberType::Validator(s), round_prober_handle)
         } else {
-            // Register all the observer peers in the peers pool
-            for peer in context.parameters.tonic.observer_peers.iter() {
-                // If the peer's public key is in the committe, then register it as a validator.
-                if let Some((index, _)) = context
-                    .committee
-                    .authorities()
-                    .find(|(_, authority)| authority.network_key == peer.public_key)
-                {
-                    peers_pool
-                        .register_validator(
-                            index,
-                            vec![PeerServer::Validator, PeerServer::Observer],
-                        )
-                        .expect("Failed to register validator");
-                } else {
-                    // Otherwise this is another Observer peer and we register it as such. This peer we know that
-                    // it will support the Observer server.
-                    peers_pool.register_observer(peer.public_key.clone());
-                }
-            }
-
             // Observer node: subscribe to specified peer(s) using ObserverSubscriber
             let observer_client = network_manager.observer_client();
             let observer_service = Arc::new(ObserverService::new(
@@ -816,7 +795,7 @@ mod tests {
             ..Default::default()
         };
 
-        let (observer_commit_consumer, _observer_commit_receiver) = CommitConsumerArgs::new(0, 0);
+        let (observer_commit_consumer, observer_commit_receiver) = CommitConsumerArgs::new(0, 0);
         let observer = ConsensusAuthority::start(
             network_type,
             0,
@@ -832,6 +811,8 @@ mod tests {
             0,
         )
         .await;
+
+        commit_receivers.push(observer_commit_receiver);
 
         // Give Observer more time to connect and sync
         sleep(Duration::from_secs(5)).await;
@@ -852,7 +833,7 @@ mod tests {
             let mut expected_transactions = submitted_transactions.clone();
             loop {
                 let committed_subdag =
-                    tokio::time::timeout(Duration::from_secs(1), receiver.recv())
+                    tokio::time::timeout(Duration::from_secs(3), receiver.recv())
                         .await
                         .unwrap()
                         .unwrap();
--- a/consensus/core/src/authority_node.rs
+++ b/consensus/core/src/authority_node.rs
@@ -34,6 +34,7 @@ use crate::{
     },
     observer_service::ObserverService,
     observer_subscriber::ObserverSubscriber,
+    peers_pool::{PeerServer, PeersPool},
     round_prober::{RoundProber, RoundProberHandle},
     round_tracker::RoundTracker,
     storage::rocksdb_store::RocksDBStore,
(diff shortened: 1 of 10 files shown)


consensus/core/src/network/mod.rs
```diff
@@ -61,6 +61,15 @@ impl Display for PeerId {
     }
 }
 
+impl PeerId {
+    pub(crate) fn hostname(&self, context: &Context) -> String {
+        match self {
+            PeerId::Validator(index) => context.committee.authority(*index).hostname.to_string(),
+            PeerId::Observer(_) => format!("[Observer]{:?}", self),
```
mwtian: Just `Display` instead of `Debug`?

consensus/core/src/authority_node.rs
```diff
@@ -824,7 +831,7 @@ mod tests {
             let mut expected_transactions = submitted_transactions.clone();
             loop {
                 let committed_subdag =
-                    tokio::time::timeout(Duration::from_secs(1), receiver.recv())
+                    tokio::time::timeout(Duration::from_secs(3), receiver.recv())
```
mwtian: Were there test failures?

consensus/core/src/error.rs
```diff
@@ -92,7 +92,13 @@ pub enum ConsensusError {
     SignatureVerificationFailure(FastCryptoError),
 
     #[error("Synchronizer for fetching blocks directly from {0} is saturated")]
-    SynchronizerSaturated(AuthorityIndex),
+    SynchronizerSaturated(String),
+
+    #[error("Peer {0} is unavailable")]
+    PeerUnavailable(String),
+
+    #[error("Peer {0} not found")]
```
mwtian: Be more specific on where the Peer is not found?

consensus/core/src/peers_pool.rs
```diff
@@ -0,0 +1,412 @@
...
+use crate::{
+    context::Context,
+    error::{ConsensusError, ConsensusResult},
+    network::{NodeId, PeerId},
+};
+
+/// The server types that a peer can support.
+/// A peer may support multiple server types.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) enum PeerServer {
```
mwtian: How about `PeerService`?

consensus/core/src/peers_pool.rs
```diff
@@ -0,0 +1,412 @@
...
+    /// The observer server for streaming blocks and serving observer requests
+    #[allow(dead_code)]
+    Observer,
+}
+
+/// Information about a registered peer
+#[derive(Debug, Clone)]
+struct PeerInfo {
+    /// The server types this peer supports
+    supported_servers: BTreeSet<PeerServer>,
```
mwtian: `supported_services`?

consensus/core/src/synchronizer.rs
```diff
@@ -171,7 +170,7 @@ impl InflightBlocksMap {
 enum Command {
     FetchBlocks {
         missing_block_refs: BTreeSet<BlockRef>,
-        peer_index: AuthorityIndex,
+        peer: Box<PeerId>,
```
mwtian: Maybe we should support passing PeerId by value. The Observer variant can use a Box internally.

consensus/core/src/synchronizer.rs
```diff
@@ -285,14 +297,14 @@ where
         // Spawn the tasks to fetch the blocks from the others
         let mut fetch_block_senders = BTreeMap::new();
         let mut tasks = JoinSet::new();
-        for (index, _) in context.committee.authorities() {
-            if index == context.own_index {
-                continue;
-            }
+
+        // Create fetch tasks for all available peers (validators and observers)
```
mwtian: We can start with this logic. But now with observer, the set of peers does change. How is removing peers handled now?

consensus/core/src/synchronizer.rs
```diff
@@ -1112,26 +1157,29 @@ where
...
-        let mut peers: Vec<AuthorityIndex> = context
-            .committee
-            .authorities()
-            .filter_map(|(index, _)| (index != context.own_index).then_some(index))
-            .collect();
+        // Get available peers from the PeersPool
+        let mut peers = peers_pool.get_available_peers();
+        let available_peers_count = peers.len();
+        assert!(
+            available_peers_count > 0,
```
mwtian: Maybe just assert !peers.is_empty()

consensus/core/src/synchronizer.rs
```diff
@@ -1189,11 +1238,21 @@ where
...
+
+        // Get available peers from the PeersPool
+        let mut peers = peers_pool.get_available_peers();
+
         // Distribute the same number of authorities into each peer to sync.
-        // When running this function, context.committee.size() is always greater than 1.
+        // Use the number of available peers from the pool, capped at MAX_PERIODIC_SYNC_PEERS
+        let available_peers_count = peers.len();
+        assert!(
+            available_peers_count > 0,
```
mwtian: Maybe just assert !peers.is_empty()

consensus/core/src/network/mod.rs
```diff
@@ -61,6 +61,15 @@ impl Display for PeerId {
     }
 }
 
+impl PeerId {
+    pub(crate) fn hostname(&self, context: &Context) -> String {
```
mwtian, line 73: I think we should differentiate:
- hostname(): for logging, prints 4-8 digits of the key
- labelname(): for metrics, not print public keys

consensus/core/src/peers_pool.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/// Pool of available peers (both validators and observers) that can be used for communication.
+/// This pool tracks which peers are registered, what servers they support, and their availability for synchronization.
```
mwtian: For register, maybe "known" is better?

consensus/core/src/peers_pool.rs
```diff
@@ -0,0 +1,412 @@
...
+            }
+        }
+
+        s
+    }
+
+    /// Registers a validator as available in the pool with its supported servers.
+    /// Note: this method will override the peer if it already exists in the pool.
+    #[allow(dead_code)]
+    pub(crate) fn register_validator(
```
mwtian, line 88: A bit confusing is that {validator, observer} are both services, and node types.

consensus/core/src/peers_pool.rs
```diff
@@ -0,0 +1,412 @@
...
+            }
+        } else {
+            // Peer not registered (should not happen for validators after initialization)
+            false
+        }
+    }
+
+    /// Gets all available peers based that are compatible with the current node.
+    #[allow(dead_code)]
+    pub(crate) fn get_available_peers(&self) -> Vec<PeerId> {
```
mwtian: Is this more like `get_known_peers()`? There is no check for actual availability?

consensus/core/src/network/mod.rs
akichidis: Good idea!

consensus/core/src/network/mod.rs
akichidis: Let me check if currently display is supported

consensus/core/src/authority_node.rs
akichidis: Hhhm not really, let me check again why I've changed that to `3`. It might be an oversight

consensus/core/src/peers_pool.rs
akichidis: I went for server to match the gRPC server naming. But service sounds equally good to me

consensus/core/src/peers_pool.rs
akichidis: That's true. The only solution I've found was to introduce more terminology which I've tried to avoid. I think, as you suggested, if we stay with the `Service` for defining the interface availability, hopefully it's going to be clearer

consensus/core/src/peers_pool.rs
akichidis: Ah yes, `known` is better

consensus/core/src/synchronizer.rs
akichidis: I should have left a comment here, that now we don't handle the removing of peers. This will happen as part of the "scaling" milestone where state sync will be utilised to populate the service availability. If you are concerned about the code functionality per se at this point then what I can do is:
* Remove the `remove` functionality from the pool
* Pool should be initialised on construction, and no separate `registration` should be offer publicly at this point

so we make this more static.

In the future the dynamic registration/removal of peers would be handled by utilising signals/events emitted by the `SyncPool` it self. The synchornizer will basically have a `task` running listening to the events and spinning up/winding down the running tasks for the available peers. @mwtian I can craft this functionality earlier, but might not be used soon.

consensus/core/src/network/mod.rs
```diff
@@ -43,20 +43,55 @@ use crate::{
...
 
 /// Identifies a peer in the network, which can be either a validator or an observer.
+/// The Observer variant is boxed to keep the enum small, since `NodeId` (32 bytes) is
+/// much larger than `AuthorityIndex` (4 bytes).
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub(crate) enum PeerId {
     /// A validator node identified by its authority index.
     Validator(AuthorityIndex),
     /// An observer node identified by its network public key.
     #[allow(dead_code)]
```
mwtian, line 53: This is no longer deadcode?

consensus/core/src/network/mod.rs
```diff
@@ -43,20 +43,55 @@ use crate::{
...
 impl Display for PeerId {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
             PeerId::Validator(authority) => write!(f, "[{}]", authority),
-            PeerId::Observer(node_id) => write!(f, "[{:?}]", node_id),
+            PeerId::Observer(node_id) => {
+                let bytes = node_id.to_bytes();
+                write!(
+                    f,
+                    "[Observer:{:02x}{:02x}{:02x}{:02x}]",
```
mwtian: `impl Debug for ConciseAuthorityPublicKeyBytesRef` is another example.

mwtian (approved):

PR #25985, opened 2026-03-25
Add sui-kvstore and sui-kv-rpc to CI cluster

nickvikeras:
## Description

Integrate the archival stack with the CI tests. This uses the bigtable emulator from the GCP SDK, which is already installed on the CI image.

## Test plan

I added some panics to the sui-kv-rpc methods to ensure the tests crashed to "prove" the data was now being fetched from the bigtable emulator.

### DIFF
--- a/crates/sui-kv-rpc/src/lib.rs
+++ b/crates/sui-kv-rpc/src/lib.rs
@@ -79,7 +79,8 @@ impl KvRpcServer {
             .await?
             .pop()
             .expect("failed to fetch genesis checkpoint from the KV store");
-        let chain_id = ChainIdentifier::from(genesis.summary.digest());
+        let summary = genesis.summary.expect("genesis checkpoint missing summary");
+        let chain_id = ChainIdentifier::from(summary.digest());
         Ok(Self::init(
             client,
             chain_id,
--- a/crates/sui-kvstore/src/lib.rs
+++ b/crates/sui-kvstore/src/lib.rs
@@ -124,9 +124,9 @@ pub struct BigTableIndexer {
 
 #[derive(Clone, Debug)]
 pub struct CheckpointData {
-    pub summary: CheckpointSummary,
-    pub contents: CheckpointContents,
-    pub signatures: AuthorityStrongQuorumSignInfo,
+    pub summary: Option<CheckpointSummary>,
+    pub contents: Option<CheckpointContents>,
+    pub signatures: Option<AuthorityStrongQuorumSignInfo>,
 }
 
 #[derive(Clone, Debug)]
--- a/crates/sui-indexer-alt-reader/src/checkpoints.rs
+++ b/crates/sui-indexer-alt-reader/src/checkpoints.rs
@@ -78,11 +78,11 @@ impl Loader<CheckpointKey> for BigtableReader {
             .checkpoints(&checkpoint_keys)
             .await?
             .into_iter()
-            .map(|c| {
-                (
-                    CheckpointKey(c.summary.sequence_number),
-                    (c.summary, c.contents, c.signatures),
-                )
+            .filter_map(|c| {
+                Some((
+                    CheckpointKey(c.summary.as_ref()?.sequence_number),
+                    (c.summary?, c.contents?, c.signatures?),
+                ))
             })
             .collect())
     }
--- a/crates/sui-kv-rpc/src/v2/get_checkpoint.rs
+++ b/crates/sui-kv-rpc/src/v2/get_checkpoint.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use sui_data_ingestion_core::{CheckpointReader, create_remote_store_client};
+use sui_kvstore::tables::checkpoints::col;
 use sui_kvstore::{BigTableClient, CHECKPOINTS_PIPELINE, KeyValueStoreReader};
 use sui_rpc::field::{FieldMask, FieldMaskTree, FieldMaskUtil};
 use sui_rpc::merge::Merge;
@@ -30,6 +31,7 @@ pub async fn get_checkpoint(
         })?;
         FieldMaskTree::from(read_mask)
     };
+    let columns = checkpoint_columns(&read_mask);
     let checkpoint = match request.checkpoint_id {
         Some(CheckpointId::Digest(digest)) => {
             let digest = digest.parse::<CheckpointDigest>().map_err(|e| {
@@ -38,12 +40,12 @@ pub async fn get_checkpoint(
                     .with_reason(ErrorReason::FieldInvalid)
             })?;
             client
-                .get_checkpoint_by_digest(digest)
+                .get_checkpoint_by_digest_filtered(digest, Some(&columns))
                 .await?
                 .ok_or(CheckpointNotFoundError::digest(digest.into()))?
         }
         Some(CheckpointId::SequenceNumber(sequence_number)) => client
-            .get_checkpoints(&[sequence_number])
+            .get_checkpoints_filtered(&[sequence_number], Some(&columns))
             .await?
             .pop()
             .ok_or(CheckpointNotFoundError::sequence_number(sequence_number))?,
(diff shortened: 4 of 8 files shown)


crates/sui-indexer-alt-e2e-tests/src/lib.rs
```diff
@@ -122,6 +129,15 @@ pub struct OffchainCluster {
     /// doesn't get cleaned up until the cluster is stopped.
     #[allow(unused)]
     dir: TempDir,
+
+    /// BigTable Dependencies (BigTable emulator process and client for watermark polling).
+    archival: BigTable,
+}
+
+struct BigTable {
```
amnn: nit: could we follow the same pattern as the other parts of the cluster where they are stored directly in `OffchainCluster`, with the read interfaces clustered at the top, and the server handles clustered at the bottom? (I would also be fine with segregating each interface into its own component -- what I care about is the consistency -- but that seems like a bigger change)

```rust
pub struct OffchainCluster {
    /// The address the consistent store is listening on.
    consistent_listen_address: SocketAddr,

    /// The address the JSON-RPC server is listening on.
    jsonrpc_listen_address: SocketAddr,

    /// The address the GraphQL server is listening on.
    graphql_listen_address: SocketAddr,

    /// Read access to BigTable
    bigtable_client: BigTableClient

    /// Read access to the temporary database.
    db: Db,

    /// The pipelines that the indexer is populating.
    pipelines: Vec<&'static str>,

    /// Handles to all running services. Held on to so the services are not dropped (and therefore
    /// aborted) until the cluster is stopped.
    #[allow(unused)]
    services: Service,

    /// Handle to the BigTable emulator process.
    #[allow(unused)]
    bigtable_emulator: BigTableEmulator,

    /// Hold on to the database so it doesn't get dropped until the cluster is stopped.
    #[allow(unused)]
    database: TempDb,

    /// Hold on to the temporary directory where the consistent store writes its data, so it
    /// doesn't get cleaned up until the cluster is stopped.
    #[allow(unused)]
    dir: TempDir,
}
```

crates/sui-indexer-alt-e2e-tests/src/lib.rs
```diff
@@ -378,6 +401,18 @@ impl OffchainCluster {
             consistent_store_statement_timeout_ms: None,
         };
 
+        let (archival, archival_service) =
+            start_archival(client_args.clone(), kv_rpc_address).await?;
+
+        let graphql_kv_args = GraphQlKvArgs {
+            ledger_grpc_url: Some(
+                format!("http://{kv_rpc_address}")
```
amnn, line 421: nit: consistency with how the consistent reader args are set-up (whatever your preference on how it's done, just make them both the same).

crates/sui-indexer-alt-e2e-tests/src/lib.rs
```diff
@@ -722,3 +783,60 @@ pub async fn write_checkpoint(path: &Path, checkpoint: Checkpoint) -> anyhow::Re
...
+    create_tables(emulator.host(), INSTANCE_ID)
+        .await
+        .context("Failed to create BigTable tables")?;
+
+    let bigtable_client =
+        BigTableClient::new_local(emulator.host().to_string(), INSTANCE_ID.to_string())
+            .await
+            .context("Failed to create BigTable client")?;
+
+    // Use a separate registry to avoid metric name collisions with the Postgres indexer.
```
amnn: All our other services introduce a prefix to their indexing/ingestion metrics so that they don't collide while they can live in the same registry:

https://github.com/MystenLabs/sui/blob/76e09574c2df63131bf14ee6faca4e0797a46f66/crates/sui-indexer-alt-consistent-store/src/indexer.rs#L79

Could we do something similar for archival and/or postgres? It's a bit of disruption in our ops, because we then need to update all the metric names in our dashboards and alerts, but it then means we can combine these services in one process with a shared metrics service, which is quite useful.

For other services, this is a must because this is how `sui start --with-graphql` works. For BigTable, it's not quite a must yet, unless we want to give people the option to run the local network talking to a BigTable emulator, but I want to remove as many differences between set-ups as possible.

crates/sui-indexer-alt-reader/src/ledger_grpc_reader.rs
```diff
@@ -69,13 +69,16 @@ impl LedgerGrpcReader {
...
         if let Some(timeout) = args.statement_timeout() {
             endpoint = endpoint.timeout(timeout);
         }
-        let channel = endpoint.tls_config(tls_config)?.connect_lazy();
+        let channel = if uri.scheme_str() == Some("https") {
+            let tls_config = ClientTlsConfig::new().with_native_roots();
+            endpoint.tls_config(tls_config)?.connect_lazy()
+        } else {
+            endpoint.connect_lazy()
+        };
```
amnn:
```suggestion

        if uri.scheme_str() == Some("https") {
            let tls_config = ClientTlsConfig::new().with_native_roots();
            endpoint = endpoint.tls_config(tls_config)?;
        }

        let channel = endpoint.connect_lazy();
```

crates/sui-kv-rpc/src/lib.rs
amnn, line 1: Not necessarily all from this PR but there are some import style inconsistencies here (mix of single imports + nested, fully-qualified references vs not, scoped imports vs top-level).

crates/sui-kvstore/src/lib.rs
```diff
@@ -299,6 +300,29 @@ pub trait KeyValueStoreReader {
...
+        host: String,
+        instance_id: String,
+        client_args: ClientArgs,
+        registry: &Registry,
+    ) -> Result<Self> {
+        let client = BigTableClient::new_local(host, instance_id).await?;
+        let store = BigTableStore::new(client);
+        Self::new(
+            store,
+            IndexerArgs::default(),
```
amnn: nit: I would prefer not to introduce this pattern where incidental configuration that we're not using right now gets defaulted out in an intermediate level, because it leads to confusing as the codebase evolves (Is there something special about the local set-up that means we can't configure these things? If we need to configure them, should we introduce a new function here that accepts just that config or should we introduce just that config as a new parameter to the existing function and fix-up existing call-sites? etc).

Instead, could we just use the existing constructor, but pass in a `BigTableStore` with a local client, and set the defaults up in the test set-up?

crates/sui-kvstore/src/testing.rs
```diff
@@ -0,0 +1,184 @@
...
+
+use anyhow::Context;
+use anyhow::Result;
+use anyhow::bail;
+use futures::future::try_join_all;
+use tokio::process::Command as TokioCommand;
+
+pub const INSTANCE_ID: &str = "bigtable_test_instance";
+
+pub const TABLES: &[&str] = &[
```
amnn, line 26: Is there some way for us to know that a new table needs to be added here?

crates/sui-kvstore/src/testing.rs
```diff
@@ -0,0 +1,184 @@
...
+pub struct BigTableEmulator {
+    child: Child,
+    host: String,
+    // Keep stdout open so the emulator doesn't get SIGPIPE and die.
+    _stdout_drain: std::thread::JoinHandle<()>,
+}
+
+impl BigTableEmulator {
+    pub fn start() -> Result<Self> {
+        let mut child = Command::new(cbtemulator_path())
```
amnn: Out of interest, is it possible to reuse `get_available_port` to avoid having to parse `stdout` and then set-up a drain for it?

https://github.com/MystenLabs/sui/blob/76e09574c2df63131bf14ee6faca4e0797a46f66/crates/sui-pg-db/src/temp.rs#L314-L342

That way you can pipe `stdout` to `/dev/null` as well.

(I think it would be fine to copy `get_available_port` in here rather than take a confusing dependency).

crates/sui-kvstore/src/testing.rs
```diff
@@ -0,0 +1,184 @@
...
+                    bail!(
+                        "cbt createfamily {table} failed: {}",
+                        String::from_utf8_lossy(&output.stderr)
+                    );
+                }
+
+                Ok(())
+            }
+        })
+        .collect();
```
amnn: nit: I don't think you need to collect into a `Vec` in order to call `try_join_all`, it would accept the iterator directly.

crates/sui-kvstore/src/testing.rs
nickvikeras: These tests will fail when the indexer runs and tries to write to the new table.

crates/sui-indexer-alt-e2e-tests/src/lib.rs
nickvikeras: I was injecting the prefix here in the real binary - https://github.com/MystenLabs/sui/blob/1d9fe9dbb78ca64623a8587b801beaef0dbea8a1/crates/sui-kvstore/src/bin/sui-kvstore-alt.rs?plain=1#L115

So I think that is just wrong and I should be injecting it through this prefix. None of the actual metric names should change in the real prod environments though.

PR #25988, opened 2026-03-25
Render event/object json in archival

nickvikeras:
## Description 

Add support for rendering object/event json for ledger service in archival. Loads the layout from the bigtable objects table with an LRU cache (using the same package store type everything else uses for this).

## Test plan 

Added e2e tests.

---

## Release notes

- [ ] Protocol: 
- [x] gRPC: Render object json and event json in archival ledger service.
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-reader/src/checkpoints.rs
+++ b/crates/sui-indexer-alt-reader/src/checkpoints.rs
@@ -78,11 +78,11 @@ impl Loader<CheckpointKey> for BigtableReader {
             .checkpoints(&checkpoint_keys)
             .await?
             .into_iter()
-            .map(|c| {
-                (
-                    CheckpointKey(c.summary.sequence_number),
-                    (c.summary, c.contents, c.signatures),
-                )
+            .filter_map(|c| {
+                Some((
+                    CheckpointKey(c.summary.as_ref()?.sequence_number),
+                    (c.summary?, c.contents?, c.signatures?),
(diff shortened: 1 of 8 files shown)


crates/sui-kv-rpc/src/v2/mod.rs
```diff
@@ -12,15 +13,33 @@ use sui_rpc::proto::sui::rpc::v2::{
...
+    resolver: &PackageResolver,
+    struct_tag: &StructTag,
+    contents: &[u8],
+) -> Option<prost_types::Value> {
+    let type_tag = TypeTag::Struct(Box::new(struct_tag.clone()));
+    let layout = resolver.type_layout(type_tag).await.ok()?;
+    ProtoVisitor::new(MAX_JSON_MOVE_VALUE_SIZE)
+        .deserialize_value(contents, &layout)
+        .ok()
+}
```
bmwill, line 41: I wonder if there is a way that we could unify some of the rendering logic that now exists here and in the sui-rpc-api crate. Thats obviously a larger change than we probably want to make right now but just something to think about.

nickvikeras: Yeah I had the same thought as I was looking at this.

PR #25990, opened 2026-03-26
Use seeded RNG for validator selection in TestCluster

alex-mysten:
## Summary

Fixes flaky `test_reconfig_with_committee_change_stress` by using seeded RNG for validator selection instead of always picking the first validator.

After reconfiguration, the first validator in the BTreeMap may still be catching up, causing transaction timeouts. Random selection distributes load across validators.

## Test plan

Reproduced and verified fix locally.

### DIFF
--- a/crates/test-cluster/src/lib.rs
+++ b/crates/test-cluster/src/lib.rs
@@ -4,11 +4,10 @@
 use futures::{StreamExt, future::join_all};
 use jsonrpsee::http_client::{HttpClient, HttpClientBuilder};
 use mysten_common::fatal;
-use rand::{distributions::*, rngs::OsRng, seq::SliceRandom};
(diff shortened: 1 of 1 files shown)


crates/test-cluster/src/lib.rs
```diff
@@ -945,11 +948,12 @@ impl TestCluster {
         client_addr: Option<SocketAddr>,
     ) -> anyhow::Result<(TransactionEffects, TransactionEvents)> {
         let agg = self.authority_aggregator();
-        // Pick a validator to submit to
-        let (_, client) = agg
-            .authority_clients
+        // Pick a validator to submit to, rotating through validators on each call
+        let clients = &agg.authority_clients;
+        let index = self.validator_round_robin.fetch_add(1, Ordering::Relaxed) % clients.len();
```
mystenmark: would it be simpler to just choose randomly?

PR #25991, opened 2026-03-26
feat: proposal for OWS encrypted keystore backend

kevarifin14:
## Problem

`sui.keystore` stores private keys as plaintext Base64 strings in a JSON file:
```json
["AO2qBb...", "ALsXBb..."]
```
File permissions (0o600) are the only protection.

## Proposal

Add an `Ows` variant to the `Keystore` enum in `crates/sui-keys/src/keystore.rs`:

```rust
pub enum Keystore {
    File(FileBasedKeystore),
    InMem(InMemKeystore),
    Ows(OwsKeystore),  // new
}
```

`OwsKeystore` would implement `AccountKeystore` by decrypting keys from an [OWS](https://openwallet.sh) vault (AES-256-GCM, scrypt KDF) via the published `ows-lib` crate.

One wallet created with `ows wallet create` works in Sui CLI, Solana CLI, Foundry, and any other OWS-integrated tool.

### DIFF
--- a/crates/sui-keys/src/ows_keystore_proposal.md
+++ b/crates/sui-keys/src/ows_keystore_proposal.md
@@ -1,0 +1,11 @@
+// NOTE: This is a proposal for adding OWS keystore support.
+// The Keystore enum would gain an Ows variant:
+//
+//   pub enum Keystore {
+//       File(FileBasedKeystore),
+//       InMem(InMemKeystore),
+//       Ows(OWSKeystore),  // <-- new
+//   }
+//
+// OWSKeystore would implement AccountKeystore by decrypting keys
+// from the OWS vault via ows-lib crate.


stefan-mysten:
Hi @kevarifin14,

Thanks for proposing this feature for the Sui CLI.

Could you elaborate a bit more how the OWS works, what would be the main changes, what security issues this might open?

A possible better way to support OWS is to add it as a new signer here: https://github.com/MystenLabs/rust-signers
@jordanjennings-mysten is the owner and driver of that project and he might guide you in a better way.

Finally, we're in the process of rebuilding the CLI and we wouldn't want to put more effort in the current one than needed. That being said, having OWS support seems valuable and probably the `rust-signers` is the way to go here.

PR #25997, opened 2026-03-26
[Block Streaming] introduce the BlockSyncService and implement the Observer client and server sync endpoints

akichidis:
## Description 

This PR is extracting the common processing logic for synchronising blocks and commits to the `BlockSyncService` as the same logic - at least for now - will be used for both the `AuthorityService` and the `ObserverService` endpoints.

The PR also implements the client & server endpoints `fetch_blocks` and `fetch_commits` for the Observer server. After this PR all the Observer endpoints should be working end to end. The `commit_syncer` is also refactored to use the `PeersPool`.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/commit_syncer.rs
+++ b/consensus/core/src/commit_syncer.rs
@@ -504,25 +504,33 @@ where
                         return (commit_range.end(), commits);
                     }
                     Ok(Err(e)) => {
-                        let peer_name = peer.hostname(&inner.context);
-                        warn!("Failed to fetch {commit_range:?} from {peer_name}: {}", e);
+                        warn!(
+                            "Failed to fetch {commit_range:?} from {}: {}",
+                            peer.hostname(&inner.context),
+                            e
+                        );
                         inner
                             .context
                             .metrics
                             .node_metrics
                             .commit_sync_fetch_once_errors
-                            .with_label_values(&[peer_name.as_str(), e.name()])
+                            .with_label_values(&[peer.labelname(&inner.context).as_str(), e.name()])
                             .inc();
                     }
                     Err(_) => {
-                        let peer_name = peer.hostname(&inner.context);
-                        warn!("Timed out fetching {commit_range:?} from {peer}",);
+                        warn!(
+                            "Timed out fetching {commit_range:?} from {}",
+                            peer.hostname(&inner.context)
+                        );
                         inner
                             .context
                             .metrics
                             .node_metrics
                             .commit_sync_fetch_once_errors
-                            .with_label_values(&[peer_name.as_str(), "FetchTimeout"])
+                            .with_label_values(&[
+                                peer.labelname(&inner.context).as_str(),
+                                "FetchTimeout",
+                            ])
                             .inc();
                     }
                 }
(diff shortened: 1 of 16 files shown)


consensus/core/src/network/mod.rs
```diff
@@ -284,6 +284,8 @@ pub(crate) trait ObserverNetworkService: Send + Sync + 'static {
         &self,
         peer: NodeId,
         block_refs: Vec<BlockRef>,
+        highest_accepted_rounds: Vec<Round>,
```
mwtian: I think these are called fetch_after_rounds and fetch_missing_ancestors now.

mwtian (approved):
I think we just need to "update" from the new validator service names and comments, e.g. use fetch_after_rounds and fetch_missing_ancestors in the fetch_blocks() call.

consensus/core/src/observer_service.rs
```diff
@@ -44,6 +45,8 @@ pub(crate) struct ObserverService {
     commit_vote_monitor: Arc<CommitVoteMonitor>,
     transaction_vote_tracker: TransactionVoteTracker,
     synchronizer: Arc<SynchronizerHandle>,
+    #[allow(dead_code)]
```
mwtian: This can be removed now?

consensus/core/src/commit_syncer.rs
```diff
@@ -809,8 +801,13 @@ impl<VC: ValidatorNetworkClient, OC: ObserverNetworkClient> Inner<VC, OC> {
...
                 // start is inclusive, so first commit must be at the start index.
                 if commit.index() != commit_range.start() {
+                    // Extract authority index if it's a validator, otherwise use MAX
+                    let authority = match peer {
+                        PeerId::Validator(idx) => idx,
+                        PeerId::Observer(_) => AuthorityIndex::MAX,
+                    };
                     return Err(ConsensusError::UnexpectedStartCommit {
-                        peer,
+                        peer: authority,
```
mwtian: I don't think this error needs bytes compatibility, so it can be updated to accommodate the new PeerId type.

consensus/core/src/commit_syncer.rs
```diff
@@ -617,7 +607,7 @@ where
                     // 5. Verify the same number of blocks are returned as requested.
                     if request_block_refs.len() != serialized_blocks.len() {
                         return Err(ConsensusError::UnexpectedNumberOfBlocksFetched {
-                            authority: target_authority,
+                            peer: peer.to_string(), // Convert to string for error
```
mwtian: Unnecessary comment?

consensus/core/src/commit_syncer.rs
```diff
@@ -647,7 +637,7 @@ where
                         );
                         if *requested_block_ref != received_block_ref {
                             return Err(ConsensusError::UnexpectedBlockForCommit {
-                                peer: target_authority,
+                                peer: peer.to_string(), // Convert to string for error
```
mwtian: Unnecessary comment?

mwtian (approved):

PR #26000, opened 2026-03-26
[move] compressed type layouts

tzakian:
## Description 

Introduces a compressed, deduplicated representation of Move type layouts in `move-core-types`. Type layout trees (which can be deeply nested with significant subtree duplication) are flattened into a shared node table where identical subtrees are stored once. Primitive types are encoded inline in a tagged `u16` reference and never occupy table entries.

This is the foundational data structure for replacing the recursive tree-based `MoveTypeLayout` throughout the codebase. This PR adds the types, builder, view layer, and deserialization support — it does not yet change any existing consumers.

### Design

**Encoding:** Each layout node reference (`LayoutRef`) is a `u16` where bit 15 distinguishes leaves from table indices:
- Bit 15 set → primitive type encoded in the low bits (Bool, U8, ..., Signer — 9 variants)
- Bit 15 clear → index into the compound node table (max 32,767 entries)

**Node table:** Only compound types (Vector, Struct, Enum) are stored. For the annotated variant, field names (`Identifier`) and type tags (`StructTag`) are stored inline in nodes — no separate interning tables.

**Sharing:** The node table is behind `Arc<[MoveTypeNode]>`. Cloning a layout or extracting a sub-layout (e.g., "give me the layout of field 3") is just an `Arc` bump + new root index. No data is copied.

**Two variants:**
- `compressed::runtime` — field layouts only (no names/tags), matching tree `runtime_value::MoveTypeLayout`
- `compressed::annotated` — field names + struct tags inline in nodes, matching tree `annotated_value::MoveTypeLayout`

Both share `LayoutRef`, `LeafType`, `LayoutHandle`, and `ResolvedRef` from `compressed::mod.rs`.

### Example

A tree layout for `struct Wrapper { x: Inner, y: Inner }` where `Inner = struct { a: u64, b: bool }`:

**Tree representation** (7 nodes, `Inner` duplicated):
```
Struct([Struct([U64, Bool]), Struct([U64, Bool])])
```

**Compressed representation** (2 nodes, `Inner` shared):
```
nodes: [
  Struct { fields: [U64↑, Bool↑] },     // index 0 = Inner
  Struct { fields: [0, 0] },             // index 1 = Wrapper (both fields → index 0)
]
root: 1
```
(↑ = leaf ref encoded inline, not in table)

### Public API

**Building:**
```rust
let mut b = MoveTypeLayoutBuilder::new();
let u64_h = b.u64();      // returns LayoutHandle (leaf, no table entry)
let bool_h = b.bool();
let inner = b.struct_layout(&tag, &[(&field_a, u64_h), (&field_b, bool_h)])?;
let wrapper = b.struct_layout(&tag2, &[(&field_x, inner), (&field_y, inner)])?; // deduped
let layout = b.build(wrapper);
```

**Inflation back to tree:**
```rust
let tree: MoveTypeLayout = layout.inflate()?; // fallible if unknown enum variants
```

### Unknown enum variants

Enum variants can be marked as `None` (layout unknown but variant exists). Views return `VariantFieldView::Unknown` for these. Deserialization and inflation return errors if they encounter unknown variants. This supports partial layout knowledge for enums where not all variants have been resolved (_this is future looking_).

### Test Plan

31 tests covering:
- **Roundtrip:** Compress tree → inflate back, verify equality (all types: primitives, vectors, structs, enums, nested)
- **Deduplication:** Verify shared subtrees produce fewer nodes (primitive dedup, struct dedup, shared subtree dedup)
- **View navigation:** Struct field access, enum variant lookup, vector element resolution
- **Deserialization parity:** BCS deserialization via tree path vs compressed path produces identical `MoveValue` results (primitives, structs, vectors, enums, nested structs, nested vectors, complex nested types)

```bash
cd external-crates/move
SUI_SKIP_SIMTESTS=1 cargo nextest run -p move-core-types --lib -- compressed
# 31 tests, all pass
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
+++ b/external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
@@ -263,6 +263,7 @@ impl MoveLayoutView {
             MoveLayoutView::Enum(ev) => {
                 let variants = ev
                     .variants()
+                    .iter()
                     .map(|vl| match vl.fields() {
                         Some(fields) => {
                             let field_layouts = fields
@@ -406,6 +407,7 @@ impl MoveDatatypeLayout {
             MoveDatatypeLayout_::Enum(move_enum_layout) => {
                 let variants = move_enum_layout
                     .variants()
+                    .iter()
                     .map(|vl| match vl {
                         VariantLayout::Known { name, tag, fields } => {
                             let field_layouts = fields
@@ -590,15 +592,15 @@ impl MoveEnumLayout {
     }
 
     /// Iterate over all variants.
-    pub fn variants(&self) -> impl ExactSizeIterator<Item = &VariantLayout> {
-        self.variants.iter()
+    pub fn variants(&self) -> &[VariantLayout] {
+        &self.variants
     }
 }
 
 impl fmt::Display for MoveEnumLayout {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{} {{ ", self.type_)?;
-        for (i, vl) in self.variants().enumerate() {
+        for (i, vl) in self.variants().iter().enumerate() {
             if i > 0 {
                 write!(f, ", ")?;
             }
--- a/external-crates/move/crates/move-core-types/src/compressed/runtime/layout.rs
+++ b/external-crates/move/crates/move-core-types/src/compressed/runtime/layout.rs
@@ -90,7 +90,7 @@ pub enum MoveDatatypeLayout {
 /// The enum layout of an enum type, as a view into a shared pool.
 #[derive(Debug, Clone)]
 pub struct MoveEnumLayout {
-    pub variants: Box<[VariantLayout]>,
+    pub(crate) variants: Box<[VariantLayout]>,
 }
 
 /// The struct layout of a struct type, as a view into a shared pool.
@@ -242,6 +242,7 @@ impl MoveLayoutView {
             MoveLayoutView::Enum(ev) => {
                 let variants = ev
                     .variants()
+                    .iter()
                     .map(|vfv| match vfv {
                         VariantLayout::Known(fv) => {
                             fv.fields().map(|f| f.inflate()).collect::<AResult<_>>()
@@ -339,15 +340,15 @@ impl MoveEnumLayout {
     }
 
     /// Iterate over all variants.
-    pub fn variants(&self) -> impl ExactSizeIterator<Item = &VariantLayout> {
-        self.variants.iter()
+    pub fn variants(&self) -> &[VariantLayout] {
+        &self.variants
     }
 }
 
 impl fmt::Display for MoveEnumLayout {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "enum ")?;
-        for (tag, vfv) in self.variants().enumerate() {
+        for (tag, vfv) in self.variants().iter().enumerate() {
             write!(f, "variant_tag: {} {{ ", tag)?;
             match vfv {
                 VariantLayout::Known(fv) => {


tzakian:
For generating the tree-layout -- that is just as an interim compatibility layer (hopefully). I would love/plan on this going away completely, but in the interim it would allow us to "inflate at boundaries" as we push this across the codebase.

Rc is useful to allow for cheaper clones/copies when creating sub-layouts that need to be stored (e.g., what we do in some places in the tracer). Originally this started as a `Box` but Rc-ing (well, soon to be `Arc`-ing so they're Send/Sync) seemed the sensible thing to do there.

I think once we start looking at the views, it makes a bit more sense -- I actually generally quite like where this has come out and from an internal representation POV (and I believe @tnowacki and @cgswords are overall happy with this modulo some internal representation discussions from chatting with them about this, but @cgswords take is even more extreme than this ;) ).

In terms of interning `Identifiers` and tags, this seemed like the reasonable thing to do in this case (and if we're interning nodes interning other possibly shared data seems to make sense to me). But if not doing that would make you happier I can not do that, but it feels like making things take more space unnecessarily. 

In terms of sharing between runtime and annotated layouts -- good point, I think not sharing anything between the two is the right call. I will do that and update this PR with the most recent state in a little bit.

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,1018 @@
...
+}
+
+impl<'a> MoveEnumView<'a> {
+    /// The enum's type tag.
+    pub fn type_(&self) -> &'a StructTag {
+        self.type_
+    }
+
+    pub fn is_type(&self, t: &TypeTag) -> bool {
+        matches!(t, TypeTag::Struct(s) if **s == self.type_().clone())
```
akmysti: Is `clone()` needed here?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,1018 @@
...
+
+    /// Access a field's raw ref by index (for sublayout).
+    pub(crate) fn raw_field(&self, i: usize) -> Option<(&'a Identifier, LayoutRef)> {
+        self.fields
+            .get(i)
+            .map(|(name_idx, layout_ref)| (&self.pool.strings[*name_idx as usize], *layout_ref))
+    }
+
+    /// Look up a field by name, returning its layout view.
+    pub fn field_by_name(&self, name: &str) -> Option<MoveLayoutView<'a>> {
```
akmysti: Any plans to make this cheaper than `O(n)` or this is not a concern to begin with?

akmysti: Any plans to make this cheaper than `O(n)` or this is not a concern to begin with?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
tzakian: Nope!

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
tzakian: Not a concern for now (IMO) -- we shouldn't be doing this _too_ frequently, and also the number of fields is quite low and therefore linear search is generally faster.

external-crates/move/crates/move-core-types/src/compressed/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+
+/// The result of resolving a [`LayoutRef`].
+pub(crate) enum ResolvedRef {
+    Leaf(LeafType),
+    Index(usize),
+}
+
+/// An opaque handle to a layout node returned by the builder.
+///
+/// Handles are only useful for passing back into the same builder (to compose
```
akmysti, line 65: Is there a way/need to track the builder that created it?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,857 @@
...
+
+/// The shared node table backing a [`MoveTypeLayout`].
+pub(crate) type MoveTypeLayoutPool = [MoveTypeNode];
+
+// --- Owned layout types ---
+
+/// A deduplicated, flat representation of an annotated [`MoveTypeLayout`] tree.
+/// Names and type tags are stored inline in nodes. Cloning is cheap — the
+/// node table is shared via `Arc`.
+#[derive(Debug, Clone)]
```
akmysti, line 73: Will it help to compare (at least equality) two MoveTypeLayouts

akmysti:
In the commit message there is `let value = MoveValue::simple_deserialize_compressed(&bytes, &layout)?;`, dont see that function in this PR.

external-crates/move/crates/move-core-types/src/unit_tests/compressed_layout_test.rs
```diff
@@ -0,0 +1,735 @@
...
+        ],
+    }));
+    let compressed = AC::MoveTypeLayout::try_from(&layout).unwrap();
+    assert_eq!(compressed.node_count(), 1);
+    let inflated = compressed.inflate().unwrap();
+    assert_eq!(inflated, layout);
+}
+
+#[test]
+fn annotated_string_dedup() {
```
akmysti: this looks like a roundtrip test not dedup.

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,857 @@
...
+    pub(crate) fields: MoveFieldView<'a>,
+}
+
+/// The result of looking up a variant in an annotated enum view.
+#[derive(Debug, Clone, Copy)]
+pub enum VariantFieldView<'a> {
+    /// The variant's field layout is known.
+    Known(MoveFieldView<'a>),
+    /// The variant exists but its field layout is not available.
+    Unknown,
```
akmysti: Don't see any (negative) test for Unknown field view.

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
tzakian: Good question. I think we should explicitly _not_ derive Eq, PartialEq, (etc.) on these as two layouts may be directly non-equal but in fact be equal (due to permutation of the pool).

I think this is worth adding as a comment and will add that in :)

external-crates/move/crates/move-core-types/src/compressed/mod.rs
tzakian: Not really (or at least no easily...). I played around with some possible things but it would make dealing with these things _insanely_ hard if not bordering on impossible I think.

external-crates/move/crates/move-core-types/src/compressed/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+pub mod annotated;
+pub mod runtime;
+
+use std::sync::Arc;
+
+// =============================================================================
+// Shared types used by both runtime and annotated compressed layouts
+// =============================================================================
+
+pub(crate) type Shared<T> = Arc<T>;
```
tnowacki: Do we want to just remove this for `Arc` everywhere? (might be more clear)
Or do you like the type alias?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,824 @@
...
+    pub(crate) fields: MoveFieldsLayout,
+}
+
+/// The result of looking up a variant in an annotated enum layout.
+#[derive(Debug, Clone)]
+pub enum VariantLayout {
+    /// The variant's field layout is known.
+    Known {
+        name: Identifier,
+        tag: u16,
```
tnowacki: Should we use the file format's type alias here? I think its something like VariantTag?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,824 @@
...
+    Signer,
+    Vector(Box<MoveTypeLayout>),
+    Struct(Box<MoveStructLayout>),
+    Enum(Box<MoveEnumLayout>),
+}
+
+/// A compressed layout that is known to be a struct or enum (not a primitive
+/// or vector). This mirrors the tree-based [`crate::annotated_value::MoveDatatypeLayout`].
+#[derive(Debug, Clone)]
+pub enum MoveDatatypeLayout_ {
```
tnowacki: Should this be `pub(crate)`?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,824 @@
...
+        &self.type_
+    }
+
+    /// Check whether this struct's type tag matches the given [`TypeTag`].
+    pub fn is_type(&self, t: &TypeTag) -> bool {
+        matches!(t, TypeTag::Struct(s) if &**s == self.type_())
+    }
+
+    /// Access the fields layout.
+    pub fn field_view(&self) -> &MoveFieldsLayout {
```
tnowacki: This isn't really a view anymore. `fields_layout`?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,824 @@
...
+    /// The variant's name.
+    pub fn name(&self) -> &Identifier {
+        match self {
+            VariantLayout::Known { name, .. } => name,
+            VariantLayout::Unknown { name, .. } => name,
+        }
+    }
+
+    /// The variant's tag.
+    pub fn tag(&self) -> u16 {
```
tnowacki: ditto on type alias here

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,824 @@
...
+// --- MoveTypeLayoutBuilder ---
+
+impl MoveTypeLayoutBuilder {
+    pub fn new() -> Self {
+        Self {
+            nodes: IndexSet::new(),
+        }
+    }
+
+    fn intern(&mut self, node: MoveTypeNode) -> AResult<LayoutHandle> {
```
tnowacki: nit, rename per discussion

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,824 @@
...
+        self.intern(MoveTypeNode::Enum(MoveEnumNode {
+            type_: type_tag.clone(),
+            variants: variant_entries,
+        }))
+    }
+
+    /// Recursively intern a tree-based annotated layout.
+    /// Tree-based enum layouts always have known variants, so all variants
+    /// are wrapped in `Some`.
+    pub fn intern_tree(&mut self, layout: &AV::MoveTypeLayout) -> AResult<LayoutHandle> {
```
tnowacki: From tree? The name "intern" here is strange to me (nits nonwithstanding)

external-crates/move/crates/move-core-types/src/compressed/mod.rs
tzakian: I'm fine using Arc directly, and may be more clear so will just inline I think.

This may have been done in a fit of rage after going from Box -> Rc -> Arc and was like...I'm just gonna alias it lol but I think Arc is pretty settled at this point so inlining makes sense.

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
tzakian: 🤔 Possibly, but I have a sneaking suspicion that it will need to be `public` eventually. For now I'll make it pub(crate) and then if we need to relax that to `pub` we can take a look at it at that point.

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
tzakian: gah... bad refactor thanks for catching

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
tzakian: ah this is below the file format so we can't use `VariantTag` here sadly (circular deps and whatnot)

tnowacki: Should we just create yet-another-type-alias? Might be better than a raw u16

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,830 @@
...
+        &self.type_
+    }
+
+    /// Check whether this struct's type tag matches the given [`TypeTag`].
+    pub fn is_type(&self, t: &TypeTag) -> bool {
+        matches!(t, TypeTag::Struct(s) if &**s == self.type_())
+    }
+
+    /// Access the fields layout.
+    pub fn fields_layout(&self) -> &MoveFieldsLayout {
```
tnowacki, line 501: I keep reading this and stumbling like "attorneys general"

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,830 @@
...
+    std::sync::LazyLock::new(|| Arc::from(Vec::<MoveTypeNode>::new()));
+
+// --- Node types ---
+
+/// A (field_name, layout_ref) pair for struct/enum fields.
+pub(crate) type AnnotatedFieldEntry = (Identifier, LayoutRef);
+
+/// A single variant entry: (variant_name, tag, optional fields).
+/// `None` fields means the variant exists but its layout is unknown.
+pub(crate) type AnnotatedVariantEntry = (Identifier, u16, Option<Box<[AnnotatedFieldEntry]>>);
```
cgswords: The boxes here are not strictly: they are not held as part of the `MoveTypeLayoutPool` defined here. I am not sure if that is intentional, just making a comment. You could encode this linearly in the `MoveTypeLayoutPool` instead if you wanted to.

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,830 @@
...
+    }
+
+    fn leaf(ty: LeafType) -> Self {
+        MoveTypeLayout {
+            pool: EMPTY_POOL.clone(),
+            root: LayoutRef::leaf(ty),
+        }
+    }
+
+    pub fn bool() -> Self {
```
cgswords, line 173: Can you `const` these? Or is the allocation stopping us?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,830 @@
...
+            }
+            MoveLayoutView::Enum(ev) => {
+                let variants = ev
+                    .variants()
+                    .map(|vl| match vl.fields() {
+                        Some(fields) => {
+                            let field_layouts = fields
+                                .fields()
+                                .map(|(name, layout)| {
+                                    Ok(AV::MoveFieldLayout::new(name.clone(), layout.inflate()?))
```
cgswords, line 272: This layout is rather eager; do we want it to be?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,830 @@
...
+                AV::MoveTypeLayout::Enum(Box::new(TreeEnumLayout {
+                    type_: ev.type_().clone(),
+                    variants,
+                }))
+            }
+        })
+    }
+
+    /// Check whether this layout matches the given [`TypeTag`].
+    pub fn is_type(&self, t: &TypeTag) -> bool {
```
cgswords: Nit: `is_type_tag`?

cgswords: Or `PartialEq`? 🤔

external-crates/move/crates/move-core-types/src/compressed/runtime/layout.rs
```diff
@@ -0,0 +1,550 @@
...
+
+/// A deduplicated, flat representation of a [`RV::MoveTypeLayout`] tree.
+/// Cloning is cheap — the pool is shared via `Arc`.
+///
+/// NOTE: `Eq`/`PartialEq`/`Hash` are intentionally not derived. Two layouts
+/// representing the same type may have different pool orderings (node
+/// permutations), so structural equality on the raw fields would produce
+/// false negatives. Compare by inflating to tree form or by comparing views.
+#[derive(Debug, Clone)]
+pub struct MoveTypeLayout {
```
cgswords, line 60: How would you feel about abstracting this into a `MoveTypeLayout<Pool>`? And then we could newtype out `MoveTypeLayout(MoveTypeLayout<RuntimePool>)` and `MoveTypeLayout(MoveTypeLayout<AnnotationPool>)`?

external-crates/move/crates/move-core-types/src/compressed/runtime/layout.rs
```diff
@@ -0,0 +1,550 @@
...
+    U16,
+    U32,
+    U64,
+    U128,
+    U256,
+    Address,
+    Signer,
+    Vector(Box<MoveTypeLayout>),
+    Struct(Box<MoveStructLayout>),
+    Enum(Box<MoveEnumLayout>),
```
cgswords, line 80: Same question here.

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,830 @@
+// Copyright (c) The Diem Core Contributors
```
cgswords, line 1: Inconsistent license header

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,830 @@
...
+    }
+}
+
+/// Resolve a [`LayoutRef`] against the pool into a [`MoveLayoutView`].
+///
+/// Panics if the reference points to an out-of-bounds table index.
+fn resolve_ref(pool: &Arc<MoveTypeLayoutPool>, r: LayoutRef) -> MoveLayoutView {
+    match r.resolve() {
+        ResolvedRef::Leaf(leaf) => leaf_to_layout_view(leaf),
+        ResolvedRef::Index(idx) => match &pool[idx] {
```
cgswords, line 823: Panic intentionally here?

external-crates/move/crates/move-core-types/src/compressed/runtime/layout.rs
```diff
@@ -0,0 +1,550 @@
...
+/// node table.
+pub struct MoveTypeLayoutBuilder {
+    nodes: IndexSet<MoveTypeNode>,
+}
+
+// --- Display helper ---
+
+/// Helper type that uses `T`'s `Display` implementation as its own `Debug` implementation,
+/// to allow other `Display` implementations to take advantage of structured formatting helpers.
+struct DebugAsDisplay<'a, T>(&'a T);
```
cgswords, line 129: This is cool but should not go in this file.

external-crates/move/crates/move-core-types/src/compressed/runtime/layout.rs
```diff
@@ -0,0 +1,550 @@
...
+#[derive(Debug, Clone)]
+pub enum MoveDatatypeLayout {
+    Struct(Box<MoveStructLayout>),
+    Enum(Box<MoveEnumLayout>),
+}
+
+/// The enum layout of an enum type, as a view into a shared pool.
+#[derive(Debug, Clone)]
+pub struct MoveEnumLayout {
+    pub variants: Box<[VariantLayout]>,
```
cgswords:
```suggestion
    pub(crate) variants: Box<[VariantLayout]>,
```

And add an accessor

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
cgswords: Also, naming these fields might be sanity-saving later, even though it is less ergonomic to match on.

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
```diff
@@ -0,0 +1,830 @@
...
+pub(crate) enum MoveDatatypeLayout_ {
+    Struct(Box<MoveStructLayout>),
+    Enum(Box<MoveEnumLayout>),
+}
+
+/// Datatype layout with a reference to the original layout for inflation and conversion.
+#[derive(Debug, Clone)]
+pub struct MoveDatatypeLayout {
+    self_layout: MoveTypeLayout,
+    inner: MoveDatatypeLayout_,
```
cgswords, line 109: This field (and the type above) seem redundant based on usage in this file: we only use it for inflation, and I believe we could do that directly by consulting `self_layout`.

tzakian: It is, although when we go into the inner layout we would then need to trust that the `MoveTypeLayout` is actually for an enum/data layout vs already holding it structurally. It felt a bit nicer to me to do this, but nothing prevents just implementing it as a newtype wrapper and trusting that the internal type is a datatype.

external-crates/move/crates/move-core-types/src/compressed/runtime/layout.rs
tzakian: This was replicating the structure we have elsewhere for this around the layouts, so I'd like to keep here for now, but we can then reconfigure where things are after the larger changes if that's ok?

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
tzakian: Don't really have an option other than to make this fallible which would make this API super annoying to deal with I think. I think trusting internal consistency here makes sense to me.

external-crates/move/crates/move-core-types/src/compressed/runtime/layout.rs
tzakian: Will give this a shot and lets see... 🤞🏻

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
tzakian: I think this amount of eagerness is fine and makes the API much nicer to work with. We can maybe re-explore once we have this pushed out into usage sites and can see how much pain being lazier would be but IMO the pain to actual perf improvement here is minimal.

external-crates/move/crates/move-core-types/src/compressed/annotated/layout.rs
tzakian: allocation is stopping it sadly :'(

external-crates/move/crates/move-core-types/src/compressed/runtime/layout.rs
tzakian: Tried it out and we don't really gain much for a lot of complexity and overhead. I think lets keep this as-is for now (and it reflects the separation that we have today with the runtime and annotated layouts as well).

PR #26004, opened 2026-03-26
[forking 1/n] basic skeleton + design docs

stefan-mysten:
## Description 

This PR provides the basic skeleton for the forking tool, design docs, and the stacked PRs list that will follow.

Stack:
https://github.com/MystenLabs/sui/pull/26004
## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-forking/poc.md
+++ b/crates/sui-forking/poc.md
@@ -88,13 +88,13 @@ Note that seeding can also be done from a file:
 ```json
 {
     "network": "testnet",
-    "checkpoint": "12345678",
+    "checkpoint": 12345678,
     "addresses": [
         "0x1234567890abcdef1234567890abcdef12345678",
         "0xabcdef1234567890abcdef1234567890abcdef12"
     ],
     "objects": [
-        "0xabcdef1234567890abcdef1234567890abcdef12
+        "0xabcdef1234567890abcdef1234567890abcdef12"
     ]
 }
 ```
@@ -105,9 +105,9 @@ The initial seed will be dumped to a file `generated_{network}_{checkpoint}.json
 ```json
 {
     "network": "testnet",
-    "checkpoint": "12345678",
+    "checkpoint": 12345678,
     "objects": [
-        "0xabcdef1234567890abcdef1234567890abcdef12
+        "0xabcdef1234567890abcdef1234567890abcdef12"
     ]
 }
 ```
@@ -124,7 +124,7 @@ sui-forking start --network testnet
 
 This command:
 
-- Starts a local “*forked*” network on port 9000 (default) - this is used to interact with the forked network specific commands, e.g., advance-checkpoints, advance-clock, status. You can do so via the CLI or via a gRPC API.
+- Starts a local “*forked*” network on port 9000 (default) - this is used to interact with the forked network, e.g., through the Sui CLI or programmatically through gRPC clients. This will also expose a gRPC API for manually controlling the network (clock, checkpoints, epoch).
 - Starts the RPC server on port 9000 (default) - this is the gRPC endpoint you can connect the Sui client to interact with the network.
 - Allows you to execute transactions against this local state and fetches objects on-demand from the real network
 
--- a/crates/sui-forking/prs.md
+++ b/crates/sui-forking/prs.md
@@ -4,9 +4,9 @@
 
 - this should have the minimal `sui-forking` crate
 - Cargo.toml
-- lib.rs
+- main.rs
 - [README.md](http://README.md) file
-- design.md
+- poc.md
 - prs.md
 
 ## Forking PR2 -> update forking-data-store
@@ -25,11 +25,10 @@
 - use the forking-data-store for the simulacrum instantiation
 - implement the checkpoint related APIs to initialize the forked network
 
-## Forking PR6 -> Implement transaction execution
+## Forking PR6 -> implement transaction execution
 - implement transaction execution logic in sui-forking & related gRPC APIs
 
 Future PRs: TBD (probably a PR per line below)
-- implement subscription service
 - implement simulate transaction API
 - implement rest of gRPC APIs (dynamic fields, epoch, etc)
 - implement the forking CLI and related gRPC APIs to interact with the network
--- a/crates/sui-forking/README.md
+++ b/crates/sui-forking/README.md
@@ -31,7 +31,7 @@ The table below summarizes when to use each option:
 | Existing on-chain packages and objects | Available from the fork point (fetched/cached on demand) | Not available unless you deploy/create them locally |
 | External dependency at runtime | Needs network access to source chain for first-time object fetches | Fully local once started |
 | Execution model | Single validator, lock-step, sequential execution | Multi-validator local network flow |
-| Checkpoint/time/epoch control | Explicit control through `advance-checkpoint`, `advance-clock`, `advance-epoch` | Driven by normal local network progression |
+| Checkpoint/time/epoch control | Explicit control through `advance-checkpoint`, `advance-clock` | Driven by normal local network progression |
 | Best for | Testing against real deployed packages and realistic chain state | Fast local development from clean state |
 | Startup cost | Higher (state bootstrap + potential object downloads) | Lower (local genesis and startup) |
 | Determinism/reproducibility | Deterministic from selected checkpoint + seeded objects | Deterministic from local genesis/configuration |


crates/sui-forking/src/lib.rs
amnn, line 1: Should we re-use `bin_version` (If so, it belongs in `main.rs`).

crates/sui-forking/design.md
```diff
@@ -0,0 +1,198 @@
...
+- Develop locally while maintaining consistency with production state
+- Run integration tests against forked network state and using packages deployed on the real live network
+
+Important to note: the forking tool spins up a network that is not generating checkpoints automatically. The network requires manual intervention
+
+# Design
+
+### **High level diagram**
+
+![image.png](image.png)
```
amnn:
```suggestion
![architecture.png](architecture.png)
```

crates/sui-forking/design.md
```diff
@@ -0,0 +1,198 @@
...
+### **Data Layer**
+
+`simulacrum` requires to use a `store` where all the required data (checkpoints, transactions, objects, etc) lives. To this end, a `ServiceStore` type that implements the required traits from `simulacrum` (e.g., `SimulatorStore`, `BackingStore`, `ObjectStore`, etc.), is implemented.
+
+Ultimately, the goal is to get to something like this for this store:
+
+```jsx
+pub struct ServiceStore {
+    /// Capability-routed composite store:
+    /// - transactions/epochs/checkpoints: memory -> filesystem (note in POC there is no historical data provided)
```
amnn: Let's drop support for historical queries for the POC.

crates/sui-forking/design.md
```diff
@@ -0,0 +1,198 @@
...
+    /// - objects: memory -> filesystem -> GraphQL
+    store: ForkDataStore,
+
+    // The checkpoint at which this forked network was forked
+    forked_at_checkpoint: u64,
+
+    /// Optional validator-set override used when building epoch state for checkpoint production.
```
amnn: By dropping support for advance epoch initially, we can avoid this override and the related complications.

crates/sui-forking/design.md
```diff
@@ -0,0 +1,198 @@
...
+**Store Layer Composition**
+
+The `CompositeStore` is a store defined in `forking-data-store` that routes each data capability to a store type. The building blocks come from the `forking-data-store` crate:
+
+- `WriteThroughStore<Primary, Secondary>`: writes to both; reads from primary first, falls back to secondary, writes to secondary, and then writes to primary.
+- `ReadThroughStore<Primary, Fallback>`: reads from primary, falls back to secondary and caches result into primary.
+
+The concrete layers:
+
+```rust
```
amnn: There's a fair bit of complexity here with different kinds of data needing different data hierarchies. It should be possible to simplify this so all kinds of data follow the same basic structure: a two-layer store of a "cache" in front of GraphQL.

- Updates from execution result in writes to the cache.
- Reads from GraphQL also write back into the cache, to reduce reliance on GraphQL over time.
- To begin with (for the POC), the cache can be purely in-memory, and we can worry about introducing persistence and overrides in the filesystem later. (cc @dariorussi, I think @stefan-mysten mentioned you had discussed something along these lines).

We stil need to take some care around cache keys (especially for object look-ups) to make sure that e.g. when you fetch an object at a version V that does not exist yet, we cache the latest object at the fork point, to avoid caching a stale entry (that the version will never exist) or repeatedly calling back to GraphQL for that query, until a version of that object is written.

Overall this should be simpler and also more immediately featureful as any point lookup should work (even for historical objects). It only leaves historical pagination for after the POC.

crates/sui-forking/design.md
```diff
@@ -0,0 +1,198 @@
...
+>
+```
+
+Objects have a third tier (GraphQL) because they are fetched on-demand from the
+live network when not found locally. Transactions, epochs, and checkpoints are
+only produced locally (or fetched once at startup via gRPC), so two tiers suffice.
+
+**Filesystem Layout**
+
+The user has the option to specify where to store the network’s state through a `--data-dir` flag at startup. If none is provided, the default `~/.forking_store` will be used.
```
amnn: Simplification for the POC, just always expect to be provided a `--data-dir`.

crates/sui-forking/design.md
```diff
@@ -0,0 +1,198 @@
...
+ <data_dir>/forking/<network>/forked_at_checkpoint_<N>
+	 objects
+	 checkpoints
+	 transactions
+```
+
+Storing data on filesystem enables resuming a previously forked session — on restart, if the user provides a checkpoint to fork from that was previously used, the tool will resume from that state that should exist on the filesystem.
+In the case that the user wants to fork from that checkpoint again on a fresh & clean state, the tool provides a `--reset` flag to remove that directory and start fresh.
+
+### **Startup object seeding**
```
amnn: Simpler and more expressive:

- `--address` adds an address for seeding (works in the consistent range), loads that address's objects and adds them to the seed.
- `--object` add the object by ID directly to the seed.

They don't need to be mutually exclusive, and each can be supplied multiple times.

You can also consider accepting a seed file that just has a list of object IDs, which is something you can also generate as part of initializing a network, so that it's easy to seed by address(es) when a checkpoint is recent, and then re-use that seed later (because at that point the seed has been normalized to only mention objects).

crates/sui-forking/design.md
```diff
@@ -0,0 +1,198 @@
...
+
+At startup, the user has the choice to seed addresses or objects, to make the forked network “aware” of them. There are two different modes on how this works, due to the data-limitations that we have in GraphQL:
+
+- `--accounts`: discover owned objects through GraphQL at startup time for checkpoints not older than 1h.
+- `--objects`: provide explicit object IDs to prefetch at startup time, required for checkpoints older than 1h.
+
+note that these two args are mutually exclusive.
+
+### Validators
```
amnn: (no longer necessary for the POC, hopefully)

crates/sui-forking/prs.md
amnn, line 1: From our discussion -- after this PR, and once you have buy-in on the design, it may be better to start with a basic version of the store, as the foundational component, and then come back to the minimal flow that uses that store.

amnn: (You do not need to complete the store implementation before building the first interaction in `sui-forking`, but seeing as we know that data fetching will end up there, it makes sense to grow that code from there).

crates/sui-forking/design.md
stefan-mysten: For POC this is not valid anymore as we're going with in memory only.

crates/sui-forking/design.md
```diff
@@ -0,0 +1,239 @@
...
+
+    // The checkpoint at which this forked network was forked
+    forked_at_checkpoint: u64,
+}
+```
+
+**Store Layer (forking-data-store)**
+
+In the initial POC, the store will be an in-memory store with a GraphQL client as the backing source for historical data (checkpoints, epochs, objects, transactions).
+
```
stefan-mysten: Switch to FileSystem cache.

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+
+In the initial POC, the tool will have a store that persists data on disk, and resolves various queries through a GraphQL client as the backing source for historical data (checkpoints, epochs, objects, transactions).
+
+** Checkpoints | Epochs | Transactions **
+
+Expected read flow:
+- a checkpoint|epoch|tx is identified by its sequence number|number|digest
+- on read, check on disk first
+- on miss, fetch from backing source
+- if the data is found but is later than the fork checkpoint, return `None`
```
amnn: This is not a case you need to worry about because GraphQL does this for you, given the following query pattern:

```graphql
query TransactionAtCheckpoint($tx: String, $cp: UInt53) {
    checkpoint(sequenceNumber: $cp) {
        query {
            transaction(digest: $tx) {
                # ...
            }
        }
    }
}
```

If the transaction with that digest exists at that checkpoint, it will be provided, otherwise you will get `null` (same for checkpoints, epochs, objects, etc). This is the point that @dariorussi was making in the meeting, I think.

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+- an object is identified by its ID and version (or other query, e.g., latest version, version at checkpoint, etc)
+- on read, check on disk first
+- on miss, fetch from backing source by ID and version/query
+- if the object is found but is later than the fork checkpoint, cache the latest object at the forked checkpoint and return `None`.
+- if the object is not found, return `None`
+
+When an object is updated (e.g., by executing a transaction that changes the object), the new version of the object is persisted to disk and the metadata around latest version is updated.
+
+```rust
+pub trait CheckpointReader {
```
amnn: It is hard to judge whether this set of traits and types makes sense, because you can't see how they are going to be used. The point of breaking your changes up is to introduce concepts when they are well-motivated. I would suggest dropping most if not all code from your design at this stage, to focus on the high level shape.

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+At startup, the user has the choice to seed addresses or objects, to make the forked network “aware” of them.
+
+`--address` adds an address for seeding (works in the consistent range), loads that address's objects and adds them to the seed.
+`--object` add the object by ID directly to the seed.
+
+Note that seeding can also be done from a file:
+
+```json
+{
+    "network": "testnet",
```
amnn, line 90: Need to give some thought to what the identifier would be here, to account for forking from a devnet, localnet(s), another simulated network or forked network, etc.

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+    "checkpoint": "12345678",
+    "addresses": [
+        "0x1234567890abcdef1234567890abcdef12345678",
+        "0xabcdef1234567890abcdef1234567890abcdef12"
+    ],
```
amnn, line 95: The entire point of the seeds file is to capture the initial (object) seeds AFTER you have converted addresses into their sets of owned objects, so that if you wanted to generate a fresh version of the forked network at some later date when the checkpoint is no longer recent, you don't need to perform that mapping. This field should not exist at all.

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+
+Start a local forked network at the latest checkpoint:
+
+```bash
+sui-forking start --network testnet
+```
+
+This command:
+
+- Starts a local “*forked*” network on port 9001 (default) - this is used to interact with the forked network, e.g., advance-checkpoints, request gas, advance-clock, advance-epoch, etc. You can do so via the CLI commands with the `sui-forking` binary or via the REST API (see below).
```
amnn: If you are using gRPC for interacting with this, then you would use the same port to interact with both the fullnode gRPC and to control the network.

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+**The startup flow**
+- Initialize store layer (forking-data-store)
+- Fetch the latest checkpoint (or the checkpoint specified by the user)
+- Wait for commands to advance the network (checkpoint, clock) on the forking service, or gRPC requests.
+
+### POC CLI
+
+The forking tool provides a CLI to interact with the forking-server for various actions. In addition to the `sui-forking start` command explained previously, there are a few other commands available:
+
+**Faucet - request SUI tokens**
```
amnn: How does the faucet actually work? (Where are the coins coming from?)

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Execution Engine
+
+Under the hood, the tool uses `simulacrum` to manage the state of the network and execute transactions. In a nutshell, simulacrum has an API for creating & handling checkpoints, objects, transactions, transaction events, executing transactions, etc.
+
+When a transaction execution request comes in from gRPC, it will be routed by the gRPC API and passed to the `simulacrum`. Before executing the transaction, there are a few more steps needed to successfully execute the transaction:
+- fetch any missing input objects (this delegates fetching data to the data-layer)
+- sign the transaction with a dummy private key (allows for impersonating senders)
+- execute the transaction and get back the effects
+- create a checkpoint
```
mdgeorge4153, line 44: The intro says that the forking tool doesn't auto-crank but this says it does. I remember there being discussion back and forth about this, but I don't remember where it settled.

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
+# Forking Tool Design, Implementation, & PR execution - POC
+
+`sui-forking` allows developers to start a local network in lock-step mode and execute transactions against some initial state derived from the live Sui network. This enables you to:
+
+- Depend on existing on-chain packages and data
+- Test contracts that interact with real deployed packages
+- Develop locally while maintaining consistency with production state
+- Run integration tests against forked network state and using packages deployed on the real live network
```
mdgeorge4153, line 8: Is GraphQL support planned as a follow-up? Might be worth saying that it's explicitly out of scope or planned for the future or whatever.

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+    pub fn write_epoch(&self, epoch: Epoch) -> Result<(), StoreError> {
+        // persist to local store
+}
+```
+
+As data is persisted to disk, the user needs to provide a directory where the network state should be stored. Upon restarting the tool, if the directory exists and contains valid data, the tool can reuse the existing data and continue from there. This allows the user to maintain the state of the forked network across restarts.
+
+### **Startup object seeding**
+
+At startup, the user has the choice to seed addresses or objects, to make the forked network “aware” of them.
```
mdgeorge4153, line 80: My understanding coming out of the discussion was that for the POC, only data that is written to the fork would be stored, all the reads would be passed through (possibly going through a shared global cache layer, but that's out of scope for the POC). If that's the case, I don't understand what the seeding does.

mdgeorge4153: Put another way, storing objects locally that are identical to what graphQL would return is just a performance optimization, which would be better handled at a lower layer of the stack.

crates/sui-forking/poc.md
mdgeorge4153: I was wondering about this too. Would there be any complexity introduced by extending gRPC this way?

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+
+The command accepts a checkpoint to fork from. This must not larger than the latest known checkpoint the RPC is aware of. It will error if the user requests a checkpoint that is not available.
+
+- `--checkpoint <number>`: The checkpoint to fork from
+- `--network <network>`: Network to fork from: `mainnet (default)`, `testnet`, `devnet`, or a custom one (`-network <CUSTOM-GRAPHQL-ENDPOINT>` ). The latter is useful for “forking” from a custom local network  / private network. It requires to have a GraphQL service running and a fullnode as well.
+- `--data-dir <path>`: The directory to persist the network state (checkpoints, epochs, transactions, objects, etc). This allows the user to maintain the state of the forked network across restarts.
+
+**The startup flow**
+- Initialize store layer (forking-data-store)
+- Fetch the latest checkpoint (or the checkpoint specified by the user)
```
mdgeorge4153, line 139: Related to my comment above, it seems like if you're always reading on demand, I'm not sure why you need to fetch the checkpoint here (other than to check that it exists)

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,261 @@
...
+sui-forking faucet --address <address> --amount <amount> # Max is 10M SUI
+```
+
+**Advance Checkpoint**
+
+```bash
+sui-forking advance-checkpoint
+```
+
+Advances the checkpoint of the local network by 1.
```
mdgeorge4153, line 152: Maybe an ignorant question: do you also need an API for starting a new epoch?

crates/sui-forking/poc.md
amnn: Not really, no, gRPC already separates methods into services and has service discovery etc, for this reason.

crates/sui-forking/poc.md
amnn: I think at some previous point we talked about something like "auto-crank by default but offer a flag to opt-out"

crates/sui-forking/poc.md
amnn: Planned follow-up yes (certainly out of scope for initial version)

crates/sui-forking/poc.md
amnn: Seeding is to initialize the indices for owned object queries, the discussion yesterday was about how point lookups were handled through the various data layers.

crates/sui-forking/poc.md
amnn: `advance-epoch` was part of the initial proposal, but it comes with some annoying complications that got it punted for a later version.

crates/sui-forking/poc.md
stefan-mysten: Got it, I was looking yesterday for this and missed the `query` part.

crates/sui-forking/poc.md
stefan-mysten: Yep, you're right. Though we probably want to keep a map from objects to owner, if we expect that this file might be also consumed by humans. The case I am thinking of is that someone might want to check if the address they're thinking of was part of this automatically generated seed file.

crates/sui-forking/poc.md
stefan-mysten: Yep, forgot to rectify this after our discussion. gRPC it is!

crates/sui-forking/poc.md
stefan-mysten: The coins come from a dummy genesis at startup time.

crates/sui-forking/poc.md
stefan-mysten: We can offer a flag to opt out, but not sure that wouldn't break transaction execution (at least for the usual workflow with the CLI it would). For POC I would prefer to keep it simple and automatically create a checkpoint.

stefan-mysten: > doesn't auto-crank but this says it does.

In the case of transaction execution it does. Otherwise, when the network is running it does not automatically create checkpoints nor epochs.

crates/sui-forking/poc.md
stefan-mysten: > initialize the indices

Not sure what you mean by initialize the indices. My understanding is that we need to fetch and store those objects on filesystem.

crates/sui-forking/poc.md
stefan-mysten: When you start the `simulacrum`, you need to pass to it a custom state instead of a new "genesis". That custom state is composed of:
- a `VerifiedCheckpoint` - so we actually need to fetch it. We could maybe "lazily" initialize the simulacrum, but don't see the advantage of doing it.
- the `SystemState` at that checkpoint (or right after that checkpoint). This is a bit trickier because it should also handle the case when you fork at the last checkpoint in an epoch.

crates/sui-forking/poc.md
stefan-mysten: Good question, let me add it back and mention that it will not be part of POC. I will add a section for various things that are planned for post POC.

crates/sui-forking/poc.md
stefan-mysten: I added it to a post POC section.

crates/sui-forking/poc.md
stefan-mysten: Here's my proposal:
```
"network": "testnet",
"network": "mainnet",
"network": "devnet-CHAIN_ID_DIGEST"
"network": "custom-CHAIN_ID_DIGEST"
```

crates/sui-forking/poc.md
amnn: It would be fine to include the addresses as an informational thing, although I don't know how much we want to encourage people interacting with the file directly.

crates/sui-forking/poc.md
amnn: Okay, there is something funky going on there, (which seems related to our previous discussions about what 'genesis' means for a forked network), but we can broach that later.

amnn: I would potentially leave that out of scope for now to avoid having to broach it for the POC.

crates/sui-forking/poc.md
amnn: No, you technically do not. You just need to seed an index of "who owns what" (owner -> object id + version (+ digest, optionally))

crates/sui-forking/poc.md
stefan-mysten: Sure, we can leave it out for POC.

crates/sui-forking/poc.md
stefan-mysten: Technically we don't, but why do we want to keep different ways of handling objects? For a transaction, input objects will be fetched and stored, but not for owned objects?

crates/sui-forking/poc.md
stefan-mysten: The use case I was thinking about is:
- have two seed files for same network, same checkpoint, but different addresses.
- you get generated two seed files.

Now you want to reuse a particular one, but you'd manually have to check to see which one has the right objects (if there's no object -> address mapping)

crates/sui-forking/poc.md
amnn: For a transaction, you have resolved inputs, and you are fetching their contents. For an owned object query, you have an owner, and you need to resolve their objects. You do not necessarily need to resolve their object contents, and in fact it would be better (more consistent) not to do that, but to factor this into two steps (resolve object refs, and then rely on the same mechanism used by execution to fetch resolved objects to fetch the object contents if you need them).

crates/sui-forking/poc.md
```diff
@@ -0,0 +1,175 @@
...
+### gRPC Interfaces
+
+Interacting with the forking tool is similar to interacting with a real network, and it’s made possible through gRPC. There are four main interfaces that the gRPC layer needs to implement:
+
+**Ledger Service**
+- this provides APIs for requesting & providing basic data (objects, epoch, checkpoint, transaction, etc)
+- provides a service info API that is used for fetching chain id, timestamp, epoch, highest available checkpoint, version.
+
+**State Service** 
+- this provides APIs for requesting and providing data related to balances, owned objects, dynamic fields, or coin information. In the context of the forking tool, this is used as a way to access live object data through the owned objects API.
```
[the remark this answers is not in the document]
stefan-mysten: We will deal with this later.

crates/sui-forking/poc.md
stefan-mysten: We punt this for a bit later when we will deal with seeds.

PR #26007, opened 2026-03-26
sui-indexer-alt: make Indexer field behavior consistent

evan-wall-mysten:
## Description 

1. Renames `default_next_checkpoint` to `first_checkpoint` for consistency with `last_checkpoint`. This helps to centralize the `next_checkpoint` calculation in `add_pipeline` where the network tip fallback logic will go.
2. Renames `first_ingestion_checkpoint` to `next_checkpoint` for consistency with `next_sequential_checkpoint`.
3. Use existing `IngestionService` range bounds handling instead of defaulting `last_checkpoint` to `u64::max`.

## Test plan 

Added additional assertions to exisiting unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/lib.rs
+++ b/crates/sui-indexer-alt-framework/src/lib.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::collections::BTreeSet;
+use std::ops::Bound;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -119,17 +120,21 @@ pub struct Indexer<S: Store> {
     /// Service for downloading and disseminating checkpoint data.
     ingestion_service: IngestionService,
 
-    /// The next checkpoint for a pipeline without a committer watermark to start processing from,
-    /// which will be 0 by default. Pipelines with existing watermarks will ignore this setting and
-    /// always resume from their committer watermark + 1.
-    ///
-    /// Setting this value indirectly affects ingestion, as the checkpoint to start ingesting from
-    /// is the minimum across all pipelines' next checkpoints.
-    default_next_checkpoint: u64,
+    /// Optional override of the checkpoint lowerbound. When set, pipelines without a committer
+    /// watermark will start processing at this checkpoint.
+    first_checkpoint: Option<u64>,
 
     /// Optional override of the checkpoint upperbound. When set, the indexer will stop ingestion at
     /// this checkpoint.
     last_checkpoint: Option<u64>,
+
+    /// The minimum `next_checkpoint` across all pipelines. This is the checkpoint for the indexer
+    /// to start ingesting from.
+    next_checkpoint: u64,
+
+    /// The minimum `next_checkpoint` across all sequential pipelines. This is used to initialize
+    /// the regulator to prevent ingestion from running too far ahead of sequential pipelines.
+    next_sequential_checkpoint: Option<u64>,
 
     /// An optional task name for this indexer. When set, pipelines will record watermarks using the
     /// delimiter defined on the store. This allows the same pipelines to run under multiple
@@ -152,15 +157,6 @@ pub struct Indexer<S: Store> {
     /// Pipelines that have already been registered with the indexer. Used to make sure a pipeline
     /// with the same name isn't added twice.
     added_pipelines: BTreeSet<&'static str>,
-
-    /// The checkpoint for the indexer to start ingesting from. This is derived from the committer
-    /// watermarks of pipelines added to the indexer. Pipelines without watermarks default to 0,
-    /// unless overridden by [Self::default_next_checkpoint].
-    first_ingestion_checkpoint: u64,
-
-    /// The minimum next_checkpoint across all sequential pipelines. This is used to initialize
-    /// the regulator to prevent ingestion from running too far ahead of sequential pipelines.
-    next_sequential_checkpoint: Option<u64>,
 
     /// The service handles for every pipeline, used to manage lifetimes and graceful shutdown.
     pipelines: Vec<Service>,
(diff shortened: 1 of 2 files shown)


crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -376,23 +375,26 @@ impl<S: Store> Indexer<S> {
...
         let InitWatermark {
             checkpoint_hi_inclusive,
             reader_lo,
         } = conn
             .init_watermark(
                 &pipeline_task,
                 InitWatermark {
-                    checkpoint_hi_inclusive: self.default_next_checkpoint.checked_sub(1),
-                    reader_lo: self.default_next_checkpoint,
+                    checkpoint_hi_inclusive: next_checkpoint.checked_sub(1),
```
wlmyng: ngl even though im familiar with this code it always trips me up slightly, and i have to recall that when next_checkpoint is 0, checkpoint_hi_inclusive is None, reader_lo is 0, and next_checkpoint, seeing None, sets to 0

crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -376,23 +375,26 @@ impl<S: Store> Indexer<S> {
         let pipeline_task =
             pipeline_task::<S>(P::NAME, self.task.as_ref().map(|t| t.task.as_str()))?;
 
+        // Pipeline will initialize `next_checkpoint` to this value if no watermark record exists.
```
wlmyng: i think this comment is more accurate down below at L397. At L379 this is just the `next_checkpoint` that this pipeline will process

crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -119,18 +120,22 @@ pub struct Indexer<S: Store> {
...
-    /// The next checkpoint for a pipeline without a committer watermark to start processing from,
-    /// which will be 0 by default. Pipelines with existing watermarks will ignore this setting and
-    /// always resume from their committer watermark + 1.
-    ///
-    /// Setting this value indirectly affects ingestion, as the checkpoint to start ingesting from
-    /// is the minimum across all pipelines' next checkpoints.
-    default_next_checkpoint: u64,
+    /// Optional override of the checkpoint lowerbound. When set, the indexer will start ingestion at
+    /// this checkpoint.
+    first_checkpoint: Option<u64>,
```
wlmyng:
```suggestion
    /// Optional override of the checkpoint lowerbound. When set, pipelines without a committer watermark will start processing from this checkpoint.
    first_checkpoint: Option<u64>,
```

`first_checkpoint` influences but does not force indexer to ingest at this checkpoint. For example, could have an existing pipeline w/ committer at 1, `--first-checkpoint` is 5, we'll start at 1

crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -119,18 +120,22 @@ pub struct Indexer<S: Store> {
...
 
     /// Optional override of the checkpoint upperbound. When set, the indexer will stop ingestion at
     /// this checkpoint.
     last_checkpoint: Option<u64>,
 
+    /// The minimum `next_checkpoint` across all pipelines. This is the checkpoint for the indexer
```
wlmyng, line 131: can we include a little bit on how this `next_checkpoint` is determined? or do you think that's extraneous

wlmyng (commented):
looks reasonable to me

crates/sui-indexer-alt-framework/src/lib.rs
evan-wall-mysten: I changed this a little bit to make it more clear that the initial `next_checkpoint` is just a proposal and the real value is not known until `init_watermark` is called.

crates/sui-indexer-alt-framework/src/lib.rs
evan-wall-mysten: Changed but used `at` instead of `from` to indicate it is inclusive like `last_checkpoint`.

crates/sui-indexer-alt-framework/src/lib.rs
evan-wall-mysten: I think it will become out of sync with the actual calculation because that is in a different location.

crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -376,23 +375,26 @@ impl<S: Store> Indexer<S> {
         let pipeline_task =
             pipeline_task::<S>(P::NAME, self.task.as_ref().map(|t| t.task.as_str()))?;
 
+        // Create a new record based on `proposed_next_checkpoint` if one does not exist.
+        // Otherwise, use the existing record and disregard the proposed value.
+        let proposed_next_checkpoint = self.first_checkpoint.unwrap_or(0);
```
amnn, line 380: For my uunderstanding, this is the part that will change later, so that the it starts at `latest_checkpoint - retention` if `first_checkpoint` is not set?

evan-wall-mysten: Correct, the linter forced me to use `.unwrap_or(0)`, but it will be something like 
```
let proposed_next_checkpoint = if let Some(first_checkpoint) = self.first_checkpoint {
    first_checkpoint
} else if let Some(retention) = retention {
    self.latest_checkpoint.saturating_sub(retention)
} else {
    0
};
```

crates/sui-indexer-alt-framework/src/lib.rs
evan-wall-mysten: I think the long term fix is to make `checkpoint_hi_inclusive` exclusive, but making that change requires some planning because of the schema change involved.

PR #26008, opened 2026-03-26
Restrict unused inputs in gasless transactions

mystenmark:
## Summary
- Add protocol config value `gasless_max_unused_inputs` to limit how many unused inputs a gasless PTB can have
- Implement validation in `validate_gasless_transaction()` to count unused inputs and reject transactions exceeding the limit
- When the config value is `None` (not set), there is no limit (backwards compatible with current behavior)

## Test plan
- [x] Unit tests added: `test_too_many_unused_inputs_fails`, `test_unused_inputs_within_limit_succeeds`, `test_no_unused_input_limit_allows_any`
- [x] All existing gasless validation tests pass (15 tests total)

### DIFF
--- a/crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
+++ b/crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
@@ -652,11 +652,14 @@ async fn test_combined_ab_and_coins_needed() {
             vec![(ab_amount, limited_sender)],
         )
         .build();
-    let (_, effects) = test_env
+    let (digest, effects) = test_env
         .exec_tx_directly(fund_ab_tx)
         .await
         .expect("Fund AB should succeed");
     assert!(effects.status().is_ok());
+    // Wait for the settlement transactions to commit so that the accumulator
+    // reflects limited_sender's AB before gas selection queries it.
+    test_env.cluster.wait_for_tx_settlement(&[digest]).await;
 
     // Now limited_sender has:
     // - Coin: 10 SUI


crates/sui-types/src/transaction.rs
```diff
@@ -996,8 +996,28 @@ impl ProgrammableTransaction {
         for command in &self.commands {
             command.validate_gasless_transaction(&allowed_token_types)?;
         }
+
+        // Check that the number of unused inputs doesn't exceed the limit
+        if let Some(max_unused) = config.gasless_max_unused_inputs_as_option() {
```
mystenmark: rather than using as_option just add a method that returns the default value of `usize::max` when the config is unset

crates/sui-types/src/transaction.rs
```diff
@@ -996,8 +996,28 @@ impl ProgrammableTransaction {
...
+            );
+        }
+
         Ok(())
     }
+
+    fn count_unused_inputs(&self) -> u64 {
+        let num_inputs = self.inputs.len() as u16;
+        (0..num_inputs)
+            .filter(|&i| !self.commands.iter().any(|cmd| cmd.is_input_arg_used(i)))
```
mystenmark: this looks quadratic to me! big no-no!

crates/sui-types/src/transaction.rs
```diff
@@ -996,6 +996,66 @@ impl ProgrammableTransaction {
...
+        Ok(())
+    }
+
+    fn validate_gasless_unused_inputs(&self, config: &ProtocolConfig) -> UserInputResult {
+        // Collect all used input indices in O(commands * args_per_command)
+        let used_inputs: std::collections::BTreeSet<u16> = self
+            .commands
+            .iter()
+            .flat_map(|cmd| cmd.input_arguments())
+            .collect();
```
mystenmark: since indices are not sparse we should use a `Vec<bool>` instead of a set container

crates/sui-types/src/transaction.rs
```diff
@@ -1421,6 +1481,25 @@ impl Command {
...
     }
+
+    fn input_arguments(&self) -> impl Iterator<Item = u16> + '_ {
+        self.arguments().filter_map(|arg| match arg {
+            Argument::Input(i) => Some(*i),
+            _ => None,
+        })
+    }
+
+    fn arguments(&self) -> Box<dyn Iterator<Item = &Argument> + '_> {
```
mystenmark: instead of `Box<dyn ...>` we should be able to use `impl Iterator` and `Either::{Right,Left}` in the body

crates/sui-types/src/transaction.rs
```diff
@@ -996,6 +996,72 @@ impl ProgrammableTransaction {
...
+    fn validate_gasless_unused_inputs(&self, config: &ProtocolConfig) -> UserInputResult {
+        let mut used_inputs = vec![false; self.inputs.len()];
+        for idx in self.commands.iter().flat_map(|cmd| cmd.input_arguments()) {
+            if let Some(slot) = used_inputs.get_mut(idx as usize) {
+                *slot = true;
+            }
+        }
+
+        let max_unused_pure = config
+            .gasless_max_unused_inputs_as_option()
```
mystenmark: add a method `gasless_max_unused_inputs` to ProtocolConfig.

the method should internally apply the default of usize::max

same for the other one

crates/sui-types/src/transaction.rs
```diff
@@ -1421,6 +1487,29 @@ impl Command {
...
+            Command::MoveCall(c) => Either::Left(Either::Left(c.arguments.iter())),
+            Command::TransferObjects(args, arg)
+            | Command::MergeCoins(arg, args)
+            | Command::SplitCoins(arg, args) => {
+                Either::Left(Either::Right(args.iter().chain(once(arg))))
+            }
+            Command::MakeMoveVec(_, args) => Either::Right(Either::Left(args.iter())),
+            Command::Upgrade(_, _, _, arg) => Either::Right(Either::Right(Either::Left(once(arg)))),
+            Command::Publish(_, _) => {
+                Either::Right(Either::Right(Either::Right(std::iter::empty())))
```
mystenmark: All arg types are either `Vec<Argument>` or `Argument` so we only need Either::Left and Either::Right, not all this nesting

aschran (approved):

crates/sui-types/src/transaction.rs
```diff
@@ -1008,6 +1008,66 @@ impl ProgrammableTransaction {
...
+        let mut unused_pure_count = 0usize;
+
+        for (i, input) in self.inputs.iter().enumerate() {
+            if used_inputs[i] {
+                continue;
+            }
+            // This input is unused
+            match input {
+                CallArg::Pure(bytes) => {
+                    unused_pure_count += 1;
```
tnowacki: How is this unused? What if it is... used? That is as an argument for some command

crates/sui-types/src/transaction.rs
```diff
@@ -1008,6 +1008,66 @@ impl ProgrammableTransaction {
...
+                        "Gasless transactions do not allow unused Object inputs (input {})",
+                        i
+                    )));
+                }
+                CallArg::FundsWithdrawal(_) => {
+                    return Err(UserInputError::Unsupported(format!(
+                        "Gasless transactions do not allow unused FundsWithdrawal inputs (input {})",
+                        i
+                    )));
+                }
```
tnowacki: We should just ignore these for this function. This is verified in `validate_gasless_transaction`

crates/sui-types/src/transaction.rs
```diff
@@ -5179,3 +5259,383 @@ impl TransactionKey {
         }
     }
 }
+
+#[cfg(test)]
+mod gasless_validation_tests {
```
alex-mysten: I just wrote some nice transactional tests; these could probably supplement those

crates/sui-types/src/transaction.rs
```diff
@@ -1438,6 +1498,26 @@ impl Command {
...
+    fn arguments(&self) -> impl Iterator<Item = &Argument> + '_ {
+        let (args, single): (&[Argument], Option<&Argument>) = match self {
+            Command::MoveCall(c) => (&c.arguments, None),
+            Command::TransferObjects(args, arg)
+            | Command::MergeCoins(arg, args)
+            | Command::SplitCoins(arg, args) => (args, Some(arg)),
+            Command::MakeMoveVec(_, args) => (args, None),
+            Command::Upgrade(_, _, _, arg) => (&[], Some(arg)),
+            Command::Publish(_, _) => (&[], None),
+        };
```
alex-mysten, line 1521: this looks very similar to is_argument_used. deduplication opportunity?

crates/sui-types/src/transaction.rs
```diff
@@ -1008,6 +1008,66 @@ impl ProgrammableTransaction {
...
+        for (i, input) in self.inputs.iter().enumerate() {
+            if used_inputs[i] {
+                continue;
+            }
+            // This input is unused
+            match input {
+                CallArg::Pure(bytes) => {
+                    unused_pure_count += 1;
+                    fp_ensure!(
+                        bytes.len() <= max_unused_bytes,
```
alex-mysten: Should we validate the size of used inputs?

crates/sui-types/src/transaction.rs
mystenmark: See above where we check `used_inputs[i]`

crates/sui-types/src/transaction.rs
mystenmark: yes, good point!

crates/sui-types/src/transaction.rs
mystenmark: yes, good idea

claude: we can implement is_argument_used in terms of this function

crates/sui-types/src/transaction.rs
mystenmark: will do!

crates/sui-types/src/transaction.rs
mystenmark: claude: verify that we have a test cases that for unused objects/fundswithdrawals and then delete these match arms

mystenmark: as far as i can tell, this is not true, no other test fails if i remove these match arms

crates/sui-transactional-test-runner/src/test_adapter.rs
```diff
@@ -317,6 +319,12 @@ impl AdapterInitConfig {
         if enable_gasless {
             protocol_config.enable_gasless_for_testing();
         }
+        if let Some(max_bytes) = gasless_max_pure_input_bytes {
+            protocol_config.set_gasless_max_pure_input_bytes_for_testing(max_bytes);
```
tnowacki, line 327: Nit might want to assert that gasless is enabled for these two flags. Not that it will matter for very long though

crates/sui-types/src/transaction.rs
```diff
@@ -1440,17 +1499,27 @@ impl Command {
...
+    }
+
+    fn input_arguments(&self) -> impl Iterator<Item = u16> + '_ {
+        self.arguments().filter_map(|arg| match arg {
+            Argument::Input(i) => Some(*i),
+            _ => None,
+        })
+    }
+
+    fn arguments(&self) -> impl Iterator<Item = &Argument> + '_ {
```
tnowacki, line 1512: 🎉

crates/sui-types/src/transaction.rs
```diff
@@ -1440,17 +1499,27 @@ impl Command {
...
+        let (args, single): (&[Argument], Option<&Argument>) = match self {
```
tnowacki, line 1513: It is a bit strange, but at least you don't have to box the iterator this way or use either?

crates/sui-types/src/transaction.rs
```diff
@@ -1022,6 +1021,66 @@ impl ProgrammableTransaction {
...
+                        i
+                    )));
+                }
+                CallArg::FundsWithdrawal(_) if !is_used => {
+                    return Err(UserInputError::Unsupported(format!(
+                        "Gasless transactions do not allow unused FundsWithdrawal inputs (input {})",
+                        i
+                    )));
+                }
+                _ => {}
```
tnowacki: We aren't adding new CallArgs very often, but it might be worth considering pushing each case into the branch with an `fp_ensure` so that this match is exhaustive

crates/sui-types/src/transaction.rs
mystenmark: claude came up with this. I suggested Either, it went bonkers (used a different left/right/left chain for every branch), i told it that was dumb, and then it did this. which is admittedly clever.

crates/sui-transactional-test-runner/src/test_adapter.rs
mystenmark: yeah, makes sense

PR #26009, opened 2026-03-26
[forking 2/n] stand up forking server with mininum data fetching

stefan-mysten:
## Description 

This PR brings up the forking server with minimum data fetching.
- Add the core modules for the sui-forking crate: network selection (Network), GraphQL data fetching              
  (GraphQLQueryClient), persistent store (ServiceStore), and server startup (start_server)                          
- Introduce ForkDataProvider trait to abstract network data fetching (checkpoint + protocol version), decoupling
  startup logic from the concrete GraphQL client                                                                    
- Make setup_simulacrum and start_server generic over SimulatorStore, so tests can swap in InMemoryStore while production uses ServiceStore                                                                                      
- Add an integration test that boots the server with mocked dependencies (genesis checkpoint + InMemoryStore) and verifies /health and /status endpoints over HTTP            

Stack:
https://github.com/MystenLabs/sui/pull/26009
https://github.com/MystenLabs/sui/pull/26004

## Test plan 

The tests are starting the network from a new genesis using an InMemoryStore (from simulacrum) at checkpoint 0 and with the MAX protocol version existing in the codebase.
```
cargo nextest run
cargo test -- --nocapture # for some output
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-forking/src/graphql.rs
+++ b/crates/sui-forking/src/graphql.rs
@@ -1,21 +1,22 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use anyhow::{Result, bail};
-use sui_graphql::{CheckpointResponse, Client};
-use sui_types::crypto::{AggregateAuthoritySignature, AuthorityQuorumSignInfo};
+use anyhow::Result;
+use anyhow::bail;
+use tracing::info;
+
+use sui_graphql::CheckpointResponse;
+use sui_graphql::Client;
+use sui_types::crypto::AggregateAuthoritySignature;
+use sui_types::crypto::AuthorityQuorumSignInfo;
 use sui_types::message_envelope::Envelope;
 use sui_types::messages_checkpoint::VerifiedCheckpoint;
-use tracing::info;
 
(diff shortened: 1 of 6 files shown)


crates/sui-forking/src/graphql.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) Mysten Labs, Inc.
```
jordanjennings-mysten, line 1: Is this more or less a proxy? graphql_proxy.rs?

crates/sui-forking/Cargo.toml
```diff
@@ -8,10 +8,24 @@ edition = "2024"
 
 [dependencies]
```
jordanjennings-mysten, line 9: I see you're doing mocks might be worth a look at https://crates.io/crates/mockall

crates/sui-forking/src/graphql.rs
stefan-mysten: Yes, probably should call it `graphql_client` or `proxy` as you suggested.

PR #26011, opened 2026-03-26
Fix coin reservation panic during checkpoint replay with MVCC

mystenmark:
## Summary
- Use MVCC to read accumulator objects at correct version before settlement transactions modify them

## Test plan
- [x] `MSIM_TEST_SEED=1 cargo simtest -p sui-benchmark -E 'test(test_composite_workload)'` passes
- [x] Seed search with 10 random seeds passes

### DIFF
--- a/crates/sui-e2e-tests/tests/address_balance_tests.rs
+++ b/crates/sui-e2e-tests/tests/address_balance_tests.rs
@@ -22,7 +22,7 @@ use sui_types::{
     SUI_FRAMEWORK_PACKAGE_ID, TypeTag,
     accumulator_root::AccumulatorValue,
     balance::Balance,
-    base_types::{ObjectID, ObjectRef, SuiAddress, dbg_addr},
+    base_types::{ObjectID, ObjectRef, SequenceNumber, SuiAddress, dbg_addr},
     coin_reservation::{CoinReservationResolverTrait, ParsedObjectRefWithdrawal},
     digests::{ChainIdentifier, CheckpointDigest},
     effects::{InputConsensusObject, TransactionEffectsAPI},
@@ -50,6 +50,7 @@ impl CoinReservationResolverTrait for NoOpResolver {
         &self,
         _sender: SuiAddress,
         _coin_reservation: ParsedObjectRefWithdrawal,
+        _accumulator_version: Option<SequenceNumber>,
     ) -> UserInputResult<FundsWithdrawalArg> {
         panic!("Not used in these tests")
     }
--- a/crates/sui-types/src/coin_reservation.rs
+++ b/crates/sui-types/src/coin_reservation.rs
@@ -258,8 +258,10 @@ impl CoinReservationResolver {
         coin_reservation: ParsedObjectRefWithdrawal,
         accumulator_version: Option<SequenceNumber>,
     ) -> UserInputResult<FundsWithdrawalArg> {
-        let (owner, type_tag) = self
-            .get_owner_and_type_for_object(coin_reservation.unmasked_object_id, accumulator_version)?;
+        let (owner, type_tag) = self.get_owner_and_type_for_object(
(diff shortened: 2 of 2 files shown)


crates/sui-benchmark/src/workloads/composite/operations.rs
```diff
@@ -83,6 +90,8 @@ pub struct OperationResources {
     pub test_coin_cap: Option<(ObjectID, SequenceNumber)>,
     pub test_coin_type: Option<TypeTag>,
     pub immutable_object: Option<ObjectRef>,
+    pub chain_identifier: Option<ChainIdentifier>,
+    pub current_epoch: Option<EpochId>,
```
mystenmark, line 94: we can already get the current epoch from the system state, you can see how to do this in make_transaction_batch

crates/sui-core/src/accumulators/coin_reservations.rs
```diff
@@ -33,19 +33,28 @@ impl CachingCoinReservationResolver {
     fn get_owner_and_type_cached(
         &self,
         object_id: ObjectID,
+        accumulator_version: Option<SequenceNumber>,
     ) -> UserInputResult<(SuiAddress, TypeTag)> {
-        self.cache.get_with(object_id, || {
-            self.inner.get_owner_and_type_for_object(object_id)
-        })
+        // When a specific version is requested (MVCC), bypass the cache since the cached
+        // value may be at a different version.
```
mystenmark: this is not necessary - the only information returned by this method are the owner and type tag, which cannot change over time. the cache is always coherent

crates/sui-core/src/accumulators/transaction_rewriting.rs
```diff
@@ -12,11 +12,16 @@ use sui_types::transaction::{CallArg, ObjectArg, ProgrammableTransaction, Transa
 /// Returns `Some(rewritten_inputs)` if any inputs were rewritten, where each bool indicates whether
 /// the corresponding input was converted from a coin reservation. Returns `None` if nothing
 /// was rewritten.
+///
+/// `accumulator_version` is the version of the accumulator root object to use for MVCC lookup.
+/// This is required during checkpoint replay to read the accumulator state at the correct version,
+/// before any settlement transactions have modified it.
```
mystenmark: comment not necessary

crates/sui-core/src/accumulators/transaction_rewriting.rs
```diff
@@ -54,8 +61,10 @@ fn rewrite_programmable_transaction_for_coin_reservations(
             // 2. The scheduler reserves funds before allowing the transaction to execute. If the
             //    accumulator were deleted (balance dropped to 0), the reservation would fail and
             //    the transaction would not enter execution.
+            // 3. During checkpoint replay with MVCC, we read the accumulator at the version before
+            //    any settlement transactions have modified it.
```
mystenmark: comment not necessary

crates/sui-core/src/authority.rs
```diff
@@ -1917,19 +1917,27 @@ impl AuthorityState {
...
         Result<(), ExecutionError>,
     ) {
-        let rewritten_inputs = rewrite_transaction_for_coin_reservations(
-            self.chain_identifier,
-            &*self.coin_reservation_resolver,
-            sender,
-            &mut kind,
-        );
+        // Skip rewriting if execution_params already indicates an error - the transaction will fail
+        // anyway, and trying to rewrite could fail if the accumulator was deleted.
```
mystenmark: the comment is false and the code changes it documents should not be needed

mystenmark:
RE: current_epoch comment - The epoch is already obtained from `system_state.epoch` in `make_transaction_batch` and passed through `resolve_resources_for_op`. Could you clarify if you're suggesting a different approach?

aschran (approved):

PR #26012, opened 2026-03-26
Add CoinReservationWithdraw to composite workload and fix MVCC panic

mystenmark:
## Summary
- Add `CoinReservationWithdraw` operation to the composite workload to test coin reservation transactions alongside other operations
- Add reachability assertion to ensure coin reservation withdrawals are processed during simtest
- Fix panic during checkpoint replay when settlement transactions execute before user transactions using coin reservations

## Details

The MVCC fix addresses a race condition during checkpoint replay where:
1. User transaction uses a coin reservation (encoded ObjectRef pointing to accumulator field)
2. Settlement/Barrier transaction deletes or modifies the accumulator field
3. During checkpoint replay, the settlement can execute first, causing the subsequent user transaction to panic when trying to resolve the coin reservation

The fix passes `accumulator_version` through the resolver chain, allowing MVCC reads at the correct version. When a specific version is requested, the cache is bypassed to ensure we read the correct historical state.

## Test plan
- [x] Existing tests pass
- [x] New composite workload test exercises coin reservation withdrawals
- [x] Simtest confirms variety of transaction outcomes for CoinReservationWithdraw

### DIFF
--- a/crates/sui-benchmark/tests/simtest.rs
+++ b/crates/sui-benchmark/tests/simtest.rs
@@ -1592,7 +1592,7 @@ mod test {
         .with_probability(AddressBalanceOverdraw::NAME, 0.3)
         .with_probability(AccumulatorBalanceRead::NAME, 0.3)
         .with_probability(AuthenticatedEventEmit::NAME, 0.1)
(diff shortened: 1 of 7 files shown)


mystenmark:
Duplicate - see #26011

PR #26017, opened 2026-03-27
Improve &str to Digest error message

evan-wall-mysten:
## Description 

Improves the error message when a `Digest` cannot be created from a `&str` to help with debugging and centralizes the logic to do so.

## Test plan 

Added unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-types/src/digests.rs
+++ b/crates/sui-types/src/digests.rs
@@ -371,13 +371,7 @@ impl std::str::FromStr for CheckpointDigest {
     type Err = anyhow::Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let mut result = [0; 32];
-        let buffer = Base58::decode(s).map_err(|e| anyhow::anyhow!(e))?;
-        if buffer.len() != 32 {
-            return Err(anyhow::anyhow!("Invalid digest length. Expected 32 bytes"));
-        }
-        result.copy_from_slice(&buffer);
-        Ok(CheckpointDigest::new(result))
+        Ok(Self::new(digest_from_base58(s)?))
     }
 }
 
@@ -464,13 +458,7 @@ impl std::str::FromStr for CheckpointContentsDigest {
     type Err = anyhow::Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let mut result = [0; 32];
-        let buffer = Base58::decode(s).map_err(|e| anyhow::anyhow!(e))?;
-        if buffer.len() != 32 {
-            return Err(anyhow::anyhow!("Invalid digest length. Expected 32 bytes"));
-        }
-        result.copy_from_slice(&buffer);
-        Ok(CheckpointContentsDigest::new(result))
+        Ok(Self::new(digest_from_base58(s)?))
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/sui-types/src/digests.rs
```diff
@@ -1169,6 +1127,23 @@ impl fmt::Display for CheckpointArtifactsDigest {
...
+    let buffer = Base58::decode(s).map_err(|e| anyhow::anyhow!(e))?;
+    if buffer.len() < 32 {
+        return Err(anyhow::anyhow!(
+            "Invalid digest length. Expected 32-byte base58 string, but got {s}"
+        ));
+    } else if buffer.len() > 32 {
+        // truncate in case `s` is very long
+        return Err(anyhow::anyhow!(
+            "Invalid digest length. Expected 32-byte base58 string, but got {s} (truncated)"
+        ));
```
bmwill: s isn't truncated from what i can see?

evan-wall-mysten: Yeah my mistake. I fixed that and added some tests then forgot to push it.

crates/sui-types/src/digests.rs
```diff
@@ -1169,9 +1127,26 @@ impl fmt::Display for CheckpointArtifactsDigest {
...
 
+fn digest_from_base58(s: &str) -> anyhow::Result<[u8; 32]> {
+    let mut result = [0; 32];
+    let buffer = Base58::decode(s).map_err(|e| anyhow::anyhow!(e))?;
+    let len = buffer.len();
+    if len < 32 {
+        return Err(anyhow::anyhow!(
+            "Invalid digest length. Expected 32-byte base58 string, but got [{s}] with length {len}"
+        ));
+    } else if len > 32 {
```
tpham-mysten, line 1138: nit: else if here is unnecessary

crates/sui-types/src/digests.rs
```diff
@@ -1169,9 +1127,26 @@ impl fmt::Display for CheckpointArtifactsDigest {
...
+        return Err(anyhow::anyhow!(
+            "Invalid digest length. Expected 32-byte base58 string, but got [{s}] (truncated) with length {len}"
```
tpham-mysten: There seems to be no truncation happening here. Do I miss anything?

crates/sui-types/src/digests.rs
```diff
@@ -1169,9 +1127,26 @@ impl fmt::Display for CheckpointArtifactsDigest {
     }
 }
 
+fn digest_from_base58(s: &str) -> anyhow::Result<[u8; 32]> {
+    let mut result = [0; 32];
+    let buffer = Base58::decode(s).map_err(|e| anyhow::anyhow!(e))?;
+    let len = buffer.len();
+    if len < 32 {
+        return Err(anyhow::anyhow!(
+            "Invalid digest length. Expected 32-byte base58 string, but got [{s}] with length {len}"
```
tpham-mysten: nit: just a bit confusing regarding "32-byte base58 string",

The 32-byte constraint is on the decoded byte bot the base58 string itself. Maybe 

> Expected base58 string decoding to 32 bytes

is more explicit?

crates/sui-types/src/digests.rs
evan-wall-mysten: Yeah I accidentally left it out then was getting some error pushing my changes last night but it should be here now.

crates/sui-types/src/digests.rs
evan-wall-mysten: Yeah I'll change it since the base58 string is actually 44 bytes I think.

crates/sui-types/src/digests.rs
evan-wall-mysten: It is for the "truncated" branch which I fixed

PR #26019, opened 2026-03-27
[graphql-stream] WebSocket subscription infrastructure [1/n]

tpham-mysten:
## Description 

Add WebSocket infrastructure for GraphQL streaming subscriptions.

What this PR does:
  - Adds a `StreamSubscription` type to the GraphQL schema with a placeholder checkpoints subscription (returns `FeatureUnavailable` until streaming is implemented)
  - Adds WebSocket endpoint at `/graphql/subscriptions` using `graphql-transport-ws protocol`
  - Adds `--checkpoint-stream-url` CLI flag (`SubscriptionArgs`) to enable subscriptions per instance
  - Adds `SubscriptionConfig` with configurable `broadcast_buffer` size (default: 256)
  - Adds `CheckpointStreamTask` and `ProcessedCheckpoint` stubs for follow-up PRs
  - Guards subscriptions at `on_connection_init` — instances without `--checkpoint-stream-url` reject WebSocket connections with "Subscriptions are not enabled on this instance"
  
What this PR does not do (which will arrive in follow-up PRs)
  - No actual checkpoint streaming — `CheckpointStreamTask` does not connect to gRPC yet
  - No subscription resolvers — checkpoints returns `FeatureUnavailable`
  - No `process_checkpoint()` or BCS deserialization
  - No `Scope` changes for streamed data

Extension changes for WebSocket compatibility:
  - `query_limits`: skip `ContentLength` payload check for WebSocket (no HTTP body)
  - `rich`: skip Meter for subscriptions (resolve from streamed data, no DB queries)
  - `timeout`: skip per-request timeout for subscriptions (long-lived streams)

## Test plan 

- `cargo nextest run -p sui-indexer-alt-graphql)`
- cargo clippy
- Manually tested with local GraphQL server:
    - Without `--checkpoint-stream-url`: subscription in GraphiQL shows "Subscriptions are not enabled on this instance."
    - Without `--checkpoint-stream-url`: direct WebSocket client receives close code with the same message
    - Regular queries/mutations continue to work normally

---

## Stack

- #26019
- #26094
- #26117
- #26170
- #26194

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
+++ b/crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
@@ -1,8 +1,9 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::HashMap;
 use std::sync::Arc;
-use std::sync::Mutex;
+use std::sync::OnceLock;
 use std::time::Duration;
 
 use async_graphql::Response;
@@ -36,7 +37,9 @@ pub(crate) struct Timeout(Arc<TimeoutConfig>);
 
 struct TimeoutExt {
     config: Arc<TimeoutConfig>,
-    operation_type: Mutex<OperationType>,
+    /// Map from operation name to its type, populated once during parsing.
+    /// `None` key represents an anonymous (single) operation.
+    operation_types: OnceLock<HashMap<Option<String>, OperationType>>,
 }
 
 impl Timeout {
@@ -49,7 +52,7 @@ impl ExtensionFactory for Timeout {
     fn create(&self) -> Arc<dyn Extension> {
         Arc::new(TimeoutExt {
             config: self.0.clone(),
-            operation_type: Mutex::new(OperationType::Query),
+            operation_types: OnceLock::new(),
         })
     }
 }
@@ -65,9 +68,12 @@ impl Extension for TimeoutExt {
     ) -> ServerResult<ExecutableDocument> {
         let document = next.run(ctx, query, variables).await?;
 
-        if let Some((_, op)) = document.operations.iter().next() {
-            *self.operation_type.lock().unwrap() = op.node.ty;
-        }
+        let types: HashMap<_, _> = document
+            .operations
+            .iter()
+            .map(|(name, op)| (name.map(|n| n.to_string()), op.node.ty))
+            .collect();
+        let _ = self.operation_types.set(types);
 
         Ok(document)
     }
@@ -78,7 +84,12 @@ impl Extension for TimeoutExt {
         operation_name: Option<&str>,
         next: NextExecute<'_>,
     ) -> Response {
-        let operation_type = *self.operation_type.lock().unwrap();
+        let key = operation_name.map(|s| s.to_string());
+        let operation_type = self
+            .operation_types
+            .get()
+            .and_then(|types| types.get(&key).copied())
+            .unwrap_or(OperationType::Query);
 
         // Subscriptions are long-lived streams — a per-request timeout does not apply.
         if operation_type == OperationType::Subscription {
--- a/crates/sui-indexer-alt-graphql/src/api/types/available_range.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/available_range.rs
@@ -632,7 +632,12 @@ mod field_piplines_tests {
                 );
             }
         }
-        insta::assert_snapshot!(output);
+        let snapshot_name = if cfg!(feature = "staging") {
+            "registry_collect_pipelines_snapshot_staging"
+        } else {
+            "registry_collect_pipelines_snapshot"
+        };
+        insta::assert_snapshot!(snapshot_name, output);
     }
 
     fn formatted_output_str(
--- a/crates/sui-indexer-alt-graphql/src/api/types/snapshots/sui_indexer_alt_graphql__api__types__available_range__field_piplines_tests__registry_collect_pipelines_snapshot_staging.snap
+++ b/crates/sui-indexer-alt-graphql/src/api/types/snapshots/sui_indexer_alt_graphql__api__types__available_range__field_piplines_tests__registry_collect_pipelines_snapshot_staging.snap
@@ -1634,6 +1634,9 @@ SplitCoinsCommand.amounts
 StoreExecutionTimeObservationsTransaction._
   => {}
 
+Subscription.checkpoints
+  => {}
+
 Transaction.id
   => {}
 


crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -425,8 +445,46 @@ async fn graphql(
...
+) -> impl IntoResponse {
+    ws.protocols(ALL_WEBSOCKET_PROTOCOLS)
+        .on_upgrade(move |stream| {
+            let mut data = async_graphql::Data::default();
+            data.insert(Session::new(addr));
+
+            GraphQLWebSocket::new(stream, schema, protocol)
+                .with_data(data)
+                .on_connection_init(move |_| async move {
+                    if !subscriptions_enabled {
```
tpham-mysten, line 498: This logic is quite interesting, and it is related to a design topic: **How to guard subscriptions on non-subscription instances?**

For more context, We plan to have dedicated instances for GraphQL streaming for an ease of maintenance, monitoring, and scalability, so there would be GraphQL instances that do not support subscription. We then need to design a guarding solution, and it comes with a few questions to discuss:

1. Should the subscription schema be visible on non-subscription instances?

Hiding it keeps the schema accurate per-instance but could confuse developers who see subscriptions documented and available in one instance but not the others. Therefore, my preferred solution is to show it, and try to return a clear error like "Subscription is not available on this GraphQL instance", which is explicit and clear to clients.

2. How could we enable and disable GraphQL streamings?

We plan to just use `--checkpoint-stream-url` args, which is a raw checkpoints gRPC streaming that serves GraphQL streaming. The subscription is enabled when the flag is defined and disabled when unset.

3. How to display the subscription-unavailable error to clients?

There are two types of clients that connect to the subscription endpoint: `GraphiQL` and direct WebSocket clients. `GraphiQL` is the browser-based IDE that developers use for interactive testing, served at `GET /graphql` and is a critical part of the GraphQL developer experience. Direct WebSocket clients are programmatic clients that connect to `/graphql/subscriptions` to consume streaming data.

The challenge is that `GraphiQL` uses the browser's WebSocket API (via the [graphql-ws](https://github.com/enisdenjo/graphql-ws) JavaScript library). The WebSocket spec requires that browsers ["must not convey any failure information to scripts"](https://websockets.spec.whatwg.org/) when a connection fails, to prevent network probing attacks. As a result, when we reject a connection before the WebSocket upgrade completes, GraphiQL can only display [{isTrusted: true}](https://www.loom.com/i/5bef43793a9d4ff29711a72dc22e9550), a bare Event object with no error details. Direct WebSocket clients outside the browser are not affected and receive the error normally.

With this constraint, we explored the following options:

**Option 1: Reject before WebSocket upgrade with HTTP error**

Return a 503 with {"errors": [{"message": "Subscriptions are not enabled on this instance."}]} before completing the WebSocket handshake.
  - Pros: No WebSocket resource cost. Direct clients receive a clear JSON error.
  - Cons: GraphiQL displays only {isTrusted: true} due to the browser restriction above.

**Option 2: Don't set `subscription_endpoint` in GraphiQL**
Without a subscription endpoint configured, GraphiQL would send subscription queries as regular HTTP POST instead of WebSocket.
  - Pros: No WebSocket connection at all.
  - Cons: GraphiQL detects the subscription operation type before sending and throws its own [client-side error](https://loom.com/i/02f23c6a4c354fd09d4a49f2b307659e): "not properly configured for websocket subscriptions yet." Our server never receives the request, so we cannot return our own error message.

**Option 3: Reject in each subscription resolver**
The WebSocket upgrade succeeds and each resolver individually checks for the broadcaster.
  - Pros: Works for both `GraphiQL` and direct clients. Error is a proper GraphQL error.
  - Cons: Every subscription resolver must check and handle the missing broadcaster case, which is repetitive and error-prone as we add more subscriptions

**Option 4: Reject at on_connection_init (current approach)**
The WebSocket upgrade completes, then we immediately check whether streaming is enabled. If not, we reject with a clear message through the graphql-transport-ws protocol.
  - Pros: GraphiQL displays "Subscriptions are not enabled on this instance" [error](https://loom.com/i/e8bbe2384e0f47fbaec5b4d421f5159b). Direct WebSocket clients receive the same close code and reason string. Subscription resolvers can assume the broadcaster is always available, and no None checks needed.
  - Cons: Requires completing a WebSocket handshake before the rejection can be delivered, since the graphql-transport-ws protocol operates after the upgrade. However, the connection is closed immediately after `connection_init`, so the resource cost is minimal. This mainly matters for production non-subscription instances where GraphiQL is typically disabled (`--no-ide`), and the WebSocket handshake cost would only come from direct clients attempting to connect.

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use async_graphql::Context;
+use async_graphql::Subscription;
+
+use crate::api::types::checkpoint::Checkpoint;
+use crate::error::RpcError;
+
+#[derive(Default)]
+pub struct StreamSubscription;
```
amnn: nit: let's just call this type `Subscription`, like we call the top-level query type `Query`, and the top-level mutation type `Mutation`.

crates/sui-indexer-alt-graphql/src/extensions/query_limits/mod.rs
```diff
@@ -107,8 +107,10 @@ impl Extension for QueryLimitsCheckerExt {
         variables: &Variables,
         next: NextParseQuery<'_>,
     ) -> ServerResult<ExecutableDocument> {
-        let &ContentLength(length) = ctx.data_unchecked();
-        if length > self.limits.max_payload_size() as u64 {
+        // ContentLength is not available for WebSocket subscriptions (no HTTP body).
```
amnn, line 110: Is there something else that prevents clients from sending just very large requests when initially starting a subscription?

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -74,6 +76,14 @@ impl Extension for TimeoutExt {
             Ordering::Relaxed,
         );
 
+        self.is_subscription.store(
+            document
+                .operations
+                .iter()
+                .any(|(_, op)| op.node.ty == OperationType::Subscription),
+            Ordering::Relaxed,
+        );
```
amnn: Maybe we should just store the `OperationType`?

crates/sui-indexer-alt-graphql/src/snapshots/sui_indexer_alt_graphql__tests__schema.graphql.snap
```diff
@@ -4018,6 +4018,15 @@ type StoreExecutionTimeObservationsTransaction {
 	_: Boolean
 }
 
+type StreamSubscription {
```
amnn: If this is not yet something that people can use, I would suggest using the `staging` feature for this, and for us to finally do the work to ship the `staging` build to CI and devnet so that we can iterate on the schema without committing to something on a production network.

crates/sui-indexer-alt-graphql/src/config.rs
amnn, line 1: nit: file order

crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -425,8 +445,46 @@ async fn graphql(
...
+        GraphiQLSource::build()
+            .endpoint(GRAPHQL_PATH)
+            .subscription_endpoint(SUBSCRIPTION_PATH)
+            .finish(),
+    )
+}
+
+/// Whether subscriptions are enabled on this instance (i.e., `--checkpoint-stream-url` was set).
+#[derive(Clone, Copy)]
+struct SubscriptionsEnabled(bool);
```
amnn: file order

crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -367,9 +376,14 @@ pub async fn start_rpc(
...
 
+    let streaming_task = subscription_args
+        .checkpoint_stream_url
+        .map(|uri| task::streaming::CheckpointStreamTask::new(uri, &config.subscription));
+
     let rpc = rpc
-        .route("/graphql", post(graphql))
-        .route("/graphql/health", get(health::check))
+        .route(GRAPHQL_PATH, post(graphql))
+        .route(SUBSCRIPTION_PATH, get(graphql_ws))
```
amnn: Should this route only be enabled when subscriptions are enabled?

crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -64,12 +67,17 @@ use url::Url;
...
 use crate::api::query::Query;
+use crate::api::subscription::StreamSubscription;
 use crate::error::PanicHandler;
 use crate::extensions::logging::Logging;
 use crate::extensions::logging::Session;
 use crate::metrics::RpcMetrics;
 use crate::middleware::version::Version;
 
+const GRAPHQL_PATH: &str = "/graphql";
+const SUBSCRIPTION_PATH: &str = "/graphql/subscriptions";
```
amnn: From what I can tell, the usual convention is to keep the path the same, but route based on the protocol (`wss`/`ws` vs `https`/`http`).

crates/sui-indexer-alt-graphql/src/extensions/query_limits/mod.rs
amnn: Also, is it possible to handle this part of the request over HTTP, before upgrading to websocket for the actual subscription? Then you would have access to a content length to validate the initial request.

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Sounds good! Done

crates/sui-indexer-alt-graphql/src/extensions/query_limits/mod.rs
tpham-mysten: The query isn't available during the HTTP upgrade; it's a bodyless GET with only standard WebSocket headers. The GraphQL query arrives later as a WebSocket message (inside a subscribe frame), after the connection is already established.  I verified this too locally by logging the upgrade request headers.

Once the subscribe message arrives, our `input::check` (AST depth/nodes) and `output::check` (estimated response size) still run before execution, so query complexity is bounded. For byte-size limits, we can configure [max_message_size](https://docs.rs/tungstenite/latest/tungstenite/protocol/struct.WebSocketConfig.html#structfield.max_message_size) on the WebSocket to match our max_payload_size. Does that sound right to you?

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
tpham-mysten: Sounds good! Done

crates/sui-indexer-alt-graphql/src/snapshots/sui_indexer_alt_graphql__tests__schema.graphql.snap
tpham-mysten: Sounds good. My original plan was to guard the implementation to return `RpcError::FeatureUnavailable` for non-staging, but yeah guarding the schema is better. 

Done! the Subscription type is gated behind `#[cfg(feature = "staging")]`, falling back to `EmptySubscription` in production.

crates/sui-indexer-alt-graphql/src/config.rs
tpham-mysten: Done

crates/sui-indexer-alt-graphql/src/lib.rs
tpham-mysten: This is one of discussions items in this comment: https://github.com/MystenLabs/sui/pull/26019/changes#r3002027220

crates/sui-indexer-alt-graphql/src/lib.rs
tpham-mysten: Sounds good, and you're right (same path is the standard convention). Thanks for pointing it out. I was not putting much thoughts into this before.

The only tricky thing is GraphiQL is also using `GET` on `/graphql`, but we can differentiate by checking for the WebSocket upgrade. Both are handled in the new `graphql_get` handler:

```
match (ws, protocol) {Ok.
        (Some(ws), Some(protocol)) => {
            handle_ws(ws, protocol, schema, addr, subscriptions_enabled).into_response()
        }
        _ if ide_enabled => graphiql().into_response(),
}
```

crates/sui-indexer-alt-graphql/src/lib.rs
tpham-mysten: Done

crates/sui-indexer-alt-graphql/src/extensions/query_limits/mod.rs
amnn: Ah got it, that's unfortunate, but I guess is done to enable subscription multiplexing.

crates/sui-indexer-alt-graphql/src/api/types/snapshots/sui_indexer_alt_graphql__api__types__available_range__field_piplines_tests__registry_collect_pipelines_snapshot.snap
amnn, line 1: This file may now need a `staging` variant, not sure.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -37,7 +36,7 @@ pub(crate) struct Timeout(Arc<TimeoutConfig>);
 
 struct TimeoutExt {
     config: Arc<TimeoutConfig>,
-    is_mutation: AtomicBool,
+    operation_type: Mutex<OperationType>,
```
amnn: [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) is a slightly better fit.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
```diff
@@ -66,13 +65,9 @@ impl Extension for TimeoutExt {
...
         let document = next.run(ctx, query, variables).await?;
 
-        self.is_mutation.store(
-            document
-                .operations
-                .iter()
-                .any(|(_, op)| op.node.ty == OperationType::Mutation),
-            Ordering::Relaxed,
-        );
+        if let Some((_, op)) = document.operations.iter().next() {
```
amnn: This is doing something slightly different now -- only picking the first operation. Is that a problem? Is it possible to trojan horse one operation behind another?

I remember at one point we had a restriction that you could only have one operation per document, but I'm not sure we still have that (and I don't recall the circumstances in which it's valid to send a document with multiple operations in it, whether their types can differ etc).

crates/sui-indexer-alt-graphql/src/lib.rs
amnn: I think with the merged route, this consideration goes away.

crates/sui-indexer-alt-graphql/src/api/types/snapshots/sui_indexer_alt_graphql__api__types__available_range__field_piplines_tests__registry_collect_pipelines_snapshot.snap
tpham-mysten: Done. I added the staging variant.

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/extensions/timeout.rs
tpham-mysten: > Is it possible to trojan horse one operation behind another?

Good catch. A document can have multiple named operations with different types, and only one is executed based on operation_name. For example:

```
  subscription Foo { checkpoints { sequenceNumber } }
  query Bar { checkpoint { digest } }
```

If operation_name: "Bar", the old code would pick Foo (subscription) as the type and skip the timeout, even though Bar (query) is the one actually executed.

I've fixed this by storing a map of operation name -> type during `parse_query`, then looking up the correct one by `operation_name` in execute.

PR #26020, opened 2026-03-27
[move-compiler] Macro support in debug info

awelc:
## Description 

This PR adds support for recording precise information about control flow related to macros at the bytecode level into debug info.
This is essential so the debugger treats macro calls similarly to function calls and shows correct call stack frame transitions. Without it, inlining and optimizations can cause some frame transitions to be lost.
Consider the following example:

```module A::m {
    macro fun inner($x: u64): u64 {
        $x + 1
    }

    macro fun outer($x: u64): u64 {
        inner!($x)
    }

    public fun test(v: u64): u64 {
        outer!(v)
    }
}
```
When `inner` is inlined in `outer`, which is then inlined in `test`, the compilation process results in the first macro instruction in the bytecode being from `inner`, and the information that `inner` is reached via `outer` is lost. This gets even worse if `outer` has some non-optimized code following the call to `inner`, because upon returning from the call" we will see both instructions from `inner` and `outer`, making frame transitions asymmetric. The debugger cannot reconcile this based on what is in the bytecode, trace, and debug info prior to this PR.

This PR attempts to solve this by providing frame transition information in debug info (aka source map) for each "important" inlined instruction. The "important" instruction is one that either:
- starts a new inlined macro body
- starts a new inlined lambda body
- starts a new inlined argument evaluation

In other words, debug info contains a sparse map that the debugger can consult to see if a frame transition is needed before a given instruction and what this frame transition should be.

On the technical level, this is done by preserving two pieces of information throughout compilation
- hierarchy of colors that describe progressively nested inlining levels. 
- color for each bytecode instruction (though of course these do not end up being in the final generated bytecode)

Example of hierarchy of colors the code snippet above would be
0 - outer function (the one calling into a macro) color
1 - `outer` macro body color
2 - `inner` macro body color
3 - argument color

With this information, even if the first inlined instruction in outer instruction is the one from `inner` (and is stamped with `inner`'s color), we know based on the hierarchy of colors that in addition to a frame for `inner` we also need to have a frame for `outer`

In the implementation, we don't really use numeric colors for these but rather encode information about macro "frame transition" directly (`MacroInfo` struct).


external-crates/move/crates/move-compiler/src/shared/mod.rs
```diff
@@ -654,6 +640,33 @@ impl CompilationEnv {
...
+
+    pub fn add_macro_frame_info(
+        &self,
+        module: ModuleIdent_,
+        function: Symbol,
+        info: macro_frames::ColorFrameInfo,
+    ) {
+        self.macro_frame_infos
+            .write()
+            .unwrap()
```
cgswords: These unwraps are a touch dubious, but maybe okay. We do it everywhere else.

PR #26021, opened 2026-03-27
[consensus] fail over to block sync when commit sync makes no progress

mwtian:
## Description 

There are a number of scenarios when commit sync is slow or not making progress. Fail over to block sync in these cases.

## Test plan 

PT

### DIFF
--- a/consensus/core/src/synchronizer.rs
+++ b/consensus/core/src/synchronizer.rs
@@ -1036,9 +1036,14 @@ where
         let commit_threshold = current_commit_index
             + self.context.parameters.commit_sync_batch_size * COMMIT_LAG_MULTIPLIER;
         let now = Instant::now();
+        let metrics = &self.context.metrics.node_metrics;
 
         // Commit is not lagging.
         if quorum_commit_index <= commit_threshold {
+            metrics
+                .synchronizer_periodic_sync_decision
+                .with_label_values(&["true", "default"])
+                .inc();
             // Reset last commit state.
             self.last_changed_commit_index = current_commit_index;
             self.last_commit_change_time = now;
@@ -1053,11 +1058,19 @@ where
             if current_commit_index
                 < self.last_changed_commit_index + self.context.parameters.commit_sync_batch_size
             {
+                metrics
+                    .synchronizer_periodic_sync_decision
+                    .with_label_values(&["true", "commit_catchup::run"])
+                    .inc();
                 // Not enough progress has been made yet. Keep running periodic sync.
                 // Do not update last commit state yet.
                 // Run periodic sync.
                 return true;
             } else {
+                metrics
+                    .synchronizer_periodic_sync_decision
+                    .with_label_values(&["false", "commit_catchup::end"])
+                    .inc();
                 // Enough progress has been made. Disable commit sync failover and reset last commit state.
                 self.last_changed_commit_index = current_commit_index;
                 self.last_commit_change_time = now;
@@ -1071,17 +1084,24 @@ where
         if current_commit_index == self.last_changed_commit_index {
             // Enter commit sync failover if not enough progress has been made.
             if now.duration_since(self.last_commit_change_time) >= COMMIT_PROGRESS_TIMEOUT {
+                metrics
+                    .synchronizer_periodic_sync_decision
+                    .with_label_values(&["true", "commit_catchup::start"])
+                    .inc();
                 self.commit_sync_failover = true;
                 // Run periodic sync.
                 return true;
             }
-            // Do not update last commit state when commit index is not changing.
         } else {
-            // Update last commit state.
+            // IMPORTANT: Only update last commit state when commit index is changing.
             self.last_changed_commit_index = current_commit_index;
             self.last_commit_change_time = now;
         }
 
+        metrics
+            .synchronizer_periodic_sync_decision
+            .with_label_values(&["false", "commit_lag"])
+            .inc();
         false
     }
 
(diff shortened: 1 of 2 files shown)


consensus/core/src/authority_service.rs
```diff
@@ -391,13 +392,21 @@ impl<C: CoreThreadDispatcher> ValidatorNetworkService for AuthorityService<C> {
...
+    //    - response returns max_blocks_per_sync blocks.
+    // 2. Periodic sync:
+    //    - Highest accepted rounds must be specified.
+    //    - Missing block refs are optional.
+    //    - breadth_first is false.
+    //    - response returns max_blocks_per_fetch blocks.
+    // 3. Commit sync:
+    //    - Missing block refs are specified.
+    //    - Highest accepted rounds are empty.
+    //    - breadth_first should be false but is ignored.
```
akichidis: Why it should be false? As default state you mean?

consensus/core/src/authority_service.rs
```diff
@@ -407,6 +416,9 @@ impl<C: CoreThreadDispatcher> ValidatorNetworkService for AuthorityService<C> {
     ) -> ConsensusResult<Vec<Bytes>> {
         fail_point_async!("consensus-rpc-response");
 
+        if block_refs.is_empty() && (breadth_first || highest_accepted_rounds.is_empty()) {
+            return Err(ConsensusError::InvalidFetchBlocksRequest("When no block refs are provided, highest_accepted_rounds must be provided and breadth_first must be false".to_string()));
+        }
```
akichidis: So we want this endpoint to work into two basic modes:
1. Specify `block_refs` - fetch any of those blocks
2. Specify the `highest_accepted_rounds` - try to fetch blocks for each authority from the provided `highest_accepted_rounds` up to whatever max round block this node already has

On top of the above there is some flavours of options to configure additional fetching.

To be honest, I find the validation/modes a bit subtle. Do you believe we could break down a bit the code or the state management here?

consensus/core/src/authority_service.rs
```diff
@@ -463,63 +476,72 @@ impl<C: CoreThreadDispatcher> ValidatorNetworkService for AuthorityService<C> {
...
+                    )?;
+                    if let Some(block) = fetched.into_iter().next() {
+                        let next_start = block.round() + 1;
+                        blocks.push(block);
+                        if blocks.len() >= max_response_num_blocks {
+                            blocks.truncate(max_response_num_blocks);
+                            break;
+                        }
+                        if next_start < limit_round {
+                            heap.push(Reverse((next_start, authority, limit_round)));
```
akichidis, line 539: IIUC the intention here is to fetch blocks across all the authorities but ensuring some "fairness" so we fill up the fetched blocks from lower -> higher round across all authorities, right?

consensus/core/src/synchronizer.rs
```diff
@@ -919,9 +926,19 @@ where
...
+            .get_missing_blocks()
+            .await
+            .map_err(|_err| ConsensusError::Shutdown)?;
+        if self.commit_sync_failover {
+            // Keep missing blocks to those that must be included in fetch request.
+            // Filtered out missing blocks that will eventually be fetched with highest accepted rounds.
+            let highest_accepted_rounds =
+                Self::get_highest_accepted_rounds(dag_state.clone(), &context);
+            missing_blocks
+                .retain(|block| block.round <= highest_accepted_rounds[block.author.value()]);
```
akichidis: Are we doing this for request optimisations reasons? Or because the endpoint will respond better to this? Otherwise I would suggest just keeping it simpler.

consensus/core/src/synchronizer.rs
```diff
@@ -935,16 +952,28 @@ where
...
+                let results = if missing_blocks.is_empty() {
+                    let _scope = monitored_scope("BlockSync::Periodic::HighestAcceptedRounds");
+                    // Fetch blocks from a random peer using highest accepted rounds (commit sync failover)
+                    Self::fetch_blocks_with_highest_accepted_rounds(
+                        context.clone(),
+                        blocks_to_fetch.clone(),
+                        network_client,
+                        dag_state,
+                    )
+                    .await
```
akichidis: My understanding is that the cases for which we want to keep sync running is when:
* no certified commits can be provided by peers
* for whatever reason, commit sync is dealing with difficulties (ex timing out etc)

or you are looking on some other cases?

consensus/core/src/synchronizer.rs
```diff
@@ -994,15 +1025,121 @@ where
...
+                // Skip periodic sync because of commit lag.
+                return false;
+            }
+        }
+
+        // The node is commit lagging and not in commit sync failover yet.
+        if current_commit_index == self.last_changed_commit_index {
+            // Enter commit sync failover if not enough progress has been made.
+            if now.duration_since(self.last_commit_change_time) >= COMMIT_PROGRESS_TIMEOUT {
+                self.commit_sync_failover = true;
```
akichidis, line 1091: Do you believe it would be useful to have a metric to tell when we get into commit sync failover?

consensus/core/src/authority_service.rs
mwtian: Yes

consensus/core/src/synchronizer.rs
mwtian: This is the main case. Basically we want periodic block sync to be default enabled, unless we know commit sync must be running and not stuck.

consensus/core/src/synchronizer.rs
mwtian: Updated comment.

consensus/core/src/authority_service.rs
mwtian: The change will become too large if I attempt to also simplify the periodic sync logic. That will be done in a future PR.

PR #26023, opened 2026-03-27
[adapter][hardening] move to `safe_unwrap/safe_assert` in natives

tzakian:
## Description 

More safety is always a good thing :)

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-move-natives/src/accumulator.rs
+++ b/sui-execution/latest/sui-move-natives/src/accumulator.rs
@@ -48,8 +48,7 @@ fn emit_event(
     native_charge_gas_early_exit!(context, event_emit_cost_params.event_emit_cost_base);
 
     let amount = safe_unwrap_err!(safe_unwrap!(args.pop_back()).value_as::<u64>());
-    let recipient =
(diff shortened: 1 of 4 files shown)


external-crates/move/crates/move-binary-format/src/lib.rs
```diff
@@ -158,8 +158,10 @@ macro_rules! safe_unwrap_err {
         match $e {
             Ok(x) => x,
             Err(e) => {
-                let err = PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
-                    .with_message(format!("{}:{} {:#}", file!(), line!(), e));
+                let err = move_binary_format::errors::PartialVMError::new(
```
tnowacki, line 161: Should this be $crate ?

tzakian: It probably should be, but matches the other macro definitions right by it. I'll fix up in a followup PR though.

PR #26025, opened 2026-03-27
Replace debug_assertions with testing feature gate for test-only code

alex-mysten:
## Summary
- Introduces a `testing` Cargo feature in `sui-types` to gate test-only code instead of `#[cfg(debug_assertions)]`
- Adds a `testing` feature to `sui-transactional-test-runner` that propagates `sui-types/testing`
- Updates all upstream test crates to enable the feature via dev-dependencies
- Adds CLAUDE.md guidance on when to use `#[cfg(test)]` vs `#[cfg(feature = "testing")]`

## Test plan
- [x] `cargo check --release -p sui-transactional-test-runner` passes (test-only code excluded)
- [x] `cargo check --release -p sui-types` passes
- [x] Gasless transactional tests pass with feature enabled
- [x] `cargo xclippy` clean

### DIFF
--- a/crates/sui-transactional-test-runner/Cargo.toml
+++ b/crates/sui-transactional-test-runner/Cargo.toml
@@ -61,6 +61,10 @@ sui-framework-snapshot.workspace = true
 sui-storage.workspace = true
 typed-store.workspace = true
 
+[[bin]]
+name = "sui-transactional-test-runner"
+required-features = ["testing"]
+
 [features]
 testing = ["sui-types/testing"]
 
--- a/crates/sui-transactional-test-runner/src/lib.rs
+++ b/crates/sui-transactional-test-runner/src/lib.rs
@@ -3,55 +3,71 @@
 
 //! This module contains the transactional test runner instantiation for the Sui adapter
 
+#[cfg(feature = "testing")]
 pub mod args;
+#[cfg(feature = "testing")]
 pub mod cursor;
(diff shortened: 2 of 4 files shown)


crates/sui-transactional-test-runner/Cargo.toml
```diff
@@ -61,5 +61,8 @@ sui-framework-snapshot.workspace = true
 sui-storage.workspace = true
```
tnowacki, line 61: Is it possible to put the whole test runner under "testing"?

alex-mysten: I did some digging here, it would be invasive to gate the whole crate under "testing". It seems like there should be a simple canonical way to mark a crate and upstreams as test-only, but that does not seem to be the case. Claude identified a few options for excluding the crate from release builds, but they have tradeoffs / increase cognitive load:

1. Use `default-members` - this would exclude it from any monorepo-wide build (only buildable using -p)
2. Remove it as a workspace member - we'd have to manually manage dependent crate versions in the Cargo.toml
3. separate, independently managed workspaces - too complex

I think it's simpler as-is

tnowacki: I would just slap the attribute on each module. But I can look once it lands

alex-mysten: Done :)

PR #26029, opened 2026-03-27
Composite workload: test accumulator create/delete cycles via coin reservations

mystenmark:
## Summary

Adds `CoinReservationWithdraw` to the composite benchmark workload. The goal is to repeatedly create and delete accumulators: each withdrawal drains an account's accumulator to zero (deleting it), then deposits the funds into the partner's accumulator (creating it if needed). Over time, accounts oscillate between having and not having an accumulator for a given coin type.

- Payloads are paired (0↔1, 2↔3, etc.); each successful withdrawal deposits the balance into the partner's accumulator via `balance::send_funds`
- End-of-test assertion verifies that at least one account completed the round-trip (≥2 successful withdrawals), proving accumulator deletion and recreation both worked

## Test plan

- [ ] `cargo simtest -p sui-benchmark test_composite_workload` passes for 20 seeds

### DIFF
--- a/crates/sui-benchmark/src/workloads/composite/mod.rs
+++ b/crates/sui-benchmark/src/workloads/composite/mod.rs
@@ -148,11 +148,6 @@ pub struct OperationSetStats {
 #[derive(Debug)]
 pub struct CompositionMetrics {
     stats: HashMap<OperationSet, OperationSetStats>,
-    /// Counts successful coin reservation withdrawals per account address.
-    /// Used to verify that paired accounts successfully transfer back and forth:
-    /// a count >= 2 for any account proves the round-trip works (first success drains
-    /// the seeded balance, second success requires the partner to have deposited back).
-    coin_reservation_success_count: HashMap<SuiAddress, u64>,
 }
 
 impl Default for CompositionMetrics {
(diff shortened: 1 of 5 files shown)


crates/sui-benchmark/tests/simtest.rs
```diff
@@ -1631,6 +1631,21 @@ mod test {
                 accum_read_stats.success_count > 0,
                 "expected at least one accumulator balance read"
             );
+
+            // Verify the coin reservation round-trip pattern works: an account achieving
+            // >= 2 successful withdrawals proves funds flowed back from its partner
+            // (first success drains seeded balance; second requires partner to have deposited).
+            let max_coin_res_successes = metrics.max_coin_reservation_success_count();
```
alex-mysten: Seems possibly brittle assumption that any coin reservation will drain the account
do we want to / can we observe accumulator deletion/creation rather than relying on proxy?

crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -1043,6 +1083,10 @@ impl Payload for CompositePayload {
                         metrics.record_insufficient_funds(tx_info.op_set.clone());
                     } else if effects.is_ok() {
                         metrics.record_success(tx_info.op_set.clone());
+                        // Track successful coin reservation withdrawals for timeout checking
```
alex-mysten: timeouts are relevant because the simtest assertion will fail?

alex-mysten (approved):
LGTM!

PR #26036, opened 2026-03-28
[move][feature] Add object::new_with_salt and new_with_salt_as for deterministic object ID creation

theigwe:
## Description 

Introduces two new Move framework functions for deterministic object ID creation.

Motivation: Current object IDs are derived from Hash(0xf1 || tx_digest || counter), binding them to a specific transaction and making them unknowable before execution. This prevents counterfactual instantiation patterns where funds or state need to be sent to an address before the object exists.

How it works: The new scheme hashes Hash(0xf2 || creator_address || salt) — the same creator and salt always yield the same object ID, independent of when or in which transaction the object is created.

Unlike regular IDs, no transaction counter is involved; uniqueness is enforced via a post-execution backing-store collision check in TemporaryStore before effects are committed.

New API:
- object::new_with_salt(salt, ctx) — sender-scoped; ID namespace is implicitly owned by ctx.sender(), no squatting possible
- object::new_with_salt_as(creator, _proof, salt, ctx) — factory-scoped; _proof is the &UID of an object whose address equals creator, proving ownership of that namespace without requiring ctx.sender()

Key design decisions:
- in new_with_salt, ctx.sender() is used as creator to prevent salt-squatting across address namespaces
- new_with_salt_as uses &UID as an unforgeable capability — Move's type system guarantees possession, no additional auth mechanism needed
- Salted IDs are tracked separately from regular IDs in ObjectRuntimeState — regular IDs are provably collision-free by counter; only salted IDs need the backing-store check
- Collision detection runs post-execution in TemporaryStore before effects are committed (the native has no backing-store access)
- Feature-gated behind protocol version 121, enabled on devnet/testnet first

Changes:
- New HashingIntentScope::SaltedObjectId = 0xf2 discriminant
- ObjectID::derive_id_with_salt in sui-types
- Protocol config version 121 with enable_salted_object_ids flag and object_new_with_salt_cost_base = 260
- Native new_with_salt_impl wired through the Move natives layer
- object::new_with_salt and object::new_with_salt_as in sui-framework

## Test plan

1. Protocol config snapshot tests - verified version 121 snapshots are correct and enable_salted_object_ids is off on Mainnet, on for Devnet/Testnet: 
```shell
cargo test -p sui-protocol-config
```                                                                                                                                             
2. Type-check across all touched crates — confirmed no compilation errors:                                                                                                           
```shell
cargo check -p shared-crypto -p sui-types -p sui-protocol-config -p sui-framework -p sui-move-natives-latest -p sui-adapter-latest -p sui-execution
```                                  
3. Move framework unit tests — test_deterministic, test_different_salt, test_different_creator in object_salted_id_tests.move cover the core ID derivation properties:
```shell               
cargo test -p sui-framework-tests
```                                                                                                                                                
4. Transactional tests — salted_objects/basic.move and salted_objects/collision.move cover end-to-end object creation, transfer, and re-creation-after-deletion scenarios:
```shell           
cargo test -p sui-adapter-transactional-tests
```
---                                                                                                                                                                                  
## Release notes                                                                                                                                                                   
- [✔︎] Protocol:
Introduces protocol version 121 with a new enable_salted_object_ids feature flag. Activated on Devnet and Testnet; not enabled on Mainnet. Adds a new HashingIntentScope::SaltedObjectId = 0xf2 discriminant for deterministic ID derivation and a new SaltedObjectIdAlreadyExists execution error kind for collision detection. No action required for node operators on Mainnet at this time.  
                                                                                                                               
- [✔︎] Nodes (Validators and Full nodes):                                                                                                                                                 
Validators and full nodes on Devnet/Testnet must upgrade to a binary that supports protocol version 121 before the version is activated on those networks. The new post-execution collision check in TemporaryStore adds a minor overhead for transactions that use new_with_salt or new_with_salt_as.

### DIFF
--- a/crates/shared-crypto/src/intent.rs
+++ b/crates/shared-crypto/src/intent.rs
@@ -194,4 +194,5 @@ pub(crate) mod private {
 pub enum HashingIntentScope {
     ChildObjectId = 0xf0,
     RegularObjectId = 0xf1,
+    SaltedObjectId = 0xf2,
 }
--- a/crates/sui-framework/packages/sui-framework/sources/object.move
+++ b/crates/sui-framework/packages/sui-framework/sources/object.move
@@ -58,6 +58,9 @@ const SUI_ADDRESS_ALIAS_STATE_ID: address = @0xa;
 /// Sender is not @0x0 the system address.
 const ENotSystemAddress: u64 = 0;
 
+/// Sender is not the creator address passed to `new_with_salt`.
+const ENotCreator: u64 = 2;
+
 /// An object ID. This is used to reference Sui Objects.
 /// This is *not* guaranteed to be globally unique--anyone can create an `ID` from a `UID` or
 /// from an object, and ID's can be freely copied and dropped.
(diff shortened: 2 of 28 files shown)


theigwe:
@tnowacki

Thank you for taking time to respond.

The referenced functionality serves my use case for now.

If I figure out further useful improvement,  I will raise an SIP for that.

PR #26041, opened 2026-03-30
sui-rpc-resolver in Analytics Indexer

nickvikeras:
## Description

The analytics indexer has long had a bespoke package cache for type layouts. It was a 2-level cache (in-memory LRU, then rocksdb), and only the in-memory layer was being invalidated for system packages on epoch changes, which led to a stale cache and stalled pipelines when display registry rolled out.

Migrating to the sui-rpc-resolver (mostly) addresses this. Some more context in <https://linear.app/mysten-labs/issue/DP-246/use-sui-rpc-resolver-in-analytics-indexer>

A new pseudohandler invalidates the package cache on epoch changes which is much simpler than the old code.

I think there is still a hypothetical race where we are subscribed to Fullnode A, and call Fullnode B to load type layouts after the cache invalidation, and Fullnode B hasn't yet processed the epoch change. In practice it's extraordinarily unlikely we hit this, and we can't do much about it without consistent-store-like gRPC APIs anyway.

## Test plan

Analytics indexer unit tests are passing.


evan-wall-mysten:
> I think there is still a hypothetical race where we are subscribed to Fullnode A, and call Fullnode B to load type layouts after the cache invalidation, and Fullnode B hasn't yet processed the epoch change.

If this happens, is does the indexer need to be restarted? Is there a way to detect if this happened and at least log an error?

crates/sui-analytics-indexer/src/handlers/system_package_eviction.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+#[async_trait]
+impl Processor for SystemPackageEviction {
+    const NAME: &'static str = "SystemPackageEviction";
+    type Value = ();
+
+    async fn process(&self, checkpoint: &Arc<Checkpoint>) -> Result<Vec<()>> {
+        let epoch = checkpoint.summary.data().epoch;
+        if self.last_epoch.swap(epoch, Ordering::Relaxed) != epoch {
```
evan-wall-mysten, line 40: Is it possible to do something like this to handle the edge case you mentioned in the description?
```
if old_epoch < new_epoch {
  evict()
} else if old_epoch > new_epoch {
  // block until the fullnode you are connected to updates to the new epoch
}
```

nickvikeras: The issue is the full node APIs don't have a checkpoint-height argument like the consistent store. It just gives you the data for whatever checkpoint height it's at. And since we connect to the fullnodes through the k8s service name (which could resolve to many different full nodes) there isn't really a way to reliably verify the fullnode is at a certain height before querying it.

nickvikeras:
> > I think there is still a hypothetical race where we are subscribed to Fullnode A, and call Fullnode B to load type layouts after the cache invalidation, and Fullnode B hasn't yet processed the epoch change.
> 
> If this happens, is does the indexer need to be restarted? Is there a way to detect if this happened and at least log an error?

Yeah it would need to be restarted. I don't think there is really a way to know it happened programmatically though (see my other reply). I think it's also pretty unlikely to actually occur in-practice. 

I think the only real way to fix it systemically is to merge the consistent store with grpc, or to return checkpoint height metadata in the response headers.

PR #26042, opened 2026-03-30
sui-indexer-alt-framework: make transient_backoff, retry_transient_with_slow_monitor helper functions

evan-wall-mysten:
## Description 

This change reduces the amount of copied and pasted code by creating reusable helper methods for the ingestion client calls. These can be reused in a future change to retry getting the latest checkpoint number.

## Test plan 

No new functionality, existing unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
@@ -1,6 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::future::Future;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
@@ -18,6 +19,7 @@ use object_store::azure::MicrosoftAzureBuilder;
 use object_store::gcp::GoogleCloudStorageBuilder;
 use object_store::http::HttpBuilder;
 use object_store::local::LocalFileSystem;
+use prometheus::Histogram;
 use sui_futures::future::with_slow_future_monitor;
 use sui_rpc::Client;
 use sui_rpc::client::HeadersInterceptor;
@@ -28,11 +30,10 @@ use tracing::error;
 use tracing::warn;
 use url::Url;
 
-use crate::ingestion::Error as IngestionError;
+use crate::ingestion::Error as IE;
 use crate::ingestion::MAX_GRPC_MESSAGE_SIZE_BYTES;
 use crate::ingestion::Result as IngestionResult;
 use crate::ingestion::decode;
-use crate::ingestion::error::Error::FetchError;
 use crate::ingestion::store_client::StoreIngestionClient;
 use crate::metrics::CheckpointLagMetricReporter;
 use crate::metrics::IngestionMetrics;
@@ -300,7 +301,7 @@ impl IngestionClient {
         let fetch = || async move {
             use backoff::Error as BE;
             self.checkpoint(checkpoint).await.map_err(|e| match e {
-                IngestionError::NotFound(checkpoint) => {
+                IE::NotFound(checkpoint) => {
                     debug!(checkpoint, "Checkpoint not found, retrying...");
                     self.metrics.total_ingested_not_found_retries.inc();
                     BE::transient(e)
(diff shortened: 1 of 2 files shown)


crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
```diff
@@ -342,18 +343,17 @@ impl IngestionClient {
...
                         BE::permanent(IngestionError::NotFound(checkpoint))
                     }
-                    CheckpointError::Transient { reason, error } => self.metrics.inc_retry(
-                        checkpoint,
-                        reason,
-                        IngestionError::FetchError(checkpoint, error),
-                    ),
+                    CheckpointError::Transient { reason, error } => {
+                        self.metrics
+                            .inc_retry(checkpoint, reason, FetchError(checkpoint, error))
```
amnn: I ended up confused by this for a while -- I hadn't spotted that we had imported a specific enum variant as part of top-level imports. We should not do that, because it goes against reader expectations -- it makes this call seem like we're constructing a struct rather than an enum variant. This is especially confusing for errors, because it makes it hides the signal that there are other potential error variants.

Always refer to enum variants scoped by their type (it's fine to import the type using some shorter prefix if it's referenced a lot, like `IE` in this case). Can you undo that previous change to introduce the `FetchError` import please?

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
```diff
@@ -415,67 +408,79 @@ impl IngestionClient {
...
-                };
-
-                // Keep backing off until we are waiting for the max interval, but don't give up.
-                let backoff = ExponentialBackoff {
-                    max_interval: MAX_TRANSIENT_RETRY_INTERVAL,
-                    max_elapsed_time: None,
-                    ..Default::default()
-                };
+/// Keep backing off until we are waiting for the max interval, but don't give up.
+fn default_backoff() -> ExponentialBackoff {
```
amnn: "default" is not a good name for this -- it implies that it could be used for anything. `transient_backoff` would work.

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
```diff
@@ -415,67 +408,79 @@ impl IngestionClient {
...
+    ExponentialBackoff {
+        max_interval: MAX_TRANSIENT_RETRY_INTERVAL,
+        max_elapsed_time: None,
+        ..Default::default()
+    }
+}
 
-                backoff::future::retry(backoff, request)
+/// Retry an fallible async operation with exponential backoff and slow-operation monitoring.
```
amnn:
```suggestion
/// Retry a fallible async operation with exponential backoff and slow-operation monitoring.
```

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
```diff
@@ -415,67 +408,79 @@ impl IngestionClient {
             .total_ingested_objects
             .inc_by(data.object_set.len() as u64);
 
-        let chain_id = *self.get_or_init_chain_id(checkpoint).await?;
+        let client = self.client.clone();
```
amnn: Not strictly to do with this PR but could you/should you run this request concurrently with fetching the checkpoint data?

amnn: We may also want to measure the elapsed time for this request as well. It should be minimal, but if for some reason it is slow or requires many retries, we would not have visibility on that -- the metrics would suggest that checkpoint fetching was fast, but we would be seeing inexplicable delays.

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
```diff
@@ -415,67 +408,79 @@ impl IngestionClient {
...
+        max_interval: MAX_TRANSIENT_RETRY_INTERVAL,
+        max_elapsed_time: None,
+        ..Default::default()
+    }
+}
 
-                backoff::future::retry(backoff, request)
+/// Retry an fallible async operation with exponential backoff and slow-operation monitoring.
+/// All errors are treated as transient and retried.
+async fn retry_with_slow_monitor<F, Fut, T>(
```
amnn:
```suggestion
async fn retry_transient_with_slow_monitor<F, Fut, T>(
```

amnn: You probably also want to tweak the `SLOW_OPERATION_WARNING_THRESHOLD` based on which operation it is -- fetching a checkpoint is expected to take longer than fetching the chain ID or latest checkpoint number.

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
evan-wall-mysten: I changed the signature of `retry_transient_with_slow_monitor` to include a metric so that the latency of all usages can be monitored.

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
evan-wall-mysten: I renamed the method, but I'll leave `SLOW_OPERATION_WARNING_THRESHOLD` at 60s for everything now to keep it simple since it looks like the p99 is under 1s (I think this is measuring that https://metrics.sui.io/goto/ffhsgpb4mmyv4d?orgId=stacks-447897) which is probably true for the other operations.

PR #26043, opened 2026-03-30
[e2e] fix test_alias_changes flake by waiting for enable tx settlement

lxfind:
## Summary
- Fix intermittent failure in `test_alias_changes` (failed on iteration 19/20, seed 4616056488367146196)
- Root cause: after the `enable` tx creates the `AddressAliases` object on the submitting validator, the next transaction's alias claim mismatches on other validators that haven't executed `enable` yet, causing consensus rejection
- Add `wait_for_tx_settlement` after `enable`, matching the existing pattern already used for `add` and `remove` in the same test

## Test plan
- [ ] Run `test_alias_changes` with `--count 20` to verify the flake is fixed
- [ ] Confirm `test_alias_race` still passes (it intentionally tests the rejection path)


aschran:
ty for the fix!

PR #26046, opened 2026-03-30
sui-indexer-alt-e2e-tests: bump graphql_zklogin_tests graphql timeout

evan-wall-mysten:
## Description 

Increases the graphql timeout again to reduce flakes locally like in #25914.

## Test plan 

`cargo nextest run -p sui-indexer-alt-e2e-tests` passing locally.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql_zklogin_tests.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql_zklogin_tests.rs
@@ -102,7 +102,7 @@ impl FullCluster {
         // Trigger an epoch change and wait until GraphQL sees Epoch 1
         onchain.trigger_reconfiguration().await;
         onchain.wait_for_authenticator_state_update().await;
-        tokio::time::timeout(Duration::from_secs(20), async {
+        tokio::time::timeout(Duration::from_secs(30), async {
             let mut interval = interval(Duration::from_millis(200));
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-e2e-tests/tests/graphql_zklogin_tests.rs
```diff
@@ -102,7 +102,7 @@ impl FullCluster {
         // Trigger an epoch change and wait until GraphQL sees Epoch 1
         onchain.trigger_reconfiguration().await;
         onchain.wait_for_authenticator_state_update().await;
-        tokio::time::timeout(Duration::from_secs(20), async {
+        tokio::time::timeout(Duration::from_secs(30), async {
```
tpham-mysten, line 105: Is 30 a safe threshold then? What is the duration of recent failures?

evan-wall-mysten: The tests were taking a little over 20s when I just ran them so 30s should be safe for now. If they continue to take longer then we can reinvestigate.

PR #26050, opened 2026-03-30
sui-indexer-alt-framework: CheckpointStream peekable, timeout

evan-wall-mysten:
## Description 

This change makes the stream inside CheckpointStream peekable and moves the stream read timeout into `GrpcStreamingClient` so it can be shared across all users. A follow up change will add another usage to determine the network tip when the indexer starts.

## Test plan 

Update MockStreamingClient and related tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
@@ -1,14 +1,14 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::pin::Pin;
 use std::time::Duration;
 
 use anyhow::Context;
 use anyhow::anyhow;
 use async_trait::async_trait;
-use futures::Stream;
 use futures::StreamExt;
+use futures::stream::BoxStream;
+use futures::stream::Peekable;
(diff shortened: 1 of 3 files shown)


crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
```diff
@@ -85,31 +87,49 @@ impl CheckpointStreamingClient for GrpcStreamingClient {
             .map_err(|e| Error::StreamingError(anyhow!("Chain ID parse error: {e}")))?
             .into();
 
-        let converted_stream = response.into_inner().map(|result| match result {
+        let stream = response.into_inner().map(|result| match result {
+            Err(e) => Err(Error::RpcClientError(e)),
```
tpham-mysten: tiny nit: Is this intended change? Seems like the common pattern is `Err` after `Ok`.

evan-wall-mysten: Accidental - will switch it back

PR #26056, opened 2026-03-30
[consensus] probe before fetching commits and blocks from a peer

mwtian:
## Description 

When fetching commits and blocks, progressively long timeouts are needed in case there are a large amount of data. Probe the peers before fetching to reduce the delay to skip over unavailable peers.

## Test plan 

PT

### DIFF
--- a/consensus/config/src/parameters.rs
+++ b/consensus/config/src/parameters.rs
@@ -100,7 +100,7 @@ pub struct Parameters {
     #[serde(default = "Parameters::default_commit_sync_request_timeout")]
     pub commit_sync_request_timeout: Duration,
 
-    // Timeout for the reachability probe (get_latest_rounds()) before committing to a full fetch.
+    // Timeout for the connectivity probe against a peer before committing to a full fetch.
     // Should be short to quickly skip unreachable peers.
     #[serde(default = "Parameters::default_commit_sync_probe_timeout")]
     pub commit_sync_probe_timeout: Duration,
@@ -214,7 +214,7 @@ impl Parameters {
     }
 
     pub(crate) fn default_commit_sync_probe_timeout() -> Duration {
-        Duration::from_secs(1)
+        Duration::from_secs(2)
     }
 
     pub(crate) fn default_commit_sync_batches_ahead() -> usize {
--- a/consensus/core/src/authority_service.rs
+++ b/consensus/core/src/authority_service.rs
@@ -32,7 +32,10 @@ use crate::{
     core_thread::CoreThreadDispatcher,
     dag_state::DagState,
     error::{ConsensusError, ConsensusResult},
-    network::{BlockStream, ExtendedSerializedBlock, PeerId, ValidatorNetworkService},
+    network::{
+        BlockStream, ExtendedSerializedBlock, NodeId, ObserverBlockStream, ObserverNetworkService,
+        PeerId, ValidatorNetworkService,
+    },
     round_tracker::RoundTracker,
     stake_aggregator::{QuorumThreshold, StakeAggregator},
     storage::Store,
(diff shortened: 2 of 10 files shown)


consensus/core/src/authority_service.rs
```diff
@@ -679,6 +679,16 @@ impl<C: CoreThreadDispatcher> ObserverNetworkService for AuthorityService<C> {
             "Observer fetch commits not yet implemented".to_string(),
         ))
     }
+
+    async fn handle_get_latest_rounds(
+        &self,
+        _peer: NodeId,
+    ) -> ConsensusResult<(Vec<Round>, Vec<Round>)> {
+        // TODO: implement observer get_latest_rounds, similar to validator get_latest_rounds.
+        Err(ConsensusError::NetworkRequest(
```
akichidis: The PR is not reviewed yet, but something to keep in mind is that I've extracted common application logic to the `BlockSyncService` https://github.com/MystenLabs/sui/pull/25997/changes#diff-e419a2c434171dce71c787bfff7e22cfec84ac0563d1999394289d16ec03e765R26 so both the `AuthorityService` and the `ObserverService` can re-use. Maybe not something to worry about now, but just to create some awareness.

consensus/core/src/commit_syncer.rs
```diff
@@ -549,6 +548,17 @@ where
...
+        // 0. Probe the target to check reachability before committing to the full fetch.
+        // This skips unreachable and slow peers quickly.
+        let probe_timeout = inner.context.parameters.commit_sync_probe_timeout;
+        inner
+            .network_client
+            .get_latest_rounds(
+                crate::network::PeerId::Validator(target_authority),
+                probe_timeout,
+            )
+            .await?;
```
akichidis, line 560: Have we thought of doing the probing on the background with some periodic cadence? Even though the check should be indeed quite quick, maybe it's too busy? I guess on the grand scheme of things it should be fine, but just iterating on the idea.

akichidis: Also as we discussed offline this might make less sense for the Observer nodes, but we can absolutely disable the check for Observer nodes.

consensus/core/src/commit_syncer.rs
```diff
@@ -986,6 +996,14 @@ mod tests {
         ) -> ConsensusResult<(Vec<Bytes>, Vec<Bytes>)> {
             unimplemented!("Unimplemented")
         }
+
+        async fn get_latest_rounds(
```
akichidis: As you proposed offline, maybe we could have a more generic endpoint to get info, so a bit more details could be shared (ex commit index etc) which could allow us use it in the future as a substitute for the commit vote monitor.

mwtian:
> If this gets deployed will the authority service respond with an error? Won't commit sync have a problem with it?

What kind of error or problem do you envision?

consensus/core/src/commit_syncer.rs
mwtian: I removed this trait from observer network for now.

akichidis:
> > If this gets deployed will the authority service respond with an error? Won't commit sync have a problem with it?
> 
> What kind of error or problem do you envision?

I mean will the endpoint reply with an error now or a successful response?

akichidis:
> > > If this gets deployed will the authority service respond with an error? Won't commit sync have a problem with it?
> > 
> > 
> > What kind of error or problem do you envision?
> 
> I mean will the endpoint reply with an error now or a successful response?

Actually @mwtian never mind - just saw that we re-use the `get_latest_rounds` that already exists

consensus/config/src/parameters.rs
```diff
@@ -199,6 +209,14 @@ impl Parameters {
         }
     }
 
+    pub(crate) fn default_commit_sync_request_timeout() -> Duration {
+        Duration::from_secs(10)
+    }
+
+    pub(crate) fn default_commit_sync_probe_timeout() -> Duration {
+        Duration::from_secs(1)
```
akichidis: 1 second might be too tight? I mean we might want to allow some slack especially under some network mild instability conditions. I would suggest putting something between 2-3 seconds

PR #26059, opened 2026-03-31
[move-cli] adds lint command

damirka:
## Description 

- no integration to sui-move yet (similar to docgen)
- default All level, same build config as the rest of commands

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-cli/tests/build_tests/lint_command/args.exp
+++ b/external-crates/move/crates/move-cli/tests/build_tests/lint_command/args.exp
@@ -27,7 +27,6 @@ warning[Lint W04004]: unneeded return
    = This warning can be suppressed with '#[allow(lint(unneeded_return))]' applied to the 'module' or module member ('const', 'fun', or 'struct')
 
 External Command `LC_ALL=C ls -A1`:
-Move.lock
 Move.toml
 args.exp
 args.txt


external-crates/move/crates/move-cli/src/base/lint.rs
```diff
@@ -0,0 +1,33 @@
...
+        path: Option<&Path>,
+        mut config: BuildConfig,
+    ) -> anyhow::Result<()> {
+        let rerooted_path = reroot_path(path)?;
+        let env = find_env::<F>(&rerooted_path, &config)?;
+
+        config.lint_flag.set(LintLevel::All);
+
+        config
+            .compile_package::<F, _>(&rerooted_path, &env, &mut std::io::stdout())
```
tnowacki: Do we need to actually produce bytecode here? I think this does but I forget

external-crates/move/crates/move-cli/tests/build_tests/help/args.exp
```diff
@@ -9,6 +9,7 @@ Commands:
   disassemble  Disassemble the Move bytecode pointed to
   decompile    Decompile Move bytecode into Move source code
   docgen       Generate Rust style documentation for Move packages
+  lint         Run Move linters on the package at `path`. If no path is provided defaults to current directory
```
tnowacki, line 12: Probably need some tests here for what actually happens when you run lint

damirka: Added a CLI test!±

external-crates/move/crates/move-cli/src/base/lint.rs
damirka: There wasn't a way to achieve that with existing set of functions, or I failed to find one, so extended `check` to return diagnostics.

external-crates/move/crates/move-cli/src/base/lint.rs
```diff
@@ -0,0 +1,31 @@
...
+impl Lint {
+    pub async fn execute<F: MoveFlavor>(
+        self,
+        path: Option<&Path>,
+        mut config: BuildConfig,
+    ) -> anyhow::Result<()> {
+        let rerooted_path = reroot_path(path)?;
+        let env = find_env::<F>(&rerooted_path, &config)?;
+
+        config.lint_flag.set(LintLevel::All);
```
akmysti, line 25: Is there a plan to pass configure linter here? There might be lints we don't want to enable in certain cases.

damirka: Only in follow ups, currently this command does what `sui move build --lint` is doing.

external-crates/move/crates/move-cli/tests/build_tests/lint_command/args.exp
```diff
@@ -0,0 +1,18 @@
+Command `lint`:
+BUILDING LintTest
+warning[W09008]: unused function
+  ┌─ ./sources/example.move:8:5
+  │
+8 │ fun t1(): u64 { return EXAMPLE }
```
tnowacki: Why not make it public? This isn't a lint

external-crates/move/crates/move-cli/tests/build_tests/lint_command/args.txt
```diff
@@ -0,0 +1 @@
+lint
```
tnowacki, line 1: can you cat the build directory? what is the output there?

external-crates/move/crates/move-compiler/src/command_line/compiler.rs
```diff
@@ -458,8 +458,9 @@ impl Compiler {
     }
 
     pub fn check_and_report(self) -> anyhow::Result<MappedFiles> {
-        let (files, res) = self.check()?;
-        unwrap_or_report_diagnostics(&files, res);
```
tnowacki, line 462: This seems like a  good change. Should we just change the behavior of `unwrap_or_report_diagnostics`?

external-crates/move/crates/move-cli/tests/build_tests/lint_command/args.txt
damirka: I've added `> ls build` but its output only points that there's no `build` folder.

external-crates/move/crates/move-compiler/src/command_line/compiler.rs
damirka: Are you suggesting that we always `report_warnings` in `unwrap_or_report_diagnostics`?

tzakian: That would make sense to me.

After this change, every callsite to `unwrap_or_report_diagnostics` then subsequently calls `report_warnings` on the returned warnings.

external-crates/move/crates/move-cli/src/base/lint.rs
```diff
@@ -0,0 +1,31 @@
...
+pub struct Lint;
+
+impl Lint {
+    pub async fn execute<F: MoveFlavor>(
+        self,
+        path: Option<&Path>,
+        mut config: BuildConfig,
+    ) -> anyhow::Result<()> {
+        let rerooted_path = reroot_path(path)?;
+        let env = find_env::<F>(&rerooted_path, &config)?;
```
akmysti, line 23: does find_env use LintLevel for anything? in that case `config.lint_flag.set` after this can problematic.

damirka: Nope, it doesn't affect find_env. It's as if --lint flag was passed. Next steps for this command and others are deprecating these flags (eg --doc).

external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
```diff
@@ -151,9 +151,15 @@ pub fn unwrap_or_report_pass_diagnostics<T, Pass>(
     }
 }
 
-pub fn unwrap_or_report_diagnostics<T>(files: &MappedFiles, res: Result<T, Diagnostics>) -> T {
+pub fn unwrap_or_report_diagnostics<T>(
```
tnowacki, line 156: I would probably document that this reports warnings too

external-crates/move/crates/move-cli/tests/build_tests/lint_command/args.txt
```diff
@@ -0,0 +1,3 @@
...
+lint --mode custom
+> test -d /path/to/dir && echo "build/ EXISTS" || echo "build/ NOT FOUND"
```
tnowacki:
```
/path/to/dir
```
I think you have an issue here and need to swap out `/path/to/dir`

tnowacki: And `LC_ALL=C ls -A1` might be more informative

PR #26065, opened 2026-03-31
[move-package-alt] Replace Dependency<T> with DependencyContext

mdgeorge4153:
## Summary

- Introduce a node/edge separation for pinned dependencies: `Pinned` (pub(crate)) represents a package's source location (node), while `PinnedDependency` (pub) pairs a `Pinned` with a `DependencyContext` describing how a parent references it (edge)
- Replace the generic `Dependency<DepInfo>` wrapper with a plain `DependencyContext` struct holding shared dependency metadata (name, environment, override flag, etc.)
- Collapse all methods from `impl Pinned` into `PinnedDependency` — `Pinned` is now a pure data enum
- `Package::dep_for_self` remains `Pinned` (not `PinnedDependency`), since the root package has no parent edge context
- Remove `FetchedDependency` in favor of a free `fetch::fetch` function
- `unfetched_path` and `fetch` now take `chain_id: &EnvironmentID` for future on-chain cache path support

## Test plan

- [x] All existing tests pass
- [ ] Verify no downstream breakage in dependent crates

---

## PR Stack
2. **#26065 — Separate node/edge concepts for pinned dependencies** ← this PR


stefan-mysten (approved):

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -112,19 +120,20 @@ impl PinnedDependencyInfo {
         let mut non_system_deps: Vec<CombinedDependency> = Vec::new();
 
         for dep in deps.into_iter() {
-            if let Combined::System(sys) = &dep.0.dep_info {
+            if let ManifestDependencyInfo::System(sys) = &dep.dep_info {
```
mdgeorge4153: Probably better to use the type alias here.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -112,19 +120,20 @@ impl PinnedDependencyInfo {
...
                             dep: sys.system.clone(),
                             valid: valid_list.to_string(),
                         })?;
-                let file = dep.0.containing_file;
-                let pinned_dep = PinnedDependencyInfo(dep.0.map(|_| {
-                    Pinned::from_lockfile(file, lockfile_dep)
-                        .expect("system dependencies are valid pins")
-                }));
+                let file = dep.context.containing_file;
+                let pinned_dep = PinnedDependencyInfo {
```
mdgeorge4153: I thought we renamed `PinnedDependencyInfo` to `PinnedDependency` to be parallel with `CombinedDependency` and `ResolvedDependency`

external-crates/move/crates/move-package-alt/src/dependency/resolve.rs
```diff
@@ -20,27 +20,28 @@ use tokio::process::Command;
...
-/// externally resolved or system dependencies
-pub type Resolved = ResolverDependencyInfo;
+/// The dep_info type for the resolved stage.
+pub(super) type Resolved = ResolverDependencyInfo;
 
 pub type ResolverResult<T> = Result<T, ResolverError>;
 
 /// A [ResolvedDependency] is like a [CombinedDependency] except that it no longer has
-/// externally resolved dependencies
+/// externally resolved or system dependencies.
```
mdgeorge4153, line 39: Double checking - does `ResolvedDependencyInfo` have a `System` variant? If this is the invariant, it shouldn't.

external-crates/move/crates/move-package-alt/src/dependency/resolve.rs
```diff
@@ -95,9 +96,9 @@ impl ResolvedDependency {
             BTreeMap::new();
 
         for dep in deps.iter() {
-            if let Combined::External(ext) = &dep.0.dep_info {
+            if let ManifestDependencyInfo::External(ext) = &dep.dep_info {
```
mdgeorge4153: Probably keep the type alias here

mdgeorge4153: and in other places

external-crates/move/crates/move-package-alt/src/dependency/mod.rs
```diff
@@ -6,83 +6,41 @@ mod combine;
...
-struct Dependency<DepInfo> {
-    /// The name given to this dependency in the manifest. For modern manifests, this is the same
-    /// as the name used for the package in the source code, while for legacy manifests this name
-    /// may be different (it is still normalized to be a valid identifier but does not correspond
-    /// to the named address).
-    name: PackageName,
-
-    dep_info: DepInfo,
+pub(super) struct DependencyContext {
+    /// The name given to this dependency in the manifest.
```
mdgeorge4153: we should keep the extended doc comments here - the different fields can be somewhat confusing and there is some subtlety around naming for example.

mdgeorge4153 (commented):
Left some comments

external-crates/move/crates/move-package-alt/src/dependency/resolve.rs
mdgeorge4153: Correct -- `ResolverDependencyInfo` only has `Local`, `Git`, and `OnChain` variants (no `System` or `External`). The doc comment on `ResolvedDependency` is accurate.

external-crates/move/crates/move-package-alt/src/dependency/mod.rs
```diff
@@ -6,83 +6,63 @@ mod combine;
...
-struct Dependency<DepInfo> {
+pub(super) struct DependencyContext {
     /// The name given to this dependency in the manifest. For modern manifests, this is the same
     /// as the name used for the package in the source code, while for legacy manifests this name
     /// may be different (it is still normalized to be a valid identifier but does not correspond
     /// to the named address).
-    name: PackageName,
-
-    dep_info: DepInfo,
+    pub(super) name: PackageName,
```
mdgeorge4153, line 30: Question: given that the struct is `pub(super)`, is `pub` preferable here, or should it be `pub(super)`?

external-crates/move/crates/move-package-alt/src/dependency/mod.rs
```diff
@@ -6,83 +6,63 @@ mod combine;
...
-            dep_info: f(self.dep_info),
-            use_environment: self.use_environment,
-            is_override: self.is_override,
-            addresses: self.addresses,
-            containing_file: self.containing_file,
-            rename_from: self.rename_from,
-            modes: self.modes,
+impl DependencyContext {
+    /// Create a minimal context for dependencies constructed outside of the manifest pipeline
+    /// (e.g. from lockfile pins or root dependencies). Uses a placeholder name and empty fields.
```
mdgeorge4153: I think we should have names for deps coming from the lockfile, and also the other fields. Maybe we should just have a special case for the root dependency. I also find the dummy values to be a bit of a code smell - it might be worth making an enum?

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -9,33 +9,36 @@ use sha2::{Digest, Sha256};
...
-pub enum Pinned {
+pub(super) enum Pinned {
     Local(PinnedLocalDependency),
     Git(PinnedGitDependency),
     OnChain(OnChainDepInfo),
     Root(PackagePath),
 }
 
-/// Invariant: if a PinnedDependencyInfo has `dep_info` `Root`, then its `containing_file` is either a
+/// Invariant: if a PinnedDependency has `dep_info` `Root`, then its `containing_file` is either a
```
mdgeorge4153, line 55: This comment is in the wrong place. Also, since `PinnedDependency` is the main public type, I think it should be moved to the top.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -138,56 +151,94 @@ impl PinnedDependencyInfo {
...
+            Pinned::Root(path) => path.path().to_path_buf(),
+        }
+    }
+
+    /// Return an abbreviated string (without braces) showing the dependency
+    pub fn abbreviated(&self) -> String {
+        self.dep_info.abbreviated()
+    }
+
+    /// Returns a hash that uniquely identifies this pinned dependency's source location.
```
mdgeorge4153: What is this for? I don't remember there being a `unique_hash` method.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -138,56 +151,94 @@ impl PinnedDependencyInfo {
...
+    pub fn unique_hash(&self) -> u64 {
+        self.dep_info.unique_hash()
+    }
+
+    /// Create a pinned dependency from a lockfile pin. Uses a minimal `DependencyContext` since
```
mdgeorge4153: We definitely shouldn't be using a stub here.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -138,56 +151,94 @@ impl PinnedDependencyInfo {
...
+    /// Access the inner pinned dependency info. Only visible within the `dependency` module.
+    pub(super) fn pinned(&self) -> &Pinned {
+        &self.dep_info
     }
 }
 
 impl Pinned {
     /// Is the dependency the root?
-    pub fn is_root(&self) -> bool {
+    pub(super) fn is_root(&self) -> bool {
```
mdgeorge4153: I think we can just inline this - it's only used in `PinnedDependency.is_root` correct?

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -198,7 +249,7 @@ impl Pinned {
     /// We do not set the `rename-from` field, since when we are creating the pinned dependency we
     /// don't yet know what the rename-from field  should be. The caller is responsible for calling
     /// [Self::with_rename_from] if they need to establish the rename-from check invariant.
-    pub fn from_lockfile(
+    pub(super) fn from_lockfile(
```
mdgeorge4153: Similarly for other methods in `impl Pinned` - can/should we just move them into the `PinnedDependency` type?

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -287,14 +338,17 @@ impl LocalDepInfo {
                 inner: parent_git.inner.relative_tree(self.local)?,
             }),
             Pinned::Local(parent_local) => Pinned::Local(PinnedLocalDependency {
-                absolute_path_to_package: parent.unfetched_path().join(&self.local).clean(),
+                absolute_path_to_package: parent_local
```
mdgeorge4153: would this be simpler if `pin` took a `PinnedDependency` instead of a `Pin`?

external-crates/move/crates/move-package-alt/src/dependency/mod.rs
mdgeorge4153: [CLAUDE] Since the struct is `pub(super)`, there's no difference in visibility either way. Left as `pub(super)` for consistency with the other dependency types.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Done — `PinnedDependency` is now defined above `Pinned` with the main doc comment.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] `unique_hash()` is pre-existing — it's used in `package_impl.rs` to generate `dummy_addr` for unpublished packages (`OriginalID::from(dep.unique_hash() as u16)`). I just moved it from `Pinned` to `PinnedDependency`.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Done — inlined as `matches!(self.dep_info, Pinned::Root(_))`. The `impl Pinned` block is now empty and removed.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Done — all `Pinned` methods, `From<Pinned>` impls, and `Display for Pinned` collapsed into `PinnedDependency`. `Pinned` is now a pure data enum.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Done — `pin()` now takes `&PinnedDependency` and `&EnvironmentID`, so it can use `parent.unfetched_path(chain_id)` instead of inlined field access.

external-crates/move/crates/move-package-alt/src/dependency/mod.rs
mdgeorge4153: [CLAUDE] Resolved — `dep_for_self` is back to `Pinned` (a node concept). `DependencyContext::stub()` and `PinnedDependency::root()` are removed. `Pinned` is `pub(crate)` with a doc comment explaining the node/edge distinction and a TODO about potentially moving it out of the `dependency` module.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Resolved — no more stubs. `from_lockfile` on `PinnedDependency` now takes `(&Pin, &EnvironmentID)` and stores real `address_override` data. `dep_for_self` is `Pinned` directly, no context wrapper needed.

external-crates/move/crates/move-package-alt/src/dependency/fetch.rs
```diff
@@ -34,29 +30,25 @@ pub enum FetchError {
...
-}
-
-impl From<FetchedDependency> for PackagePath {
-    fn from(value: FetchedDependency) -> Self {
-        value.0.dep_info
-    }
+/// Ensure that the dependency's files are present on the disk and return a path to them.
+/// Assumes that `pinned` is already normalized - paths of any local dependencies are relative
+/// to the current working directory, and local dependencies of git dependencies have been
+/// transformed into git dependencies. `chain_id` is passed through to [Pinned::unfetched_path].
```
mdgeorge4153, line 36: Referring to `Pinned::unfetched_path` isn't appropriate for the doc comment, both because `Pinned` has more restrictive visibility, but more importantly because it describes implementation rather than interface. Instead, we should say that `chain_id` is used to determine the location for on-chain dependencies. It's okay to refer to `Pinned::unfetched_path` for more information.

amnn (commented):
@mdgeorge4153, purely for my own education, can you share the motivation for this change? I.e. what does this help achieve for on-chain deps or otherwise why was this change worth doing. 

For future, this is usually good context to include in the PR description or in commit messages -- the description here reads more like a summary of what changed, which can be useful but is not as interesting as the why.

PR #26069, opened 2026-03-31
Remove backwards compat support for non-TLS connections to validator server

aschran:
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-http/src/config.rs
+++ b/crates/sui-http/src/config.rs
@@ -24,7 +24,6 @@ pub struct Config {
     pub(crate) max_connection_age: Option<Duration>,
     pub(crate) tls_handshake_timeout: Duration,
     pub(crate) max_pending_connections: usize,
-    pub(crate) allow_insecure: bool,
 }
 
 impl Default for Config {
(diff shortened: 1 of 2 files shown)


crates/sui-http/src/config.rs
```diff
@@ -187,14 +193,34 @@ impl Config {
...
+        Config {
+            max_pending_connections: max,
+            ..self
+        }
+    }
+
+    /// Previously allowed accepting insecure connections when a tls_config is provided.
+    /// This is no longer supported and must be set to `false`.
+    #[deprecated(note = "allow_insecure is no longer supported and must be set to false")]
     pub fn allow_insecure(self, allow_insecure: bool) -> Self {
```
bmwill: why not just remove this?

aschran: I thought it would be friendlier to crash with an error that it's not supported, vs making configs fail to parse

bmwill: afaik this doesn't live in any configs anywhere

aschran: removed

PR #26071, opened 2026-03-31
[move] Thread MoveFlavor instance through package pipeline

mdgeorge4153:
## Summary

- Change `MoveFlavor` trait methods from associated functions to instance methods (`&self`)
- Make `system_deps` and `implicit_dependencies` async to support future RPC-based queries
- Thread `Arc<F>` through the package loading pipeline from `PackageLoader::load()` down to dependency pinning and legacy manifest parsing

This is a prerequisite for on-chain bytecode dependencies (which need the flavor instance to fetch packages via RPC) and for fixing DVX-1814 (network-aware system dependency resolution).

## Test plan

- [ ] CI passes (all existing tests should pass unchanged — this is a pure refactor with no behavior change)
- [ ] `cargo check` passes across the full workspace


mdgeorge4153:
Splitting into stacked PRs for easier review.

PR #26077, opened 2026-03-31
pick(26061) fix(graphql): fix Display v2 scoping issue

amnn:
## Description

Fixes an issue where GraphQL would bound the version of the Display v2 format for an object by the version of the type being displayed, which would cause valid Display v2 formats to be discarded.

## Test plan

Introduced a regression test:

```
$ cargo nextest run            \
  --test transactional_tests   \
  -- graphql/objects/display/display_v2_scope
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/objects/display/display_v2_scope.move
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/objects/display/display_v2_scope.move
@@ -1,0 +1,60 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//# init --accounts A --addresses test=0x0 --simulator --protocol-version 117
+
+//# publish --sender A
+module test::mod {
+  use std::string::String;
+  use sui::display_registry::DisplayCap;
+  use sui::display_registry::DisplayRegistry;
+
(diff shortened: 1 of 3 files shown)


crates/sui-indexer-alt-e2e-tests/tests/graphql/objects/display/display_v2_scope.move
```diff
@@ -0,0 +1,60 @@
...
+
+//# programmable --sender A --inputs object(0xd) @A
+//> 0: test::mod::display(Input(0));
+//> 1: TransferObjects([Result(0)], Input(1))
+
+//# create-checkpoint
+
+//# run-graphql
+{
+  object(address: "@{obj_2_0}") {
```
wlmyng, line 50: Isn't this going to fetch the latest obj version though? I'm not sure what the test is testing for

amnn: It fetches the latest version of the object, but that object's version is less than the version of the Display object, because the Display object was created by a later transaction that used the same gas coin.

PR #26080, opened 2026-03-31
sui-indexer-alt-framework: reduce copied and pasted test code

evan-wall-mysten:
It is hard to see what these test cases are testing because most of the code is copied and pasted boilerplate to initialize the indexer and database. This change adds helper methods to reduce this boiler plate.

---

Added helper methods:
  - create_test_indexer — creates an Indexer<MockStore> with a temp dir and default client/ingestion config. Takes an optional ingestion_data: Option<(u64, u64)> to generate synthetic checkpoint
  data.
  - set_committer_watermark — calls conn.set_committer_watermark with only checkpoint_hi_inclusive set, defaulting the rest.
  - add_concurrent — registers a concurrent pipeline with default config on an indexer.
  - add_sequential — registers a sequential pipeline with default config on an indexer.
  - test_init_watermark — (pre-existing) sets up an indexer with a single pipeline and returns its watermarks.

Added helper macros:
  - assert_out_of_order — asserts the total_watermarks_out_of_order metric value for a given pipeline name.

Additional changes:
  - Removed test_indexer_ingestion_existing_watermarks_ignore_first_checkpoint — it was redundant with the no-first-checkpoint variant (identical watermarks, identical assertions; the
  "first_checkpoint ignored" behavior is already unit-tested by test_next_checkpoint_ignore_first_checkpoint).
  - Simplified test_multiple_sequential_pipelines_next_checkpoint — replaced two verbose set_committer_watermark calls (with all explicit fields) with set_watermark, since only
  checkpoint_hi_inclusive matters for the test's assertions.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/lib.rs
+++ b/crates/sui-indexer-alt-framework/src/lib.rs
@@ -602,6 +602,93 @@ mod tests {
     test_pipeline!(SequentialHandler, "sequential_handler");
     test_pipeline!(MockCheckpointSequenceNumberHandler, "test");
 
+    /// If `ingestion_data` is `Some((num_checkpoints, checkpoint_size))`, synthetic ingestion
+    /// data will be generated in the temp directory before creating the indexer.
+    async fn create_test_indexer(
+        store: MockStore,
+        indexer_args: IndexerArgs,
+        registry: &Registry,
+        ingestion_data: Option<(u64, u64)>,
+    ) -> (Indexer<MockStore>, tempfile::TempDir) {
+        let temp_dir = tempfile::tempdir().unwrap();
+        if let Some((num_checkpoints, checkpoint_size)) = ingestion_data {
+            synthetic_ingestion::generate_ingestion(synthetic_ingestion::Config {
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-framework/src/lib.rs
```diff
@@ -602,6 +602,91 @@ mod tests {
...
+        indexer: &mut Indexer<MockStore>,
+        handler: H,
+    ) {
+        indexer
+            .sequential_pipeline(handler, SequentialConfig::default())
+            .await
+            .unwrap();
+    }
+
+    fn assert_out_of_order(metrics: &IndexerMetrics, pipeline: &str, expected: u64) {
```
amnn: Use macros to abstract assertions, to preserve the function (test) name and line number in the backtrace output (otherwise all the test failures will point to this function).

PR #26085, opened 2026-03-31
[sui-framework][std] Added additional `option` macros

tnowacki:
## Description 

- Added `std::option::map_mut` 
  - We originally felt this might be confusing, but looking at it now, I don't feel the same way, especially as we add more `mut` functions 
- Added `std::option::is_none_or`
- Added `std::option::fold`
- Added `std::option::fold_ref`
- Added `std::option::fold_mut`

## Test plan 

- New tests 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-framework/packages/move-stdlib/sources/option.move
+++ b/crates/sui-framework/packages/move-stdlib/sources/option.move
@@ -232,7 +232,7 @@ public macro fun is_some_and<$T>($o: &Option<$T>, $f: |&$T| -> bool): bool {
     o.is_some() && $f(o.borrow())
 }
 
-/// Return `true` if the value is `None`, or the predicate `f` returns `true` for the contained
+/// Return `true` if the value is `None`, or if the predicate `f` returns `true` for the contained
 /// value.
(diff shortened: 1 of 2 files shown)


crates/sui-framework/packages/move-stdlib/sources/option.move
```diff
@@ -225,6 +232,41 @@ public macro fun is_some_and<$T>($o: &Option<$T>, $f: |&$T| -> bool): bool {
     o.is_some() && $f(o.borrow())
 }
 
+/// Return `true` if the value is `None`, or the predicate `f` returns `true` for the contained
```
awelc: It's probably just me but without the `if`, it took me longer to parse the comment than it probably should have
```suggestion
/// Return `true` if the value is `None`, or if the predicate `f` returns `true` for the contained
```

tnowacki:
> If we open the floodgates of `map_mut`, shall we do the same for vector?

At some point, yeah

PR #26094, opened 2026-04-01
[graphql-stream] Implement checkpoint subscription streaming [2/n]

tpham-mysten:
## Description 

- Wire `CheckpointStreamTask` to connect to a fullnode's gRPC `SubscribeCheckpoints` endpoint, deserialize incoming proto checkpoints, and broadcast them to GraphQL subscription resolvers via a broadcast channel
- Implement the checkpoints subscription resolver — yields Checkpoint objects with pre-processed streamed data, resolving fields (summary, contents, signature) directly from memory instead of database fetches
- Add `Scope::for_streamed_checkpoint` with `checkpoint_viewed_at: None` since streamed data isn't bounded by indexed state
- Add e2e tests using a real TestCluster with gRPC streaming to verify sequential checkpoint delivery and field correctness via insta snapshot redactions

Not yet supported:
  - Streaming from a particular checkpoint (start cursor) — will be tackled in Phase 2 alongside backfilling
  - Proper handling of Connection and stream errors — will be tackled in Phase 2 alongside backfilling
  - Nested subscription fields (transactions, events, etc.) — will be tackled in a follow-up PR

## Test plan 

- `cargo nextest run -p sui-indexer-alt-e2e-tests --test graphql_subscription_tests`
- Local Graphql server testing and verify checkpoints are streamed correctly

---

## Stack

- #26019
- #26094
- #26117
- #26170
- #26194

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/Cargo.toml
+++ b/crates/sui-indexer-alt-e2e-tests/Cargo.toml
@@ -6,6 +6,9 @@ license = "Apache-2.0"
 publish = false
 edition = "2024"
 
+[features]
+staging = ["sui-indexer-alt-graphql/staging"]
+
 [lints]
 workspace = true
 
@@ -41,7 +44,7 @@ sui-kvstore.workspace = true
 sui-indexer-alt-consistent-api.workspace = true
 sui-indexer-alt-consistent-store.workspace = true
 sui-indexer-alt-framework = { workspace = true, features = ["postgres", "cluster"] }
-sui-indexer-alt-graphql = { workspace = true, features = ["staging"] }
+sui-indexer-alt-graphql.workspace = true
 sui-indexer-alt-jsonrpc.workspace = true
 sui-indexer-alt-reader.workspace = true
 sui-indexer-alt-schema.workspace = true
@@ -66,7 +69,7 @@ insta.workspace = true
 jsonrpsee.workspace = true
 telemetry-subscribers.workspace = true
 tokio-stream.workspace = true
-tokio-tungstenite = { workspace = true, features = ["native-tls"] }
+tokio-tungstenite = { workspace = true, features = ["rustls-tls-webpki-roots"] }
 tonic.workspace = true
 futures.workspace = true
 
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/main.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/main.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+#![cfg(feature = "staging")]
+
 mod testing;
 
 mod checkpoint_subscription;
--- a/crates/sui-indexer-alt-graphql/src/api/subscription.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/subscription.rs
@@ -26,10 +26,10 @@ impl Subscription {
         &self,
         ctx: &Context<'_>,
     ) -> Result<impl futures::Stream<Item = Result<Checkpoint, RpcError>>, RpcError> {
-        let package_store: Arc<PackageCache> = ctx.data::<Arc<PackageCache>>()?.clone();
+        let package_store = ctx.data::<Arc<PackageCache>>()?.clone();
         let limits: &Limits = ctx.data()?;
         let resolver_limits = limits.package_resolver();
-        let broadcaster: &CheckpointBroadcaster = ctx.data::<CheckpointBroadcaster>()?;
+        let broadcaster: &CheckpointBroadcaster = ctx.data()?;
 
         Ok(async_stream::stream! {
             let mut receiver = broadcaster.resubscribe();
--- a/crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
+++ b/crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
@@ -27,6 +27,7 @@ use crate::config::SubscriptionConfig;
 
 use super::processed_checkpoint::ProcessedCheckpoint;
 
+// TODO: Make these configurable via SubscriptionConfig.
 const MAX_GRPC_MESSAGE_SIZE_BYTES: usize = 128 * 1024 * 1024;
 const CONNECTION_TIMEOUT: Duration = Duration::from_secs(30);
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3938,16 +3938,6 @@ dependencies = [
 
 [[package]]
 name = "core-foundation"
-version = "0.9.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
-dependencies = [
- "core-foundation-sys",
- "libc",
-]
-
-[[package]]
-name = "core-foundation"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b55271e5c8c478ad3f38ad24ef34923091e0548492a266d19b3c0b4d82574c63"
@@ -5931,21 +5921,6 @@ name = "foldhash"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
-
-[[package]]
-name = "foreign-types"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
-dependencies = [
- "foreign-types-shared",
-]
-
-[[package]]
-name = "foreign-types-shared"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
 
 [[package]]
 name = "form_urlencoded"
@@ -9794,23 +9769,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "native-tls"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87de3442987e9dbec73158d5c715e7ad9072fda936bb03d19d7fa10e00520f0e"
-dependencies = [
- "libc",
- "log",
- "openssl",
- "openssl-probe",
- "openssl-sys",
- "schannel",
- "security-framework 2.11.1",
- "security-framework-sys",
- "tempfile",
-]
-
-[[package]]
 name = "neptune"
 version = "13.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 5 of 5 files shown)


crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -29,4 +58,84 @@ impl CheckpointStreamTask {
...
     }
+
+    /// Connect to the fullnode's gRPC SubscribeCheckpoints endpoint.
+    async fn connect(&self) -> anyhow::Result<Streaming<SubscribeCheckpointsResponse>> {
+        let endpoint = Endpoint::from(self.uri.clone()).connect_timeout(CONNECTION_TIMEOUT);
+
+        let mut client = SubscriptionServiceClient::connect(endpoint)
+            .await
+            .context("Failed to connect to checkpoint stream")?
+            .max_decoding_message_size(MAX_GRPC_MESSAGE_SIZE_BYTES);
```
evan-wall-mysten, line 79: Is `max_encoding_message_size <= MAX_GRPC_MESSAGE_SIZE_BYTES` in the upstream server? If not then this client will reject messages that the server is sending unnecessarily.

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -16,10 +24,46 @@ impl Subscription {
...
+        ctx: &Context<'_>,
     ) -> Result<impl futures::Stream<Item = Result<Checkpoint, RpcError>>, RpcError> {
-        Err::<futures::stream::Empty<_>, _>(RpcError::FeatureUnavailable {
-            what: "Checkpoint Subscriptions",
+        let package_store: Arc<PackageCache> = ctx.data::<Arc<PackageCache>>()?.clone();
+        let limits: &Limits = ctx.data()?;
+        let resolver_limits = limits.package_resolver();
+        let broadcaster: CheckpointBroadcaster = ctx.data::<CheckpointBroadcaster>()?.clone();
+
+        Ok(async_stream::stream! {
```
evan-wall-mysten, line 34: Will some of this code be reusable for the other top level streams?

evan-wall-mysten:
> Nested subscription fields (transactions, events, etc.) — will be tackled in a follow-up PR

Is that PR going to validate that a single item of the top level stream where each nested pagable field contains a min of 1 item will fit within the max item size? Also will it handle resizing the nested pagable fields?

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: This matches the existing pattern in sui-indexer-alt-framework [streaming_client.rs](https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs#L70), which uses the same 128 MB limit for SubscriptionServiceClient.

The upstream server doesn't set max_encoding_message_size (defaults to usize::MAX), so this is a client-side safety bound rather than a server-matching limit

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Yes! the broadcast receive loop with lagged/closed error handling and scope creation would be shared across any top-level stream. I will refactor them later when implementing these logic

tpham-mysten (commented):
Thanks @evan-wall-mysten!

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -16,10 +24,46 @@ impl Subscription {
...
+            let mut receiver = broadcaster.subscribe();
+            loop {
+                match receiver.recv().await {
```
Danny-Devs, line 37: Noticed a subtle channel lifetime interaction: if the gRPC stream ends cleanly (`stream.next()` returns `None` in `CheckpointStreamTask::run()`), the task's `broadcast::Sender` is dropped — but the clone stored in the GraphQL context via `rpc.data(task.broadcaster())` keeps the channel alive. Since `RecvError::Closed` requires all senders to be dropped, `recv().await` here would block indefinitely. The error path is well-handled by the Service framework, but the clean-exit path seems like it could fall through silently.

Since there's only one producer in this system, would it make sense to store a `broadcast::Receiver` in the context instead of a `Sender`, and use `resubscribe()` to create per-subscriber receivers? That way the stream task holds the only `Sender`, and `Closed` fires naturally when it dies. `Receiver` is `Send + Sync`, and `resubscribe()` calls the same internal `new_receiver()` as `subscribe()` — so behavior would be identical.

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/testing/harness.rs
```diff
@@ -0,0 +1,152 @@
...
+
+pub struct SubscriptionStream {
+    stream: futures::stream::SplitStream<
+        tokio_tungstenite::WebSocketStream<
+            tokio_tungstenite::MaybeTlsStream<tokio::net::TcpStream>,
+        >,
+    >,
+}
+
+impl SubscriptionStream {
```
amnn: nit: FWIW, these operations already exist as stream modifiers, but in `tokio_stream`, rather than `futures::stream`. You can

- apply a [`timeout`](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.timeout) on a stream, 
- [`map`](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.map) its contents in order to get access to the payload,
- [`take`](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.take) a fixed quantity of items,
- [`collect`](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.collect) them into a `Vec<_>`.

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/checkpoint_subscription.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use sui_macros::sim_test;
+use test_cluster::TestClusterBuilder;
+
+use super::testing::*;
```
amnn: nit: import styling (no `super`, no wildcards).

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/main.rs
amnn, line 1: Does this structure prevent Rust's test runner from treating the `testing` module as  an integration test? My understanding is that every `*.rs` file inside the `tests` directory is compiled on its own and treated like a standalone test file, so this set-up where helper modules are in the `test` directory can cause issues.

That is why `sui-indexer-alt-e2e-tests` exists as its own crate -- its library target is supposed to contain all the helper types, which are used by individual `*.rs` test binaries in the `tests` directory.

crates/sui-indexer-alt-e2e-tests/Cargo.toml
```diff
@@ -39,7 +39,7 @@ sui-kvstore.workspace = true
 sui-indexer-alt-consistent-api.workspace = true
 sui-indexer-alt-consistent-store.workspace = true
 sui-indexer-alt-framework = { workspace = true, features = ["postgres", "cluster"] }
-sui-indexer-alt-graphql.workspace = true
+sui-indexer-alt-graphql = { workspace = true, features = ["staging"] }
```
amnn: I would expect this to mess with the other tests we have, which are parametrized on this feature (e.g. the transactional tests have a normal mode and a staging mode).

The right way to set-up a staging test is to gate the inclusion of the test on the staging featuer flag, but not force the dependency to include staging, so that this crate's tests can be run with staging on (in which case it will include your tests) or off.

crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
amnn, line 1: It would be nice if we could use the same pattern everywhere for making fresh data available to GraphQL, which is to stash it in the `Scope` and check for it in the relevant `*Content` type's `fetch` function.

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -16,10 +24,46 @@ impl Subscription {
     /// This subscription is not yet available for use.
     async fn checkpoints(
         &self,
-        _ctx: &Context<'_>,
+        ctx: &Context<'_>,
     ) -> Result<impl futures::Stream<Item = Result<Checkpoint, RpcError>>, RpcError> {
-        Err::<futures::stream::Empty<_>, _>(RpcError::FeatureUnavailable {
-            what: "Checkpoint Subscriptions",
+        let package_store: Arc<PackageCache> = ctx.data::<Arc<PackageCache>>()?.clone();
```
amnn: Is the double type annotation necessary? (also below)

```suggestion
        let package_store: Arc<PackageCache> = ctx.data()?.clone();
```

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -2,20 +2,49 @@
...
+use tonic::transport::Endpoint;
 use tonic::transport::Uri;
+use tracing::info;
 
 use crate::config::SubscriptionConfig;
 
 use super::processed_checkpoint::ProcessedCheckpoint;
 
+const MAX_GRPC_MESSAGE_SIZE_BYTES: usize = 128 * 1024 * 1024;
+const CONNECTION_TIMEOUT: Duration = Duration::from_secs(30);
```
amnn, line 32: Can we make these configs? (Can be a follow-up -- just add a TODO in that case).

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Thanks for catching this, @Danny-Devs! I have applied the fix - storing a Receiver in the context so subscribers get `RecvError::Closed` when the stream task drops its Sender.

This is the right behavior for now since we terminate `CheckpointStreamTask` on errors. Worth noting that this may evolve in Phase 2 when we add proper error handling and backfilling. In that phase, we likely won't want to terminate the task when gRPC subscription fails — checkpoints in the blockchain should keep coming, and a gRPC failure is a transient infrastructure issue, not a terminal state. The task could fall back to other sources (e.g. remote checkpoint stores) while the gRPC connection recovers.

In that world, hanging (waiting for recovery) may actually be preferable to closing the channel — subscribers see a delay in upcoming checkpoints rather than a disconnect that forces them to reconnect and retry. The framework can auto-heal when the gRPC subscription comes back, which is a nicer experience for clients

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/testing/harness.rs
tpham-mysten: Sounds good! Done

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/checkpoint_subscription.rs
tpham-mysten: Done

crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
tpham-mysten: This is similar to the pattern we discussed with `ExecutedTransaction` a few months ago. Not sure if it's still relevant here, but if the data makes sense for all nested fields, we will put it in Scope works. In this case, checkpoint-level data can contain all objects and versions across all transactions, and exposing it via Scope to each transaction's resolvers could be a room for future bugs (e.g. cross-transaction object access). That's why I keep `streamed_data` on Checkpoint and only put per-transaction data in Scope.

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/main.rs
tpham-mysten: The main.rs entry point makes this a single test binary — testing/ is a module within it, not a separate integration test. Rust only creates separate test binaries from top-level *.rs files or directories with main.rs.

That said, if the helpers become useful across multiple test binaries, I'll move them to the library target.

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Ah nope. Updated

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Added a TODO as we can revisit them all together when doing backfilling

crates/sui-indexer-alt-e2e-tests/Cargo.toml
tpham-mysten: Sounds good. I have added it.

tpham-mysten:
@evan-wall-mysten Sorry I missed one of your question

> Nested subscription fields (transactions, events, etc.) — will be tackled in a follow-up PR
>> Is that PR going to validate that a single item of the top level stream where each nested pagable field contains a min of 1 item will fit within the max item size? Also will it handle resizing the nested pagable fields?

I'm not sure what "max item size" refers to? Is it WebSocket frame limit? GraphQL response size? gRPC message size?

> Also will it handle resizing the nested pagable fields?

I have not thought about this. Right now, the pagination wil be quite the same as Query API, and you can take a look on https://github.com/MystenLabs/sui/pull/26117, but worth considering as we move toward production readiness

crates/sui-indexer-alt-graphql/src/api/subscription.rs
Danny-Devs: Glad to help!

PR #26095, opened 2026-04-01
[move] disallow orphaned tables 

tzakian:
## Description 

Bit of cleanliness enforcement -- don't allow orphaned jumps. 

## Test plan 

CI + some local testing

---

## Release notes

- [X] Protocol: Shifts 120 to 121, and adds an additional check in 120 to enforce cleanliness around jumps.  
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority/snapshots/sui_core__authority__execution_time_estimator__tests__execution_time_observer_v121.snap
+++ b/crates/sui-core/src/authority/snapshots/sui_core__authority__execution_time_estimator__tests__execution_time_observer_v121.snap
@@ -1,0 +1,159 @@
+---
+source: crates/sui-core/src/authority/execution_time_estimator.rs
+expression: snapshot_data
+---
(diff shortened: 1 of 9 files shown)


crates/sui-protocol-config/src/lib.rs
```diff
@@ -4765,6 +4773,9 @@ impl ProtocolConfig {
                     cfg.transfer_receive_object_type_cost_per_byte = Some(2);
                 }
                 120 => {
+                    cfg.feature_flags.disallow_jump_orphans = true;
+                }
+                121 => {
                     // Re-enable unpaid amplification deferral protection (testnet + devnet)
```
tnowacki, line 4779: Why do we need two different protocol configs like this??

PR #26096, opened 2026-04-01
sui-indexer-alt-framework: treat ingestion client deser errors as transient across impls

evan-wall-mysten:
## Description 

Previously the `rpc_client` and `store_client` ingestion client impls had different behavior for retrying deserialization errors and metering the checkpoint bytes.

Changes in this PR:

Commit 1: 
- upgrade sui-rpc to e494a36a76a0aab8c5d66d5557995faee5c1fb09 to pull in changes from https://github.com/MystenLabs/sui-rust-sdk/pull/237

Commit 2:
- port https://github.com/MystenLabs/sui-rust-sdk/pull/234 to this repo which uses new API from commit 1
- CheckpointData: Converted from an enum with Raw/Checkpoint variants to a Checkpoint. Deserialization now happens in client implementations rather than in
  IngestionClient.
- rpc_client.rs: Uses `ByteCountMakeCallbackHandler` to populate num_bytes.
- store_client.rs: Decodes raw bytes into a Checkpoint before returning, with deserialization failures as transient errors.
- ingestion_client.rs: Removed Raw-path decode logic. Simplified tests with helper methods, consolidated redundant tests, added assertions for ingestion metrics.

## Test plan 

Updated unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Remove `CheckpointData::Raw` raw variant. Treat all ingestion errors as transient (retryable). Populate `total_ingested_bytes` during gRPC ingestion.


crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
```diff
@@ -59,13 +60,17 @@ impl IngestionClientTrait for RpcClient {
             })?
             .into_inner();
 
+        let num_bytes = response.encoded_len() as u64;
```
evan-wall-mysten: `encoded_len()` is faster than deserialization, but needs to look at each field in the response to calculate the size - adding an extension to the response that keeps track of response bytes https://github.com/MystenLabs/sui-rust-sdk/pull/234

bmwill: that unfortunately isn't sufficient under a number of situations, specifically if compression is involved, that middleware will result in an inaccurate byte accounting

amnn: Is it safe to treat this value as an accurate representation of the uncompressed size? If so, as long as we are clear about that (maybe updating the ingestion metric description?), it still seems useful to track that information (better to have it than not).

We should separately have an idea of ingress/egress for the indexer, but it would seem like a better fit to track that as aggregate bytes coming into/out of the instance, so it's not something we would track in the indexing framework.

nickvikeras:
What is the reason for marking a deserialization error as transient? If we are just trying to make all the ingestion paths have consistent behavior, would it make more sense to mark them all as permanent? Or is there some case where we expect it to succeed on a retry?

crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
bmwill: I was specifically talking about the middleware that is being added in https://github.com/MystenLabs/sui-rust-sdk/pull/234. What is done as-is in this PR, while maybe being slightly inefficient, does yield the correct byte size of the uncompressed payload.

bmwill:
> What is the reason for marking a deserialization error as transient? If we are just trying to make all the ingestion paths have consistent behavior, would it make more sense to mark them all as permanent? Or is there some case where we expect it to succeed on a retry?

to add on to this, if we have a deser error, we really don't have many options depending on the semantics you want:
- gracefully degrade and omit the part of the checkpoint we cant understand (this is not good because it can lead to dataloss
- treat it as transient and retry. This doesn't make much sense because if the payload is "correct" (and the place you fetched it from isn't malicious) then you'll never succeed unless the software itself is updated. If we have multiple upstreams we could try and fetch from others to see if we end up with the same effect. ultimately we'll require raising an alert to have someone come and update the binary.

evan-wall-mysten:
> > What is the reason for marking a deserialization error as transient? If we are just trying to make all the ingestion paths have consistent behavior, would it make more sense to mark them all as permanent? Or is there some case where we expect it to succeed on a retry?
> 
> to add on to this, if we have a deser error, we really don't have many options depending on the semantics you want:
> 
> * gracefully degrade and omit the part of the checkpoint we cant understand (this is not good because it can lead to dataloss
> * treat it as transient and retry. This doesn't make much sense because if the payload is "correct" (and the place you fetched it from isn't malicious) then you'll never succeed unless the software itself is updated. If we have multiple upstreams we could try and fetch from others to see if we end up with the same effect. ultimately we'll require raising an alert to have someone come and update the binary.

I think the retries only help in the case that the upstream store has an invalid item in it because ingestion will be stuck until it can be corrected there. If the indexer has a bug in it then it has to be restarted anyways which is similar to crashing.

amnn:
> What is the reason for marking a deserialization error as transient?

The general principle is that the indexer avoids panicking unless it is unavoidable (meaning not panicking would result in committing some unknown or known bad state), to preserve forward progress.

If there is a deserialization issue with a particular checkpoint, that issue is local to that checkpoint. We won't try to make do with the checkpoint (the indexer is forced to retry), but other checkpoints can progress through the indexer fine, so it just signals the issue through metrics/alerts/log messages.

As @evan-wall-mysten mentioned, if it is an upstream issue (corrupt file written to a bucket, or bug in the upstream service), it then suffices to fix that upstream service and none of the forward progress on other checkpoints is lost, but if it's a bug in the indexer, then yes, you need to restart. Always panicking prevents us from preserving forward progress in the former case.

crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
evan-wall-mysten: The metric that this powers is defined as
```
            total_ingested_bytes: register_int_counter_with_registry!(
                name("total_ingested_bytes"),
                "Total number of bytes fetched from the remote store, this metric will not \
                be updated when data are fetched over gRPC.",
                registry,
            )
```

I'll update the description because gRPC is no longer excluded, but if compression is used does that mean fewer bytes are fetched?

nickvikeras:
> > What is the reason for marking a deserialization error as transient?
> 
> The general principle is that the indexer avoids panicking unless it is unavoidable (meaning not panicking would result in committing some unknown or known bad state), to preserve forward progress.
> 
> If there is a deserialization issue with a particular checkpoint, that issue is local to that checkpoint. We won't try to make do with the checkpoint (the indexer is forced to retry), but other checkpoints can progress through the indexer fine, so it just signals the issue through metrics/alerts/log messages.
> 
> As @evan-wall-mysten mentioned, if it is an upstream issue (corrupt file written to a bucket, or bug in the upstream service), it then suffices to fix that upstream service and none of the forward progress on other checkpoints is lost, but if it's a bug in the indexer, then yes, you need to restart. Always panicking prevents us from preserving forward progress in the former case.

Got it makes a lot of sense, just wasn't obvious to me. Might be worth documenting somewhere for posterity.

bmwill:
>Always panicking prevents us from preserving forward progress in the former case.

Yeah i wasn't necessarily suggesting we panic in this instance (that would be bad).

crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
amnn: > if compression is used does that mean fewer bytes are fetched?

Yes, based on the discussion we had out-of-band, the compressed size would be what we want here.

bmwill: Then you should be able to read this header for that info https://docs.rs/http/latest/http/header/constant.CONTENT_LENGTH.html

evan-wall-mysten: It looks like this header is not present
```
grpcurl -v \
  -d '{
    "read_mask": {
      "paths": [
        "summary.bcs",
        "signature",
        "contents.bcs",
        "transactions.transaction.bcs",
        "transactions.effects.bcs",
        "transactions.effects.unchanged_loaded_runtime_objects",
        "transactions.events.bcs",
        "objects.objects.bcs"
      ]
    }
  }' \
  fullnode.mainnet.sui.io:443 \
  sui.rpc.v2.LedgerService/GetCheckpoint

Resolved method descriptor:
rpc GetCheckpoint ( .sui.rpc.v2.GetCheckpointRequest ) returns ( .sui.rpc.v2.GetCheckpointResponse );

Request metadata to send:
(empty)

Response headers received:
access-control-allow-origin: *
access-control-expose-headers: *
alt-svc: h3=":443"; ma=2592000
content-type: application/grpc
date: Fri, 03 Apr 2026 14:01:15 GMT
server: sui-node/1.68.1-3c0f387ebb40
server-timing: finish_request;dur=1
vary: origin, access-control-request-method, access-control-request-headers
via: 1.1 google
x-sui-chain: mainnet
x-sui-chain-id: 35834a8a
x-sui-checkpoint-height: 261090353
x-sui-epoch: 1086
x-sui-lowest-available-checkpoint: 256182448
x-sui-lowest-available-checkpoint-objects: 256182448
x-sui-timestamp: 2026-04-03T14:01:14.670Z
x-sui-timestamp-ms: 1775224874670
grpcurl -v \
  -d '{
    "read_mask": {
      "paths": [
        "summary.bcs",
        "signature",
        "contents.bcs",
        "transactions.transaction.bcs",
        "transactions.effects.bcs",
        "transactions.effects.unchanged_loaded_runtime_objects",
        "transactions.events.bcs",
        "objects.objects.bcs"
      ]
    }
  }' \
  fullnode.mainnet.sui.io:443 \
  sui.rpc.v2.LedgerService/GetCheckpoint

Resolved method descriptor:
rpc GetCheckpoint ( .sui.rpc.v2.GetCheckpointRequest ) returns ( .sui.rpc.v2.GetCheckpointResponse );

Request metadata to send:
(empty)

Response headers received:
access-control-allow-origin: *
access-control-expose-headers: *
alt-svc: h3=":443"; ma=2592000
content-type: application/grpc
date: Fri, 03 Apr 2026 14:01:15 GMT
server: sui-node/1.68.1-3c0f387ebb40
server-timing: finish_request;dur=1
vary: origin, access-control-request-method, access-control-request-headers
via: 1.1 google
x-sui-chain: mainnet
x-sui-chain-id: 35834a8a
x-sui-checkpoint-height: 261090353
x-sui-epoch: 1086
x-sui-lowest-available-checkpoint: 256182448
x-sui-lowest-available-checkpoint-objects: 256182448
x-sui-timestamp: 2026-04-03T14:01:14.670Z
x-sui-timestamp-ms: 1775224874670
```

---

And we should not depend on it according to claude:

> Can `Content-Length` Be Used in gRPC over HTTP/2?

1. **Spec doesn't define `Content-Length`** — The gRPC protocol specifies `content-type`, `grpc-encoding`, `grpc-status`, etc., but not `Content-Length`. ([source](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md))

2. **gRPC has its own framing** — Each message is preceded by a 1-byte compressed flag and 4-byte message length, making HTTP-level content length redundant. ([source](https://grpc.github.io/grpc/core/md_doc__p_r_o_t_o_c_o_l-_h_t_t_p2.html))

3. **DATA frames ≠ message boundaries** — The spec states DATA frame boundaries have no relation to message boundaries and implementations should not assume alignment. ([source](https://chromium.googlesource.com/external/github.com/grpc/grpc/+/HEAD/doc/PROTOCOL-HTTP2.md))

4. **Proxies may strip it** — Envoy's gRPC bridge removes `content-length` because it may conflict with the gRPC frame size. ([source](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/grpc_http1_bridge_filter))

bmwill: 🫠 sounds good. Looks like we do need to handle being able to collect this info in a middleware then.

I'm not a big fan of the approach in https://github.com/MystenLabs/sui-rust-sdk/pull/234 because this middleware is given special treatment (its built in) with no mechanism for supplying other middlewares that we may want/need in the future. I think the correct approach would be to introduce the ability to have users provide arbitrary middleware which would be applied to any and all requests sent via the client. 

I anticipated needing this and have something half-done locally. Let me spend a bit this morning to clean it up and i can share it. This would mean that the middleware you added in that PR above would instead be one defined and set here in the framework itself.

evan-wall-mysten: Ok yeah that sounds better, let me know when it's ready.

bmwill: https://github.com/MystenLabs/sui-rust-sdk/pull/237

evan-wall-mysten: Thanks, I updated `sui-rust-sdk` to the version from your merged PR, ported the change from https://github.com/MystenLabs/sui-rust-sdk/pull/234 to here, and replaced `encoded_len()`.

crates/sui-indexer-alt-consistent-store/src/restore/formal_snapshot.rs
```diff
@@ -146,14 +147,16 @@ impl FormalSnapshot {
...
+        // transaction. We only use `end_of_epoch_checkpoints` here, which never increments the
+        // byte counter, so an unregistered counter is fine.
         let client = StoreIngestionClient::new(
             HttpBuilder::new()
                 .with_url(snapshot_args.remote_store_url.to_string())
                 .with_client_options(ClientOptions::new().with_allow_http(true))
                 .build()
                 .map(Arc::new)
                 .context("Failed to connect to remote checkpoint store")?,
+            IntCounter::new("unused_total_ingested_bytes", "unused").unwrap(),
```
bmwill: why is this thing named "unused"?

evan-wall-mysten: This `IntCounter` doesn't do anything in `format_snapshot.rs` because the `checkpoint` method that increments the counter is unused.

I changed this to use an `Option<IntCounter>` and passed `None` here to make it more obvious that it does nothing.

PR #26100, opened 2026-04-01
sui-indexer-alt-framework: replace futures::stream::Peekable with tokio_stream::adapters::Peekable

evan-wall-mysten:
## Description 

This PR changes the `CheckpointStream::stream` from `futures::stream::Peekable` (takes `self: Pin<&mut Self>:`) to `tokio_stream::adapters::Peekable` (takes `&mut self`) to simplify call sites from `Pin::new(&mut stream).peek()` to `stream.peek()`.

## Test plan 

This affects the compiler, no runtime changes.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
@@ -8,12 +8,12 @@ use anyhow::anyhow;
 use async_trait::async_trait;
 use futures::StreamExt;
 use futures::stream::BoxStream;
-use futures::stream::Peekable;
 use sui_rpc::headers::X_SUI_CHAIN_ID;
 use sui_rpc::proto::sui::rpc::v2::SubscribeCheckpointsRequest;
 use sui_rpc::proto::sui::rpc::v2::subscription_service_client::SubscriptionServiceClient;
 use sui_types::digests::ChainIdentifier;
 use sui_types::messages_checkpoint::CheckpointDigest;
+use tokio_stream::adapters::Peekable;
 use tonic::Status;
 use tonic::transport::Endpoint;
 use tonic::transport::Uri;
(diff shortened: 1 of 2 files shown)


crates/sui-indexer-alt-framework/src/ingestion/streaming_client.rs
```diff
@@ -109,15 +109,15 @@ fn wrap_stream(
     stream: impl futures::Stream<Item = Result<Checkpoint>> + Send + 'static,
     statement_timeout: Duration,
 ) -> Peekable<BoxStream<'static, Result<Checkpoint>>> {
-    tokio_stream::StreamExt::timeout(stream, statement_timeout)
+    let stream = tokio_stream::StreamExt::timeout(stream, statement_timeout)
```
amnn, line 112: Do we need to use both `futures::StreamExt` and `tokio_stream::StreamExt`, or could we standardise on the latter in this file? That would allow this call to become:

```
stream
    .timeout(statement_timeout)
    .map(move |result| ...)
    .peekable()
```

evan-wall-mysten: Yeah that is still the worst part about this and I couldn't figure out a way around it yet.

`futures::StreamExt` is needed for `.boxed` and `tokio_stream::StreamExt` does not have an equivalent. I was trying to figure out if there was a way to avoid the `BoxStream` completely, but I didn't find a good solution yet.

The good part though is that this is at least limited to where the stream is being created and not to the stream users anymore.

amnn: You can scope the trait `use` more tightly to the functions that need them, so that there using both doesn't trigger an ambiguity.

PR #26104, opened 2026-04-01
wip(indexer-alt): ingestion-based obj_versions pruning

amnn:
## Description

An alternate approach for pruning `obj_versions` that fetches checkpoints in the pruning range and uses their input objects and wrapped/deleted objects to determine which records to prune.

## Test plan

New E2E test:

```
$ cargo nextest run -p sui-indexer-alt-e2e-tests -- obj_versions_pruning_tests
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt/src/handlers/obj_versions.rs
```diff
@@ -63,4 +77,57 @@ impl Handler for ObjVersions {
...
+        for cp in &checkpoints {
+            for tx in &cp.checkpoint.transactions {
+                let lamport = tx.effects.lamport_version();
+                for change in tx.effects.object_changes() {
+                    if let Some(v) = change.input_version {
+                        ids.push(change.id.to_vec());
+                        versions.push(v.value() as i64);
+                    }
+
+                    if change.output_version.is_none() {
```
wlmyng: This handles wrapped/ deleted? On obj_versions, when an object is wrapped or deleted, we track it with the transaction's lamport version, so this is needed to handle wrapped/ deleted object records

crates/sui-indexer-alt/src/handlers/obj_versions.rs
```diff
@@ -63,4 +77,57 @@ impl Handler for ObjVersions {
...
+            }
+        }
+
+        if ids.is_empty() {
+            return Ok(0);
+        }
+
+        let query = query!(
+            r#"
+            DELETE FROM
```
wlmyng: I can kick off an indexer to see how this performs and iterate on the delete query if necessary

wlmyng (commented):
For context, the other pruning approach for `obj_versions` consists of 4 phases, where per checkpoint chunk:
1. delete intermediate versions for objects with multiple modifications
2. delete the remaining local entries that are superseded by a newer version
3. delete the predecessor below pruner_hi
4. clean up wrapped / deleted objects

The challenge with this approach is that the reference data is also the work queue. Phase 2, needed for correctness, becomes expensive as it is a per-object lookup. Dense early-chain checkpoints created hundreds of thousands of objects, and at any time we could observe similar if not greater load in the network. In my approach, the pruning implementation consults the table, which in the worst case becomes a HashAggregate and spill to disk to complete the delete query.

I think this approach is promising especially since we had similar logic for `obj_info` (which constructed a reference table) and as @amnn had called out in our meeting, fullnode uses this to prune as well. It should be less work on the db in the very least, since now postgres just needs to delete based on a list of keys

Happy to run a few benchmarks with this implementation

crates/sui-indexer-alt/src/handlers/obj_versions.rs
amnn: Yes please! It is in your hands now :)

crates/sui-indexer-alt/src/handlers/obj_versions.rs
```diff
@@ -63,4 +89,116 @@ impl Handler for ObjVersions {
...
+        let client = self.client.clone();
+        let fetcher: tokio::task::JoinHandle<Result<(), Break<anyhow::Error>>> =
+            tokio::spawn(async move {
+                stream::iter(from..to_exclusive)
+                    .try_for_each_send_spawned(
+                        ConcurrencyConfig::adaptive(33, 1, 2000),
+                        move |cp| {
+                            let client = client.clone();
+                            async move {
+                                let checkpoint =
```
amnn, line 110: To make best use of the adaptive concurrency, you should also do the work of turning the checkpoint into deletion records in here.

PR #26107, opened 2026-04-02
sui: ensure external keys files exist

jordanjennings-mysten:
## Description 

If external.keystore or external.aliases does not exist create them.

## Test plan 

shell tests
- prompt_if_no_config: no sui directory
- .sui/client exists but client.yml is old and does not have external_keys field
- external keys in .client yaml is correct but no external.keystore + external.aliases
- no external keys in .client.yaml but have existing external.keystore + external.aliases

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/tests/shell_tests/external_keys_bootstrap/missing_external_files.sh
+++ b/crates/sui/tests/shell_tests/external_keys_bootstrap/missing_external_files.sh
@@ -9,6 +9,7 @@ export SUI_CONFIG_DIR="$PWD/config"
 
 set -o pipefail
 sui external-keys list-keys missing-signer 2>&1 \
+  | sed 's/program not found.*/<REDACTED>/g' \
   | sed 's/No such file or directory (os error 2)/<REDACTED>/g'
 echo "exit_code: $?"
 
@@ -18,4 +19,5 @@ ls config
 
 echo ""
 echo "client config:"
-sed 's/active_address:.*$/active_address: <REDACTED>/g' config/client.yaml
+sed 's|  External: .*|  External: <SANDBOX_DIR>/config/external.keystore|g' config/client.yaml \
+  | sed 's/active_address:.*$/active_address: <REDACTED>/g'
--- a/crates/sui/tests/shell_tests/external_keys_bootstrap/missing_external_files.snap
+++ b/crates/sui/tests/shell_tests/external_keys_bootstrap/missing_external_files.snap
@@ -1,7 +1,5 @@
 ---
 source: crates/sui/tests/shell_tests.rs
-expression: result
-input_file: crates/sui/tests/shell_tests/external_keys_bootstrap/missing_external_files.sh
 ---
 ----- script -----
 # Copyright (c) Mysten Labs, Inc.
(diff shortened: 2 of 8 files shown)


crates/sui-keys/src/external.rs
```diff
@@ -150,6 +152,24 @@ impl External {
             BTreeMap::default()
         };
 
+        if !aliases_file_exists {
+            let aliases_store = serde_json::to_string_pretty(&aliases)
+                .map_err(|e| anyhow!("Serialization error: {}", e))?;
```
amnn: Use `.context`, and describe what you're serializing.


```suggestion
                .context("error serializing aliases")?;
```

crates/sui-keys/src/external.rs
```diff
@@ -150,6 +152,24 @@ impl External {
...
+                format!(
+                    "Cannot write aliases to file: {}",
+                    aliases_store_directory.display()
+                )
+            })?;
+        }
+
+        if !keystore_file_exists {
+            let keys_store = serde_json::to_string_pretty(&keys)
+                .map_err(|e| anyhow!("Serialization error: {}", e))?;
```
amnn:
```suggestion
                .context("error serializing keystore")?;
```

jordanjennings-mysten:
> This looks good though I am wondering in which context this gets triggered. Is it when you have a ~/.sui/sui_config directory that's empty that an error occurs?

this had me thinking.. I definitely need to call prompt_if_no_config(). it assumes .sui directory.

stefan-mysten:
> > This looks good though I am wondering in which context this gets triggered. Is it when you have a ~/.sui/sui_config directory that's empty that an error occurs?
> 
> this had me thinking.. I definitely need to call prompt_if_no_config(). it assumes .sui directory.

Right, but that's mostly a `sui` thing rather than a `keystore` thing. What is the problem this PR is trying to address?

jordanjennings-mysten:
turns out there are 4 scenarios:
- prompt_if_no_config: no sui directory
- .sui/client exists but client.yml is old and does not have external_keys field
- external keys in .client yaml is correct but no external.keystore + external.aliases
- the opposite of the previous

the tests I added should clear this up.

crates/sui/src/sui_commands.rs
```diff
@@ -523,7 +523,10 @@ impl SuiCommand {
                 cmd,
             } => {
                 let client_path = sui_config_dir()?.join(SUI_CLIENT_CONFIG);
-                let mut config = PersistedConfig::<SuiClientConfig>::read(&client_path)?;
+                prompt_if_no_config(&client_path, false).await?;
```
stefan-mysten, line 526: `sui_config_dir` will create the `.sui/sui_config` folder if I recall correctly, so not sure you need to call `prompt_if_no_config`. It's probably fine though.

jordanjennings-mysten: I think the problem is client.yml wont exist. I played with this a bit today and it does seem like we need to generate the client.yml

stefan-mysten: Actually, you probably need it because `sui_config_dir` only creates the directories.

PR #26109, opened 2026-04-02
Fix compilation warnings in move-vm-runtime and move-regex-borrow-graph

akmysti:
## Description 

Gate debug/testing-only code behind #[cfg(any(debug_assertions, feature = "testing"))] to eliminate warnings in optimized builds:
- Unused imports in linkage_context.rs, state.rs, functions.rs
- Dead code: InternedDisplay trait/impls, FieldHandle.owner, FieldInstantiation.owner, StackFrame::iter, NativeContext.state
- move-regex-borrow-graph: successors_idx, edge_weight, all_edges_idx, check_invariants only used in debug/test

## Test plan 

```
cargo bench -p language-benchmarks --no-run
```
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
+++ b/external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
@@ -129,6 +129,7 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
     }
 
     /// Returns the direct successors of the specified reference by NodeIndex
+    #[cfg(debug_assertions)]
     fn successors_idx(
         &self,
         r: NodeIndex,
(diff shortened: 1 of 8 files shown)


tnowacki (commented):
A lot of these cfg attrs feel a bit invasive and strangely overfit. Could we just allow/expect warnings? might be a bit more intentional

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -129,6 +129,7 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
     }
 
     /// Returns the direct successors of the specified reference by NodeIndex
+    #[allow(dead_code)]
```
tnowacki: Is there a better one here? Like unused function or something similar? Since dead code could also be unreachable code (like following a `return` or a `todo!`)

akmysti: `unused` is more loose (https://stackoverflow.com/questions/64556750/what-is-the-difference-between-the-dead-code-and-unused-lints) but the code is currently only called from test so `dead_code` seems more appropriate atm. lmk if you prefer `unused` as that should also fix the warning.

PR #26114, opened 2026-04-02
Prioritize admission of tx to consensus by gas price

aschran:
## Description 

Adds a bounded priority queue that prioritizes transaction admission by gas price during congestion, replacing the previous probabilistic load shedding for consensus overload.

Queue activates only above a configurable load threshold; no overhead during normal operations.

## Test plan 

Added unit and e2e tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-core/src/admission_queue.rs
```diff
@@ -0,0 +1,689 @@
...
+            return false; // duplicate insert
+        }
+
+        if self.total_len < self.capacity {
+            self.push_entry(entry, keys);
+            self.metrics.queue_depth.set(self.total_len as i64);
+            return true; // directly inserted
+        }
+
+        let min_price = self.min_gas_price().unwrap();
```
alex-mysten, line 139: Is is possible that an unvalidated config could cause a panic here (if capacity == 0, for instance)

crates/sui-core/src/authority_server.rs
```diff
@@ -958,45 +960,82 @@ impl ValidatorService {
...
-            )
-            .await?
-        } else {
-            let futures = consensus_transactions.into_iter().map(|t| {
+        }
+
+        let consensus_positions = if is_ping_request || self.should_bypass_admission_queue() {
+            // Submit directly to consensus, bypassing the admission queue.
+            self.metrics.admission_queue_direct_bypasses.inc();
+            if is_soft_bundle_request || is_ping_request {
```
alex-mysten: If we move up the `tx_groups` above this block, can we flatten this branch?

crates/sui-core/src/authority_server.rs
```diff
@@ -958,45 +960,82 @@ impl ValidatorService {
...
+                let rx = aq
+                    .try_insert(gas_price, txns, submitter_client_addr)
+                    .await?;
+                receivers.push(rx);
+            }
+            let results = future::try_join_all(receivers.into_iter().map(|rx| async move {
+                rx.await.map_err(|_| {
+                    SuiError::from(SuiErrorKind::TooManyTransactionsPendingConsensus)
+                })?
+            }))
```
alex-mysten: During reconfig, would transactions accepted into the admission queue and waiting for consensus positions surface as TooManyTransactionsPendingConsensus?

aschran: during reconfig these transactions fail with ValidatorHaltedAtEpochEnd, and then are retried via the handle_submit_transaction retry loop (which catches ValidatorHaltedAtEpochEnd)

crates/sui-core/src/admission_queue.rs
aschran: good point, added an assert at queue construction time that capacity>0

crates/sui-core/src/authority_server.rs
aschran: good idea, done

mystenmark (commented):
Looks good overall. I think we need a few things:
- testing in PTN
- a plan for a gradual rollout - running on mysten validators initially might be enough
- ability to totally disable the system via config
- a dead-man's switch, so if no transactions are submitted for 30 seconds, the whole system fails open.

aschran:
> * ability to totally disable the system via config
> * a dead-man's switch, so if no transactions are submitted for 30 seconds, the whole system fails open.

@mystenmark added these two, PTAL

crates/sui-core/src/admission_queue.rs
```diff
@@ -0,0 +1,846 @@
...
+        for entry in entries {
+            self.queue
+                .metrics
+                .queue_wait_latency
+                .observe(entry.enqueue_time.elapsed().as_secs_f64());
+            let adapter = self.consensus_adapter.clone();
+            let es = self.epoch_store.clone();
+            spawn_monitored_task!(submit_queue_entry(entry, adapter, es));
+        }
+        *self.last_drain.lock().unwrap() = Instant::now();
```
mwtian, line 510: Is it ok to update `last_drain` when `available` is 0?

crates/sui-core/src/authority_server.rs
```diff
@@ -990,47 +1014,74 @@ impl ValidatorService {
...
+                for txns in tx_groups {
+                    let gas_price = Self::extract_gas_price(&txns);
+                    let rx = aq
+                        .try_insert(gas_price, txns, submitter_client_addr)
+                        .await?;
+                    receivers.push(rx);
+                }
+                let results = future::try_join_all(receivers.into_iter().map(|rx| async move {
+                    rx.await.map_err(|_| {
+                        SuiError::from(SuiErrorKind::TooManyTransactionsPendingConsensus)
```
mwtian, line 1077: Can we use a different error variant for txn hitting inflight limit + being outbid? Maybe the error message can be more explicit in the outbid aspect. `TooManyTransactionsPendingConsensus` can remain as an error on total pending.

mwtian (commented):
This PR can be a good 1st step towards better control on what gets submitted to consensus. In future, have we considered refactoring consensus submission more significantly? One proposal: instead of pushing transactions into consensus, consensus can pull transactions  instead. Sui can pass a `SuiTransactionCollector` to consensus, similar to `SuiTxVerifier`. `SuiTransactionCollector` will hold pending user and system transactions, and return to consensus a list of prioritized transactions bounded by total size and count when polled. Then there is no latency concern. Priority can also be managed more tightly without inflight submissions.

crates/sui-core/src/authority_server.rs
aschran: after chat with @mystenmark  - we are going to allow duplicate submission to the queue (since it's currently allowed outside of the queue) with no error, and just count duplicate submissions as spam for dos protection purposes

crates/sui-core/src/admission_queue.rs
aschran: good catch, added a test + fixed

mystenmark (approved):

crates/sui-core/src/authority_server.rs
```diff
@@ -978,47 +1002,78 @@ impl ValidatorService {
...
-                .await?
+            consensus_transactions
                 .into_iter()
-                .flatten()
+                .map(|t| vec![t])
                 .collect()
         };
 
+        let consensus_positions: Vec<ConsensusPosition> = match submit_mode {
+            AdmissionQueueSubmitMode::Bypass | AdmissionQueueSubmitMode::Disabled => {
```
mystenmark, line 1052: i wonder if it would be simpler to put the mode inside the queue. Then, if it is in bypass or disabled, it always flushes all entries immediately whenever it wakes up.

PR #26129, opened 2026-04-05
docs fix: Update address reference link in Sui Move concepts

ksemele:
docs fix

## Description 

Update address reference link in Sui Move concepts because previous one doesn't exists anymore


ksemele:
@jessiemongeon1 ?

PR #26138, opened 2026-04-06
Reorganize gas charging pipeline into explicit phases

dariorussi:
Split the monolithic `charge_gas` method and restructure `execute_transaction` into a clear phase-based pipeline with uniform error handling.
`execute_transaction` is the function to look at, it's the right starting point.

## What changed

**execution_engine.rs**: Replace the nested `and_then` closure + monolithic `charge_gas` call with an explicit `and_then` chain of four phases:
1. Charge input objects
2. Execute PTB + round computation cost
3. Finalize storage (collect costs, budget check, gasless validation)
4. Validate effects limits (meter size, written objects size)

On any phase failure, `reset_charges` resets the store to the input-only state (drops writes, re-smashes gas, re-touches mutable inputs) and prepares the gas status for the final charge. `GasCharger::charge` is then called unconditionally by the caller on both the success and error paths. All writes are removed on error; gas and input objects are the only things we touch in a transaction with errors.

**gas_charger.rs**: Split `charge_gas` into three focused methods with clean signatures:
- `round_computation(is_move_abort) -> Result<(), ExecutionError>` — round raw gas into pricing tiers. Returns the OOG error to the caller instead of mutating a shared result.
- `finalize_storage() -> Result<(), ExecutionError>` — collect storage + budget check. No internal retries or resets; pure collect + check.
- `charge(&Result<T, E>) -> GasCostSummary` — apply payment. Takes the execution result by shared reference; only reads it to decide skip-charge cases (gasless-on-error, address-balance withdraw shortfall).

The OOG retry logic (`handle_storage_and_rebate_v1/v2`) is eliminated. Recovery is now a caller-side sequence: `reset_charges` → `charge`. If the first `finalize_storage` attempt inside `reset_charges` can't cover input-only storage (or we're on gas model v1), it resets again and sets computation to the full budget with zero storage, then the caller's `charge` applies the payment.

**API cleanup**: no function in the gas-charging API takes `&mut Result` any more. Errors flow via return values and the caller merges them explicitly.

**Renames**: `bucketize_computation` → `round_computation_cost` across the trait and all execution versions.

## Behavioral notes

All existing gas tests pass (21/21). The refactoring preserves behavior for all common scenarios. Three edge cases improve:

1. Limit checks no longer overwrite execution errors (e.g., MoveAbort was previously replaced by EffectsTooLarge, losing the abort gas cap)
2. Conservation re-check after recovery uses the correct cost summary
3. Gasless check failure skips unnecessary storage evaluation

## Test plan

- Existing gas tests in `crates/sui-core/src/unit_tests/gas_tests.rs` (21/21 pass)
- New differential snapshot tests in `crates/sui-core/src/unit_tests/gas_diff_tests.rs` — 11 tests covering happy path, computation OOG (storage OK and fallback), storage OOG (minimal OK and fallback), and Move abort (generous, tight, multi-coin). Each captures the full `GasCostSummary` + `ExecutionStatus` + effects counts as insta snapshots. Snapshots are byte-identical between `main` and this branch, proving the refactor is behavior-preserving on every path exercised.
- Not yet covered (follow-ups): `EffectsTooLarge` / `WrittenObjectsTooLarge` interacting with execution errors, early execution errors, conservation recovery, gasless transaction paths.

Co-Authored-By: My Dog (1M bones) <noreply@my_dog.com> and Claude Opus 4.6 (1M context) <noreply@anthropic.com>

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-types/src/gas_model/gas_v2.rs
+++ b/crates/sui-types/src/gas_model/gas_v2.rs
@@ -189,7 +189,7 @@ mod checked {
         gas_budget: u64,
         // Computation cost after execution. This is the result of the gas used by the `GasStatus`
         // properly bucketized.
-        // Starts at 0 and it is assigned in `bucketize_computation`.
+        // Starts at 0 and it is assigned in `round_computation_cost`.
         computation_cost: u64,
         // Whether to charge or go unmetered
         charge: bool,
@@ -409,7 +409,7 @@ mod checked {
             &mut self.gas_status
         }
 
-        fn bucketize_computation(&mut self, aborted: Option<bool>) -> Result<(), ExecutionError> {
+        fn round_computation_cost(&mut self, aborted: Option<bool>) -> Result<(), ExecutionError> {
             let gas_used = self.gas_status.gas_used_pre_gas_price();
             let effective_gas_price = if self
                 .cost_table
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -235,6 +235,10 @@ pub mod move_integration_tests;
 mod gas_tests;
 
 #[cfg(test)]
+#[path = "unit_tests/gas_diff_tests.rs"]
+mod gas_diff_tests;
+
+#[cfg(test)]
 #[path = "unit_tests/gas_data_tests.rs"]
 mod gas_data_tests;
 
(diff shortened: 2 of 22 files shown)


crates/sui-types/src/gas_model/gas_v2.rs
```diff
@@ -409,7 +409,7 @@ mod checked {
             &mut self.gas_status
         }
 
-        fn bucketize_computation(&mut self, aborted: Option<bool>) -> Result<(), ExecutionError> {
+        fn round_computation_cost(&mut self, aborted: Option<bool>) -> Result<(), ExecutionError> {
```
dariorussi, line 412: apology but very few of the changes here are about renaming this function

PR #26146, opened 2026-04-07
Add direct call inlining for zero-parameter leaf functions

akmysti:
## Description 
Introduce a bytecode-level inlining optimization that replaces DirectCall
    instructions with the callee's body for small, zero-parameter, zero-local
    leaf functions. The two-pass approach first identifies inline candidates,
    then builds new bytecode with expanded call sites and adjusted branch
    targets (including VariantSwitch jump tables).

## Test plan 

Includes comprehensive unit tests (functional + bytecode comparison)

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/language-benchmarks/Cargo.toml
+++ b/external-crates/move/crates/language-benchmarks/Cargo.toml
@@ -20,6 +20,10 @@ move-binary-format.workspace = true
 move-stdlib.workspace = true
 move-package.workspace = true
 
+[features]
+default = []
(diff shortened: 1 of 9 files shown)


dariorussi (commented):
is this just to put an idea out there?
I am not clear we need or want inlining any time soon and certainly not before gas work.
Can you clarify the intent here?

PR #26158, opened 2026-04-08
Add timestamp-based epoch close

mwtian:
## Summary
- Each validator can now transition from AcceptAllCerts to RejectAllCerts directly when the consensus commit timestamp exceeds the epoch reconfiguration timestamp, instead of waiting for 2f+1 EndOfPublish messages
- Simplifies epoch transitions and reduces latency by skipping EndOfPublish voting for epoch change
- EndOfPublish quorum is preserved as a fallback for manual/forced epoch close
- Gated behind `timestamp_based_epoch_close` feature flag (devnet and tests only)

## Test plan
- [x] `cargo insta test -p sui-protocol-config --accept` (snapshot updated)
- [x] `cargo check -p sui-core -p sui-node` passes
- [x] `SUI_SKIP_SIMTESTS=1 cargo nextest run -p sui-core`
- [x] `cargo simtest -p sui-e2e-tests`


crates/sui-core/src/consensus_adapter.rs
```diff
@@ -347,7 +347,10 @@ impl ConsensusAdapter {
         // Send EndOfPublish if needed.
         // This handles the case where the node crashed after setting reconfig lock state
         // but before the EndOfPublish message was sent to consensus.
-        if epoch_store.should_send_end_of_publish() {
+        // Skip when timestamp_based_epoch_close is enabled (epoch close is timestamp-driven).
+        if epoch_store.should_send_end_of_publish()
+            && !epoch_store.protocol_config().timestamp_based_epoch_close()
```
lxfind: In the case if EOP is manually done by an admin, we need to always preserve the crash recovery?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2860,12 +2928,19 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
-                    ConsensusTransactionKind::UserTransaction(_)
-                        | ConsensusTransactionKind::UserTransactionV2(_)
-                        | ConsensusTransactionKind::CertifiedTransaction(_)
-                ) {
+                // When timestamp_based_epoch_close is enabled, validators don't
+                // automatically send EndOfPublish, so skip this filter.
+                if !self
+                    .epoch_store
+                    .protocol_config()
+                    .timestamp_based_epoch_close()
```
lxfind: I think this may have the same problem if EOP was manually triggered

crates/sui-core/src/consensus_adapter.rs
mwtian: Makes sense. We can recover manual eop.

crates/sui-core/src/consensus_handler.rs
mwtian: Removed.

crates/sui-core/src/consensus_handler.rs
```diff
@@ -1422,14 +1434,25 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
         self.send_end_of_publish_if_needed().await;
     }
 
-    fn handle_eop(
+    fn handle_close_epoch(
         &self,
         state: &mut CommitHandlerState,
+        timestamp_based_epoch_close: bool,
```
mystenmark: nit: no need for this parameter since we could just read it from the protocol config

mystenmark (approved):
This looks good to me, but please make sure to run a split/upgrade test on antithesis before landing

crates/sui-core/src/consensus_handler.rs
mwtian: Refactored.

PR #26171, opened 2026-04-09
sui-analytics-indexer: implement accepts_chain_id

evan-wall-mysten:
## Description 

Replaced the `accepts_chain_id` stub with an impl that stores the 32-byte `chain_id` at `_metadata/chain_id/{pipeline_task}`. Uses `PutMode::Create` to claim the file on first call; on `AlreadyExists`, reads it back and returns whether it matches.

## Test plan 

Added new unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Implement `accepts_chain_id` in analytics indexer.

### DIFF
--- a/crates/sui-analytics-indexer/src/store/live.rs
+++ b/crates/sui-analytics-indexer/src/store/live.rs
@@ -34,6 +34,10 @@ impl LiveStore {
         Self { object_store }
     }
 
+    pub(crate) fn object_store(&self) -> &Arc<dyn ObjectStore> {
+        &self.object_store
+    }
+
     /// Determine the watermark by scanning file names in the object store.
     ///
     /// 1. Find epoch directories under `{pipeline}/epoch_*`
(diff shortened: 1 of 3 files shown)


crates/sui-analytics-indexer/src/store/mod.rs
```diff
@@ -629,18 +641,51 @@ impl Connection for AnalyticsConnection<'_> {
...
+            Ok(_) => Ok(true),
+            Err(ObjectStoreError::AlreadyExists { .. }) => {
+                let bytes = object_store.get(&path).await?.bytes().await?;
+                let stored: [u8; 32] = bytes.as_ref().try_into().map_err(|_| {
+                    anyhow::anyhow!(
+                        "stored chain_id at {} has wrong length: {}",
+                        path,
+                        bytes.len()
+                    )
+                })?;
```
amnn: nit: use `with_context` instead of `map_err(|e| anyhow!(...))`, and you can prepend `.ok()` if you want to drop the previous message:



```suggestion
                let stored: [u8; 32] = bytes.as_ref().try_into().ok().with_context(|| {
                    format!(
                        "stored chain_id at {} has wrong length: {}",
                        path,
                        bytes.len()
                    )
                })?;
```

amnn (commented):
Change looks reasonable to me, but @nickvikeras is the expert here.

crates/sui-analytics-indexer/src/store/live.rs
```diff
@@ -222,6 +226,30 @@ mod tests {
...
+        let config = test_config(object_store.clone());
+        let store = AnalyticsStore::new(object_store.clone(), config, test_metrics());
+        (object_store, store)
+    }
+
+    async fn read_stored_chain_id(
+        object_store: &Arc<dyn object_store::ObjectStore>,
+        pipeline_task: &str,
+    ) -> Vec<u8> {
+        let path = ObjectPath::from(format!("_metadata/chain_id/{pipeline_task}"));
```
nickvikeras, line 242: Should this file be global for the indexer? It doesn't really make sense to have pipeline tasks within one indexer that index different chain ids does it?

evan-wall-mysten: I originally had a similar thought, but @amnn mentioned that currently people have setups where data from multiple chains is indexed together, just not in the same pipeline.

nickvikeras: Okay sounds good. If it was intentional i'm fine with it.

nickvikeras (approved):

PR #26173, opened 2026-04-09
[light-client] trust ratcheting improvements

alex-mysten:
## Summary

Faster trust ratcheting in the authenticated-events light client.

- **Parallel epoch fetching** with single-mutex serialization across concurrent ratchets, narrower checkpoint reads (`summary.bcs` + `signature` only), and `Arc<Committee>` in `EpochCache`.
- **`epoch_source` over gRPC.** `EpochDataFetcher` tries the archive endpoint (`archive.{network}.sui.io:443`) first, falls back to the fullnode. Archive misses are retried on the fullnode; fullnode misses surface as a specific "missing epochs" error.
- **Bundled testnet committee** at epoch 1029 (first epoch where `enable_authenticated_event_streams` was active). Cuts the from-scratch testnet ratchet from ~1085 epochs to ~60 — measured 2.3s.
- **`EpochCache` starts at any epoch** (not hard-coded 0), and `genesis_committee` → `starting_committee` to reflect that.
- **`ClientConfig`** moved to builder pattern; validation runs at construction.

## Test plan

- [x] `SUI_SKIP_SIMTESTS=1 cargo nextest run -p sui-light-client --lib` (16 passed)
- [x] testnet e2e one-off — 60-epoch ratchet in 2.3s, 10 events verified
- [x] `cargo xclippy -D warnings`, `cargo fmt --all`


crates/sui-light-client/src/authenticated_events/epoch_cache.rs
```diff
@@ -1,69 +1,102 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+use std::sync::Arc;
 use sui_types::committee::Committee;
 
 pub(crate) struct EpochCache {
-    completed_committees: Vec<(u64, u64, Committee)>,
+    // Committees for every epoch in `[start_epoch, current_epoch_number)`.
+    // `completed_committees[i]` is the committee for epoch `start_epoch + i`.
+    completed_committees: Vec<Arc<Committee>>,
```
bmwill, line 10: why don't we just use a Map?

PR #26178, opened 2026-04-09
[execution] Add Move VM telemetry reporting to ExecutionMetrics

tzakian:
## Description

Make the Move VM runtime telemetry module public and add MoveVMTelemetry (boundary type) and MoveVMTelemetryMetrics (Prometheus gauges) to ExecutionMetrics. The latest execution adapter calls try_update after each transaction. This lazily reads the VM telemetry report every 30 seconds and sets the Prometheus gauges covering cache state, timing, and stack sizes.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-types/src/metrics.rs
+++ b/crates/sui-types/src/metrics.rs
@@ -1,9 +1,12 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::sync::atomic::{AtomicU64, Ordering};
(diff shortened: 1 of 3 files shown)


cgswords (approved):
LGTM but I think we should get someone from an ops team to look at it, too.

PR #26179, opened 2026-04-09
gas data tests across execution entry points

dariorussi:
## Description 

We check that bad gas data produces expected and consistent results across entry point.
Bogus data for address balance and free transaction is not really exercised yet and should come with a later PR (@mystenmark @alex-mysten @aschran if you have suggestions)

Co-Authored-By: My Dog (1M bones) <noreply@my_dog.com> and Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Test plan 

This is it

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/unit_tests/gas_data_tests.rs
+++ b/crates/sui-core/src/unit_tests/gas_data_tests.rs
@@ -1,0 +1,658 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::fmt;
+use std::sync::Arc;
+
+use super::*;
+
+use crate::authority::authority_test_utils::{
+    init_state_validator_with_fullnode, submit_and_execute,
+};
+use move_core_types::identifier::Identifier;
+use move_core_types::language_storage::{StructTag, TypeTag};
+use sui_json_rpc_types::SuiTransactionBlockEffectsAPI;
+use sui_protocol_config::ProtocolConfig;
+use sui_types::base_types::FullObjectRef;
+use sui_types::collection_types::Table as TableType;
+use sui_types::crypto::{AccountKeyPair, get_key_pair};
+use sui_types::effects::TransactionEffectsAPI;
+use sui_types::error::{SuiError, SuiErrorKind, UserInputError};
+use sui_types::gas::GasCostSummary;
+use sui_types::gas_coin::GasCoin;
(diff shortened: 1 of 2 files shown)


crates/sui-core/src/unit_tests/gas_data_tests.rs
```diff
@@ -0,0 +1,657 @@
...
+                );
+                data.gas_data_mut().payment = payment.clone();
+                env.fullnode
+                    .simulate_transaction(data, TransactionChecks::Disabled, mock_gas)
+                    .map(|r| format!("{:?}", r.effects.status()))
+                    .map_err(|e| format!("{e:?}"))
+            }
+            _ => unreachable!(),
+        };
+        println!("{entry_point}: {result:?}");
```
tzakian: Should there be an assertion on the `result` here?

tzakian (approved):
One question on a test but otherwise LGTM

crates/sui-core/src/unit_tests/gas_data_tests.rs
```diff
@@ -0,0 +1,657 @@
...
+        "'{entry_point}': expected Ok, got: {result:?}"
+    );
+}
+
+// =============================================================================
+// Tests
+// =============================================================================
+
+#[tokio::test]
+async fn test_simple_transfer_gas_all_paths() {
```
alex-mysten, line 347: what about a valid gas coin with insufficient balance for gas_budget? thinking about how that would translate to address balances too

alex-mysten (approved):
This looks good to me - one comment - happy to add the address balance / gasless variants once merged

crates/sui-core/src/unit_tests/gas_data_tests.rs
dariorussi: we should have those tests elsewhere, this was mostly to track bad gas data and transaction data related to gas.
To be honest I would like to review all gas tests because we had few issues and I believe we can organize them better

crates/sui-core/src/unit_tests/gas_data_tests.rs
dariorussi: well definitely there should NOT be a `println`

dariorussi:
addressed feedback

PR #26181, opened 2026-04-09
sui-indexer-alt-consistent-store: implement accepts_chain_id

evan-wall-mysten:
## Description 

**Commit 1:**

- Adds a new $chain_id RocksDB column family to the consistent store's internal bookkeeping (alongside existing $watermark and $restore CFs).
- Implements Db::accepts_chain_id() — on the first call for a given pipeline it persists the chain ID and returns true; on subsequent calls it returns true only if the chain ID matches the stored value. This guards against accidentally pointing the indexer at a database populated from a different chain.                                               
- Wires the method into the store::Connection trait impl, replacing the previous TODO stub that always returned true.

**Commit 2:**

Introduces a TestDb helper struct that wraps a tempfile::TempDir and provides .store() and .db() convenience methods for creating a Store or raw (Db, TestSchema).           

## Test plan 

Added unit tests for new features.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Implement `accepts_chain_id` in consistent store.

### DIFF
--- a/crates/sui-indexer-alt-consistent-store/src/store/mod.rs
+++ b/crates/sui-indexer-alt-consistent-store/src/store/mod.rs
@@ -264,6 +264,21 @@ mod tests {
         fn store(&self) -> Store<TestSchema> {
             Store::open(self.0.path().join("db"), DbConfig::default(), 4, None).unwrap()
         }
+
+        fn db(&self) -> (Arc<Db>, TestSchema) {
+            let db_options: rocksdb::Options = DbConfig::default().into();
+            let db = Arc::new(
+                Db::open(
+                    self.0.path().join("db"),
+                    db_options.clone(),
+                    4,
+                    TestSchema::cfs(&db_options),
+                )
+                .unwrap(),
+            );
+            let schema = TestSchema::open(&db).unwrap();
+            (db, schema)
+        }
     }
 
     async fn wait_until<F, R>(f: F) -> Result<(), Elapsed>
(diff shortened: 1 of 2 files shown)


wlmyng (commented):
can we split out the refactors and optimizations from the `accepts_chain_id` implementation into separate PRs? Not blocking my review, but just thinking it would be easier to land each chunk of work on its own

Edit: 

- what's the motivation for using OptimisticTransactionDB?
- separate PRs means we can land the race condition fix/ optimizations individually

wlmyng (commented):
This was a cool PR to read through, but is it not the case that the write path is reachable exactly once for a pipeline? Regardless of store, we record the `chain_id` exactly once for a pipeline. After that first write, every subsequent call - within that run, in the next run, and so on - enters the `Some` branch and is just a read. The invariant is that subsequent indexer runs must only read. So I'm not sure if there is a race condition that justifies the other changes?

As for the `restore_at` race - I think there are already a few things that protect against possible race conditions. Two processes can't run against the same rocksdb, and within the `Restore` command, the `Finalizer` doesn't execute until every restore worker has joined. We also protect against trying to restore a pipeline that has begun indexing, and only allow a restore to resume if the watermark matches.

I do think there could be a gap, where the restore flow doesn't call `accepts_chain_id` (because this is wired into the framework's processor task, which doesn't run during restoration)

Edit: ah, I suppose at the very first startup for a pipeline, concurrent processors see no `chain_id`, and they might all want to write .. but IngestionClient is constructed once per Indexer with a single configured source + we have a OnceCell to collapse concurrent upserts - should mean that we don't actually have a potential race at hand?

crates/sui-indexer-alt-consistent-store/src/store/mod.rs
```diff
@@ -250,6 +255,10 @@ mod tests {
         }
     }
 
+    fn open_test_store(path: &Path) -> Store<TestSchema> {
```
amnn: Does it make sense to re-use this another tests? It seems to follow the pattern used by the other tests as well.

amnn (commented):
Similar comments to @wlmyng -- I'm not seeing the existing race around restoration, or the new potential race for setting the chain ID, because we can only have one writer to the underlying RocksDB instance, and that is protected by `Db` which ensures that per-pipeline we follow one sequential flow through restoration, to initialisation, to ingestion.

evan-wall-mysten:
Removed transactional changes and will add to another PR.

amnn (approved):
LGTM!

amnn:
> Removed transactional changes and will add to another PR.

But do we still need those transactional changes?

PR #26182, opened 2026-04-09
chore(indexer): nits from #26116

amnn:
## Description

Addressing some post-land nits from #26116. The main one is that we should not refer to `anyhow::Result` in an unguarded way, as it leads to confusion about the error type, and we should not re-export types that are accessible through other means (i.e. their module is not private to the crate), as it leads to inconsistencies when the language server tries to resolve them.

## Test plan

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/cluster.rs
+++ b/crates/sui-indexer-alt-framework/src/cluster.rs
@@ -15,7 +15,6 @@ use url::Url;
 
 use crate::Indexer;
 use crate::IndexerArgs;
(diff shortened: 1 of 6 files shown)


amnn:
Incorporated into #26198

PR #26192, opened 2026-04-10
[forking 5/n] add transaction store APIs

stefan-mysten:
## Description 

This PR implements 
- the transaction related APIs on filesystem to be able to read/write from disk transaction data (data, effects, events)
- the initial transaction execution logic and a test that executes a system transaction (advance clock).

## Test plan 
```
cargo nextest run
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-forking/src/filesystem.rs
```diff
@@ -83,6 +101,84 @@ impl FilesystemStore {
...
 
+    /// Return the directory path for storing transaction data.
+    fn transactions_dir(&self) -> PathBuf {
+        self.root.join(TRANSACTIONS_DIR)
+    }
+
+    /// Return the directory for a specific transaction.
+    fn transaction_dir(&self, digest: &TransactionDigest) -> PathBuf {
+        self.transactions_dir().join(digest.to_string())
+    }
```
jordanjennings-mysten, line 112: these names look very similar

crates/sui-forking/src/execution.rs
```diff
@@ -0,0 +1,109 @@
...
+    context: &Context,
+    object_id: &ObjectID,
+) -> Result<(), anyhow::Error> {
+    let simulacrum = context.simulacrum().read().await;
+    let data_store = simulacrum.store();
+    let obj = data_store.get_object(object_id).ok().flatten();
+    obj.ok_or_else(|| anyhow::anyhow!("Object {} not found in store during execution", object_id))?;
+
+    Ok(())
+}
```
stefan-mysten: We probably don't need this, because `simulacrum` will fetch these objects through DataStore.

crates/sui-forking/src/execution.rs
```diff
@@ -0,0 +1,109 @@
...
+
+/// Execute a transaction and commit it to the store. This will fetch and store all input objects
+/// before execution to ensure the Simulacrum has all necessary data to execute the transaction.
+pub async fn execute_transaction(
+    context: &Context,
+    tx_data: TransactionData,
+) -> Result<ExecutionResult, RpcError> {
+    // Fetch and cache input objects
+    fetch_input_objects(context, &tx_data).await?;
```
stefan-mysten: No need.

crates/sui-forking/src/execution.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub effects: TransactionEffects,
+    pub _execution_error: Option<ExecutionError>,
+}
+
+/// Execute a transaction and commit it to the store. This will fetch and store all input objects
+/// before execution to ensure the Simulacrum has all necessary data to execute the transaction.
+pub async fn execute_transaction(
+    context: &Context,
+    tx_data: TransactionData,
+) -> Result<ExecutionResult, RpcError> {
```
jordanjennings-mysten: at some point I would either use thiserror here to keep this layer free of RPC knowledge. There also might be an error type in sui-types that gives you what you want here with the conversions into rpc errors for free.

jordanjennings-mysten (approved):

crates/sui-forking/src/filesystem.rs
stefan-mysten: Will fix in next PR, thank you!

crates/sui-forking/src/execution.rs
stefan-mysten: Removed, returns `anyhow::Error` instead.

PR #26194, opened 2026-04-10
[graphql-stream] Add package caching from streamed checkpoints [5/n]

tpham-mysten:
## Description 

Adds in-memory package caching for streaming subscriptions so that type resolution works for packages seen on the gRPC stream before `kv_packages` has indexed them

 ## Key logic
 
 **Dedicated streaming index with watermark-gated eviction.** Streamed checkpoints populate a separate in-memory store (DashMap). Packages are only evicted once the `kv_packages` watermark has indexed the checkpoint that introduced them — guaranteeing the DB can serve them after eviction. The eviction task runs periodically (default 5 min), reads the watermark via an observable `watch::Receiver<Arc<Watermarks>>` on `WatermarkTask`, and drains queued entries below the watermark. A checked removal (`DashMap::remove_if`) handles system package upgrades where the same ID is re-indexed at a later checkpoint.

**Startup readiness gate.** On cold start, the stream begins at some checkpoint C, but packages from earlier checkpoints live only in the DB. Subscriptions block until `kv_packages` has caught up to C-1, at which point type resolution is guaranteed to succeed regardless of which checkpoint a subscriber receives.

**Layered fetch path.** Resolution flows: streaming index → DB. Each layer is isolated and can be tested independently.

## Test plan 
- Unit tests for the streaming index, LRU cache, and eviction task (with paused-time deterministic scheduling)
- E2E subscription tests including a full create/update/wrap/unwrap lifecycle validating `inputState`/`outputState` JSON resolution through the streaming path
- `cargo fmt` and `cargo clippy` clean

---

## Stack
- #26019
- #26094
- #26117
- #26170
- #26194

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/testing/harness.rs
```diff
@@ -221,3 +235,45 @@ pub fn checkpoint_tx_digests(item: &Value) -> Vec<&str> {
...
+    timeout(Duration::from_secs(30), async {
+        loop {
+            if let Ok(mut conn) = reader.connect().await {
+                #[derive(QueryableByName)]
+                struct W {
+                    #[diesel(sql_type = BigInt)]
+                    checkpoint_hi_inclusive: i64,
+                }
+                if let Ok(rows) = conn
+                    .results::<_, _, W>(sui_sql_macro::query!(
```
amnn: Note that you don't really need `query!` for this, you can just use a normal diesel DSL based query (`query!` is only really intended for cases where the DSL does not work, because it gives you worse type safety guarantees in exchange for more query flexibility).

amnn (commented):
Seeding the existing package cache is not the right move here, I don't think, for the reason that you mentioned in the description -- entries can be evicted from this cache.

Instead, for subscriptions, you need to introduce a store in front of the regular `kv_packages` supplied store (and LruCache) which is fed from recent subscribed checkpoints. It should be configured to act as a rotating window of packages that gets checked first before falling back to the database.

crates/sui-indexer-alt-graphql/src/task/streaming/streaming_package_store.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Maintains a dedicated LRU cache of packages seen in streamed checkpoints,
+/// isolated from the query path's cache. On fetch, checks the streaming cache
+/// first before falling through to the inner store (typically the shared
+/// PackageCache backed by the database).
+///
+/// This prevents query traffic from evicting packages needed by streaming
+/// subscribers, while still falling back to the database for packages not
+/// recently seen in the stream (e.g., system packages from genesis).
+pub(crate) struct StreamingPackageStore<S> {
+    packages: Mutex<LruCache<AccountAddress, Arc<Package>>>,
```
amnn: You don't want to use an `LruCache` here, because the eviction policy would not be right (the issue is not just that query traffic could cause an eviction, even subscription traffic alone could cause problems). Instead:

- use something like `DashMap` for `packages` to allow for concurrent access, store the checkpoint the package came from along with the package.
- store a separate structure mapping checkpoint to the list of packages that were introduce by that checkpoint, call this `checkpoints` (this is a concurrent queue of some kind).
- have it populated by the checkpoint streaming task (as it is today), but nit, this is not really caching, because we cannot evict entries at will, let's just call it `index_packages`.
- introduce a separate task, that you start up concurrently with the checkpoint streaming task. 
  - Its job is to regularly check on the watermarks that are being updated by the `WatermarkTask`, and based on the `kv_packages` watermark, 
  - it evicts entries from `packages` by popping values from the beginning of `checkpoints` while they are below the `kv_packages` watermark, and removes the packages associated with that checkpoint, `C`, from `packages`, as long as the checkpoint associated with that package in `packages` is still `C` (this is to deal with the case of system packages that are updated in place).
  - you need to be a little careful with the concurrency here, to avoid a TOCTOU bug (check the checkpoint for the package, then it gets updated before you go to delete it). `DashMap` has atomic checked removal primitives for this.

The easiest way to model a concurrent queue in Rust/with Tokio is to use an unbounded `mpsc::channel`, and then convert its received into a peekable `ReceiverStream`.

This new eviction task does not need to run often -- it can be every ~5 minutes or something.

tpham-mysten: Yeah I was actually thinking about this approach, and I think we briefly discussed it during our pair programming too. I was holding it off for a couple reasons:

1. Regardless of which caching/storage approach we use, we still need a bound on it for when kv_packages is lagging behind, plus a strategy for handling that (e.g. stalling the stream, returning a partial error, using @Deferred, etc.). So I deferred it to Phase 2 to tackle together.
2. The LRU approach works well enough for now since the number of new packages per checkpoint (in mainnet) is very low ([p99 = 1](https://loom.com/i/8204a0532ee043259a26381f4ef54610)). An LRU with a fixed capacity handles eviction fine in practice. It's true this breaks down when `kv_packages` lags, but that's the same problem the separate-pruning-task approach would face.

That said, happy to discuss the approach itself now:

- IMO the separate-task approach is better than LRU in pretty much all aspects besides implementation complexity (though even that isn't bad, just the race considerations you mentioned). One thing I noticed though: once `kv_packages` catches up, your approach prunes all entries, so future reads for those packages would always hit the DB. That could be inefficient for frequently loaded packages. I'm thinking we could add an LRU layer in between, so the lookup flow becomes: index_packages → LRU → DB (the pruner task only removes from index_packages). WDYT? If that sounds good, I'll implement it in this PR and handle the error/bound scenarios in Phase 2.

amnn: > * I'm thinking we could add an LRU layer in between, so the lookup flow becomes: index_packages → LRU → DB

Agreed, this is what I had in mind also.

I do think it's worth fixing in this PR -- the main issue I have with the LRU approach is that we have no way of guaranteeing that the package being evicted is definitely accessible elsewhere (the correlation between the package being the least recently used, and it being indexed by `kv_packages` is quite weak, which would lead to hard to predict behaviour when things go wrong).

Even if we didn't have an answer for what to do if `kv_packages` lags with the task-based approach, the way the system behaves when the pipeline lags is very predictable and easy to detect (increased memory usage and/or cache size, which is something we can monitor).

In practice, I think we can set a checkpoint window that is so generous that the service would be considered unhealthy because of lag before we ran into memory issues from the in-memory package lookaside table (like, an hour or more).

One other thing you'll need to think about is how to deal with initialisation. Once the service starts accepting checkpoints by streaming at checkpoint `C`, you need to wait until the `kv_packages` watermark is at least `C` before you accept new subscriptions.

tpham-mysten: Sounds good. I have implemented the eviction task approach in this PR and the logic to deal with the gap on initialization. High level details are included in the PR description and also crates/sui-indexer-alt-graphql/src/task/streaming/mod.rs file

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/testing/harness.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/task/streaming/lru_package_store.rs
amnn, line 1: Do you still need this component? I thought you could use the existing LRU cache support in sui-package-resolver. If you introduce this new cache, then you would also need to re-implement system package eviction, whereas if you rely on the shared cache used by regular queries, you can rely on the `system_package_task` to do this for you:

https://github.com/MystenLabs/sui/blob/1a844db79902ad31e454be5ee18feb9a37f36240/crates/sui-indexer-alt-reader/src/system_package_task.rs#L27-L38

crates/sui-indexer-alt-graphql/src/task/streaming/package_eviction_task.rs
```diff
@@ -0,0 +1,267 @@
...
+use sui_futures::service::Service;
+use tracing::debug;
+
+use crate::task::watermark::KV_PACKAGES_PIPELINE;
+use crate::task::watermark::Pipeline;
+use crate::task::watermark::WatermarksLock;
+
+use super::StreamedPackageStore;
+
+/// Queue that records each streamed checkpoint's package IDs for eventual eviction.
```
amnn: Could you use an unbounded tokio `mpsc` channel for this instead? 

- The task that fills the streaming package store, sends the list of package IDs down the channel,
- The receiving side wraps it in a peekable stream:
  - In a loop it fetches the latest watermark, peeks the next element in the stream, takes it if its below the watermark fetched, evicts those packages.
  - ...repeat until the peek shows a checkpoint that is ahead of the watermark.

This saves you having to re-implement a concurrent queue using `Mutex` and `VecDeque` and should be more efficient.

crates/sui-indexer-alt-graphql/src/task/streaming/subscription_readiness.rs
```diff
@@ -0,0 +1,55 @@
...
+                let Some(&first_cp) = self.first_checkpoint.get() else {
+                    return false;
+                };
+                let target = first_cp.saturating_sub(1);
+                w.per_pipeline()
+                    .get(KV_PACKAGES_PIPELINE)
+                    .is_some_and(|p| p.hi().checkpoint() >= target)
+            })
+            .await
+            .map_err(|_| anyhow!("Watermark task shut down before subscriptions became ready"))?;
```
amnn:
```suggestion
            .await
            .ok()
            .context("Watermark task shut down before subscriptions became ready")?;
```

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -156,11 +165,18 @@ impl WatermarkTask {
         self.watermarks.clone()
     }
 
+    /// Receiver for observing watermark updates. Use `wait_for` with a predicate to await
+    /// specific pipeline conditions.
+    pub(crate) fn watermarks_receiver(&self) -> watch::Receiver<Arc<Watermarks>> {
```
amnn: nit (optional): convention

```suggestion
    pub(crate) fn watermarks_rx(&self) -> watch::Receiver<Arc<Watermarks>> {
```

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -225,7 +241,9 @@ impl WatermarkTask {
                     "Watermark updated"
                 );
 
-                *watermarks.write().await = Arc::new(w);
+                let w = Arc::new(w);
+                *watermarks.write().await = w.clone();
+                let _ = watermarks_tx.send(w);
```
amnn, line 249: Is it worth performing a broader refactor (as a follow-up) to also switch the context to receive watermarks vs the watch channel? I think `WatermarkLock` is kind of acting like a broadcast "channel" between the watermark task and individual requests at the moment, so we have kind of a split responsibility

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -234,12 +234,25 @@ pub struct SubscriptionConfig {
...
+    /// Capacity of the LRU layer in the streaming package pipeline.
+    ///
+    /// Sits between the streaming index (for packages not yet in the DB) and the shared
+    /// PackageCache → DB. Caches frequently accessed packages to avoid DB round-trips.
+    /// Eviction here is harmless — the DB always has these packages.
+    pub package_lru_capacity: u64,
+
+    /// How often (in milliseconds) the eviction task checks the `kv_packages` watermark
+    /// and evicts indexed packages from the streaming index.
+    pub eviction_interval_ms: u64,
```
amnn: This probably needs to be clarified that it's related to packages.


```suggestion
    pub package_eviction_interval_ms: u64,
```

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -26,12 +27,17 @@ impl Subscription {
...
     ) -> Result<impl futures::Stream<Item = Result<Checkpoint, RpcError>>, RpcError> {
-        let package_store = ctx.data::<Arc<PackageCache>>()?.clone();
+        let package_store = ctx.data::<Arc<StreamingPackageStore>>()?.clone();
         let limits: &Limits = ctx.data()?;
         let resolver_limits = limits.package_resolver();
         let broadcaster: &CheckpointBroadcaster = ctx.data()?;
+        let readiness = ctx.data::<Arc<SubscriptionReadiness>>()?.clone();
 
         Ok(async_stream::stream! {
+            if let Err(e) = readiness.wait_for_ready().await {
```
amnn: Rather than adding the readiness check on each subscription entrypoint, can we wait to advertise the schema until the readiness criteria is met?

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -142,9 +158,25 @@ impl CheckpointStreamTask {
...
             while let Some(result) = stream.next().await {
                 let response = result.context("Checkpoint stream error")?;
                 if let Some(checkpoint) = response.checkpoint {
+                    let sequence_number = checkpoint
+                        .sequence_number
+                        .context("Checkpoint without sequence_number")?;
+                    if !first_checkpoint_recorded {
+                        self.readiness.record_first_checkpoint(sequence_number);
+                        first_checkpoint_recorded = true;
+                    }
```
amnn, line 172: nit (optional): I'm not sure what the semantics are regarding `response.checkpoint` being `None`, but you could abstract this slightly differently using `peek` (make the stream peekable, `peek` it before you start your main loop, and use that to set the first checkpoint). Purely a code organisation thing.

amnn (commented):
I think you may still have a small problem with stale cache entries in your subscription-specific caching component, could you take a look? I think the fix is as simple as not having that component, and just relying on the shared Lru cache that's managed by the `SystemPackageTask`.

crates/sui-indexer-alt-graphql/src/config.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Ack! I tried it, but the flag still feels more concise and clear enough for this logic, so I will keep it as-is.

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Yes we can, I have moved the readiness.wait_for_ready() to sui/crates/sui-indexer-alt-graphql/src/lib.rs

crates/sui-indexer-alt-graphql/src/task/streaming/package_eviction_task.rs
tpham-mysten: Yeah sounds good! Applied it

crates/sui-indexer-alt-graphql/src/task/watermark.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/task/watermark.rs
tpham-mysten: Agreed! WatermarksLock is effectively a broadcast channel. Added a TODO; following up separately since it touches every request handler.

crates/sui-indexer-alt-graphql/src/task/streaming/lru_package_store.rs
tpham-mysten: `Thanks for the catch! It was overlooking the system package and also the `system_package_task`. I have then removed this store.

tpham-mysten (commented):
> I think you may still have a small problem with stale cache entries in your subscription-specific caching component, could you take a look? I think the fix is as simple as not having that component, and just relying on the shared Lru cache that's managed by the SystemPackageTask.

Thanks for the catch. I was overlooking them and have simply just removed them as you suggested.

crates/sui-indexer-alt-graphql/src/task/watermark.rs
amnn: Sounds good -- note that the lock type should mainly be used in one place (the root request handler that sets up the context for all GraphQL field resolvers), all the other field resolvers use that cached watermark value so that they all see the same watermark and don't contend for the lock.

amnn (approved):
LGTM, thanks!

PR #26198, opened 2026-04-10
sui-indexer-alt-framework: fix indexer startup during streaming hang

evan-wall-mysten:
## Description 

**Commit 1:**

Adds back #26096 and #26116 which were reverted in #26188. 

**Commit 2:**

Specify `timeout` along with `connect_timeout` when creating the  gRPC client.
* `connect_timeout` only bounds the TCP/TLS handshake - the low-level socket connect. Once the TCP connection is established, `connect_timeout` has no further effect. The `subscribe_checkpoints` RPC is a separate HTTP/2 request sent over the already-established connection, so it's not covered by that setting. If a server accepts TCP, but stalls before sending response headers, the client hangs the indefinitely.
* `timeout` is the counterpart that bounds individual RPC calls on the channel.

**Commit 3:**

Include changes from #26182.

## Test plan 

Added new unit test to prove indexer does not hang during startup when streaming hangs.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/mod.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/mod.rs
@@ -14,6 +14,7 @@ use serde::Deserialize;
 use serde::Serialize;
 use sui_futures::service::Service;
 use tokio::sync::mpsc;
+use tracing::warn;
 
 pub use crate::config::ConcurrencyConfig as IngestConcurrencyConfig;
 use crate::ingestion::broadcaster::broadcaster;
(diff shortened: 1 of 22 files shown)


amnn:
@evan-wall-mysten, could you include the changes from https://github.com/MystenLabs/sui/pull/26182 in your change? (Or review that and I can land it after, that's fine too)

crates/sui-indexer-alt-framework/src/ingestion/mod.rs
amnn, line 1: Spurious change

crates/sui-indexer-alt-framework/src/lib.rs
amnn, line 1: We should not make this change -- I'm not sure what's happening in devnet but this should ordinarily be a cheap operation and if it is not, we should fix that, rather than add the complexity here to make any delay on fetching the latest checkpoint less obvious because it doesn't happen all the time.

evan-wall-mysten: Removed

amnn (approved):

PR #26201, opened 2026-04-11
sui-indexer-alt-graphql: handle OutOfRange in watermark task

ebmifa:
## Description

Fix a bug where the GraphQL watermark task gets permanently stuck when the consistent store's snapshot window doesn't cover the indexer DB's checkpoint.

### The bug

In `watermark_from_consistent()`, when the consistent store returns `OutOfRange`, the function bails with an error. The caller (`WatermarkTask::run`) catches this error, logs a warning, and `continue`s — skipping the watermark update entirely. On the next poll (500ms later), the same thing happens. The watermark never gets updated, and the health endpoint's `checkpoint_lag_ms` grows at 1s/s until it hits the threshold and returns 500.

This requires a manual pod restart to recover — there is no self-healing.

### When it happens

After any deploy that restarts the consistent store. The snapshot window starts narrow and may not cover the indexer DB's latest checkpoint. The window eventually expands to cover it, but only after `snapshots / CPS` seconds (~2.9 hours with `snapshots=50000` on devnet).

On testnet/mainnet this is masked by snapshot restore (pre-populates the window) and 2 replicas (rolling updates keep one replica's window intact). But if both replicas ever restart simultaneously or snapshot restore fails, the same bug would hit production.

### The fix

When `OutOfRange` is returned, parse the consistent store's current `max_checkpoint` from the gRPC error metadata and retry `available_range` with that checkpoint. This gives GraphQL a valid (if slightly behind) consistent watermark instead of no watermark at all, allowing it to serve requests immediately.

The watermark will be at the consistent store's tip rather than the indexer DB's tip, so there may be a brief period where the consistent pipeline shows slightly higher lag than the other pipelines. This converges as the consistent store's window expands.

## Test plan

- [x] `cargo check -p sui-indexer-alt-graphql` passes
- [ ] CI
- [ ] Verify on devnet: after CI deploy restarts consistent-store, GraphQL should recover without manual intervention

@evan-wall-mysten @joeymeere for review — you know this code best.

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/task/watermark.rs
+++ b/crates/sui-indexer-alt-graphql/src/task/watermark.rs
@@ -519,13 +519,47 @@ async fn watermark_from_consistent(
         }
 
         Err(consistent_reader::Error::OutOfRange(status)) => {
-            let unknown = AsciiMetadataValue::from_static("<unknown>");
+            // The requested checkpoint is outside the consistent store's current
+            // range. This can happen after a deploy restarts the consistent store
+            // and its snapshot window hasn't expanded to cover the indexer DB's
+            // watermark yet. Instead of bailing (which causes the watermark task
+            // to skip the update entirely, leaving GraphQL stuck returning 500),
+            // retry with the consistent store's current max checkpoint.
+            if let Some(max_val) = status.metadata().get(CHECKPOINT_HEIGHT_METADATA)
+                && let Ok(max_cp) = max_val.to_str().unwrap_or("").parse::<u64>()
+            {
+                warn!(
+                    requested = checkpoint,
+                    consistent_max = max_cp,
+                    "Checkpoint out of consistent range, retrying with consistent store max"
+                );
+                if let Ok(AvailableRangeResponse {
+                    min_checkpoint: Some(min_checkpoint),
+                    max_checkpoint: Some(max_checkpoint),
+                    max_epoch: Some(max_epoch),
+                    total_transactions: Some(total_transactions),
+                    max_timestamp_ms: Some(max_timestamp_ms),
+                    stride: _,
+                }) = consistent_reader.available_range(max_cp).await
+                {
+                    return Ok(Some(WatermarkRow {
+                        pipeline: "consistent".to_owned(),
+                        epoch_hi_inclusive: max_epoch as i64,
+                        checkpoint_hi_inclusive: max_checkpoint as i64,
+                        tx_hi: total_transactions as i64,
+                        timestamp_ms_hi_inclusive: max_timestamp_ms as i64,
+                        epoch_lo: 0,
+                        checkpoint_lo: min_checkpoint as i64,
+                        tx_lo: 0,
+                    }));
+                }
+            }
 
+            let unknown = AsciiMetadataValue::from_static("<unknown>");
             let min = status
                 .metadata()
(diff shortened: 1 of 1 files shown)


ebmifa:
Note: This is a proposed fix based on observed behavior in devnet CI, not a confirmed bug. The OutOfRange bail in watermark_from_consistent may be intentionally preventing the watermark from updating with stale consistent-store data when the indexer DB is ahead.

The concern with the current behavior: when OutOfRange fires, the entire watermark update is skipped (the continue on line 201), and there is no recovery path — GraphQL returns 500 until the consistent store window expands to cover the indexer DB checkpoint (which can take hours depending on the snapshots setting).

If the bail IS intentional, a better fix might be to change the caller to update the non-consistent pipelines even when the consistent watermark fails, rather than skipping the entire update. That way GraphQL indexer-backed queries could still work while the consistent store catches up.

Happy to rework this based on your guidance — you know the consistency invariants best. @evan-wall-mysten @joeymeere

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -519,13 +519,47 @@ async fn watermark_from_consistent(
...
+                    "Checkpoint out of consistent range, retrying with consistent store max"
+                );
+                if let Ok(AvailableRangeResponse {
+                    min_checkpoint: Some(min_checkpoint),
+                    max_checkpoint: Some(max_checkpoint),
+                    max_epoch: Some(max_epoch),
+                    total_transactions: Some(total_transactions),
+                    max_timestamp_ms: Some(max_timestamp_ms),
+                    stride: _,
+                }) = consistent_reader.available_range(max_cp).await
```
tpham-mysten, line 543: A cleaner approach maybe to enrich error metadata on server side [here](https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt-consistent-store/src/rpc/state.rs#L99), so we don't need to make another call within error handling to get these info?

amnn (changes requested):
I am pretty sure this fix is not correct. This scenario happens when the other stores (postgres, bigtable) are behind the trailing edge of the consistent store.

If you ask the consistent store for a watermark without telling it that the other stores are behind, you will get back a response, it will be combined with the watermarks from other stores using `min`, which means it will not affect those watermarks (because the consistent store is ahead), and GraphQL will report that it is healthy can can serve requests at the older checkpoint.

This would be wrong, because if you then tried to make a query that required access to the consistent store, it would fail, because it's outside the consistent store's available range. Bailing at this point is designed exactly to prevent this bug.

The right fix for this issue, is that the GraphQL service should not be marked as healthy, and therefore should not start serving requests until it its stores are in sync with each other. If that's taking longer than expected, that needs to be investigated and fixed.

GraphQL's health check is also now supposed to wait until the watermark task has received its first definite response before it marks the service as healthy -- if that's not working correctly, we should look into that too.

ebmifa:
Closing based on amnn's feedback — the OutOfRange bail is intentional to prevent GraphQL from serving queries that would fail at the consistent store layer. The right fix is to address why the stores take too long to sync after a deploy (consistent-store snapshot window sizing) rather than masking the mismatch. Thanks for the review.

ebmifa:
Will try lowing the consistent-store snapshots for devnet, to see if it helps https://github.com/MystenLabs/sui-operations/pull/7611

PR #26202, opened 2026-04-11
[graphql-stream] Add top-level transaction subscription [6/n]

tpham-mysten:
## Description 

Add `subscription { transactions(filter: ...) { ... } }` that yields individual transactions from finalized checkpoints, with optional `TransactionFilter`

## Test plan 

- Unit + e2e tests
- Local GraphQL server 

---

## Stack
- #26019
- #26094
- #26117
- #26170
- #26194
- #26202

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -46,24 +49,84 @@ impl Subscription {
...
+    /// checkpoints. Transactions are ordered by checkpoint, then by position within
+    /// the checkpoint.
+    ///
+    /// This subscription is not yet available for use.
+    async fn transactions(
+        &self,
+        ctx: &Context<'_>,
+        filter: Option<TransactionFilter>,
+    ) -> Result<impl futures::Stream<Item = Result<Transaction, RpcError>>, RpcError> {
+        let package_store: Arc<PackageCache> = ctx.data::<Arc<PackageCache>>()?.clone();
```
amnn, line 73: nit: do you need the type annotation at both locations? I think just the annotation on the local variable should be enough.

amnn (approved):
Needs a rebase but I looked at the last commit, and it looks reasonable to me!

I'm also curious -- how will we rate limit the number of entries we send through in a single response? We need to get access to the output node size for the simulation request, divide the max output node size by that quantity and ensure that a single response from the subscription service doesn't exceed this size (and we probably also need to rate limit the number of responses we send through per second too).

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Ah nope. Fixed it

tpham-mysten:
Thanks for pointing it out! I honestly did not think about it, but the suggestion looks reasonable. I will think more about and incorporate them into Phase 2 design soon.

PR #26203, opened 2026-04-11
Fix native call double-pop in gas meter stack height tracking

akmysti:
## Description

`charge_native_function_before_execution` was popping args from the gas meter's `stack_height_current` that `charge_call` had already popped. The resulting negative intermediate heights were masked by `saturating_sub` in `pop_stack`.

`pop_stack` has now a single `checked_sub` which branches on the result: `Some(stack_height)` uses it; `None` with `gas_model_version <= 11` saturates to `0` (legacy double-pop is intentional); `None` with `gas_model_version > 11` panics in debug, saturates to `0` in release and reports an invariant violation.

## Test plan

Existing tests.

---

## Release notes

- [x] Protocol: Adds protocol version 123 (`gas_model_version` 12). Fixes a stack-height double-decrement in the gas meter for native calls. Internal accounting correctness only; no user-visible change to successful transactions.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/coin_registry_tests.rs
+++ b/crates/sui-e2e-tests/tests/coin_registry_tests.rs
@@ -12,6 +12,7 @@ async fn test_create_coin_registry_object() {
     let _guard =
         sui_protocol_config::ProtocolConfig::apply_overrides_for_testing(|_, mut config| {
             config.set_execution_version_for_testing(3);
+            config.set_gas_model_version_for_testing(11);
             // The new consensus handler requires these flags, and they are irrelevant to the test
             config.set_ignore_execution_time_observations_after_certs_closed_for_testing(true);
             config.set_record_time_estimate_processed_for_testing(true);
--- a/crates/sui-e2e-tests/tests/coin_registry_tests.rs
+++ b/crates/sui-e2e-tests/tests/coin_registry_tests.rs
@@ -12,6 +12,7 @@ async fn test_create_coin_registry_object() {
     let _guard =
         sui_protocol_config::ProtocolConfig::apply_overrides_for_testing(|_, mut config| {
             config.set_execution_version_for_testing(3);
+            // set up config values that would otherwise be incompatible with the execution version
             config.set_gas_model_version_for_testing(11);
             // The new consensus handler requires these flags, and they are irrelevant to the test
             config.set_ignore_execution_time_observations_after_certs_closed_for_testing(true);
--- a/crates/sui-protocol-config/src/lib.rs
+++ b/crates/sui-protocol-config/src/lib.rs
@@ -324,6 +324,7 @@ const TESTNET_USDC: &str =
 //              Enable bulletproofs verification on devnet.
 //              Enable defer_unpaid_amplification on mainnet.
 // Version 123: Add timestamp_based_epoch_close feature flag and enable in tests.
+//              Fix native call double-pop in gas meter stack height tracking (gas_model v12).
 
 #[derive(Copy, Clone, Debug, Hash, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
 pub struct ProtocolVersion(u64);
@@ -4880,6 +4881,7 @@ impl ProtocolConfig {
                     if chain != Chain::Mainnet && chain != Chain::Testnet {
                         cfg.feature_flags.timestamp_based_epoch_close = true;
                     }
+                    cfg.gas_model_version = Some(12);
                 }
                 // Use this template when making changes:
                 //
--- a/crates/sui-protocol-config/src/lib.rs
+++ b/crates/sui-protocol-config/src/lib.rs
@@ -28,7 +28,7 @@ use tracing::{info, warn};
 
 /// The minimum and maximum protocol versions supported by this build.
 const MIN_PROTOCOL_VERSION: u64 = 1;
-const MAX_PROTOCOL_VERSION: u64 = 122;
+const MAX_PROTOCOL_VERSION: u64 = 123;
 
 const TESTNET_USDC: &str =
     "0xa1ec7fc00a6f40db9693ad1415d0c193ad3906494428cf252621037bd7117e29::usdc::USDC";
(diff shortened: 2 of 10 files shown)


crates/sui-types/src/gas_model/tables.rs
```diff
@@ -570,3 +577,74 @@ pub fn initial_cost_schedule_v5() -> CostTable {
...
+        }
+        assert_eq!(
+            s.stack_height_current(),
+            1,
+            "v11: saturation prevents correct accumulation (should be 10, got 1)"
+        );
+    }
+
+    #[test]
+    #[cfg_attr(not(debug_assertions), ignore)]
```
tnowacki: 😵‍💫 
I am not sure I can parse when this test is or is not used

crates/sui-types/src/gas_model/tables.rs
```diff
@@ -570,3 +577,74 @@ pub fn initial_cost_schedule_v5() -> CostTable {
         stack_height_tiers,
     }
 }
+
+#[cfg(test)]
+mod tests {
```
tnowacki: Having tests here like this feels a bit strange to me. Can we not observe the change through the protocol config in a test?

crates/sui-types/src/gas_model/tables.rs
tnowacki: I think I kind of get it? Maybe? But in general it feels a bit strange to purposefully trigger a debug assert that should never trigger

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4800,6 +4801,13 @@ impl ProtocolConfig {
                         .early_return_receive_object_mismatched_type = true;
                 }
                 122 => {}
+                123 => {
```
tnowacki: We can probably stick this in 122, no?

crates/sui-types/src/gas_model/tables.rs
tnowacki: My point is that this doesn't really show that we had a bug, nor does it show we are fixing one, right? We are just simulating "bugged" behavior

sui-execution/latest/sui-adapter/src/gas_meter.rs
```diff
@@ -148,18 +150,21 @@ impl<G: DerefMut<Target = GasStatus>> GasMeter for SuiGasMeter<G> {
...
-        // Calculate the size decrease of the stack from the above pops.
-        let stack_reduction_size = args.try_fold(
-            AbstractMemorySize::new(pops),
-            |acc, elem| -> PartialVMResult<_> { Ok(acc + abstract_memory_size(&self.0, elem)?) },
-        )?;
-        // Track that this is going to be popping from the operand stack. We also increment the
-        // instruction count as we need to account for the `Call` bytecode that initiated this
-        // native call.
-        self.0.charge(1, 0, pops, 0, stack_reduction_size.into())
+        if fix_native_double_pop(self.0.gas_model_version) {
```
tnowacki, line 153: Do we have a way to double check this logic anywhere? 
Like we do in the PTB interpreter for example:

https://github.com/MystenLabs/sui/blob/main/sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/interpreter.rs#L437-L440

akmysti: that assert_invariant should be sufficient, i guess. Earlier that assert was not triggering because saturating_sub hid the bug.

crates/sui-types/src/gas_model/tables.rs
akmysti: removed this test

crates/sui-types/src/gas_model/tables.rs
akmysti: removed this test.

sui-execution/latest/sui-adapter/src/gas_meter.rs
tnowacki: My point was do we want to add some debug asserts in the interpreter in some spots

dariorussi:
moved to Draft as I am testing few things. I'll open it again once it is ready

sui-execution/latest/sui-adapter/src/gas_meter.rs
dariorussi: I could not spot a good place in the interpreter to put a check that did not feel redundant.
The `debug_fatal` in pop should be good enough I think.
Thoughts?

dariorussi:
ok this is ready for review again.
Everything should have been addressed.
Please take a look

crates/sui-e2e-tests/tests/coin_registry_tests.rs
```diff
@@ -12,6 +12,7 @@ async fn test_create_coin_registry_object() {
     let _guard =
         sui_protocol_config::ProtocolConfig::apply_overrides_for_testing(|_, mut config| {
             config.set_execution_version_for_testing(3);
+            config.set_gas_model_version_for_testing(11);
```
tnowacki, line 16: (and for the other 3 tests) why do we need this?

cgswords (approved):
LGTM, though we might want to double-check with some playback.

crates/sui-types/src/gas_model/tables.rs
tnowacki: I don't think this is called anywhere, but can we add the same check here?

crates/sui-e2e-tests/tests/coin_registry_tests.rs
dariorussi: it felt strange to use an execution version of the past (execution_version = 3) with a gas model that was implemented in newer versions (latest). 
So I changed in all places were execution was taken back in time. To be honest I even wondered about the sanity of a test that picks a protocol config that is not "compatible" with its execution version.
Maybe I am overthinking this

crates/sui-types/src/gas_model/tables.rs
dariorussi: yeah uncalled now but even more of a reason

crates/sui-e2e-tests/tests/coin_registry_tests.rs
tnowacki: ah. no I see that now. Seems great. 
maybe a small comment will help future versions of me read something this obvious

crates/sui-types/src/gas_model/tables.rs
dariorussi: I actually removed the function given it is not called and we are changing that code, so it seems foolish to keep it around

crates/sui-e2e-tests/tests/coin_registry_tests.rs
dariorussi: added small comment

PR #26204, opened 2026-04-12
Fix integer test type mismatches

skinnypete65:
## Description

- Fix the negative `div_ceil` tests in the `u16`, `u32`, `u64`, `u128`, and `u256` stdlib test modules so each test exercises the matching integer type.
- Fix `u128_tests::test_pow_overflow` to invoke `u128::pow` instead of `u16::pow`.
- This restores the intended per-type negative coverage and removes false confidence from mislabeled tests.

## Test plan

```sh
SUI_SKIP_SIMTESTS=1 cargo nextest run --profile ci --cargo-quiet -p sui-framework-tests --test move_tests -- move-stdlib
```

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-framework/packages/move-stdlib/tests/u128_tests.move
+++ b/crates/sui-framework/packages/move-stdlib/tests/u128_tests.move
@@ -83,9 +83,9 @@ fun test_div_ceil() {
     integer_tests::test_div_ceil!(MAX, CASES);
 }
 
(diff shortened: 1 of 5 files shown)


tnowacki (approved):
Thanks for the fix!

PR #26208, opened 2026-04-13
display: differentiate between scoped and latest object loads

amnn:
## Description

Previously Display's `Store` trait only supported one kind of object load. For most implementors, this was fine, but for GraphQL, we need to be able to differentiate between an object load that is scoped within the confines of a parent object, and an object load that is targeting the latest version of an object.

This changes introduces that distinction, so that it's possible to use some part of the parent object as the key for a dynamic field on some other object or derived object rooted on some other object.

## Test plan

New unit tests for `sui-display`:

```
$ cargo nextest run -p sui-display
```

New E2E tests for versioning behavior:

```
$ cargo nextest run -p sui-indexer-alt-e2e-tests
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Fixes an issue where a derived object or dynamic field load could be incorrectly bounded by the version of the object containing the derived object key, instead of the parent.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-display/src/v2/interpreter.rs
+++ b/crates/sui-display/src/v2/interpreter.rs
@@ -27,7 +27,10 @@ pub struct Interpreter<S: V::Store> {
 
     /// Cache of the objects that have been fetched so far. This cache is never evicted -- it is
     /// used to keep objects alive for the lifetime of the interpreter.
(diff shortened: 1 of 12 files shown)


manolisliolios (approved):
looks reasonable to me!

PR #26219, opened 2026-04-13
[move][signed ints] Core types and downstream matches for signed integer support [1/n]

cgswords:
## Description 

Adds `I8..I256` variants to `MoveValue`, `MoveTypeLayout`, `TypeTag`,
the annotated / runtime visitors, `TypeInput`, and the parser, and
propagates exhaustive-match handling across every crate that consumes
these types. Sui-layer rejection at transaction-input and public-schema
boundaries is in place.

VM runtime feature work (the `Value::I*` variants, JIT opcode
execution, arithmetic, gas schedule entries) is deferred to a later
branch. This branch contains only the minimum rejection stubs in
`move-vm-runtime` needed to keep the workspace compiling.

## Test plan 

New tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-display/src/v2/visitor/extractor.rs
```diff
@@ -91,6 +91,56 @@ impl<'v> AV::Visitor<'v, 'v> for Extractor<'v, '_> {
...
 
+    /// Signed integer types are not supported on Sui, and the display `Value`
+    /// enum has no variants for them.
+    fn visit_i8(
+        &mut self,
+        _: &AV::ValueDriver<'_, 'v, 'v>,
+        _: i8,
+    ) -> Result<Self::Value, Self::Error> {
+        Ok(None)
+    }
```
tnowacki, line 102: Should we TODO this or otherwise find some way of debug failing if we hit these?

crates/sui-display/src/v2/value.rs
```diff
@@ -283,6 +283,9 @@ impl Value<'_> {
                 L::U64 => bcs::from_bytes::<u64>(data).ok(),
                 L::U128 => bcs::from_bytes::<u128>(data).ok()?.try_into().ok(),
                 L::U256 => bcs::from_bytes::<U256>(data).ok()?.try_into().ok(),
+                // Signed integer types are not supported on Sui, so they cannot
+                // be coerced to `u64`.
+                L::I8 | L::I16 | L::I32 | L::I64 | L::I128 | L::I256 => None,
```
tnowacki, line 288: Should we just cast them here if they are positive and smaller than u64 max?

crates/sui-json/src/lib.rs
```diff
@@ -642,6 +648,13 @@ fn layout_of_primitive_typetag(tag: &TypeTag) -> Option<MoveTypeLayout> {
         TypeTag::U256 => MTL::U256,
         TypeTag::Address => MTL::Address,
         TypeTag::Signer => return None,
+        // Signed integer types are not accepted as pure transaction inputs on Sui.
```
tnowacki, line 651: But they will be... how will we catch these later?

external-crates/move/crates/move-core-types/src/unit_tests/annotated_visitor_test.rs
```diff
@@ -657,6 +712,8 @@ fn peek_field_test() {
             Ok(None)
         }
 
+        visitor_default! { <'b, 'l> i8, i16, i32, i64, i128, i256 = Ok(None) }
+
```
tnowacki, line 716: Why did the other cases above not use this macro?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
```diff
@@ -537,6 +537,20 @@ impl<'pc, 'vm, 'state, 'linkage, 'extensions> Env<'pc, 'vm, 'state, 'linkage, 'e
...
+                | TypeInput::I32
+                | TypeInput::I64
+                | TypeInput::I128
+                | TypeInput::I256 => {
+                    return Err(ExecutionError::from_kind(
+                        ExecutionErrorKind::TypeArgumentError {
+                            argument_idx: checked_as!(type_arg_idx, u16)?,
+                            kind: TypeArgumentError::TypeNotFound,
+                        },
+                    ));
```
tnowacki, line 552: Bit of strange error but... sure why not
Out of curiosity, is this reachable?

tnowacki: Maybe in dev inspect?

tnowacki (commented):
Overall LGTM but @tzakian might want to look at the layout/value/visitor sections

crates/sui-display/src/v2/value.rs
tzakian: +1

tzakian: That or this is a case of "this should be a `todo!()`"

crates/sui-display/src/v2/value.rs
```diff
@@ -776,6 +779,11 @@ impl<'s> TryFrom<Value<'s>> for Atom<'s> {
                 L::U64 => A::U64(bcs::from_bytes(bytes)?),
                 L::U128 => A::U128(bcs::from_bytes(bytes)?),
                 L::U256 => A::U256(bcs::from_bytes(bytes)?),
+                // Signed integer types are not supported on Sui, and the
+                // display `Atom` has no variants for them.
+                L::I8 | L::I16 | L::I32 | L::I64 | L::I128 | L::I256 => {
+                    return Err(FormatError::TransformInvalid("unexpected signed integer"));
```
tzakian, line 785: `todo!()`?

crates/sui-json-rpc-types/src/sui_move.rs
```diff
@@ -463,6 +467,15 @@ impl From<MoveValue> for SuiMoveValue {
             MoveValue::U64(value) => SuiMoveValue::String(format!("{value}")),
             MoveValue::U128(value) => SuiMoveValue::String(format!("{value}")),
             MoveValue::U256(value) => SuiMoveValue::String(format!("{value}")),
+            // Signed integer types are not accepted as pure transaction inputs on Sui, but
```
tzakian, line 470: What does this comment even mean lol

crates/sui-package-resolver/src/lib.rs
```diff
@@ -1123,11 +1123,11 @@ impl OpenSignatureBody {
...
             S::U256 => O::U256,
-            // Signed integer signatures are not yet supported in the package resolver
+            // Signed integer types are not accepted on Sui. The open signature
+            // body enum has no variants for them, and they should never appear
+            // in signatures of modules deployed on Sui.
             S::I8 | S::I16 | S::I32 | S::I64 | S::I128 | S::I256 => {
-                return Err(Error::UnexpectedError(std::sync::Arc::new(
-                    std::io::Error::other("Signed integer types are not yet supported"),
-                )));
+                return Err(Error::UnexpectedSignedInteger);
```
tzakian, line 1130: uhhhh....uhhhh maybe I'm missing things with higher up in the stack here but this comment and change are not making sense to me

crates/sui-replay-2/src/artifacts/move_call_info.rs
```diff
@@ -194,27 +194,42 @@ impl MoveCallInfo {
...
             TypeInput::U256 => MoveType::U256,
+            // Signed integer types are not supported at the Sui layer.
+            TypeInput::I8
+            | TypeInput::I16
+            | TypeInput::I32
+            | TypeInput::I64
+            | TypeInput::I128
+            | TypeInput::I256 => {
+                return Err(anyhow!(
+                    "signed integer type {:?} is not supported at the Sui layer",
```
tzakian, line 218: This is wrong, although maybe it's a case of "fixed higher in the stack"? But if so it feels like doing a lot of double-work by making a previously infallible function fallible, to then move it back to being fallible?

external-crates/move/crates/move-core-types/src/unit_tests/annotated_visitor_test.rs
tzakian: I'm guessing historical reasons -- the cases above were written before this macro existed IIRC.

external-crates/move/crates/move-model/src/builder/exp_translator.rs
```diff
@@ -392,6 +392,12 @@ impl ExpTranslator<'_, '_, '_> {
             (_, MoveValue::U64(n)) => Value::Number(BigInt::from_u64(*n).unwrap()),
             (_, MoveValue::U128(n)) => Value::Number(BigInt::from_u128(*n).unwrap()),
             (_, MoveValue::U256(n)) => Value::Number(BigInt::from(n)),
+            (_, MoveValue::I8(n)) => Value::Number(BigInt::from(*n)),
+            (_, MoveValue::I16(n)) => Value::Number(BigInt::from(*n)),
+            (_, MoveValue::I32(n)) => Value::Number(BigInt::from(*n)),
+            (_, MoveValue::I64(n)) => Value::Number(BigInt::from(*n)),
+            (_, MoveValue::I128(n)) => Value::Number(BigInt::from(*n)),
+            (_, MoveValue::I256(n)) => Value::Number(n.to_string().parse::<BigInt>().unwrap()),
```
tzakian, line 400: `BigInt::from` doesn't work on i256?

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -3039,6 +3039,12 @@ impl<'d> serde::de::DeserializeSeed<'d> for SeedWrapper<&MoveTypeLayout> {
             L::Address => AccountAddress::deserialize(deserializer).map(Value::address),
             L::Signer => AccountAddress::deserialize(deserializer).map(Value::signer),
 
+            L::I8 | L::I16 | L::I32 | L::I64 | L::I128 | L::I256 => {
+                Err(<D::Error as serde::de::Error>::custom(
```
tzakian, line 3043: IMO these should be `todo!`s but realize it's kinda moot as it's just gonna get removed above this...

sui-execution/v0/sui-adapter/src/programmable_transactions/context.rs
```diff
@@ -1029,6 +1029,16 @@ mod checked {
...
             TypeTag::U256 => Type::U256,
+            // Signed integer types are not supported at the Sui layer for this
+            // execution version, and the VM runtime has no signed integer types.
+            TypeTag::I8
+            | TypeTag::I16
+            | TypeTag::I32
+            | TypeTag::I64
+            | TypeTag::I128
+            | TypeTag::I256 => {
+                return verification_error(StatusCode::TYPE_RESOLUTION_FAILURE);
```
tzakian, line 1040: IMO for non-latest versions these should be `unreachable!()`

tzakian (commented):
One of the main/overarching comments (but probably too late for this) is that the errors added here should probably be `todo!`s...

Left a bunch of comments, until skipping over the sui-side changes as otherwise it'll not be really useful. I think most of my comments are just mentioning things that will be fixed up higher in the stack.

Main things is `unreachable!` in non-latest versions, and a question on `BigInt`

PR #26223, opened 2026-04-13
[trace-view] Added support for reading large traces

awelc:
## Description 

This PR adds support for handling large traces. It replaces previous archiver with a more capable one and implements decoding/uncompressing of the trace in chunks so that we keep peak memory footprint low. 

We don't actually have to stream-process the trace as during parsing we filter a lot of events that would otherwise contribute to memory consumption. This allowed us to keep the changes ex to trace parsing part of the implementation.

## Test plan 

Tested on otherwise failing trace for mainnet `EfWjbxdSDotgj8ZusSuqXpU99EKhAmN92gf1bqa1SHg5` transaction. Also verified that everything works with smaller traces (e.g., with the `HKb8Ji74mNE8U6tH46M9znvgq7srztNbdAEVNatmchLM` mainnet transaction)


tzakian (approved):
Love it!

PR #26227, opened 2026-04-13
[move][move-compiler] Diagnostic prefixes turned into families, interned

cgswords:
## Description 

Toward reworking allow lists, intern prefixes as u8s (also, rename the families to help my brain a bit). This is a small change, but should allow us to do some fun optimizations in the next PR.

## Test plan 

Everything still works.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
+++ b/external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
@@ -9,7 +9,10 @@ use crate::{
     Flags,
     command_line::COLOR_MODE_ENV_VAR,
     diagnostics::{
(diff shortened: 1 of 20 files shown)


external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
damirka, line 694: The build will fail somewhere, right? I don't see the use of this function being changed in this PR

PR #26229, opened 2026-04-14
sui-indexer-alt-framework: harden streaming backoff bounds

Rifvckzie13:
## Description

Validate streaming backoff configuration at service construction time and harden the fallback math used by the ingestion broadcaster.

This change rejects zero or inverted backoff settings, uses saturating arithmetic for fallback and buffer-window calculations, and clamps fallback progress so retries always advance by at least one checkpoint.

## Test plan

- cargo fmt -p sui-indexer-alt-framework --check
- cargo test -p sui-indexer-alt-framework --lib reject_
- cargo test -p sui-indexer-alt-framework --lib setup_streaming_task_
- cargo test -p sui-indexer-alt-framework --lib streaming_connection_

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [x] Indexing Framework: Validate streaming backoff config and use saturating fallback math so retries cannot stall without progress.


amnn (commented):
Hi @Rifvckzie13, thanks for the contribution -- could you share some more details for the motivation of this change? Was there something that you were trying to do that caused you to run into issues?

Typically if there was an issue, we would handle it by validating inputs before the service starts so that we didn't need to keep interacting with the parameters in a defensive fashion while the service is running.

Rifvckzie13:
@amnn Thanks. This came from code inspection rather than a production incident.

The main issue I wanted to address was that the streaming backoff settings are externally configured but were not validated at service construction time. On the pre-change path, if `streaming_backoff_initial_batch_size = 0`, a streaming connection failure can fall back to an empty ingestion range (`ingestion_end == checkpoint_hi`), so the broadcaster can keep retrying without making forward progress. Similarly, `streaming_backoff_max_batch_size < streaming_backoff_initial_batch_size` violates the intended backoff invariant.

Would you prefer this PR to focus only on the startup validation and tests?

amnn:
Got it -- yes it would be good to focus on each validation parameter that needed to be validated in this way in their own PR with their own tests. For `streaming_backoff_initial_batch_size`, it sounds like we can prevent the undesirable behaviour by using representing it using `NonZeroUsize` in the config? Then when we start up the broadcaster, we can convert it into a `usize` and we will know that it's fine.

Rifvckzie13:
@amnn Done, thanks. I narrowed this to `streaming_backoff_initial_batch_size` using `NonZeroUsize` and added the focused deserialization test;

amnn (approved):
Looks great, thank you!

PR #26230, opened 2026-04-14
[sui-indexer-alt-reader] Add gRPC metrics middleware and refactor clients to use it

wlmyng:
## Description 

`LedgerGrpcReader` and `FullnodeClient` have the same metrics, just renamed differently, so this PR consolidates the boilerplate into a `GrpcMetricsLayer` middleware that both clients now wrap their channels with.

## Potential followups

- Refactor `FullnodeClient` to push `Option` to the caller, removing `Error::NotConfigured`. Callers hold `Option<FullnodeClient>` and decide at the boundary whether the feature is configured.    
- `requests_cancelled` counter, which we can support by replacing the `boxFuture` with a `pin_project` setup - something like                  crates/sui-indexer-alt-jsonrpc/src/metrics/middleware.rs                           
- `TimeoutLayer` 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-reader/src/fullnode_client.rs
+++ b/crates/sui-indexer-alt-reader/src/fullnode_client.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::sync::Arc;
-
 use anyhow::Context;
 use prometheus::Registry;
 use prost_types::FieldMask;
@@ -13,10 +11,12 @@ use sui_types::transaction::Transaction;
 use sui_types::transaction::TransactionData;
 use tonic::transport::Channel;
 use tonic::transport::ClientTlsConfig;
+use tower::Layer;
 use tracing::instrument;
 use url::Url;
 
-use crate::metrics::FullnodeClientMetrics;
+use crate::metrics::GrpcMetricsLayer;
+use crate::metrics::GrpcMetricsService;
 
 #[derive(clap::Args, Debug, Clone)]
 pub struct FullnodeArgs {
@@ -28,8 +28,7 @@ pub struct FullnodeArgs {
 /// A client for executing and simulating transactions via the full node gRPC service.
 #[derive(Clone)]
 pub struct FullnodeClient {
-    execution_client: TransactionExecutionServiceClient<Channel>,
-    metrics: Arc<FullnodeClientMetrics>,
+    execution_client: TransactionExecutionServiceClient<GrpcMetricsService<Channel>>,
 }
 
 #[derive(thiserror::Error, Debug)]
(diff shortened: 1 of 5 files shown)


crates/sui-indexer-alt-reader/src/fullnode_client.rs
```diff
@@ -106,12 +104,15 @@ impl FullnodeClient {
...
-            self.execution_client.clone(),
-            |mut client| async move { client.execute_transaction(request).await },
-        )
-        .await
+        let Some(mut client) = self.execution_client.clone() else {
+            return Err(Error::NotConfigured);
+        };
+
+        client
+            .execute_transaction(request)
```
amnn:
```suggestion
        self.execution_client
            .clone()
            .ok_or(Error::NotConfigured)?
            .execute_transaction(request)
```

crates/sui-indexer-alt-reader/src/metrics.rs
```diff
@@ -149,69 +158,89 @@ impl FullnodeClientMetrics {
...
+        let metrics = self.metrics.clone();
+        metrics
+            .requests_received
+            .with_label_values(&[&method])
+            .inc();
+        let timer = metrics.latency.with_label_values(&[&method]).start_timer();
+
+        let fut = self.inner.call(req);
+        Box::pin(async move {
+            let result = fut.await;
```
amnn, line 267: Do you need to handle the case where the future is dropped while it is awaiting here? E.g. if a request is dropped while it's waiting on a request, then you'll see a mismatch between requests received and succeeded/failed, and the timer will not ever return.

amnn (commented):
Just need to handle the "requests canceled" case -- there are a couple of examples in the codebase of how to setup the guard object to deal with this.

crates/sui-indexer-alt-reader/src/metrics.rs
wlmyng: Ah yes, I think I can handle that with a drop guard

crates/sui-indexer-alt-reader/src/metrics.rs
```diff
@@ -37,19 +44,42 @@ pub(crate) struct ConsistentReaderMetrics {
...
+    metrics: Arc<GrpcMetrics>,
+}
+
+struct MetricsGuard {
+    metrics: Arc<GrpcMetrics>,
+    method: String,
+    timer: Option<HistogramTimer>,
+}
+
+impl MetricsGuard {
```
amnn: nit: file order

amnn (approved):
LGTM!

PR #26231, opened 2026-04-14
[move] Thread MoveFlavor instance through package pipeline

mdgeorge4153:
## Summary

- Change `MoveFlavor` trait methods from associated functions to instance methods (`&self`)
- Make `system_deps` and `implicit_dependencies` async to support future RPC-based queries
- Thread `Arc<F>` through the package loading pipeline from `PackageLoader::load()` down to dependency pinning and legacy manifest parsing

This is a prerequisite for on-chain bytecode dependencies (which need the flavor instance to fetch packages via RPC) and for fixing DVX-1814 (network-aware system dependency resolution).

## Test plan

- [ ] CI passes — pure refactor with no behavior change


crates/sui-move/src/cache_package.rs
```diff
@@ -37,7 +37,7 @@ impl CachePackage {
             name: self.environment_name.clone(),
             id: self.environment_id.clone(),
         };
-        let info = cache_package::<SuiFlavor>(&env, &dep.dep).await?;
+        let info = cache_package::<SuiFlavor>(&env, &dep.dep, &SuiFlavor).await?;
```
mdgeorge4153: You can probably drop the turbofishes here and elsewehere because the `SuiFlavor` is obvious from context.

crates/sui-move/src/coverage.rs
```diff
@@ -21,7 +21,7 @@ impl Coverage {
         build_config: BuildConfig,
     ) -> anyhow::Result<()> {
         self.coverage
-            .execute::<SuiFlavor>(path, build_config)
+            .execute::<SuiFlavor>(path, build_config, Arc::new(SuiFlavor))
```
mdgeorge4153: Here's another place where dropping the turbofish seems reasonable

crates/sui/src/sui_commands.rs
```diff
@@ -786,7 +783,7 @@ impl SuiCommand {
             }
             SuiCommand::FireDrill { fire_drill } => run_fire_drill(fire_drill).await,
             SuiCommand::Analyzer => {
-                analyzer::run::<SuiFlavor>(Some(Flavor::Sui));
+                analyzer::run::<SuiFlavor>(Arc::new(SuiFlavor), Some(Flavor::Sui));
```
mdgeorge4153, line 786: I'm not sure where `Flavor::Sui` is defined, but I wonder if it makes sense to have a deterministic mapping so that only one of the flavors needs to be passed.

external-crates/move/crates/move-analyzer/src/symbols/runner.rs
```diff
@@ -67,12 +53,13 @@ impl SymbolicatorRunner {
     }
 
     /// Create a new runner
-    pub fn new<F: MoveFlavor>(
+    pub fn new<F: MoveFlavor + 'static>(
```
mdgeorge4153, line 57: Why is the `'static` needed here?

external-crates/move/crates/move-analyzer/src/code_action.rs
```diff
@@ -85,6 +85,7 @@ pub fn on_code_action_request<F: MoveFlavor>(
     request: &Request,
     ide_files_root: VfsPath,
     pkg_dependencies: Arc<Mutex<CachedPackages>>,
+    move_flavor: Arc<F>,
```
mdgeorge4153, line 88: I'm wondering - the `move_flavor` seems to trickle down to a whole bunch of places, is there already a context argument of some kind that we can add the flavor to instead?

external-crates/move/crates/move-cli/src/base/summary.rs
```diff
@@ -105,8 +107,9 @@ impl Summary {
             )
         } else {
             let path = reroot_path(path)?;
-            let env = find_env::<F>(&path, &config)?;
-            let root_pkg: RootPackage<F> = config.package_loader(&path, &env).load().await?;
+            let env = find_env::<F>(&path, &config, &flavor)?;
+            let root_pkg: RootPackage<F> =
+                config.package_loader(&path, &env).load(flavor).await?;
```
mdgeorge4153: I wonder if the flavor object should be part of the package loader or even the config. Does that make things simpler?

external-crates/move/crates/move-cli/src/base/test.rs
```diff
@@ -160,8 +164,11 @@ pub async fn run_move_unit_tests<F: MoveFlavor, V: VMTestSetup + Sync, W: Write
 
     // Load the package (package graph diagnostics are only needed for CLI commands so
     // ignore them by passing a vector as the writer)
-    let env = find_env::<F>(pkg_path, &build_config)?;
-    let root_pkg: RootPackage<F> = build_config.package_loader(pkg_path, &env).load().await?;
+    let env = find_env::<F>(pkg_path, &build_config, &flavor)?;
```
mdgeorge4153, line 167: Similarly, I wonder if putting a flavor field in a build_config is a good idea

external-crates/move/crates/move-package-alt/design/DESIGN.md
mdgeorge4153, line 1: While this move makes sense, I don't think it should be part of this PR

external-crates/move/crates/move-package-alt-compilation/src/lib.rs
```diff
@@ -97,8 +97,12 @@ use std::path::Path;
...
 /// is passed, it will try to find it in the list of available environments, and error if it cannot
 /// be found.
-pub fn find_env<F: MoveFlavor>(path: &Path, config: &BuildConfig) -> anyhow::Result<Environment> {
-    let envs = RootPackage::<F>::environments(path)?;
+pub fn find_env<F: MoveFlavor>(
+    path: &Path,
+    config: &BuildConfig,
+    flavor: &F,
+) -> anyhow::Result<Environment> {
+    let envs = RootPackage::<F>::environments(path, flavor)?;
```
mdgeorge4153, line 105: I don't think the `environments` command should need access to a flavor object, correct? In fact, I'm not even sure why `environments` is parameterized, maybe it should be moved outside of `RootPackage`

external-crates/move/crates/move-package-alt/src/cli/build.rs
```diff
@@ -36,7 +36,8 @@ impl Build {
     pub async fn execute(&self) -> anyhow::Result<()> {
         let path = self.path.clone().unwrap_or_else(|| PathBuf::from("."));
 
-        let envs = RootPackage::<Vanilla>::environments(&path)?;
+        let flavor = Vanilla;
```
mdgeorge4153: I'd just inline this

external-crates/move/crates/move-package-alt/src/cli/update_deps.rs
```diff
@@ -37,7 +37,8 @@ impl UpdateDeps {
     pub async fn execute(&self) -> anyhow::Result<()> {
         let path = self.path.clone().unwrap_or(PathBuf::from("."));
 
-        let envs = RootPackage::<Vanilla>::environments(&path)?;
+        let flavor = Vanilla;
```
mdgeorge4153: inline here as well

external-crates/move/crates/move-package-alt/src/compatibility/legacy_parser.rs
```diff
@@ -75,10 +75,11 @@ pub struct LegacyPackageMetadata {
 /// the unsupported sections: `[addresses]`, `[dev-addresses]`, or `[dev-dependencies]`. Although
 /// these fields are not technically required in the old system, we want to process manifests that
 /// don't have them using the modern parser.
-pub fn try_load_legacy_manifest<F: MoveFlavor>(
+pub async fn try_load_legacy_manifest<F: MoveFlavor>(
```
mdgeorge4153, line 78: I'm not sure why this needs to be async

mdgeorge4153: Oh I see, because we do implicit dependency checking as part of the loading, and the set of implicit deps could in principle require fetching from the network

external-crates/move/crates/move-package-alt/src/graph/builder.rs
```diff
@@ -60,13 +60,15 @@ struct PackageCache<F: MoveFlavor> {
 pub struct PackageGraphBuilder<'a, F: MoveFlavor> {
     cache: PackageCache<F>,
     config: &'a PackageConfig,
+    flavor: &'a F,
```
mdgeorge4153, line 63: Perhaps the `flavor` should be part of the `PackageConfig`

external-crates/move/crates/move-package-alt/src/package/package_loader.rs
```diff
@@ -12,6 +15,12 @@ pub struct PackageLoader {
     config: PackageConfig,
 }
 
+/// A [PackageLoader] paired with a [MoveFlavor] instance, ready to load a package.
+pub struct FlavoredLoader<F: MoveFlavor> {
```
mdgeorge4153: I think this should probably just be folded into `PackageLoader`

mdgeorge4153 (commented):
I think the main thing is that there are several places where we have a config object floating around already (buildconfig, packageconfig, packageloader), and adding the flavor object to the config will prevent adding a whole bunch of arguments everywhere.

crates/sui-move/src/cache_package.rs
```diff
@@ -37,7 +37,7 @@ impl CachePackage {
             name: self.environment_name.clone(),
             id: self.environment_id.clone(),
         };
-        let info = cache_package::<SuiFlavor>(&env, &dep.dep).await?;
+        let info = cache_package(&env, &dep.dep, &SuiFlavor).await?;
```
mdgeorge4153, line 40: It would be better to add the SuiFlavor constructor here instead of a later PR (SuiFlavor::new)

crates/sui-move/src/coverage.rs
```diff
@@ -21,7 +21,7 @@ impl Coverage {
         build_config: BuildConfig,
     ) -> anyhow::Result<()> {
         self.coverage
-            .execute::<SuiFlavor>(path, build_config)
+            .execute(path, build_config, Arc::new(SuiFlavor))
```
mdgeorge4153, line 24: I thought we were putting the flavor into the build_config, so that this code wouldn't change?

mdgeorge4153: Probably the Move BuildConfig

mdgeorge4153: Before doing this, check to see how invasive of a change it would be and ask me to make a call

external-crates/move/crates/move-analyzer/src/symbols/compilation.rs
```diff
@@ -423,7 +424,7 @@ pub fn get_compiled_pkg<F: MoveFlavor>(
         None
     };
 
-    let root_pkg = load_root_pkg::<F>(&build_config, pkg_path)?;
+    let root_pkg = load_root_pkg::<F>(&build_config, pkg_path, move_flavor)?;
```
mdgeorge4153, line 427: you can drop the turbofish here

external-crates/move/crates/move-cli/src/base/coverage.rs
```diff
@@ -145,13 +148,14 @@ impl Coverage {
...
         mut config: BuildConfig,
+        flavor: Arc<F>,
         differential: Option<String>,
         test: Option<String>,
     ) -> anyhow::Result<()> {
         // Make sure we always compile the package in test mode so we get correct source maps.
         config.test_mode = true;
         let package = config
-            .compile_package::<F, _>(&path, env, &mut Vec::new())
+            .compile_package::<F, _>(&path, env, flavor, &mut Vec::new())
```
mdgeorge4153, line 158: should be able to drop the turbofish here also

external-crates/move/crates/move-cli/src/base/summary.rs
```diff
@@ -105,8 +107,9 @@ impl Summary {
             )
         } else {
             let path = reroot_path(path)?;
-            let env = find_env::<F>(&path, &config)?;
-            let root_pkg: RootPackage<F> = config.package_loader(&path, &env).load().await?;
+            let env = find_env::<F>(&path, &config, &flavor)?;
```
mdgeorge4153, line 110: We changed it so `find_environment` doesn't need a flavor I think, does `find_env` need one?

external-crates/move/crates/move-package-alt-compilation/src/lib.rs
mdgeorge4153: Keeping `environments()` flavor-specific — it combines the flavor-provided defaults (mainnet/testnet for Sui) with manifest-declared environments. Both pieces are needed, and the flavor is the natural source for the defaults.

external-crates/move/crates/move-analyzer/src/symbols/runner.rs
mdgeorge4153: The \`'static\` bound is needed because the flavor is moved into a \`thread::spawn\` closure (which requires \`'static\` captures). There's a comment in the code explaining this.

external-crates/move/crates/move-analyzer/src/code_action.rs
mdgeorge4153: I looked into this — there is no existing context object that flows through these functions. The flavor threads through from `run()` → `on_request()` → `on_code_action_request()` etc. Adding a context struct is possible but would be a larger refactor of the move-analyzer request handling. Leaving as-is for now.

external-crates/move/crates/move-cli/src/base/summary.rs
mdgeorge4153: Investigated putting flavor in BuildConfig — it's a clap-derived struct with Serialize/Deserialize/Default/PartialEq, so adding a generic type param is very invasive (94 files). Keeping flavor separate from BuildConfig for now.

external-crates/move/crates/move-cli/src/base/test.rs
mdgeorge4153: Investigated putting flavor in BuildConfig — it's a clap-derived struct with Serialize/Deserialize/Default/PartialEq, so adding a generic type param is very invasive (94 files). Keeping flavor separate from BuildConfig for now.

crates/sui-move/src/coverage.rs
mdgeorge4153: Investigated putting flavor in BuildConfig — it's a clap-derived struct with Serialize/Deserialize/Default/PartialEq, so adding a generic type param is very invasive (94 files). Keeping flavor separate from BuildConfig for now.

external-crates/move/crates/move-package-alt/src/graph/builder.rs
mdgeorge4153: Agreed this would be cleaner. PackageConfig is internal (only 5 files), so the blast radius is small. Will do as a follow-up.

crates/sui/src/sui_commands.rs
mdgeorge4153: `Flavor::Sui` is in `move_compiler::editions::Flavor` — it's the compiler flavor, separate from `MoveFlavor`/`SuiFlavor` which is the package management flavor. A deterministic mapping would make sense but the two serve different purposes (compiler config vs package management state). Worth revisiting but out of scope for this PR.

external-crates/move/crates/move-cli/src/base/summary.rs
mdgeorge4153: Yes — `find_env` calls `RootPackage::environments()` which needs the flavor for default environments (mainnet/testnet). `find_environment` in sui-package-alt is Sui-specific and constructs `SuiFlavor` internally, so it doesn't need one from the caller.

external-crates/move/crates/move-package-alt/src/package/package_impl.rs
```diff
@@ -299,10 +305,11 @@ impl<F: MoveFlavor> Package<F> {
 }
 
 /// Ensure that the dependency given by `dep_info` is cached on disk, and return information
-/// about its publication in `env`
+/// about its publication in `env`. `flavor` is used for system dep resolution and validation.
 pub async fn cache_package<F: MoveFlavor>(
     env: &Environment,
     manifest_dep: &ManifestDependencyInfo,
+    flavor: &F,
```
stefan-mysten: This has both F: MoveFlavor and flavor, it's a bit confusing.

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
```diff
@@ -39,21 +39,25 @@ pub trait MoveFlavor: Debug + Send + Sync {
...
+    /// Return ALL the system dependencies for the requested `environment`.
+    async fn system_deps(
+        &self,
+        environment: &EnvironmentID,
+    ) -> BTreeMap<SystemDepName, LockfileDependencyInfo>;
 
-    /// Return the default system dependencies for the requested environment.
-    fn implicit_dependencies(
+    /// Return the default system dependencies for the requested `environment`.
+    async fn implicit_dependencies(
```
stefan-mysten, line 57: Why are these now async?

external-crates/move/crates/move-package-alt/src/package/package_impl.rs
```diff
@@ -237,20 +242,21 @@ impl<F: MoveFlavor> Package<F> {
...
+        flavor: &F,
     ) -> PackageResult<Vec<CombinedDependency>> {
-        let implicits = F::implicit_dependencies(env.id());
+        let implicits = flavor.implicit_dependencies(env.id()).await;
         let is_implicit = implicits.contains_key(manifest.package.name.as_ref());
 
         let system_dependencies = if manifest.package.implicit_dependencies && !is_implicit {
             debug!("adding implicit dependencies");
-            F::implicit_dependencies(env.id())
+            flavor.implicit_dependencies(env.id()).await
```
stefan-mysten: Why is this called again? Reuse the `implicits` variable defined above?

external-crates/move/crates/move-package-alt/src/graph/builder.rs
```diff
@@ -349,13 +361,15 @@ impl<F: MoveFlavor> PackageCache<F> {
         }
     }
 
-    /// Return a reference to a cached [Package], loading it if necessary
+    /// Return a reference to a cached [Package], loading it if necessary.
+    /// `flavor` is passed through to [Package::load].
     pub async fn fetch(
```
stefan-mysten, line 354: From claude
```
graph/builder.rs PackageCache::fetch — potential race in cache | warning                                                                                                                                                                                  
  Two concurrent callers could both pass the cell.get() check, both call Package::load(), and both call cell.get_or_init(). The loser's Arc is discarded, but the code returns the locally-constructed node rather than the stored one. Not currently          
  triggerable (sequential loop), but would surface if parallel fetching is enabled.                                                                                                                                                                            
  Fix: Return the value from get_or_init instead of the local node.         
```

external-crates/move/crates/move-package-alt/src/graph/mod.rs
```diff
@@ -128,7 +131,11 @@ impl<F: MoveFlavor> PackageGraph<F> {
     ///  - if `overrides` contains the package, we replace its publication
     ///  - if the package has a system address (see F::is_system_address), we keep it
     ///  - otherwise, we drop it
-    pub fn make_ephemeral(&mut self, overrides: BTreeMap<EphemeralDependencyInfo, Publication<F>>) {
+    pub fn make_ephemeral(
```
stefan-mysten, line 131: I think the calls to this method in tests need updating.

external-crates/move/crates/move-package-alt/design/DESIGN.md
stefan-mysten, line 1: README.md still references See DESIGN.md.

external-crates/move/crates/move-package-alt/src/package/package_loader.rs
```diff
@@ -8,8 +11,9 @@ use crate::{
 };
 
 /// A Builder for the [RootPackage] type
-pub struct PackageLoader {
+pub struct PackageLoader<F: MoveFlavor> {
     config: PackageConfig,
+    flavor: Arc<F>,
 }
```
stefan-mysten, line 16: Should this use a lifetime for flavor instead of an Arc?

stefan-mysten (commented):
Analyzer files use `move_flavor` (to disambiguate from `Option<Flavor>` compiler edition), while CLI files use `flavor`. Might be a good idea to fix this inconsistency in this PR.

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
mdgeorge4153: Because knowing the set of implicit dependencies (or the versions) should depend on the protocol version; to determine that we'll need to do an RPC request, which is async

external-crates/move/crates/move-package-alt/src/package/package_loader.rs
mdgeorge4153: No, I'm not entirely sure but I think the root package will still need to hold on to the flavor after it has been loaded - if this was a reference with a lifetime parameter, the root package would need to have a lifetime annotation too, and it starts to get out of hand.

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
stefan-mysten: So in the case of Vanilla flavor, it will return an empty map? What I am trying to make sure is that the move-part of code will never do an RPC call unless it's really needed.

Alternatively, this could be left sync and the Flavor implementation does a block_on to resolve the async data fetch.

external-crates/move/crates/move-package-alt/src/package/package_impl.rs
stefan-mysten: Ignore, just read it wrong.

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
mdgeorge4153: yes. It's impossible for the Move side to do an RPC call - it would need a dependency on the non-external-crates. Making it sync doesn't solve anything here - the move side could still just do a block_on and call an RPC anyway.

stefan-mysten: I was thinking of letting this be sync for keeping all APIs consistent. Not a blocker at all, was just curious.

mdgeorge4153: Yeah, IMO `block_on!` is a major code smell and should only be used if it saves you from asyncifying a huge part of the codebase; it shouldn't be baked into the design.

stefan-mysten (approved):

PR #26233, opened 2026-04-14
[move] Implement on-chain dependency pinning

mdgeorge4153:
## Summary

- `Pinned::OnChain` carries `published_at: PublishedID` — this is the source locator for on-chain deps, analogous to git SHA for git deps
- Fill in on-chain pinning: verify addresses present during pinning, construct `Pinned::OnChain { published_at }`
- `unfetched_path` works naturally for all `Pinned` variants including `OnChain` (cache path: `~/.move/on-chain/<chain_id>/<published_at>/`)
- `from_lockfile` takes `(&LockfileDependencyInfo, Option<&PublishAddresses>)` to extract `published_at` for on-chain deps
- Return `FetchError::OnChainNotSupported` when fetching on-chain deps (not yet implemented)
- `OnChainDepMissingAddress` error includes the environment name
- Add `design/dvx-2115-address-identity.md` documenting the relationship between source identity, compiled identity, and address identity (DVX-2115)

## Test plan

- [x] All existing tests pass
- [x] Clippy and rustfmt clean
- [ ] End-to-end test with on-chain dep (blocked on fetch implementation)

---

## PR Stack
2. ~~#26065 — Separate node/edge concepts for pinned dependencies~~ ✅ merged
3. **#26233 — On-chain dependency pinning** ← this PR

### DIFF
--- a/external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
+++ b/external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
@@ -1,0 +1,129 @@
+# Address Identity in the Dependency Graph
+
+## Problem (TODO)
+
+The dependency system distinguishes between "node" concepts (where a package's source
+lives) and "edge" concepts (how a parent references a dependency). The `Pinned` enum
+represents nodes, and `DependencyContext` represents edges. However, `published_at` and
+`original_id` don't fit cleanly into either category, and the current placement leads to
+design problems. The items below are not yet resolved.
+
+## Background: What Are Addresses?
+
+Every published Move package has two addresses:
+
+- **`original_id`**: The address the package was first published at. This serves as the
+  package's identity in the Move type system — types are identified by `original_id::module::Type`.
+- **`published_at`**: The address where the current version of the package lives on-chain.
+  After an upgrade, `published_at` changes but `original_id` stays the same.
+
+These addresses can come from several sources:
+
+1. **The package's own files** — for legacy packages, this may be in `Move.toml` or
+   `Move.lock`. For modern packages, it is in `Move.published`.
+2. **`dep-replacements` in the parent's manifest** — a parent can override the addresses
+   for a dependency, e.g. when the dependency's own manifest has incorrect or missing
+   address info.
+
+## The Identity Question
+
+Consider this scenario: a git package `foo` is published twice on the same network — once
+at `0xA` and once at `0xB`. These are genuinely different on-chain packages that happen to
+share the same source code. A downstream project could depend on both (e.g., to bridge
+between them).
+
+This is an unusual corner case and something users probably shouldn't do, but it's
+important because it exposes a fundamental question about the design model. We could
+detect this situation and fail, but the fact that we're tripping on these kinds of issues
+suggests we need to think more clearly about the relationship between source identity,
+compiled identity, and address identity.
+
+In the current system:
+- Both edges point to the same fetched source (same git SHA, same local path on disk)
+- But they have different `published_at` / `original_id` via `dep-replacements`
+- The compiled bytecode differs because the self-address differs
+- The type systems are disjoint — `0xA::module::T` and `0xB::module::T` are unrelated types
+
+### Are these the same node or different nodes?
+
+**From a source perspective**: same node. The files on disk are identical.
+
+**From a compiled/on-chain perspective**: different nodes. They produce different bytecode,
+have different type identities, and are independently upgradeable.
+
+## Design Decision
+
+Addresses are part of the **compiled identity** of a package, but not part of the **source
+identity**. This means:
+
+1. **Source fetching and caching** should be based on the source location alone (git SHA +
+   path, local path, etc.). Two deps with the same source but different addresses should
+   share the same fetched files on disk. The current source cache layout does not need to
+   change.
+
+2. **Compiled identity** (for ephemeral publication, linkage, and type checking) must
+   account for addresses. The same source with different address overrides produces
+   different compilation artifacts. While we don't currently cache compiled artifacts, it
+   seems plausible that we would do so in the future, and such a cache would need to be
+   keyed by (source location + addresses).
+
+3. **For on-chain dependencies**, `published_at` is additionally the **source locator** —
+   it determines which bytecode to fetch from the network. So for on-chain deps,
+   `published_at` is unambiguously part of both the source identity and the compiled
+   identity.
+
+## Implications for `Pinned`
+
+If we treat the same source with different addresses as different nodes in the dependency
+graph, then addresses become part of the node identity for **all** `Pinned` variants, not
+just `OnChain`. This would mean `PinnedLocalDependency`, `PinnedGitDependency`, etc. would
+all carry address information.
+
+`unfetched_path` (for source fetching) would ignore addresses, since the source cache is
+keyed by source location alone. But ephemeral publication keys, linkage, and any future
+compiled artifact cache would use addresses as part of the key.
+
+### Immediate step: `Pinned::OnChain` should carry `published_at`
+
+Since `published_at` determines *what to fetch* for on-chain deps (analogous to the git
+SHA for git deps), it belongs in the `Pinned` enum regardless of the broader design:
+
+```rust
+pub(crate) enum Pinned {
+    Local(PinnedLocalDependency),
+    Git(PinnedGitDependency),
+    OnChain { published_at: PublishedID },
+    Root(PackagePath),
+}
+```
+
+This lets `unfetched_path` work naturally for all variants without panics or sentinel
+values.
+
+### Ephemeral publication keys should include addresses
+
+Currently, ephemeral publication (`Pub.<env>.toml`) entries are keyed by
+`EphemeralDependencyInfo`, which wraps `LocalDepInfo` — just the absolute local path to the
+source. The `published_at` and `original_id` are stored as data alongside the key, not as
+part of the key itself.
+
+This means two deps with the same source but different address overrides would collide on
+the same key. If we support the same source at multiple addresses (whether we treat it as
+different nodes or not), the ephemeral publication key should incorporate the address
+override.
+
+### Linkage should validate address consistency
+
+DESIGN.md (lines 609-623) describes a check for conflicting addresses: "if two
+dependencies in the graph have the same original ID but different published addresses, then
+there is a conflict." This check is not yet implemented. With the model above, this
+remains important — two edges claiming the *same* `original_id` with *different*
+`published_at` for what should be the *same* compiled package is a conflict.
+
+## Status
+
+- [x] Put `published_at` into `Pinned::OnChain` (immediate, unblocks on-chain deps)
+- [ ] Decide whether addresses belong on all `Pinned` variants or only `OnChain`
+- [ ] Update ephemeral publication keys to account for address overrides
+- [ ] Implement the address consistency check from DESIGN.md
+- [ ] Consider whether `original_id` should also be in `Pinned::OnChain` or remain in context
--- a/external-crates/move/crates/move-package-alt/src/dependency/pin.rs
+++ b/external-crates/move/crates/move-package-alt/src/dependency/pin.rs
@@ -14,7 +14,7 @@ use crate::{
     git::{GitCache, GitError, GitTree},
     package::paths::PackagePath,
     schema::{
-        EnvironmentID, EnvironmentName, EphemeralDependencyInfo, LocalDepInfo,
+        ConstTrue, EnvironmentID, EnvironmentName, EphemeralDependencyInfo, LocalDepInfo,
         LockfileDependencyInfo, LockfileGitDepInfo, ManifestGitDependency, ModeName,
         OnChainDepInfo, PackageName, RenderToml, RootDepInfo,
     },
@@ -48,7 +48,7 @@ pub struct PinnedDependency {
 pub(crate) enum Pinned {
     Local(PinnedLocalDependency),
     Git(PinnedGitDependency),
-    OnChain(OnChainDepInfo),
+    OnChain,
     Root(PackagePath),
 }
 
@@ -95,7 +95,14 @@ impl PinnedDependency {
             let transformed = match dep.dep_info {
                 Resolved::Local(ref loc) => loc.clone().pin(parent, environment_id)?,
                 Resolved::Git(ref git) => git.pin().await?,
-                Resolved::OnChain(_) => todo!(),
+                Resolved::OnChain(_) => {
+                    if dep.context.addresses.is_none() {
+                        return Err(PackageError::OnChainDepMissingAddress {
+                            name: dep.context.name.to_string(),
+                        });
+                    }
+                    Pinned::OnChain
+                }
             };
 
             result.push(PinnedDependency {
@@ -193,7 +200,21 @@ impl PinnedDependency {
 
     /// Return the absolute path to the directory that this dependency would be fetched into.
     pub fn unfetched_path(&self, chain_id: &EnvironmentID) -> PathBuf {
-        self.dep_info.unfetched_path(chain_id)
+        match &self.dep_info {
+            Pinned::OnChain => {
+                let published_at = &self
+                    .context
+                    .addresses
+                    .as_ref()
+                    .expect("on-chain deps have addresses")
+                    .published_at;
+                PathBuf::from(move_command_line_common::env::MOVE_HOME.as_str())
+                    .join("on-chain")
+                    .join(chain_id)
+                    .join(published_at.to_string())
+            }
+            _ => self.dep_info.unfetched_path(chain_id),
+        }
     }
 }
 
@@ -209,7 +230,9 @@ impl Pinned {
         match &self {
             Pinned::Git(dep) => dep.inner.path_to_tree(),
             Pinned::Local(dep) => dep.absolute_path_to_package.clone(),
-            Pinned::OnChain(_dep) => todo!(),
+            Pinned::OnChain => {
+                unreachable!("on-chain deps need context; use PinnedDependency::unfetched_path")
+            }
             Pinned::Root(path) => path.path().to_path_buf(),
         }
     }
@@ -233,7 +256,7 @@ impl Pinned {
                     .clean(),
                 relative_path_from_root_package: loc.local.to_path_buf().clean(),
             })),
-            LockfileDependencyInfo::OnChain(chain) => Ok(Pinned::OnChain(chain.clone())),
+            LockfileDependencyInfo::OnChain(_) => Ok(Pinned::OnChain),
             LockfileDependencyInfo::Git(git) => Ok(Pinned::Git(git.clone().try_into()?)),
             LockfileDependencyInfo::Root(_) => Ok(Pinned::Root(PackagePath::new(
                 containing_file
@@ -257,7 +280,7 @@ impl Pinned {
                 let rev = fmt_truncated(git.inner.sha(), 6, 2);
                 format!(r#"git = "{repo}", path = "{path}", rev = "{rev}""#)
             }
-            Pinned::OnChain(_on_chain) => "on-chain = true".to_string(),
+            Pinned::OnChain => "on-chain = true".to_string(),
             Pinned::Root(_) => "local = \".\"".to_string(),
         }
     }
@@ -318,7 +341,7 @@ impl LocalDepInfo {
                 absolute_path_to_package: parent.unfetched_path(chain_id).join(&self.local).clean(),
                 relative_path_from_root_package: self.local.clean(),
             }),
-            Pinned::OnChain(_) => todo!(),
+            Pinned::OnChain => return Err(PackageError::OnChainLocalDep),
         };
 
         Ok(info)
@@ -342,7 +365,9 @@ impl From<Pinned> for LockfileDependencyInfo {
                 rev: git.inner.sha().clone(),
                 path: git.inner.path_in_repo().to_path_buf(),
             }),
-            Pinned::OnChain(on_chain) => Self::OnChain(on_chain),
+            Pinned::OnChain => Self::OnChain(OnChainDepInfo {
+                on_chain: ConstTrue,
+            }),
             Pinned::Root(_) => Self::Root(RootDepInfo { root: true }),
         }
     }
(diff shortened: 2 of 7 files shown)


mdgeorge4153 (commented):
[CLAUDE] Review notes:

1. **`from_lockfile` duplicates `from_lockfile_source`** — The local/git/root arms are copy-pasted between the two functions (lines 254-284 vs 298-316). `from_lockfile` could delegate to `from_lockfile_source` for the non-on-chain cases to avoid duplication.

2. **`from_lockfile_source` panics on on-chain** (line 296) — Since this returns `PackageResult`, it should return an error rather than panicking. Even though the current call site filters out on-chain deps first, a panic is unnecessarily harsh for a private helper.

3. **`name: "unknown"` in `from_lockfile`** (line 267) — When the lockfile is missing `address_override` for an on-chain dep, the error uses `"unknown"` as the name. Could we extract a better identifier from the `Pin` or the lockfile context (e.g., the package ID key from the lockfile iteration)?

4. **No tests** — There are no tests for the new pinning paths. A unit test for `PinnedOnChainDependency::cache_path()` and one for `from_lockfile` with an on-chain pin would help catch regressions.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -34,10 +34,33 @@ use super::{
...
+/// A pinned on-chain dependency, identified by its chain ID and published address.
+/// The cache directory is `~/.move/on-chain/<chain_id>/<published_at>/`.
+#[derive(Clone, Debug)]
+pub struct PinnedOnChainDependency {
+    /// The chain identifier for the network this package is on.
+    pub chain_id: EnvironmentID,
+    /// The address the package is published at on-chain.
+    pub published_at: PublishedID,
+    /// The original publishing address (may differ from `published_at` after upgrades).
+    pub original_id: OriginalID,
```
mdgeorge4153: All of this information is contained in the DependencyContext, so it shouldn't be duplicated here. I think this type is redundant and the `OnChain` variant of `Pinned` can just be a unit

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -88,7 +111,18 @@ impl PinnedDependency {
             let transformed = match dep.dep_info {
                 Resolved::Local(ref loc) => loc.clone().pin(parent)?,
                 Resolved::Git(ref git) => git.pin().await?,
-                Resolved::OnChain(_) => todo!(),
+                Resolved::OnChain(_) => {
+                    let addresses = dep.context.addresses.as_ref().ok_or_else(|| {
+                        PackageError::OnChainDepMissingAddress {
```
mdgeorge4153, line 106: I'm not sure whether pinning is the right phase for this check, but I think it's probably good.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -198,26 +232,69 @@ impl Pinned {
...
+                    .parent()
+                    .expect("files have parents")
+                    .to_path_buf(),
+            )?)),
+        }
+    }
+
+    /// Create a pinned dependency from just a `LockfileDependencyInfo` source (without the full
+    /// [Pin] context). Only valid for non-on-chain dependencies (local, git, root).
+    /// Panics if called with an on-chain source.
```
mdgeorge4153: I'm not sure why this is here - it's very weird to panic for one variant.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -336,7 +414,9 @@ impl From<Pinned> for LockfileDependencyInfo {
                 rev: git.inner.sha().clone(),
                 path: git.inner.path_in_repo().to_path_buf(),
             }),
-            Pinned::OnChain(on_chain) => Self::OnChain(on_chain),
+            Pinned::OnChain(_) => Self::OnChain(OnChainDepInfo {
+                on_chain: true.try_into().expect("true is ConstTrue"),
```
mdgeorge4153: This is wild. It should just be `ConstTrue`.

external-crates/move/crates/move-package-alt/src/errors/mod.rs
```diff
@@ -130,6 +130,15 @@ pub enum PackageError {
         "Packages with old-style Move.toml files cannot depend on new-style packages. See https://docs.sui.io/references/package-managers/package-manager-migration for instructions."
     )]
     LegacyDependsOnModern,
+
+    #[error(
+        "On-chain dependency `{name}` requires a `published-at` address. Add a \
```
mdgeorge4153: Until we decouple the `published-at` and `original-id` fields, the error message should say that both fields are required.

external-crates/move/crates/move-package-alt/src/errors/mod.rs
```diff
@@ -130,6 +130,15 @@ pub enum PackageError {
...
+         `[dep-replacements.<env>.{name}]` entry with `published-at` and `original-id` fields."
+    )]
+    OnChainDepMissingAddress { name: String },
+
+    #[error("On-chain packages cannot have local transitive dependencies")]
```
mdgeorge4153: I wonder if this should be an internal error instead of a user-facing error, since the system is generating the on-chain dep manifests and not the user.

mdgeorge4153:
[CLAUDE] TODO: The `From<Pinned> for EphemeralDependencyInfo` currently panics for `Pinned::OnChain`. We need to think through what should happen when on-chain deps appear during ephemeral publication: (1) skip them since they're already published, (2) return a clear error, (3) include them as-is using their cache path.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Already addressed in the current code — `Pinned::OnChain` is a unit variant.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Agreed — pinning seems like the natural place since that's where we transition from resolved deps (which may lack addresses) to pinned deps (which must have them for on-chain).

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Fixed in the next PR (#26234) — removed `Pinned::unfetched_path` entirely and inlined path computation at each call site. The `OnChain` variant now returns a sentinel path `<on-chain>` instead of panicking, and most callers no longer need the method at all.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Already fixed — uses `ConstTrue` directly.

external-crates/move/crates/move-package-alt/src/errors/mod.rs
mdgeorge4153: [CLAUDE] Already fixed — the error message now says "requires `published-at` and `original-id` addresses".

external-crates/move/crates/move-package-alt/src/errors/mod.rs
mdgeorge4153: [CLAUDE] Good point. For now this error is reachable during pinning when a user declares an on-chain dep without providing `published-at`/`original-id` in dep-replacements, so it is user-facing. Once we generate manifests automatically, this could become an internal error.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -193,7 +200,21 @@ impl PinnedDependency {
...
     /// Return the absolute path to the directory that this dependency would be fetched into.
     pub fn unfetched_path(&self, chain_id: &EnvironmentID) -> PathBuf {
-        self.dep_info.unfetched_path(chain_id)
+        match &self.dep_info {
+            Pinned::OnChain => {
+                let published_at = &self
+                    .context
+                    .addresses
+                    .as_ref()
+                    .expect("on-chain deps have addresses")
```
mdgeorge4153: Does this `expect` come from an invariant? If so, is that invariant documented in the `Pinned` type?

mdgeorge4153: It may also be worth adding more detail to the expect message to more clearly identify the invariant (e.g. pinned on-chain deps have addresses)

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -209,7 +230,9 @@ impl Pinned {
         match &self {
             Pinned::Git(dep) => dep.inner.path_to_tree(),
             Pinned::Local(dep) => dep.absolute_path_to_package.clone(),
-            Pinned::OnChain(_dep) => todo!(),
+            Pinned::OnChain => {
+                unreachable!("on-chain deps need context; use PinnedDependency::unfetched_path")
```
mdgeorge4153: This feels a little gross to me. If we need context to give the path for a dep, then we need context to give the path for a dep. Let's discuss

external-crates/move/crates/move-package-alt/src/errors/mod.rs
mdgeorge4153: Are you sure? I would expect the OnChainDepMissingAddress error in that case. My comment was about OnChainLocalDep

mdgeorge4153: [CLAUDE] You're right — my earlier reply confused the two errors. `OnChainLocalDep` fires when an on-chain package has a local transitive dependency, which is indeed a system-generated scenario (the user wouldn't create this). This should probably be an internal/debug error rather than a user-facing one. Will fix in #26234.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Yes — the invariant is that `Pinned::OnChain` is only constructed (in `PinnedDependency::pin`) after checking that `dep.context.addresses.is_some()`. The expect message should be more specific, e.g. "pinned on-chain deps always have addresses (verified during pinning)". The invariant isn't currently documented on the `Pinned` type — will add that and improve the expect message in #26234.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Agreed this is gross. The sentinel was added to avoid the `unreachable!()` panic you flagged earlier. Options I see:

1. **Return `Option<PathBuf>`** — callers that know they're not on-chain can unwrap, but it ripples through signatures
2. **Remove `Pinned::unfetched_path` entirely** — inline the match at each call site (there are only ~3 callers outside `PinnedDependency`), so the `OnChain` arm never needs to produce a path
3. **Keep the method but make it `debug_assert!(false)` for OnChain** — documents the expectation without crashing in release

Option 2 feels cleanest to me — the callers already guard against OnChain before calling. Happy to discuss.

mdgeorge4153: [CLAUDE] Fixed in latest push to #26234 — `Pinned::OnChain` now carries `published_at: PublishedID`, so `unfetched_path` works naturally for all variants with no sentinel or panic. Also added `design/address-identity.md` documenting the broader question of whether addresses should be part of node identity for all `Pinned` variants.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Fixed — `Pinned::OnChain` now carries `published_at` directly, so the `expect` is gone. The invariant is documented on the variant: `published_at` is always set during pinning (verified in `PinnedDependency::pin`).

mdgeorge4153: Please fix the documentation problems in this PR.

external-crates/move/crates/move-package-alt/design/address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+
+DESIGN.md (lines 609-623) describes a check for conflicting addresses: "if two
+dependencies in the graph have the same original ID but different published addresses, then
+there is a conflict." This check is not yet implemented. With the model above, this
+remains important — two edges claiming the *same* `original_id` with *different*
+`published_at` for what should be the *same* compiled package is a conflict.
+
+## Status
+
+- [ ] Put `published_at` into `Pinned::OnChain` (immediate, unblocks on-chain deps)
```
mdgeorge4153: We should check this off.

external-crates/move/crates/move-package-alt/design/address-identity.md
```diff
@@ -0,0 +1,129 @@
+# Address Identity in the Dependency Graph
+
+## Problem
```
mdgeorge4153: Maybe we could make it clearer that this is a TODO instead of something that's currently implemented.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -204,25 +216,32 @@ impl Pinned {
...
     /// Create a [Pinned] from a lockfile dependency info entry.
     ///
     /// `containing_file` is the lockfile that the dependency comes from, used to resolve relative
-    /// paths.
+    /// paths. `published_at` is required for on-chain dependencies and ignored for other types.
     pub(crate) fn from_lockfile(
         containing_file: FileHandle,
-        pin: &LockfileDependencyInfo,
+        source: &LockfileDependencyInfo,
+        published_at: Option<&PublishedID>,
```
mdgeorge4153: This doesn't seem right - isn't this already included in the `LockfileDependencyInfo`?

mdgeorge4153: The `Option` here that is only used depending on the variant of another argument is another code smell.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -48,12 +48,14 @@ pub struct PinnedDependency {
 pub(crate) enum Pinned {
     Local(PinnedLocalDependency),
     Git(PinnedGitDependency),
-    OnChain(OnChainDepInfo),
+    /// An on-chain dependency, identified by the address it is published at.
+    /// Invariant: `published_at` is always set during pinning (verified in
+    /// [PinnedDependency::pin]).
```
mdgeorge4153: I think it's better not to include the "verified in" part, because the fact that it's an invariant means it better be verified in all constructors, and talking about how an invariant is maintained is likely to become out of date.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -233,7 +255,14 @@ impl Pinned {
                     .clean(),
                 relative_path_from_root_package: loc.local.to_path_buf().clean(),
             })),
-            LockfileDependencyInfo::OnChain(chain) => Ok(Pinned::OnChain(chain.clone())),
+            LockfileDependencyInfo::OnChain(_) => {
+                let published_at = published_at
+                    .ok_or_else(|| PackageError::OnChainDepMissingAddress {
+                        name: "unknown".to_string(),
```
mdgeorge4153: I don't love the "unknown" that gets generated here - it makes an unactionable and confusing error message for the user. But I understand how maybe fixing it is hard. Is there any easy way to improve it?

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -298,27 +329,32 @@ impl TryFrom<LockfileGitDepInfo> for PinnedGitDependency {
 
 impl LocalDepInfo {
     /// Takes a local dependency and its `parent`, and transforms local dependencies based on their
-    /// parent. `chain_id` is passed through to [Pinned::unfetched_path].
+    /// parent.
     ///
     /// 1. If the parent is a git dependency, we convert local transitive deps to git.
     /// 2. If the parent is a local dependency, we normalize the path based on the parents.
```
mdgeorge4153, line 339: Maybe a description for on-chain deps is warranted here, just for completeness (i.e. 3. The parent cannot be an on-chain dependency because on-chain packages can't have local deps.)

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -351,6 +389,9 @@ impl From<Pinned> for LockfileDependencyInfo {
 impl Pinned {
     /// Convert to an [EphemeralDependencyInfo]. For ephemeral dependencies, local dependencies
     /// (including the root) are stored as absolute paths.
+    ///
+    /// Only valid for Local, Git, and Root variants. On-chain dependencies cannot be represented
+    /// as ephemeral local paths.
```
mdgeorge4153: Why? This doesn't seem right.

external-crates/move/crates/move-package-alt/src/errors/mod.rs
```diff
@@ -130,6 +130,12 @@ pub enum PackageError {
         "Packages with old-style Move.toml files cannot depend on new-style packages. See https://docs.sui.io/references/package-managers/package-manager-migration for instructions."
     )]
     LegacyDependsOnModern,
+
+    #[error(
+        "On-chain dependency `{name}` requires `published-at` and `original-id` addresses. Add a \
```
mdgeorge4153, line 135: It might be useful to include the actual `env` as part of the error message. If it's hard, don't bother.

external-crates/move/crates/move-package-alt/design/address-identity.md
mdgeorge4153: [CLAUDE] Done — checked off.

external-crates/move/crates/move-package-alt/design/address-identity.md
mdgeorge4153: [CLAUDE] Done — changed heading to "Problem (TODO)" and added a note that the items below are not yet resolved.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Good catch — `LockfileDependencyInfo::OnChain` only contains `{on-chain: true}`, not the address. The `published_at` is in `Pin.address_override`. Changed `from_lockfile` to take `&Pin` directly, and added a separate `from_lockfile_source` for non-on-chain deps (used by system deps).

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Good point — removed the "verified in" part. The invariant comment now just states the invariants without referencing specific constructors.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] This path is only reached when loading from a malformed lockfile (missing `address_override` for an on-chain dep). Changed the placeholder to `"(from lockfile)"` to make the source clearer. In the normal pinning path, the real package name and env are included.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Done — added item 3 documenting the on-chain case.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] You're right — the doc was misleading. The issue is that `canonicalize()` requires the path to exist on disk, which won't be true for on-chain deps until they're fetched. Updated the doc to explain that this requires the source to be fetched to disk and that on-chain packages can't currently be used as ephemeral publication keys.

external-crates/move/crates/move-package-alt/src/errors/mod.rs
mdgeorge4153: [CLAUDE] Done — added `env` field to the error variant. The pinning path passes the actual environment ID; the lockfile path uses `"(unknown)"` since the env isn't readily available there.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153, line 1: It is probably better to create a linear issue in the backlog for the package management linear and use the ticket ID in the file name

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -342,15 +384,22 @@ impl From<Pinned> for LockfileDependencyInfo {
...
 }
 
 impl Pinned {
-    /// Convert to an [EphemeralDependencyInfo]. For ephemeral dependencies, local dependencies
-    /// (including the root) are stored as absolute paths.
+    /// Convert to an [EphemeralDependencyInfo] for use as a key in ephemeral publication files.
+    /// The path is canonicalized so that ephemeral pubfiles can be reused across packages in
+    /// a single filesystem.
+    ///
+    /// Requires that the package source has already been fetched to disk (so the path can be
```
mdgeorge4153, line 400: This is still incorrect, the on-chain deps aren't special here. We just haven't implemented on-chain dep fetching yet. If this is causing tests to fail, maybe we need to add the `mkdir` step to `fetch` for on-chain deps before failing, but I don't think even that is necessary.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: [CLAUDE] Will create a Linear ticket and rename the file.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Fixed — removed the special-casing language. On-chain deps aren't different here, we just haven't implemented fetching yet.

mdgeorge4153:
[CLAUDE] Alternative approach: #26552 uses separate `OnChainFlag` and `OnChainAddress` types instead of a single `OnChainDepInfo` with `address_override`. Compare both before merging.

mdgeorge4153:
Closed in favor of #26552

PR #26234, opened 2026-04-14
[move] Add stub generation from bytecode for on-chain dependencies

mdgeorge4153:
## Summary

- New `stubs` module in `move-package-alt` that renders Move source stubs from compiled bytecode
- Given a `normalized::Module`, produces source with `native fun` declarations for public/package functions and full definitions for structs and enums
- Handles: abilities (`has` clause with commas, constraints with `+`), phantom type params, entry modifier, references, generics, multiple return values
- Includes snapshot test using pre-compiled bytecode fixture (`tests/data/stub_test/example.mv`)
- Also addresses review comments from #26065 and #26233: removes `unreachable!()` panics, fixes doc comments, removes unused `chain_id` parameters from `fetch`, `LocalDepInfo::pin`, `to_ephemeral`, and `make_ephemeral`

## Test plan

- [x] All 176 tests pass
- [x] Clippy clean

---

## PR Stack
2. #26065 — Separate node/edge concepts for pinned dependencies
4. **#26234 — Stub generation from bytecode** ← this PR


external-crates/move/crates/move-package-alt/src/stubs.rs
```diff
@@ -0,0 +1,331 @@
...
+pub fn render_module(address: &AccountAddress, module: &normalized::Module<Symbol>) -> String {
+    let mut out = String::new();
+
+    writeln!(
+        &mut out,
+        "/// Auto-generated stub for on-chain package at {}",
+        address.to_canonical_string(true)
+    )
+    .unwrap();
+    writeln!(&mut out, "module {}::{} {{", address.to_canonical_string(true), module.name()).unwrap();
```
mdgeorge4153: We could conceivably use the package name instead of the address literal here, but I think maybe the address literal is preferable?

external-crates/move/crates/move-package-alt/src/stubs.rs
```diff
@@ -0,0 +1,331 @@
...
+    file_format::{AbilitySet, DatatypeTyParameter, Visibility},
+    normalized::{self, Type},
+};
+use move_core_types::account_address::AccountAddress;
+use move_symbol_pool::Symbol;
+
+/// Render a single compiled `module` as a Move source stub.
+/// The module declaration uses the given `address` for its address.
+pub fn render_module(address: &AccountAddress, module: &normalized::Module<Symbol>) -> String {
+    let mut out = String::new();
```
mdgeorge4153: Is it more idiomatic to take a writer as a parameter than returning a string?

external-crates/move/crates/move-package-alt/src/stubs.rs
```diff
@@ -0,0 +1,331 @@
...
+fn render_enum(out: &mut String, e: &normalized::Enum<Symbol>) {
+    let tparams = render_datatype_type_params(&e.type_parameters);
+    let abilities = render_has_abilities(&e.abilities);
+    let has_clause = if abilities.is_empty() {
+        String::new()
+    } else {
+        format!(" has {abilities}")
+    };
+
+    writeln!(out, "    public enum {}{tparams}{has_clause} {{", e.name).unwrap();
```
mdgeorge4153: Consider factoring out the common code between this and `render_struct`

external-crates/move/crates/move-package-alt/src/stubs.rs
```diff
@@ -0,0 +1,331 @@
...
+/// Render a function as a `native fun` declaration.
+fn render_function(out: &mut String, f: &normalized::Function<Symbol>) {
+    let vis = match f.visibility {
+        Visibility::Public => "public ",
+        Visibility::Friend => "public(package) ",
+        Visibility::Private => "",
+    };
+    let entry = if f.is_entry { "entry " } else { "" };
+
+    let tparam_names: Vec<String> = (0..f.type_parameters.len())
```
mdgeorge4153: can you reuse `datatype_tparam_names` here?

external-crates/move/crates/move-package-alt/src/stubs.rs
```diff
@@ -0,0 +1,331 @@
...
+    };
+
+    writeln!(
+        out,
+        "    {vis}{entry}native fun {}{tparams}({}){ret};",
+        f.name,
+        params.join(", ")
+    )
+    .unwrap();
+}
```
mdgeorge4153: This function is too long for my taste; I would prefer to see it subdivided further.

external-crates/move/crates/move-package-alt/src/stubs.rs
```diff
@@ -0,0 +1,331 @@
...
+                format!("{phantom}T{i}")
+            } else {
+                format!("{phantom}T{i}: {constraints}")
+            }
+        })
+        .collect();
+    format!("<{}>", rendered.join(", "))
+}
+
+/// Render an ability set as a list of ability names.
```
mdgeorge4153: Would it be simpler to just return a single string? I doubt we use the vector for anything other than passing it to `render_has_abilities`

external-crates/move/crates/move-package-alt/src/stubs.rs
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    /// Render the pre-compiled `example.mv` test fixture and snapshot the output. The fixture is
+    /// built from `tests/data/stub_test/sources/example.move`; if you change the source, rebuild
+    /// the fixture with `cargo run -p move-cli -- build --path tests/data/stub_test` and copy
+    /// `build/stub_test/bytecode_modules/example.mv` back to `tests/data/stub_test/example.mv`.
+    #[test]
+    fn render_example_stub() {
+        let bytes = include_bytes!("../tests/data/stub_test/example.mv");
```
mdgeorge4153: surprised to see `include_bytes` here instead of reading the file dynamically; is that idiomatic?

external-crates/move/crates/move-package-alt/tests/data/stub_test/sources/example.move
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// A small module that exercises various Move features for stub generation testing.
+module 0x1::example {
```
mdgeorge4153: We should use a less special address like `0xcafe`

external-crates/move/crates/move-package-alt/tests/data/stub_test/example.mv
mdgeorge4153, line 1: We should include information on how `example.mv` was generated. Ideally, it would be part of the test itself (so that `example.mv` is generated instead of checked in). But if that's not possible, we should at least include a README explaining how to rebuild `example.mv` in case we need to change it.

mdgeorge4153: (and ideally a note on why that can't be part of the test)

external-crates/move/crates/move-package-alt/src/stubs/snapshots/move_package_alt__stubs__move_source__tests__render_example_stub.snap
```diff
@@ -0,0 +1,34 @@
+---
+source: crates/move-package-alt/src/stubs.rs
+assertion_line: 329
+expression: rendered
+---
+/// Auto-generated stub for on-chain package at 0x0000000000000000000000000000000000000000000000000000000000000001
+module 0x0000000000000000000000000000000000000000000000000000000000000001::example {
+    public struct Coin has copy, drop, store {
```
mdgeorge4153, line 8: There's no struct or enum with no abilities

external-crates/move/crates/move-package-alt/src/snapshots/move_package_alt__stubs__tests__render_example_stub.snap
```diff
@@ -0,0 +1,34 @@
...
+    public enum Color has copy, drop {
+        Red,
+        Green,
+        Blue {
+            r: u8,
+            g: u8,
+            b: u8,
+        },
+    }
+    public native fun create_coin(p0: u64): 0x0000000000000000000000000000000000000000000000000000000000000001::example::Coin;
```
mdgeorge4153: We should consider using a name here (e.g. `self`) - in the generated `Move.toml` we would always name the package `self`

external-crates/move/crates/move-package-alt/src/stubs.rs
```diff
@@ -0,0 +1,331 @@
+// Copyright (c) Mysten Labs, Inc.
```
mdgeorge4153: We're not generating `Move.toml` files - is that a future PR?

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153, line 1: this should probably be `stubs/move.rs`, and we should have `stubs/manifest.rs`

external-crates/move/crates/move-package-alt/src/snapshots/move_package_alt__stubs__tests__render_example_stub.snap
```diff
@@ -0,0 +1,34 @@
...
+    public native fun coin_value(p0: &0x0000000000000000000000000000000000000000000000000000000000000001::example::Coin): u64;
+    public native fun destroy_coin(p0: 0x0000000000000000000000000000000000000000000000000000000000000001::example::Coin): u64;
```
mdgeorge4153: coin_value and destroy_coin test the same thing. There is no test with multiple arguments, parameterized functions, etc.

external-crates/move/crates/move-package-alt/tests/data/stub_test/example.mv
mdgeorge4153: thinking about this more, we should definitely make this more mechanical, and split it into a large test suite that individually hits a variety of move constructs. This suggests to me that we want to split stubs out into its own crate.

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/tests/data/stub_test/sources/example.move
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/tests/data/stub_test/example.mv
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/src/stubs/snapshots/move_package_alt__stubs__move_source__tests__render_example_stub.snap
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153: Addressed in latest push.

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153: Agreed — address literal is correct here since the generated stub uses the on-chain address. Using a package name like `self` would require the generated Move.toml to define a named address mapping, which is the manifest generation work in a future PR.

external-crates/move/crates/move-package-alt/src/stubs.rs
mdgeorge4153: Yes — Move.toml generation will be in `stubs/manifest.rs` in a future PR (Step 4 of the on-chain deps plan).

external-crates/move/crates/move-package-alt/src/snapshots/move_package_alt__stubs__tests__render_example_stub.snap
mdgeorge4153: Good idea. This requires coordination with the generated Move.toml (which would define the `self` named address). Will address when we add manifest generation.

external-crates/move/crates/move-package-alt/src/snapshots/move_package_alt__stubs__tests__render_example_stub.snap
mdgeorge4153: Added `transfer(coin, to, amount)` for multi-arg, `swap<T: copy + drop>` for constrained type params, and `Empty` struct with no abilities.

mdgeorge4153:
[CLAUDE] Replaced by #26555 — retargeted against main since this doesn't depend on onchain-pinning.

PR #26235, opened 2026-04-14
Add addr_bal_deposit load test workload

mystenmark:
## Summary
- Adds a purpose-built benchmark workload (`addr_bal_deposit`) for testing address balance deposit parallelism
- Each sender's address balance is seeded from a gas coin during init (1M SUI per sender in stress, 100K SUI in simtest); after that, all transactions use address-balance gas with no coin objects
- Transaction flow per target: `balance_from_sender` withdrawal → `balance::redeem_funds` → `balance::send_funds`
- Supports multiple target addresses per transaction (e.g. deposit to 4 accounts in one tx)
- All stress workers deposit to the same target(s) to evaluate concurrent deposit throughput
- Adds `--deposit-target-address` CLI flag to the stress binary (comma-separated for multiple targets)
- Adds `accumulator_deposits` and `accumulator_withdrawals` prometheus counters to `AuthorityMetrics`, incremented in the settlement scheduler
- Includes simtest with 4 target addresses verifying >100 successes with zero aborts/failures, plus on-chain balance verification for all targets

## Usage
```bash
# Single target
stress bench --deposit-target-address 0xABC --target-qps 1000 --num-workers 12

# Multiple targets (each tx deposits to all 4)
stress bench --deposit-target-address 0xABC,0xDEF,0x123,0x456 --target-qps 1000
```

## Test plan
- [x] `test_addr_bal_deposit_workload` simtest passes with 4 target addresses
- [x] All 4 target address balances verified non-zero after test
- [x] Accumulator deposit logs confirmed firing on all validators in simtest
- [x] `test_composite_workload` and `test_composite_workload_shared_randomness` pass (no regressions)
- [x] `cargo xclippy` clean
- [x] Stress binary CLI path traced end-to-end: same code path as simtest
- [ ] Run on live test network, verify `accumulator_deposits` metric increments

### DIFF
--- a/crates/sui-benchmark/src/workloads/addr_bal_deposit.rs
+++ b/crates/sui-benchmark/src/workloads/addr_bal_deposit.rs
@@ -24,7 +24,7 @@ const GAS_BUDGET: u64 = 50_000_000;
 
 #[derive(Clone)]
 pub struct AddrBalDepositConfig {
-    pub target_address: SuiAddress,
+    pub target_addresses: Vec<SuiAddress>,
     pub deposit_amount: u64,
     pub seed_amount: u64,
     pub metrics: Option<Arc<Mutex<AddrBalDepositMetrics>>>,
@@ -33,7 +33,7 @@ pub struct AddrBalDepositConfig {
 impl std::fmt::Debug for AddrBalDepositConfig {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("AddrBalDepositConfig")
-            .field("target_address", &self.target_address)
+            .field("target_addresses", &self.target_addresses)
             .field("deposit_amount", &self.deposit_amount)
             .field("seed_amount", &self.seed_amount)
             .finish()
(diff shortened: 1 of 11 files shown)


crates/sui-benchmark/src/workloads/addr_bal_deposit.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        let workload_params = WorkloadParams {
+            group,
+            target_qps,
+            num_workers,
+            max_ops,
+            duration,
+        };
+
+        let _ = reference_gas_price;
```
alex-mysten: remove from params?

crates/sui-benchmark/tests/simtest.rs
```diff
@@ -1727,6 +1730,151 @@ mod test {
         assert!(shared_plus_randomness_cancellations > 0);
     }
 
+    #[sim_test(config = "test_config()")]
+    async fn test_addr_bal_deposit_workload() {
```
alex-mysten, line 1735: what's the rationale for including this as a simtest? my understanding is that this is more useful as a performance benchmark, but we're not asserting anything about perf

alex-mysten (approved):
two small comments

crates/sui-benchmark/tests/simtest.rs
mystenmark: Yes, but we need a test that shows that the workload is generating valid transactions that execute successfully. (this was useful to get claude to produce working code and it will probably be needed to prevent bitrot).

PR #26237, opened 2026-04-14
`simulate_transaction`: recommend 0 gas price for gasless-eligible tx

aschran:
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/grpc_simulate_gasless_tests.rs
+++ b/crates/sui-e2e-tests/tests/grpc_simulate_gasless_tests.rs
@@ -1,0 +1,217 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Tests for the gRPC SimulateTransaction auto-gasless behavior.
+//!
+//! When the caller submits an unresolved transaction that is gasless-eligible and does not set
+//! an explicit `gas_payment.price`, the simulate flow should auto-switch to `price = 0` so that
+//! the returned resolved transaction reflects the gasless shape.
+
+use move_core_types::identifier::Identifier;
+use move_core_types::language_storage::TypeTag;
+use sui_macros::sim_test;
+use sui_rpc::proto::sui::rpc::v2::SimulateTransactionRequest;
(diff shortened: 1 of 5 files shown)


crates/sui-e2e-tests/tests/grpc_simulate_gasless_tests.rs
```diff
@@ -0,0 +1,220 @@
...
+
+async fn connect(
+    test_env: &TestEnv,
+) -> TransactionExecutionServiceClient<tonic::transport::Channel> {
+    TransactionExecutionServiceClient::connect(test_env.cluster.rpc_url().to_owned())
+        .await
+        .unwrap()
+}
+
+#[cfg_attr(not(msim), ignore)]
```
bmwill: why ignore when not msim? I find this is actually harmful to those who don't always run everythng under msim

aschran: removed

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -92,64 +98,88 @@ pub fn simulate_transaction(
...
-
-            let simulation_result = executor
-                .simulate_transaction(
-                    estimation_transaction,
-                    TransactionChecks::Enabled,
-                    true, /* allow mock gas coin */
-                )
-                .map_err(anyhow::Error::from)?;
+    let perform_gas_selection = request.do_gas_selection() && checks.enabled();
+    let simulation_result = 'simulate: {
```
bmwill, line 102: maybe make this a function then? i tend to have a very difficult time reading these scope breaking setups.

aschran: It seemed cleaner this way. the function would have so many arguments and imo be harder to read bc it would require jumping back and forth to piece the logic together

mystenmark (approved):

PR #26240, opened 2026-04-14
[indexer-alt-reader] make FullnodeClient require execution client

wlmyng:
## Description 

Callers take on the responsibility of instantiating the fullnode client, rather than the client itself managing whether an execution service is configured or not, to simplify callsites using the client and reduce checks to initialization 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/mutation.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/mutation.rs
@@ -63,8 +63,7 @@ impl Mutation {
     ) -> Result<ExecutionResult, RpcError<TransactionInputError>> {
         // Get the gRPC client from context. If the service was started without a fullnode URL
         // there is no client to dispatch the transaction to.
-        let fullnode_client: &Option<FullnodeClient> = ctx.data()?;
-        let Some(fullnode_client) = fullnode_client.as_ref() else {
+        let Some(fullnode_client) = ctx.data_opt::<FullnodeClient>() else {
             return Err(feature_unavailable("executing transactions"));
         };
 
--- a/crates/sui-indexer-alt-graphql/src/api/mutation.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/mutation.rs
@@ -21,6 +21,7 @@ use crate::api::types::execution_result::ExecutionResult;
 use crate::api::types::transaction_effects::TransactionEffects;
 use crate::error::RpcError;
 use crate::error::bad_user_input;
+use crate::error::feature_unavailable;
 use crate::error::upcast;
 use crate::scope::Scope;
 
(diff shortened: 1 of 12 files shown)


crates/sui-indexer-alt-graphql/src/api/mutation.rs
```diff
@@ -60,8 +61,12 @@ impl Mutation {
...
         signatures: Vec<Base64>,
     ) -> Result<ExecutionResult, RpcError<TransactionInputError>> {
-        // Get the gRPC client from context
-        let fullnode_client: &FullnodeClient = ctx.data()?;
+        // Get the gRPC client from context. If the service was started without a fullnode URL
+        // there is no client to dispatch the transaction to.
+        let fullnode_client: &Option<FullnodeClient> = ctx.data()?;
+        let Some(fullnode_client) = fullnode_client.as_ref() else {
+            return Err(feature_unavailable("executing transactions"));
+        };
```
amnn, line 68: Could we use the existing pattern for optional context entries, by not adding the client to the context if it's not been configured?

```suggestion
        // Get the gRPC client from context. If the service was started without a fullnode URL
        // there is no client to dispatch the transaction to.
        let Some(fullnode_client): Option<&FullnodeClient> = ctx.data_opt() 
        else {
            return Err(feature_unavailable("executing transactions"));
        };
```

crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -315,9 +315,17 @@ pub async fn start_rpc(
...
     let metrics = rpc.metrics();
 
-    // Create gRPC full node client wrapper
-    let fullnode_client =
-        FullnodeClient::new(Some("graphql_fullnode"), fullnode_args, registry).await?;
+    // Create gRPC full node client wrapper. If no URL was configured, the client is stored as
+    // `None` in the schema data and resolvers that depend on it return `FeatureUnavailable`.
+    let fullnode_client = if let Some(url) = fullnode_args.fullnode_rpc_url {
+        Some(
+            FullnodeClient::new(Some("graphql_fullnode"), url, registry)
```
amnn: I wonder if you could still preserve the way the client is constructed, but just adapt its signature so that `FullnodeClient::new` returned an `Result<Option<Self>, _>`? That way you would not need to reach into the arguments structure outside of the module that defined it.

amnn (approved):
Nice simplification -- suggested a couple more, thanks @wlmyng

crates/sui-indexer-alt-graphql/src/api/mutation.rs
wlmyng: ooh, thanks for the reminder on `data_opt`, yes that should work

crates/sui-indexer-alt-graphql/src/lib.rs
wlmyng: In fact, I think the better setup is an `Option<FullnodeArgs>`, with non-optional `fullnode_rpc_url`, and this is passed into `FullnodeClient`, which continues to return `Result<Self, Error>` because it would only be called if fullnode url is actually configured. Missed this spot the first time around

wlmyng: I see FullnodeArgs is used as a cli clap arg, but this should still be fine because the `Url` field is required

amnn: Are you sure that works? I can't remember the full details, but I remember there were problems related to flattening an `Option<Arg>` type.

wlmyng: Ah yes... wow, that's annoying. I've put up a fix here: https://github.com/MystenLabs/sui/pull/26257

PR #26242, opened 2026-04-14
[move][compressed-layouts] Arc instead of Box LayoutRefs

tzakian:
## Description 

`Arc` them instead of `Box`ing to make clones bit cheaper.

## Test plan 

CI/existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


tnowacki (commented):
When are these items cloned?

tzakian:
Hmmm good question. I think we probably don't need to do this. I think I'm going to close this out, and then we can look at adding sharing back in more judiciously once this is plumbed out.

PR #26245, opened 2026-04-14
[forking 6/n] add support for checkpoint contents

stefan-mysten:
## Description 

This PR implements the following in preparation for the next PR where we'll be adding gRPC support for basic operations with checkpoints, objects, and transactions:
- support for checkpoint contents
  - additional graphql query support for fetching checkpoint content BCS
  - filesystem support to read and write checkpoints to disk
- moves various unit tests into their own folder for better separation

## Test plan 

```
cd crates/sui-forking && cargo nextest run
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-forking/src/store.rs
```diff
@@ -86,16 +86,101 @@ impl DataStore {
...
+    /// fetched remotely — they only exist if the local executor produced
+    /// them, so a miss returns `None`.
+    pub(crate) fn get_checkpoint_by_sequence_number(
         &self,
-        checkpoint: Option<CheckpointSequenceNumber>,
+        sequence: CheckpointSequenceNumber,
     ) -> anyhow::Result<Option<VerifiedCheckpoint>> {
-        let checkpoint = checkpoint.unwrap_or(self.forked_at_checkpoint);
-        let verified_checkpoint = self.gql.get_verified_checkpoint(Some(checkpoint))?;
+        if let Some(checkpoint) = self.local.get_checkpoint_by_sequence_number(sequence)? {
```
stefan-mysten, line 99: Should start thinking of splitting this DataStore into historical lookups, and local forked filesystem.

jordanjennings-mysten (approved):

jordanjennings-mysten:
I wouldn't do this today maybe leave for a later refactor:

If you pull out the part about how you are encoding an object, checkpoint etc into a key, value (in the case of FS filename as key + serialized type as the value stored to file) you can make filesystem more generic and flexible (which also helps with new types you need to serialize to disk) and also make database backing easier later on.

ObjectFSEncode -> key value of file path and serialized type
Filesystem -> writes a generic key value to the correct folder based on the key

later on

DatabaseEncode -> possibly the same key and value encoding but might also have its own format, depending on if its postgres or just a key value store
Database -> Writes a database encoded K/V or something more complex.

stefan-mysten:
> I wouldn't do this today maybe leave for a later refactor:
> 
> If you pull out the part about how you are encoding an object, checkpoint etc into a key, value (in the case of FS filename as key + serialized type as the value stored to file) you can make filesystem more generic and flexible (which also helps with new types you need to serialize to disk) and also make database backing easier later on.
> 
> ObjectFSEncode -> key value of file path and serialized type Filesystem -> writes a generic key value to the correct folder based on the key
> 
> later on
> 
> DatabaseEncode -> possibly the same key and value encoding but might also have its own format, depending on if its postgres or just a key value store Database -> Writes a database encoded K/V or something more complex.

This is interesting. Let's discuss as part of the `sui-data-store` work, as the plan is for this to share the common lookups through that, and take advantage of the caching layer there.

PR #26246, opened 2026-04-14
[WIP] Party permissioned objects

tnowacki:
## Description 

Foundational types and infrastructure for permissioned consensus-owned ("party") objects. Adds `Owner::Party`, `ObjectPermissions`, `Party`, and the loading + typing pipeline pieces that thread permission state through PT execution. Construction paths and most read-side conversions are explicitly stubbed (`todo!("Party WIP")`) — see TODO. Defenses in place to reject any `Owner::Party` input or effect until the rest lands.

### Open Questions

- Currently we are relying on the `SharedObjectMutability::Immutable` and `SharedObjectMutability::Mutable`. If instead we had something like `SharedObjectMutability::Party` we could simplify the SDK/transaction builder logic since it would not require refactoring it to pass in the sender. We could still support `Mutable` and `Immutable` for party objects (as is done on this PR), but an extra mode would help simplify things on the client side a bit.

### Done

- Added `Owner::Party { start_version, permissions }` variant
- `ObjectPermission` is a small bitflag enum with the permissions from the `sui::party` Move module
- `ObjectPermissions` is a bitflag set of permissions
- `Party` struct that mirrors `sui::party::Party` Move struct
  - Cannot be used to represent a consensus address owner
  - Sorts the members and requires uniqueness
- Custom serde validation:
  - `ObjectPermissions` via `try_from = "u64"` — rejects unknown bits and mutable permissions without `MutableUsage`
  - `Party` via `try_from = "RawPartySerde"` — rejects unsorted members, duplicate addresses, and non-canonical forms
- Signing checks for immutable/mutable access in `sui-transaction-checks` implemented (outside of `NonExclusiveWrite` mutability, which remains TODO)
- `is_consensus()` now includes `Party`
- Loading layer: `Owner::Party` and `Owner::ConsensusAddressOwner` route through a unified `SharedObjectKind::Party(ObjectPermissions)` shape in `loading/ast.rs` + `loading/translate.rs`, capturing sender-resolved permissions for downstream use
- Typing pipeline: `Command_::incurs_post_execution_checks` (renamed from `consumed_shared_objects: Vec<ObjectID>`) computed per command, drives the `private_entry_arguments` hot-clique analysis
- Defense-in-depth:
  - Input check in `sui-transaction-checks::check_one_object` early-bails on `Owner::Party` with `UserInputError::Unsupported`
  - Effect-writer assertion in `temporary_store::into_effects` rejects any written object carrying `Owner::Party`

### TODO

- No creation path yet in Move
- No execution implementations (~12 `todo!("Party WIP")` sites remain)
- `test_scenario.rs` (latest) — test scenario support
- Deeper RPC/SDK conversions
- Balance change tracking for party-owned coins
- Indexer support
- Transaction builder / CLI PTB builder support

## Test plan 

- Ran tests
- Replay validation 

---

## Release notes

- [X] Protocol: New flag for dealing with party object extensions for incurring post execution writes
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-types/src/object.rs
```diff
@@ -493,6 +493,255 @@ impl Data {
...
+    /// Returns `None` if it could be represented as a single `ConsensusAddressOwner`
+    pub fn new(
+        default_permissions: ObjectPermissions,
+        mut members: Vec<(SuiAddress, ObjectPermissions)>,
+    ) -> Option<Self> {
+        if Self::is_consensus_address_owner(default_permissions, &members) {
+            return None;
+        }
+
+        if !members.is_sorted_by_key(|(a, _)| *a) {
```
akmysti: Assuming line 677 explicitly requires members to be sorted, what is the rationale to sort it here?

crates/sui-types/src/object.rs
```diff
@@ -493,6 +493,255 @@ impl Data {
...
+            }
+            write!(f, "{}", permission)?;
+            first = false;
+        }
+        write!(f, "}}")
+    }
+}
+
+#[derive(
+    Eq, PartialEq, Debug, Clone, Deserialize, Serialize, Hash, JsonSchema, Ord, PartialOrd,
```
akmysti, line 675: How do we validate that a Deserialized PartyPermissions object has members sorted by  SuiAddress, and it doesn't have any duplicate members ?

crates/sui-types/src/object.rs
```diff
@@ -493,6 +493,255 @@ impl Data {
...
+
+    pub const fn intersect(self, other: Self) -> Self {
+        Self(self.0 & other.0)
+    }
+
+    pub const fn union(self, other: Self) -> Self {
+        Self(self.0 | other.0)
+    }
+
+    pub const fn difference(self, other: Self) -> Self {
```
akmysti, line 590: Assuming this will be used/tested somewhere ?

crates/sui-types/src/object.rs
```diff
@@ -519,6 +768,10 @@ pub enum Owner {
         // The owner of the object.
         owner: SuiAddress,
     },
+    PartyPermissioned {
```
aschran: personally I would vote for just `Party` here. every owner type is "permissioned" so the suffix feels a bit redundant

crates/sui-types/src/object.rs
```diff
@@ -519,6 +768,10 @@ pub enum Owner {
...
+        start_version: SequenceNumber,
+        permissions: PartyPermissions,
```
aschran: should this be boxed to avoid increasing Owner size too much?

crates/sui-types/src/object.rs
```diff
@@ -553,7 +809,8 @@ impl Owner {
             Self::Shared {
                 initial_shared_version,
             } => Some(*initial_shared_version),
-            Self::ConsensusAddressOwner { start_version, .. } => Some(*start_version),
+            Self::ConsensusAddressOwner { start_version, .. }
```
aschran, line 810: nit - update function comment

crates/sui-types/src/object.rs
```diff
@@ -493,6 +493,255 @@ impl Data {
...
+    pub fn new(
+        default_permissions: ObjectPermissions,
+        mut members: Vec<(SuiAddress, ObjectPermissions)>,
+    ) -> Option<Self> {
+        if Self::is_consensus_address_owner(default_permissions, &members) {
+            return None;
+        }
+
+        if !members.is_sorted_by_key(|(a, _)| *a) {
+            members.sort_by_key(|(a, _)| *a);
```
aschran: The is_sorted_by_key guard before sort_by_key is redundant — Rust's sort is already O(n) on sorted input.

crates/sui-types/src/object.rs
```diff
@@ -493,6 +493,255 @@ impl Data {
...
+        }
+        for window in members.windows(2) {
+            let [(addr1, _), (addr2, _)] = window.try_into().unwrap();
+            if addr1 == addr2 {
+                // Not unique
+                return None;
+            }
+        }
```
aschran: nit-potentially more simple as

```
if members.windows(2).any(|w| w[0].0 == w[1].0) {
    return None;
}
```

crates/sui/src/client_ptb/builder.rs
```diff
@@ -150,6 +150,10 @@ impl<'a> Resolver<'a> for ToObject {
                     Tx::SharedObjectMutability::Immutable
                 },
             },
+            Owner::PartyPermissioned { .. } => {
+                // We need to know the sender to determine mutability
```
aschran: why does the `is_mut` thing used above for shared obj not work here?

aschran: (this appears in a few other places so I will not repeat the same question, but consider this one to apply to all locations)

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -671,6 +676,47 @@ mod checked {
...
+                            input_initial_shared_version == *actual_initial_shared_version,
+                            UserInputError::SharedObjectStartingVersionMismatch
+                        );
+                        // Check the owner has permissions for this kind of mutability
+                        let sender_permissions = permissions.permissions_for(owner);
+                        match mutability {
+                            SharedObjectMutability::Immutable => {
+                                // TODO better error kind here
+                                fp_ensure!(
+                                    sender_permissions.can_access_immutably_at_signing(),
```
aschran, line 694: it seems that if you don't have Read but do have other permissions, you still should be able to access immutably at signing? (ie "write" is a superset of "read"?) wdyt?

aschran (commented):
looks nice overall!

crates/sui-transaction-checks/src/lib.rs
tzakian: I don't have super strong opinions here, but I think if we wanted to have some level of parallelism with unix/linux file permissions types, then needing an explicit read would logically follow.

crates/sui-types/src/object.rs
```diff
@@ -493,6 +493,255 @@ impl Data {
...
+#[derive(
+    Eq, PartialEq, Debug, Clone, Deserialize, Serialize, Hash, JsonSchema, Ord, PartialOrd,
+)]
+#[cfg_attr(feature = "fuzzing", derive(proptest_derive::Arbitrary))]
+pub struct PartyPermissions {
+    /// The default permissions, used for any party member not explicitly listed in `members`.
+    default_permissions: ObjectPermissions,
+    /// The permissions for each party member.
+    /// NOTE this should be sorted by SuiAddress
+    members: Vec<(SuiAddress, ObjectPermissions)>,
```
tzakian: Getting to Aditya's comment above, thoughts on a BTreeMap here? Or do we not require that permissions be reduced? It would also simplify the logic in `new` as well I think.

tzakian: Although I have a sneaking suspicion that this is hard/impossible to do due to this actually being a shared thing with a Move datastructure...but worth asking..

tnowacki: Yeah I started down this road with that thought. But I actually don't think they need to be tied together here in this way. We already don't directly expose owner in Move anywhere this would be a problem... so let me try the map which should remove a lot of these small issues

crates/sui-types/src/object.rs
tnowacki: Addresses are real big. It's actually the same size as `ConsensusAddressOwner` at 40 bytes.
Though I can still box it if you want

crates/sui/src/client_ptb/builder.rs
tnowacki: Ah I suppose we could... it just feels a bit strange since you might have permission to do that. But maybe that isn't the SDKs problem?

crates/sui-transaction-checks/src/lib.rs
tnowacki: I think the semantics here are fine but maybe just the name is off? 
We want a Permission that correspond to the following Sui behaviors. I will give them letters since I will need to refer to them, but don't get caught up in the names/letters
- (I) Can take the object as an immutable input
- Can take the object as a mutable input and:
  - (M) maybe modify the value (but not its location/existence in storage)
  - (T) transfer it (value must remain the same if it we the sender does not have (M)) 
    - wrapping is weird here, see below
  - (D) delete it 
  
And we can separate wrapping out from (T) if we want but it is tricky since we will not be able to determine if the value has been modified so we would have to define it as
- (W) wrap it and maybe modify the value. But if it is not wrapped, that modification is not allowed 😵‍💫

crates/sui/src/client_ptb/builder.rs
tnowacki: I did update this and other spots with utilizing that flag + a todo

crates/sui-types/src/object.rs
```diff
@@ -493,6 +494,307 @@ impl Data {
...
+            Self::Wrap => write!(f, "Wrap"),
+            Self::MutableUsage => write!(f, "MutableUsage"),
+            Self::ImmutableUsage => write!(f, "ImmutableUsage"),
+        }
+    }
+}
+
+/// A bitflag set of `ObjectPermission`s.
+#[derive(
+    Eq, PartialEq, Debug, Clone, Copy, Deserialize, Serialize, Hash, JsonSchema, Ord, PartialOrd,
```
tzakian, line 560: Should we be defensive on the `Deserialize` here to make sure we only allow/deserialize valid `ObjectPermissions`? e.g.,

slap a `#[serde(try_from = "u64", into = "u64")]` on this and do  something like 

```
 impl TryFrom<u64> for ObjectPermissions {
      type Error = &'static str;
      fn try_from(bits: u64) -> Result<Self, Self::Error> {
          Self::new(bits).ok_or("invalid ObjectPermissions bits")
      }
  }
```

tzakian (approved):
One comment, but otherwise LGTM

crates/sui-types/src/object.rs
```diff
@@ -493,6 +494,363 @@ impl Data {
...
+    MutableUsage = 1 << 1,
+    Write = 1 << 2,
+    Delete = 1 << 3,
+    InternalTransfer = 1 << 4,
+    PublicTransfer = 1 << 5,
+    Wrap = 1 << 6,
+}
+
+impl ObjectPermission {
+    pub const fn from_u64(bits: u64) -> Option<Self> {
```
aschran: is it worth logging any kind of error/debug_fatal if more than one bit is set in the argument?

crates/sui-types/src/object.rs
```diff
@@ -493,6 +494,363 @@ impl Data {
...
+        (self.0 & p) == p
+    }
+
+    /// Can it change the Move value of the object?
+    pub fn can_write(&self) -> bool {
+        self.can(ObjectPermission::Write)
+    }
+
+    /// Can the object's UID be deleted?
+    // TODO this might also gate derived object restoration
```
aschran: not sure I fully undersatnd this TODO - what is left to-do here? or is it more of a "note to self"?

crates/sui-types/src/object.rs
```diff
@@ -493,6 +494,363 @@ impl Data {
...
+pub struct ObjectPermissionsIterator {
+    set: ObjectPermissions,
+    idx: usize,
+}
+
+impl Iterator for ObjectPermissionsIterator {
+    type Item = ObjectPermission;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        const ORDER: &[ObjectPermission] = &[
```
aschran: curious how you chose this order and why it differs from enum definition order?

aschran: also can you do anything here so that if a permission is added in the future, the build fails here & we can't forget to fix it?

crates/sui-types/src/object.rs
```diff
@@ -519,6 +877,17 @@ pub enum Owner {
         // The owner of the object.
         owner: SuiAddress,
     },
+    /// Object is sequenced via consensus with per-address permissions.
+    /// Each address can be granted a subset of {Read, Write, Delete, Transfer}.
```
aschran: this line looks outdated?

crates/sui-types/src/object.rs
```diff
@@ -1398,4 +1778,21 @@ mod tests {
         test_for_value(u32::MAX as u64 + 1);
         test_for_value(u64::MAX);
     }
+
+    #[test]
+    fn test_owner_variant_sizes() {
+        use super::{Party, SequenceNumber};
+
+        // AddressOwner / ObjectOwner
+        assert_eq!(std::mem::size_of::<SuiAddress>(), 32);
```
aschran, line 1861: what's the point of asserting specific individual variant sizes? could the test just be one assert that sizeof Owner is less than some limit e.g. 64 or 128 or whatever we want?

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -671,6 +683,47 @@ mod checked {
...
+                                    UserInputError::IncorrectUserSignature {
+                                        error: format!(
+                                            "Sender address {owner:?} does not have mutable access permissions for object {object_id:?} with party ownership. The required permission is {}, but the permissions for the sender for this object are {sender_permissions}",
+                                            ObjectPermission::MutableUsage,
+                                        ),
+                                    }
+                                )
+                            }
+                            SharedObjectMutability::NonExclusiveWrite => {
+                                todo!("Party WIP")
```
aschran, line 724: confirming this is ok because it's prevented from reaching this point elsewhere?

sui-execution/v3/sui-adapter/src/programmable_transactions/context.rs
```diff
@@ -1818,6 +1818,7 @@ mod checked {
                 // protected by transaction input checker
                 invariant_violation!("ObjectOwner objects cannot be input")
             }
+            Owner::Party { .. } => todo!("Party WIP"),
```
aschran: should this be "Does not exist"?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
```diff
@@ -50,30 +51,25 @@ pub enum InputArg {
...
-pub enum ObjectMutability {
-    Mutable,
-    Immutable,
-    NonExclusiveWrite,
+#[derive(Debug)]
+#[cfg_attr(debug_assertions, derive(Clone))]
+pub struct ObjectArg {
+    pub kind: ObjectArgKind,
+    /// Permissions, potentially refined/limited based on the input argument. For example if a
+    /// shared object is used but marked as read-only, the permissiones would be refined to being
```
aschran: nit: spelling

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/translate.rs
```diff
@@ -1333,105 +1326,106 @@ mod consumed_shared_objects {
     /// deleted or re-shared.
```
aschran: does this need any udpating? it's maybe technically still true but a bit confusing now that the content of the function references `incurs_post_execution_checks` and not `consumed_shared_objects`?

sui-execution/latest/sui-move-natives/src/transfer.rs
```diff
@@ -248,6 +248,7 @@ pub fn party_transfer_internal(
         owner: address.into(),
     };
     object_runtime_transfer(context, owner, ty, obj)?;
+    // TODO check permissions for transfer
```
aschran, line 251: maybe add the string "Party WIP" on these just so we can find them more easily later?

aschran (approved):

crates/sui-transaction-checks/src/lib.rs
mystenmark: User transactions cannot specify this mode at all (yet) so it should fail in validity check

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -533,6 +536,13 @@ mod checked {
         object: &Object,
         system_transaction: bool,
     ) -> UserInputResult {
+        // Defense-in-depth: Owner::Party is not yet supported.
+        if matches!(object.owner, Owner::Party { .. }) {
+            return Err(UserInputError::Unsupported(
+                "Party-owned objects are not yet supported".to_string(),
+            ));
```
mystenmark, line 543: a bit confused here - we're shipping the code but not enabling it? how are we testing anything with this here?

tnowacki: While this PR is old, so the intent has been lost a bit to time, but yes we are shipping this without actually building `Party` objects anywhere. In this way, all of this code is malleable. @aschran and I discussed this and landed on doing this rather than doing a branch or anything, because of how it can be contained. 

Maybe not entirely clear from the comment above this part, but this code cannot be reached. It was only added as additional belt-and-suspenders kind of check.

aschran: Yes, fwiw this is similar to how I added ConsensusAddressOwner a while back

crates/sui-transaction-checks/src/lib.rs
tnowacki: We don't construct any party objects, so this code is unreachable

crates/sui-types/src/object.rs
tnowacki: It was mostly done for `Party` to make sure that one doesn't drift even if it gets Boxed or something.. 
I do test the size of Owner below to be `48`

crates/sui-types/src/object.rs
tnowacki: Good callout. This code went through a few refactorings on the bit layout, and I forgot to change this one.

crates/sui-types/src/object.rs
tnowacki: More of a note to self, but I will update it a tad

crates/sui-types/src/object.rs
tnowacki: I am a bit limited since this needs to be `const`, but I will try to add `debug_assert`

PR #26248, opened 2026-04-15
[sui-framework] Add df/dof `replace`. Fix `remove_opt`

tnowacki:
## Description 

- Adds `sui::dynamic_field::replace`
- Adds `sui::dynamic_object_field::replace`
- Fixes `dynamic_object_field::remove_opt` 
- Additional tests for macros on type mismatch

## Test plan 

- New tests

---

## Release notes

- [x] Protocol:

- Adds `sui::dynamic_field::replace`.
- Adds `sui::dynamic_object_field::replace`.

- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-framework/packages/sui-framework/tests/dynamic_field_tests.move
```diff
@@ -155,6 +155,44 @@ fun delete_uid_with_fields() {
...
+    let sender = @0x0;
+    let mut scenario = test_scenario::begin(sender);
+    let mut id = scenario.new_object();
+    add(&mut id, 0u64, 42u64);
+    let old = dynamic_field::replace<u64, u64, u64>(&mut id, 0, 100);
+    assert_eq!(old, option::some(42));
+    assert_eq!(*borrow<u64, u64>(&id, 0), 100);
+    scenario.end();
+    id.delete();
+}
```
akmysti, line 204: Missing test when the type is misrepresented:

```rust
    let mut id = scenario.new_object();
    add(&mut id, 0u64, 42u64);

    // Field is u64. We mistakenly try to extract it as a u8. Should abort.
    let _old = dynamic_field::replace<u64, u64, u8>(&mut id, 0u64, 100u64);
```

akmysti:
We probably want to test add/remove objects on same name. Something like:

Add(Name, Value1) 
Replace(Name, Value1, Value2)
Replace(Name, Value2, Value3)
Replace(Name, Value3, Value1)

tnowacki:
> We probably want to test add/remove objects on same name. Something like:

We have replace_existing and replace_different_type tests which do that with one step. What is the case you think not covered there?

akmysti:
> > We probably want to test add/remove objects on same name. Something like:
> 
> We have replace_existing and replace_different_type tests which do that with one step. What is the case you think not covered there?

I was thinking of a 'Roundrip test' that replaces a few times and goes back to the original value. feel free to ignore if you dont find it adding any value.

tnowacki:
> > > We probably want to test add/remove objects on same name. Something like:
> > 
> > 
> > We have replace_existing and replace_different_type tests which do that with one step. What is the case you think not covered there?
> 
> I was thinking of a 'Roundrip test' that replaces a few times and goes back to the original value. feel free to ignore if you dont find it adding any value.

It is a good callout, and I think it would be worth adding if `replace` was more than a wrapper around add/remove. And we have some good coverage for add/remove loops in the transactional tests

akmysti (approved):

damirka (approved):

PR #26250, opened 2026-04-15
[sui-rosetta] Parse ConsolidateAllStakedSuiToFungible on read path

xia88niuniu:
## Summary

Adds typed parsing for `ConsolidateAllStakedSuiToFungible` transactions on the Rosetta read path. Previously `/construction/parse` and `/block/transaction` returned a single opaque `ProgrammableTransaction` generic op for these PTBs, leaving clients to decode the embedded PTB themselves. They now emit a typed `ConsolidateAllStakedSuiToFungible` op with `staked_sui_ids` and `fss_ids` extracted from PTB inputs.

## Changes

**Parser (`crates/sui-rosetta/src/operations.rs`):**

- New predicates `is_convert_to_fss_call` and `is_join_fss_call` recognize the signature MoveCalls `0x3::sui_system::convert_to_fungible_staked_sui` and `0x3::staking_pool::join_fungible_staked_sui`.
- Dispatch scans commands once at the top of `parse_programmable_transaction`; if either signature call is present, runs a new `parse_consolidate` sub-parser. Shape mismatch falls through to the existing generic op fallback, so third-party PTBs that happen to call the same Move functions are not mislabeled.
- `parse_consolidate` performs strict shape validation covering all three valid PTB shapes produced by the existing `consolidate_to_fungible_pt` builder:
  - pure FSS merge (`S=0, F>=2`): only `join_fungible_staked_sui` calls
  - convert-only (`S>=1, F=0`): `convert` calls + trailing `TransferObjects` to sender
  - mixed (`S>=1, F>=1`): existing FSS merges + converts + new-FSS merges + cross-merge
- Input classification walks `convert` and `join` argument refs to split PTB inputs into StakedSui vs FSS object IDs; any overlap rejects (falls through).

**Metadata — backward-compatible:**

- `OperationMetadata::ConsolidateAllStakedSuiToFungible.validator` relaxed to `Option<SuiAddress>`. Parsed ops emit `validator: None` because the Move function takes only the object ref, not the validator address — it is not recoverable from PTB bytes in the sync parse path. Write-side destructurer still returns `Error::MissingInput` if absent, preserving old behavior for write callers. Old JSON input with `{\"validator\": \"0x...\"}` continues to deserialize.
- Two new parse-populated fields: `staked_sui_ids: Vec<ObjectID>` and `fss_ids: Vec<ObjectID>`, defaulting to empty on write input (ignored by write destructurer via `..`).

**Test-visibility change (`crates/sui-rosetta/src/types/internal_operation.rs`):**

- Re-export `consolidate_to_fungible_pt` as `pub(crate)` so the unit tests in `operations.rs` can construct PTBs through the same builder used by production code, guaranteeing test-to-production shape parity.

## Behavior

`/construction/parse` response for a Consolidate PTB (S=2, F=2):
```json
{
  "operations": [{
    "type": "ConsolidateAllStakedSuiToFungible",
    "account": { "address": "<sender>" },
    "metadata": {
      "ConsolidateAllStakedSuiToFungible": {
        "staked_sui_ids": ["0x...", "0x..."],
        "fss_ids": ["0x...", "0x..."]
      }
    }
  }],
  "account_identifier_signers": [...]
}
```

`/block/transaction` layers the typed op with the usual `Gas` op and balance-change ops from effects processing (no `SuiBalanceChange` for the sender since Consolidate has no net SUI delta).

## Test plan

- [x] 22 new unit tests in `operations.rs::tests` covering:
  - 11 happy paths: every (S, F) combination accepted by the builder, including pure merge (`S=0, F={2,3,5}`), convert-only (`S={1,3}, F=0`), and mixed (`S={1,2,3}, F={1,2,3}`), plus a classification-correctness test.
  - 4 fall-through cases: rogue `MergeCoins`, unrelated MoveCall, wrong first input, extra command after `TransferObjects` — each asserts the parser falls through to the generic `ProgrammableTransaction` op.
  - 2 robustness: empty PTB, PTB with only `MergeCoins` (non-FSS).
  - 2 serialization compat: old JSON input still deserializes; new parse output serializes with proper field-skip.
  - 1 write-side preservation: missing validator on write input still yields `MissingInput`.
- [x] 19 new e2e tests in `end_to_end_tests.rs`:
  - 7 `/construction/parse` round-trip tests across the (S, F) matrix plus multi-validator isolation.
  - 7 `/block/transaction` tests submitting the TX and re-parsing through `try_from_executed_transaction`.
  - 2 new write-error tests (invalid validator, missing validator); 3 others (no-stakes, single-FSS, unactivated-only) are already covered by existing `test_consolidate_noop_*` tests.
  - 2 robustness: garbage bytes and truncated TX data to `/construction/parse` return proper Rosetta error responses.
- [x] Full `cargo nextest run -p sui-rosetta` suite: 136 tests pass, 0 skipped.
- [x] `cargo clippy -p sui-rosetta --tests` clean.

### DIFF
--- a/crates/sui-rosetta/src/operations.rs
+++ b/crates/sui-rosetta/src/operations.rs
@@ -30,7 +30,9 @@ use sui_types::base_types::{ObjectID, SequenceNumber, SuiAddress};
 use sui_types::gas_coin::GasCoin;
 use sui_types::governance::{ADD_STAKE_FUN_NAME, WITHDRAW_STAKE_FUN_NAME};
 use sui_types::sui_system_state::SUI_SYSTEM_MODULE_NAME;
-use sui_types::{SUI_FRAMEWORK_PACKAGE_ID, SUI_SYSTEM_ADDRESS, SUI_SYSTEM_PACKAGE_ID};
+use sui_types::{
+    SUI_FRAMEWORK_PACKAGE_ID, SUI_SYSTEM_ADDRESS, SUI_SYSTEM_PACKAGE_ID, SUI_SYSTEM_STATE_OBJECT_ID,
+};
 
 use crate::types::internal_operation::{
     ConsolidateAllStakedSuiToFungible, MergeAndRedeemFungibleStakedSui, PayCoin, PaySui, Stake,
@@ -277,11 +279,15 @@ impl Operations {
         let metadata = op.metadata.ok_or_else(|| {
             Error::MissingInput("ConsolidateAllStakedSuiToFungible metadata".to_string())
         })?;
-        let OperationMetadata::ConsolidateAllStakedSuiToFungible { validator } = metadata else {
+        let OperationMetadata::ConsolidateAllStakedSuiToFungible { validator, .. } = metadata
+        else {
             return Err(Error::InvalidInput(
                 "Cannot find validator from ConsolidateAllStakedSuiToFungible metadata.".into(),
             ));
         };
+        let validator = validator.ok_or_else(|| {
+            Error::MissingInput("validator required for ConsolidateAllStakedSuiToFungible".into())
+        })?;
         Ok(InternalOperation::ConsolidateAllStakedSuiToFungible(
             ConsolidateAllStakedSuiToFungible { sender, validator },
         ))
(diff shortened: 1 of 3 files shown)


xia88niuniu:
Closing — superseded by a consolidated PR that ships both FungibleStakedSui parsers together.

PR #26258, opened 2026-04-15
Fix watermark init bug

nickvikeras:
## Description

In (#26041) I migrated the analytics indexer to use the sui-rpc-resolver rather than it's older custom package cache implementation. I used a "pseudo pipeline" to invalidate the cache on epoch boundaries. This pipeline always returned a watermark of 0 at startup, which caused the indexer to stall when it was deployed.

The fix is to just choose the max of the other pipelines as this pipelines starting watermark.

## Test plan

Manualy deploying to testnet now to verify the fix.

### DIFF
--- a/crates/sui-analytics-indexer/src/store/mod.rs
+++ b/crates/sui-analytics-indexer/src/store/mod.rs
@@ -44,6 +44,7 @@ use tracing::warn;
 use crate::config::FileFormat;
 use crate::config::IndexerConfig;
 use crate::handlers::CheckpointRows;
+use crate::handlers::system_package_eviction::SYSTEM_PACKAGE_EVICTION_PIPELINE;
 use crate::metrics::Metrics;
 use crate::schema::RowSchema;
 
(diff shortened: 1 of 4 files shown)


bmwill (approved):

crates/sui-analytics-indexer/src/store/mod.rs
```diff
@@ -664,18 +672,48 @@ impl Connection for AnalyticsConnection<'_> {
...
     /// In migration mode: reads from watermark metadata file.
+    ///
+    /// `SystemPackageEviction` is a pseudohandler that never writes files, so its
+    /// watermark cannot be derived from its own `output_prefix`. It is always registered
+    /// last in `build_analytics_indexer`, so by the time this fires for it, each other
+    /// pipeline's watermark has already been computed and cached in `initial_watermarks`.
+    /// Return the max across those pipelines. The cache only matters at tip — it starts
+    /// empty on restart and only needs invalidation to clear entries staled by a live
+    /// epoch transition. Starting at the farthest-along pipeline means eviction sits idle
+    /// while slower pipelines backfill, then wakes up at tip where invalidation matters.
```
wlmyng, line 683: can't we keep track of a running max instead? it seems we only do this once at an indexer's startup, once per pipeline, the eviction doesn't seem to touch this `initial_watermarks` map

PR #26260, opened 2026-04-15
[cleanup] remove benchmark client type

mwtian:
## Description 

This is unused now.

## Test plan 

CI

### DIFF
--- a/crates/sui-benchmark/src/drivers/bench_driver.rs
+++ b/crates/sui-benchmark/src/drivers/bench_driver.rs
@@ -220,23 +220,6 @@ struct Stats {
     pub bench_stats: BenchmarkStats,
 }
 
-#[derive(Debug)]
-pub enum ClientType {
-    // Used for Mysticeti Fast Path
-    TransactionDriver,
-    // Used for original tx certification then fast path + Mysticeti
-    QuorumDriver,
-}
-
-impl std::fmt::Display for ClientType {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            ClientType::TransactionDriver => write!(f, "transaction_driver"),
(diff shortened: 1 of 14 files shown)


crates/sui-benchmark/src/drivers/bench_driver.rs
```diff
@@ -842,11 +825,11 @@ async fn run_bench_worker(
                 let square_latency_ms = latency.as_secs_f64().powf(2.0);
                 metrics
                     .latency_s
-                    .with_label_values(&[&payload.to_string(), &client_type.to_string()])
+                    .with_label_values(&[payload_str.as_str(), "transaction_driver"])
```
akichidis: Since we have only transaction driver, can't we not populate the `transaction_driver` value?

crates/sui-benchmark/src/drivers/bench_driver.rs
```diff
@@ -1039,9 +1022,10 @@ async fn run_bench_worker(
                     let num_in_flight_metric = metrics.num_in_flight.with_label_values(&[&payload.to_string()]);
                     let res = worker.execution_proxy
                         .execute_transaction_block(tx.clone())
-                        .then(|(client_type, res)| async move  {
-                            metrics.num_submitted.with_label_values(&[&payload.to_string(), &client_type.to_string()]).inc();
-                            handle_execute_transaction_response(res, start, tx, payload, committee, client_type)
+                        .then(|res| async move  {
+                            let payload_str = payload.to_string();
+                            metrics.num_submitted.with_label_values(&[payload_str.as_str(), "transaction_driver"]).inc();
```
akichidis: Same as above, do we need any more the client type in metrics?

akichidis (approved):
Left a comment regarding the necessity of passing on the metrics now the client type.

mwtian:
Removing labels can break metrics. I want to keep it for now until we know for sure it won't be needed.

PR #26264, opened 2026-04-16
[gasless] Add remaining balance verification for withdrawals

alex-mysten:
## Summary

- Adds `gasless_verify_remaining_balance` protocol config flag (enabled in v122)
- At voting time, validates that a gasless withdrawal does not leave the sender's accumulator balance in an invalid state (remaining must be 0 or >= min_transfer)
- At execution time, validates the same invariant against actual accumulator net balance changes, covering partial redemptions and `withdrawal_split` flows
- Transactional tests covering boundary conditions for both validation paths

## Test plan

- [x] `cargo nextest run -p sui-adapter-transactional-tests gasless` (8 tests pass)
- [x] `cargo xclippy -D warnings`
- [x] Protocol config snapshots updated

### DIFF
--- a/crates/sui-adapter-transactional-tests/tests/gasless/gasless_remaining_balance.move
+++ b/crates/sui-adapter-transactional-tests/tests/gasless/gasless_remaining_balance.move
@@ -1,0 +1,100 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Tests dust prevention: withdrawals must not leave a remaining balance
+// between 0 and min_transfer in the sender's account.
+// gasless_verify_remaining_balance is enabled by --enable-gasless.
+
+//# init --addresses test=0x0 --accounts A B --enable-gasless --enable-accumulators
+
+//# publish --sender A
+#[allow(deprecated_usage)]
+module test::usdc {
+    use sui::coin;
+
+    public struct USDC has drop {}
+
+    fun init(otw: USDC, ctx: &mut TxContext) {
+        let (treasury_cap, metadata) = coin::create_currency(
+            otw, 6, b"USDC", b"USD Coin", b"", option::none(), ctx,
+        );
+        transfer::public_freeze_object(metadata);
+        transfer::public_transfer(treasury_cap, ctx.sender());
+    }
+}
+
+//# gasless-allow-token test::usdc::USDC --min-transfer 1000
+
+//# programmable --sender A --inputs 5000 object(1,1) @A
+//> 0: sui::coin::mint<test::usdc::USDC>(Input(1), Input(0));
+//> 1: sui::coin::into_balance<test::usdc::USDC>(Result(0));
+//> 2: sui::balance::send_funds<test::usdc::USDC>(Result(1), Input(2));
+
+//# create-checkpoint
+
+//# view-funds sui::balance::Balance<test::usdc::USDC> A
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs withdraw<sui::balance::Balance<test::usdc::USDC>>(4500) @B
+// Withdraw 4500 from 5000, leaving 500 dust - should fail at voting time
+//> 0: sui::balance::redeem_funds<test::usdc::USDC>(Input(0));
+//> 1: sui::balance::send_funds<test::usdc::USDC>(Result(0), Input(1));
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs withdraw<sui::balance::Balance<test::usdc::USDC>>(4000) @B
+// Withdraw 4000 from 5000, leaving 1000 (== min) - should succeed
+//> 0: sui::balance::redeem_funds<test::usdc::USDC>(Input(0));
+//> 1: sui::balance::send_funds<test::usdc::USDC>(Result(0), Input(1));
+
+//# create-checkpoint
+
+//# view-funds sui::balance::Balance<test::usdc::USDC> A
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs withdraw<sui::balance::Balance<test::usdc::USDC>>(1000) @B
+// Withdraw all remaining 1000, leaving 0 - should succeed
+//> 0: sui::balance::redeem_funds<test::usdc::USDC>(Input(0));
+//> 1: sui::balance::send_funds<test::usdc::USDC>(Result(0), Input(1));
+
+//# create-checkpoint
+
+//# view-funds sui::balance::Balance<test::usdc::USDC> A
+
+//# programmable --sender A --inputs 10000 object(1,1) @A
+// Refund A for send-back-to-self tests
+//> 0: sui::coin::mint<test::usdc::USDC>(Input(1), Input(0));
+//> 1: sui::coin::into_balance<test::usdc::USDC>(Result(0));
+//> 2: sui::balance::send_funds<test::usdc::USDC>(Result(1), Input(2));
+
+//# create-checkpoint
+
+//# view-funds sui::balance::Balance<test::usdc::USDC> A
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs withdraw<sui::balance::Balance<test::usdc::USDC>>(3000) @B 2500 @A
+// Withdraw 3000, send 2500 to B, send 500 (dust) back to self - should fail at execution
+//> 0: sui::balance::redeem_funds<test::usdc::USDC>(Input(0));
+//> 1: sui::balance::split<test::usdc::USDC>(Result(0), Input(2));
+//> 2: sui::balance::send_funds<test::usdc::USDC>(Result(1), Input(1));
+//> 3: sui::balance::send_funds<test::usdc::USDC>(Result(0), Input(3));
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs withdraw<sui::balance::Balance<test::usdc::USDC>>(3000) @B 2000 @A
+// Withdraw 3000, send 2000 to B, send 1000 back to self (>= min) - should succeed
+//> 0: sui::balance::redeem_funds<test::usdc::USDC>(Input(0));
+//> 1: sui::balance::split<test::usdc::USDC>(Result(0), Input(2));
+//> 2: sui::balance::send_funds<test::usdc::USDC>(Result(1), Input(1));
+//> 3: sui::balance::send_funds<test::usdc::USDC>(Result(0), Input(3));
+
+//# create-checkpoint
+
+//# view-funds sui::balance::Balance<test::usdc::USDC> A
+
+//# programmable --sender A --address-balance-gas --gas-price 0 --gas-budget 0 --inputs withdraw<sui::balance::Balance<test::usdc::USDC>>(2000) @B 1500u256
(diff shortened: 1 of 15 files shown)


crates/sui-core/src/accumulators/funds_read.rs
```diff
@@ -49,4 +50,41 @@ pub trait AccountFundsRead: Send + Sync {
...
+                Balance::maybe_get_balance_type_param(type_tag).unwrap_or_else(|| type_tag.clone());
+            if let Some(&min_amount) = min_amounts.get(&coin_type)
+                && min_amount > 0
+                && remaining < min_amount as u128
+            {
+                return Err(SuiErrorKind::UserInputError {
+                    error: UserInputError::InvalidWithdrawReservation {
+                        error: format!(
+                            "Gasless withdrawal from {} leaves {remaining} which is \
+                             below minimum {min_amount} for token type {coin_type}",
```
aschran, line 79: maybe say "below non-zero minimum" so it's a bit clearer that users may zero out their balances?

aschran (approved):

mystenmark:
Looks like this has the odd property if your balance is `R + A + A` where A is the minimum balance, you could reserve `R +  A` but would then be unable to redeem only `R + A - 1`.

I do not think this is a problem than anyone will run into and its too hard to fix, but just wanted to make sure i'm understanding this correctly.

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -618,6 +622,26 @@ impl<'backing> TemporaryStore<'backing> {
             }
         }
 
+        if let Some(reservations) = withdrawal_reservations {
```
tnowacki, line 625: Why do we need the withdrawal information here? Can't we just check the net_totals and filter out those that aren't for the sender?

crates/sui-core/src/accumulators/funds_read.rs
tnowacki: +1, I read this like three times and checked the test and was still confused lol. 
Maybe something like "Invalid gasless withdrawal from {}. Gasless transactions must either use the entire balance, or leave at least {min_amount} for token type {coin_type}, and the remaining amount of {remaining} is below that threshold" 
A bit wordy still, but trying to spell out it is either 0 or above the threshold

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -569,7 +569,11 @@ impl<'backing> TemporaryStore<'backing> {
     /// - No new objects were created or existing objects mutated (written_objects is empty)
     /// - The set of deleted objects exactly equals the set of input Coin objects
     /// - Each recipient receives at least the minimum transfer amount per token type
-    pub fn check_gasless_execution_requirements(&self) -> Result<(), String> {
+    /// - Unused withdrawal reservation (reservation - actual split) is 0 or >= min_amount
+    pub fn check_gasless_execution_requirements(
+        &self,
```
tnowacki, line 574: Also, can we check this up front before execution? Obviously we cannot in general, but the current form of gasless is restrictive enough that we maybe could? (it could be at least approximate it with failing to allow execution when it cannot be statically determined)

alex-mysten: This is exactly where I started - statically analyzing the PTB based on arguments and move commands at voting time to decide how much of the balance was consumed.

At voting time, none of the ptb structure or arguments have been semantically validated. There's a lot of validation surface area that would be moved up from execution to voting time - argument arity, type, overflow considerations, etc etc that we need to consider.

So the answer is yes, we can do this, but I'm not sure we want to be in the business of performing move argument validation upstream of execution. Also it would be brittle. If we have a feature requirement where the amount to send is computed, the static-validation model breaks.

sui-execution/latest/sui-adapter/src/temporary_store.rs
alex-mysten: We have two independent checks we want to apply:
1. at voting time, sender is prevented from sending a Withdrawal that would leave only dust in their balance.
2. at execution time, of that Withdrawal, they may partially or fully consume it, but they must not leave dust

net totals would not be sufficient as it knows nothing about the sender's actual balance

alex-mysten:
> Looks like this has the odd property if your balance is `R + A + A` where A is the minimum balance, you could reserve `R + A` but would then be unable to redeem only `R + A - 1`.
> 
> I do not think this is a problem than anyone will run into and its too hard to fix, but just wanted to make sure i'm understanding this correctly.

@mystenmark correct. a ptb that leaves only dust remaining from the amount that they withdrew would be rejected. a small advantage of going the static analysis route is that we'd just compare against user balance at submission, so this case wouldn't happen. but this has the tradeoffs I called out above

sui-execution/latest/sui-adapter/src/temporary_store.rs
tnowacki: Yeah I agree on the brittle ness. I mostly saw it as a spam protection system, not so much as a way of actually validating the "correctness" of the txn. 

I think things would be different if we thought that this was all the expressivity we would have under gasless transactions--now and forever. But since the scope will increase, it is not the spot we should rely on.

crates/sui-adapter-transactional-tests/tests/gasless/gasless_remaining_balance.move
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) Mysten Labs, Inc.
```
tnowacki, line 1: As a small nit, could you split these into failing and non-failing transactions? It makes it a lot easier for me to understand what behavior is allowed vs disallowed.

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -683,6 +691,37 @@ mod checked {
...
+            if let CallArg::FundsWithdrawal(fw) = input
+                && let Some(coin_type) = fw.type_arg.get_balance_type_param()
+            {
+                let owner = match fw.withdraw_from {
+                    WithdrawFrom::Sender => sender,
+                    WithdrawFrom::Sponsor => {
+                        debug_fatal!(
+                            "WithdrawFrom::Sponsor is not allowed in gasless transactions"
+                        );
+                        return None;
```
tnowacki, line 714:
```suggestion
                        continue;
```

tnowacki: We probably want to harden here by still attempting to look at all the reservations

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -683,6 +691,37 @@ mod checked {
...
+    ) -> Option<BTreeMap<(SuiAddress, TypeTag), u64>> {
+        let TransactionKind::ProgrammableTransaction(pt) = transaction_kind else {
+            debug_fatal!("Gasless transaction must be a ProgrammableTransaction");
+            return None;
+        };
+        let sender = tx_ctx.sender();
+        let mut reservations = BTreeMap::<(SuiAddress, TypeTag), u64>::new();
+        for input in &pt.inputs {
+            if let CallArg::FundsWithdrawal(fw) = input
+                && let Some(coin_type) = fw.type_arg.get_balance_type_param()
```
tnowacki, line 706: I feel we probably shouldn't silently fail right now when this is none? Maybe? 
In short, I think it is impossible for this to be anything other than a balance at the moment so it should maybe be like 

```suggestion
           let CallArg::FundsWithdrawal(fw) = input else { continue };
           let Some(coin_type) = fw.type_arg.get_balance_type_param() else {
               debug_fatal!("we should only have balance types for withdrawals at this point");
               continue;
            };
```

sui-execution/latest/sui-adapter/src/temporary_store.rs
tnowacki: Do we have a minimum transfer amount? Or just a minimum balance remaining

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -618,6 +622,26 @@ impl<'backing> TemporaryStore<'backing> {
...
 
+        if let Some(reservations) = withdrawal_reservations {
+            for ((owner, token_type), &reserved) in reservations {
+                let net = net_totals
+                    .get(&(*owner, token_type.clone()))
+                    .copied()
+                    .unwrap_or(0);
+                let remaining = (reserved as i128).saturating_add(net);
+                if remaining > 0
+                    && let Some(&min_amount) = allowed_types.get(token_type)
```
tnowacki, line 633:
```suggestion
                    && let Some(&min_balance_remaining) = allowed_types.get(token_type)
```

Saying `min_balance_remaining` instead of `min_amount` might have made it easier for me to grok this at the first go around (worth changing in the other spot too maybe)

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -683,6 +691,37 @@ mod checked {
...
+                    WithdrawFrom::Sender => sender,
+                    WithdrawFrom::Sponsor => {
+                        debug_fatal!(
+                            "WithdrawFrom::Sponsor is not allowed in gasless transactions"
+                        );
+                        return None;
+                    }
+                };
+                let Reservation::MaxAmountU64(amount) = fw.reservation;
+                let entry = reservations.entry((owner, coin_type)).or_default();
```
tnowacki, line 718: Nit only for my general distrust of `Default` (you also don't use `unwrap_or_default()` below so maybe just consistency 🤷 take or leave it, I don't care really. We just hate `Default` in external crates land)
```suggestion
                let entry = reservations.entry((owner, coin_type)).or_insert(0);
```

sui-execution/latest/sui-adapter/src/temporary_store.rs
alex-mysten: We have both. Minimum transfer amount is already implemented

crates/sui-core/src/accumulators/funds_read.rs
alex-mysten: Addressed

crates/sui-adapter-transactional-tests/tests/gasless/gasless_remaining_balance_accept.move
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) Mysten Labs, Inc.
```
alex-mysten: Split these as suggested

sui-execution/latest/sui-adapter/src/execution_engine.rs
alex-mysten: We now look at all the reservations. I left the debug_assert since we would probably want to check this again if we did ever change the rule to disallow sponsor reservations

sui-execution/latest/sui-adapter/src/execution_engine.rs
alex-mysten: Addressed

sui-execution/latest/sui-adapter/src/temporary_store.rs
alex-mysten: Addressed

sui-execution/latest/sui-adapter/src/execution_engine.rs
alex-mysten: Addressed

tnowacki (approved):

PR #26265, opened 2026-04-16
[genesis] Directly splash system packages at genesis

tzakian:
## Description 

Due to the way that we were previously publishing system packages at genesis we were running them through the non-system-package publication path, and all we were doing was using it to create the object.

Just skip all of that and just splash the package down in genesis.

## Test plan 

CI 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


dariorussi:
wow that is quite a clean up.
Adding @bmwill to double check but I like it a lot

bmwill (approved):

bmwill:
One note that we should still worry about is what do we do when the system package gets larger?

dariorussi:
> One note that we should still worry about is what do we do when the system package gets larger?

though they will likely never get that large?

PR #26268, opened 2026-04-16
sui-indexer-alt-framework-store-traits: add shared connection test macros

evan-wall-mysten:
## Description 

Adds shared harness-based tests for the store trait contracts in `sui-indexer-alt-framework-store-traits`.

The shared macros are split by capability:

- `connection_tests!`
- `concurrent_connection_tests!`
- `sequential_connection_tests!`

They cover common `Connection`, `ConcurrentConnection`, and `SequentialConnection` behavior including watermark initialization, committer watermark progression and stale-write rejection, chain-id acceptance/rejection, reader/pruner stale-write rejection, and sequential transaction rollback.

This also adds a trait-crate in-memory `MockStore` for exercising the shared tests. The framework test store is now `FallibleMockStore`; it delegates shared watermark and chain-id behavior to the trait-crate `MockStore` while retaining framework-specific failure, data, transaction, and pruning controls.

Adopters in this branch:

- `sui-indexer-alt-object-store`
- `sui-pg-db`
- `sui-kvstore`

Implementation-specific tests remain where the behavior is storage-specific, such as object-store watermark format migration and pruner wait/saturation behavior.

---

Failed macro assertions backtraces contain line with assertion:
`/Users/evanwall/workspace/sui2/crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs:59:9`

and line where macro is called from crate's test mod:
`sui_indexer_alt_object_store::tests::connection_tests::init_watermark_fresh_without_checkpoint::{{closure}}
             at ./src/lib.rs:480:5`

Example:
```
test tests::connection_tests::init_watermark_fresh_without_checkpoint ... FAILED

failures:

---- tests::connection_tests::init_watermark_fresh_without_checkpoint stdout ----

thread 'tests::connection_tests::init_watermark_fresh_without_checkpoint' (5476533) panicked at /Users/evanwall/workspace/sui2/crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs:59:9:
assertion failed: false
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/panicking.rs:80:14
   2: core::panicking::panic
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/panicking.rs:150:5
   3: sui_indexer_alt_framework_store_traits::testing::Harness::init_watermark_fresh_without_checkpoint::{{closure}}
             at /Users/evanwall/workspace/sui2/crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs:59:9
   4: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /Users/evanwall/.rustup/toolchains/1.92-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/future/future.rs:133:9
   5: sui_indexer_alt_object_store::tests::connection_tests::init_watermark_fresh_without_checkpoint::{{closure}}
             at ./src/lib.rs:480:5
```

## Test plan 

New macro tests pass.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: sui-indexer-alt-framework-store-traits: add store/connection test macros


crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+const PIPELINE: &str = "pipeline";
+const EPOCH_HI: u64 = 7;
+const CHECKPOINT_HI: u64 = 200;
+const TX_HI: u64 = 42;
+const TIMESTAMP_MS_HI: u64 = 99;
+const READER_LO: u64 = 123;
+const PRUNER_HI: u64 = 77;
+const ONE_HOUR_MS: u64 = 3_600_000;
+
+fn watermark() -> CommitterWatermark {
```
amnn: nit: the function name here is quite generic -- the body seems to suggest it's some sort of sentinel high watermark, is there a more specific name that captures what this represents better?

Also it seems to be a private function, can we move it lower in the file order?

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+/// async fn make_store() -> ((), MyStore) { ((), MyStore::new()) }
+///
+/// sui_indexer_alt_framework_store_traits::store_tests! {
+///     setup: make_store,
+///     concurrent,
+///     sequential,
+/// }
+/// ```
+#[macro_export]
+macro_rules! store_tests {
```
amnn: Could we separate this into three macros that can be called individually?

```rust
connection_tests!(harness);
concurrent_tests!(harness);
sequential_tests!(harness);
```

The current set-up introduces syntax specific to this macro, and it also gives the impression that you could do things like:

```rust
store_tests!(setup: ..., sequential, concurrent);
```

When in fact you can't because the pattern only detects `concurrent, sequential`.

crates/sui-indexer-alt-object-store/src/lib.rs
```diff
@@ -164,14 +164,19 @@ impl Connection for ObjectStoreConnection {
     ) -> anyhow::Result<Option<InitWatermark>> {
         let object_path = watermark_path(pipeline_task);
         let reader_lo = checkpoint_hi_inclusive.map_or(0, |cp| cp + 1);
+        // Match the postgres impl (and the `set_reader_watermark` contract): anchor
+        // `pruner_timestamp_ms` at wall-clock time on first write, so the pruner's
+        // `(pruner_timestamp + delay) - now` computation is meaningful before any reader-lo
+        // advancement has happened.
+        let pruner_timestamp_ms = SystemTime::now().duration_since(UNIX_EPOCH)?.as_millis() as u64;
```
amnn: This change is a bit of a flag because it's implementing the wrong behaviour -- when setting the pruning watermark, it's important the timestamp comes from a source that all accessors can synchronize on to minimise skew, i.e. not the indexer's local timestamp. This is why in `sui-pg-db`, we use the timestamp coming form the DB.

We should revisit the contracts in the tests if they encourage folks to implement the pruner timestamp in this way.

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+macro_rules! __store_tests_emit_connection {
+    (setup: $setup:path) => {
+        mod connection_tests {
+            use super::*;
+
+            #[::tokio::test]
+            async fn init_watermark_fresh_without_checkpoint() {
+                let (_guard, store) = $setup().await;
+                $crate::testing::init_watermark_fresh_without_checkpoint(store).await;
+            }
```
amnn: Can you reduce the boilerplate here with a macro that generates this wrapper? Something like:

```rust
macro_rule! __test_case {
    $Harness:path :: $case:ident) => {
        #[::tokio::test]
        async fn $case() {
            $Harness::new().await.$case().await;
        }
    }
}

macro_rules! connection_tests {
    ($Harness:path) => {
        __test_case!($Harness::init_watermark_fresh_without_checkpoint);
        __test_case!($Harness::init_watermark_fresh_with_checkpoint);
        // ...
    }
}
```

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+// `store_tests!` macro — emits thin `#[tokio::test]` wrappers that delegate to
+// the helpers above.
+// =============================================================================
+
+/// Generate the union of `Connection`/`ConcurrentConnection`/`SequentialConnection` trait tests
+/// against a caller-provided `setup` function.
+///
+/// `setup` is the path to an `async fn` that returns `(Guard, Store)`. The `Guard` is bound for
+/// the test body's lifetime — callers that don't need a real guard return `()` as the first
+/// element of the tuple.
```
amnn: It may be neater to structure your set-up around traits, something like. This way the constraints on the store are documented in one place (the specialised harness traits), and harnesses can hide whether or not they need to keep something alive for the duration of the test behind the `Harness` trait impl.

```rust
#[async_trait]
trait Harness {
    type S: Store;

    async fn new() -> Self;
    fn store() -> &S;

    async fn init_watermark_fresh_without_checkpoint(&self) {
        let mut conn = self.store().connect().await.unwrap();
        let init = conn.init_watermark(PIPELINE, None).await.unwrap().unwrap();
        assert_eq!(init.checkpoint_hi_inclusive, None);
    }

    // ... other generic connection/store traits.
}

#[async_trait]
trait ConcurrentHarness: Harness where Self::S: ConcurrentStore {
    // concurrent store specific tests
}

#[async_trait]
trait SequentialHarness: Harness where Self::S: SequentialStore {
    // sequential store specific tests
}

macro_rules! connection_tests {
    ($Harness:path) => {
        __test_case!($Harness::init_watermark_fresh_without_checkpoint);
        // ... etc
    }
}

macro_rules! concurrent_tests {
    ($Harness:path) => {
        impl ConcurrentHarness for $Harness {}
        // ...
    }
}

macro_rules! sequential_tests {
    ($Harness:path) => {
        impl SequentialHarness for $Harness {}
        // ....
    }
}
```

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+        conn.set_committer_watermark(PIPELINE, watermark())
+            .await
+            .unwrap()
+    );
+
+    let stored = conn.committer_watermark(PIPELINE).await.unwrap().unwrap();
+    assert_eq!(stored.epoch_hi_inclusive, EPOCH_HI);
+    assert_eq!(stored.checkpoint_hi_inclusive, CHECKPOINT_HI);
+    assert_eq!(stored.tx_hi, TX_HI);
+    assert_eq!(stored.timestamp_ms_hi_inclusive, TIMESTAMP_MS_HI);
```
amnn: nit: Re-use the fields from `watermark()`?

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+    };
+    assert!(conn.set_committer_watermark(PIPELINE, lower).await.unwrap());
+    assert!(
+        conn.set_committer_watermark(PIPELINE, watermark())
+            .await
+            .unwrap()
+    );
+
+    let stored = conn.committer_watermark(PIPELINE).await.unwrap().unwrap();
+    assert_eq!(stored.checkpoint_hi_inclusive, CHECKPOINT_HI);
```
amnn: nit (optional): prevents us having to skip up to the top of the file to see how `watermark` is defined.

```suggestion
    let hi = watermark();
    let lo = CommitterWatermark {
        checkpoint_hi_inclusive: hi.checkpoint_hi_inclusive / 2
        ..hi
    };

    assert!(conn.set_committer_watermark(PIPELINE, lo).await.unwrap());
    assert!(conn.set_committer_watermark(PIPELINE, hi).await.unwrap());

    let stored = conn.committer_watermark(PIPELINE).await.unwrap().unwrap();
    assert_eq!(stored.checkpoint_hi_inclusive, hi.checkpoint_hi_inclusive);
```

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+    assert!(
+        conn.set_reader_watermark(PIPELINE, READER_LO)
+            .await
+            .unwrap()
+    );
+    let watermark = conn.reader_watermark(PIPELINE).await.unwrap().unwrap();
+    assert_eq!(watermark.reader_lo, READER_LO);
+}
+
+pub async fn pruner_watermark_wait_for_ms<S: ConcurrentStore>(store: S) {
```
amnn: Given what I mentioned below, my vote would be to remove this test (IIUC, it's new, right?)

We need to provide too much slack for CI for this to be useful, and limiting the slack/skew is the property that we would care to test for this implementation. 

That's difficult to achieve, but having a test that seems to accept a high skew gives legitimacy to implementations that are wrong (like the ones added to object store and KV store).

Instead, in this case, can we:

- Remove this test,
- undo those additions to Object Store and KV store,
- add generous documentation to the trait for `set_pruner_watermark` to explain what is expected of Store impls? I believe there are some docs elsewhere, maybe in the framework where this is used, that are similar, but it seems helpful to surface on the trait docs.

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+    // Generous slack for slow CI.
+    assert!(
+        watermark.wait_for_ms > (ONE_HOUR_MS as i64 - 100_000)
+            && watermark.wait_for_ms <= ONE_HOUR_MS as i64,
+        "wait_for_ms = {}",
+        watermark.wait_for_ms,
+    );
+}
+
+pub async fn pruner_watermark_saturates_when_ready<S: ConcurrentStore>(store: S) {
```
amnn: Ditto, let's remove this test.

amnn (approved):
Mostly looks good, I think the macro approach you have gone for is better than the custom runner approach -- I've left some suggestions to hopefully reduce the boilerplate a little more.

The only issue that needs to be addressed before landing are the changes around pruner watermark (testing `wait_for_ms` behaviour, and that driving changes in stores that don't currently use `wait_for_ms` and have been forced to implement something wrong there to conform.

Let me know if the suggested fix there makes sense, otherwise feel free to re-request review, and I will take another look. Otherwise accepted to unblock.

crates/sui-indexer-alt-framework-store-traits/src/testing/mock_store.rs
```diff
@@ -0,0 +1,652 @@
...
+/// Configuration for simulating failures in tests
+#[derive(Default)]
+pub struct Failures {
+    /// Number of failures to simulate before allowing success
+    pub failures: usize,
+    /// Counter for tracking total transaction attempts
+    pub attempts: AtomicUsize,
+}
+
+/// Mock store for testing pipelines driven by one or more indexers.
```
wlmyng: With the refactor, I wonder if a doc comment like ... 
```
/// In-memory `Store`/ `Connection` for testing framework mechanics. Operators should largely test their real store implementations.
```

I suppose we can't really make this pub(crate) since the framework uses it. Maybe we can move it into `sui-indexer-alt-framework`?

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+    let mut conn = store.connect().await.unwrap();
+
+    conn.init_watermark(PIPELINE, Some(CHECKPOINT_HI))
+        .await
+        .unwrap();
+    conn.set_committer_watermark(PIPELINE, watermark())
+        .await
+        .unwrap();
+
+    // Second init must surface the existing checkpoint, not the input.
```
wlmyng: This might be better off as a `///` comment? Above each helper, a description of what is being tested. For this one it's the idempotent-init of `init_watermark`

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+// =============================================================================
+// ConcurrentConnection-trait test helpers
+// =============================================================================
+
+pub async fn reader_watermark_roundtrip<S: ConcurrentStore>(store: S) {
+    let mut conn = store.connect().await.unwrap();
+    concurrent_bootstrap(&mut conn, CHECKPOINT_HI).await;
+
+    let watermark = conn.reader_watermark(PIPELINE).await.unwrap().unwrap();
+    assert_eq!(watermark.checkpoint_hi_inclusive, CHECKPOINT_HI);
```
wlmyng: this coudl be in the concurrent_bootstrap itself?

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+
+    assert!(
+        conn.set_reader_watermark(PIPELINE, READER_LO)
+            .await
+            .unwrap()
+    );
+    let watermark = conn.reader_watermark(PIPELINE).await.unwrap().unwrap();
+    assert_eq!(watermark.reader_lo, READER_LO);
```
wlmyng: similarly here ... in our testing we could assert that the particular watermark is indeed set

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
wlmyng: We could inject a clock into `MockStore` .. or swap the mock's time for `tokio`, then use `tokio::time` and `#[tokio::test(start_paused=true)]`, given that the pruner runs on an interval. We just need mockstore to agree on the clock source

Although using paused time probably defaults the purpose of this test?

wlmyng: Yeah also I think this is a framework-logic test (does the pruner/committer/collector et al task do the right thing), where the store is a dependency, but what we want are store-contract tests where the store is the subject

crates/sui-indexer-alt-object-store/src/lib.rs
```diff
@@ -849,16 +698,6 @@ mod tests {
...
-        let mut conn = store_conn().await;
-
-        let chain_id = [1u8; 32];
-        assert!(conn.accepts_chain_id(PIPELINE, chain_id).await.unwrap());
-        // Second call with the same chain_id should also accept
-        assert!(conn.accepts_chain_id(PIPELINE, chain_id).await.unwrap());
-    }
-
     #[tokio::test]
     async fn test_accepts_chain_id_mismatching_rejects() {
```
wlmyng: i think this one can also be removed?

wlmyng: or, could leave a comment about why we keep this and the above `writes_and_accepts` even though already tested by `store_tests!`

crates/sui-indexer-alt-object-store/src/lib.rs
```diff
@@ -660,7 +620,12 @@ mod tests {
         );
         assert_eq!(migrated_watermark.reader_lo, migrated_reader_lo);
         assert_eq!(migrated_watermark.pruner_hi, migrated_reader_lo);
-        assert_eq!(migrated_watermark.pruner_timestamp_ms, 0);
+        // Migrated watermark anchors pruner_timestamp_ms at the time init_watermark ran.
+        assert!(
```
wlmyng: imo this feels more like `object-store` crate testing framework logic behavior? I think the store test just needs to verify that we /can/ write timestamps

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
```diff
@@ -0,0 +1,598 @@
...
+            async fn transaction_rolls_back_on_error() {
+                let (_guard, store) = $setup().await;
+                $crate::testing::transaction_rolls_back_on_error(store).await;
+            }
+        }
+    };
+}
+
+#[cfg(test)]
+mod sanity_tests {
```
wlmyng: do we want to think of `MockStore` as some canonical reference impl? `MockStore` also comes with a bunch of other things like failure injection, polling observers, sequential-pipeline recording, framework-related test scaffolding that don't necessarily apply here

maybe we can move both `MockStore` and these sanity checks to the framework?

Or

a stripped down `MockStore` in-memory, which the framework extends on with all the other scaffolding (that is specific to framework tasks testing)

wlmyng (commented):
pretty dope

crates/sui-indexer-alt-framework-store-traits/src/testing/mock_store.rs
evan-wall-mysten: Updated the comment. I put these mocks here because they are used in the unit tests of the macro and adding it to `sui-indexer-alt-framework` would create a circular dependency with `sui-indexer-alt-framework-store-traits`.

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
evan-wall-mysten: I added `concurrent_bootstrap` just to reduce boiler plate. These assertions are in just one of the test cases to avoid needing duplicate asserts and because putting them in the helper shows the wrong line number when they break.

crates/sui-indexer-alt-object-store/src/lib.rs
evan-wall-mysten: Removed

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
evan-wall-mysten: I put `MockStore` in `sui-indexer-alt-framework-store-traits` to avoid a circular dependency with `sui-indexer-alt-framework`. I'll move the parts that are relevant to the framework tests into the framework.

crates/sui-indexer-alt-object-store/src/lib.rs
evan-wall-mysten: Removed

crates/sui-indexer-alt-framework-store-traits/src/testing/mod.rs
evan-wall-mysten: Added an assertion on the watermark instead of just the field.

PR #26269, opened 2026-04-16
[sui-package-resolver] Parametrize `Resolver` by async/sync and introduce `SyncResolver`

tzakian:
## Description

Sadly there is no easy way of parametrizing by sync/async other than through macro generation. This adds a `SyncResolver` which is basically the `Resolver` but with a sync backing store and exposes a sync interface.

This also adds a `datatype_layout` function which is just a specialization of the `type_layout` function to make it easier in PRs to come. I can remove it for now if that's preferable, but we will need to add it back later then, so figured might as well do it now.

## Test plan 

CI/existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-package-resolver/src/lib.rs
+++ b/crates/sui-package-resolver/src/lib.rs
@@ -25,6 +25,7 @@ use move_binary_format::file_format::StructDefinitionIndex;
 use move_binary_format::file_format::StructFieldInformation;
 use move_binary_format::file_format::TableIndex;
 use move_binary_format::file_format::Visibility;
+use move_binary_format::partial_vm_error;
 use move_command_line_common::display::RenderResult;
 use move_command_line_common::display::try_render_constant;
 use move_command_line_common::error_bitset::ErrorBitset;
@@ -76,8 +77,6 @@ pub struct Resolver<S> {
 #[derive(Debug)]
 pub struct SyncResolver<S> {
     package_store: S,
-    // Read by the sync resolution methods added in a later commit.
-    #[allow(dead_code)]
     limits: Option<Limits>,
 }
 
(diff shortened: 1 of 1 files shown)


crates/sui-package-resolver/src/lib.rs
```diff
@@ -305,12 +318,18 @@ struct ResolutionContext<'l> {
 /// Interface to abstract over access to a store of live packages.  Used to override the default
 /// store during testing.
 #[async_trait]
-pub trait PackageStore: Send + Sync + 'static {
+pub trait PackageStore: Send + Sync {
```
tzakian, line 321: Don't need this `'static` AFAICT so removing it.

crates/sui-package-resolver/src/lib.rs
```diff
@@ -332,6 +351,23 @@ impl<S: PackageStore> PackageStore for Arc<S> {
     }
 }
 
+/// Every [`BackingPackageStore`] implementation is usable as a [`SyncPackageStore`]. The blanket
+/// also covers `&B`, `Arc<B>`, `Box<B>` for any `B: BackingPackageStore` via
+/// [`BackingPackageStore`]'s own forwarding impls.
+impl<S: BackingPackageStore> SyncPackageStore for S {
```
amnn, line 357: I was originally going to ask whether it's possible to move this into `sui-types` -- the answer is no I think, because it would create a circular dependency between crates (I was just thinking about it because in this case we are implementing `SyncPackageStore` "on" `BackingPackageStore` so usually this would be colocated with the thing the trait is implemented on, but that's not exclusively the case and it's fine that it's here).

crates/sui-package-resolver/src/lib.rs
```diff
@@ -424,6 +484,29 @@ impl<S: PackageStore> Resolver<S> {
...
+            layout => {
+                // This error is unexpected because a `StructTag` should always resolve to either a
+                // struct or enum layout, and if it doesn't, it likely means there is a bug in the
+                // resolver logic.
+                Err(Error::UnexpectedError(Arc::new(partial_vm_error!(
+                    UNKNOWN_INVARIANT_VIOLATION_ERROR,
+                    "StructTag {:#?} did not resolve to a struct or enum layout: {:#?}",
+                    struct_tag,
+                    layout
+                ))))
```
amnn, line 505: Does this need to use `partial_vm_error` -- do we feel brave enough to change the error wording/structure to completely separate this from the VM error type?

crates/sui-package-resolver/src/lib.rs
```diff
@@ -636,6 +719,241 @@ impl<S: PackageStore> Resolver<S> {
...
+            /* visit_fields */ false,
+            /* visit_phantoms */ true,
+        )?;
+
+        // (2). Use that information to relocate package IDs in the type.
+        context.canonicalize_type(&mut tag)?;
+        Ok(tag)
+    }
+
+    pub fn type_layout(&self, mut tag: TypeTag) -> Result<MoveTypeLayout> {
```
amnn, line 744: Duplicate the doc comments? (for LSP)

crates/sui-package-resolver/src/lib.rs
```diff
@@ -636,6 +719,241 @@ impl<S: PackageStore> Resolver<S> {
...
+
+        // (2). Use that information to relocate package IDs in the signature.
+        for sig in def.parameters.iter_mut().chain(def.return_.iter_mut()) {
+            context.relocate_signature(&mut sig.body)?;
+        }
+
+        Ok(def)
+    }
+
+    pub fn pure_input_layouts(
```
amnn, line 837: There's potentially an avenue for deduplicating some of this stuff (as a follow-up)?

It would require using a similar pattern to the resolution context with function signatures (it's already a bit sub-optimal that we are fetching signatures sequentially in a loop in the async case).

amnn (approved):
LGTM, modulo the comments/questions, but they are not blocking.

PR #26270, opened 2026-04-16
[forking 8/n] add RPC support through `sui-rpc-api` and its gRPC interfaces

stefan-mysten:
## Description 

This PR adds support for querying and submitting transactions to the forked network via gRPC. The `sui-rpc-api` crate provides mostly all the required implementation, except implementing a few traits which this PR does.
- implements `ReadStore` for `DataStore`
- implements `RpcStateReader` for `DataStore`
- adds `DataStore` to `Context` as a field so that is easily shared with the RPC service
- wires up the RPC service through the existing `RpcService` type at startup
- implements missing APIs to handle historical checkpoints, transactions, objects.

Note: I left aside transaction execution as that will be added in the next PR

## Test plan 

A few new unit tests and local manual tests via CLI and grpc queries.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-forking/src/context.rs
```diff
@@ -14,13 +14,22 @@ use crate::store::DataStore;
...
+    /// through the `RwLock` that protects the simulacrum.
+    data_store: DataStore,
     chain_identifier: Chain,
 }
 
 impl Context {
-    pub(crate) fn new(simulacrum: Simulacrum<OsRng, DataStore>, chain_identifier: Chain) -> Self {
+    pub(crate) fn new(
+        simulacrum: Simulacrum<OsRng, DataStore>,
+        data_store: DataStore,
```
jordanjennings-mysten: are you basically passing data store twice here? can you reach into simulacrum and use that?

jordanjennings-mysten (approved):

crates/sui-forking/src/context.rs
stefan-mysten: You're right, fixed.

crates/sui-forking/src/store.rs
```diff
@@ -628,6 +734,160 @@ impl SimulatorStore for DataStore {
...
+    }
+
+    fn get_latest_checkpoint(&self) -> StorageResult<VerifiedCheckpoint> {
+        self.get_highest_verified_checkpoint()
+            .map_err(|e| StorageError::custom(e.to_string()))?
+            .ok_or_else(|| StorageError::missing("no checkpoint persisted yet"))
+    }
+
+    fn get_highest_verified_checkpoint(&self) -> StorageResult<VerifiedCheckpoint> {
+        DataStore::get_highest_verified_checkpoint(self)
```
amnn, line 753:
```suggestion
        self.get_highest_verified_checkpoint()
```

Also below

amnn: Scratch that, I see now why it's like this (`ReadStore` and `DataStore` have functions with the same name) -- that's kind of unfortunate.

stefan-mysten: I am happy to change the `DataStore` API as needed, but keeping them the same keeps me sane, to not try to understand if there are any differences.

PR #26271, opened 2026-04-16
typed-store: fix safe_drop_db lock race in tidehunter builds

andll:
## Summary
- In tidehunter builds, `safe_drop_db` was a cfg-gated override that only invoked `TideHunterDb::drop_db`. Pure-RocksDB stores (e.g. `rpc-index`) re-initialize by dropping the handle, calling `safe_drop_db`, then reopening — the non-TH variant's `rocksdb::DB::destroy` retry loop was what papered over the async `Drop` of the previous handle. Without it, re-open panics with `"lock hold by current process"` because the path is still in RocksDB's in-process `locked_files_` set when the pending `UnlockFile` hasn't fired yet.
- Replace the two cfg-gated variants with a single runtime-dispatched function: detect tidehunter paths by the presence of `shape.yaml` and route to the matching cleanup; everything else (including pure-RocksDB stores in a tidehunter-enabled binary) uses `rocksdb::DB::destroy` with the retry loop.
- Unified return type to `std::io::Error`. All callers (`rpc_index.rs`, `db_tool`, `authority_per_epoch_store_pruner`, `consensus_store_pruner`, tests) either use `?` into `anyhow::Result`, `.expect(...)`, or match on `Display`, so no caller-side changes needed.

## Test plan
- [x] `cargo check -p typed-store -p sui-core -p sui-tool`
- [x] `USE_TIDEHUNTER=1 cargo check -p typed-store -p sui-core -p sui-tool --features typed-store/tidehunter`
- [x] `cargo clippy -p typed-store --lib --no-deps` (both feature configs)
- [x] `cargo nextest run -p typed-store --lib test_safe_drop_db` (both feature configs)
- [ ] Weekly Antithesis run with this branch to confirm the rpc-index panic is gone

### DIFF
--- a/crates/typed-store/src/rocks/mod.rs
+++ b/crates/typed-store/src/rocks/mod.rs
@@ -2047,8 +2047,19 @@ pub fn open_cf_opts_secondary<P: AsRef<Path>>(
 }
 
 // Drops a database if there is no other handle to it, with retries and timeout.
-#[cfg(not(tidehunter))]
-pub async fn safe_drop_db(path: PathBuf, timeout: Duration) -> Result<(), rocksdb::Error> {
+// Detects the storage variant (RocksDB vs. tidehunter) from the directory
+// contents and dispatches to the matching cleanup. Both variants coexist in
+// tidehunter builds — some stores (e.g. rpc-index) are pure RocksDB even when
+// the tidehunter feature is enabled elsewhere in the binary.
+pub async fn safe_drop_db(path: PathBuf, timeout: Duration) -> Result<(), std::io::Error> {
+    #[cfg(tidehunter)]
+    if is_tidehunter_db(&path) {
+        return safe_drop_tidehunter_db(path, timeout).await;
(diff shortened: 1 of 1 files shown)


crates/typed-store/src/rocks/mod.rs
```diff
@@ -2047,8 +2047,19 @@ pub fn open_cf_opts_secondary<P: AsRef<Path>>(
 }
 
 // Drops a database if there is no other handle to it, with retries and timeout.
-#[cfg(not(tidehunter))]
```
johnjmartin, line 2050: intentionally removing this?

johnjmartin (approved):

crates/typed-store/src/rocks/mod.rs
andll: This is just artifact of how diff is shown, this function is just renamed to `safe_drop_rocksdb`

PR #26275, opened 2026-04-16
move][signed ints] Add signed integers to the move stdlib [4/n] 

cgswords:
## Description 

Add signed to the the external-crates stdlib

## Test plan 

Move tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-stdlib/sources/i128.move
+++ b/external-crates/move/crates/move-stdlib/sources/i128.move
@@ -1,0 +1,99 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#[defines_primitive(i128)]
+module std::i128;
+
+use std::string::String;
+
+/// Maximum value for an `i128`
+public macro fun max_value(): i128 {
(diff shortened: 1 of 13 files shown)


external-crates/move/crates/move-stdlib/sources/i128.move
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) Mysten Labs, Inc.
```
tnowacki, line 1: No mysten copy rights on the external crates side of things--all Move Core contributors (or whatever it is)

tnowacki (changes requested):
Our source of truth these days is really the sui stdlib. And we cut over to here on occasion, though it has been a while. 

Do you need these for tests or something? If not, what are your thoughts on waiting a release or two of having signed in support in sui and the compiler before adding these? My thought is we can kind of soft launch signed ints and then later come in and add stdlib support. I think this might be the easiest for external breakages of all sorts

PR #26278, opened 2026-04-17
Remove dead code gated on mysticeti_fastpath / disable_preconsensus_locking

mwtian:
## Summary

- `mysticeti_fastpath` is enabled on all chains since Mainnet v96; `disable_preconsensus_locking` since v105. Both accessors always return `true` now.
- Delete the call-site conditionals in `sui-core` and `sui-single-node-benchmark`: unwrap the live (always-taken) branches, drop the dead (never-taken) branches.

### Notable changes
- `AuthorityState::wait_for_transaction_execution_for_testing` loses its dead `!disable_preconsensus_locking()` enqueue path and its now-unused `epoch_store` parameter; updated two callers.
- `vote_transactions` early-return, `handle_submit_transaction_inner` `UnsupportedFeatureError`, and the "Ignoring MFP transaction" skip block are all deleted as unreachable.
- `to_consensus_protocol_config` passes `true` literally for the `mysticeti_fastpath` argument that still flows into `consensus-core`.

## Test plan

- [x] `cargo check -p sui-core -p sui-single-node-benchmark`
- [x] `cargo clippy -p sui-core -p sui-single-node-benchmark -p sui-protocol-config -- -D warnings`
- [x] `cargo fmt`
- [x] `cargo nextest run -p sui-protocol-config` — snapshot tests pass with zero diff (lib.rs untouched)
- [x] `cargo nextest run -p sui-core --lib --no-fail-fast` — 650/650 pass
- [x] CI simtests / e2e

### DIFF
--- a/crates/sui-core/src/consensus_handler.rs
+++ b/crates/sui-core/src/consensus_handler.rs
@@ -1198,7 +1198,7 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
             &commit_info,
             &consensus_commit,
         );
-        // Buffer owned object locks for batch write when preconsensus locking is disabled
+        // Buffer owned object locks for batch write.
         if !owned_object_locks.is_empty() {
             state.output.set_owned_object_locks(owned_object_locks);
         }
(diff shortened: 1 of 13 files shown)


crates/sui-core/src/consensus_handler.rs
akichidis: Probably that comment is irrelevant now

akichidis (approved):

PR #26279, opened 2026-04-17
consensus_adapter: remove staggered submission and other unused logic

mwtian:
## Summary

Clean up `ConsensusAdapter` and related code. ~920 lines removed, no behavior change on any live path.

- Remove the submit-delay / position-shuffle machinery in `ConsensusAdapter` (helper chain, orphaned fields, `CheckConnection` / `ConnectionMonitorStatus`, the no-longer-read `low_scoring_authorities` map and `latency_observer`, and the `new_mysticeti_certificate` constructor). All of it was dead given `CertifiedTransaction` is never constructed and rejected at `consensus_validator`.
- Drop the config knobs that fed it: `ConsensusConfig::max_submit_position`, `submit_delay_step_override_millis`, and the matching builder surface across `sui-swarm`, `sui-swarm-config`, `test-cluster`, plus the now-inert simtest calls.
- Treat `UserTransaction` (v1) symmetrically to `CertifiedTransaction`: never constructed, rejected at the validator; drop active-logic pattern matches. Variants stay in the enum for bincode ordinal compat.

### Metric changes

- Six metrics removed: `sequencing_certificate_authority_position`, `_positions_moved`, `_preceding_disconnected`, `_amplification_factor`, `sequencing_estimated_latency`, `sequencing_resubmission_interval_ms`.
- `sequencing_certificate_latency`: first label renamed `position` → `submitted` (values `submitted` / `skipped`).
- `classify()` no longer emits `shared_certificate` / `owned_certificate` (collapsed to `_deprecated_certificate`); `UserTransaction` (v1) → `_deprecated_user_transaction`.

## Test plan

- [x] `cargo check --all-targets` on `sui-core`, `sui-node`, `sui-types`, `sui-config`, `sui-swarm-config`, `sui-single-node-benchmark`.
- [x] `cargo clippy --all-targets` — clean.
- [x] `cargo nextest run --lib -p sui-core` for `consensus_adapter`, `consensus_validator`, `consensus_handler`, `scoring_decision` — 29/29 pass.
- [x] `cargo nextest run -p sui-config -p sui-swarm-config` — 25/25 pass (network-config snapshot updated).
- [x] CI: full `cargo simtest`.


crates/sui-core/src/scoring_decision.rs
```diff
@@ -11,39 +10,33 @@ use tracing::debug;
...
+/// metrics and a count gauge.
+/// The list of lower-scoring authorities are returned for testing.
+pub(crate) fn update_low_scoring_authorities_metrics(
     sui_committee: &Committee,
     consensus_committee: &ConsensusCommittee,
     reputation_score_sorted_desc: Option<Vec<(AuthorityIndex, u64)>>,
     metrics: &Arc<AuthorityMetrics>,
     consensus_bad_nodes_stake_threshold: u64,
-) {
+) -> HashMap<AuthorityName, u64> {
```
akichidis: Can't consensus emit those metrics from now on?

akichidis (approved):

crates/sui-core/src/scoring_decision.rs
mwtian: In a future PR, we can definitely try migrating consensus to emit these metrics and not output scores anymore.

crates/sui-config/src/node.rs
```diff
@@ -955,15 +955,6 @@ pub struct ConsensusConfig {
     /// Default to 20_000 inflight limit, assuming 20_000 txn tps * 1 sec consensus latency.
     pub max_pending_transactions: Option<usize>,
 
-    /// When defined caps the calculated submission position to the max_submit_position. Even if the
-    /// is elected to submit from a higher position than this, it will "reset" to the max_submit_position.
-    pub max_submit_position: Option<usize>,
```
aschran, line 960: do we care that removing these params entirely (vs deprecating) could break existing node configs?

crates/sui-types/src/messages_consensus.rs
```diff
@@ -470,15 +470,13 @@ pub enum ConsensusTransactionKind {
 impl ConsensusTransactionKind {
     pub fn as_user_transaction(&self) -> Option<&Transaction> {
         match self {
-            ConsensusTransactionKind::UserTransaction(tx) => Some(tx),
```
aschran, line 473: should these enum variants themselves be marked deprecated?

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -620,42 +272,22 @@ impl ConsensusAdapter {
...
-                first_kind,
-                ConsensusTransactionKind::CertifiedTransaction(_)
+                transactions[0].kind,
+                ConsensusTransactionKind::UserTransactionV2(_)
             );
-
+            if !is_user_tx_batch {
+                return Err(SuiErrorKind::InvalidTxKindInSoftBundle.into());
+            }
             for transaction in &transactions[1..] {
```
aschran: any reason to still be checking index 0 separately from the rest?

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -869,28 +484,24 @@ impl ConsensusAdapter {
...
             debug!("Submitting {:?} to consensus", transaction_keys);
+            guard.submitted = true;
+
+            // The full submission future: acquire the submit semaphore permit
+            // then run the inner submit+status loop. Holding the semaphore
+            // acquire inside this future lets the race below cancel a
+            // submission while it is still queued behind the inflight limit.
+            let submit_fut = async {
+                let _permit: SemaphorePermit = self
+                    .submit_semaphore
```
aschran: not entirely sure I understand the subtlety of including the semaphore in this future vs what was done before? is this PR supposed to just be mechanical changes (removing dead code) or is there more to it than that?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2747,8 +2734,7 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
                     // TODO(fastpath): Add metrics for rejected transactions.
                     if matches!(
                         parsed.transaction.kind,
-                        ConsensusTransactionKind::UserTransaction(_)
-                            | ConsensusTransactionKind::UserTransactionV2(_)
+                        ConsensusTransactionKind::UserTransactionV2(_)
```
aschran: i see this `matches!` pattern all over the place, can these all now be `is_user_transaction` instead?

crates/sui-config/src/node.rs
mwtian: IIRC parsing yaml still works with unknown fields, which has been verified in unit tests.

crates/sui-types/src/messages_consensus.rs
mwtian: It's a good idea. But no deprecated variant in ConsensusTransactionKind has been marked this way yet. So we probably want to do this in a future PR to avoid opening a can of worms here.

crates/sui-core/src/scoring_decision.rs
mwtian: This function, related fields and metrics are removed now.

crates/sui-core/src/consensus_adapter.rs
mwtian: Since we are removing the delay function, to keep the resulting code clean, it is necessary to split checking processed via consensus vs checkpoint into the sync and async parts. Not allowing the wait to acquire inflight semaphore to be cancelled seems like an oversight before, so the semaphore was moved. I can revert this and make the removal a separate change.

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -590,11 +577,7 @@ impl ConsensusAdapter {
         }
         debug!("{transaction_keys:?} processed by consensus");
 
-        let is_user_tx = is_soft_bundle
```
aschran, line 973: intentional that you lost the `is_soft_bundle ||`?

mwtian: Yes this condition has never been necessary. I can add a comment and rename the variable to make it clear.

aschran (approved):

PR #26282, opened 2026-04-17
[move][move-compiler] Revise warning filter system, add #[expect], #[deny], #[warn] support

cgswords:
## Description 

Replaces the WarningFilters/WarningFiltersBuilder/WarningFiltersTable system
with a FilterScope/FilterStack design, which does less interning in order to
track location information for messages and error reporting in a number of
cases.

Also adds #[expect(...)] and #[deny(...)] diagnostic attributes, plus support for them:

- #[expect(...)]: suppresses a warning but emits W10009 if unfulfilled
- #[deny(...)]: upgrades warning severity to error with secondary label

Key changes:

- FilterScope: immutable override set behind a leaked &'static Copy handle
- FilterStack: simple Vec<FilterScope> replacing the old linked-list scope
- Filtering now works by taking the diagnostic and returning a FilterResult.
- Overrides are now accumulated via an `Override` type (`Spanned<Override_>`),
  instead of the filter builder.
- `finalize()` called per-item in to_bytecode for unfulfilled expect checking
- Wildcard filters for `expect` are rejected at parse time
- Parser / attribute handling cleanup (abstraction for code reuse)

## Test plan 

New tests, plus old tests retain their behavior.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
```diff
@@ -0,0 +1,528 @@
+// Copyright (c) Mysten Labs, Inc.
```
cgswords, line 1: I couldn't convince github that this was a file move 😢

external-crates/move/crates/move-compiler/tests/move_2024/ide_mode/control_exp_autocomplete.snap
```diff
@@ -9,7 +9,7 @@ warning[W10007]: issue with attribute value
   ┌─ tests/move_2024/ide_mode/control_exp_autocomplete.move:3:9
   │
 3 │ #[allow(ide_path_autocomplete)]
-  │         ^^^^^^^^^^^^^^^^^^^^^ Unknown warning filter 'ide_path_autocomplete'
+  │         ^^^^^^^^^^^^^^^^^^^^^ Unknown diagnostic filter 'ide_path_autocomplete'
```
cgswords, line 12: I am not attached to this name change; the error just made more sense to me this way. Happy to revert

external-crates/move/crates/move-compiler/src/parser/ast.rs
```diff
@@ -216,6 +216,12 @@ pub enum Attribute_ {
     Allow {
         allow_set: BTreeSet<(Option<Name>, Name)>,
     },
+    Deny {
```
damirka, line 219: Should we have `Warn` here as well?

cgswords: What does `Warn` do?

damirka: Warns? Emits a warning if a global / build config says allow.

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
```diff
@@ -0,0 +1,617 @@
...
+    /// Suppress the diagnostic but retain it in `filtered_source_diagnostics`.
+    Allow,
+    /// Emit the diagnostic as a warning (the default level). Overrides `Allow`/`Deny`/`Expect`
+    /// from an outer scope, restoring normal warning behavior.
+    Warn,
+    /// Upgrade the diagnostic's severity to `NonblockingError`.
+    Deny,
+    /// Suppress the diagnostic and mark this filter entry as fulfilled. Unfulfilled `Expect`
+    /// entries emit an "unfulfilled expect" diagnostic at [`FilterScope::finalize`] time.
+    Expect,
```
tnowacki, line 88: Is this in the right position? And it should be overridden for sure by a deny... I think?

like 
```
#[expect(w)]
...
#[deny(w)]
```
that inner deny should trigger, no?

tnowacki: To me, `expect` seems like a flag on `allow` more than anything

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
```diff
@@ -0,0 +1,617 @@
...
+}
+
+/// What a filter entry matches against. Variants are ordered by decreasing specificity;
+/// during resolution the most specific match wins within a scope.
+///
+/// This type is internal to the filter module. External code uses [`DiagnosticsID`] (with
+/// wildcard sentinels); conversion happens via `From<DiagnosticsID>` at the filter boundary.
+#[derive(Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
+enum FilterTarget {
+    /// Matches a single diagnostic by its exact code ID.
```
tnowacki:
```suggestion
    /// Matches a single diagnostic by its exact code ID (which contains an EernalPrefix)
```
Or something like that somewhere just it caught me off guard at first

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
```diff
@@ -0,0 +1,617 @@
...
+struct ExpectState {
+    loc: Loc,
+    target: FilterTarget,
+    fulfilled: AtomicBool,
+}
+
+/// An immutable set of entries plus expect-fulfillment side-car.
+#[derive(Debug)]
+pub(crate) struct FilterScopeData {
+    filter_entries: BTreeMap<FilterTarget, FilterEntry>,
```
tnowacki: Why not just `BTreeMap<FilterTarget, FilterKind>`?

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
```diff
@@ -0,0 +1,617 @@
...
+    };
+    FilterScope(Arc::new(FilterScopeData {
+        filter_entries: BTreeMap::from([(target, sp(Loc::invalid(), o))]),
+        expects: vec![],
+    }))
+});
+
+static UNUSED_FOR_TEST_FILTER_SCOPE: LazyLock<FilterScope> = LazyLock::new(|| {
+    let cat = Category::UnusedItem as u8;
+    let codes = [
```
tnowacki: nit, maybe this would be nice as constant outside... thought its fine

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
```diff
@@ -0,0 +1,617 @@
...
+    LazyLock::new(|| {
+        let cat_unused = Category::UnusedItem as u8;
+        macro_rules! code {
+            ($cat:ident :: $code:ident) => {
+                DiagnosticsID::exact(None, Category::$cat as u8, $cat::$code as u8)
+            };
+        }
+        // Deliberate leak: these slices live for the process lifetime (behind LazyLock)
+        // and are referenced by every FilterScope. The total size is small and bounded.
+        fn leak(v: Vec<DiagnosticsID>) -> KnownFilterExpansion {
```
tnowacki, line 288: Is this just space saving? Or do you need the `Box<[]>` for something else?

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
```diff
@@ -0,0 +1,617 @@
...
+                FilterTarget::Prefix(prefix) => DiagnosticsID::all(prefix),
+                FilterTarget::AllForDependency => DiagnosticsID::all(None),
+            };
+            (id, sp(*loc, o.kind))
+        })
+    }
+
+    /// Emit diagnostics for every `#[expect(...)]` entry that was never matched.
+    /// NB: `Copy` could ostensibly let you call this multiple times, but that would be a misuse
+    /// of the API and also hopefully deduped by the diagnostic framework.
```
tnowacki: Does this mean `Clone`? 

And if I am understanding it correctly, could we forcibly set `fulfilled` to `true` to prevent any double reporting? Thought that feels like just trading the way in which we are brittle

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
```diff
@@ -0,0 +1,617 @@
...
+    /// Resolve a diagnostic against the active scope stack.
+    ///
+    /// Warnings that pass through get an `#[allow(...)]` hint note when a known filter
+    /// name exists. `Deny` upgrades severity to `NonblockingError`.
+    pub fn filter(
+        &self,
+        mut diag: Diagnostic,
+        known_filter_names: &BTreeMap<DiagnosticsID, (FilterPrefix, FilterName)>,
+    ) -> FilterResult {
+        let key = diag.info().id();
```
tnowacki: Should we early return if the diag is already an error?

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
```diff
@@ -0,0 +1,617 @@
...
+            FilterKind::Deny => {
+                diag = diag.set_severity(Severity::NonblockingError);
+                if resolved.loc != Loc::invalid() {
+                    diag.add_secondary_label((resolved.loc, "the lint level is defined here"));
+                }
+                FilterResult::Emit(diag)
+            }
+            FilterKind::Expect => {
+                mark_expect_fulfilled(resolved.scope, resolved.target);
+                FilterResult::Discarded
```
tnowacki: Should this be Filtered?

damirka (approved):
LGTM!

external-crates/move/crates/move-compiler/src/expansion/translate.rs
```diff
@@ -1651,39 +1647,121 @@ fn warning_filter_(context: &Context, attributes: &E::Attributes) -> WarningFilt
...
+        };
+        for id in filters {
+            use std::collections::btree_map::Entry;
+            match filter_entries.entry(id) {
+                Entry::Vacant(e) => {
+                    e.insert(sp(name_loc, filter_kind));
+                }
+                Entry::Occupied(e) if e.get().value == filter_kind => {
+                    // Same kind, no conflict — keep the existing entry.
+                }
```
tnowacki, line 1745: Does this mean we allow `[allow(unused), allow(unused)]`?

external-crates/move/crates/move-compiler/tests/linter/nested_expect_lint_inner_warn_lint.move
```diff
@@ -0,0 +1,7 @@
+// Test that an inner #[warn(lint(...))] overrides an outer #[expect(lint(...))],
+// causing the lint to fire as a warning and the outer expect to be unfulfilled.
+#[expect(lint(constant_naming))]
+module 0x42::m {
+    #[warn(lint(constant_naming))]
+    const Another_BadName: u64 = 42;
```
tnowacki, line 6: I think this is consistent with Rust, but I find this so fuckin strange lol
Like it says "I expect a warning" and the inner says "please warn". So I would expect the "expect" to be satisfied by the warning

external-crates/move/crates/move-compiler/tests/move_check/warning_suppression/conflict_warn_allow.move
```diff
@@ -0,0 +1,8 @@
+// Test: #[warn] and #[allow] for the same code on the same item conflicts.
+#[warn(unused_variable)]
+#[allow(unused_variable)]
```
tnowacki, line 3: Do we have a test for a double `#[allow(unused_variable)]`

external-crates/move/crates/move-compiler/tests/move_check/warning_suppression/expect_all.snap
```diff
@@ -0,0 +1,34 @@
...
+info:
+  flavor: core
+  edition: legacy
+  lint: false
+---
+warning[W10007]: issue with attribute value
+  ┌─ tests/move_check/warning_suppression/expect_all.move:3:10
+  │
+3 │ #[expect(all)]
+  │          ^^^ Wildcard filters are not allowed in '#[expect(...)]'. Use '#[allow(all)]' instead
```
tnowacki: I'm not sure I love the term "wildcard" filters here. I might just say something like "Only specific filters are allowed in expect"

external-crates/move/crates/move-compiler/tests/move_check/warning_suppression/expect_all.snap
```diff
@@ -0,0 +1,34 @@
...
+4 │ module 0x42::m { fun var(a: u64) { } }
+  │                          ^ Unused parameter 'a'. Consider removing or prefixing with an underscore: '_a'
+  │
+  = This warning can be suppressed with '#[allow(unused_variable)]' applied to the 'module' or module member ('const', 'fun', or 'struct')
+
+warning[W10007]: issue with attribute value
+  ┌─ tests/move_check/warning_suppression/expect_all.move:7:10
+  │
+7 │ #[expect(unused)]
+  │          ^^^^^^ Wildcard filters are not allowed in '#[expect(...)]'. Use '#[allow(unused)]' instead
```
tnowacki: Similarly I find the term wildcard a bit strange here. Maybe "category" or "category based"?

external-crates/move/crates/move-compiler/tests/move_check/warning_suppression/expect_wildcard_prefixed.snap
```diff
@@ -0,0 +1,12 @@
...
+info:
+  flavor: core
+  edition: legacy
+  lint: false
+---
+warning[W10007]: issue with attribute value
+  ┌─ tests/move_check/warning_suppression/expect_wildcard_prefixed.move:2:15
+  │
+2 │ #[expect(lint(all))]
+  │               ^^^ Unknown diagnostic filter 'lint(all)'
```
tnowacki: Is this intentional? Seems like the wrong kind of rejection?

external-crates/move/crates/move-compiler/tests/move_check/warning_suppression/warn_shadows_deny.move
```diff
@@ -0,0 +1,8 @@
+// Test: #[deny] on module upgrades to error, #[warn] on function restores to warning.
+#[deny(unused_variable)]
+module 0x42::m {
+    #[warn(unused_variable)]
```
tnowacki, line 4: Hmmmm. Maybe this is why expect works the way it does?
How does this case work with allow?

tnowacki: Maybe this whole precedence thing is fucked. Maybe you just always take the most recent attribute?

external-crates/move/crates/move-compiler/tests/move_check/warning_suppression/conflict_warn_allow.move
tnowacki: NVM found it

tnowacki (changes requested):
Requesting changes mostly so this doesn't accidentally get automerged before discussing things here. 
Looks mostly good structurally, just not sure exactly about how this whole precedence thing shakes out

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
cgswords: As discussed, this is just a best-effort for when things collide in the same scope. That said, I agree that `deny` should override, so I reordered them.

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
cgswords: Completely for space-saving, to literally make the code a touch shorter.

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
cgswords: This comment is out of date; it was `Copy` at one point when the representation was a bit more-compact, but I relaxed it back to Clone when I switched it to Arcs and this comment was simply missed.

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
cgswords: Moved

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
cgswords: My thinking was: if we ever wanted something that would be an error able to be converted back to a warning, we might want this carve-out. Easy enough to swap back in that case, though. Moved up to the top, even before that.

external-crates/move/crates/move-compiler/tests/linter/nested_expect_lint_inner_warn_lint.move
cgswords: I think it makes sense when you view it as a stack, but we could be far pickier about layering these things in a follow-up PR if you would like.

external-crates/move/crates/move-compiler/src/expansion/translate.rs
cgswords: Test and its output are below, as you pointed out.

external-crates/move/crates/move-compiler/tests/move_check/warning_suppression/warn_shadows_deny.move
cgswords: Yes, that is what we do today. It's what Rust does as well. Happy to 🚲 🏠  further if you want.

external-crates/move/crates/move-compiler/src/diagnostics/filter.rs
cgswords: We can, and it lets us get rid of `FilterEntry` altogether.

external-crates/move/crates/move-compiler/tests/move_check/warning_suppression/expect_all.snap
```diff
@@ -0,0 +1,38 @@
...
+4 │ module 0x42::m { fun var(a: u64) { } }
+  │                          ^ Unused parameter 'a'. Consider removing or prefixing with an underscore: '_a'
+  │
+  = This warning can be suppressed with '#[allow(unused_variable)]' applied to the 'module' or module member ('const', 'fun', or 'struct')
+
+warning[W10007]: issue with attribute value
+  ┌─ tests/move_check/warning_suppression/expect_all.move:7:10
+  │
+7 │ #[expect(unused)]
+  │          ^^^^^^ 'unused' matches multiple diagnostics, but 'expect' may only be used for individual diagnostics
```
tnowacki, line 28: 👍 looks great!

tnowacki (approved):

PR #26284, opened 2026-04-17
ci: build release Linux binaries in rust:1.92-bullseye (glibc 2.31 floor)

ebmifa:
## Summary
Build Linux release binaries inside `rust:1.92-bullseye` instead of on the runner host, pinning the glibc floor at 2.31. This aligns `release.yml` with the convention already established by every Dockerfile in this repo.

## Why
The `ubuntu-ghcloud` Larger Runner was recently migrated from Ubuntu 22.04 → 24.04. Host glibc jumped from 2.35 to 2.39. Binaries built directly on the runner now link against `memcpy@GLIBC_2.38` / `fcntl@GLIBC_2.39` etc. — symbol versions that don't exist on older deployment hosts. Internal reports already confirmed sui-node failing to load with:

```
/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found
/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found
```

## Approach: mirror the existing Dockerfile convention
Every Dockerfile in the repo (sui-node, sui-tool, sui-indexer-alt-*, sui-services, stress, sui-proxy, sui-kvstore, etc.) already uses `FROM rust:1.92-bullseye AS builder`. The glibc-2.31 floor is the established internal standard for docker-shipped infra. **This PR applies the same standard to release.yml** — no new policy, just closing a gap.

Why build inside a container rather than changing the runner image:
- **Decouples the binary's ABI floor from the runner OS.** Future runner upgrades (24.04 → 26.04 → …) won't affect released binaries.
- **Aligns with the "manylinux" pattern** used by PyPI wheel builders, Rust's `cross` tool, etc.
- **Keeps the runner upgrade intact** — the ubuntu-ghcloud migration to 24.04 was the right call; this PR just fixes the side effect on release binaries.

## Scope
- `release.yml` only. Other workflows (`rust.yml`, `nightly.yml`, etc.) test on the runner and don't ship binaries, so they stay on the host.
- Linux matrix entries (`ubuntu-ghcloud`, `ubuntu-arm64`) only; Windows/macOS unchanged (no glibc concern).
- sccache skipped on Linux release builds — host sccache isn't inside the container, and warmup doesn't populate bullseye-container cache keys. Release runs per-tag, not PR-CI-critical; the slower cold build is an acceptable tradeoff.

## Out of scope (flagged for separate work)
- **sui-operations**: validator docker images are produced there; platform team is handling separately.
- **sccache-warmup.yml**: if we later want release builds cached, warmup would need to build in the same container too. Deferring until someone feels the cold-build pain.
- **No Dockerfile changes**: every Dockerfile already uses `rust:1.92-bullseye` — no drift to fix.

## Safety net: glibc-floor verification
Added a post-build check that runs `objdump -T target/release/sui | grep GLIBC | sort -V | tail -1` and fails the job if the highest-required symbol version exceeds 2.31. Catches future regressions (e.g., someone accidentally building on host instead of in-container).

## Test plan
- [ ] `workflow_dispatch` on this branch against a throwaway tag. Verify Linux job completes successfully and `Verify glibc floor` step prints `OK: within glibc 2.31 floor`
- [ ] Download the produced `sui-*-linux-x86_64.tgz` artifact, extract, run `objdump -T sui-node | grep GLIBC | sort -V | tail -1` → expect `GLIBC_2.31` or lower
- [ ] Run the extracted `sui-node --version` on an Ubuntu 20.04 / Debian 11 host to confirm backward compat
- [ ] Windows/macOS jobs unaffected (unchanged build path)
- [ ] Once merged, recut any release tags cut between the ubuntu-ghcloud 24.04 flip and this PR landing

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -162,25 +162,81 @@ jobs:
           sudo rm -rf ~/Library/Developer/Xcode/iOS\ DeviceSupport/*
           df -h /
 
-      - name: Setup sccache
-        if: ${{ env.s3_existing_archive == '' }}
+      # sccache on Windows/macOS only. Linux release builds run inside a
+      # rust:1.92-bullseye container (see the "Cargo build (Linux...)" step below)
+      # to pin the glibc floor at 2.31 — matching the convention used by every
+      # Dockerfile in this repo. Host-installed sccache isn't available inside
+      # that container, and warmup doesn't populate bullseye-container-keyed
+      # entries, so sccache is skipped on Linux for release builds.
+      - name: Setup sccache (non-Linux)
+        if: ${{ runner.os != 'Linux' && env.s3_existing_archive == '' }}
         id: sccache
         continue-on-error: true
         uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # pin@v0.0.9
 
-      - name: Enable sccache
-        if: steps.sccache.outcome == 'success'
+      - name: Enable sccache (non-Linux)
+        if: ${{ runner.os != 'Linux' && steps.sccache.outcome == 'success' }}
         shell: bash
         run: echo "RUSTC_WRAPPER=sccache" >> "$GITHUB_ENV"
 
       - name: Register cargo problem matcher
         run: echo "::add-matcher::.github/matchers/cargo.json"
 
-      - name: Cargo build for ${{ matrix.os }} platform
-        if: ${{ env.s3_existing_archive == '' }}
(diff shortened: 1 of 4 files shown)


ebmifa:
Closing: team decided to roll back the ubuntu-ghcloud runner to 22.04 rather than keep 24.04 with container-based release builds. Release binaries built on a 22.04 host will match the glibc floor of deployment hosts natively, so this container indirection isn't needed. See team discussion for context.

PR #26292, opened 2026-04-17
[forking 9/n] add transaction executor service

stefan-mysten:
## Description 

This PR implements the transaction executor service for gRPC. For now, only `execute_transaction` is implemented, and `simulate_transaction` is left for a future PR.

## Test plan 

Added one unit test that executes a transaction through the gRPC interface.
```
cargo nextest run -- test_tx_execution
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-forking/src/rpc/executor.rs
```diff
@@ -0,0 +1,138 @@
...
+                finality_info: EffectsFinalityInfo::QuorumExecuted(effects.executed_epoch()),
+            },
+            events,
+            input_objects,
+            output_objects,
+            auxiliary_data: None,
+        })
+    }
+
+    fn simulate_transaction(
```
stefan-mysten, line 127: `simulacrum` does not yet support simulation. There's a separate PR to add it: https://github.com/MystenLabs/sui/pull/26321. Once that's in, we can easily plug it.

crates/sui-forking/src/rpc/executor.rs
```diff
@@ -0,0 +1,141 @@
...
+        let (effects, exec_error, checkpoint_seq) = {
+            let mut sim = self.context.simulacrum().write().await;
+            let (effects, exec_error) =
+                sim.execute_transaction(request.transaction).map_err(|e| {
+                    TransactionSubmissionError::TransactionDriverInternalError(SuiError::from(
+                        format!("forked execution failed: {e}"),
+                    ))
+                })?;
+            let checkpoint = sim.create_checkpoint();
+            let checkpoint_seq = checkpoint.data().sequence_number;
```
stefan-mysten, line 63: We should store this on disk.

stefan-mysten: Note to self: SimulatorStore saves the data during `create_checkpoint`.

crates/sui-forking/src/rpc/executor.rs
```diff
@@ -0,0 +1,141 @@
...
+        // Execute under the simulacrum write lock, then seal the transaction
+        // into a fresh checkpoint so downstream reads see it as finalized.
+        // The lock is dropped before reading back events/objects so
+        // concurrent reads aren't blocked.
+        let (effects, exec_error, checkpoint_seq) = {
+            let mut sim = self.context.simulacrum().write().await;
+            let (effects, exec_error) =
+                sim.execute_transaction(request.transaction).map_err(|e| {
+                    TransactionSubmissionError::TransactionDriverInternalError(SuiError::from(
+                        format!("forked execution failed: {e}"),
```
jordanjennings-mysten: is there any variant of error where you would want to bubble up the real error so the behavior is the same as a real node? (outside of execution errors)

I think no but not 100% sure.

crates/sui-forking/src/rpc/executor.rs
```diff
@@ -0,0 +1,141 @@
...
+                    TransactionSubmissionError::TransactionDriverInternalError(SuiError::from(
+                        format!("failed to resolve output objects for {digest}: {e}"),
+                    ))
+                })?,
+            )
+        } else {
+            None
+        };
+
+        let executed_epoch = effects.executed_epoch();
```
jordanjennings-mysten, line 111: this has me curious if you need to be careful about the mechanics around epochs since quite a few important things happen on the epoch.

crates/sui-forking/src/tests/rpc_executor.rs
```diff
@@ -0,0 +1,121 @@
...
+use sui_types::transaction::{GasData, Transaction, TransactionData, TransactionKind};
+use sui_types::transaction_driver_types::ExecuteTransactionRequestV3;
+use sui_types::transaction_executor::TransactionExecutor;
+
+use crate::context::Context;
+use crate::rpc::executor::ForkedTransactionExecutor;
+use crate::store::DataStore;
+
+#[tokio::test]
+async fn test_tx_execution() {
```
jordanjennings-mysten, line 138: maybe a few error tests

jordanjennings-mysten (approved):

crates/sui-forking/src/rpc/executor.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        Ok(ExecuteTransactionResponseV3 {
+            effects: FinalizedEffects {
+                effects,
+                // The forked network is single-node with no consensus; we
+                // report the effects as executed within their embedded epoch.
+                finality_info: EffectsFinalityInfo::QuorumExecuted(executed_epoch),
```
stefan-mysten: This is probably wrong. I wonder if the correct thing is to return
`finality_info: EffectsFinalityInfo::Checkpointed(epoch_id, checkpoint_seq)`

crates/sui-forking/src/rpc/executor.rs
stefan-mysten: Curious about the specifics here. Could you share more details? Are you thinking of `EndOfEpochTransaction` type?

crates/sui-forking/src/rpc/executor.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub(crate) fn new(context: Arc<Context>) -> Self {
+        Self { context }
+    }
+}
+
+#[async_trait]
+impl TransactionExecutor for ForkedTransactionExecutor {
+    async fn execute_transaction(
+        &self,
+        request: ExecuteTransactionRequestV3,
```
stefan-mysten, line 50: This is probably wrong too, should be `ExecuteTransactionRequest`.

stefan-mysten: Actually no, the trait specifies `ExecuteTransactonRequestV3`.

crates/sui-forking/src/rpc/executor.rs
stefan-mysten: That's a good point. I think we'd have to get better errors from simulacrum to be able to do that. It currently returns `ExecutionError`. I'd have to play a bit more with it to understand what can be exposed here.

crates/sui-forking/src/tests/rpc_executor.rs
stefan-mysten: Added more tests.

crates/sui-forking/src/rpc/executor.rs
jordanjennings-mysten: I was thinking more for example do we miss out on anything with the lack of epoch changes? I assume the epoch stays static for the life of the fork but would that cause issues with the fork? protocol changes are an interesting example if in the next epoch an important protocol change were to happen would that cause any problems with your fork to not change to the next epoch at some time.

PR #26303, opened 2026-04-18
[jsonrpc-alt] in-house validators apy calculation

wlmyng:
## Description 

The implementation on fullnode jsonrpc involves getting the latest system state, and for each active validator and inactive pool, getting all historical exchange rates. APY is calculated per adjacent pair, and then a sample is selected for averaging to get the APY. This is a bit wasteful, but understandable since `getValidatorsApy` shares a cache (evicted on epoch change) with `getStakes` and `getStakesById` - and these two apis support historical estimated rewards calculation.

Now that we've split the `getStakes` APIs out, we can simplify the APY calculation. Observing that the exchange rate is calculated from the `staking_pool_sui_balance` and `pool_token_balance`, and this is indexed today on `kv_epoch_starts`, we can instead retrieve the 31 latest entries from the table, extract just the active validator data, and calculate APY accordingly. 

In practice, the calculated APY match to the last 

Without caching, a local setup speaking to mainnet postgres db, takes at most a second, and around 400ms once warm.

| validator | legacy | new |
|---|---|---|
| `0x1e1985024aafe50a8e4eafc5a89eb7ecd58ba08c39f37688bee00bd55c8b2059` | 0.01568213561669605**5** | 0.01568213561669604**8** |
| `0x61953ea72709eed72f4441dd944eec49a11b4acabfc8e04015e89c63be81b6ab` | 0.01527990128315504**8** | 0.01527990128315504**7** |
| `0xc21c40c94c712f17244d8518688dbaaa81ba232c15cb13dbba138ead55451e43` | 0.01538794574356528 | 0.01538794574356528**6** |
| `0xd939e3fe7ea4d503f84767dca0c58b7ec1c71f085638a4c0611aa64aa71b5fcf` | 0.01525615104199363**2** | 0.01525615104199363**3** |
| `0x0de5a09a0b029b0c4f1017a58dc78a2341511a7f6ba4925287aa836c1166dbbe` | 0.01552167944976588 | 0.01552167944976587**9** |
| `0xf72816357f8e5638c90ec5b0f7eecee362bac0ef7e64822ea6d686730251a7a0` | 0.01551620642763676 | 0.01551620642763675**9** |
| `0xefc45ee5de49f376b66b73f6bab9fbc0c452b18e1cfc5035c268c6786dd8bb63` | 0.01570598497104809**2** | 0.0157059849710481 |
| `0xb8068ad94d7f0448059ae3ea1d877f2af54792b2849d80e2753201adfa532411` | 0.01541293648769391**2** | 0.01541293648769391**3** |
| `0x62e0917d86c476ed6beb5594bab91fea285671df1716874cc6cbde5c695da64e` | 0.01505770851626768**8** | 0.01505770851626769**5** |
| `0x13d90417a640de83e0b83683f12028961de06195ab004156df80111a9499081b` | 0.01536802734917224**4** | 0.01536802734917224**3** |
| `0x3d9fb148e35ef4d74fcfc36995da14fc504b885d5f2bfeca37d6ea2cc044a32d` | 0.01536601659804334 | 0.01536601659804333**9** |
| `0x7ab4a4f88e3062a8af7c4c43add6e1d6d864f3511ed67c352989c81ff048fccd` | 0.0155494855880644 | 0.01554948558806440**1** |
| `0x111111dccbe54f2eab9a1374cd1763ed08e6cbb4f241f053d8153cc73283ab63` | 0.01534896886092208 | 0.01534896886092208**1** |
| `0x63c99c4c48e9a417038a5f96fa7e3a0dbaacaa56a0d766f39efc8bd6cdcb2196` | 0.01428231136856860**3** | 0.01428231136856860**5** |

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-jsonrpc/src/api/governance.rs
+++ b/crates/sui-indexer-alt-jsonrpc/src/api/governance.rs
@@ -2,26 +2,25 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::collections::BTreeMap;
+use std::collections::HashMap;
 use std::convert::Infallible;
 
 use anyhow::Context as _;
 use diesel::ExpressionMethods;
 use diesel::QueryDsl;
 use futures::future;
-
 use jsonrpsee::core::RpcResult;
-use jsonrpsee::http_client::HttpClient;
 use jsonrpsee::proc_macros::rpc;
 use move_core_types::language_storage::StructTag;
-
 use sui_indexer_alt_reader::consistent_reader::proto::owner::OwnerKind;
 use sui_indexer_alt_reader::governance::RewardsKey;
 use sui_indexer_alt_reader::governance::ValidatorAddressKey;
+use sui_indexer_alt_schema::epochs::StoredEpochStart;
 use sui_indexer_alt_schema::schema::kv_epoch_starts;
-use sui_json_rpc_api::GovernanceReadApiClient;
 use sui_json_rpc_types::DelegatedStake;
 use sui_json_rpc_types::Stake;
 use sui_json_rpc_types::StakeStatus;
+use sui_json_rpc_types::ValidatorApy;
 use sui_json_rpc_types::ValidatorApys;
 use sui_open_rpc::Module;
 use sui_open_rpc_macros::open_rpc;
@@ -36,6 +35,8 @@ use sui_types::governance::STAKED_SUI_STRUCT_NAME;
 use sui_types::governance::STAKING_POOL_MODULE_NAME;
 use sui_types::governance::StakedSui;
 use sui_types::sui_serde::BigInt;
+use sui_types::sui_system_state::PoolTokenExchangeRate;
+use sui_types::sui_system_state::SuiSystemState;
 use sui_types::sui_system_state::SuiSystemStateTrait;
 use sui_types::sui_system_state::SuiSystemStateWrapper;
 use sui_types::sui_system_state::sui_system_state_inner_v1::SuiSystemStateInnerV1;
@@ -49,8 +50,11 @@ use crate::data::latest_epoch;
 use crate::data::load_live;
 use crate::data::load_live_deserialized;
 use crate::error::RpcError;
-use crate::error::client_error_to_error_object;
 use crate::error::rpc_bail;
+
+/// Number of most recent epochs to load from `kv_epoch_starts` when computing validator APYs.
+/// Matches the upper bound on adjacent-epoch samples used by `compute_apy`.
+const APY_EPOCH_WINDOW: i64 = 31;
 
 #[open_rpc(namespace = "suix", tag = "Governance API")]
 #[rpc(server, namespace = "suix")]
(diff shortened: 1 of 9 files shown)


crates/sui-indexer-alt-jsonrpc/src/api/governance.rs
```diff
@@ -339,3 +312,100 @@ async fn delegated_stakes_response(
...
+///
+/// Rates are derived from `staking_pool_sui_balance` and `pool_token_balance` of each active
+/// validator in the latest `APY_EPOCH_WINDOW` rows of `kv_epoch_starts`. These are the same numbers
+/// that `advance_epoch` writes into each staking pool's `exchange_rates` table. APYs are calculated
+/// from adjacent pairs of rates, and then filtered and averaged to produce each validator's APY.
+/// This mirrors the legacy fullnode jsonrpc's `backfill_rates` implementation.
+///
+/// In safe mode, a pool's sui and token balance carry over unchanged, which produces 0% APY and
+/// also gets filtered out.
+async fn validators_apy_response(ctx: &Context) -> Result<ValidatorApys, RpcError> {
```
amnn, line 355: You could consider putting a cache in `ctx` for this information, so that multiple requests for this data don't end up doing wasted work? It's going to be the same numbers for the rest of the day right?

amnn: An `LruCache` with capacity of 1, keyed by `epoch` would do the trick.

amnn (approved):
Glorious to see delegation disappear. Left a suggestion to add caching for this, it can be a follow up though.

crates/sui-indexer-alt-jsonrpc/src/api/governance.rs
amnn: Actually it doesn't need to go in the context -- it could be initialised on the type that implements `GovernanceApiServer`.

wlmyng: went ahead and did it

crates/sui-indexer-alt-jsonrpc/src/api/governance.rs
```diff
@@ -150,35 +158,29 @@ impl GovernanceApiServer for Governance {
...
-    }
-}
+            .context("Failed to fetch latest epoch for APY cache lookup")
+            .map_err(RpcError::<Infallible>::from)?;
 
-impl RpcModule for Governance {
-    fn schema(&self) -> Module {
-        GovernanceApiOpenRpc::module_doc()
-    }
+        if let Some(hit) = self.apy_cache.lock().unwrap().get(&epoch).cloned() {
```
amnn, line 171: nit: would `RwLock` be a better fit here? Assuming this is a relatively common query, reads should be way more common than writes.

PR #26321, opened 2026-04-20
[core/simulacrum] refactor simulate transaction logic to be more easily reused

stefan-mysten:
## Description 

This PR refactors the `simulate_transaction` API in `AuthorityState` to make it reusable in simulacrum. Initially, I duplicated most code for simulacrum but it's easy for them to diverge, thus this refactoring allows simulacrum to use the same code path, sans the `AuthorityState` stuff.

The main split is
- have a type for various parameters that we need to provide for simulation. This way, we can create that from the data we have (simulacrum, fullnode, etc)
- validate the inputs
- prepare the simulation
- execute the simulation
- build the output data

## Test plan 

Existing tests and added new test in `sui-core`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/simulacrum/src/epoch_state.rs
+++ b/crates/simulacrum/src/epoch_state.rs
@@ -1,207 +1,30 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::HashSet;
 use std::sync::Arc;
 
 use anyhow::Result;
-use move_core_types::language_storage::TypeTag;
-use once_cell::sync::Lazy;
 use sui_config::{
     transaction_deny_config::TransactionDenyConfig, verifier_signing_config::VerifierSigningConfig,
-};
-use sui_core::accumulators::funds_read::AccountFundsRead;
-use sui_core::simulation::{
-    SimulationContext, SimulationParams, simulate_transaction_with_context,
 };
 use sui_execution::Executor;
 use sui_protocol_config::{Chain, ProtocolConfig, ProtocolVersion};
 use sui_types::{
-    SUI_ACCUMULATOR_ROOT_OBJECT_ID,
-    accumulator_root::{AccumulatorKey, AccumulatorObjId, AccumulatorValue},
-    base_types::{ObjectID, ObjectRef, SequenceNumber, SuiAddress},
-    coin_reservation::{CoinReservationResolverTrait, ParsedObjectRefWithdrawal},
     committee::{Committee, EpochId},
-    digests::{ChainIdentifier, TransactionDigest},
+    digests::ChainIdentifier,
     effects::TransactionEffects,
-    error::{SuiResult, UserInputError, UserInputResult},
     execution_params::ExecutionOrEarlyError,
     gas::SuiGasStatus,
     inner_temporary_store::InnerTemporaryStore,
     metrics::{BytecodeVerifierMetrics, ExecutionMetrics},
-    storage::BackingPackageStore,
     sui_system_state::{
         SuiSystemState, SuiSystemStateTrait,
         epoch_start_sui_system_state::{EpochStartSystemState, EpochStartSystemStateTrait},
     },
-    transaction::{
-        FundsWithdrawalArg, InputObjectKind, InputObjects, ReceivingObjects, TransactionData,
-        TransactionDataAPI, VerifiedTransaction,
-    },
-    transaction_executor::{SimulateTransactionResult, TransactionChecks},
+    transaction::{TransactionDataAPI, VerifiedTransaction},
 };
 
 use crate::SimulatorStore;
-
-static EMPTY_CERTIFICATE_DENY_SET: Lazy<HashSet<TransactionDigest>> = Lazy::new(HashSet::new);
-
-fn invalid_withdraw_reservation(error: impl Into<String>) -> UserInputError {
-    UserInputError::InvalidWithdrawReservation {
-        error: error.into(),
-    }
-}
-
-struct SimulatorCoinReservationResolver<'a> {
-    store: &'a (dyn SimulatorStore + Send + Sync),
-}
-
-impl SimulatorCoinReservationResolver<'_> {
-    fn get_owner_and_type_for_object(
-        &self,
-        object_id: ObjectID,
-        accumulator_version: Option<SequenceNumber>,
-    ) -> UserInputResult<(SuiAddress, TypeTag)> {
-        let object =
-            AccumulatorValue::load_object_by_id(self.store, accumulator_version, object_id)
-                .map_err(|e| {
-                    invalid_withdraw_reservation(format!(
-                        "could not load coin reservation object id {e}"
-                    ))
-                })?
-                .ok_or_else(|| {
-                    invalid_withdraw_reservation(format!(
-                        "coin reservation object id {object_id} not found"
-                    ))
-                })?;
-
-        let move_object = object.data.try_as_move().ok_or_else(|| {
-            invalid_withdraw_reservation(format!(
-                "coin reservation object id {object_id} is not a move object"
-            ))
-        })?;
-        let type_tag = move_object
-            .type_()
-            .balance_accumulator_field_type_maybe()
-            .ok_or_else(|| {
-                invalid_withdraw_reservation(format!(
-                    "coin reservation object id {object_id} is not a balance accumulator field"
-                ))
-            })?;
-        let (key, _): (AccumulatorKey, AccumulatorValue) = move_object.try_into().map_err(|e| {
-            invalid_withdraw_reservation(format!("could not load coin reservation object id {e}"))
-        })?;
-        Ok((key.owner, type_tag))
-    }
-}
-
-impl CoinReservationResolverTrait for SimulatorCoinReservationResolver<'_> {
-    fn resolve_funds_withdrawal(
-        &self,
-        sender: SuiAddress,
-        coin_reservation: ParsedObjectRefWithdrawal,
-        accumulator_version: Option<SequenceNumber>,
-    ) -> UserInputResult<FundsWithdrawalArg> {
-        let (owner, type_tag) = self.get_owner_and_type_for_object(
-            coin_reservation.unmasked_object_id,
-            accumulator_version,
-        )?;
-
-        if sender != owner {
-            return Err(invalid_withdraw_reservation(format!(
-                "coin reservation object id {} is owned by {}, not sender {}",
-                coin_reservation.unmasked_object_id, owner, sender
-            )));
-        }
-
-        Ok(FundsWithdrawalArg::balance_from_sender(
-            coin_reservation.reservation_amount(),
-            type_tag,
-        ))
-    }
-}
-
-struct SimulatorAccountFundsRead<'a> {
-    store: &'a (dyn SimulatorStore + Send + Sync),
-}
-
-impl AccountFundsRead for SimulatorAccountFundsRead<'_> {
-    fn get_latest_account_amount(&self, id: &AccumulatorObjId) -> (u128, SequenceNumber) {
-        latest_account_amount(self.store, id)
-    }
-
-    fn get_account_amount_at_version(
-        &self,
-        id: &AccumulatorObjId,
-        version: SequenceNumber,
-    ) -> u128 {
-        AccumulatorValue::load_object_by_id(self.store, Some(version), *id.inner())
-            .expect("account-balance reads in simulacrum must succeed")
-            .map(|object| {
-                let move_object = object
-                    .data
-                    .try_as_move()
-                    .expect("accumulator account objects must be move objects");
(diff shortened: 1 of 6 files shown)


crates/simulacrum/src/epoch_state.rs
```diff
@@ -172,4 +410,24 @@ impl EpochState {
             );
         Ok((inner_temp_store, gas_status, effects, result))
     }
+
+    pub fn simulate_transaction(
+        &self,
+        store: &(dyn SimulatorStore + Send + Sync),
+        deny_config: &TransactionDenyConfig,
+        verifier_signing_config: &VerifierSigningConfig,
```
stefan-mysten, line 418: I think we don't want these here, but I'll play more with it and fix it as needed.

crates/sui-forking/src/store.rs
```diff
@@ -543,6 +543,10 @@ impl SimulatorStore for DataStore {
         }
     }
 
+    fn backing_package_store(&self) -> &dyn BackingPackageStore {
+        self
+    }
+
```
stefan-mysten, line 549: We're already implementing BackingPackageStore so this needs to go.

crates/sui-transactional-test-runner/src/simulator_persisted_store.rs
```diff
@@ -387,6 +387,10 @@ impl SimulatorStore for PersistedStore {
         }
     }
 
+    fn backing_package_store(&self) -> &dyn sui_types::storage::BackingPackageStore {
+        self
+    }
+
```
stefan-mysten, line 393: Same as above, not needed.

crates/sui-core/src/authority.rs
```diff
@@ -2473,17 +2567,10 @@ impl AuthorityState {
 
     pub fn simulate_transaction(
```
stefan-mysten, line 2568: This function is refactored, and it's basically now split into 4 parts in `simulation.rs` file.
- validate simulation request
- prepare for execution
- execute
- get output data

crates/sui-core/src/authority.rs
```diff
@@ -965,6 +967,98 @@ pub struct AuthorityState {
     post_processing_semaphore: Arc<tokio::sync::Semaphore>,
 }
 
+struct AuthoritySimulationContext<'a> {
+    authority: &'a AuthorityState,
+    epoch_store: &'a AuthorityPerEpochStore,
+}
+
+impl SimulationContext for AuthoritySimulationContext<'_> {
+    fn protocol_config(&self) -> &sui_protocol_config::ProtocolConfig {
```
amnn: A lot of these functions accept no arguments and I think (?) they are requesting data that simulation would need to fetch unconditionally. Given that, does it make sense to include them in the trait, vs accepting the return type as a parameter to the abstracted simulation function?

stefan-mysten: That's reasonable, will rework this.

crates/sui-core/src/authority.rs
```diff
@@ -1009,36 +1035,18 @@ impl AuthorityState {
         receiving_objects_refs: &[ObjectRef],
         protocol_config: &ProtocolConfig,
     ) -> SuiResult<BTreeMap<AccumulatorObjId, (u64, TypeTag)>> {
-        // Note: the deny checks may do redundant package loads but:
-        // - they only load packages when there is an active package deny map
-        // - the loads are cached anyway
-        sui_transaction_checks::deny::check_transaction_for_signing(
+        crate::simulation::pre_object_load_checks(
```
alex-mysten, line 1038: pre_object_load_checks is used at voting time by the validator to determine if a transaction should be accepted. It feels misleading to delegate to a module called "simulation". These are the real checks used by the validator at signing time, not simulation-only concerns.

crates/sui-core/src/authority.rs
```diff
@@ -2510,245 +2511,36 @@ impl AuthorityState {
...
-        // TODO see if we can spin up a VM once and reuse it
-        let executor = sui_execution::executor(
-            protocol_config,
-            true, // silent
+        simulate_transaction_with_context(
+            &params,
+            &context,
+            transaction,
+            checks,
+            allow_mock_gas_coin,
```
alex-mysten, line 2542: Why did we choose to exclude these from the params struct?

crates/sui-core/src/simulation.rs
```diff
@@ -0,0 +1,526 @@
...
+    mock_gas_id: Option<ObjectID>,
+}
+
+/// Run transaction simulation using the provided context.
+///
+/// # Errors
+///
+/// Returns an error if the transaction is unsupported for simulation, fails input validation, or
+/// fails any shared pre-execution checks.
+pub fn simulate_transaction_with_context<C: SimulationContext>(
```
alex-mysten, line 116: nit - `with_context` suffix implies there exists some alternative method of simulation. can we drop it?

crates/sui-core/src/simulation.rs
```diff
@@ -0,0 +1,526 @@
...
+    ctx: &C,
+    transaction: TransactionData,
+    checks: TransactionChecks,
+    allow_mock_gas_coin: bool,
+) -> SuiResult<SimulateTransactionResult> {
+    validate_simulation_request(params, &transaction)?;
+    let prepared = prepare_simulation(params, ctx, transaction, checks, allow_mock_gas_coin)?;
+    let executed = execute_prepared_simulation(params, prepared)?;
+    let suggested_gas_price = ctx.suggested_gas_price(&executed.transaction);
+    Ok(build_simulation_result(executed, suggested_gas_price))
```
alex-mysten, line 127: Have we audited the test coverage for this path to ensure that the refactor preserves behavior?

crates/sui-core/src/authority.rs
```diff
@@ -2510,245 +2511,36 @@ impl AuthorityState {
...
-        // create an address balance withdrawal for gas.
-        // Skip mock gas for gasless transactions — they don't use gas coins.
-        let is_gasless = protocol_config.enable_gasless() && transaction.is_gasless_transaction();
-        let mock_gas_object = if allow_mock_gas_coin && transaction.gas().is_empty() && !is_gasless
-        {
-            let obj = Object::new_move(
-                MoveObject::new_gas_coin(
-                    OBJECT_START_VERSION,
-                    ObjectID::MAX,
-                    DEV_INSPECT_GAS_COIN_VALUE,
```
alex-mysten, line 2546: Should this const live in simulation?

alex-mysten (commented):
Appreciate the refactor, the separation into stages makes this easier to reason about!

crates/sui-core/src/authority.rs
stefan-mysten: Got it. Can we move it out from `AuthorityStore` struct and make it into a free function there, so that the simulation can reuse it?

crates/sui-core/src/simulation.rs
stefan-mysten: Will remove.

crates/sui-core/src/authority.rs
stefan-mysten: Yes, good catch.

crates/sui-core/src/simulation.rs
stefan-mysten: I will have another check.

crates/sui-core/src/authority.rs
stefan-mysten: It's referenced in a few places in `authority.rs`, so I think it's fine to stay there.

crates/sui-core/src/authority.rs
stefan-mysten: The function sig of `simulate_transaction` in `TransactionExecutor` is 
https://github.com/MystenLabs/sui/blob/0ec8fd7b2f0fa7144d097f7a8781249cb18c5574/crates/sui-types/src/transaction_executor.rs?plain=1#L27-L32

So I thought keeping it like that, but sure, we can move it to the params struct. I'll check for issues with lifetimes.

crates/sui-core/src/authority.rs
stefan-mysten: Moved!

crates/sui-core/src/simulation.rs
```diff
@@ -0,0 +1,526 @@
...
+    pub transaction_deny_config: &'a TransactionDenyConfig,
+    pub verifier_signing_config: &'a VerifierSigningConfig,
+    pub certificate_deny_set: &'a HashSet<TransactionDigest>,
+    pub bytecode_verifier_metrics: Arc<BytecodeVerifierMetrics>,
+    pub execution_metrics: Arc<ExecutionMetrics>,
+    pub package_store: &'a dyn BackingPackageStore,
+    pub backing_store: &'a dyn BackingStore,
+    pub coin_reservation_resolver: &'a dyn CoinReservationResolverTrait,
+    pub account_funds_read: &'a dyn AccountFundsRead,
+}
```
dariorussi, line 59: this feels way too much of a kitchen sink to me and I have difficulty understanding what is abstracting.
I'll make more comments around hoping they will clarify a bit more.
One suggestion I got was to split this in 2 struct
```
pub struct SimulationData<'a> { /* the 10 concrete fields */ }
pub struct SimulationStores<'a> { /* the 4 trait objects */ }
```
which is an option though I am not sure I'd go there.
It still feels the break up is not quite there for me

crates/sui-core/src/simulation.rs
```diff
@@ -0,0 +1,526 @@
...
+        receiving_object_refs: &[ObjectRef],
+    ) -> SuiResult<(InputObjects, ReceivingObjects)>;
+
+    /// Returns a caller-specific suggested gas price, if simulation should surface one.
+    ///
+    /// Fullnodes can use this to report congestion-aware pricing, while simulators that do not
+    /// compute such a signal can keep the default `None`.
+    fn suggested_gas_price(&self, _tx: &TransactionData) -> Option<u64> {
+        None
+    }
```
dariorussi, line 80: what is this doing here?
I do not understand this api. Simulacrum/forking will never use this and it feels an exclusive responsibility of a fullnode/validator. In that respect the fullnode/validator code can compute this and it would never enter the simulation pipeline. Basically keep that in the authority side.
Is that possible?

crates/sui-core/src/simulation.rs
```diff
@@ -0,0 +1,526 @@
...
+        mock_gas_id,
+    })
+}
+
+/// Runs deny list checks and processes funds withdrawals. Called before loading
+/// input objects, since these checks don't depend on object state.
+///
+/// This is the single canonical implementation shared by both the fullnode
+/// simulation path and the `AuthorityState` signing path.
+pub fn pre_object_load_checks(
```
tzakian, line 280: This feels like the wrong place/wrong way round for where this function should live IMO.

I think it should probably still reside in `authority.rs` since it is crucial to the validator execution/signing paths, but then simulate can use this function.

crates/sui-core/src/authority.rs
```diff
@@ -965,6 +965,34 @@ pub struct AuthorityState {
...
+        _tx_digest: &TransactionDigest,
+        input_object_kinds: &[InputObjectKind],
+        receiving_object_refs: &[ObjectRef],
+    ) -> SuiResult<(InputObjects, ReceivingObjects)> {
+        self.authority.input_loader.read_objects_for_signing(
+            // We don't want to cache this transaction since it's a simulation.
+            None,
+            input_object_kinds,
+            receiving_object_refs,
+            self.epoch_store.epoch(),
```
dariorussi, line 985: there is an `EpochId` saved in `SimulationParams` and this one used here.
I am not sure what is the correlation between them but it feels strange to me

dariorussi (commented):
some feedback though I am looking at this more and I suspect it will take me the rest of the day to build a good sense of what is going on.
It is not clear how the refactor makes anything more clear to me.
On a side note please, if @alex-mysten or anyone from core accepts this do not wait for me as I will just give you feedback and it's for them to accept it.

crates/sui-core/src/simulation.rs
tzakian: +1. I think separating this into "data" and "behavior" may make this a bit nicer.

crates/sui-core/src/simulation.rs
stefan-mysten: Got moved back into `authority.rs` but did not push the commits yet as I was working on more tests to ensure correctness with the original code.

stefan-mysten:
Thanks all for the reviews. Mulching a bit more over this, to see if it can be a lot simpler. Might have another PR to compare.

stefan-mysten:
Closing this in favor of doing some refactorings + tests before moving things around and adding more stuff.

PR #26323, opened 2026-04-20
sui-benchmark: retry soft bundles on retriable wait_for_effects responses

aschran:
## Summary

The bench soft-bundle proxy already retries on submit-time retriable rejections. Extend the same retry (10 attempts / 60s) to cover retriable `wait_for_effects` responses — `Rejected` with a retriable reason and `Expired` — which similarly mean the tx was never ordered.

This removes spurious epoch-boundary failures the composite alias-revoke workload was working around, and lets that workaround (per-batch checkpoint polls + digest tracking set) be deleted.

## Behavior change

Workloads that count `RetriableFailure` events will see fewer; most now resolve to `Success` or `PermanentFailure` via proxy-level retry. `Rejected{error: None}` still surfaces as `UnknownRejection`.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/src/workloads/composite/mod.rs
+++ b/crates/sui-benchmark/src/workloads/composite/mod.rs
@@ -467,6 +467,37 @@ struct AliasState {
     cycle_state: AliasRevokeCycleState,
 }
 
+impl AliasState {
+    /// Apply the result of polling `is_transaction_checkpointed` for the currently
+    /// pending Add or Remove alias tx. If checkpoint-confirmed, transitions to the
+    /// next cycle state.
+    fn confirm_pending_alias_checkpoint(&mut self, checkpointed: bool) {
+        if !checkpointed {
+            return;
+        }
+        let next_state = match &self.cycle_state {
+            AliasRevokeCycleState::AddPending {
+                tx_digest: Some(digest),
+            } => {
+                info!("Add alias tx {digest} checkpoint confirmed");
+                Some(AliasRevokeCycleState::Active {
+                    successful_alias_txs: 0,
+                })
+            }
+            AliasRevokeCycleState::RemovePending {
+                tx_digest: Some(digest),
+            } => {
+                info!("Remove alias tx {digest} checkpoint confirmed");
+                Some(AliasRevokeCycleState::Revoked)
+            }
+            _ => None,
+        };
+        if let Some(s) = next_state {
+            self.cycle_state = s;
+        }
+    }
+}
+
 #[derive(Clone, Copy)]
 enum BatchTxKind {
     /// Standard composite PTB.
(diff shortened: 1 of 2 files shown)


crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -733,6 +736,36 @@ impl CompositePayload {
...
+        if !pending_pre_revoke_alias_txs.is_empty() {
+            let checkpointed = join_all(
+                pending_pre_revoke_alias_txs
+                    .iter()
+                    .map(|digest| self.fullnode_proxies[0].is_transaction_checkpointed(digest)),
+            )
+            .await;
+            if let Some(alias_state) = self.alias_state.as_mut() {
+                for (digest, checkpointed) in
+                    pending_pre_revoke_alias_txs.into_iter().zip(checkpointed)
```
alex-mysten: isn't zip illegal? or not in test

crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -771,11 +804,14 @@ impl CompositePayload {
...
                 successful_alias_txs,
-            } if *successful_alias_txs >= self.config.alias_txs_before_revoke => true,
+            } if *successful_alias_txs >= self.config.alias_txs_before_revoke => {
+                alias_state.in_flight_pre_revoke_alias_txs.is_empty()
+            }
             AliasRevokeCycleState::Active { .. } => {
                 rng.gen_bool(self.config.alias_tx_probability as f64)
             }
-            AliasRevokeCycleState::Revoked | AliasRevokeCycleState::NeedAdd => true,
+            AliasRevokeCycleState::Revoked => alias_state.in_flight_pre_revoke_alias_txs.is_empty(),
```
alex-mysten: is this state reachable now?

crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -733,6 +736,36 @@ impl CompositePayload {
...
+                    .iter()
+                    .copied()
+                    .collect::<Vec<_>>()
+            })
+            .unwrap_or_default();
+        if !pending_pre_revoke_alias_txs.is_empty() {
+            let checkpointed = join_all(
+                pending_pre_revoke_alias_txs
+                    .iter()
+                    .map(|digest| self.fullnode_proxies[0].is_transaction_checkpointed(digest)),
```
alex-mysten: is it possible that an "inflight" transaction could never be checkpointed, such as in case of failure or denial?

crates/sui-benchmark/src/workloads/composite/mod.rs
aschran: wow that one snuck by me! it is indeed failing in CI, thanks for pointing out

crates/sui-benchmark/src/workloads/composite/mod.rs
aschran: yes, & refactored to add tests for this

crates/sui-benchmark/src/workloads/composite/mod.rs
aschran: No: response status and on-chain outcome are independent. The executor's `RetriableFailure` only means *that RPC attempt* failed transiently — it doesn't imply anything about whether validator quorum already finalized the tx. So a digest in this set is in one of two cases: (a) the tx really didn't land, in which case the driver retries until it gets a definitive `Success` / `PermanentFailure` / `UnknownRejection` (each of which drains it via `handle_alias_tx_result`), or (b) the tx already landed but the executor's response was retriable — typical around epoch change — which is what the checkpoint poll drains.

PR #26326, opened 2026-04-20
[test] add send_funds PTB benchmark

mystenmark:
## Summary

- Adds a `bench-programmable` subcommand to the Sui transactional test runner that wraps a programmable transaction with a Criterion `prepare_txn` loop (release mode only).
- Adds `crates/sui-adapter-transactional-tests/tests/address_balances/send_funds_bench.move`, which benchmarks `sui::balance::send_funds` for a SUI withdrawal from the sender's address balance, paying gas from that same address balance.
- Each bench iteration gets its own Criterion baseline keyed by task number (e.g. `benchmark_tx_task_3`).

## Running the benchmark

```
cargo nextest run --release -p sui-adapter-transactional-tests --test-threads 1 --no-capture -E 'test(/send_funds_bench/)'
```

Criterion output shows up inline; results are also cached in `target/criterion/` for baseline comparison across runs.

## Sample timings (local)

| Case | Commands | prepare_txn time |
|------|----------|------------------|
| 1 withdraw + 1 send | 2 | ~118 µs |
| 4 withdraws + 4 sends | 8 | ~317 µs |

## Test plan

- [x] Debug-mode run passes snapshot (`cargo nextest run -p sui-adapter-transactional-tests -E 'test(/send_funds_bench/)'`)
- [x] Release-mode run produces Criterion timings for both bench tasks
- [x] Existing `address_balances` tests still pass (regular `programmable` path unchanged in behavior)

### DIFF
--- a/crates/sui-adapter-transactional-tests/tests/address_balances/send_funds_bench.move
+++ b/crates/sui-adapter-transactional-tests/tests/address_balances/send_funds_bench.move
@@ -1,0 +1,39 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Benchmark the send_funds command with a SUI withdrawal from A's address balance
+// sent to B, with gas also paid from A's address balance.
+
+//# init --addresses test=0x0 --accounts A B C D E --enable-accumulators --enable-address-balance-gas-payments
+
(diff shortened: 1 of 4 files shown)


tzakian (approved):
👍🏻

PR #26327, opened 2026-04-21
[sui-rpc-api] Add custom service support in rpc api

stefan-mysten:
## Description 

This PR enables a new gRPC service to be included through the RpcService type.

## Test plan 

Existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-rpc-api/src/lib.rs
+++ b/crates/sui-rpc-api/src/lib.rs
@@ -1,13 +1,17 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::convert::Infallible;
+use std::sync::Arc;
+
 use mysten_network::callback::CallbackLayer;
 use reader::StateReader;
-use std::sync::Arc;
 use subscription::SubscriptionServiceHandle;
 use sui_types::storage::RpcStateReader;
 use sui_types::transaction_executor::TransactionExecutor;
 use tap::Pipe;
+use tonic::server::NamedService;
+use tower::Service;
 
 pub mod client;
 mod config;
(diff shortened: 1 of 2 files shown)


bmwill:
What's the motivation for this change?

stefan-mysten:
> What's the motivation for this change?

Sorry, left a more detailed description on Slack. I will explain here as well.

For the forking tool, I am reusing the `RpcService` to start the rpc service and take advantage of the existing impl. I am adding a new grpc interface for interacting with the forked network (advance checkpoint, clock, etc), and I would like to register it as well. I initially did it via axum + custom routing, but it felt a bit weird?

This also does not feel 100% right because I guess it isn't supposed to be a generic RpcService builder, so wanted to see what would be the suggested way to go.

bmwill (approved):

crates/sui-rpc-api/src/lib.rs
```diff
@@ -57,6 +61,9 @@ pub struct RpcService {
     server_version: Option<ServerVersion>,
     metrics: Option<Arc<RpcMetrics>>,
     config: Config,
+    extra_routes: axum::Router,
```
amnn, line 64: I'm not following why you need to add extra routes -- I thought your aim was to add extra gRPC services -- don't they all get served under the same route?

bmwill: No, gRPC works with path based routing, every service is routed by "/package.servicename/methodname". Difference services do not share a path prefix and it is against spec to nest gRPC services anywhere other than the root "/" path.

PR #26328, opened 2026-04-21
[forking 10/n] add forking service grpc for interacting with the network

stefan-mysten:
## Description 

This PR adds a proto description for the ForkingService that enables users to interact with the forked network through the common RPC apis. The main new APIs are:
```
sui.forking.ForkingService/GetStatus
sui.forking.ForkingService/AdvanceClock
sui.forking.ForkingService/AdvanceCheckpoint
```

The changes to `sui-rpc-api` are now pushed in a different PR (https://github.com/MystenLabs/sui/pull/26327), so I plan to merge that first before this one, and remove the changes here.

## Test plan 

Existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-forking/proto/sui/forking/v1alpha/forking_service.proto
amnn, line 1: I'm curious to get @bmwill or maybe @nickvikeras' take on whether we should version this schema. The tool that talks to this service is distributed with the binary that runs the service, so it seems unnecessary, but it does seem to be a convention we follow (it comes from [AIP 185](https://google.aip.dev/185), I believe), and we do follow this in the case of the `ConsistentService` which is officially only an internal component of GraphQL (so they are deployed together).

crates/sui-forking/build.rs
amnn, line 1: It can be quite helpful to split the proto into its own crate that handles the codegen etc for you, so that clients don't need to depend on the server implementation.

amnn: Also curious to get @bmwill's or @nickvikeras's take on whether to house the logic to sync generated files in `build.rs` or within a test. My recollection is that `sui-rpc-api` originally used a test for this (and I copied that into `sui-indexer-alt-consistent-store`, but I see now that `sui-rpc-api` no longer does this, so the evolution there might be interesting.

amnn (approved):
Core change looks reasonable here -- I've looped in @bmwill and @nickvikeras to share their thoughts from building/maintaining a more substantial schema.

crates/sui-forking/build.rs
bmwill: I personally prefer to not use build.rs to do codegen because it requires your consumers to take a dep on a protoc compiler. That and LSP is more useful. 

Note that the actual gRPC interface code lives in sui-rpc which doesn't live in this repo. Eventually, assuming we expect others to depend on this interface, we should also pull it out there with canonical definitions in sui-apis repo

crates/sui-forking/proto/sui/forking/v1alpha/forking_service.proto
bmwill: Yes this is a hard requirement that we versions the apis, before they are stable please use alphav1 or betav1. Also look at setting up buf lint

crates/sui-forking/build.rs
stefan-mysten: Just to make sure I understand @bmwill, you're suggesting that the forking interface should go into `sui-rpc` at some point. Is it ok for now to keep it here until this whole tool gets to a reasonable stable version?

crates/sui-forking/proto/sui/forking/v1alpha/forking_service.proto
stefan-mysten: Noted, will add versioning.

crates/sui-forking/build.rs
stefan-mysten: Thinking of having a script that generates the proto files. I expect them not to change often, maybe just one more time when adding the `advance-epoch`, so for now seems a reasonable option to avoid using `build.rs`.

bmwill: Yes, I think it's actually ideal to leave the interfaces here while iterating, then once we're ready to stabilize the whole thing we can evaluate having the APIs live with the others.

stefan-mysten: Great, let's do that.

PR #26329, opened 2026-04-21
docs: automated updates for framework_change

jessiemongeon1:
## Summary

Automated documentation updates triggered by recent release notes.

### Changes

**Triggered by [MystenLabs/sui#26248](https://github.com/MystenLabs/sui/pull/26248)**: [sui-framework] Add df/dof `replace`. Fix `remove_opt`
- 🔴 `docs/content/develop/objects/dynamic-fields.mdx`

### Review details

<details>
<summary><code>docs/content/develop/objects/dynamic-fields.mdx</code></summary>

The documentation page needs to be updated to reflect the new `replace` function added to `sui::dynamic_field` and `sui::dynamic_object_field`.

### Section: "Adding dynamic fields" (last paragraph)

**Current text:**
> "A transaction that attempts to add a field with the same `<Name>` type and value as one that is already defined fails. You can modify fields in-place by borrowing them mutably and you can overwrite them safely by removing the old value first."

**What's wrong:** This paragraph is now incomplete/inaccurate. It describes only two ways to update a field value (borrow mutably or remove then add), but omits the new `replace` function which provides a direct, safe way to overwrite an existing field value.

**Suggested update:** Add mention of the `replace` function as a third option:
> "A transaction that attempts to add a field with the same `<Name>` type and value as one that is already defined fails. You can modify fields in-place by borrowing them mutably, overwrite them safely by removing the old value first, or use the `replace` function to swap out the value directly."

### New section needed: "Replacing a dynamic field value"

A new subsection should be added (likely between "Accessing dynamic fields" and "Removing a dynamic field") documenting the `replace` function, similar to how `add`, `borrow`/`borrow_mut`, and `remove` are each documented. It should include:

- The function signature from `dynamic_field` (via `ImportContent` or inline code block)
- A note that `sui::dynamic_object_field` has an equivalent function
- Explanation that `replace` requires the field to already exist (aborts otherwise), and returns the old value
- Clarification of when to use `replace` vs. `borrow_mut` vs. `remove` + `add`

</details>

---

This PR was automatically generated by the **Docs Impact Monitor**. Please review the changes carefully before merging.

Each file edit was generated by Claude based on the release notes and a review of the existing documentation content.


jessiemongeon1:
## Style Guide Audit

**9** violation(s) found. All must be fixed before merge.

### `docs/content/develop/objects/dynamic-fields.mdx` (9 violation(s))

_9 violation(s) (0 regex, 9 claude)_

- **Line 23** — active-voice
  - Current: `are considered wrapped and is not accessible`
  - Fix: `you consider them wrapped and cannot access them`
- **Line 134** — active-voice
  - Current: `is rejected for having invalid inputs`
  - Fix: `the transaction fails with invalid inputs`
- **Line 193** — active-voice
  - Current: `renders them all inaccessible`
  - Fix: `makes them all inaccessible`
- **Line 195** — active-voice
  - Current: `is particularly undesirable for onchain collection types`
  - Fix: `particularly affects onchain collection types`
- **Line 221** — active-voice
  - Current: `are hidden when viewed from external storage`
  - Fix: `external storage hides them`
- **Line 221** — active-voice
  - Current: `keeps those objects visible`
  - Fix: `you can view those objects`
- **Line 334** — active-voice
  - Current: `is guaranteed to succeed at runtime`
  - Fix: `always succeeds at runtime`
- **Line 348** — active-voice
  - Current: `is only considered equal to itself`
  - Fix: `the system only considers it equal to itself`
- **Line 356** — active-voice
  - Current: `This is unlikely to be the definition of equality that you want.`
  - Fix: `You likely do not want this definition of equality.`

---

PR #26330, opened 2026-04-21
[fix] use prod vk for testnet

joyqvq:
## Description 

there is a regression in grpc vs json rpc: https://github.com/MystenLabs/sui/blob/main/crates/sui-json-rpc/src/read_api.rs#L1178

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-rpc-api/src/grpc/v2/signature_verification_service.rs
+++ b/crates/sui-rpc-api/src/grpc/v2/signature_verification_service.rs
@@ -175,10 +175,10 @@ fn verify_signature(
     };
 
     let mut zklogin_verifier = match service.chain_id().chain() {
-        sui_protocol_config::Chain::Mainnet | sui_protocol_config::Chain::Testnet=> sui_crypto::zklogin::ZkloginVerifier::new_mainnet(),
-        sui_protocol_config::Chain::Unknown => {
-            sui_crypto::zklogin::ZkloginVerifier::new_dev()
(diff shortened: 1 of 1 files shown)


crates/sui-rpc-api/src/grpc/v2/signature_verification_service.rs
```diff
@@ -175,8 +175,8 @@ fn verify_signature(
     };
 
     let mut zklogin_verifier = match service.chain_id().chain() {
-        sui_protocol_config::Chain::Mainnet => sui_crypto::zklogin::ZkloginVerifier::new_mainnet(),
-        sui_protocol_config::Chain::Testnet | sui_protocol_config::Chain::Unknown => {
+        sui_protocol_config::Chain::Mainnet | sui_protocol_config::Chain::Testnet=> sui_crypto::zklogin::ZkloginVerifier::new_mainnet(),
```
bmwill: Has this always been the case 😅

bmwill (approved):

crates/sui-rpc-api/src/grpc/v2/signature_verification_service.rs
joyqvq: yes since we had it in json rpc and also graphql! https://github.com/MystenLabs/sui/blob/main/crates/sui-json-rpc/src/read_api.rs#L1178

PR #26335, opened 2026-04-21
sui-indexer-alt-framework: add BackfillService

evan-wall-mysten:
## Description

backfill.rs implements a shared backfill service for pipeline subscribers that need historical checkpoint data.

  - BackfillService
      - Owns subscriptions before startup.
      - Creates BackfillHandles for each pipeline range.
      - Runs one async service loop that handles requests, fetch completions, and cache eviction timers.
  - BackfillHandle
      - Per-pipeline client handle.
      - Sends progress updates with set_pipeline_lo.
      - Requests the next historical checkpoint chunk with next_chunk.
      - Sends a drop/close signal when the handle is dropped.
  - BackfillState
      - Tracks remaining ranges for active handles.
      - Fetches checkpoints through IngestionClient.
      - Caches chunk results so overlapping handles share the same fetch work.
      - Tracks which handles are pending or expected consumers of each cached chunk.
      - Uses an index from handle ID to cache chunks so handle close does not scan the whole cache.

  Events handled:

  - SetPipelineLo
      - Sent by a handle when its pipeline advances its committed low watermark.
      - Updates that handle’s active range.
      - If the range becomes empty, closes the handle’s remaining cache state.
  - NextChunk
      - Sent by a handle to request its next backfill chunk.
      - Computes the aligned chunk bounds.
      - Creates a cache entry and fetch tasks if needed.
      - Adds the handle as a pending waiter for that chunk.
      - Dispatches immediately if the chunk is already complete and no expected consumers remain.
  - DropHandle
      - Sent from BackfillHandle::drop.
      - Removes that handle from active ranges, expected-consumer sets, pending waiters, and the cache index.
      - Sends None to any pending request for the dropped handle.
      - May unblock and dispatch chunks that were waiting only for that handle.
  - Fetch
      - Internal task result for one checkpoint fetch.
      - Inserts the checkpoint envelope into the matching cache entry.
      - Once the chunk is complete, either dispatches pending waiters or starts an eviction timer if expected consumers remain.
  - Evict
      - Internal timer event after expected_consumer_wait_duration.
      - Releases a completed cached chunk even if expected consumers never arrived.
      - Dispatches pending waiters before removing the cache entry.

  Shutdown behavior:

  - If all request senders are gone, the service exits.
  - On shutdown, pending handles receive None instead of hanging.

## Test plan 

`BackfillService` is not yet connected to the `Indexer` so this change only adds unit tests. 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/pipeline/concurrent/backfill.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/concurrent/backfill.rs
@@ -1,0 +1,1053 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::BTreeMap;
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::collections::hash_map::Entry;
+use std::ops::Range;
+use std::time::Duration;
+
+use sui_futures::service::Service;
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+use tokio::task::JoinSet;
+use tokio::time::Instant;
+use tracing::info;
+
+use crate::ingestion::ingestion_client::CheckpointEnvelope;
+use crate::ingestion::ingestion_client::IngestionClient;
+
+const BACKFILL_REQUEST_CHANNEL_SIZE: usize = 1024;
+
+/// Configuration for [`BackfillService`].
+#[derive(Clone)]
+pub struct BackfillConfig {
+    /// Bucket size, and also the maximum number of checkpoints any single
+    /// [`BackfillHandle::next_chunk`] call returns.
+    pub chunk_size: u64,
+
+    /// Maximum distance above a bucket's upper bound for a handle to be considered an expected
+    /// consumer of that bucket. This bounds cache retention to consumers likely to share the same
+    /// completed fetch.
+    pub expected_consumer_max_distance: u64,
+
+    /// How long to keep a completed cache entry around for expected consumers.
+    pub expected_consumer_wait_duration: Duration,
+
+    /// Polling interval to retry fetching checkpoints that do not exist yet.
+    pub retry_interval: Duration,
+}
+
+/// A source of checkpoint data for backfills.
+pub struct BackfillService {
+    backfill_config: BackfillConfig,
+    ingestion_client: IngestionClient,
+    pipeline_ranges: HashMap<BackfillHandleId, Range<u64>>,
+    request_tx: mpsc::Sender<BackfillRequest>,
+    request_rx: mpsc::Receiver<BackfillRequest>,
+}
+
+/// Per-pipeline subscription handle returned by [`BackfillService::subscribe`].
+pub struct BackfillHandle {
+    backfill_handle_id: BackfillHandleId,
+    request_tx: mpsc::Sender<BackfillRequest>,
+}
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+struct BackfillHandleId(usize);
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+struct CacheEntryId(usize);
+
+enum BackfillRequest {
+    SetPipelineLo {
+        backfill_handle_id: BackfillHandleId,
+        pipeline_lo: u64,
+    },
+    NextChunk {
+        backfill_handle_id: BackfillHandleId,
+        reply: oneshot::Sender<Option<Vec<CheckpointEnvelope>>>,
+    },
+}
+
+struct CacheEntry {
+    /// Generation ID for this `chunk_hi`, used to ignore stale fetch completions after eviction.
+    cache_entry_id: CacheEntryId,
+    /// Expected consumers keep a completed chunk warm without blocking forever on distant handles.
+    remaining_backfill_handle_ids: HashSet<BackfillHandleId>,
+    /// Requests wait here so concurrent consumers share one fetch fan-out.
+    pending_handles: HashMap<BackfillHandleId, PendingHandle>,
+    /// Ordered storage lets a shared cache entry serve per-handle sub-ranges deterministically.
+    checkpoint_envelopes: BTreeMap<u64, CheckpointEnvelope>,
+    /// A count avoids assuming fetch completions arrive contiguously or in order.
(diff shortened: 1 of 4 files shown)


crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
amnn, line 1: Note that the aim of factoring this service out in this way, with the interfac eit has, was so that it could belong as part of ingestion.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/backfill.rs
amnn, line 1: This service is written in quite a different style from the rest of the framework, and what we had discussed earlier in the week (maybe there's a reason for this, but I couldn't spot one), sketching that out a bit more, it looks something like this:

- `BackfillService` works like `IngestionService` in that you initialize the service, add subscribers, and then call `run` to spawn an actor task as a `Service`.
- From then on communication between the service and its subscribers happens through channels -- I think you could probably use a single channel that accepts a message enum (there are two message kinds: "I changed pipeline_lo", and "I want my next chunk"). 
- Shared state is managed as local variables in that actor task (not an `Arc<...>` that is shared across multiple handles). The shared state contains the boundaries of each subscriber, and a map of channels to talk to pending chunks.

Apart from helping with consistency and understandability across this codebase, this actor style makes it so that most of the local state is hidden behind a sequential actor which is much much easier to reason about.

amnn (commented):
Just some initial comments -- the style here is quite different and it would be good to bring this in line with the rest of the codebase.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/backfill.rs
wlmyng: It seems the new machinery a backfill command would require is a watermark task that pushes `reader_lo` down on contiguity. I'm not sure if we actually need a separate BackfillService? We could have an `IngestionService` for forward, and set up another one for backfill. Each pipeline subscribes to one or both as needed. It's true that disjoint pipelines will wait until the backfill ingestion service can service them but I think that's totally fine? This ends up being a mirror of the forward ingestion case

Everything downstream, process, commit, is almost the same, `commit_watermark` is purely responsible for bumping up `commit_hi`, so the overlapping conflict comes down to pruning vs backfilling

Pruning and backfilling want `reader_lo` advanced in opposite directions, but a pipeline shouldn't be attempting both at the same time anyways - it's either extending coverage, or has sufficient retention and is periodically pruning. And I think that a pipeline would backfill at most once within the lifetime of an indexer run

evan-wall-mysten: @amnn in the original plan I was going to have the new version of the reader_lo task's event loop be what drives backwards ingestion. BackfillService was just keeping track of the state of each pipeline, fetching checkpoints, and returning them to handlers. 

I refactored BackfillService to have its own event loop which can interact with the reader_lo task's event loop. BackfillService now contains BackfillState which serves a similar role to the original version of BackfillService, but the collections inside do not need to be threadsafe anymore.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/backfill.rs
```diff
@@ -0,0 +1,1267 @@
...
+    /// consumer of that bucket. This bounds cache retention to consumers likely to share the same
+    /// completed fetch.
+    pub expected_consumer_max_distance: u64,
+
+    /// How long to keep a completed cache entry around for expected consumers.
+    pub expected_consumer_wait_duration: Duration,
+
+    /// Polling interval to retry fetching checkpoints that do not exist yet.
+    pub retry_interval: Duration,
+}
```
wlmyng, line 42: But is there a reason why we can't just create a separate `IngestionService` for pipelines that need backfill, starting from the most recent checkpoint and working backwards?

crates/sui-indexer-alt-framework/src/pipeline/concurrent/backfill.rs
```diff
@@ -0,0 +1,1267 @@
...
+                    request = request_rx.recv() => {
+                        let Some(request) = request else {
+                            // All handles have been dropped; pending tasks no longer have consumers.
+                            break;
+                        };
+                        match request {
+                            BackfillRequest::SetPipelineLo { backfill_handle_id, pipeline_lo, .. } => state.handle_set_pipeline_lo(backfill_handle_id, pipeline_lo),
+                            BackfillRequest::NextChunk { backfill_handle_id, reply, .. } => state.handle_next_chunk(backfill_handle_id, reply),
+                            BackfillRequest::DropHandle { backfill_handle_id } => state.close_handle(backfill_handle_id),
+                        }
```
wlmyng, line 194: It seems the backfill service logic is largely around maintaining a cache of checkpoints for the various subscribers that need backfill, but could we not just mirror the forward scenario? The main issue I can think of is that there's no semblance of order so we could be fetching a bunch of checkpoints from some non-contiguous point instead of actually working backwards. But then, could we just have an orchestrator that tells the ingestion service to do smaller batches going backwards?

crates/sui-indexer-alt-framework/src/pipeline/concurrent/backfill.rs
```diff
@@ -0,0 +1,1267 @@
...
+            } = pending_handle;
+            let checkpoint_envelopes = cache_entry
+                .checkpoint_envelopes
+                .range(response_range)
+                .map(|(_, v)| v.clone())
+                .collect();
+            // A dropped receiver means the caller no longer needs the backfill result.
+            let _ = reply.send(Some(checkpoint_envelopes));
+        }
+    }
```
wlmyng, line 477: When we start an indexer we should be able to determine the full range we need to backfill for the indexer (calculated across all pipelines in need of backfill) right?

Even if over the course of the indexer run this reader/ pruner shifts, I think it's ok for us to overshoot. When we hand back to the pruner, it'll just have some more work to do

PR #26337, opened 2026-04-21
Fix coin reservation panic during checkpoint replay

mystenmark:
## Summary

Fix panics in coin reservation transactions during checkpoint replay. The core issue is that
coin reservation transactions can execute before the settlement transaction that creates the
accumulator account objects they depend on.

### Changes

1. **Don't cache not-found errors in CachingCoinReservationResolver** — Not-found errors are
   transient (the object will appear once the settlement executes). Caching them poisons the
   cache for all subsequent lookups.

2. **Skip transaction rewriting on early execution error** — When a transaction is already
   known to fail (e.g. InsufficientFundsForWithdraw), skip `rewrite_transaction_for_coin_reservations`
   to avoid panicking on a deleted accumulator object. (Cherry-pick of lost commit 7be4694fc6)

3. **Add CoinReservationWithdraw to CompositeWorkloadConfig::balanced** — Ensures the balanced
   workload config exercises coin reservation withdrawals.

4. **Add accumulator root dependency for coin reservation transactions** — When the execution
   scheduler enqueues a transaction that has coin reservations, add the accumulator root object
   at the assigned accumulator version as an input dependency. This makes the scheduler wait
   for the settlement transaction (which writes the accumulator root at that version) to
   complete before executing the coin reservation transaction. This is the primary fix.

## Test plan
- [x] `seed-search.py --num-seeds 200 test_coin_reservation_checkpoint_replay` — 200/200 pass
- [x] `seed-search.py --num-seeds 200 test_simulated_load_reconfig_with_crashes_and_delays` — 200/200 pass

### DIFF
--- a/crates/sui-benchmark/src/workloads/composite/mod.rs
+++ b/crates/sui-benchmark/src/workloads/composite/mod.rs
@@ -310,6 +310,7 @@ impl CompositeWorkloadConfig {
         probabilities.insert(ObjectBalanceOverdraw::NAME, 0.1);
         probabilities.insert(AuthenticatedEventEmit::NAME, 0.1);
         probabilities.insert(ImmutableObjectRead::NAME, 0.2);
+        probabilities.insert(CoinReservationWithdraw::NAME, 0.1);
         Self {
             probabilities,
             alias_tx_probability: 0.3,
--- a/crates/sui-benchmark/tests/simtest.rs
+++ b/crates/sui-benchmark/tests/simtest.rs
@@ -1595,6 +1595,7 @@ mod test {
     #[sim_test(config = "test_config()")]
     async fn test_composite_workload() {
         sui_protocol_config::ProtocolConfig::poison_get_for_min_version();
+
         let test_cluster = build_test_cluster(4, 10000, 1).await;
 
         let protocol_config = sui_protocol_config::ProtocolConfig::get_for_version(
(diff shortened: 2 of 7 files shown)


crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -851,84 +879,101 @@ impl CheckpointExecutor {
...
+    /// could execute before the accumulator it references exists on this node.
     #[instrument(level = "info", skip_all)]
-    fn schedule_transaction_execution(
+    fn schedule_settlement_transactions(
         &self,
         ckpt_state: &CheckpointExecutionState,
         tx_data: &CheckpointTransactionData,
-    ) -> Vec<TransactionDigest> {
+    ) -> (
+        Vec<TransactionDigest>,
```
aschran: return type is a bit complicated, maybe document what each of these are with a comment or wrap in a struct?

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -585,14 +585,42 @@ impl CheckpointExecutor {
         pipeline_handle: &mut PipelineHandle,
     ) -> CheckpointExecutionState {
         let sequence_number = checkpoint.sequence_number;
-        let (mut ckpt_state, tx_data, unexecuted_tx_digests) = {
+        let (mut ckpt_state, tx_data, settlement_digests, remaining_txns, unexecuted_tx_digests) = {
```
aschran: nit - the word "remaining" here is confusing out of context, remaining from what?

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -585,14 +585,42 @@ impl CheckpointExecutor {
...
+                tx_data,
+                settlement_digests,
+                remaining_txns,
+                unexecuted_tx_digests,
+            )
         };
 
+        finish_stage!(pipeline_handle, ExecuteSettlements);
+
+        // Wait for settlement transactions to finish before enqueuing other transactions.
```
aschran: this feels a bit abstraction-breaking, bc (iiuc) we normally just enqueue everything and rely on the execution scheduler to resolve/wait for missing dependencies. so it seems strange to be treating this one particular missing dependency differently. is it because we are able to wait for missing versions but not entire missing objects? just out of curiosity would it be hard to augment the execution scheduler to be able to handle missing objects, and if so why?

sorry if my mental model of this is all wrong lol

mystenmark: this is fair, a better way might be to add a shared object version dependency to the accumulator root for any transactions using coin reservations. i will try that

aschran (approved):

PR #26338, opened 2026-04-21
sui-indexer-alt-framework: separate pruner from prune watermark

evan-wall-mysten:
## Description 

This change splits the concurrent pruner into two tokio tasks to mirror the existing committer + commit_watermark pattern: one task distributes work, a second task owns DB watermark persistence.

## Test plan 

New unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Added new `PrunerConfig::pruner_channel_size`.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
@@ -23,6 +23,7 @@ use crate::pipeline::concurrent::collector::collector;
 use crate::pipeline::concurrent::commit_watermark::commit_watermark;
 use crate::pipeline::concurrent::committer::committer;
 use crate::pipeline::concurrent::main_reader_lo::track_main_reader_lo;
+use crate::pipeline::concurrent::prune_watermark::prune_watermark;
 use crate::pipeline::concurrent::pruner::pruner;
 use crate::pipeline::concurrent::reader_watermark::reader_watermark;
 use crate::pipeline::processor::processor;
@@ -33,6 +34,7 @@ mod collector;
 mod commit_watermark;
 mod committer;
 mod main_reader_lo;
+mod prune_watermark;
 mod pruner;
 mod reader_watermark;
 
@@ -142,6 +144,9 @@ pub struct ConcurrentConfig {
 
     /// Size of the channel between the committer and the watermark updater.
     pub committer_channel_size: Option<usize>,
+
+    /// Size of the channel between the pruner and the pruner watermark updater.
+    pub pruner_channel_size: Option<usize>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
@@ -257,6 +262,7 @@ pub(crate) fn pipeline<H: Handler>(
         processor_channel_size,
         collector_channel_size,
         committer_channel_size,
+        pruner_channel_size,
     } = config;
 
     let concurrency = fanout.unwrap_or(ConcurrencyConfig::Adaptive {
@@ -278,6 +284,8 @@ pub(crate) fn pipeline<H: Handler>(
     //docs::/#buff
     let committer_channel_size = committer_channel_size.unwrap_or(num_cpus::get());
     let (committer_tx, watermark_rx) = mpsc::channel(committer_channel_size);
+    let pruner_channel_size = pruner_channel_size.unwrap_or(64);
+    let (pruner_tx, prune_watermark_rx) = mpsc::channel::<(u64, u64)>(pruner_channel_size);
     let main_reader_lo = Arc::new(SetOnce::new());
 
     let handler = Arc::new(handler);
@@ -330,7 +338,15 @@ pub(crate) fn pipeline<H: Handler>(
     let s_reader_watermark =
         reader_watermark::<H>(pruner_config.clone(), store.clone(), metrics.clone());
 
-    let s_pruner = pruner(handler, pruner_config, store, metrics);
+    let s_pruner = pruner(
+        handler,
+        pruner_config.clone(),
+        store.clone(),
+        pruner_tx,
+        metrics.clone(),
+    );
+
+    let s_prune_watermark = prune_watermark::<H>(pruner_config, prune_watermark_rx, store, metrics);
 
     s_processor
         .merge(s_collector)
@@ -339,6 +355,7 @@ pub(crate) fn pipeline<H: Handler>(
         .attach(s_track_reader_lo)
         .attach(s_reader_watermark)
         .attach(s_pruner)
+        .attach(s_prune_watermark)
 }
 
 #[cfg(test)]
--- a/crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
@@ -1,0 +1,406 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::BTreeMap;
+use std::sync::Arc;
+use std::time::Duration;
+
+use backoff::ExponentialBackoff;
+use sui_futures::service::Service;
+use tokio::sync::mpsc;
+use tracing::error;
+use tracing::info;
+use tracing::warn;
+
+use crate::metrics::IndexerMetrics;
+use crate::pipeline::concurrent::Handler;
+use crate::pipeline::concurrent::PrunerConfig;
+use crate::pipeline::logging::LoggerWatermark;
+use crate::pipeline::logging::WatermarkLogger;
+use crate::store::ConcurrentConnection;
+use crate::store::Store;
+
+/// If reading the initial `pruner_hi` fails, retry starting at this interval.
+const INITIAL_READ_RETRY_INTERVAL: Duration = Duration::from_millis(100);
+
+/// Cap the retry interval at this value while reading the initial `pruner_hi`.
+const MAX_READ_RETRY_INTERVAL: Duration = Duration::from_secs(1);
+
+/// The prune watermark task is responsible for advancing the pipeline's `pruner_hi` watermark in
+/// the database. It receives successfully pruned ranges `(from, to_exclusive)` from the `pruner`
+/// task over an mpsc channel, buffers out-of-order completions in a `BTreeMap`, and advances
+/// `pruner_hi` through the contiguous prefix of completed ranges.
+///
+/// Splitting watermark updates into their own task mirrors the `committer` + `commit_watermark`
+/// pattern: the pruner task focuses on distributing work, while this task focuses on persisting
+/// progress. At startup, the task reads the current DB `pruner_hi` to establish the baseline for
+/// contiguity — the first received chunk may have `from > pruner_hi` if an earlier chunk failed
+/// and is still being retried, so lazy-initialization from the first message is not safe.
+///
+/// If `config` is `None`, the task will shutdown immediately.
+pub(super) fn prune_watermark<H: Handler>(
+    config: Option<PrunerConfig>,
+    mut rx: mpsc::Receiver<(u64, u64)>,
+    store: H::Store,
+    metrics: Arc<IndexerMetrics>,
+) -> Service {
+    Service::new().spawn_aborting(async move {
+        if config.is_none() {
+            info!(pipeline = H::NAME, "Skipping prune watermark task");
+            return Ok(());
+        };
+
+        let initial_pruner_hi = read_initial_pruner_hi::<H>(&store).await;
+
+        info!(
+            pipeline = H::NAME,
+            initial_pruner_hi, "Starting prune watermark task"
+        );
+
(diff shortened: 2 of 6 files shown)


crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
```diff
@@ -0,0 +1,406 @@
...
+        info!(pipeline = H::NAME, "Stopping prune watermark task");
+        Ok(())
+    })
+}
+
+/// Read the initial `pruner_hi` from the DB, retrying with exponential backoff on connection or
+/// query failures. If the pipeline has no watermark row yet, returns 0 — in that case the pruner
+/// task will also find no watermark and produce no messages, so the value is unused until a
+/// watermark is written.
+async fn read_initial_pruner_hi<H: Handler>(store: &H::Store) -> u64 {
```
wlmyng, line 108: i think this could return `Option<u64>` - so that "0 is a safe fallback" is pushed to the callsite.

wlmyng: it's a bit confusing because we have a possible `0` result representing two scenarios:

1. 0 coming from inner representing "no watermark row"
2. But we could also get a 0 from the outer, unreachable-branch placeholder since it'll never be reached (`BE::transient`)

imo we should return the `Option` to represent, no pruner watermark, and maybe do a `.expect()` because we do expect to not get a permanent error / retry indefinitely

crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
```diff
@@ -0,0 +1,406 @@
...
+    // `Err` arm is unreachable — `unwrap_or(0)` is a safe fallback for the type system.
+    backoff::future::retry(backoff, read).await.unwrap_or(0)
+}
+
+/// Write `pruner_hi` to the database. Returns the elapsed write time on success (so the caller can
+/// pass it to the logger) and `Err(())` on connection or DB error.
+async fn write_pruner_watermark<H: Handler>(
+    store: &H::Store,
+    pruner_hi: u64,
+    metrics: &IndexerMetrics,
```
wlmyng, line 150: It doesn't seem ideal to have the fn own half the metrics, the caller own the other half, and the caller matches a `Result` manufactured by the fn solely solely to shuttle outcome information back. I think we should just keep this inlined (as was originally the case)

Or, lift everything out into the function (perhaps renamed to `advance_pruner_watermark`)

crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
```diff
@@ -0,0 +1,406 @@
...
+            "Gap between (0, 5) and (10, 15) should stall pruner_hi at 5"
+        );
+
+        // Fill the gap. Both (5, 10) and the buffered (10, 15) should flush.
+        setup.tx.send((5, 10)).await.unwrap();
+        wait_for_pruner_hi(&setup.store, 15, Duration::from_secs(1)).await;
+    }
+
+    #[tokio::test]
+    async fn test_watermark_advances_despite_connection_flakiness() {
```
wlmyng, line 330: i think we could use tokio's paused time `#[tokio::test(start_paused = true)]`

are we testing here that `read_initial_pruner_hi` retries indefinitely? I think what's happening is:

                                                                                                                             
  1. read_initial_pruner_hi first attempt → connect() fails → BE::transient → retry after 100ms.                                                                                                      
  2. read_initial_pruner_hi retry → connect() fails → retry after ~200ms. 
  3. read_initial_pruner_hi retry → connect() fails → retry after ~400ms.                                                                                                                             
  4. read_initial_pruner_hi retry → connect() succeeds, read pruner_hi = 0.                                                                                                                           
  5. Main loop processes (0, 5) → write_pruner_watermark → connect() succeeds.                                                                                                                        
  6. Main loop processes (5, 10) → write_pruner_watermark → connect() succeeds.

wlmyng: I think we need the test to allow the initial read(s) to succeed, fail the next K (assert pruner_hi doesn't advance while failures are outstanding), retry succeed and maek the expectations. I think paused time would be useful here as we can control "when" the failures occur with paused time + `wait_for_connection_attempts`

And the test could be like:
```
// currently at pruner_hi = 5
setup.tx.send((5, 10))
wait_for_connection_attempts()

setup.tx.send((10, 15))
wait_for_connection_attempts()

setup.tx.send((15, 20))
// land pruner_hi
expect a single successful write to db that pushes pruner_hi to 20
```

wlmyng (commented):
mainly comments around the tests

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
```diff
@@ -278,6 +284,8 @@ pub(crate) fn pipeline<H: Handler>(
     //docs::/#buff
     let committer_channel_size = committer_channel_size.unwrap_or(num_cpus::get());
     let (committer_tx, watermark_rx) = mpsc::channel(committer_channel_size);
+    let pruner_channel_size = pruner_channel_size.unwrap_or(64);
```
amnn, line 287: Is there a reason to use `64` here and not the same sizing as `committer_channel_size` (or to share `num_cpus` between the two of them or something?)

crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs
```diff
@@ -278,6 +284,8 @@ pub(crate) fn pipeline<H: Handler>(
...
+    let (pruner_tx, prune_watermark_rx) = mpsc::channel::<(u64, u64)>(pruner_channel_size);
```
amnn, line 288:
```suggestion
    let (pruner_tx, prune_watermark_rx) = mpsc::channel(pruner_channel_size);
```

crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        info!(
+            pipeline = H::NAME,
+            initial_pruner_hi, "Starting prune watermark task"
+        );
+
+        // Buffer completed ranges that are not yet contiguous with `highest_pruned`.
+        let mut precommitted: BTreeMap<u64, u64> = BTreeMap::new();
+        let mut highest_pruned: u64 = initial_pruner_hi;
+        let mut highest_watermarked: u64 = initial_pruner_hi;
```
amnn, line 63:
```suggestion
        let mut highest_pruned = initial_pruner_hi;
        let mut highest_watermarked = initial_pruner_hi;
```

crates/sui-indexer-alt-framework/src/pipeline/concurrent/pruner.rs
```diff
@@ -334,16 +267,36 @@ mod tests {
...
+        let (tx, rx) = mpsc::channel(16);
+        let s_pruner = pruner(
+            handler,
+            Some(config.clone()),
+            store.clone(),
+            tx,
+            metrics.clone(),
+        );
+        let s_prune_watermark = prune_watermark::<DataPipeline>(Some(config), rx, store, metrics);
+        (s_pruner, s_prune_watermark)
```
amnn, line 297: Just merge the services together.


```suggestion
    ) -> Service {
        let (tx, rx) = mpsc::channel(16);
        let s_pruner = pruner(
            handler,
            Some(config.clone()),
            store.clone(),
            tx,
            metrics.clone(),
        );
        let s_prune_watermark = prune_watermark::<DataPipeline>(Some(config), rx, store, metrics);
        s_pruner.merge(s_prune_watermark)
```

crates/sui-indexer-alt-framework/src/pipeline/concurrent/pruner.rs
amnn, line 160: Note that the pruner still has a pretty big bottleneck, because it waits for each batch of work to completely finish before it schedules the next batch to be pruned. Now that watermark updates have been moved into their own task, it should become simpler to get rid of that bottleneck, by introducing yet another task (this should probably be a follow-up, but it's well worth doing).

The line this comment is on becomes a `send` call, down a channel. On the other side is the equivalent of a pruner "processor" -- a task that is given a channel receiver, turns it into a stream, and then calls `try_for_each_spawned` or one of its variants on it. It seems sensible to use one of the adaptive concurrency variants from the start here.

The body of the `try_for_each_spawned` is responsible for doing the pruning, and handling retries. Each instance of the body is handed the channel for the pruner watermark task to send it updates.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        let mut logger = WatermarkLogger::new("pruner");
+
+        while let Some((from, to_exclusive)) = rx.recv().await {
```
amnn, line 67: Can we have this task follow the pattern of `commit_watermark`? The main difference is that `commit_watermark` rate limits writes to the DB to only happen  on a commit watermark interval (with some jitter), to smooth and thin out load the to DB, whereas currently, this task will potentially perform a write for each pruner chunk if it can.

For pruners chunks are already groupings of checkpoints, so load on the DB just from watermark updates will be lower, but it seems helpful to have fewer patterns in the codebase for doing essentially the same operation, unless we have a good reason to differentiate them.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
```diff
@@ -0,0 +1,406 @@
...
+                    pipeline = H::NAME,
+                    "Failed to read initial pruner watermark, retrying: {e}"
+                );
+                BE::transient(())
+            })
+    };
+
+    // `max_elapsed_time: None` means the operation never terminates with a permanent error, so the
+    // `Err` arm is unreachable — `unwrap_or(0)` is a safe fallback for the type system.
+    backoff::future::retry(backoff, read).await.unwrap_or(0)
```
amnn, line 142: This would be a stronger signal for what you're trying to show here -- I would not need the comment if I saw this.

```suggestion
    backoff::future::retry(backoff, read).await.expect("unexpected permanent error");
```

crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
amnn: Agreed with the confusion(s) here, and some suggestions for how to improve:

- The unreachable `unwrap_or` could be turned into an `expect`.
- If a watermark row cannot be found, this task should retry, not exit returning 0 (the `pruner` also retries if a watermark row cannot be found).

0 will no longer be a safe default once you support ingesting backwards -- and with these changes, it should no longer be a problem (the only time you would see a `0` from this system was if the pipeline really were unpruned since genesis.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
amnn: 💯 I would opt for inlining.

crates/sui-indexer-alt-framework/src/pipeline/concurrent/prune_watermark.rs
```diff
@@ -0,0 +1,406 @@
...
+            }
+            if start.elapsed() > timeout {
+                let actual = store
+                    .watermark(DataPipeline::NAME)
+                    .map(|w| w.pruner_hi)
+                    .unwrap_or(u64::MAX);
+                panic!("Timed out waiting for pruner_hi={expected}; last observed {actual}");
+            }
+            tokio::time::sleep(Duration::from_millis(10)).await;
+        }
```
amnn, line 290: It would be better to use the `timeout` and `interval` helpers from tokio rather than rebuild them from scratch yourself. Also is it safer to implement this as an inequality rather than a precise equality check on `pruner_hi`?


```suggestion
        let res = time::timeout(timeout, async move {
            let interval = time::interval(Duration::from_millis(10));
            interval.tick().await;
            while let Some(w) = store.watermark(DataPipeline::NAME)
                && w.pruner_hi < expected
            {
                interval.tick().await;
            }
        })
        .await;
        
        assert!(
            res.is_ok(), 
            "Timed out waiting for pruner_hi={expected}; last observed {:?}",
            store.watermark(DataPipeline::NAME).map(|w| w.pruner_hi),
        );
```

amnn (commented):
Main thrust of the change looks good -- left some suggestions mostly to align with what how commit waterark is implemented. Will leave it  to @wlmyng to give the final stamp once their comments are addressed -- +1 to all of them from me. Thanks!

PR #26342, opened 2026-04-21
Avoid direct disk write of RejectUserCerts in consensus handler

mystenmark:
## Summary

- Refactors `close_user_certs` so the consensus handler path (`handle_close_epoch`) transitions reconfig state in-memory only, without a direct disk write. Disk persistence happens through the commit batch via `advance_eop_state_machine` → `state.output.store_reconfig_state`.
- The consensus adapter's `close_epoch()` path retains its direct disk write (needed for `recover_end_of_publish` crash recovery).
- Extracts `record_epoch_close_time()` helper to avoid duplication.

## Motivation

In the timestamp-based epoch close path, `close_user_certs` writes `RejectUserCerts` directly to the `reconfig_state` DB column **outside** the atomic commit batch. If the validator crashes between this write and the batch flush, on restart the consensus handler loads `RejectUserCerts` into `initial_reconfig_state`. Today this is safe only because `should_accept_consensus_certs()` returns `true` for `RejectUserCerts`, so `filter_consensus_txns` still includes user txns — matching the first-run behavior. But this is a fragile invariant: any future change that gates filtering on `should_accept_user_certs` instead would silently introduce a fork-after-crash-recovery bug.

This PR eliminates the hazard by ensuring the consensus handler never writes reconfig state outside the commit batch.

## Test plan

- [x] `cargo check -p sui-core` passes
- [ ] Existing simtests covering epoch close (timestamp-based and EOP-quorum) should pass unchanged — no behavioral change, only persistence path change

### DIFF
--- a/crates/sui-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/sui-core/src/authority/authority_per_epoch_store.rs
@@ -3117,21 +3117,6 @@ impl AuthorityPerEpochStore {
         }
     }
 
-    /// Record epoch close for the timestamp-based epoch transition path.
-    /// Sets epoch_close_time for metrics and fires user_certs_closed_notify
-    /// (needed by consensus adapter for submit delay cancellation).
-    /// Does NOT modify reconfig_state — that is handled by the consensus handler directly.
-    pub fn record_epoch_close_for_timestamp_based_transition(&self) {
-        let mut epoch_close_time = self.epoch_close_time.write();
-        if epoch_close_time.is_none() {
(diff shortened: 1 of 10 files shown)


mwtian:
`RejectUserCerts` will still be (much less frequently) set in DB async to consensus handler after this change. I can understand this change for consistency in consensus handler though.

If we have the appetite for more changes around end of epoch, I would propose completely deprecate the `RejectUserCerts` variant and related logic to reject user txn submissions. Now `RejectUserCerts` is no longer needed for correctness, so removing it should simplify logic.

PR #26344, opened 2026-04-21
[jsonrpc-alt] graceful fallback when the package resolver cannot resolve an input based on MoveCall signature

wlmyng:
## Description 

`sui_getTransactionBlock` in `sui-indexer-alt-jsonrpc` currently returns `-32603 Internal Error: Failed to resolve types in transaction data: Function not found: ...` whenever the package resolver can't find a function referenced by a PTB. The whole response does not need to fail though, because raw input bytes are still available, and can be returned.

Graphql and legacy `sui-json-rpc` end up serving some version of the `PureValue { value_type: None, value: bytes }`, so we make a similar change for `jsonrpc-alt` for parity. 

## Test plan 

unit test, and locally checked that now we return results instead of -32603

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-json-rpc-types/src/sui_transaction.rs
+++ b/crates/sui-json-rpc-types/src/sui_transaction.rs
@@ -1933,6 +1933,7 @@ impl SuiProgrammableTransactionBlock {
             Ok(layouts) => layouts,
             Err(_) => vec![None; value.inputs.len()],
         };
+
         let ProgrammableTransaction { inputs, commands } = value;
         Ok(SuiProgrammableTransactionBlock {
             inputs: inputs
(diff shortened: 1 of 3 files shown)


crates/sui-json-rpc-types/src/sui_transaction.rs
```diff
@@ -1925,7 +1925,14 @@ impl SuiProgrammableTransactionBlock {
...
     ) -> Result<Self, anyhow::Error> {
-        let input_types = package_resolver.pure_input_layouts(&value).await?;
+        // If the resolver can't infer layouts (e.g. a MoveCall references a function the resolver
+        // can't find), fall back to rendering every pure input as untyped bytes rather than
+        // failing the whole conversion. Matches the legacy `sui-json-rpc` behavior and the
+        // `sui-indexer-alt-graphql` behavior at `programmable/mod.rs`.
+        let input_types = match package_resolver.pure_input_layouts(&value).await {
+            Ok(layouts) => layouts,
+            Err(_) => vec![None; value.inputs.len()],
+        };
```
amnn, line 1935: minor nit: blank line after?

amnn (approved):
LGTM, thanks @wlmyng !

amnn:
@wlmyng this still seems relevant right?

PR #26345, opened 2026-04-22
[consensus] refactor core.rs tests to use CoreTestFixture

mwtian:
## Summary
- Route 7 single-Core tests in `consensus/core/src/core.rs` through the existing `CoreTextFixture` helper, deleting ~30 lines of setup boilerplate per test. Net diff: **+306 / −531** (−225).
- Extend the fixture: expose `transaction_client`, and add a `new_with_prepopulated_blocks(...)` constructor that seeds the store before `DagState` is built — needed for recovery tests.
- Remove the `// TODO: create helper to avoid the duplicated code here.` that lived in `test_core_set_propagation_delay_per_authority`.
- Other minor cleanups.

## Refactored tests
- `test_core_set_min_propose_round`
- `test_smart_ancestor_selection`
- `test_excluded_ancestor_limit`
- `test_core_set_propagation_delay_per_authority`
- `test_multiple_commits_advance_threshold_clock`
- `test_core_recover_from_store_for_partial_round`
- `test_core_propose_after_genesis`

## Tests intentionally left on manual setup
`test_core_recover_from_store_for_full_round` and `test_commit_and_notify_for_block_status` both need to register block-status subscriptions on `TransactionConsumer` **before** `Core::new_validator` runs `recover_validator()`, which synchronously drains the subscription map via `try_commit()`. The fixture constructs Core internally, so a pre-Core subscription hook would have to flow through the constructor signature. Not worth the API clutter for two tests — leaving them inline.

## Test plan
- [x] `SUI_SKIP_SIMTESTS=1 cargo nextest run -p consensus-core --lib` — 203/203 pass
- [x] `cargo clippy -p consensus-core --all-targets` — clean

### DIFF
--- a/consensus/core/src/core.rs
+++ b/consensus/core/src/core.rs
@@ -928,20 +928,20 @@ impl CoreSignalsReceivers {
 pub(crate) async fn create_cores(
     context: Context,
     authorities: Vec<Stake>,
-) -> Vec<CoreTextFixture> {
+) -> Vec<CoreTestFixture> {
     let mut cores = Vec::new();
 
     for index in 0..authorities.len() {
         let own_index = AuthorityIndex::new_for_test(index as u32);
         let core =
-            CoreTextFixture::new(context.clone(), authorities.clone(), own_index, false).await;
+            CoreTestFixture::new(context.clone(), authorities.clone(), own_index, false).await;
         cores.push(core);
     }
     cores
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


consensus/core/src/core.rs
akichidis: I guess we should rename this to `CoreTestFixture` ?

akichidis (approved):
Nice refactoring!

PR #26346, opened 2026-04-22
Provision mode for external keys

jordanjennings-mysten:
## Description 

Adds a "provision mode" argument to the CLI which adds 3 options to how keys can be generated:
- RecoverableAssumed: default mode which is only supported by ledger signer where a backup flow is an expected part of the signer, ledgers walk you through this.
- MnemonicBacked: generate a key outside of the signer, import it and return the mnemonic to the user
- NonRecoverable: generate on device, skip backups.

Ledgers support RecoverableAssumed since they walk users through a backup flow, however yubikeys are not backed up by default on generation and do not support recoverable assumed. Yubikeys require users to provision with either mnemonicbacked or nonrecoverable to explicitly approve of the risk of on device generation.

## Test plan 

unit tests and QA

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [x] CLI: external-keys subcommand receives --provision-mode [recoverable-assumed|mnemonic-backed|non-recoverable] 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-keys/src/external.rs
+++ b/crates/sui-keys/src/external.rs
@@ -13,7 +13,8 @@ use base64;
 use base64::{Engine as _, engine::general_purpose};
 use bcs;
 use fastcrypto::traits::{EncodeDecodeBase64, VerifyingKey};
-use jsonrpc::client::Endpoint;
+use jsonrpc::client::{Endpoint, JsonRpcError};
+use jsonrpc::types::RemoteError;
 use mockall::{automock, predicate::*};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_json::{Value as JsonValue, json};
@@ -25,6 +26,26 @@ use std::process::Stdio;
 use sui_types::base_types::SuiAddress;
 use sui_types::crypto::{PublicKey, Signature, SuiKeyPair, SuiSignature, SuiSignatureInner};
 use tokio::process::Command;
+
+const PROVISION_MODE_NOT_SUPPORTED_ERROR_CODE: i32 = -32012;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ExternalExecError {
+    #[error(transparent)]
+    JsonRpc(#[from] JsonRpcError),
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+    #[error("Command failed with status: {0}")]
+    CommandFailed(std::process::ExitStatus),
+    #[error("Command returned null result")]
+    NullResult,
+}
+
+impl From<RemoteError> for ExternalExecError {
+    fn from(error: RemoteError) -> Self {
+        JsonRpcError::from(error).into()
+    }
+}
 
 #[derive(Debug)]
 /// External keystore for managing keys and aliases with an external signer.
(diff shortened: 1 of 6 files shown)


crates/sui-keys/src/external.rs
```diff
@@ -53,6 +74,30 @@ pub struct ExternalKey {
     pub key_id: String,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct CreateKeyResponse {
+    pub public_key: PublicKey,
+    pub key_id: String,
+    pub mnemonic: Option<String>,
```
stefan-mysten: This has Debug on it, so it could expose the mnemonic if by mistake this info is printed in debug-logs. Could maybe have a custom Debug on it, or skip it if it's not really need it.

crates/sui-keys/src/external.rs
```diff
@@ -53,6 +74,30 @@ pub struct ExternalKey {
...
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, Default, PartialEq, Eq)]
+/// Provision mode requested when creating a key on an external signer.
+pub enum ProvisionMode {
```
stefan-mysten: ProvisionMode serde wire format is implicit PascalCase.  
```suggestion
#[derive(Serialize, Deserialize, Debug, Clone, Copy, Default, PartialEq, Eq)]
#[serde(rename_all = "kebab-case")]
/// Provision mode requested when creating a key on an external signer.
pub enum ProvisionMode {
```

crates/sui-keys/src/external.rs
```diff
@@ -384,22 +419,61 @@ impl AccountKeystore for External {
...
+                    Err(add_error) => {
+                        let error = match provision_mode_error {
+                            Some(message) => anyhow!(
+                                "{message}. Also failed to add the first unindexed key: {add_error}"
+                            ),
+                            None => anyhow!(
+                                "Failed to create a new key on the external signer: {create_error}. Also failed to add the first unindexed key: {add_error}"
+                            ),
+                        };
+                        return Err(error);
```
stefan-mysten: I am a bit wary about this, when logic is handled in an Err branch. Could this create an unindexed key that is not desired?

crates/sui-keys/src/external.rs
```diff
@@ -53,6 +74,30 @@ pub struct ExternalKey {
     pub key_id: String,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct CreateKeyResponse {
+    pub public_key: PublicKey,
+    pub key_id: String,
```
stefan-mysten, line 83: Isn't this an `ExternalKey` - should it be reused?
```suggestion
    CreateKeyResponse { 
               #[serde(flatten)]
               key: ExternalKey,
               mnemonic: Option<String> 
   }
```

crates/sui/src/external_signer.rs
```diff
@@ -7,16 +6,40 @@ use clap::*;
...
 use serde_json::json;
-use sui_keys::external::External;
+use std::fmt::{Debug, Display, Formatter};
+use sui_keys::external::{External, ProvisionMode};
 use sui_keys::keystore::{AccountKeystore, GenerateOptions, GeneratedKey, Keystore};
 use tracing::info;
 
+#[derive(ValueEnum, Clone, Copy, Debug, PartialEq, Eq)]
+#[clap(rename_all = "kebab-case")]
+pub enum ProvisionModeArg {
```
stefan-mysten, line 16: I am confused, you cannot reuse that one in external.rs?

jordanjennings-mysten: well its a separate crate type so we cannot put the clap derives on it. otherwise you end up with a bit of parser code to convert to the type which feels less clean. either case you will end up with about the same number of lines of code.

crates/sui-keys/src/external.rs
jordanjennings-mysten: I actually had been thinking about this.. I might improve this later where its a bit more explicit but in the case with ledger its expected behavior. It becomes less expected with yubikey but I'll have to revisit this with some better signer metadata so the signers can maybe describe some more details about their provision modes.

jordanjennings-mysten: I think I'll go the safe route here only add unindex when its the error that ledger would return, making ledger the only signer which would fall back. eventually I'll add some better error codes or metadata.

crates/sui-keys/src/external.rs
jordanjennings-mysten: since mnemonic is here I'd rather keep it separate.

stefan-mysten (approved):

PR #26349, opened 2026-04-22
Fix warnings in release mode

dariorussi:
## Description 

Added `#[allow(unused)]` to the function in `graph_map.rs` as in other test only function that already existed in the file.
Added `#[cfg(debug_assertions)]` in `fn successors_idx(..)` given that function is only called from `debug_assertion` code

## Test plan 

Built and verified no warnings

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
+++ b/external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
@@ -129,6 +129,7 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
     }
 
     /// Returns the direct successors of the specified reference by NodeIndex
+    #[cfg(debug_assertions)]
     fn successors_idx(
         &self,
         r: NodeIndex,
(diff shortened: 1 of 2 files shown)


external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -129,6 +129,7 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
     }
 
     /// Returns the direct successors of the specified reference by NodeIndex
+    #[cfg(debug_assertions)]
```
tnowacki, line 132: I'd prefer allow unused

tnowacki (approved):
Is there some way in CI we can stop these changes from happening?

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
tnowacki: Though it doesn't matter since non public

dariorussi:
> Is there some way in CI we can stop these changes from happening?

you mean stop warning from happening?

PR #26350, opened 2026-04-22
Move expensive checkpoint decoding off IO threads

nickvikeras:
## Description

I was debugging throughput bottlenecks in the bigtable indexer and found that checkpoint decoding is quite expensive (about 10% of overall CPU) and the tasks are course-grained enough to warrant moving off of the tokio io threads.

```
  ┌────────────────────────────────┬─────────┬─────────┬──────────┐
  │              step              │   avg   │   p50   │   p99    │
  ├────────────────────────────────┼─────────┼─────────┼──────────┤
  │ total (zstd + prost + TryFrom) │ 3.97 ms │ 4.37 ms │ 18.82 ms │
  ├────────────────────────────────┼─────────┼─────────┼──────────┤
  │ prost decode alone             │ 0.50 ms │ 0.54 ms │ 4.71 ms  │
  ├────────────────────────────────┼─────────┼─────────┼──────────┤
  │ TryFrom (proto→native) alone   │ 2.40 ms │ 1.80 ms │ 13.72 ms │
  └────────────────────────────────┴─────────┴─────────┴──────────┘
```

This frees up the io threads to keep doing io dispatch work to keep work moving through the system.

## Test plan

I'm running it in my bigtable bitmap index backfill. This wasn't the thing that finally moved the needle, but I never tested without it after resolving the primary bottleneck. It just seems like the correct thing to do given the cost of these tasks regardless.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
@@ -56,8 +56,12 @@ impl IngestionClientTrait for RpcClient {
         // `ByteCountMakeCallbackHandler` request layer attached in
         // `IngestionClient::with_grpc`, so it does not need to be tracked here.
         let response = response.into_inner();
-        Checkpoint::try_from(response.checkpoint())
-            .map_err(|e| CheckpointError::Decode(ProtoConversion(e)))
+        tokio::task::spawn_blocking(move || {
+            Checkpoint::try_from(response.checkpoint())
+                .map_err(|e| CheckpointError::Decode(ProtoConversion(e)))
+        })
+        .await
+        .map_err(|e| CheckpointError::Fetch(anyhow!("decode task panicked: {e}")))?
     }
 
     async fn latest_checkpoint_number(&self) -> anyhow::Result<u64> {
--- a/crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
@@ -56,6 +56,8 @@ impl IngestionClientTrait for RpcClient {
         // `ByteCountMakeCallbackHandler` request layer attached in
         // `IngestionClient::with_grpc`, so it does not need to be tracked here.
         let response = response.into_inner();
+        // Proto -> Checkpoint conversion is multi-ms of CPU work; offload to the
+        // blocking pool so it doesn't stall the reactor.
         tokio::task::spawn_blocking(move || {
             Checkpoint::try_from(response.checkpoint())
                 .map_err(|e| CheckpointError::Decode(ProtoConversion(e)))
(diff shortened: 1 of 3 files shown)


crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
```diff
@@ -56,8 +56,12 @@ impl IngestionClientTrait for RpcClient {
         // `ByteCountMakeCallbackHandler` request layer attached in
         // `IngestionClient::with_grpc`, so it does not need to be tracked here.
         let response = response.into_inner();
-        Checkpoint::try_from(response.checkpoint())
-            .map_err(|e| CheckpointError::Decode(ProtoConversion(e)))
+        tokio::task::spawn_blocking(move || {
```
wlmyng, line 61: worth a comment about sending it to blocking pool?

wlmyng (approved):
I think we could consider pushing `spawn_blocking` into the `decode::checkpoint`? Or a helper that does the `spawn_blocking`? Would make it cleaner at callsites. However, this optimization doesn't seem to be a necessity - but worth adding a comment to why we are doing this

nickvikeras:
> I think we could consider pushing `spawn_blocking` into the `decode::checkpoint`? Or a helper that does the `spawn_blocking`? Would make it cleaner at callsites. However, this optimization doesn't seem to be a necessity - but worth adding a comment to why we are doing this

I initially tried putting it in the decode function but it was kind of awkward because the store client does decompression _and_ decoding and you want all of that to happen in one task ideally, so you'd need different helper methods for rpc and store client paths, and that seemed like overkill.

Adding comments is a good call cuz it's not obvious this is expensive.

crates/sui-indexer-alt-framework/src/ingestion/rpc_client.rs
evan-wall-mysten: It would be nice to explain that it isn't being moved for performance, but to make sure that it allows scheduled tasks run on time and prevent occasional spikes in measured latency/lag.

amnn (commented):
Just a note that [this article](https://ryhl.io/blog/async-what-is-blocking/) is quite a good resource for thinking about this kind of problem. 

TL;DR `spawn_blocking` is mostly useful for offloading blocking I/O, as its thread pool is quite large (100s of threads), and less good at offloading CPU intensive work. `rayon` is better for that use case (even if you are not using its helpers to auto-parallelize iterative/streaming operations), because its thread pool **is** tuned for CPU intensive work.

This may not be a huge deal for us, because adaptive concurrency limits may help to choke the inflow of checkpoints and prevent us from creating a lot of scheduler pressure, but just wanted to flag this, because using `rayon` may be a relatively simple way to protect the rest of the threads/processes on the system from the consequences of the a lot (100s) of heavy-to-deserialize checkpoints arriving at once.

PR #26353, opened 2026-04-22
[sui-adapter] Add type linkage caching as well as type conversion and loaded function caches

tzakian:
## Description 

This adds per-transaction caches to the PTB layer that cache type conversions that are done, as well as type linkage caches and loaded functions.

## Test plan 

Make sure existing tests + local benchmarking 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
+++ b/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
@@ -6,7 +6,9 @@ use crate::static_programmable_transactions::{
     loading::ast::{LoadedFunction, Type},
 };
 use indexmap::IndexSet;
-use move_core_types::{account_address::AccountAddress, language_storage::TypeTag};
+use move_core_types::{
+    account_address::AccountAddress, identifier::IdentStr, language_storage::TypeTag,
+};
 use move_vm_runtime::execution as vm_runtime;
 use std::{
     cell::RefCell,
@@ -86,6 +88,8 @@ struct PerTxCache_ {
     /// the version-specific package ID because private/entry functions can disappear across
     /// package versions, so different versions resolve differently.
     function_cache: HashMap<LoadedFunctionKey, Rc<LoadedFunction>>,
+
+    defining_id_map: HashMap<(ObjectID, Identifier, Identifier), ObjectID>,
 }
 
 /// Early-returns `Ok($empty)` from the enclosing method when PTB caching is disabled. The second
@@ -112,6 +116,7 @@ impl<'pc> PerTxCache<'pc> {
                 vm_to_type: HashMap::new(),
                 type_to_vm: HashMap::new(),
                 function_cache: HashMap::new(),
+                defining_id_map: HashMap::new(),
             }),
         }
     }
@@ -278,6 +283,42 @@ impl<'pc> PerTxCache<'pc> {
             "duplicate insert into function_cache"
         );
         Ok(())
+    }
+
+    pub(super) fn insert_type_to_defining_id(
+        &self,
+        package: ObjectID,
+        module: &IdentStr,
+        name: &IdentStr,
+        defining_id: ObjectID,
+    ) -> Result<(), ExecutionError> {
+        gated!(self.protocol_config, ());
+        let previous = self
+            .borrow_mut()?
+            .defining_id_map
+            .insert((package, module.to_owned(), name.to_owned()), defining_id);
+        assert_invariant!(
+            previous.is_none(),
+            "duplicate insert into defining_id_map: ({}::{}::{})",
+            package,
+            module,
+            name
+        );
+        Ok(())
+    }
+
+    pub(super) fn lookup_type_to_defining_id(
+        &self,
+        package: ObjectID,
+        module: &IdentStr,
+        name: &IdentStr,
+    ) -> Result<Option<ObjectID>, ExecutionError> {
+        gated!(self.protocol_config, None);
+        Ok(self
+            .borrow()?
+            .defining_id_map
+            .get(&(package, module.to_owned(), name.to_owned()))
+            .cloned())
     }
 }
 
--- a/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
+++ b/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
@@ -150,7 +150,7 @@ impl<'pc> PerTxCache<'pc> {
         Ok(())
     }
 
-    pub(super) fn lookup_type_by_input(
+    pub(super) fn lookup_type_input(
         &self,
         input: &TypeInput,
     ) -> Result<Option<Type>, ExecutionError> {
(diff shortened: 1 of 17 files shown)


sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
```diff
@@ -0,0 +1,337 @@
...
+        &self,
+        key: TypeLinkageCacheKey,
+        linkage: ExecutableLinkage,
+    ) -> Result<(), ExecutionError> {
+        gated!(self.protocol_config, ());
+        self.borrow_mut()?.type_linkage_cache.insert(key, linkage);
+        Ok(())
+    }
+
+    pub(super) fn lookup_type_by_input(
```
tnowacki:
```suggestion
    pub(super) fn lookup_type_input(
```
?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
```diff
@@ -0,0 +1,337 @@
...
+        &self,
+        input: &TypeInput,
+    ) -> Result<Option<Type>, ExecutionError> {
+        gated!(self.protocol_config, None);
+        Ok(self.borrow()?.type_input_to_type.get(input).cloned())
+    }
+
+    pub(super) fn insert_type_input(
```
tnowacki, line 339:
```suggestion
    pub(super) fn insert_type_by_input(
```
or?

I'm a bit confused by the `by` or not

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
```diff
@@ -0,0 +1,337 @@
...
+    ) -> Result<(Rc<vm_runtime::Type>, Type), ExecutionError> {
+        let vm_type = Rc::new(vm_type);
+        gated!(self.protocol_config, (vm_type, ty));
+        assert_invariant!(
+            !matches!(vm_type.as_ref(), vm_runtime::Type::TyParam(_)),
+            "cannot cache unresolved TyParam: {:?}",
+            vm_type
+        );
+        let mut c = self.borrow_mut()?;
+        if let Some(existing) = c.vm_to_type.get(&vm_type) {
```
tnowacki: Should we just assert that this is None?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
```diff
@@ -0,0 +1,337 @@
...
+        if let Some(existing) = c.type_to_vm.get(&ty) {
+            assert_invariant!(
+                existing == &vm_type,
+                "cache inconsistency on type_to_vm: existing {:?} vs new {:?}",
+                existing,
+                vm_type
+            );
+        }
+        c.vm_to_type.insert(vm_type.clone(), ty.clone());
+        c.type_to_vm.insert(ty.clone(), vm_type.clone());
```
tnowacki: How common is the overwrite here?
Should we gate here to where it is not already there to minimize clones?

tnowacki: Oh they are all Rcs? Less of a concern then

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    pub(super) fn insert_function(
+        &self,
+        key: LoadedFunctionKey,
+        function: Rc<LoadedFunction>,
+    ) -> Result<(), ExecutionError> {
+        gated!(self.protocol_config, ());
+        let mut c = self.borrow_mut()?;
+        if let Some(existing) = c.function_cache.get(&key) {
```
tnowacki: Should this be None always?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
```diff
@@ -0,0 +1,337 @@
...
+    ) -> Result<(Rc<TypeTag>, Type), ExecutionError> {
+        let tag = Rc::new(tag);
+        gated!(self.protocol_config, (tag, ty));
+        assert_invariant!(
+            !matches!(ty, Type::Reference(_, _)),
+            "TypeTag cannot represent a reference type: {:?}",
+            ty
+        );
+        let mut c = self.borrow_mut()?;
+        if let Some(existing) = c.tag_to_type.get(&tag) {
```
tnowacki: Should this be none always?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
tzakian: I like your suggestion of `insert_type_input`/`lookup_type_input` -- will do that rename.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
tzakian: Yep. here and elsewhere.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
tzakian: we should never overwrite. In the update if we double-insert it's an invariant violation.

But yea, all `Rc`s so clones are cheap.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
tzakian: for the `by` or not is if there are multiple lookups e.g., `lookup_vm_type_by_{tab, vm_type}` but that is the only one that should have this.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
```diff
@@ -0,0 +1,306 @@
...
+                type_to_tag: HashMap::new(),
+                vm_to_type: HashMap::new(),
+                type_to_vm: HashMap::new(),
+                function_cache: HashMap::new(),
+            }),
+        }
+    }
+
+    fn borrow(&self) -> Result<std::cell::Ref<'_, PerTxCache_>, ExecutionError> {
+        self.inner.try_borrow().map_err(|_| {
```
cgswords, line 160: I am a little confused about `try_borrow` here (and `try_borrow_mut` below). If these are single-threaded, this failing means something went actually insane, right? Is that not a panic-level problem?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
```diff
@@ -0,0 +1,306 @@
...
+use move_vm_runtime::execution as vm_runtime;
+use std::{
+    cell::RefCell,
+    collections::{BTreeSet, HashMap},
+    rc::Rc,
+};
+use sui_protocol_config::ProtocolConfig;
+use sui_types::{Identifier, base_types::ObjectID, error::ExecutionError, type_input::TypeInput};
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
```
cgswords, line 33: In the name of type reuse, can we use the linkage hash from the move runtime?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
tzakian: It is not :) (/this is in keeping with other `try_borrow`-> invariant violations that we have.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env/cache.rs
tzakian: We cannot/shouldn't. The `LinkageHash` represents a fully resolved linkage. A `TypeLinkageCacheKey` represents the roots of a linkage, and the value this maps to is the result of computing the linkage as defined by those roots.

PR #26356, opened 2026-04-22
[forking 12/n] Design doc for owned objects and how to `keep track` of them on disk

stefan-mysten:
## Description 

Co-designed with Claude & Codex how to handle owned objects & balances. It's not clear to me how deleted objects work in the real network. Perhaps `latest` and `live` should just merge into one marker file, but not sure.
Balances also seem to require various things that I do not understand yet. Will probably iterate on this more.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-forking/owned_objects_design.md
+++ b/crates/sui-forking/owned_objects_design.md
@@ -1,0 +1,442 @@
+# Owned Objects and Balances for `sui-forking`
+
+## Context
+
+`crates/sui-forking` currently persists:
+
+- object versions under `objects/<object_id>/<version>`
+- a `latest` file per object directory
+- transactions, effects, events, and checkpoints
+
+This is enough for object-by-id reads, but it is not enough to support:
+
+- `SimulatorStore::owned_objects`
+- RPC `list_owned_objects`
+- RPC `get_balance`
+- RPC `list_balances`
+
+There is one current limitation in the existing implementation:
+
+1. `update_objects()` ignores deleted objects.
+   New versions are written, but there is no persisted representation of an object leaving the live
+   set, so the store cannot reconstruct current ownership or balances correctly.
+
+This document proposes a design for owned objects and balances for data created locally after the
+fork starts, while fitting the filesystem model that already exists today.
+
+## Goals
+
+- Support owned object enumeration for locally executed transactions after the fork starts.
+- Support coin balance queries for locally executed transactions after the fork starts.
+- Reuse the current object-version filesystem layout instead of introducing a new primary store.
+- Keep object files as the source of truth and treat ownership/balance indexes as derived state.
+- Make it possible to rebuild indexes from disk on startup.
+- Match the ordering and query shape expected by RPC as closely as practical.
+
+## Non-goals
+
+- Full historical owned-object queries at arbitrary checkpoints.
+- Full pre-fork address inventory enumeration without explicit seeding or local materialization.
+- Replacing the current object cache layout.
+
+## High-level approach
+
+Treat owned objects and balances as derived live indexes over the locally materialized live object
+set.
+
+The source of truth remains:
+
+- `objects/<object_id>/<version>` for object contents
+- per-object live metadata that identifies which version, if any, is currently live
+
+The derived state is maintained in memory:
+
+- an owner index for address-owned objects
+- a balance index keyed by `(owner, coin_type)`
+
+If the process restarts, the derived state is rebuilt by scanning the live object set on disk.
+
+## Proposed filesystem changes
+
+### Object layout
+
+Keep:
+
+- `objects/<object_id>/<version>`: BCS-encoded `Object`
+
+Add:
+
+- `objects/<object_id>/live`: text file containing the currently live version
+
+Semantics:
+
+- If `live` exists, the object is currently live at that version.
+- If `live` does not exist, the object is not in the live set.
+- Historical versions remain on disk even after deletion or wrapping.
+
+### Why add `live` instead of reusing `latest`
+
+`latest` today tracks the highest version written to disk (via `std::cmp::max`), which is correct
+for resolving the current version of an object. However, it cannot represent an object leaving
+the live set: deletions and wrapping need a durable way to say "this object has no live version",
+and `latest` always points to some version.
+
+We should either:
+
+- replace `latest` with `live`, or
+- keep `latest` for cache/debug purposes and introduce `live` for correctness
+
+This design assumes `live` is the authoritative live marker.
+
+## Alternative: filesystem-only live scans
+
+The simplest alternative is to use the filesystem as both the source of truth and the query path.
+
+Under this model:
+
+- object versions are still stored under `objects/<object_id>/<version>`
+- liveness is still represented by `objects/<object_id>/live`
+- there is no in-memory owner index
+- there is no in-memory balance index
+
+Instead, each owned-object or balance query scans the live object set on demand:
+
+1. iterate `objects/*/live`
+2. load the referenced live object
+3. inspect its owner and type
+4. for balance queries, inspect whether it is a `Coin<T>` and extract its value
+5. sort and paginate the filtered results as needed
+
+The write path is smaller in this design:
+
+- write object versions
+- update `live` markers
+- handle deletions by removing `live`
+
+No additional in-process state needs to be updated.
+
+## In-memory indexes
+
+### Owner index
+
+Maintain an in-memory `BTreeMap<OwnerIndexKey, SequenceNumber>`.
+
+Suggested key:
+
+```rust
+struct OwnerIndexKey {
+    owner: SuiAddress,
+    object_type: StructTag,
+    inverted_balance: Option<u64>,
+    object_id: ObjectID,
+}
+```
+
+This follows the same shape as the RPC index in `sui-core`:
+
+- owner first
+- then type
+- then descending balance for coin-like objects via `!balance`
+- then object id as a stable tie-breaker
+
+This is enough to support:
+
+- `SimulatorStore::owned_objects(owner)`
+- RPC `list_owned_objects`
+- owner/type filtering
+- stable cursor pagination
+
+Only index:
+
+- `Owner::AddressOwner`
+- `Owner::ConsensusAddressOwner` normalized as address-owned
+
+Do not index:
+
+- `Owner::ObjectOwner`
+- `Owner::Shared`
+- `Owner::Immutable`
+
+Packages should also be excluded from this index, because RPC owned-object pagination expects a
+`StructTag`.
+
+### Balance index
+
+Maintain an in-memory `BTreeMap<(SuiAddress, StructTag), i128>`.
+
+Use `i128` internally so updates can apply positive and negative deltas naturally.
+Clamp to `u64` when returning RPC-facing `BalanceInfo`.
+
+This is enough to support:
+
+- RPC `get_balance`
+- RPC `list_balances`
+
+For `Coin<T>` objects:
+
+- add the balance when a live address-owned coin enters the live set
+- subtract the balance when a live address-owned coin leaves the live set
+
+## What counts toward balances
+
+### Phase 1
+
+Support balance aggregation from address-owned `Coin<T>` objects only.
+
+This is the smallest useful feature and aligns with the owned-object index.
+
+### Phase 2
+
+Optionally extend support to address-balance accumulator objects.
+
+RPC v2 renders balances as:
+
+- total indexed balance
+- optionally split into coin balance and address balance
+
+The current fullnode implementation derives address balance separately from accumulator objects.
+If we want exact parity for protocols using address-balance gas payments, we should also track
+those accumulator-backed values.
+
+This can be added independently of the owner index.
+
+## Write path for local execution
+
+The core rule is:
+
+- object files are written first
+- live metadata and in-memory indexes are updated from transaction effects
+
+For every locally executed transaction, `update_objects()` should process both deleted and written
+objects.
+
+### Deleted / wrapped objects
+
+For each `(object_id, old_version, _)` in `deleted_objects`:
+
+1. Load the old live object from `objects/<id>/live`.
+2. Remove the object's contribution from the owner index if it was address-owned.
+3. Remove the object's contribution from the balance index if it was an address-owned coin.
+4. Remove `objects/<id>/live`.
+
+This should also apply to wrapped objects if they appear in the deleted set exposed by effects.
+
+### Written objects
+
+For each written object:
+
+1. If the object already has a live version, load that old live object and remove its old index
+   contributions first.
+2. Write `objects/<id>/<new_version>`.
+3. Write `objects/<id>/live = <new_version>`.
+4. Add the new owner's contribution to the owner index if the object is address-owned.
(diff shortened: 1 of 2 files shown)


crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,442 @@
...
+for resolving the current version of an object. However, it cannot represent an object leaving
+the live set: deletions and wrapping need a durable way to say "this object has no live version",
+and `latest` always points to some version.
+
+We should either:
+
+- replace `latest` with `live`, or
+- keep `latest` for cache/debug purposes and introduce `live` for correctness
+
+This design assumes `live` is the authoritative live marker.
```
jordanjennings-mysten: so live is an empty file? I'd probably just bundle them together with a simple format 
`<live>,<version>`

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+Cons:
+
+- more implementation work than filesystem-only scans
+- local writes must update the derived indexes correctly
+- restart requires a rebuild pass over the live set
+
+This is the best tradeoff if we expect repeated RPC-style queries and want reasonable performance
+without introducing a second durable store.
+
+### Option C: durable on-disk secondary indexes
```
jordanjennings-mysten: I think indexes should live on disk just because who knows how large a fork will get potentially making rebuilds very costly. If something gets corrupted maybe give some escape hatch to rebuild indexes. The main reason I would lean this direction is that you can put a simple cache on top and get the performance you want while also making that cache more agnostic to the datastore type (db/fs/etc)

stefan-mysten: I agree with the 2nd part, but I am not too worried about making rebuilds very costly. Reading 100k files in parallel would take a couple of seconds, which is acceptable at startup. Having a set with 100k live objects is already on the larger side of a forked network, and wouldn't expect it to be common.

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+
+Semantics:
+
+- If `live` exists, the object is currently live at that version.
+- If `live` does not exist, the object is not in the live set.
+- Historical versions remain on disk even after deletion or wrapping.
+
+### Why add `live` instead of reusing `latest`
+
+`latest` today tracks the highest version written to disk (via `std::cmp::max`), which is correct
```
mdgeorge4153: If I understand the proposal correctly, if I have a live object, and I wrap it, the `latest` file points to an old version of the object, which is no longer valid information. My instinct says that turning live / latest into an enumeration (matching the state that an object can be in - owned, shared, wrapped, deleted) makes more sense. I guess this is your first option below

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+
+The source of truth remains:
+
+- `objects/<object_id>/<version>` for object contents
+- per-object metadata identifying which version, if any, is currently live
+
+The derived state is maintained in memory:
+
+- an owner index for address-owned objects
+- a balance index keyed by `(owner, coin_type)`
```
mdgeorge4153: I don't see where the underlying data is stored - I don't think you can derive it from the object set, right?

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+We should either:
+
+- replace `latest` with `live`, or
+- keep `latest` for cache/debug purposes and introduce `live` for correctness
+
+This design treats the `live` file as the authoritative liveness marker.
+
+## In-memory indexes
+
+The owner and balance indexes live inside `DataStore` as plain fields. `DataStore` derives `Clone`,
```
mdgeorge4153: Curious about why `DataStore` has clone

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+- add the balance when a live address-owned coin enters the live set
+- subtract the balance when a live address-owned coin leaves the live set
+
+## What counts toward balances
+
+### Phase 1
+
+Support balance aggregation from address-owned `Coin<T>` objects only.
+
+This is the smallest useful feature and aligns with the owned-object index.
```
mdgeorge4153: I'm not sure I follow - if you don't implement address balances, won't you get different query results than you would from a "real" RPC?

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+
+### Phase 2
+
+Optionally extend support to address-balance accumulator objects.
+
+RPC v2 renders balances as:
```
mdgeorge4153: is RPC v2 the same as gRPC?

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+those accumulator-backed values.
+
+This can be added independently of the owner index.
+
+## Write path for local execution
+
+The core rule is:
+
+- object files are written first
+- live metadata is updated atomically via write-to-temp + rename
```
mdgeorge4153: It seems like there's some concurrency control here but not enough - I assume the write-to-temp +rename strategy is to handle the case where the server dies, but it also seems like an inconsistent state if the object files are written without updating the live set.

mdgeorge4153: In particular, since your goal is to execute transactions sequentially, I wonder if it would be more straightforward to do write-to-temp + rename for the entire data store at transaction boundaries instead of just doing it for the live metadata. I think it would simplify a lot of things and I don't think performance is a big concern.

crates/sui-forking/owned_objects_design.md
mdgeorge4153: I wonder if we can make the assumptions about the size and durability of forked networks explicit and put them in the design documentation.

crates/sui-forking/owned_objects_design.md
mdgeorge4153: And if we are concerned about performance I bet there's some nice way to do COW to make duplicating the data store cheap.

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+
+Pros:
+
+- fast query path without a startup rebuild
+- avoids repeated full-store scans
+
+Cons:
+
+- highest complexity
+- requires keeping multiple durable structures in sync
```
mdgeorge4153: If we do want to go this route, this is another point in favor of atomically updating the entire data store.

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+
+## Recommended direction
+
+The recommended path is Option B:
+
+- use the filesystem as the durable source of truth
+- add `live` markers for correct liveness
+- maintain rebuildable in-memory owner and balance indexes for query speed
+
+If we want a smaller first milestone, we can land Option A first and treat Option B as a follow-up
```
mdgeorge4153: I don't think option A will actually help with option B, so I don't think this path makes sense. I'd just go for option B

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+Cons:
+
+- highest complexity
+- requires keeping multiple durable structures in sync
+- schema evolution becomes more expensive while the feature is still changing
+
+This is likely premature for the first version. It only becomes attractive if startup rebuild time
+or in-memory index size turns into a practical issue.
+
+## Recommended direction
```
mdgeorge4153: It occurs to me that you're reinventing a database here. I wonder if life would be simpler if you just used a simple in-memory database like bdb or sqlite or something instead of reinventing the wheel. Then you can get atomicity and durability for free and just focus on the schema.

mdgeorge4153: I'm not familiar with the fullnode internals, but you may even be able to reuse some of its implementation for this. That might be a big win from a maintainability standpoint if you can get it to work.

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+### RPC `list_balances`
+
+Iterate the balance map over the `(owner, coin_type)` prefix.
+
+## Pre-fork data and seeding
+
+This design is intentionally scoped to local post-fork execution and locally materialized objects.
+
+Pre-fork remote fetches (e.g. `get_latest_object` pulling from a remote fullnode) must NOT create
+`live` markers. Remote fetches populate version files for use as execution inputs, but they do not
```
mdgeorge4153: This is quite surprising to me - I would expect that if I fork a network and specify an address to seed, and then immediately query the owned objects for that address, that I'd get the same answer as if I did the same query on the real network. If I understand this correctly, I'd instead get the empty set.

mdgeorge4153: It also seems like it would give you an inconsistent state, because some objects would be marked as owned but not present in the queries. I may not understand the live object set well enough.

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+
+6. If we choose Option B, implement in-memory balance index:
+   - back RPC `get_balance`
+   - back RPC `list_balances`
+
+7. Optionally add address-balance accumulator support.
+
+## Open questions
+
+1. Should `latest` be removed entirely and replaced by `live`, or should both coexist?
```
mdgeorge4153: I'd vote yes, but maybe call it `status` and make it an enum.

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,435 @@
...
+2. Do we want balance support to include address-balance accumulators in the first version?
```
mdgeorge4153: This seems important, especially if we're transitioning towards using address balances for gas.

crates/sui-forking/owned_objects_design.md
jordanjennings-mysten: maybe you can treat the checkpoint as a WAL and dump the entire checkpoint to a file first.

crates/sui-forking/owned_objects_design.md
stefan-mysten: This requires a bit more study from my side to understand how it works and draft some requirements. As of now, simulacrum does not support address balances (and not sure if it's required on that end in any way).

crates/sui-forking/owned_objects_design.md
stefan-mysten: I went with a different approach and added a `deleted` marker to indicate if an object was deleted.

crates/sui-forking/owned_objects_design.md
stefan-mysten: Seeding was not part of this design. At startup, seeding will populate the owned object index (id, version, maybe struct type for the object) to ensure one can lazily fetch the objects owned by this address when requested.

crates/sui-forking/owned_objects_design.md
stefan-mysten: I'd like to avoid using a database for now. We can revisit this and it can be easily reworked and added if deemed necessary.

The atomicity issue is required only in two three places, and should be just fine.

crates/sui-forking/owned_objects_design.md
stefan-mysten: Reworked everything.

crates/sui-forking/owned_objects_design.md
stefan-mysten: gRPC API is versioned. Latest version for our gRPC is v2 (see `crates/sui-rpc-api/src/grpc/v2`). https://google.aip.dev/185.

crates/sui-forking/owned_objects_design.md
stefan-mysten: Very likely, yes. I need more time to understand how this works and what it's required to support it.

crates/sui-forking/owned_objects_design.md
stefan-mysten: I don't think it should but for some reason it does now, I'll look into fixing it.

crates/sui-forking/owned_objects_design.md
stefan-mysten: Fixed now.

crates/sui-forking/owned_objects_design.md
stefan-mysten: Maybe that's a good thing to explain what it's best used for, which is not a long lived network a la real network.

jordanjennings-mysten (approved):
I think your simulacrum write lock makes sure each transaction is sequential so you cannot create inconsistencies with your object store by writing a delete and then directly after using that same object you deleted. Reads I'm curious about since I'd guess you might not be waiting for that lock on reads? I'd have to look though just something to think about.

stefan-mysten:
> simulacrum write lock makes sure each transaction is sequential so you cannot create inconsistencies with your object store by writing a delete and then directly after using that same object you deleted

Yes, it shouldn't be the case.

> Reads I'm curious about since I'd guess you might not be waiting for that lock on reads? I'd have to look though just something to think about.

Maybe I am missing something, but I'd say that reads are fine. If a read is triggered while a write is in process, then the "old" data is retrieved which should be fine. Once a transaction is executed and tx data and checkpoint is saved to disk, the lock is released and a subsequent read will retrieve the new data.
By design `RwLock` allows only one writer at a time, and multiple readers. If we want to enforce order, I guess we could switch to a Mutex and then they'd have to wait for each other.

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,144 @@
...
+
+Local execution can now add:
+
+```text
+objects/
+  <object_id>/
+    deleted
+```
+
+The `deleted` file is an explicit local tombstone. When present, current-object
```
amnn, line 48: It might be simpler to always have a `latest` file, and then have that file contain a sigil if the object was deleted or wrapped, so that read path always reads `latest`, and then can decide based on its contents whether to issue a subsequent read request, or return an empty response. Otherwise you need to perform an extra interaction with the filesystem for each object read.

crates/sui-forking/owned_objects_design.md
```diff
@@ -0,0 +1,144 @@
...
+`objects/<object_id>/<version>`.
+
+The owned-object index is stored separately:
+
+```text
+indices/
+  owned_objects
+```
+
+`indices/owned_objects` is a BCS-encoded `Vec<OwnedObjectEntry>` sorted by
```
amnn, line 60: I would personally just use a RocksDB map for this... it sounds like overkill, but the point is that it is a prebuilt component that handles ordered lookups for you, and we have existing abstractions for working with them.

crates/sui-forking/owned_objects_design.md
stefan-mysten: Pushed up a PR here: https://github.com/MystenLabs/sui/pull/26672

PR #26358, opened 2026-04-23
Eager loop in sequential pipelines

nickvikeras:
## Description

I applied the throughput/scheduling learnings from the collector loop in the concurrent pipeline code to the sequential pipeline collector loop. 

I was worried that since the pending buffer is unbounded, there could be a failure mode with the greedy loop where we never actually flush and just oom, so I added a MAX_PENDING_ROWS to the sequential handler which acts a soft-cap to force a flush attempt rather than a hard limit. This prevents the loop from collecting too many rows without adding a deadlock.

## Test plan

Backfilled bigtable

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [x] Indexing Framework: Added MAX_PENDING_ROWS to sequential pipelines. It's a soft-cap that forces the batching loop to attempt a flush to the database, not a hard-cap on the amount of rows buffered in memory.


jessiemongeon1:
## Style Guide Audit

**19** violation(s) found. All must be fixed before merge.

### `docs/content/develop/accessing-data/custom-indexer/pipeline-architecture.mdx` (19 violation(s))

_19 violation(s) (0 regex, 19 claude)_

- **Line 8** — Passive voice
  - Current: `Each checkpoint is fully committed before the next one`
  - Fix: `The pipeline fully commits each checkpoint before processing the next one`
- **Line 10** — Passive voice
  - Current: `This allows you to process multiple checkpoints simultaneously for higher throughput, but requires reads to check which data is fully committed to ens`
  - Fix: `This allows you to process multiple checkpoints simultaneously for higher throughput, but your reads must check which data the pipeline fully committe`
- **Line 14** — Passive voice
  - Current: `While sequential pipelines have throughput limitations compared to concurrent, base your decision primarily on engineering complexity rather than perf`
  - Fix: `While sequential pipelines have throughput limitations compared to concurrent pipelines, base your decision primarily on engineering complexity rather`
- **Line 91** — Passive voice
  - Current: `These watermarks work together to enable safe out-of-order processing, automatic data cleanup, and recovery from failures.`
  - Fix: `These watermarks work together and enable safe out-of-order processing, automatic data cleanup, and recovery from failures.`
- **Line 96** — Passive voice
  - Current: `Both concurrent and sequential pipelines rely on `checkpoint_hi_inclusive` to understand where to resume processing on restarts.`
  - Fix: `Both concurrent and sequential pipelines use `checkpoint_hi_inclusive` to determine where to resume processing on restarts.`
- **Line 98** — Passive voice
  - Current: `Optionally, the pipeline tracks `reader_lo` and `pruner_hi`, which define safe lower bounds for reading and pruning operations, if pruning is enabled.`
  - Fix: `Optionally, the pipeline tracks `reader_lo` and `pruner_hi`, which define safe lower bounds for reading and pruning operations when you enable pruning`
- **Line 105** — Passive voice
  - Current: `The pipeline frequently cleans unpruned checkpoints to ensure storage does not grow indefinitely while maintaining the retention guarantee.`
  - Fix: `The pipeline frequently cleans unpruned checkpoints to ensure storage does not grow indefinitely while the system maintains the retention guarantee.`
- **Line 182** — Passive voice
  - Current: `A separate reader watermark update task (running periodically, configurable) advances `reader_lo` to 1001`
  - Fix: `A separate reader watermark update task (which runs periodically and is configurable) advances `reader_lo` to 1001`
- **Line 196** — Passive voice
  - Current: `When `pruner_hi` (1000) < `reader_lo` (1001), the pruner detects checkpoints outside the retention window, cleans up all elements up to `reader_lo` (d`
  - Fix: `When `pruner_hi` (1000) < `reader_lo` (1001), the pruner detects checkpoints outside the retention window, cleans up all elements up to `reader_lo` (w`
- **Line 218** — Passive voice
  - Current: `The `Processor` component is described in detail in the [Common pipeline components](#processor) section.`
  - Fix: `The [Common pipeline components](#processor) section describes the `Processor` component in detail.`
- **Line 245** — Passive voice
  - Current: ``MAX_PENDING_ROWS` is a soft cap: when exceeded, the `Collector` yields from eagerly draining its input channel to flush what it can.`
  - Fix: ``MAX_PENDING_ROWS` is a soft cap: when the collector exceeds it, the `Collector` yields from eagerly draining its input channel to flush what it can.`
- **Line 263** — Passive voice
  - Current: `Sequential pipelines use the same bounded-channel backpressure model as concurrent pipelines`
  - Fix: `Sequential pipelines use the bounded-channel backpressure model that concurrent pipelines use`
- **Line 327** — Passive voice
  - Current: `Defaults to `max(num_cpus / 2, 4)` when `None`.`
  - Fix: `The system defaults to `max(num_cpus / 2, 4)` when you set this to `None`.`
- **Line 330** — Passive voice
  - Current: `Unlike the concurrent pipeline's `MAX_PENDING_ROWS`, this is not a hard backpressure gate.`
  - Fix: `Unlike the concurrent pipeline's `MAX_PENDING_ROWS`, this does not act as a hard backpressure gate.`
- **Line 331** — Passive voice
  - Current: `Receive is never blocked on it, because a missing predecessor for the next commit might still be sitting in the input channel.`
  - Fix: `The system never blocks receive on it because a missing predecessor for the next commit might still be sitting in the input channel.`
- **Line 372** — Passive voice
  - Current: `Database writes are expensive; batching dramatically improves throughput by reducing the number of database round trips.`
  - Fix: `Database writes are expensive. Batching dramatically improves throughput by reducing the number of database round trips.`
- **Line 408** — Passive voice
  - Current: `This component enforces the critical rule that `checkpoint_hi_inclusive` can advance only when all data up to that point is committed with no gaps.`
  - Fix: `This component enforces the critical rule that `checkpoint_hi_inclusive` can advance only when the system commits all data up to that point with no ga`
- **Line 466** — Passive voice
  - Current: `It is critical for recovery, as the framework reads this table to resume from the correct checkpoint.`
  - Fix: `This is critical for recovery because the framework reads this table to resume from the correct checkpoint.`
- **Line 534** — Passive voice
  - Current: `Database becomes slow (because of high load, maintenance, or similar) and can now only handle 50 commits per second.`
  - Fix: `Database becomes slow (because of high load, maintenance, or similar reasons) and can now handle only 50 commits per second.`

---

wlmyng (approved):
looks reasonable

amnn (commented):
(LGTM!)

PR #26365, opened 2026-04-23
[graphql] Add scanTransactions and scanEvents skeleton behind staging

tpham-mysten:
## Summary

* Add skeleton `Query.scanTransactions` and `Query.scanEvents` fields, gated behind the `staging` feature, returning `feature_unavailable`.
* Introduce `CScanTransaction` and `CScanEvent` cursor types matching the signatures in #25717.
* Unblocks Phase 2 subscription backfill integration work while the team decides between bloom filter and roaring bitmap indexing strategies.

The real implementation is intentionally deferred — the skeletons exist so downstream work (subscription backfill via iterative scan extension) can target the final method signatures now.

## Test plan

- [x] `cargo check -p sui-indexer-alt-graphql --features staging` passes
- [x] `cargo check -p sui-indexer-alt-graphql` (default, no staging) passes with no new warnings
- [ ] Confirm schema SDL snapshot is unaffected in the default build (staging fields hidden)

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/query.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/query.rs
@@ -37,6 +37,8 @@ use crate::api::types::dynamic_field::DynamicField;
 use crate::api::types::epoch::CEpoch;
 use crate::api::types::epoch::Epoch;
 use crate::api::types::event::CEvent;
+#[cfg(feature = "staging")]
+use crate::api::types::event::CScanEvent;
 use crate::api::types::event::Event;
 use crate::api::types::event::filter::EventFilter;
 use crate::api::types::move_object::MoveObject;
@@ -60,6 +62,8 @@ use crate::api::types::signature_verify;
 use crate::api::types::signature_verify::IntentScope;
 use crate::api::types::signature_verify::SignatureVerifyResult;
 use crate::api::types::simulation_result::SimulationResult;
+#[cfg(feature = "staging")]
+use crate::api::types::transaction::CScanTransaction;
 use crate::api::types::transaction::CTransaction;
 use crate::api::types::transaction::Transaction;
 use crate::api::types::transaction::filter::TransactionFilter;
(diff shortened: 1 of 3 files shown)


crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
amnn, line 1: Out of interest, is the addition of `cp_sequence_number` here from the bloom filter based scanning transaction implementation, or a requirement from subscriptions? If the latter, then it may go away if we switch to roaring bitmaps, but if the former, then we will need to keep it.

crates/sui-indexer-alt-graphql/src/api/query.rs
```diff
@@ -724,6 +728,54 @@ impl Query {
         )
     }
 
+    /// The transactions that exist in the network, optionally filtered by multiple transaction filters.
+    ///
+    /// Supports combining filters (e.g. `affectedAddress` AND `function`). The checkpoint range being
+    /// scanned cannot exceed `ServiceConfig.maxScanLimit` — use `checkpointBound` to narrow it.
```
amnn, line 734: I'm not sure `checkpointBound` is a thing? The filters are `beforeCheckpoint`, `afterCheckpoint` and `atCheckpoint`. Also (nit) the doc comment should not hard-wrapped.

crates/sui-indexer-alt-graphql/src/api/query.rs
```diff
@@ -724,6 +728,54 @@ impl Query {
...
+    /// Supports a longer retention than `Query.transactions`.
+    ///
+    /// Note that this differs from `Query.transactions`, which supports a single filter with an
+    /// optional sender filter, with no checkpoint range restriction and generally a shorter retention.
+    #[cfg(feature = "staging")]
+    async fn scan_transactions(
```
amnn, line 740: One thing I've not fully understood is why having the stub implementations in the schema unblocks subscription phase 2 -- is it just so that you can synchronize the cursors etc? I would expect the actual implementation of subscription backfill to not call these functions directly, but rather share a common implementation.

crates/sui-indexer-alt-graphql/src/api/query.rs
```diff
@@ -724,6 +728,54 @@ impl Query {
...
+        // TODO(phase-2): Skeleton to unblock subscription backfill integration.
+        // Real implementation pending the team's decision on bloom filter vs
+        // roaring bitmap indexing strategy.
+        Err(feature_unavailable("scanTransactions"))
+    }
+
+    /// The events that exist in the network, optionally filtered by multiple event filters.
+    ///
+    /// Supports combining filters (e.g. `sender` AND `module` AND `type`). The checkpoint range being
+    /// scanned cannot exceed `ServiceConfig.maxScanLimit` — use `afterCheckpoint`, `beforeCheckpoint`,
```
amnn, line 758: Ah this comment is slightly better in that it doesn't mention `checkpointBounds`, but still needs to be unwrapped.

amnn (approved):
Looks good -- staging schema will need to be regenerated though (and I'm not super clear on why the stubs are needed, but that's a me problem and not a reason to not land it).

tpham-mysten:
This is not relevant anymore so close it

PR #26372, opened 2026-04-24
[consensus] add LeaderScheduleV3 sliding-window scorer

mwtian:
## Description 

Support multiple anchor blocks / leaders per round in leader schedule, and allow adjusting the number of leaders dynamically.

Computation for leader scores (comment above `LeaderScheduleV3`):

When `C >= 4`, C-1, C-2 and C-3 exist as pending commits.
Let `r` be the leader round of commit C-3.

For each authority A: consider A's block at round `r+1` (voting round) across commits
from commit C-2 to commit with leader round > `r+1`.
If there are multiple A blocks in this slot, `score[A] = 0`.

Collect the set of distinct authorities P, such that A's voting block votes to
(has an ancestor link to) P's round `r` block (leader block) in commit C-3.
Then `voted_for_stake[A] = sum_{P in distinct set} stake(P)`.

Collect the set of distinct authorities Q, such that a Q's round `r+2`
block certifies (has an ancestor link to) A's voting block in round `r+1`,
across commits from C-2 to commit with leader round > `r+2`.
Then `certified_by_stake[A] = sum_{Q in distinct set} stake(Q)`.

The overall score for A is `score[A] = voted_for_stake[A] * certified_by_stake[A]`.

## Test plan 

CI

### DIFF
--- a/consensus/config/src/consensus_protocol_config.rs
+++ b/consensus/config/src/consensus_protocol_config.rs
@@ -27,8 +27,8 @@ pub struct ConsensusProtocolConfig {
     bad_nodes_stake_threshold: u64,
     /// Whether to enable V3 logic.
     enable_v3: bool,
-    /// Number of recent commits retained by `LeaderScheduleV3` for its sliding-window scoring.
-    leader_schedule_running_length: u32,
+    /// Number of recent commits retained for leader schedule sliding-window scoring.
+    leader_schedule_window_size: u32,
 }
 
 impl Default for ConsensusProtocolConfig {
@@ -44,7 +44,7 @@ impl Default for ConsensusProtocolConfig {
             num_leaders_per_round: None,
             bad_nodes_stake_threshold: 0,
             enable_v3: false,
-            leader_schedule_running_length: 300,
+            leader_schedule_window_size: 300,
         }
     }
 }
@@ -61,7 +61,7 @@ impl ConsensusProtocolConfig {
         num_leaders_per_round: Option<usize>,
         bad_nodes_stake_threshold: u64,
         enable_v3: bool,
-        leader_schedule_running_length: u32,
+        leader_schedule_window_size: u32,
     ) -> Self {
         Self {
             protocol_version,
@@ -74,7 +74,7 @@ impl ConsensusProtocolConfig {
             num_leaders_per_round,
             bad_nodes_stake_threshold,
             enable_v3,
-            leader_schedule_running_length,
+            leader_schedule_window_size,
         }
     }
 
@@ -92,7 +92,7 @@ impl ConsensusProtocolConfig {
             num_leaders_per_round: Some(1),
             bad_nodes_stake_threshold: 30,
             enable_v3: false,
-            leader_schedule_running_length: 300,
+            leader_schedule_window_size: 300,
         }
     }
 
@@ -138,8 +138,8 @@ impl ConsensusProtocolConfig {
         self.enable_v3
     }
 
-    pub fn leader_schedule_running_length(&self) -> u32 {
-        self.leader_schedule_running_length
+    pub fn leader_schedule_window_size(&self) -> u32 {
+        self.leader_schedule_window_size
     }
 
     // Test setter methods
@@ -176,7 +176,7 @@ impl ConsensusProtocolConfig {
         self.enable_v3 = val;
     }
 
-    pub fn set_leader_schedule_running_length_for_testing(&mut self, val: u32) {
-        self.leader_schedule_running_length = val;
+    pub fn set_leader_schedule_window_size_for_testing(&mut self, val: u32) {
+        self.leader_schedule_window_size = val;
     }
 }
--- a/consensus/config/src/consensus_protocol_config.rs
+++ b/consensus/config/src/consensus_protocol_config.rs
@@ -25,7 +25,10 @@ pub struct ConsensusProtocolConfig {
     transaction_voting_enabled: bool,
     num_leaders_per_round: Option<usize>,
     bad_nodes_stake_threshold: u64,
+    /// Whether to enable V3 logic.
     enable_v3: bool,
+    /// Number of recent commits retained by `LeaderScheduleV3` for its sliding-window scoring.
+    leader_schedule_running_length: u32,
 }
 
 impl Default for ConsensusProtocolConfig {
@@ -41,6 +44,7 @@ impl Default for ConsensusProtocolConfig {
             num_leaders_per_round: None,
             bad_nodes_stake_threshold: 0,
             enable_v3: false,
+            leader_schedule_running_length: 300,
         }
     }
 }
@@ -57,6 +61,7 @@ impl ConsensusProtocolConfig {
         num_leaders_per_round: Option<usize>,
         bad_nodes_stake_threshold: u64,
         enable_v3: bool,
+        leader_schedule_running_length: u32,
     ) -> Self {
         Self {
             protocol_version,
@@ -69,6 +74,7 @@ impl ConsensusProtocolConfig {
             num_leaders_per_round,
             bad_nodes_stake_threshold,
             enable_v3,
+            leader_schedule_running_length,
         }
     }
 
@@ -86,6 +92,7 @@ impl ConsensusProtocolConfig {
             num_leaders_per_round: Some(1),
             bad_nodes_stake_threshold: 30,
             enable_v3: false,
+            leader_schedule_running_length: 300,
         }
     }
 
@@ -131,6 +138,10 @@ impl ConsensusProtocolConfig {
         self.enable_v3
     }
 
+    pub fn leader_schedule_running_length(&self) -> u32 {
+        self.leader_schedule_running_length
+    }
+
     // Test setter methods
 
     pub fn set_gc_depth_for_testing(&mut self, val: u32) {
@@ -163,5 +174,9 @@ impl ConsensusProtocolConfig {
 
     pub fn set_enable_v3_for_testing(&mut self, val: bool) {
         self.enable_v3 = val;
+    }
+
+    pub fn set_leader_schedule_running_length_for_testing(&mut self, val: u32) {
+        self.leader_schedule_running_length = val;
     }
 }
--- a/consensus/config/src/consensus_protocol_config.rs
+++ b/consensus/config/src/consensus_protocol_config.rs
@@ -46,8 +46,8 @@ impl Default for ConsensusProtocolConfig {
             num_leaders_per_round: None,
             bad_nodes_stake_threshold: 0,
             enable_v3: false,
-            leader_schedule_window_size: 300,
-            leader_schedule_update_interval: 12,
+            leader_schedule_window_size: 600,
+            leader_schedule_update_interval: 60,
         }
     }
 }
@@ -97,8 +97,8 @@ impl ConsensusProtocolConfig {
             num_leaders_per_round: Some(1),
             bad_nodes_stake_threshold: 30,
             enable_v3: false,
-            leader_schedule_window_size: 300,
-            leader_schedule_update_interval: 12,
+            leader_schedule_window_size: 600,
+            leader_schedule_update_interval: 60,
         }
     }
 
@@ -144,12 +144,15 @@ impl ConsensusProtocolConfig {
         self.enable_v3
     }
 
+    // Clamped to ≥ 1: downstream code (LeaderScheduleV3) uses these as window
+    // sizes and modulus operands, so a 0 config would divide by zero or be
+    // an empty window. Treat 0 as "effectively 1" instead of panicking.
     pub fn leader_schedule_window_size(&self) -> u32 {
-        self.leader_schedule_window_size
+        self.leader_schedule_window_size.max(1)
     }
 
     pub fn leader_schedule_update_interval(&self) -> u32 {
-        self.leader_schedule_update_interval
+        self.leader_schedule_update_interval.max(1)
     }
 
     // Test setter methods
--- a/consensus/config/src/consensus_protocol_config.rs
+++ b/consensus/config/src/consensus_protocol_config.rs
@@ -29,6 +29,8 @@ pub struct ConsensusProtocolConfig {
     enable_v3: bool,
     /// Number of recent commits retained for leader schedule sliding-window scoring.
     leader_schedule_window_size: u32,
+    /// Number of commit indices that use the same Mysticeti v3 leader schedule.
+    leader_schedule_update_interval: u32,
 }
 
 impl Default for ConsensusProtocolConfig {
@@ -45,6 +47,7 @@ impl Default for ConsensusProtocolConfig {
             bad_nodes_stake_threshold: 0,
             enable_v3: false,
             leader_schedule_window_size: 300,
+            leader_schedule_update_interval: 12,
         }
     }
 }
@@ -62,6 +65,7 @@ impl ConsensusProtocolConfig {
         bad_nodes_stake_threshold: u64,
         enable_v3: bool,
         leader_schedule_window_size: u32,
+        leader_schedule_update_interval: u32,
     ) -> Self {
         Self {
             protocol_version,
@@ -75,6 +79,7 @@ impl ConsensusProtocolConfig {
             bad_nodes_stake_threshold,
             enable_v3,
             leader_schedule_window_size,
+            leader_schedule_update_interval,
         }
     }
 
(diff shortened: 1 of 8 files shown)


consensus/config/src/consensus_protocol_config.rs
```diff
@@ -25,6 +25,11 @@ pub struct ConsensusProtocolConfig {
     transaction_voting_enabled: bool,
     num_leaders_per_round: Option<usize>,
     bad_nodes_stake_threshold: u64,
+
+    // Whether to enable V3 logic.
+    enable_v3: bool,
+    /// Number of recent commits retained by `LeaderScheduleV3` for its sliding-window scoring.
+    leader_schedule_running_length: u32,
```
akichidis: What about `leader_schedule_window_size` since in practice this is a sliding window algorithm?

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,858 @@
...
+    context: Arc<Context>,
+    next_commit_index: CommitIndex,
+    total_scores_per_authority: Vec<u64>,
+    // Sum of `num_leaders` across all entries in `scores_entry`.
+    total_num_leaders: usize,
+    // Sum of `leader_stakes` across all entries in `scores_entry`.
+    total_leader_stakes: u64,
+
+    // Holds at most the last two committed subdags, used as C-2 and C-1 when
+    // scoring the next commit. Populated from CommittedSubDag directly so no
```
akichidis: "Holds at most the last two committed subdags, used as C-2 and C-1 when scoring the next commit." 

Do I understand it correctly that technically it's not that we are scoring C, but rather we need `C-1` and `C` in order to score the authorities agains the leader blocks of `C-2` ?

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,858 @@
...
+        let voting_blocks: Vec<_> = c_minus_1
+            .blocks
+            .iter()
+            .chain(c.blocks.iter())
+            .filter(|b| b.round() == vote_round)
+            .collect();
+
+        let mut score_contributions = vec![0u64; self.context.committee.size()];
+
+        if !voting_blocks.is_empty() && !leader_refs.is_empty() {
```
akichidis: Under what circumstances the `leader_refs` would be empty since the filtering is practically for `c_minus_2` the `.filter(|b| b.round() == c_minus_2.leader.round)`

akichidis: Same with the `voting_blocks`

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,858 @@
...
+/// vote to (has an ancestor that is) a B's round `r` ("leader block") block in commit C-2.
+/// Then `score[A] += sum_{B in distinct set} stake(B)`.
+pub(crate) struct LeaderScheduleV3 {
+    context: Arc<Context>,
+    next_commit_index: CommitIndex,
+    total_scores_per_authority: Vec<u64>,
+    // Sum of `num_leaders` across all entries in `scores_entry`.
+    total_num_leaders: usize,
+    // Sum of `leader_stakes` across all entries in `scores_entry`.
+    total_leader_stakes: u64,
```
akichidis, line 63: Are those currently being used mostly for metrics/reporting reasons?

akichidis (commented):
I like the direction of this 💯

consensus/core/src/leader_schedule_v3.rs
mwtian: Yes, it is useful to have raw leader scores normalized, to understand how "perfect" they are.

consensus/core/src/leader_schedule_v3.rs
mwtian: Exactly. Let me rephrase.

consensus/core/src/leader_schedule_v3.rs
mwtian: They should never be empty. Will be updated.

consensus/core/src/core.rs
```diff
@@ -130,16 +144,22 @@ impl Core {
...
-        let mut ancestor_state_manager =
-            AncestorStateManager::new(context.clone(), dag_state.clone());
-        ancestor_state_manager.set_propagation_scores(propagation_scores);
+        let ancestor_state_manager = AncestorStateManager::new(context.clone(), dag_state.clone());
 
-        // Create the ValidatorProposer
+        // Create the ValidatorProposer.
+        let leader_waiter = if let Some(schedule) = leader_schedule_v3.as_ref() {
+            let next_commit_leader_schedule = schedule.next_commit_leader_schedule();
+            debug!(
```
akichidis: I would be fine having this as `info!`. It is logged only once on boot and it could be proven useful

consensus/core/src/proposer.rs
```diff
@@ -462,27 +450,35 @@ impl Proposer for ValidatorProposer {
...
-                    .as_millis() as u64,
-            );
-        self.context
-            .metrics
-            .node_metrics
-            .block_proposal_leader_wait_count
-            .with_label_values(&[leader_authority])
-            .inc();
+        // TODO(v3): support these metrics under v3 multi leader logic:
+        // - Leader slots can be empty when the quorum round is no greater than already committed leader round.
```
akichidis: I assume this is due to node recovery or when the quorum round has not progressed yet passed the `min_next_leader_round` ? Or there is something else?

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1510 @@
...
+        );
+
+        // Walk [C-2, C-1, C] up to and including the first commit whose
+        // leader round exceeds each bound.
+        let voting_commits =
+            Self::scan_commits_until_leader_round_above(&self.pending_commits, &c, vote_round);
+        let certifying_commits =
+            Self::scan_commits_until_leader_round_above(&self.pending_commits, &c, certify_round);
+
+        // Group r+1 voting blocks by author; len() != 1 detects equivocation.
```
akichidis: ` len() != 1 detects equivocation.` probably that's a comment for further down on the code path?

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1510 @@
...
+        assert!(!leader_refs.is_empty());
+        assert!(
+            leader_refs.contains(&c_minus_3.leader),
+            "C-3's leader {} missing from its own leader blocks {:?}",
+            c_minus_3.leader,
+            leader_refs,
+        );
+
+        // Walk [C-2, C-1, C] up to and including the first commit whose
+        // leader round exceeds each bound.
```
akichidis: Maybe let's add a comment why we choose to do the cut off here. My understanding is that  in theory we could iterate over all the pending commits and just filter out the `vote_round` and `certify_round` blocks, but we don't do it as this could take into account late arriving blocks (well mostly for the voting round) which ideally we don't want to score

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1510 @@
...
+            metrics
+                .leader_schedule_last_num_leaders
+                .set(last.num_leaders as i64);
+            metrics
+                .leader_schedule_average_num_leaders
+                .set(self.average_num_leaders());
+        }
+    }
+
+    fn running_length(&self) -> usize {
```
akichidis: maybe that could be `window_size` as well?

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1510 @@
...
+    cached_schedule: NextCommitLeaderSchedule,
+}
+
+/// This is used by the commit rule to figure out the next commit leaders.
+#[derive(Clone, Debug, Default)]
+pub(crate) struct NextCommitLeaderSchedule {
+    // Index of the next commit.
+    pub(crate) next_commit_index: CommitIndex,
+    // Minimum round of the next commit leader.
+    pub(crate) min_next_leader_round: Round,
```
akichidis: So this schedule will be valid for leader rounds that are `>= min_next_leader_round` ? For any other case it won't be used?

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1510 @@
...
+            .unwrap_or(GENESIS_ROUND)
+            + 1;
+        let num_leaders = self
+            .context
+            .protocol_config
+            .num_leaders_per_round()
+            .unwrap_or(1);
+        NextCommitLeaderSchedule {
+            next_commit_index: self.next_commit_index,
+            min_next_leader_round,
```
akichidis: Although we read the `nex_commit_index` from the memoised value, we do calculate the `min_next_leader_round` by accessing the `self.pending_commits.back()`. Could we maybe calculate both in the same place to make sure that those two are always coupled?

akichidis (commented):
leaving some comments for now

consensus/core/src/leader_schedule_v3.rs
mwtian: Updated comment. The goal is to inform committer about the min leader round for next commit, which is current commit leader round + 1.

consensus/core/src/proposer.rs
mwtian: Yes, during recovery or because of async logic, this seems possible. And we use an optimization to not wait on any leader.

consensus/core/src/leader_schedule_v3.rs
mwtian: Exactly, and the delayed blocks can be in either round.

consensus/core/src/proposer.rs
```diff
@@ -724,3 +725,86 @@ impl Proposer for ValidatorProposer {
...
+            Self::V3(schedule) => {
+                if round < schedule.min_next_leader_round {
+                    return vec![];
+                }
+                schedule
+                    .allowed_leaders
+                    .iter()
+                    .map(|leader| Slot::new(round, *leader))
+                    .collect()
+            }
```
akichidis, line 753: > Support multiple anchor blocks / leaders per round in leader schedule, and allow adjusting the number of leaders dynamically.

I might have missed something , but is there going to be a separate component doing the adjustment of leaders per round? Right now it seems that the whole set of "good" leaders will be returned - not filtering on the number of leaders is done.

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1562 @@
...
+                break;
+            }
+            accumulated_bad_stake += stake;
+            by_score.pop();
+        }
+
+        by_score.into_iter().map(|(idx, _)| idx).collect()
+    }
+
+    /// On each `add_commit(c)`, once 3 commits have accumulated, a score entry
```
akichidis: I would suggest:

_once 3 commits have accumulated -> once 3 commits are already buffered in `pending_commits` (before `c` is added)`_

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1562 @@
...
+                    certifying_blocks.push(b);
+                }
+            }
+        }
+
+        let mut score_contributions = vec![0u64; self.context.committee.size()];
+        for (a, voting_blocks) in &voting_blocks_by_author {
+            // Equivocation: more than one r+1 block from A → score[A] = 0.
+            if voting_blocks.len() != 1 {
+                continue;
```
akichidis, line 342: let's add a `warn` here? - it would be good to have some traceability for this edge case

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1562 @@
...
+    /// Number of recent committed sub-dags that must be replayed on startup
+    /// to reconstruct the full state: `leader_schedule_window_size` for the
+    /// scoring window, the pending commits held before scoring begins, and
+    /// `interval - 1` more so the replay spans back to the most recent rotation
+    /// boundary with the full scoring window already populated.
+    fn replay_length(context: &Context) -> u32 {
+        context.protocol_config.leader_schedule_window_size()
+            + MAX_PENDING_COMMITS as u32
+            + context.protocol_config.leader_schedule_update_interval()
+            - 1
```
akichidis: This is not wrong, but - it might be just me - it took me a bit to realise why this is right. An alternative on the calculation that I am leaving here in case it feels (more) intuitive. 
The goal is to recover enough commits so we trigger the last boundary point (so the current_schedule gets accurately populated) and onwards:
```
      let interval = context.protocol_config.leader_schedule_update_interval();

      // Detect the distance from last commit where the last boundary schedule change happened.                                                                                                                                                                                                                          
      let distance_past_boundary = last_commit_index % interval;
      
      // Now add the minimum amount of commits needed to calculate a full size window worth of scores.
      context.protocol_config.leader_schedule_window_size()                                                                                                                                                                                                                                               
          + MAX_PENDING_COMMITS as u32
          + distance_past_boundary  
```

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1562 @@
...
+        let mut leader_schedule = LeaderScheduleV3::new(context.clone(), vec![0; committee_size]);
+
+        let last_commit_index = dag_state.read().last_commit_index();
+        let replay_count = LeaderScheduleV3::replay_length(&context);
+        // Replay maximum replay_count commits, and starting from commit 1 at minimum.
+        let replay_start = last_commit_index.saturating_sub(replay_count) + 1;
+        if replay_start > last_commit_index {
+            // No commits to replay, return the new schedule with empty state.
+            return leader_schedule;
+        }
```
akichidis: maybe mentally could be better to have a chekc like:

```
let last_commit_index = dag_state.read().last_commit_index();
if last_commit_index == 0 {
     // No commits to replay, return the new schedule with empty state.
     return leader_schedule;
}
```

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1562 @@
...
+        let schedule = LeaderScheduleV3::new(context, vec![0; 4]);
+        assert_eq!(schedule.total_scores_per_authority(), &[0u64; 4]);
+        assert_eq!(schedule.scores_entries_len(), 0);
+        assert_eq!(schedule.pending_commits_len(), 0);
+        assert_eq!(schedule.average_num_leaders(), 0.0);
+        assert_eq!(schedule.total_leader_stakes(), 0);
+    }
+
+    #[tokio::test]
+    async fn test_first_three_commits_score_zero() {
```
akichidis: Could that test be consolidated with the `test_fourth_commit_scores_from_c_minus_3` ? 

we could basically assert that until commit 3 nothing changes

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -0,0 +1,1562 @@
...
+
+        // A=0 has two round-2 blocks → equivocation → score=0.
+        // A=1: voted_for_stake=s0, certified_by_stake=s0 → score=s0*s0.
+        let s0 = context.committee.stake(AuthorityIndex::new_for_test(0));
+        assert_eq!(schedule.total_scores_per_authority()[0], 0);
+        assert_eq!(schedule.total_scores_per_authority()[1], s0 * s0);
+    }
+
+    #[tokio::test]
+    async fn test_score_is_product_of_voted_for_and_certified_by_stake() {
```
akichidis: Is this needed? My understanding is that this path has been covered by other tests

akichidis (approved):
I am approving to unblock, but please have a look on the comments

consensus/core/src/leader_schedule_v3.rs
mwtian: Since this is not an operator issue, I can add a `debug!()`

consensus/core/src/leader_schedule_v3.rs
mwtian: Yes it has been covered but this test is more intentional.

consensus/core/src/leader_schedule_v3.rs
mwtian: Updated to compute replay_start here.

consensus/core/src/proposer.rs
mwtian: Right now I don't see the value of rotating a subset of good authorities as leaders, compared to using all of good authorities as leaders every round, as long as the threshold of good authorities can be dynamically adjusted. We can evaluate in future on if there are better strategies.

PR #26380, opened 2026-04-24
[chore] Fix advisories

stefan-mysten:
## Description 

This PR bumps:
- diesel from 2.2 to 2.3.8
- diesel-async 0.5 to 0.8
- diesel-migrations from 2.2 to 2.3
- bb8 0.8.5 to 0.9

## Test plan 

Existing CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt/src/handlers/sum_displays.rs
+++ b/crates/sui-indexer-alt/src/handlers/sum_displays.rs
@@ -10,7 +10,6 @@ use async_trait::async_trait;
 use diesel::ExpressionMethods;
 use diesel::upsert::excluded;
 use diesel_async::RunQueryDsl;
-use futures::future::try_join_all;
 use sui_indexer_alt_framework::FieldCount;
 use sui_indexer_alt_framework::pipeline::Processor;
 use sui_indexer_alt_framework::pipeline::sequential::Handler;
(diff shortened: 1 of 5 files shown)


crates/sui-indexer-alt/src/handlers/sum_displays.rs
```diff
@@ -78,21 +77,21 @@ impl Handler for SumDisplays {
...
-                    .set((
-                        sum_displays::display_id.eq(excluded(sum_displays::display_id)),
-                        sum_displays::display_version.eq(excluded(sum_displays::display_version)),
-                        sum_displays::display.eq(excluded(sum_displays::display)),
-                    ))
-                    .execute(conn)
-            });
-
-        Ok(try_join_all(updates).await?.into_iter().sum())
+        let mut updates = 0;
```
stefan-mysten, line 80: Not sure why Claude decided to split it like this. I assume this is less efficient.

stefan-mysten: Oh, it's because of changes to `diesel-async`.

ebmifa (approved):

crates/sui-indexer-alt/src/handlers/sum_displays.rs
amnn: Yes, this is less efficient -- what are the changes to `diesel-async` that necessitate this?

tpham-mysten (approved):

crates/sui-indexer-alt/src/handlers/sum_displays.rs
stefan-mysten: It seems that there was a lifetime issue in 0.5.0, because it allowed to borrow the connection while creating multiple async query futures. There's only one connection, so the borrow checker won't let multiple futures hold `&mut conn` at once.

Here's the compiler error
```
error: captured variable cannot escape `FnMut` closure body
  --> crates/sui-indexer-alt/src/handlers/sum_displays.rs:84:17
   |
79 |       async fn commit<'a>(&self, batch: &Self::Batch, conn: &mut Connection<'a>) -> Result<usize> {
   |                                                       ---- variable defined here
...
83 |               .map(|chunk: &[StoredDisplay]| {
   |                                            - inferred to be a `FnMut` closure
84 | /                 diesel::insert_into(sum_displays::table)
85 | |                     .values(chunk)
86 | |                     .on_conflict(sum_displays::object_type)
87 | |                     .do_update()
...  |
92 | |                     ))
93 | |                     .execute(conn)
   | |______________________________----^ returns a reference to a captured variable which escapes the closure body
   |                                |
   |                                variable captured here
   |
   = note: `FnMut` closures only have access to their captured variables while they are executing...
   = note: ...therefore, they cannot allow references to captured variables to escape

error: could not compile `sui-indexer-alt` (lib) due to 1 previous error
```

PR #26382, opened 2026-04-27
Fix Full node hanging on shutdown

ncucinotta:
## Description 

During maintenance operations, we noticed that sometimes our Sui nodes are hanging indefinitely on SIGTERM, resulting in a partial shutdown state unable to respond to RPC queries while the process is still running. This behavior is exclusively present on Full nodes, and not on validators.

We traced this to the blocking background task in `sui_core::verify_indexes::fix_indexes()`, unable to be dropped by tokio until finished. This runs at every startup and requires over 4 hours from node startup from on our mainnet machine (NVME SSD, 256GB RAM), during which time issuing a SIGTERM results in the aformentioned zombie state.

I am not familiar enough with the Sui codebase to determine whether this repair function introduced in https://github.com/MystenLabs/sui/pull/21559 should be removed. I opted to instead add a small shutdown signaling system from the node to the tokio Runtimes to gracefully handle terminating such long-running tasks.

## Test plan

Ran these changes rebased on top of the most recent mainnet release. Restarting the node multiple times in a row was reliable.

---

## Release notes

- [ ] Protocol: 
- [X] Nodes (Validators and Full nodes): Full nodes do not hang on shutdown after being recently restarted
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -54,7 +54,7 @@ use std::collections::BTreeSet;
 use std::fs::File;
 use std::io::Write;
 use std::path::{Path, PathBuf};
-use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
 use std::time::Instant;
 use std::time::SystemTime;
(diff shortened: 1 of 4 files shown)


amnn:
Adding @bmwill and @nickvikeras here -- they will be more familiar with how indices work on `sui-node`.

bmwill:
I think we actually want to remove this code, theres a comment saying it was temporary and should be removed.

bmwill:
https://github.com/MystenLabs/sui/pull/26663 for this

amnn:
Closing as superseded, thanks for flagging the issue @ncucinotta

PR #26383, opened 2026-04-27
[Block Streaming]  return multiple blocks per response

akichidis:
## Description 

Refactor the `BlockStreamResponse` to return a vector of blocks instead of single block. Refactored the `BroadcastStream` so it can poll the underlying channel and fetch a batch of blocks instead of a single one.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/authority_service.rs
+++ b/consensus/core/src/authority_service.rs
@@ -373,15 +373,23 @@ impl<C: CoreThreadDispatcher> ValidatorNetworkService for AuthorityService<C> {
             )
         };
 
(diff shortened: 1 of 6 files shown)


mwtian (approved):
Let's look into or add TODO on limiting max message siz, max inflight size, and max inflight count. They are more complicated with batched blocks unfortunately.

PR #26386, opened 2026-04-27
[lint] unused object with fields

damirka:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,324 @@
...
+        let mut tracked_params = BTreeMap::new();
+        for (_mutability, v, st) in &context.signature.parameters {
+            if is_qualifying_obj(
+                context.info,
+                context.pre_compiled_program.as_deref(),
+                st,
+            ) {
+                tracked_params.insert(*v, v.0.loc);
+                let locals = init_state.locals_mut();
+                if let Some(LocalState::Available(loc, _)) = locals.get(v) {
```
tnowacki: This should always be available at this point?

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,324 @@
...
+                if self.tracked_params.contains_key(var) {
+                    self.used_params.borrow_mut().insert(*var);
+                }
+                // Transitive: the local might hold a value assigned from a tracked param
+                if let Some(LocalState::Available(_, Value::UnusedObj(orig_var, _))) =
+                    state.locals().get(var)
+                {
+                    self.used_params.borrow_mut().insert(*orig_var);
+                }
+                None
```
tnowacki: I don't understand this case. Wouldn't this only be for an owned object? And we don't care about owned objects?

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,324 @@
...
+            }
+            // Field access / deref / freeze on a ref: the default handler evaluates the
+            // inner expression but discards its abstract value. We intercept to check
+            // whether the inner value tracks an unused ref param.
+            E::Borrow(_, inner, _, _) | E::Dereference(inner) | E::Freeze(inner) => {
```
tnowacki: I think we will need to really expand this area 
- Freezing definitely doesn't seem like a "usage"
- You cannot Dereference an object
- I would like to know if the `Borrow` is actually used. If you don't track that, there isn't really any point in using abstract interpretation as far as I can tell

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,324 @@
...
+    type State = State;
+    type ExecutionContext = ExecutionContext;
+
+    fn finish(
+        &mut self,
+        _final_states: BTreeMap<Label, State>,
+        mut diags: Diagnostics,
+    ) -> Diagnostics {
+        let used = self.used_params.borrow();
+        for (var, loc) in &self.tracked_params {
```
tnowacki: This is going to be a bit brittle and I'm not sure if this is the best thing to do... maybe it is fine for now?
In general, this sort of pattern will not work if there is ever a situation in which the analysis removes a param from the used set. In such a case, you need to look at the post state of the blocks and derive the set of used params from there

tnowacki (changes requested):
I think this needs some work in terms of checking if the borrowed field is actually used. OR if we think that isn't important, there is no reason to use abstract interpretation and we can just do a syntactic analysis

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
damirka: > I would like to know if the Borrow is actually used. If you don't track that, there isn't really any point in using abstract interpretation as far as I can tell

I think that's a fair point and something that I still have doubts about. Are you sure that we want to be that strict about usage of the borrowed field? I'm not sure if noise / utility ratio would be in our favor.

tnowacki: We can find out yeah? I imagine the noise will be low, since there isn't any way to "use" a reference outside of 
- Eq/Neq
- Borrow
- ReadRef
- WriteRef
- Call

Move/Copy/Freeze are just propagation 

And you can't use the object for Read/Write (though you will be able to for the fields and what not)

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,341 @@
...
+            return None;
+        }
+
+        let mut tracked_params = BTreeMap::new();
+        for (_mutability, v, st) in &context.signature.parameters {
+            if is_qualifying_obj(context.info, context.pre_compiled_program.as_deref(), st) {
+                tracked_params.insert(*v, v.0.loc);
+                let locals = init_state.locals_mut();
+                let LocalState::Available(loc, val) = locals.get_mut(v).unwrap() else {
+                    unreachable!("parameter must be available at init")
```
tnowacki: nit this should be an ice or something  else that will fail gracefully

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
tnowacki: I think to extend this for capabilities where we want to see a control flow path that ends in an abort, this won't work probably. 
I would explore swapping this out with something that processes the post states of every block

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,341 @@
...
+        &self.locals
+    }
+
+    fn join_value(v1: &Value, v2: &Value) -> Value {
+        match (v1, v2) {
+            (Value::Other, _) | (_, Value::Other) => Value::Other,
+            (Value::FieldOf(var, loc), _) | (_, Value::FieldOf(var, loc)) => {
+                Value::FieldOf(*var, *loc)
+            }
+            (Value::UnusedObj(var, loc), Value::UnusedObj(_, _)) => Value::UnusedObj(*var, *loc),
```
tnowacki: Thinking about join, I think your domain and value should both just be `BTreeSet<Var>` and I don't think you need a the `Loc`?

Imagine code like 
```
let tmp;
if (cond) tmp = obj_1 else tmp = obj_2;
foo(tmp); 
```
This would erroneously mark one of the two `obj_1` or `obj_2` as unused. 

You could imagine similar combinations for fields or parameters.

Currently as this is written I don't think it is important that each individual field is tracked... though you could imagine wanting to do so if you wanted to say _all_ fields are used. Though that will get tricky real quick

tnowacki: So the `Value` becomes `BTreeSet` 
And your domain can just be `used: BTreeSet<Var>` too.

tnowacki: Ah looks like you need to know field or not for `return`. I would probably split it into something like 
```
enum Value {
  Tracked {
    // invariant that root || field. 
    // you could split this into an enum but probably don't gain much
    root: bool,
    field: bool,
    roots: BTreeSet<Var>
  }
  None,
}
```
or something similar

external-crates/move/crates/move-compiler/tests/sui_mode/linter/unused_object_with_fields.move
```diff
@@ -0,0 +1,98 @@
...
+    public fun t_10(_e: &Empty, c: &OwnerCap) { assert_owner(c.owns) }
+
+    // not triggered: passed by value
+    public fun t_11(_e: &Empty, c: OwnerCap) { consume(c) }
+
+    // not triggered: passed by value (testing with unpack)
+    public fun t_12(_e: &Empty, c: OwnerCap) {
+        let OwnerCap { id, owns: _ } = c;
+        object::delete(id)
+    }
```
tnowacki: We should split these all into true/false positives/negatives (so four tests in total)
(and maybe 8 if we separate out caps)
(and maybe 12 if we do hot potatoes)

damirka: Hot potatoes? We're currently only looking at anything that has `key`. And the lint is "unused_object_...", do you think it makes sense to extend it?

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,527 @@
...
+    fn finish_command(&self, context: ExecutionContext, state: &mut State) -> Diagnostics {
+        // Capture this command's post-state contributions. Once a var is in
+        // `state.used`, every successor's pre-state inherits it through the
+        // join; harvesting after every command means no terminal block's
+        // contribution is lost.
+        self.used_in_post_states
+            .borrow_mut()
+            .extend(state.used.iter().copied());
+        let ExecutionContext { diags } = context;
+        diags
```
tnowacki, line 267: This falls to the same brittleness, and doesn't really do anything different than, just adding them to some global as we had before. 
You don't know if this is the last time the command is finished, and while it might do fine today, it will be brittle for the future. The only way to know that you have the final thing is through the post states.
If you don't have them, we should add them (and I may have to dig up a very old commit to fix this)

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,527 @@
...
+            // Inspecting a tracked value to drive control flow counts.
+            C::JumpIf { cond: e, .. } | C::VariantSwitch { subject: e, .. } => {
+                let vals = self.exp(context, state, e);
+                mark_used(state, &vals);
+                true
+            }
+            // `IgnoreAndPop`, `Abort`, etc. are left to the default handler
+            // so the value is evaluated for sub-expression side effects but
+            // not counted as a use on its own.
+            _ => false,
```
tnowacki: I would also treat such things as `IgnoreAndPop` and `Abort` as usage for a field

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,527 @@
...
+                            },
+                            Value::Other => Value::Other,
+                        })
+                        .collect(),
+                )
+            }
+            // Pure "view" operations — pass tracking through; consumers
+            // downstream are what mark.
+            E::Dereference(inner)
+            | E::Freeze(inner)
```
tnowacki: I would probably only treat `freeze` as a pass through. The rest to me are usages to me.
(At least for fields, but the root can't be used in any of them)

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,487 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+pub enum Value {
+    /// Carries data of one or more tracked roots.
+    /// `is_field` is true if some incoming path produced a field-derived
+    /// value (i.e. went through a `Borrow`); false means the value is a
+    /// bare unchanged reference to the param itself.
+    Tracked {
+        is_field: bool,
+        roots: BTreeSet<Var>,
+    },
```
tnowacki: I think you will need to change this (yet again sorry). 
This style won't catch something like
```move
let x = if (cond) a else &obj.f;
return x
```
That is, you need to know for each variable if it is a field or a root.

Technically once we have recursive types you might need something even more granular for cases like
```move
let x = if (cond) list else &list.next;
return list
```

tnowacki: Maybe it is `BTreeMap<Var, Kind>`
and if `Kind = bool` where `true` is a "field", then join can be `&&`?

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,487 @@
...
+            C::Mutate(lhs, rhs) => {
+                let rhs_vals = self.exp(context, state, rhs);
+                let lhs_vals = self.exp(context, state, lhs);
+                mark_used(state, &lhs_vals);
+                // The RHS counts only when the write target is itself a
+                // tracked field — i.e. a `&mut` field of an input — so the
+                // RHS value escapes via that input.
+                if lhs_vals.iter().any(|v| matches!(v, Value::Tracked { .. })) {
+                    mark_used(state, &rhs_vals);
+                }
```
tnowacki: Is this case reachable? I don't think it is?
You can't have a reference to a reference so the RHS must always have a non-reference type

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,487 @@
...
+            // Inspecting a tracked value to drive control flow counts.
+            C::JumpIf { cond: e, .. } | C::VariantSwitch { subject: e, .. } => {
+                let vals = self.exp(context, state, e);
+                mark_used(state, &vals);
+                true
+            }
+            // `IgnoreAndPop` (`let _ = e;`, `e;`) and `Abort` evaluate the
+            // sub-expression for its side effects but the value itself does
+            // not flow out of the function — fall through.
+            _ => false,
```
tnowacki: I would exhaustively match here maybe, since I think you are missing a lot of cases.
- Assign should propagate (I hope it does by default and we can just do `None` here)
- Abort should maybe be a usage? Though a weird one
- Variant switch maybe should be one? If we count a book as one for jump if why not this one?

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,487 @@
...
+            // Binop produces a fresh field-derived value carrying the union
+            // of operand roots; downstream consumers mark.
+            E::BinopExp(e1, _, e2) => {
+                let mut roots = BTreeSet::new();
+                for v in self.exp(context, state, e1) {
+                    collect_roots(v, &mut roots);
+                }
+                for v in self.exp(context, state, e2) {
+                    collect_roots(v, &mut roots);
+                }
```
tnowacki: Is this not just mark used for vals?
Why do propagate?

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,487 @@
...
+            // from which root — any later use of the packed struct treats
+            // every contributing root as used.
+            E::Pack(_, _, fields) | E::PackVariant(_, _, _, fields) => {
+                let mut roots = BTreeSet::new();
+                for (_, _, fe) in fields {
+                    for v in self.exp(context, state, fe) {
+                        collect_roots(v, &mut roots);
+                    }
+                }
+                Some(vec![tracked_field_or_other(roots)])
```
tnowacki: I don't understand why we propagate here? why isn't this just a usage?

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,476 @@
...
+    fn exp_custom(
+        &self,
+        context: &mut ExecutionContext,
+        state: &mut State,
+        e: &Exp,
+    ) -> Option<Vec<Value>> {
+        use UnannotatedExp_ as E;
+        match &e.exp.value {
+            E::BorrowLocal(_, var) => match state.locals().get(var) {
+                Some(LocalState::Available(_, value)) => Some(vec![value.clone()]),
```
tnowacki: Why does this clone the value here? Should this not always be Other?

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,476 @@
...
+        _f: &ModuleCall,
+        args: Vec<Value>,
+    ) -> Option<Vec<Value>> {
+        // A tracked value flowing into a function call has escaped.
+        state.mark_used(&args);
+        None
+    }
+}
+
+fn is_unpack_borrow(e: &Exp) -> bool {
```
tnowacki: I don't fully understand this. What is the case where this applies?

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -0,0 +1,476 @@
...
+            if let Value::Tracked(map) = v {
+                self.used.extend(map.keys().copied());
+            }
+        }
+    }
+
+    /// Like [`Self::mark_used`], but only marks roots that flow as
+    /// field-derived values. A bare root reference ([`Kind::Bare`]) is a
+    /// pass-through and does not count on its own.
+    fn mark_used_fields(&mut self, values: &[Value]) {
```
tnowacki: The name of this seems strange, and it through me off for a bbit as it seems to imply marking all of the fields used or something similar. Perhaps something like `mark_field_derived_used`

external-crates/move/crates/move-compiler/tests/sui_mode/linter/unused_object_with_fields_false_negatives.move
```diff
@@ -0,0 +1,21 @@
...
+    struct OwnerCap has key { id: UID, owns: address }
+
+    // forwarding the object to a function that ignores it is treated as
+    // use, since the lint does not look inside the callee
+    public fun forwarded_to_no_op(c: &OwnerCap) {
+        no_op(c);
+    }
+
+    #[allow(lint(unused_object_with_fields))]
+    fun no_op(_c: &OwnerCap) {}
```
tnowacki: This doesn't really seem like a false negative to me... since you had to allow the lint.

Perhaps something where the function is in another module? Or something that relies on generics might be more applicable?

The point of these "false negative" cases is to be honest about _all possible_ blind spots; I don't feel like this test does that.

external-crates/move/crates/move-compiler/tests/sui_mode/linter/unused_object_with_fields_false_positives.move
```diff
@@ -0,0 +1,18 @@
+// Cases where the lint warns even though the object is, in some sense, used.
+// These are limitations of the local analysis: a handful of operations drop
+// or don't follow the tracked value before any consumer can mark it.
+//
+// (No known false positives currently — placeholder kept for future cases.)
+
```
tnowacki: If there aren't any false positives... don't write this test 😛

damirka: Yeah, agreed, I wasn't sure whether to keep them.

external-crates/move/crates/move-compiler/tests/sui_mode/linter/unused_object_with_fields_false_negatives.move
damirka: We intentionally skip generics for now (the TreasuryCap<T> being the primary noisy case for this call). The test case is still off, agreed.

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
damirka: Patched.

external-crates/move/crates/move-compiler/tests/sui_mode/linter/unused_object_with_fields_false_positives.move
damirka: Removing now.

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
```diff
@@ -392,9 +388,9 @@ impl SimpleAbsInt for UnusedObjWithFieldsAI {
     }
 }
 
-fn is_unpack_borrow(e: &Exp) -> bool {
-    // Synthetic borrows from ref-pattern destructuring do not count just
-    // because their field binding is ignored.
+fn is_ref_pattern_field_borrow(e: &Exp) -> bool {
+    // Ref-pattern destructures (`let S { f: _ } = &s`) lower to synthetic
+    // field borrows assigned to `_`; ignored fields should remain unused.
```
tnowacki, line 393: Do we want those to be unused? If it was for `..`, that makes sense to me. But `_` seems it could go either way. What was the thought here?

external-crates/move/crates/move-compiler/tests/sui_mode/linter/unused_object_with_fields_false_negatives.move
```diff
@@ -1,19 +1,13 @@
-// Cases where the lint stays silent even though the object is, in practice,
-// not really used. Limitations of the local analysis.
+// Cases where the lint stays silent even though the object may be unused.
+// These are known blind spots.
 
 module a::m {
     use sui::object::UID;
 
-    struct OwnerCap has key { id: UID, owns: address }
+    struct GenericCap<phantom T> has key { id: UID, owns: address }
```
tnowacki, line 8: Do we have more blind spots than this one?

tnowacki: The blind spot I was referring to with generics was something like `foo(c)` where `foo<T>(c: &T)`, where it might not really be doing anything interesting with `c`. 

The assumption that calls should be considered a "usage" is an inductive one, and that breaks down if the function itself would not trigger the lint due to some limitation.

external-crates/move/crates/move-compiler/src/sui_mode/linters/unused_object_with_fields.rs
damirka: They're too similar? Almost identical? It's a weird edge case which no one but myself would hit. Thought to cover it as unused since we have the power of AbsInt.

external-crates/move/crates/move-compiler/tests/sui_mode/linter/unused_object_with_fields_false_negatives.move
damirka: Added `foo<T>` silencing case.

damirka: I guess `object::id()` falls into this bucket.

tnowacki (approved):
I think there is still a bit of work done possibly for false negatives. There are a good amount of things we don't cover and maybe should, though some of that is covered in the true negative case... so its probably fine. 
I think we should really think about how to expand this to generic types though! It is a bit of a heavy limitation

PR #26391, opened 2026-04-27
Include root package type origins in on-chain move summary

hayes-mysten:
## Summary
- `sui move summary` for on-chain packages was inserting type origins for every entry in the linkage table but skipping the root package itself, so the generated `root_package_metadata` was missing type origin info for the package being summarized.
- Fix is a single `BTreeMap::insert` in `download_package_and_deps_under` (`crates/sui/src/sui_commands.rs`) using `root_package.original_package_id()` as the key.

## Test plan
- [x] `cargo check -p sui`
- [x] `cargo xclippy` in `crates/sui`
- [x] `cargo nextest run -p sui --lib` (58/58 pass; no existing tests cover this path)
- [x] Manual: run `sui move summary --package-id <id>` against testnet and verify `root_package_metadata` contains type origins for the root package

### DIFF
--- a/crates/sui/src/sui_commands.rs
+++ b/crates/sui/src/sui_commands.rs
@@ -1749,6 +1749,11 @@ async fn download_package_and_deps_under(
         type_origins.insert(*original_id, package.type_origin_table().clone());
     }
 
+    type_origins.insert(
+        root_package.original_package_id(),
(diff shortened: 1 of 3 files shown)


tzakian (approved):
Thanks for the fix!

PR #26396, opened 2026-04-28
[consensus] track block children and per-round info in dag state

mwtian:
## Description 

Adds two pieces of per-block / per-round metadata consumed by the v3 commit rule:

- BlockInfo now tracks children, children_authorities and total_children_stake, populated in update_block_metadata() when a block is accepted.
- RoundInfo aggregates, per retained round, the set of unique authoring authorities and their total stake. It is held in a VecDeque whose front is gc_round + 1 and back is highest_accepted_round, extended on block accept and truncated on flush once entries fall to <= gc_round.

## Test plan 

CI

### DIFF
--- a/consensus/core/src/dag_state.rs
+++ b/consensus/core/src/dag_state.rs
@@ -358,7 +358,7 @@ impl DagState {
         if self.context.protocol_config.enable_v3() {
             // Update votes accounting in BlockInfo.
             for ancestor in block.ancestors() {
-                // Only update children when the link is potentially a leader vote.
+                // Only update children info when the link is potentially a leader vote.
                 if ancestor.round + 1 != block_ref.round || ancestor.round <= self.gc_round() {
                     continue;
                 }
@@ -524,6 +524,9 @@ impl DagState {
 
     #[cfg(test)]
     pub(crate) fn get_block_children(&self, block_ref: &BlockRef) -> Option<Vec<BlockRef>> {
+        if block_ref.round <= self.gc_round() {
+            return None;
+        }
         self.recent_blocks
             .get(block_ref)
             .map(|block_info| block_info.children.iter().cloned().collect())
@@ -534,6 +537,9 @@ impl DagState {
         &self,
         block_ref: &BlockRef,
     ) -> Option<BTreeSet<AuthorityIndex>> {
+        if block_ref.round <= self.gc_round() {
+            return None;
+        }
         self.recent_blocks
             .get(block_ref)
             .map(|block_info| block_info.children_authorities.clone())
@@ -541,6 +547,9 @@ impl DagState {
 
     #[cfg(test)]
     pub(crate) fn get_block_total_children_stake(&self, block_ref: &BlockRef) -> Option<Stake> {
+        if block_ref.round <= self.gc_round() {
+            return None;
+        }
         self.recent_blocks
             .get(block_ref)
             .map(|block_info| block_info.total_children_stake)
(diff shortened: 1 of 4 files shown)


consensus/core/src/dag_state.rs
```diff
@@ -872,6 +926,61 @@ impl DagState {
...
+                round: block.round(),
+                authorities: BTreeSet::new(),
+                total_stake: 0,
+            });
+        } else {
+            assert!(
+                block.round() < next_round,
+                "Attempted to update round info for block {block_ref} with round higher than next round {next_round}"
+            );
+        }
```
akichidis, line 985:
```suggestion
        // DAG can only grow one round at a time.
        assert!(
            block.round() <= next_round,
            "Attempted to update round info for block {block_ref} with round higher than next round {next_round}"
        );
        
        if block.round() == next_round {
            self.round_info.push_back(RoundInfo {
                round: block.round(),
                authorities: BTreeSet::new(),
                total_stake: 0,
            });
        }
```

consensus/core/src/dag_state.rs
```diff
@@ -872,6 +926,61 @@ impl DagState {
...
+        if block.round() <= gc_round {
+            return;
+        }
+
+        // Check if a new RoundInfo needs to be created.
+        let next_round = self
+            .round_info
+            .back()
+            .map(|info| info.round + 1)
+            .unwrap_or(gc_round + 1);
```
akichidis, line 973: Under what circumstances the `round_info` we expect to be empty? During `DagState` init without any block rounds accepted? I am trying to think why the `gc_round + 1` is necessary

consensus/core/src/dag_state.rs
```diff
@@ -346,7 +355,28 @@ impl DagState {
         self.recent_blocks
             .insert(block_ref, BlockInfo::new(block.clone()));
         self.recent_refs_by_authority[block_ref.author].insert(block_ref);
-
+        if self.context.protocol_config.enable_v3() {
+            // Update votes accounting in BlockInfo.
+            for ancestor in block.ancestors() {
+                // Only update children when the link is potentially a leader vote.
```
akichidis: So we are doing this whole accounting in order to keep track and easily access a potential leader's (`ancestor`) total voted stake and children?

consensus/core/src/dag_state.rs
mwtian: Added a comment. Basically because of the assert below, there needs to be the appropriate next_round when recovering or starting a new epoch.

consensus/core/src/dag_state.rs
mwtian: Yes, as opposed to computing these each time commit rule runs.

consensus/core/src/dag_state.rs
```diff
@@ -57,6 +57,11 @@ pub struct DagState {
     // Vec position corresponds to the authority index.
     recent_refs_by_authority: Vec<BTreeSet<BlockRef>>,
 
+    // Per-round aggregation of accepted blocks. Front holds the oldest retained
+    // round (`gc_round + 1` after eviction); back holds `highest_accepted_round`.
+    // Rounds between front and back are always contiguous in a valid DagState.
+    round_info: VecDeque<RoundInfo>,
```
akichidis, line 63: I see the reason of using the `VecDeque` here, but looking on the access patterns in the codebase and the fact that this is actually mutable data, could a `BTreeMap<Round, RoundInfo>` also make sense? Just to avoid calculating position indexing etc. Leaving as an optional suggestion.

consensus/core/src/dag_state.rs
akichidis: Got it. Yes you are right, we need `gc_round` during recovery

akichidis (approved):

consensus/core/src/dag_state.rs
mwtian: I think the calculation is fine. Added an assertion to make sure retrieved RoundInfo matches requested round.

PR #26397, opened 2026-04-28
[move-vm][housekeeping] Swap to quanta telemetry

cgswords:
## Description 

The VM currently calls `Instant::now` which can be expensive in a system under heavy load. This swaps it for `quanta`, a library that is less precise but faster. Since we are only using milliseconds, this should be just fine for us.

## Test plan 

Everything is still green.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6228,7 +6228,7 @@ dependencies = [
  "no-std-compat",
  "nonzero_ext",
  "parking_lot 0.12.3",
- "quanta",
+ "quanta 0.11.1",
  "rand 0.8.5",
  "smallvec",
 ]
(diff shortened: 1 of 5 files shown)


tzakian (approved):

tnowacki (approved):

cgswords:
Closing this PR because it does not work with simtest, and cfg-gating this change to handle that specific instance seems bad.

PR #26399, opened 2026-04-28
[Block Streaming] Introduce the NodeRole and the capabilities

akichidis:
## Description 

# Refactor: NodeRole for Observer-Mode Full Nodes

## Problem

With the introduction of block streaming, a new node type arises which we call Observer. This can be perceived as a full node that also spins up the relative consensus components to stream blocks from a peer, process them, build up the DAG and execute the commit rule. Eventually, this node will be able to execute transactions out of the commit path as a validator does. Sui nodes currently have a binary identity: validator (`consensus_config` present) or full node (no `consensus_config`). The consensus layer already supports observer mode, but the sui-node layer has no way to start consensus in observer mode. We need a third mode - **observer-mode full node** - that streams consensus blocks for faster ingestion while retaining full-node capabilities (JSON-RPC, indexing).

## Design: NodeRole with FullNodeSyncMode

`NodeRole` is an **enum with two top-level variants**: `Validator` and `FullNode`. A `FullNode` carries a `FullNodeSyncMode` that determines how it syncs data from the network. This models the real domain relationship: the observer distinction is about *how a fullnode syncs*, not a separate role.

```rust
pub enum FullNodeSyncMode {
    /// Syncs exclusively via the state-sync protocol.
    StateSyncOnly = 0,
    /// Streams consensus blocks for faster ingestion, in addition to state sync.
    ConsensusObserver = 1,
}

pub enum NodeRole {
    /// In the committee — participates in consensus and signs checkpoints.
    Validator,
    /// Not in the committee — serves RPC/indexing and syncs via the configured mode.
    FullNode(FullNodeSyncMode),
}

impl NodeRole {
    // --- Capability methods (primary API for gating behavior) ---
    pub fn runs_consensus(&self) -> bool {
        matches!(self, Validator | FullNode(ConsensusObserver))
    }
    pub fn can_propose_transactions(&self) -> bool { matches!(self, Validator) }
    pub fn should_enable_index_processing(&self) -> bool { matches!(self, FullNode(_)) }
    pub fn should_run_rpc_servers(&self) -> bool { matches!(self, FullNode(_)) }
    ...
}
```

Behavior is gated through **capability methods** on `NodeRole` — code should prefer capability methods over direct variant matching. This means if requirements change (e.g., consensus observers should also run the overload monitor), we flip one method in `NodeRole` instead of hunting through call sites.

| Method                                | Validator | FullNode (StateSyncOnly) | FullNode (ConsensusObserver) |
|:-------------------------------------:|:---------:|:------------------------:|:----------------------------:|
| `runs_consensus()`                    | yes       | no                       | yes                          |
| `can_propose_transactions()`           | yes       | no                       | no                           |
| `should_check_forks()`                | yes       | no                       | yes                          |
| `should_enable_index_processing()`    | no        | yes                      | yes                          |
| `should_run_validator_service()`      | yes       | no                       | no                           |
| `should_run_overload_monitor()`       | yes       | no                       | no                           |
| `should_run_randomness()`             | yes       | no                       | yes                          |
| `should_run_jwk_updater()`            | yes       | no                       | no                           |
| `should_run_execution_time_observer()`| yes       | no                       | no                           |
| `should_run_rpc_servers()`            | no        | yes                      | yes                          |

## Role Determination

`NodeRole` is a computed property on `AuthorityPerEpochStore` — the single source of truth. It's derived per-epoch from:
1. **Committee membership** (dynamic — changes each epoch)
2. **Observer config** (static — whether observer peers are configured in `ConsensusConfig`)

A node with observer peers **promotes to Validator** when added to the committee and **falls back to FullNode(ConsensusObserver)** when removed. All 4 epoch transitions (2×2) are handled.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-config/src/node.rs
+++ b/crates/sui-config/src/node.rs
@@ -85,10 +85,9 @@ pub struct NodeConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub consensus_config: Option<ConsensusConfig>,
 
-    /// The sync mode for full nodes. `None` means the node is configured as a
-    /// validator. `Some(mode)` means it is a full node using the given sync
-    /// strategy. This drives the `intended_node_role()` used for startup
-    /// infrastructure decisions.
+    /// The sync mode for full nodes.
+    /// When `None` is provided and this is a full node then the default is used which is `StateSyncOnly`.
+    /// For validator nodes this is expected to be `None`
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub fullnode_sync_mode: Option<FullNodeSyncMode>,
 
@@ -916,30 +915,24 @@ impl NodeConfig {
     /// create RPC servers or index stores). The authoritative per-epoch role
     /// lives on `AuthorityPerEpochStore::node_role()`.
     pub fn intended_node_role(&self) -> NodeRole {
-        match self.fullnode_sync_mode {
-            Some(mode) => {
-                // Validate that consensus config has been set correctly when the the chosen sync mode has been set to consensus observer
-                if mode == FullNodeSyncMode::ConsensusObserver {
-                    assert!(
-                        self.has_observer_config_peers(),
-                        "Peers list should be set when full node is configured to sync as consensus observer."
-                    );
-                }
-                NodeRole::FullNode(mode)
+        let has_consensus_config = self.consensus_config.is_some();
+
+        match (self.fullnode_sync_mode, has_consensus_config) {
+            (Some(FullNodeSyncMode::ConsensusObserver), _) => {
+                assert!(
+                    self.has_observer_config_peers(),
+                    "Observer peers must be configured when sync mode is ConsensusObserver"
+                );
+                NodeRole::FullNode(FullNodeSyncMode::ConsensusObserver)
             }
-            None => {
-                // If the full node sync mode has not been set, then we need to check whether the node has a consensus config set. If yes, then this is a validator.
-                if self.consensus_config.is_some() {
-                    assert!(
-                        !self.has_observer_config_peers(),
-                        "Peers list should be empty when the node is configured as validator"
-                    );
-                    NodeRole::Validator
-                } else {
-                    // If there is no fullnode sync mode set, and not consensus config set, then this is a full node with state sync only
-                    NodeRole::FullNode(FullNodeSyncMode::StateSyncOnly)
-                }
+            (Some(FullNodeSyncMode::StateSyncOnly), true) => {
+                panic!("Consensus config should not be set for a StateSyncOnly full node");
             }
+            (Some(FullNodeSyncMode::StateSyncOnly), false) => {
+                NodeRole::FullNode(FullNodeSyncMode::StateSyncOnly)
+            }
+            (None, false) => NodeRole::FullNode(FullNodeSyncMode::StateSyncOnly),
+            (None, true) => NodeRole::Validator,
         }
     }
 
--- a/crates/sui-config/src/node.rs
+++ b/crates/sui-config/src/node.rs
@@ -31,6 +31,7 @@ use sui_types::crypto::NetworkKeyPair;
 use sui_types::crypto::SuiKeyPair;
 use sui_types::messages_checkpoint::CheckpointSequenceNumber;
 use sui_types::supported_protocol_versions::{Chain, SupportedProtocolVersions};
+use sui_types::node_role::NodeRole;
 use sui_types::traffic_control::{PolicyConfig, RemoteFirewallConfig};
 
 use sui_types::crypto::{AccountKeyPair, AuthorityKeyPair, get_key_pair_from_rng};
@@ -897,6 +898,34 @@ impl NodeConfig {
 
     pub fn consensus_config(&self) -> Option<&ConsensusConfig> {
         self.consensus_config.as_ref()
+    }
+
+    /// Determines the node role from static configuration.
+    /// This is a preliminary role based on config alone — the authoritative
+    /// role that also accounts for committee membership lives on
+    /// `AuthorityPerEpochStore::node_role()`.
+    pub fn node_role(&self) -> NodeRole {
+        let has_consensus_config = self.consensus_config.is_some();
+        let has_observer_peers = self
+            .consensus_config
+            .as_ref()
+            .and_then(|c| c.parameters.as_ref())
+            .map(|p| !p.observer.peers.is_empty())
+            .unwrap_or(false);
+        match (has_consensus_config, has_observer_peers) {
+            (true, false) => NodeRole::validator(),
+            (true, true) => NodeRole::observer(),
+            _ => NodeRole::fullnode(),
+        }
+    }
+
+    /// Whether this node has observer peers configured in its consensus config.
+    pub fn has_observer_config(&self) -> bool {
+        self.consensus_config
+            .as_ref()
+            .and_then(|c| c.parameters.as_ref())
+            .map(|p| !p.observer.peers.is_empty())
+            .unwrap_or(false)
     }
 
     pub fn genesis(&self) -> Result<&genesis::Genesis> {
--- a/crates/sui-config/src/node.rs
+++ b/crates/sui-config/src/node.rs
@@ -1970,6 +1970,111 @@ remote-store-options:
         std::fs::remove_file(&service_account_file).ok();
         std::fs::remove_file(&aws_key_file).ok();
     }
+
+    mod intended_node_role_tests {
+        use super::*;
+        use crate::ConsensusConfig;
+        use consensus_config::Parameters as ConsensusParameters;
+        use fastcrypto::ed25519::Ed25519KeyPair;
+        use sui_types::node_role::{FullNodeSyncMode, NodeRole};
+
+        fn fullnode_template_config() -> NodeConfig {
+            const TEMPLATE: &str = include_str!("../data/fullnode-template.yaml");
+            serde_yaml::from_str(TEMPLATE).unwrap()
+        }
+
+        fn minimal_consensus_config() -> ConsensusConfig {
+            ConsensusConfig {
+                db_path: PathBuf::from("/tmp/consensus"),
+                db_retention_epochs: None,
+                db_pruner_period_secs: None,
+                max_pending_transactions: None,
+                parameters: Default::default(),
+                listen_address: None,
+                external_address: None,
+            }
+        }
+
+        fn consensus_config_with_observer_peers() -> ConsensusConfig {
+            let mut config = minimal_consensus_config();
+            let kp = Ed25519KeyPair::generate(&mut StdRng::from_seed([0; 32]));
+            let peer = consensus_config::PeerRecord {
+                public_key: consensus_config::NetworkPublicKey::new(kp.public().clone()),
+                address: "/ip4/127.0.0.1/udp/8080".parse().unwrap(),
+            };
+            let mut params = ConsensusParameters::default();
+            params.observer.peers = vec![peer];
+            config.parameters = Some(params);
+            config
+        }
+
+        #[test]
+        fn validator_with_consensus_config() {
+            let mut config = fullnode_template_config();
+            config.consensus_config = Some(minimal_consensus_config());
+            config.fullnode_sync_mode = None;
+
+            assert_eq!(config.intended_node_role(), NodeRole::Validator);
+        }
+
+        #[test]
+        fn fullnode_explicit_state_sync() {
+            let mut config = fullnode_template_config();
+            config.consensus_config = None;
+            config.fullnode_sync_mode = Some(FullNodeSyncMode::StateSyncOnly);
+
+            assert_eq!(
+                config.intended_node_role(),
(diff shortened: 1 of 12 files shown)


crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -466,6 +467,10 @@ pub struct AuthorityPerEpochStore {
     /// Waiters for barrier transactions. Used by execution scheduler to wait for
     /// barrier transaction (keyed by the same AccumulatorSettlement key as settlements).
     barrier_registrations: Arc<Mutex<HashMap<TransactionKey, BarrierRegistration>>>,
+
+    /// Whether the node is configured with observer peers (static from NodeConfig).
+    /// Used together with committee membership to compute NodeRole per epoch.
+    has_observer_config: bool,
```
akichidis: Not super fond of this - I'll try to find something better

aschran:
Just one initial thought based on reading the description:

can a validator be an observer also? if so, what would that mean? if not, would enum variants make more sense than bools (since not all combinations of boolean role are actually valid)?

(i also wonder if you'd ever make an observer that is NOT a full node, but maybe you would? idk)

akichidis:
> Just one initial thought based on reading the description:
> 
> can a validator be an observer also? if so, what would that mean? if not, would enum variants make more sense than bools (since not all combinations of boolean role are actually valid)?
> 
> (i also wonder if you'd ever make an observer that is NOT a full node, but maybe you would? idk)

@aschran thanks for your comment. Actually my initial approach was an `enum` yes. I do not expect an Observer node to not be a Full node as well, but technically that would be possible. Actually, analysing this further it opens the door to what constitutes today a full node. For example a node that:
* uses explicitly checkpoint state sync 
* spins up the RPC interface
* does not participate in consensus - not part of the committee and not proposing blocks) - and also not streaming any blocks
* has the ability to retain data locally for longer compared to a validator node

then yes, that I would highlight as a full node. Now, an Observer is all of the above + it has the ability to stream the consensus blocks and execute transactions out of the commit path. And this is where things get a bit more subtle as the Observer in this case is both `Full Node` (so we want to spin up all/most of the components that a full node does), but also an `Observer` (so we spin up consensus etc).

But following your comment, I believe now that sticking with an `enum` might be more straightforward. This node will no longer identify as `is_fullnode() == true` and `is_observer() == true` , but rather being a mutually exclusive type , which indeed will have some commonalities with a full node, but technically not a simple full node.

akichidis:
Thought this a bit further and following the comments of @aschran and after discussing with @mwtian offline, I think we can do the following:
* stick with two node roles (those could be name types or whatever we feel more appropriate) - validator & fullnode. I've iterated in my mind renaming etc, but this is quite a domain name now that is beyond this codebase alone.
* the key differentiator here is basically the sync method. The "pure" fullnode will only do state sync. A node that has the "observer" mode enabled, it will also sync via consensus. Thus I've added the `FullNodeSyncMode` . We never expect - or at least there is no plan currently - having a node that is an Observer but not a full node. 
* The capabilities could unlock the potential of providing finer grained node configuration, but that's not now on the immediate goals

how we codify the capabilities could be done in various ways, but I am now sticking to simple approach.

I've updated now the PR description/design and the code as well.

Let me know your thoughts

crates/sui-types/src/node_role.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    /// Whether this node can propose transactions and checkpoint signatures to consensus.
+    pub fn can_propose_transactions(&self) -> bool {
+        matches!(self, Self::Validator)
+    }
+
+    /// Whether this node should run fork detection and recovery at startup.
+    // Fork detection and recovery is only relevant for validators and full nodes that run in observer mode.
+    // Fullnodes that run in checkpoint state sync mode only don't need fork checking
+    pub fn should_check_forks(&self) -> bool {
```
mwtian: Why not use runs_consensus()?

crates/sui-types/src/node_role.rs
```diff
@@ -0,0 +1,146 @@
...
+
+impl NodeRole {
+    pub fn is_fullnode(&self) -> bool {
+        matches!(self, Self::FullNode(_))
+    }
+
+    pub fn is_validator(&self) -> bool {
+        matches!(self, Self::Validator)
+    }
+    // --- Capability methods ---
```
mwtian, line 56: It seems each capability evaluates to one of the condition below. Also `matches!(self, Self::Validator)` is equivalent to `!matches!(self, Self::FullNode(_))`.
```
matches!(
    self,
    Self::Validator | Self::FullNode(FullNodeSyncMode::ConsensusObserver)
)

matches!(self, Self::Validator)

matches!(self, Self::FullNode(_))
```
I wonder if it is better to just have a few functions which evaluate these conditions. We may not want to add a new function here each time a condition is added that needs to evaluate NodeRole.

crates/sui-core/src/authority.rs
```diff
@@ -971,12 +972,16 @@ pub struct AuthorityState {
 ///
 /// Repeating valid commands should produce no changes and return no error.
 impl AuthorityState {
+    pub fn node_role(&self, epoch_store: &AuthorityPerEpochStore) -> NodeRole {
+        epoch_store.node_role()
+    }
+
     pub fn is_validator(&self, epoch_store: &AuthorityPerEpochStore) -> bool {
-        epoch_store.committee().authority_exists(&self.name)
```
mwtian, line 975: This seems like a more precise condition for deciding if a node is a validator, with our current definition of validators.

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3842,9 +3850,21 @@ impl AuthorityPerEpochStore {
             .get_observations()
     }
 
+    /// Returns the role of this node for the current epoch, computed from
+    /// committee membership and observer configuration.
+    pub fn node_role(&self) -> NodeRole {
+        if self.committee.authority_exists(&self.name) {
```
mwtian: This should be the source of truth to decide if a node is validator.

crates/sui-config/src/node.rs
```diff
@@ -899,6 +900,34 @@ impl NodeConfig {
...
+        let has_consensus_config = self.consensus_config.is_some();
+        let has_observer_peers = self
+            .consensus_config
+            .as_ref()
+            .and_then(|c| c.parameters.as_ref())
+            .map(|p| !p.observer.peers.is_empty())
+            .unwrap_or(false);
+        match (has_consensus_config, has_observer_peers) {
+            (true, false) => NodeRole::Validator,
+            (true, true) => NodeRole::FullNode(FullNodeSyncMode::ConsensusObserver),
```
mwtian: Tbh this is not super intuitive to me. It seems setting observer config has unintended consequences on a validator.

crates/sui-node/src/lib.rs
```diff
@@ -465,8 +466,9 @@ impl SuiNode {
         }
 
         let run_with_range = config.run_with_range;
-        let is_validator = config.consensus_config().is_some();
-        let is_full_node = !is_validator;
+        // Preliminary role from config — used for infrastructure decisions before
+        // the epoch store exists. The authoritative role is on AuthorityPerEpochStore.
+        let preliminary_role = config.node_role();
```
mwtian: As mentioned earlier, I feel using config to derive node's role can be a bit unexpected. How about checking if the node is in latest committee?

crates/sui-types/src/node_role.rs
akichidis: Yes we could do that as well

crates/sui-types/src/node_role.rs
akichidis: We don't necessarily need to add a new function every time. I admit that I went a bit beyond on that aspect. Let me reduce the usage.

crates/sui-node/src/lib.rs
akichidis: Yes, but to avoid changing too much here, I've followed pretty much what we already used to do before. The node it self - for the initial bootstrap - was using the config to derive whether a node is a validator or a full node so I've followed that approach for the time being. 

The reason why the `preliminary_role` has been used is because there are a few decisions that need to happen before we even have the epoch store available to determine whether this is a validator or not. That being said, I am going to look at whether we can rearrange any of those dependencies so we can use the `committee`

crates/sui-config/src/node.rs
akichidis: Not the observer config, but the observer peers specifically. A validator has no reason to do that. But yes, config can be considered a bit sensitive here.

crates/sui-node/src/lib.rs
akichidis: Also, as you can see , on the rest of the code base the node role is derived from the epoch store where the committee check takes place.

akichidis: Ok @mwtian , I've done the change but I believe this is now surfacing some more delicate existing issues around it. Without my changes, a node type when the node bootstraps is determined by the existence of the `consensus_config` or not. If it's present then it is a `validator`. If not, then it is a `fullnode`.

This creates some edge cases where a node is configured with `consensus_config` present (as validator), but if it's not in the committee yet it is considered a `FullNode`:
* The node starts as `Full Node` and brings up the `indexer, transaction orchestrator, subscription checkpoint service, RPC APIs etc`
* Node synchronizes and in theory everything works fine
* Once it reconfigures and becomes a validator. None of the previous full node components shutdown. Not even any DB table optimisation configurations get reconfigured.
* Eventually the node might crash from the `subscription checkpoint service` as now checkpoints are not processed via state sync solely, but from commit output as well (there are some checks in the subscription checkpoint service that cause that - that's why some simtests fail).

Now, the above panic of course can be fixed, that's not the problem. But the bigger picture here is that if we go down this path we are basically changing the earlier assumptions  regarding the node initialisation. Before someone would start the node to `intentionally run as a validator due to the config` , even though not being in committee yet. The node should checkpoint state sync and eventually once reconfigured, all the right components would have been initialised - nothing from the FullNode side. We are now breaking this assumption. And it can even have some undesired effects (someone running as a full node with the RPC API etc).

That being said, I don't believe that existing behaviour is really correct either. But this gets a bit out of scope of this PR. I would prefer introducing something that would not mess up with the existing logic and tackle separately the "proper" node initialisation and transition `validator<->fullnode` with clear expectations. Although those transactions should be quite rare, we can handle them more appropriately

crates/sui-config/src/node.rs
```diff
@@ -883,6 +883,10 @@ impl NodeConfig {
         self.db_path.join("db_checkpoints")
     }
 
+    pub fn store_db_path(&self) -> PathBuf {
```
mwtian: nit: maybe `db_store_path()` to align with the concatenation order.

crates/sui-core/src/authority/authority_per_epoch_store.rs
mwtian: I think epoch store can keep a copy of the NodeConfig, if it is not the case already.

crates/sui-node/src/lib.rs
```diff
@@ -2027,18 +2099,46 @@ impl SuiNode {
...
         checkpoint_store: &CheckpointStore,
         checkpoint_metrics: &CheckpointMetrics,
-        is_validator: bool,
+        should_check_forks: bool,
         fork_recovery: Option<&ForkRecoveryConfig>,
     ) -> Result<()> {
-        // Fork detection and recovery is only relevant for validators
-        // Fullnodes should sync from validators and don't need fork checking
-        if !is_validator {
+        if !should_check_forks {
```
mwtian: I wonder if we should just skip calling this function when `should_check_forks` is `false`.

crates/sui-types/src/node_role.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Whether this node participates in the consensus protocol.
+    pub fn runs_consensus(&self) -> bool {
+        matches!(
+            self,
+            Self::Validator | Self::FullNode(FullNodeSyncMode::ConsensusObserver)
+        )
+    }
+
+    /// Whether this node can propose transactions and checkpoint signatures to consensus.
+    pub fn can_propose_transactions(&self) -> bool {
```
mwtian: I think the callsite can use `is_validator()`.

crates/sui-types/src/node_role.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    /// Whether this node should create index stores for JSON-RPC and REST API.
+    pub fn should_enable_index_processing(&self) -> bool {
+        matches!(self, Self::FullNode(_))
+    }
+
+    /// Whether this node should expose HTTP/RPC servers (JSON-RPC, REST).
+    pub fn should_run_rpc_servers(&self) -> bool {
+        matches!(self, Self::FullNode(_))
+    }
```
mwtian, line 74: I think the callsite can use just `is_fullnode()`

mwtian (approved):

PR #26401, opened 2026-04-28
[forking 13/n] Seeding objects design doc

stefan-mysten:
## Description 

This PR proposes a design for seeding objects at startup.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-forking/seeding_design.md
+++ b/crates/sui-forking/seeding_design.md
@@ -1,0 +1,210 @@
+# Seeding Design
+
+## Problem
+
+The forking tool can execute transactions against forked state but cannot answer
+"what objects does address X own?" without explicit rpc querying at startup / before first transaction.
+For example, if a user would like to execute a transaction, the forked network does not have
+yet the ownership information - which gas coins it owns. To enable this, seeding
+allows the tool to resolve ownership information at startup and build an initial ownership index.
+
+Seeding resolves object refs (id, version, digest) + lightweight metadata (owner, type)
+at startup, establishing the data needed to build an ownership index.
+
+Object contents are NOT fetched during seeding for address-based seeds. Content
+fetching is deferred to the existing `get_object_from_remote()` path used by
+execution. For explicit `--object` IDs, we reuse the existing `multiGetObjects`
+infrastructure which does fetch BCS — a useful side-effect since these objects
+will likely be accessed during execution anyway.
+
+### Checkpoint age constraint
+
+- **Recent checkpoints (<1h)**: GraphQL `address.objects()` works — can seed by address
+- **Older checkpoints**: Only individual object lookups work — must seed by object ID
+
+## CLI
+
+Two new flags on the `start` command:
+
+```
+--address <ADDR>     Seed all owned objects for this address (repeatable)
+--object <ID>        Seed a specific object by ID (repeatable)
+```
+
+Both can be combined. Addresses are resolved into object refs; explicit
+object IDs are fetched via the existing `multiGetObjects` path. Results are
+merged and deduped by object ID. It is important to note if addresses is passed,
+the checkpoint must be within the last 1h.
+
+When restarting a fork with the same `--data-dir` and `--checkpoint`, the existing
+seeding information that is stored on disk will be reused. If `--address` or `--object` is
+provided, the tool will error with a message indicating that a seed manifest already exists.
+
+## Generated files
+
+One file is written to `{data_dir}/{network}/forked_at_{checkpoint}/`:
+
+### `seed_manifest.json` — internal, full metadata for index rebuilding
+
+```json
+{
+    "network": "testnet",
+    "checkpoint": 12345678,
+    "entries": [
+        {
+            "object_id": "0x...",
+            "version": 42,
+            "digest": "...",
+            "owner": "0x...",
+            "object_type": "0x2::coin::Coin<0x2::sui::SUI>"
+        }
+    ]
+}
+```
+
+## Seed resolution workflow
+
+The seed manifest is immutable once written.
+
+Seed resolution:
+   - seed_input provided + manifest exists on disk → ERROR:
+     "A seed manifest already exists at <path>. To fork the same
+      checkpoint with different seeds, use a different --data-dir."
+   - seed_input provided, no existing manifest → resolve → write manifest + generated file
+   - no seed_input, manifest exists on disk → read it (restart case)
+   - no seed_input, no manifest → proceed without seeds
+
+## GraphQL queries
+
+### Address-owned objects query (new)
+
+Uses `Checkpoint.query` scoping (same pattern as `VersionAtCheckpointQuery`):
+
+```graphql
+query($sequenceNumber: UInt53, $address: SuiAddress!, $first: Int, $after: String) {
+  checkpoint(sequenceNumber: $sequenceNumber) {
+    query {
+      address(address: $address) {
+        objects(first: $first, after: $after) {
+          nodes {
+            address
+            version
+            digest
+            owner {
+              ... on AddressOwner { address { address } }
+              ... on ConsensusAddressOwner { address { address } }
+            }
+            contents { type { repr } }
+          }
+          pageInfo { hasNextPage endCursor }
+        }
+      }
+    }
+  }
+}
+```
+
+- Page size: 50
+- Paginate with cursor loop (same pattern as `events_query`)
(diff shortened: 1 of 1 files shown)


crates/sui-forking/seeding_design.md
```diff
@@ -0,0 +1,210 @@
+# Seeding Design
+
+## Problem
+
+The forking tool can execute transactions against forked state but cannot answer
+"what objects does address X own?" without explicit rpc querying at startup / before first transaction.
```
jordanjennings-mysten, line 6: I think objects are well contained enough that if you were not aware of an object its version and state would stay static allowing you to lazy load the object at any point in the future no?

crates/sui-forking/seeding_design.md
```diff
@@ -0,0 +1,210 @@
...
+For example, if a user would like to execute a transaction, the forked network does not have
+yet the ownership information - which gas coins it owns. To enable this, seeding
```
jordanjennings-mysten, line 8: instead of seeding can we not just pass that ownership query to the fullnode at fork checkpoint and lazy load it in? There is some lazy loading going on already so would you not need to implement similar patterns to lazy load owned objects? Either you load the package first via lazy load or you load the owned object first. The package can exist without the owned object but can the owned object exist without the package? or would that create an inconsistency with your data store

crates/sui-forking/seeding_design.md
```diff
@@ -0,0 +1,210 @@
...
+Object contents are NOT fetched during seeding for address-based seeds. Content
+fetching is deferred to the existing `get_object_from_remote()` path used by
+execution. For explicit `--object` IDs, we reuse the existing `multiGetObjects`
+infrastructure which does fetch BCS — a useful side-effect since these objects
+will likely be accessed during execution anyway.
+
+### Checkpoint age constraint
+
+- **Recent checkpoints (<1h)**: GraphQL `address.objects()` works — can seed by address
+- **Older checkpoints**: Only individual object lookups work — must seed by object ID
```
jordanjennings-mysten, line 23: I'm guessing you return an error if a user tries to seed an address at an older checkpoint

crates/sui-forking/seeding_design.md
```diff
@@ -0,0 +1,210 @@
...
+```
+
+Both can be combined. Addresses are resolved into object refs; explicit
+object IDs are fetched via the existing `multiGetObjects` path. Results are
+merged and deduped by object ID. It is important to note if addresses is passed,
+the checkpoint must be within the last 1h.
+
+When restarting a fork with the same `--data-dir` and `--checkpoint`, the existing
+seeding information that is stored on disk will be reused. If `--address` or `--object` is
+provided, the tool will error with a message indicating that a seed manifest already exists.
```
jordanjennings-mysten, line 41: This does have me thinking about the git branch concept again but I think it requires a bit too much reworking towards a content addressable storage system to do at this point.

crates/sui-forking/seeding_design.md
stefan-mysten: No, you cannot just pass that to a fullnode. We're avoiding working directly with fullnodes and using GraphQL to query for on-chain data.

GraphQL only holds ownership information for the past 1h. If you're trying to query _give me all the owned objects of this address 3 days ago_, you will get an error message.

We need to at least fetch the object id and version so that we can lazy load it when we need it (e.g., when a specific address makes a tx).

crates/sui-forking/seeding_design.md
stefan-mysten: yes!

crates/sui-forking/seeding_design.md
stefan-mysten: Definitely that's something we should think later. For now, for the POC, this seems to be the easiest thing to do.

crates/sui-forking/seeding_design.md
stefan-mysten: For objects, yes, not a problem. But if you want to execute a tx and you need to get a gas coin, then you need ownership information: what coins does this address own? And that can change (+ see the next comment)

jordanjennings-mysten (approved):

crates/sui-forking/seeding_design.md
```diff
@@ -0,0 +1,210 @@
...
+{
+    "network": "testnet",
+    "checkpoint": 12345678,
+    "entries": [
+        {
+            "object_id": "0x...",
+            "version": 42,
+            "digest": "...",
+            "owner": "0x...",
+            "object_type": "0x2::coin::Coin<0x2::sui::SUI>"
```
amnn, line 59: Is there a reason to store this information? This technically requires fetching the object contents.

crates/sui-forking/seeding_design.md
```diff
@@ -0,0 +1,210 @@
...
+        objects(first: $first, after: $after) {
+          nodes {
+            address
+            version
+            digest
+            owner {
+              ... on AddressOwner { address { address } }
+              ... on ConsensusAddressOwner { address { address } }
+            }
+            contents { type { repr } }
```
amnn, line 97: Related to my question above, in this query, fetching the address seems redundant seeing as you are passing it in to begin with. When adding an object by its ID, you would need to fetch its owner, but here you do not.

crates/sui-forking/seeding_design.md
```diff
@@ -0,0 +1,210 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+- Page size: 50
+- Paginate with cursor loop (same pattern as `events_query`)
+- Only collect `AddressOwner` and `ConsensusAddressOwner` entries -- not really sure about ConsensusAddressOwner, probably we don't need it.
```
amnn, line 109: You do need `ConsensusAddressOwner`

crates/sui-forking/seeding_design.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Individual objects (reuse existing)
+
+For explicit `--object` IDs, reuse the existing `object_query::query()`
+infrastructure (`multiGetObjects` with `ObjectFragment` that fetches `objectBcs`).
+BCS gets cached to disk. After fetching, extract metadata (owner, type, version,
```
amnn, line 115: I do not think it makes sense to re-use the machinery that seeds the on-disk store for this -- just write a query that gives you the precise information you want.

crates/sui-forking/seeding_design.md
stefan-mysten: I think it's useful because we don't need to deserialize bcs objects to find out their type when a get owned objects query with an object type filter comes through. By default, `ListOwnedObjects` read mask is set to object_id,version,object_type, so bcs is not fetched by default, but it requires object type.

This way, there's more data needed at startup, but less to do later because we stored this metadata in the `OwnedObjectEntry` type.

stefan-mysten: So I guess it's mostly a tradeoff between on-demand owned object fetching and preemptively fetching the data. I don't have a strong opinion on one over the other, though I recall you had a slight preference for the lazy option (store object refs, fetch on demand - when `ListOwnedObjects` RPC is called).

crates/sui-forking/seeding_design.md
stefan-mysten: Fixed in this PR: https://github.com/MystenLabs/sui/pull/26635

crates/sui-forking/seeding_design.md
stefan-mysten: I put up a PR here that removes this and only keeps object ref + owner for seeding and owned object index.
https://github.com/MystenLabs/sui/pull/26673

amnn: > I think it's useful because we don't need to deserialize bcs objects to find out their type when a get owned objects query with an object type filter comes through.

This is not really the purpose of the seed though, right? This is the purpose of the index that you create with this seed. I agree that your owned object **index** needs to take into account type information, so that you're not just loading every object to filter it by type.

Also note that object types are not fixed per ID, (e.g. for dynamic fields today, and soon for derived objects, and maybe for all objects with the upcoming ID leak verifier changes).

stefan-mysten: You're right, I've been slightly conflating the two and keeping them "in-sync". I will revise this in the PR.

PR #26409, opened 2026-04-28
Optimize tracing overhead: global callsite filter, disable span latency layer

mystenmark:
## Summary

- **Disable `PrometheusSpanLatencyLayer`** in sui-node via `with_disable_span_latency` config
- **Add `GlobalLevelFilter`** — rejects trace/debug callsites at `register_callsite()` time via `Interest::never()`, eliminating per-call `enabled()` walk overhead from per-layer filtering
- **Extend global filter to events** — also rejects event callsites above the EnvFilter's max level
- **Remove `enabled()` override** — avoid adding to the filter walk for passing callsites
- **Support `TOKIO_SPAN_LEVEL=off`** — change span_level to `LevelFilter` so `off` is valid

## Test plan

- [x] `cargo nextest run -p telemetry-subscribers` — 6/6 pass
- [x] `cargo check -p sui-node`

### DIFF
--- a/crates/telemetry-subscribers/src/lib.rs
+++ b/crates/telemetry-subscribers/src/lib.rs
@@ -42,26 +42,65 @@ mod file_exporter;
 pub mod span_latency_prom;
 mod test_layer;
 
-/// Global filter layer that rejects spans above a given level at callsite registration time.
+/// Global filter layer that rejects callsites above the configured levels at registration time.
 ///
 /// Without this, per-layer filtering causes the Registry to return `Interest::always()` for
 /// every callsite. That means trace-level `#[instrument]` spans are allocated in the slab,
 /// have their per-layer filters walked, and are immediately discarded — all at significant cost.
+/// The same applies to events below the env filter threshold.
 ///
-/// By rejecting high-verbosity span callsites globally, `Span::new` short-circuits to
-/// `Span::none()` for free.
-///
-/// Events are always passed through so that per-layer log filters work as before.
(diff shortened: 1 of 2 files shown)


crates/telemetry-subscribers/src/lib.rs
```diff
@@ -564,6 +672,18 @@ impl TelemetryConfig {
...
+        // callsites above the env filter's max level at registration time, preventing
+        // the Registry from dispatching callsites that every per-layer filter would
+        // immediately discard.
+        layers.push(
+            GlobalLevelFilter {
+                max_span_level: span_level,
+                max_event_level,
+            }
+            .boxed(),
+        );
```
bmwill: I don't think this is doing what you want it to. Today we don't "stack" any layers. Instead we create a Vec<dyn Layer> and pass that whole set to be registered. When an event is triggered it is fanned out to all of the layers we have configured. 

I think what you want instead is to take the Vec<dyn Layers> we have, and then stack them below this global filter so that the global filter can disable all the callsites we never want to see, and then have events/spans fanned out to the list of layers we have configured.

bmwill: Relavent code for how a Vec of layers is treated: https://docs.rs/tracing-subscriber/latest/src/tracing_subscriber/layer/mod.rs.html#1776-1893

right now they're in parallel but I think you want the global filter in series

bmwill (approved):

PR #26414, opened 2026-04-28
[graphql-stream] Recover from upstream gRPC gaps via kv-rpc [7/n]

tpham-mysten:
## Description 

* `CheckpointStreamTask` now reconnects on upstream stream errors and detects gaps when the first message after reconnect has a sequence number past the last broadcast. Missing checkpoints are fetched from kv-rpc (`LedgerService`) and broadcast in order before the live message, so subscribers see strictly contiguous, in-order checkpoints across disconnects.
* New `gap_recovery` module with a `CheckpointFetcher` trait, `recover_gap` driver (chunked, watermark-gated per chunk, infinite transient retry), and unit tests covering retry, chunk semantics, and watermark progression.
* Reconnect uses bounded exponential backoff. Permanent gRPC errors (`Unauthenticated`, `PermissionDenied`, `InvalidArgument`, `Unimplemented`) terminate the task; everything else retries indefinitely so the streaming server stays alive through upstream outages.

See the module-level doc at the top of `checkpoint_stream_task.rs` for the algorithm and design rationale.

**Out of scope (follow-ups)**:

 * Metrics for gap recovery (counter of attempts, failures, stuck-seconds gauge). TODO marker in the code.

## Test plan 

 - Unit tests in `gap_recovery::tests` — 7 tests covering `fetch_one_with_retry`, `fetch_chunk`, and `recover_gap` (including a watermark-progression scenario).
 - `cargo xclippy` clean.
 - `cargo fmt` applied.

## Stack
- #26019
- #26094
- #26117
- #26170
- #26194
- #26202
- #26414

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
+++ b/crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
@@ -310,10 +310,11 @@ impl CheckpointStreamTask {
         Ok(())
     }
 
-    /// Index packages from the checkpoint into the streaming store, signal eviction, then
-    /// process and broadcast. Indexing is needed because `kv_packages` may not yet have
-    /// these packages at stream time; recovered checkpoints from `recover_gap` skip this
-    /// step because the watermark gate guarantees they're already in the DB.
+    /// Index packages from the checkpoint into the streaming store, signal eviction,
+    /// then process and broadcast. Indexing exposes the checkpoint's packages to
+    /// subscribers immediately, ahead of `kv_packages`. Recovered checkpoints from
+    /// `recover_gap` skip this step because the gate already waits for both
+    /// `ledger_grpc` and `kv_packages` to catch up.
     fn index_and_broadcast(&self, checkpoint: ProtoCheckpoint, seq: u64) -> anyhow::Result<()> {
         let packages = extract_packages(&checkpoint);
         if !packages.is_empty() {
--- a/crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
+++ b/crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
@@ -1,15 +1,72 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+//! Checkpoint stream task.
+//!
+//! Maintains the connection to the fullnode's `SubscribeCheckpoints` gRPC
+//! stream and broadcasts every checkpoint, in order, to GraphQL subscribers.
+//!
+//! ```text
+//!   loop forever:
+//!     connect (retry on transient errors, fail on auth/config)
+//!     for each incoming checkpoint:
+//!       if seq > last_broadcast + 1:
+//!         recover_gap(last_broadcast + 1 ..= seq - 1)
+//!       broadcast(checkpoint)
+//!     on stream end: reconnect
+//! ```
+//!
+//! # Reconnect
+//!
+//! gRPC connections drop for many reasons: rolling deploys, network blips,
+//! server restarts. When the stream errors or closes, the task reconnects
+//! with bounded exponential backoff. Transient errors retry indefinitely;
+//! permanent gRPC errors (auth, config) terminate the task.
+//!
+//! # Gap recovery
+//!
+//! `SubscribeCheckpoints` resumes at the current tip on reconnect, not where
+//! we left off. So after a drop, the first new message typically has a
+//! sequence number well past the last we broadcast — checkpoints produced
+//! during the drop are now invisible on the live stream.
+//!
+//! Detection happens on every incoming message: if `seq > last_broadcast + 1`,
+//! the task fetches the missing checkpoints from kv-rpc (a `LedgerService`
+//! endpoint serving historical checkpoints) and broadcasts them in order
+//! before broadcasting the live message.
+//!
+//! ```text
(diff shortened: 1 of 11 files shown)


crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -151,47 +225,149 @@ impl CheckpointStreamTask {
...
+            *last_broadcast = Some(seq);
+        }
+
+        Ok(())
+    }
+
+    /// Index packages from the checkpoint into the streaming store, signal eviction, then
+    /// process and broadcast. Indexing is needed because `kv_packages` may not yet have
+    /// these packages at stream time; recovered checkpoints from `recover_gap` skip this
+    /// step because the watermark gate guarantees they're already in the DB.
```
nickvikeras: I think you are only waiting on the ledger service's watermark, but kv package data is still loaded from Postgres. Do you also need to gate on the Postgres package pipeline's watermark for this assumption to hold true?

nickvikeras (commented):
Backoffs and concurrency limits and stuff LGTM. Left a comment with one thing to double check.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Ah right thanks for the catch. I have then gated the kv_packages watermark too and added the tests for it

tpham-mysten:
Thanks @nickvikeras for the catch. I have added the gate for kv_packages

nickvikeras:
> Thanks @nickvikeras for the catch. I have added the gate for kv_packages

Just FYI, we are planning to eventually move the KV Package data to BigTable. I already have the data stored in BigTable, but need to find time to expose it over gRPC so GraphQL can use that as the source of this data instead of Postgres.

Once that is done, we can remove this extra gate you added in the last commit.

nickvikeras (approved):

crates/sui-indexer-alt-graphql/src/task/streaming/gap_recovery.rs
```diff
@@ -0,0 +1,501 @@
...
+/// Block until both indexer pipelines that gap recovery depends on have caught up to
+/// `target`: `ledger_grpc` (kv-rpc, serves checkpoint contents) and `kv_packages`
+/// (Postgres, serves package resolution from the DB). Recovered checkpoints don't go
+/// through `index_and_broadcast`, so subscribers resolving their packages fall through
+/// to the DB and need `kv_packages` to be ready.
+async fn wait_for_pipelines_catching_up_at(
+    target: u64,
+    watermarks_rx: &watch::Receiver<Arc<Watermarks>>,
+) -> anyhow::Result<()> {
+    let mut rx = watermarks_rx.clone();
```
amnn, line 105: nit: do you need to perform the clone here? Or could you pass in a `&mut Reciever` here, and keep a single clone of the receiver at the level of `recover_gap`?

crates/sui-indexer-alt-graphql/src/task/streaming/gap_recovery.rs
```diff
@@ -0,0 +1,501 @@
...
+/// Recover the gap `lo..=hi` by reading checkpoints from kv-rpc, processing them through the
+/// existing streaming pipeline, and broadcasting in order. Each chunk waits for the kv-rpc
+/// indexer to reach the chunk's upper bound before fetching, so partial progress is broadcast
+/// as the indexer catches up rather than waiting for the entire gap to be available.
+pub(crate) async fn recover_gap<F: CheckpointFetcher>(
+    fetcher: &F,
+    watermarks_rx: &watch::Receiver<Arc<Watermarks>>,
+    sender: &broadcast::Sender<Arc<ProcessedCheckpoint>>,
+    lo: u64,
+    hi: u64,
```
amnn, line 66: nit: conventionally this would be `hi_inclusive` in this codebase so that we distinguish between inclusive and exclusive bounds (by default `lo` is inclusive, and `hi` is exclusive, and we would use `lo_exclusive` and `hi_inclusive` if it's not the case to make the distinction obvious and off-by-one errors less likely.

crates/sui-indexer-alt-graphql/src/task/streaming/gap_recovery.rs
```diff
@@ -0,0 +1,501 @@
...
+        /// Always return `Ok(Some(make_test_proto_checkpoint(seq)))`.
+        Success,
+        /// Return `Err` for the first N calls, then `Ok(Some(...))` afterward.
+        ErrorThenSuccess(usize),
+        /// Return `Ok(None)` for the first N calls, then `Ok(Some(...))` afterward.
+        NoneThenSuccess(usize),
+    }
+
+    struct MockFetcher {
+        state: Mutex<HashMap<u64, (FetcherBehavior, usize)>>,
```
amnn, line 254: nit: DashMap

crates/sui-indexer-alt-graphql/src/task/streaming/gap_recovery.rs
tpham-mysten: That's better. Done!

tpham-mysten (commented):
Thanks @amnn! I have resolved these comments in #26476

PR #26419, opened 2026-04-29
Remove pre-chunking execution time observation code path

alex-mysten:
## Description

Observation chunking has been enabled since protocol version 102, and the `latest` execution layer only runs at protocol version 119+, so the pre-chunking observation storage path is unreachable there.

`sui-execution/v3` still serves the older versions and keeps both paths. `sui-core` is not version-gated, so when chunking is off it now starts the epoch with no stored observations instead of reading the legacy single-key dynamic field.

## Test plan

Covered by existing execution-time-estimator and congestion-control tests; their fixtures move to `observations_chunk_size: Some(18)` to match mainnet.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-adapter/src/execution_engine.rs
+++ b/sui-execution/latest/sui-adapter/src/execution_engine.rs
@@ -1126,16 +1126,14 @@ pub(crate) mod checked {
                             if let PerObjectCongestionControlMode::ExecutionTimeEstimate(params) =
                                 protocol_config.per_object_congestion_control_mode()
                             {
-                                if let Some(chunk_size) = params.observations_chunk_size {
-                                    builder = setup_store_execution_time_estimates_v2(
-                                        builder,
-                                        estimates,
-                                        chunk_size as usize,
-                                    );
-                                } else {
-                                    builder =
-                                        setup_store_execution_time_estimates(builder, estimates);
-                                }
+                                let chunk_size = params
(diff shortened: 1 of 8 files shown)


sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -1126,16 +1126,14 @@ pub(crate) mod checked {
...
-                                        estimates,
-                                        chunk_size as usize,
-                                    );
-                                } else {
-                                    builder =
-                                        setup_store_execution_time_estimates(builder, estimates);
-                                }
+                                let chunk_size = params
+                                    .observations_chunk_size
+                                    .expect("observation chunking is enabled at all protocol versions handled by this execution layer");
```
aschran, line 1131: this won't break any old epoch replays right?

aschran (approved):

sui-execution/latest/sui-adapter/src/execution_engine.rs
alex-mysten: no - execution version v4 was enabled at 119, chunking enabled at 102

PR #26429, opened 2026-04-29
Drop live_owned_object_markers table

andll:
## Summary

Two related cleanups:

1. **Drop the `live_owned_object_markers` (column family `owned_object_transaction_locks`) table.** Its `LockDetailsWrapperDeprecated` value was already deprecated; only the key was meaningful, as a "live owned object" set. Every read it served is redundant with the `objects` table, and the cache hot path for those readers already used `objects` directly — only the disk fallback used markers.
2. **Remove the redundant `check_owned_locks` recheck inside `process_certificate`.** With post-consensus locking, the safety property this check was guarding is already established earlier in the pipeline.

## Why `check_owned_locks` is gone

The check ran inside `process_certificate` right before execution and re-verified that every owned input ObjectRef still pointed to the latest live version. With post-consensus locking the same guarantee is established earlier and held until execution:

- `try_acquire_owned_object_locks_post_consensus` (`consensus_handler.rs:2871`) runs for every `UserTransactionV2` before the certificate proceeds, and blocks any conflicting tx for the same owned `(id, version, digest)`.
- `check_certificate_input` runs at the top of `process_certificate` and re-resolves input objects against the cache; a stale input fails there with the same `ObjectVersionUnavailableForConsumption` / `ObjectNotFound` errors.

So two transactions can no longer race to consume the same owned object version, and a stale ObjectRef would have been caught upstream. The hot path was paying a `compute_object_reference` (BCS-encode + Blake2b256) per owned input per cert with no failure mode left for it to catch.

This commit removes:
- `AuthorityState::check_owned_locks` + its call in `process_certificate`
- `ObjectCacheRead::check_owned_objects_are_live` trait method
- `WritebackCache` and `AuthorityStore` impls of the same

## Why dropping `live_owned_object_markers` is safe

- `check_owned_objects_are_live` is now removed entirely (see above), so its disk fallback no longer needs the marker table.
- `get_lock` (which fed `ObjectInfoResponse.lock_for_debugging`) is rewritten to read liveness via `get_latest_object_ref_if_alive` against the `objects` table; the locking tx digest still comes from the per-epoch `owned_object_locked_transactions` table.
- The marker init/delete writes in genesis insert, bulk insert, snapshot restore, and the commit path are removed.
- The field declaration and the `owned_object_transaction_locks` entry in both RocksDB and tidehunter `open()` paths are removed.
- `LockDetailsWrapperDeprecated` enum + `owned_object_transaction_locks_table_config` helper removed.

## Notes for reviewers

- **Migration**: tidehunter does not support keyspace deletion, so existing tidehunter DBs will keep `owned_object_transaction_locks` on disk but the binary will no longer reference it. RocksDB nodes will leave the CF on disk after upgrade (typed-store's `populate_missing_cfs` opens unknown-on-disk CFs with default options); a separate cleanup task can drop it.
- **Orphan fields**: `TransactionOutputs.locks_to_delete` and `new_locks_to_init` are now computed but unused in production. Left in place to keep the diff focused — removal touches `writeback_cache_tests` mocks.
- **Background**: per-conversation analysis confirmed the marker table's only readers were `check_owned_locks` (now removed as redundant) and `lock_for_debugging` (`ObjectInfoResponse`, consumed only by `sui-tool`).

## Test plan

- [x] `cargo check -p sui-core` (RocksDB + `USE_TIDEHUNTER=1 ... --features typed-store/tidehunter`)
- [x] `cargo nextest run -p sui-core --lib` — all tests pass (one flake in `jsonrpc_index::tests::test_index_cache` from a parallel RocksDB lock race; passes in isolation, unrelated)
- [x] `cargo nextest run -p sui-core --lib authority_tests transaction_tests` — 129/129 pass
- [x] `cargo clippy -p sui-core --all-targets -- -D warnings` — clean
- [x] private-testnet deploy at 12K TPS w/ tidehunter enabled
- [ ] e2e: `cargo simtest -p sui-e2e-tests`

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -1682,11 +1682,6 @@ impl AuthorityState {
             .expect("must return correct number of effects")
     }
 
-    fn check_owned_locks(&self, owned_object_refs: &[ObjectRef]) -> SuiResult {
-        self.get_object_cache_reader()
-            .check_owned_objects_are_live(owned_object_refs)
-    }
-
     /// This function captures the required state to debug a forked transaction.
     /// The dump is written to a file in dir `path`, with name prefixed by the transaction digest.
     /// NOTE: Since this info escapes the validator context,
@@ -2001,10 +1996,6 @@ impl AuthorityState {
             Err(e) => return ExecutionOutput::Fatal(e),
         };
 
-        let owned_object_refs = input_objects.inner().filter_owned_objects();
-        if let Err(e) = self.check_owned_locks(&owned_object_refs) {
-            return ExecutionOutput::Fatal(e);
-        }
         let tx_digest = *certificate.digest();
         let protocol_config = epoch_store.protocol_config();
         let transaction_data = &certificate.data().intent_message().value;
--- a/crates/sui-core/src/authority/authority_store.rs
+++ b/crates/sui-core/src/authority/authority_store.rs
@@ -22,7 +22,7 @@ use move_core_types::resolver::{ModuleResolver, SerializedPackage};
 use serde::{Deserialize, Serialize};
 use sui_config::node::AuthorityStorePruningConfig;
 use sui_macros::fail_point_arg;
-use sui_types::error::{SuiErrorKind, UserInputError};
+use sui_types::error::UserInputError;
 use sui_types::execution::TypeLayoutStore;
 use sui_types::global_state_hash::GlobalStateHash;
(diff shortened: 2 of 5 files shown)


mystenmark:
@andll The test failure here is actually a little concerning. If we are getting fewer cancellations it could be a sign that we now accept invalid transactions. Although, I would expect that to cause much bigger problems. Still we should track down the failure

PR #26431, opened 2026-04-29
Cache signed_effects_digests in memory to avoid disk reads

mystenmark:
## Summary
- Keep all uncommitted signed effects digests in a `DashMap`, populated from disk at epoch startup and pruned in `handle_finalized_checkpoint`.
- `get_signed_effects_digest` (called on every transaction) now reads only from the in-memory map instead of hitting RocksDB. Writes still persist to disk for crash recovery.
- Resolves the existing `TODO: read from cache instead of DB` comment.
- In test configurations (debug builds, msim, antithesis), every cache read is verified against the DB to catch any cache/DB divergence.

## Test plan
- [x] `cargo check -p sui-core`
- [x] `cargo xclippy` clean
- [x] Cache consistency assertion in test configurations (with retry for non-atomic write race)
- [ ] Verify with profiling that `signed_effects_digests` RocksDB reads are eliminated from the hot path
- [ ] Verify crash recovery: restart node and confirm signed effects digests are loaded from disk

### DIFF
--- a/crates/sui-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/sui-core/src/authority/authority_per_epoch_store.rs
@@ -2099,10 +2099,17 @@ impl AuthorityPerEpochStore {
         let cached = self.signed_effects_digests_cache.get(tx_digest).map(|r| *r);
         if in_test_configuration() {
             let from_db = self.tables()?.signed_effects_digests.get(tx_digest)?;
-            assert_eq!(
-                cached, from_db,
-                "signed_effects_digests cache inconsistency for {tx_digest}"
-            );
+            if cached != from_db {
+                // Cache and DB writes are not atomic, so retry after a brief delay
+                // to allow eventual consistency before panicking.
+                std::thread::sleep(std::time::Duration::from_secs(1));
+                let from_db = self.tables()?.signed_effects_digests.get(tx_digest)?;
+                let cached = self.signed_effects_digests_cache.get(tx_digest).map(|r| *r);
+                assert_eq!(
+                    cached, from_db,
+                    "signed_effects_digests cache inconsistency for {tx_digest}"
+                );
+            }
         }
         Ok(cached)
     }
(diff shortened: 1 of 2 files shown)


crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1083,6 +1089,12 @@ impl AuthorityPerEpochStore {
 
         let epoch_alive_token = CancellationToken::new();
 
+        let signed_effects_digests_cache = DashMap::new();
```
aschran, line 1090: nit-should we try to keep this roughly in similar order as the struct's fields are defined? or are we beyond the point where that even matters/is feasible anymore? haha

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2040,6 +2053,8 @@ impl AuthorityPerEpochStore {
             [(tx_digest, effects_digest)],
         )?;
         batch.write()?;
+        self.signed_effects_digests_cache
```
aschran, line 2056: i mildly worry about future regressions where we write to the table or the cache but not both, and wonder if it's worth adding a lightweight wrapper for the cache that also updates the backing store as needed, to make this type of bug impossible?

fine with leaving this as-is, just wanted to float the idea

aschran (approved):

crates/sui-core/src/authority/authority_per_epoch_store.rs
mystenmark: isn't that what this function is? i'm a bit confused what more I would do here

aschran: right, I don't think there are any bugs rn! my concern was just hypothetical: these functions are on a bloated struct that has a million fields and does a lot of different things, so it'd be easy to miss during a refactoring that this particular table and this particular dashmap must be kept in sync (especially since the table is not even in the same struct as the dashmap)

vs if we added some lightweight wrapper struct like `struct SignedEffectsDigestsCache` that had as fields the DashMap and the corresponding backing table (and did the population of the DashMap in its constructor fn, which is currently added to AuthorityPerEpochStore::new), and the only way you could get at the contents was through its accessor fns that kept the two in sync, it'd be much harder to miss the correspondence or break anything during a refactor

not a strong feeling though, maybe it'd be overkill!

mystenmark: that's reasonable - as an alternative, i added a check that the db and in-mem store always agree during tests. Do you think that's enough?

aschran: sure I think so!

PR #26437, opened 2026-04-30
[Graphql-alt] Graphql client sdk metric

tpham-mysten:
## Description 

- Add `client_sdk_type` and `client_sdk_version` labels to the `graphql_queries_received` Prometheus counter, sourced from the `client-sdk-type` and `client-sdk-version` HTTP request headers.
- Surface the same identifiers as structured fields on every request log line (succeeded, failed, cancelled).
- Empty string fallback when headers are absent, so non-SDK callers still aggregate cleanly.

Metrics that get the new labels:
  - `graphql_queries_received`

## Test plan 

- Unit tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/extensions/logging.rs
+++ b/crates/sui-indexer-alt-graphql/src/extensions/logging.rs
@@ -47,21 +47,28 @@ use crate::metrics::RpcMetrics;
 /// This custom response header contains a unique request-id used for debugging and appears in the logs.
 pub const REQUEST_ID_HEADER: HeaderName = HeaderName::from_static("x-sui-rpc-request-id");
 
-/// Headers identifying the SDK that issued the request. To capture an additional client header
-/// in the future, define another constant here, read it in `ClientInfo::from_headers`, and expose
-/// it via the appropriate metric label or log field.
+/// Header identifying the SDK that issued the request. The value flows into the
+/// `client_sdk_type` Prometheus label, matched against `SDK_TYPE_WHITELIST`. Add a new entry to
+/// the whitelist when a new first-party SDK adopts this header, otherwise its requests will be
+/// bucketed into `CLIENT_LABEL_OTHER`.
 const CLIENT_SDK_TYPE_HEADER: HeaderName = HeaderName::from_static("client-sdk-type");
+
+/// Header identifying the SDK version that issued the request. The value flows into the
+/// `client_sdk_version` Prometheus label, sanitized by `sanitize_sdk_version`. Values longer
+/// than `MAX_SDK_VERSION_LEN` or containing characters outside `[A-Za-z0-9._-]` are bucketed
+/// into `CLIENT_LABEL_OTHER`.
 const CLIENT_SDK_VERSION_HEADER: HeaderName = HeaderName::from_static("client-sdk-version");
 
-/// Maximum length of a client SDK header value we accept as a Prometheus label. Anything longer
-/// is bucketed to `CLIENT_LABEL_INVALID` so a misbehaving client cannot grow the per-series memory
-/// cost without bound.
-const MAX_CLIENT_SDK_LABEL_LEN: usize = 64;
+/// SDKs we accept verbatim as the `client_sdk_type` Prometheus label.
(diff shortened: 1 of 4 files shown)


crates/sui-indexer-alt-graphql/src/extensions/logging.rs
```diff
@@ -81,6 +97,27 @@ impl Session {
...
+        Self {
+            sdk_type: headers
+                .get(&CLIENT_SDK_TYPE_HEADER)
+                .and_then(|v| v.to_str().ok().map(String::from)),
+            sdk_version: headers
+                .get(&CLIENT_SDK_VERSION_HEADER)
+                .and_then(|v| v.to_str().ok().map(String::from)),
         }
     }
 }
```
nickvikeras, line 142: Do we need to sanitize these before using them as metric labels? I.e some kind of white list of allowed SDK types, and a regex for the expected version format?

I don't have much experience with prometheus/grafana but in other metrics systems, I have seen problems where unsanitized metric labels cause huge load or outages on the underlying time series database where the metrics get stored. 

My worry is since these are being set directly from the http header values, a malicious attacker could use it to attack our monitoring infra in the worst case.

tpham-mysten: Good catch, and it's a fair concern in principle but I think the realized risk is bounded enough as we have some existing controls:
  1. LB rate-limits each source IP to 200 RPS (sui-operations/pulumi/gcp/sui-rpc2/securityPolicy.go).
  2. Total request headers capped at 15 KiB by the GCP LB; HTTP/2 also bounded by hyper's 16 KiB max_header_list_size.

However, adding a check is also not complex, so I have added a simple length check. We can add more later if needed

crates/sui-indexer-alt-graphql/src/extensions/logging.rs
```diff
@@ -53,6 +53,16 @@ pub const REQUEST_ID_HEADER: HeaderName = HeaderName::from_static("x-sui-rpc-req
...
+/// Maximum length of a client SDK header value we accept as a Prometheus label. Anything longer
+/// is bucketed to `CLIENT_LABEL_INVALID` so a misbehaving client cannot grow the per-series memory
+/// cost without bound.
+const MAX_CLIENT_SDK_LABEL_LEN: usize = 64;
+
+/// Sentinel label value substituted when a client SDK header exceeds `MAX_CLIENT_SDK_LABEL_LEN`.
+/// Distinct from the empty string we use when the header is absent, so dashboards can tell the
+/// two apart.
+const CLIENT_LABEL_INVALID: &str = "INVALID";
+
```
nickvikeras: Just reviving your earlier response on the newest commit so it's visible -

>Good catch, and it's a fair concern in principle but I think the realized risk is bounded enough as we have some existing controls:
>
>LB rate-limits each source IP to 200 RPS (sui-operations/pulumi/gcp/sui-rpc2/securityPolicy.go).
Total request headers capped at 15 KiB by the GCP LB; HTTP/2 also bounded by hyper's 16 KiB max_header_list_size.
However, adding a check is also not complex, so I have added a simple length check. We can add more later if needed

I was more worried about sanitizing to protect _cardinality_, not length/ rate. A 64 length string can still have billions of possible values.

From the [prometheus docs](https://prometheus.io/docs/practices/naming/).

>CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

So to protect cardinality you'd need an allowlist/regex or something, not just a length guard.

tpham-mysten: Sounds good. I have added the sanity check:
- For `client-sdk-type`: Use `SDK_TYPE_WHITELIST`
- For `client-sdk-version`: Use `[A-Za-z0-9._-]` + Max length of 32

nickvikeras (approved):

PR #26443, opened 2026-04-30
rpc-index: gracefully handle db corruption

bmwill:
Gracefully handle db corruption, which can happen unexpectedly or explicitly upon updating the db version.

When updating the DB version, the merge operator for the balance column family tries to read the older values before we have a chance to nuke the db and reinitialize it. This makes that reading of the older value fail gracefully so that we can have a chance to properly reinitialize the db.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/rpc_index.rs
+++ b/crates/sui-core/src/rpc_index.rs
@@ -247,18 +247,29 @@ fn balance_delta_merge_operator(
     existing_val: Option<&[u8]>,
     operands: &MergeOperands,
 ) -> Option<Vec<u8>> {
-    let mut result = existing_val
-        .map(|v| {
-            bcs::from_bytes::<BalanceIndexInfo>(v)
-                .expect("Failed to deserialize BalanceIndexInfo from RocksDB - data corruption.")
-        })
-        .unwrap_or_default();
+    let mut result = if let Some(existing_val) = existing_val {
+        bcs::from_bytes::<BalanceIndexInfo>(existing_val)
+            .inspect_err(|e| {
+                tracing::error!(
+                    "Failed to deserialize BalanceIndexInfo from RocksDB - data corruption: {e}"
+                )
+            })
+            .ok()?
+    } else {
+        BalanceIndexInfo::default()
+    };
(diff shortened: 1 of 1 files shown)


crates/sui-core/src/rpc_index.rs
```diff
@@ -247,18 +247,29 @@ fn balance_delta_merge_operator(
...
-            bcs::from_bytes::<BalanceIndexInfo>(v)
-                .expect("Failed to deserialize BalanceIndexInfo from RocksDB - data corruption.")
-        })
-        .unwrap_or_default();
+    let mut result = if let Some(existing_val) = existing_val {
+        bcs::from_bytes::<BalanceIndexInfo>(existing_val)
+            .inspect_err(|e| {
+                tracing::error!(
+                    "Failed to deserialize BalanceIndexInfo from RocksDB - data corruption: {e}"
+                )
```
alex-mysten, line 255: we would expect to see this once in testnet and never again, right? should it be an invariant violation for visibility?

is there a reasonable test we can add?

bmwill: > we would expect to see this once in testnet and never again, right? should it be an invariant violation for visibility?

theres lots of reasons why dbs could be corrupted so its not that we expect this to happen. The reason why its seen today is a race between db teardown and merge operators running. Another option would be to just name the new CF a new name but I like this approach because its more resilient, db corruption would cause reads to some keys to fail instead of crashing the node. Also i don't think invariant violation makes sense here since a safety invariant isn't being violated.

>is there a reasonable test we can add?

unfortunately the DBMap construct would make this rather hard to explicitly test for, I have an item on my backlog to completely rework this DB to not use it, which would make testing much easier but thats a larger effort than i'd like to start today.

alex-mysten (approved):

PR #26447, opened 2026-04-30
sui-benchmark: retry soft bundles on retriable wait_for_effects responses

aschran:
## Summary

`execute_soft_bundle_with_retries` already retries on submit-time retriable rejections; extend the same retry to cover retriable `wait_for_effects` responses (`Rejected` with a retriable reason, `Expired`), which similarly mean the tx was never ordered. Dedupe the four retry sites onto a small helper.

`Rejected{error: None}` still surfaces as `UnknownRejection`.

## Behavior change

Workloads counting `RetriableFailure` events will see fewer; most now resolve to `Success` or `PermanentFailure` via proxy-level retry.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/src/lib.rs
+++ b/crates/sui-benchmark/src/lib.rs
@@ -600,6 +600,23 @@ impl ValidatorProxy for LocalValidatorAggregatorProxy {
     }
 }
 
+async fn warn_and_backoff_for_retry(
+    digests: &[TransactionDigest],
+    retry_cnt: &mut u32,
+    reason: impl std::fmt::Display,
+) {
+    let delay = Duration::from_millis(rand::thread_rng().gen_range(100..1000));
+    warn!(
+        ?digests,
+        retry_cnt = *retry_cnt,
+        "Soft bundle retry: {}. Sleeping for {:?} ...",
+        reason,
+        delay,
+    );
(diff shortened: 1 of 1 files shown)


crates/sui-benchmark/src/lib.rs
```diff
@@ -784,6 +789,32 @@ async fn execute_soft_bundle_with_retries(
 
         let wait_responses = futures::future::join_all(wait_futures).await;
 
+        // Re-submit if any wait_for_effects returned a retriable signal (Rejected
+        // with a retriable reason, or Expired) — the tx was never ordered.
+        // TODO: when error is None, poll other validators for a reject reason
+        // before giving up — our chosen validator didn't vote reject so won't
+        // have one cached, but a different validator may.
```
alex-mysten, line 796: Why leave a TODO? Seems trivial enough to add

aschran: I was just trying to keep the diff small and focused on fixing the observed failure. But can add it now if you like

alex-mysten (approved):

PR #26453, opened 2026-05-01
[graphql-stream] Add top-level events subscription [8/n]

tpham-mysten:
## Description 

Add `subscription { events(filter: { sender, module, type }) { ... } }` that yields each matching event individually as it appears in finalized checkpoints. Each yield carries the streamed checkpoint scope plus tx execution objects, so Move type and value resolution flows through the streaming package store

*Features that are not supported in this PR (and will be added in follow-up PR):*
- Resume from a historical event cursor
- Plumbing transaction contents into `Event` so chained resolution like `event.transaction.effects.objectChanges` works for streamed events
- Metrics

## Test plan 

- Unit + e2e tests
- Start a local GraphQL server

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/src/move_helpers.rs
+++ b/crates/sui-indexer-alt-e2e-tests/src/move_helpers.rs
@@ -1,0 +1,70 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Generic helpers for publishing Move packages and executing programmable transactions
+//! against a `TestCluster`. Used by integration test harnesses to avoid duplicating gas,
+//! signing, and execute boilerplate.
+
+use std::collections::BTreeSet;
+use std::path::Path;
+
+use fastcrypto::encoding::Base58;
+use fastcrypto::encoding::Encoding;
+use sui_test_transaction_builder::TestTransactionBuilder;
+use sui_types::base_types::ObjectID;
+use sui_types::base_types::ObjectRef;
+use sui_types::effects::TransactionEffects;
+use sui_types::effects::TransactionEffectsAPI;
+use sui_types::programmable_transaction_builder::ProgrammableTransactionBuilder;
+use sui_types::transaction::TransactionData;
+
+const DEFAULT_GAS_BUDGET: u64 = 5_000_000_000;
+
+/// Publish a Move package at `path` and return the new package ID.
+pub async fn publish_package(cluster: &mut test_cluster::TestCluster, path: &Path) -> ObjectID {
+    let sender = cluster.wallet.active_address().unwrap();
+    let gas = gas_for(cluster).await;
+    let rgp = cluster.wallet.get_reference_gas_price().await.unwrap();
+    let tx = cluster
+        .wallet
+        .sign_transaction(
+            &TestTransactionBuilder::new(sender, gas, rgp)
+                .publish_async(path.to_path_buf())
+                .await
+                .build(),
+        )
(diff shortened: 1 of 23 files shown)


crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -448,7 +448,7 @@ fn process_transaction(
         .deserialize()
         .context("Failed to deserialize transaction effects")?;
 
-    let events = proto
```
tpham-mysten, line 451: Simplifying to Vector to make the logic clean.

crates/sui-indexer-alt-e2e-tests/src/move_helpers.rs
tpham-mysten, line 1: Defining common helper utility to make emit_event_harness and object_wrapping_harness cleaner

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -109,6 +111,60 @@ impl Subscription {
...
+            loop {
+                match receiver.recv().await {
+                    Ok(processed) => {
+                        let timestamp_ms = Some(processed.summary.timestamp_ms);
+                        let scope = Scope::for_streamed_checkpoint(
+                            package_store.clone(),
+                            resolver_limits.clone(),
+                        );
+                        for tx in &processed.transactions {
+                            let events = tx.contents.events().unwrap_or_default();
```
evan-wall-mysten, line 143: Not sure how much this matters, but calling `events()` will clone everything internally regardless of whether it gets filtered. Is it possible to only clone items that pass the filter?

evan-wall-mysten (approved):

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Good catch. That is definitely not efficient, but yeah not sure how much it matters.

But the fix is quite simple tho, and I have updated kv_loader to return Vec<Arc<Event>>: https://github.com/MystenLabs/sui/pull/26453/changes#diff-cf2455a06b1601ecbb7807dedcb7131874d5bf97944f2437fa6d041b14312a3cL103

tpham-mysten:
Thanks @evan-wall-mysten for the catch. I have updated kv_loader to use Vec<Arc<Event>> to avoid the clone. 

I also updated `Event` to use `Arc<NativeEvent>` to make it even more efficient

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -109,6 +111,60 @@ impl Subscription {
             }
         })
     }
+
+    /// Subscribe to events as they are emitted, with optional filtering.
+    ///
+    /// Each matching event is yielded individually as it appears in finalized
```
amnn, line 117: Remember not to hard-wrap doc comments that end up being used to generate docs in the schema (also applies to `transaction`, above).

amnn (commented):
Looks good! Small post-land comment (not really related to the substance of the PR)

Issue #25750, opened 2026-03-07
Compiler panic in `to_bytecode/translate.rs`: divergent expression inside vector literal


jubnzv:
The compiler panics with an ICE in [`to_bytecode/translate.rs:972`](https://github.com/MystenLabs/sui/blob/a52ef37a51/external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs#L972) when a vector literal contains a divergent expression (e.g. `abort`) and the result is bound to a variable.

The issue is that the typing pass accepts `vector[{ abort 0 }]` (abort is valid in any expression position and the vector element type becomes uninferred), but the bytecode translation assumes all types are concrete at that point.

MRE:
```move
module 0x2a::M {
    fun f(): u64 {
        let v = vector[{ abort 0 }];
        0
    }
}
```

Reproduce steps:
```
cd external-crates/move
cargo +nightly run --release -p move-compiler --bin move-check -- mre.move
```

Output:
```
thread 'main' panicked at crates/move-compiler/src/to_bytecode/translate.rs:972:13:
ICE should not have reached compilation if there are errors
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Expected behavior: a warning about unreachable code or dead variable, not an ICE.

Git commit: a52ef37a51
Build: release
cargo 1.91.0-nightly (840b83a10 2025-07-30)
rustc 1.91.0-nightly (3672a55b7 2025-08-13)

PR #26454, opened 2026-05-01
[move][move-compiler] Fix `vector<void>` ICE

cgswords:
## Description 

Title, see #25750

## Test plan 

New tests added.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-compiler/src/typing/core.rs
+++ b/external-crates/move/crates/move-compiler/src/typing/core.rs
@@ -2388,8 +2388,24 @@ pub fn check_call_arity<S: std::fmt::Display, F: Fn() -> S>(
 pub fn solve_constraints(context: &mut Context) {
     use BuiltinTypeName_ as BT;
 
-    // Solve these constraints first to minimize error reporting.
+    // Resolve divergent type variables first, so that downstream constraints (e.g., base type)
+    // can see Void rather than an unresolved TVar.
+    {
+        let var_constraints = context.subst.tvar_constraints.clone();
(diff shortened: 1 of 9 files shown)


external-crates/move/crates/move-compiler/src/typing/core.rs
```diff
@@ -2648,12 +2660,15 @@ fn solve_base_type_constraint(context: &mut Context, loc: Loc, msg: String, ty:
...
-        | TI::Apply(_, _, _)
-        | TI::Fun(_, _) => (),
+        TI::Void => {
+            let tmsg = "Could not find a concrete type for this expression";
+            context.add_diag(diag!(
+                TypeSafety::ExpectedBaseType,
+                (loc, msg),
+                (tyloc, tmsg)
+            ))
+        }
```
tnowacki, line 2670: Should this be in expansion or does it have to be here for constraints?

tnowacki (approved):

external-crates/move/crates/move-compiler/src/typing/core.rs
cgswords: It has to be here for constraints; this is the key to the fix. We lean on the base constraint to find / prevent this situation. Moving it to elaboration would require either the "base_type: bool" flag I mentioned or will break some existing tests to have worse errors.

PR #26460, opened 2026-05-01
[move-core-types] Convert visitors to use compressed type layouts

tzakian:
## Description 

What it says on the cover.

Note that the visitors no longer hold a reference to the type layout -- this is because with how the views are now structured you do not grab inner references as you traverse (so the lifetimes no longer would work out in particular). 

It instead switches to using owned type layouts and cloning wherever needed -- this is fine however since clones are/should be exceptionally cheap (two bumps of an Arc in most cases). 

## Test plan 

Existing tests. 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-core-types/src/annotated_extractor.rs
+++ b/external-crates/move/crates/move-core-types/src/annotated_extractor.rs
@@ -67,7 +67,7 @@ where
 
     pub fn deserialize_value(
         bytes: &'b [u8],
-        layout: &'l A::MoveTypeLayout,
+        layout: crate::compressed::annotated::MoveTypeLayoutRef<'l>,
         inner: &'v mut V,
         path: Vec<Element<'p>>,
     ) -> Result<Option<V::Value>, V::Error> {
@@ -77,7 +77,7 @@ where
 
     pub fn deserialize_struct(
         bytes: &'b [u8],
-        layout: &'l A::MoveStructLayout,
+        layout: crate::compressed::annotated::MoveStructLayout<'l>,
         inner: &'v mut V,
         path: Vec<Element<'p>>,
     ) -> Result<Option<V::Value>, V::Error> {
@@ -201,7 +201,7 @@ impl<'b, 'l, V: AV::Visitor<'b, 'l>> AV::Visitor<'b, 'l> for Extractor<'_, '_, V
         // If there is a type element, check that it is a vector type with the correct element
         // type, and remove it from the path.
         let path = if let [E::Type(t), path @ ..] = self.path {
-            if !matches!(t, T::Vector(t) if driver.element_layout().is_type(t)) {
+            if !matches!(t, T::Vector(inner) if driver.element_layout().as_view().is_type(inner)) {
                 return Ok(None);
             }
             path
@@ -235,12 +235,11 @@ impl<'b, 'l, V: AV::Visitor<'b, 'l>> AV::Visitor<'b, 'l> for Extractor<'_, '_, V
         driver: &mut AV::StructDriver<'_, 'b, 'l>,
     ) -> Result<Self::Value, Self::Error> {
         use Element as E;
-        use TypeTag as T;
 
         // If there is a type element, check that it is a struct type with the correct struct tag,
         // and remove it from the path.
         let path = if let [E::Type(t), path @ ..] = self.path {
-            if !matches!(t, T::Struct(t) if driver.struct_layout().is_type(t)) {
+            if !driver.struct_layout().is_type(t) {
                 return Ok(None);
             }
             path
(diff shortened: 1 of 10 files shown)


external-crates/move/crates/move-core-types/src/annotated_visitor.rs
```diff
@@ -480,10 +486,10 @@ pub struct StructDriver<'c, 'b, 'l> {
 /// to progress the traversal (by visiting or skipping fields).
 pub struct VariantDriver<'c, 'b, 'l> {
     inner: ValueDriver<'c, 'b, 'l>,
-    layout: &'l MoveEnumLayout,
+    layout: MoveEnumLayout<'l>,
     tag: u16,
-    variant_name: &'l IdentStr,
-    variant_layout: &'l [MoveFieldLayout],
+    variant_name: &'l Identifier,
```
amnn, line 491: Shouldn't this still be `&'l IdentStr`?

external-crates/move/crates/move-core-types/src/annotated_visitor.rs
```diff
@@ -452,7 +458,7 @@ impl<'b, 'l, T: Traversal<'b, 'l> + ?Sized> Visitor<'b, 'l> for T {
 /// value being visited.
 pub struct ValueDriver<'c, 'b, 'l> {
     bytes: &'c mut Cursor<&'b [u8]>,
-    layout: Option<&'l MoveTypeLayout>,
+    layout: Option<MoveTypeLayoutRef<'l>>,
```
amnn, line 461: Naming wise, it's interesting that all the sub-layout types (`MoveStructLayout`, `MoveEnumLayout`, `MoveFieldsLayout`) don't have `Ref` in their name, but `MoveTypeLayoutRef` does.

amnn: And by interesting I mean confusing and unexpected 😅

external-crates/move/crates/move-core-types/src/annotated_visitor.rs
```diff
@@ -695,35 +699,35 @@ impl<'c, 'b, 'l> StructDriver<'c, 'b, 'l> {
...
-    /// `f` is the layout of the field that was visited) or an error if there was an underlying
-    /// deserialization error, or an error during visitation.
+    /// Returns `Ok(None)` if there are no more fields in the struct, `Ok(((n, l), v))` if there was
+    /// a field and it was successfully visited.
     pub fn next_field<V: Visitor<'b, 'l> + ?Sized>(
         &mut self,
         visitor: &mut V,
-    ) -> Result<Option<(&'l MoveFieldLayout, V::Value)>, V::Error> {
-        let Some(field) = self.peek_field() else {
+    ) -> Result<Option<((&'l Identifier, MoveTypeLayoutRef<'l>), V::Value)>, V::Error> {
```
amnn, line 716: `&'l IdentStr` would be preferable here too (I think), it's strictly less leaky than `&'l Identifier`.

external-crates/move/crates/move-core-types/src/annotated_extractor.rs
```diff
@@ -312,7 +310,7 @@ impl<'b, 'l, V: AV::Visitor<'b, 'l>> AV::Visitor<'b, 'l> for Extractor<'_, '_, V
         match field {
             // Skip over mismatched fields by name.
             E::Field(f) => {
-                while matches!(driver.peek_field(), Some(l) if l.name.as_str() != *f) {
+                while matches!(driver.peek_field(), Some(l) if l.0.as_str() != *f) {
```
amnn, line 313: nit (optional): slightly preferable.

```suggestion
                while matches!(driver.peek_field(), Some((name, _)) if name.as_str() != *f) {
```

amnn: (same above)

external-crates/move/crates/move-core-types/src/unit_tests/runtime_visitor_test.rs
```diff
@@ -16,6 +17,36 @@ use crate::{
...
+
+fn tree_visit_value<'b, V: Visitor<'b, 'b>>(
+    bytes: &'b [u8],
+    ty: &MoveTypeLayout,
+    visitor: &mut V,
+) -> Result<V::Value, V::Error>
+where
+    V::Error: std::error::Error + Send + Sync + 'static,
+{
+    let compressed: &'static CR::MoveTypeLayout = Box::leak(Box::new(compress(ty)));
```
amnn, line 32: What's the story here with the leaking?

amnn (commented):
Actual change looks good, just curious about the `Box::leak` calls in tests!

PR #26461, opened 2026-05-01
compressed type layouts conversion on Sui side

tzakian:
## Description 

Conversion on the Sui side of things. Everything outside of wire formats/data structures that derive Ser/De are converted to using compressed layouts. 

Indexer/RPC/Display code is the most heavily touched, and there are some minor touches outside of it but it's generally all pretty mechanical.

There are still some places where I think we can improve things (e.g., around absorbing sub layouts in the RPCReader trait) but I will want to put that up separately as that is less of a mechanical change. 

## Test plan 

Existing tests should pass. 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-analytics-indexer/src/handlers/tables/df.rs
```diff
@@ -64,10 +64,10 @@ impl DynamicFieldProcessor {
             .await?;
         let object_id = object.id();
 
-        let field = DFV::FieldVisitor::deserialize(move_object.contents(), &layout)?;
+        let field = DFV::FieldVisitor::deserialize(move_object.contents(), layout.as_ref())?;
```
amnn, line 67: In a comment on another PR, I asked whether it was necessary to call `as_ref` explicitly, and vaguely remembered that Rust would do that automatically, but I was actually mis-remembering auto-deref:

If we were able to implement `impl Deref<Target = MoveTypeLayoutRef> for MoveTypeLayout`, then we could keep the `&layout` and Rust would call `.deref()` for us to get a `&MoveTypeLayoutRef`. This is how `PathBuf` -> `&Path`, `String` -> `&str`, `OsString` -> `&OsStr` works, but in all these cases, the target is legitimately the variable-length internal representation of the dereferenced type, and I don't know how to do the same thing in our case where we really want to create a custom fat pointer, which refers to an array, but also points to an offset inside it.

It kind of seems like Rust's `Deref` is over-specified in requiring that its return type is a real reference ...sad times.

crates/sui-core/src/unit_tests/subscription_handler_tests.rs
```diff
@@ -33,8 +34,12 @@ fn test_to_json_value() {
         ],
     };
     let event_bytes = bcs::to_bytes(&move_event).unwrap();
+    let layout = CA::MoveTypeLayout::try_from(
+        &move_core_types::annotated_value::MoveTypeLayout::Struct(Box::new(TestEvent::layout())),
```
amnn, line 38:
```suggestion
        &A::MoveTypeLayout::Struct(Box::new(TestEvent::layout())),
```

and update imports (probably replace all of the `MoveFieldLayout`, `MoveStructLayout`, `MoveTypeLayout` with `A::..`?)

crates/sui-core/src/authority.rs
```diff
@@ -3410,7 +3406,7 @@ impl AuthorityState {
             (request.generate_layout, object.data.try_as_move())
         {
             let epoch_store = self.load_epoch_store_one_call_per_task();
-            Some(into_struct_layout(
+            Some(into_tree_struct_layout(
```
amnn, line 3409: Should we be pushing compressed layouts even further into the interface here?

crates/sui-display/src/v2/visitor/extractor.rs
```diff
@@ -178,13 +179,13 @@ impl<'v> AV::Visitor<'v, 'v> for Extractor<'v, '_> {
         // If the next accessor is an index, then try and treat this struct as a VecMap, and
         // perform a VecMap look-up.
         if let Accessor::Index(i) = accessor {
-            if !is_vec_map(&driver.struct_layout().type_) {
+            if !is_vec_map(struct_layout.type_()) {
```
amnn, line 182:
```suggestion
            if !is_vec_map(ty) {
```

crates/sui-display/src/v2/visitor/extractor.rs
amnn, line 1: The lifetime change is interesting -- what made it so that we needed the extra degree of freedom?

crates/sui-display/src/v2/mod.rs
amnn, line 1: Does it make sense to build compressed layouts directly? Maybe there's a broader question hiding here which is whether you expect the tree-based layout to persist in the long-term for some use cases?

crates/sui-display/src/v2/value.rs
```diff
@@ -10,7 +10,7 @@ use base64::engine::Engine;
 use chrono::DateTime;
 use move_core_types::account_address::AccountAddress;
 use move_core_types::annotated_value as A;
-use move_core_types::annotated_value::MoveTypeLayout;
+use move_core_types::compressed::annotated::{self as CA, BackendBuilder as _};
```
amnn, line 13: nit: no {...} imports.

crates/sui-display/src/v2/value.rs
```diff
@@ -856,7 +865,9 @@ impl<'s> TryFrom<Value<'s>> for Atom<'s> {
                     .into_iter()
                     .map(|e| match e {
                         V::U8(b) => Ok(b),
-                        V::Slice(Slice { layout, bytes, .. }) if layout == &L::U8 => {
+                        V::Slice(Slice { layout, bytes, .. })
+                            if layout.as_view() == CA::MoveTypeLayout::u8().as_view() =>
```
amnn, line 869: In other places, I've seen you convert to a `MoveTypeLayoutRef` to do the equivalence check, is there a canonical pattern to follow here?

crates/sui-display/src/v2/value.rs
```diff
@@ -946,9 +966,9 @@ pub(crate) mod tests {
             mut self,
             parent: AccountAddress,
             name: N,
-            name_layout: MoveTypeLayout,
+            name_layout: L,
```
amnn, line 969: nit: not sure how I feel about the use of a single letter type name in this position -- it's maybe a mistake on my part to import it like that at the module level, but now that we have multiple things floating around that represent layouts (tree-based layouts, compressed-layouts, compressed-layout refs, and compressed-layout refs) and arguably this is the least important one, it's shone a spotlight on that mistake.

crates/sui-indexer-alt-graphql/src/api/types/validator_set.rs
```diff
@@ -108,25 +108,25 @@ impl ValidatorSetContents {
...
+        layout: &CA::MoveTypeLayout,
     ) -> Result<Self, RpcError> {
         type ReportRecords = BTreeMap<NativeSuiAddress, Vec<NativeSuiAddress>>;
         type AtRiskValidators = BTreeMap<NativeSuiAddress, u64>;
 
         #[derive(Default)]
-        struct Contents<'b, 'l> {
-            native: Option<(&'b [u8], &'l A::MoveTypeLayout)>,
+        struct Contents<'b> {
+            native: Option<(&'b [u8], CA::MoveTypeLayout)>,
```
amnn, line 118: How come this did not become a:


```suggestion
            native: Option<(&'b [u8], CA::MoveTypeLayoutRef<'l>)>,
```

crates/sui-json-rpc-types/src/unit_tests/rpc_types_tests.rs
```diff
@@ -25,9 +26,13 @@ fn test_move_value_to_sui_coin() {
     let coin = GasCoin::new(id, value);
 
     let move_object = MoveObject::new_gas_coin(SequenceNumber::new(), id, value);
-    let layout = GasCoin::layout();
-
-    let move_struct = move_object.to_move_struct(&layout).unwrap();
+    let layout = CA::MoveTypeLayout::try_from(
+        &move_core_types::annotated_value::MoveTypeLayout::Struct(Box::new(GasCoin::layout())),
```
amnn, line 30: nit: import `annotated_value` as `A`

crates/sui-package-resolver/src/lib.rs
```diff
@@ -29,10 +29,7 @@ use move_command_line_common::display::RenderResult;
 use move_command_line_common::display::try_render_constant;
 use move_command_line_common::error_bitset::ErrorBitset;
 use move_core_types::account_address::AccountAddress;
-use move_core_types::annotated_value::MoveEnumLayout;
-use move_core_types::annotated_value::MoveFieldLayout;
-use move_core_types::annotated_value::MoveStructLayout;
-use move_core_types::annotated_value::MoveTypeLayout;
+use move_core_types::compressed::annotated::{self as CA, BackendBuilder as _, LayoutHandle};
```
amnn, line 32: nit: no {...} imports in this crate

crates/sui-package-resolver/src/lib.rs
```diff
@@ -1551,60 +1562,81 @@ impl<'l> ResolutionContext<'l> {
     /// returned alongside the layout (assuming it does not exceed `max_depth`).
     fn resolve_datatype_signature(
         &self,
+        builder: &mut CA::MoveTypeLayoutBuilder,
         data_def: &DataDef,
         type_: StructTag,
-        param_layouts: Vec<(MoveTypeLayout, usize)>,
+        param_layouts: Vec<(LayoutHandle, usize)>,
```
amnn, line 1568: Is `LayoutHandle` aliased as something else in other places?

crates/sui-package-resolver/src/lib.rs
```diff
@@ -1551,60 +1562,81 @@ impl<'l> ResolutionContext<'l> {
...
-                        name: ident(name.as_str())?,
-                        layout,
-                    })
+                    resolved_fields.push((ident(name.as_str())?, layout))
                 }
 
+                let resolved_field_refs: Vec<_> = resolved_fields
+                    .iter()
+                    .map(|(name, layout)| (name, *layout))
+                    .collect();
```
amnn, line 1587: Could you explain what is happening here? I.e. why do we need to materialize a temporary vector and then iterate over it, dereferencing layouts?

amnn (commented):
Broadly this change looks good, but one thing I'm noticing is that the view type is showing up a bit more often than I expected, now that we have a ref type. I would expect that now, the View type would only be used for pattern matching, while equality and `TypeTag` construction would be accessed via the ref type.

I'm also curious to learn more about the process of constructing a struct/enum layout, as there seems to be something funky going on there now (with materializing temporary vectors), and would like to get your thoughts on what the long-term plan is for whether the current `MoveTypeLayout` will continue to exist or not.

PR #26462, opened 2026-05-01
[move-ir] Require public on structs/enums

tnowacki:
## Description 

- 🤖 
- We are breaking everything so why not this too
- Requiring public on datatype declarations was a bit more involved than other changes due to changing the grammar a bit (`public` no longer implies a function)

## Test plan 

- More tests
- Added some missing tests from previous PRs too 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


tzakian (approved):
Looks good to me

PR #26476, opened 2026-05-04
[graphql-stream] Lift streamed checkpoint payload onto Scope [9/n]

tpham-mysten:
## Description 

- Lift the streamed `ProcessedCheckpoint` onto `Scope` once per checkpoint via `Scope::for_streamed_checkpoint(..., processed)`, so yielded subscription types stop carrying per-yield transaction context.
- Replace `Scope::with_execution_objects(per_tx_map)` with `Scope::with_streamed_tx(tx_sequence_number)`. The new builder anchors a nested scope to a specific transaction within the streamed checkpoint, and object lookups (`execution_output_object`, `execution_output_object_latest`, the `PackageStore::fetch` impl) route through a small `execution_objects_in_view()` helper that returns the anchored transaction's per-tx map.
- Behavior is preserved end to end. A transaction sees exactly its own input/output objects, the same as before, so existing tests pass without snapshot churn.

## Context

This refactor lands the [discussion](https://github.com/MystenLabs/sui/pull/26170#pullrequestreview-4099866645) on PR #26170.

## Test plan 

- unit + e2e tests

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/scope.rs
+++ b/crates/sui-indexer-alt-graphql/src/scope.rs
@@ -44,8 +44,8 @@ pub(crate) enum ObjectSource {
     Executed {
         execution_objects: ExecutionObjectMap,
     },
-    /// A streamed checkpoint with all per-tx contents and execution objects. The transaction
-    /// the scope is currently resolving for is identified by [`Scope::tx_sequence_number_viewed_at`].
+    /// A streamed checkpoint with all per-tx contents and execution objects. The active
+    /// transaction within this scope is identified by [`Scope::tx_sequence_number_viewed_at`].
     Streamed {
         checkpoint: Arc<ProcessedCheckpoint>,
     },
--- a/crates/sui-indexer-alt-graphql/src/scope.rs
+++ b/crates/sui-indexer-alt-graphql/src/scope.rs
@@ -32,6 +32,25 @@ use crate::task::watermark::Watermarks;
 pub(crate) type ExecutionObjectMap =
     Arc<BTreeMap<(ObjectID, SequenceNumber), Option<NativeObject>>>;
 
+/// Where object lookups in this scope draw their data from. Encodes the mutually exclusive
+/// modes a [`Scope`] can be in. Indexed mode hits the database/kv_loader (no in-memory payload).
+/// Executed mode is the mutation/simulate path, with a freshly executed transaction's outputs.
+/// Streamed mode is the subscription path, with an in-memory checkpoint payload.
+#[derive(Clone)]
+pub(crate) enum ObjectSource {
+    /// Reads go through the indexed-checkpoint path (kv_loader / DB). No in-memory payload.
+    Indexed,
+    /// A freshly executed transaction's input/output objects.
+    Executed {
+        execution_objects: ExecutionObjectMap,
+    },
+    /// A streamed checkpoint with all per-tx contents and execution objects. The transaction
+    /// the scope is currently resolving for is identified by [`Scope::tx_sequence_number_viewed_at`].
+    Streamed {
+        checkpoint: Arc<ProcessedCheckpoint>,
+    },
+}
+
 /// Root object bound for consistent dynamic field reads.
 ///
 /// This enables consistent dynamic field reads in the case of chained dynamic object fields,
@@ -64,25 +83,18 @@ pub(crate) struct Scope {
     /// None indicates execution context where we're viewing fresh transaction effects not yet indexed.
     checkpoint_viewed_at: Option<u64>,
 
+    /// The transaction we are scoped to, identified by global `tx_sequence_number`. In streaming
+    /// mode, anchors object lookups to that transaction's per-tx `execution_objects` within the
+    /// streamed checkpoint.
+    tx_sequence_number_viewed_at: Option<u64>,
+
     /// Root object bound for dynamic fields.
     ///
     /// This can be expressed either in terms of a specific object version or a checkpoint.
     root_bound: Option<RootBound>,
 
-    /// Cache of objects available in execution context (freshly executed transaction).
-    /// Maps (ObjectID, SequenceNumber) to optional object data.
-    /// None indicates the object was deleted or wrapped at that version.
-    /// This enables any Object GraphQL type to access fresh data without database queries.
-    execution_objects: ExecutionObjectMap,
-
-    /// The streamed checkpoint payload this scope is resolving against, if any. Holds the
-    /// per-tx contents and execution objects so nested resolvers can navigate the checkpoint
-    /// without re-plumbing per-yield context.
-    streamed_checkpoint: Option<Arc<ProcessedCheckpoint>>,
-
-    /// Within `streamed_checkpoint`, the `tx_sequence_number` of the transaction we are
-    /// resolving for. Object lookups use this transaction's own `execution_objects`.
-    tx_sequence_number_viewed_at: Option<u64>,
+    /// Where object lookups in this scope draw their data from. See [`ObjectSource`].
+    object_source: ObjectSource,
 
     /// Access to packages for type resolution.
     package_store: Arc<dyn PackageStore>,
@@ -101,10 +113,9 @@ impl Scope {
 
         Ok(Self {
             checkpoint_viewed_at: Some(watermark.high_watermark().checkpoint()),
-            root_bound: None,
-            execution_objects: Arc::new(BTreeMap::new()),
-            streamed_checkpoint: None,
             tx_sequence_number_viewed_at: None,
+            root_bound: None,
+            object_source: ObjectSource::Indexed,
             package_store: package_store.clone(),
             resolver_limits: limits.package_resolver(),
         })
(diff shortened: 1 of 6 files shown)


crates/sui-indexer-alt-graphql/src/scope.rs
```diff
@@ -74,6 +75,15 @@ pub(crate) struct Scope {
     /// This enables any Object GraphQL type to access fresh data without database queries.
     execution_objects: ExecutionObjectMap,
 
+    /// The streamed checkpoint payload this scope is resolving against, if any. Holds the
+    /// per-tx contents and execution objects so nested resolvers can navigate the checkpoint
+    /// without re-plumbing per-yield context.
+    streamed_checkpoint: Option<Arc<ProcessedCheckpoint>>,
```
wlmyng: Curious why this is called a ProcessedCheckpoint instead of a StreamedCheckpoint

crates/sui-indexer-alt-graphql/src/scope.rs
```diff
@@ -103,23 +115,28 @@ impl Scope {
...
+    /// lookups in this scope use that transaction's own `execution_objects`.
+    pub(crate) fn with_streamed_tx(&self, tx_sequence_number_viewed_at: u64) -> Self {
         Self {
             checkpoint_viewed_at: None,
-            root_bound: self.root_bound,
-            execution_objects,
+            root_bound: None,
+            execution_objects: Arc::clone(&self.execution_objects),
+            streamed_checkpoint: self.streamed_checkpoint.clone(),
+            tx_sequence_number_viewed_at: Some(tx_sequence_number_viewed_at),
```
wlmyng: is it possible for the object cache to differ from the streamed checkpoint + tx to look at?
- `for_streamed_checkpoint` creates scope with empty btreemap...
- then `with_streamed_tx` is the anchor, which clones `self.execution_objects`

When I look at `execution_objects_in_view` it seems we expect these to be mutually exclusive, either we have object map or we have streamed_checkpoint + tx anchor

is it possible to cross these wires? I think the execution objects come from an Execution, and the stream comes from Subscription, and these are all standalone payloads so it shouldn't be possible

Would it be reasonable to model this with an Enum on Scope?

wlmyng (commented):
lgtm

crates/sui-indexer-alt-graphql/src/scope.rs
tpham-mysten: Mainly to indicate that it's not the raw streamed checkpoint and has already gone through a processing layer. It lives in the task::streaming module, so use `crate::task::streaming::ProcessedCheckpoint` makes the streaming context clear at the import.

crates/sui-indexer-alt-graphql/src/scope.rs
tpham-mysten: Yeah it makes sense! Thanks for the catch

I have refactored with ObjectSource enum.

tpham-mysten (commented):
Thanks @wlmyng!

crates/sui-indexer-alt-graphql/src/scope.rs
```diff
@@ -31,6 +32,25 @@ use crate::task::watermark::Watermarks;
...
+#[derive(Clone)]
+pub(crate) enum ObjectSource {
+    /// Reads go through the indexed-checkpoint path (kv_loader / DB). No in-memory payload.
+    Indexed,
+    /// A freshly executed transaction's input/output objects.
+    Executed {
+        execution_objects: ExecutionObjectMap,
+    },
+    /// A streamed checkpoint with all per-tx contents and execution objects. The transaction
+    /// the scope is currently resolving for is identified by [`Scope::tx_sequence_number_viewed_at`].
```
wlmyng: sentence seems a little botched, i think even "the scope is identified by tx_sequence_number_viewed_at" is fine? or like, to construct a scope you'd do that?

wlmyng (approved):
looks good to me, i was tempted to see if we could further collapse Stream and Execution, but I think this is not possible because the idea is to be able to scope into a particular txn of the same checkpoint, and looking at one of the commits it'll be used later to hydrate effects at least ...

crates/sui-indexer-alt-graphql/src/scope.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/task/streaming/processed_checkpoint.rs
```diff
@@ -16,6 +18,9 @@ pub(crate) struct ProcessedCheckpoint {
     pub(crate) contents: CheckpointContents,
     pub(crate) signature: AuthorityStrongQuorumSignInfo,
     pub(crate) transactions: Vec<ProcessedTransaction>,
+    /// Index of `transactions` by global `tx_sequence_number` for O(1) lookup that scales
+    /// across many subscribers. Built once at construction.
+    by_tx_sequence_number: HashMap<u64, usize>,
```
amnn, line 23: You do not need this data structure to perform the transformation you want -- `CheckpointSummary` contains `network_total_transactions`, and `CheckpointContents` contains the list of transactions so given a global sequence number, you can find the local transaction within the processed checkpoint by checking that the following inequality holds:

```rust
let hi = summary.network_total_transactions;
let len = contents.transactions().len() as u64;
len <= hi && (hi - len) <= tx && tx < hi
```

Then the transaction you want is `transactions[hi - len + tx]` -- constant time lookup, no additional mapping required.

crates/sui-indexer-alt-graphql/src/scope.rs
```diff
@@ -237,6 +274,20 @@ impl Scope {
...
+    /// scope's [`ObjectSource`]: in `Streamed` mode, looks up the per-tx map for the transaction
+    /// identified by `tx_sequence_number_viewed_at`; in `Executed` mode, returns the freshly
+    /// extracted map; in `Indexed` mode, returns `None` so callers fall through to the DB.
+    fn execution_objects_in_view(&self) -> Option<&ExecutionObjectMap> {
+        match &self.object_source {
+            ObjectSource::Indexed => None,
+            ObjectSource::Executed { execution_objects } => Some(execution_objects),
+            ObjectSource::Streamed { checkpoint } => checkpoint
+                .transaction(self.tx_sequence_number_viewed_at?)
+                .map(|t| &t.execution_objects),
```
amnn, line 287: One of the potential benefits of this change was that you don't need to generate a different object set for each transaction -- instead you create a single object set for the entire checkpoint, and then transactions set a bound on that to contextualise that object set for themselves (For example, you can set a bound on versions using the transaction's lampor timestamp).

amnn: Thinking about it some more, I can't think of why we need to even constrain the checkpoint's object set by the current transaction -- do you have an example of a GraphQL query that behaves differently (and incorrectly) when we don't do this, but instead make the whole object set available?

Typically, we don't support scopes that are intra-checkpoint, so this is actually a departure from that rule.

amnn (commented):
Couple of thoughts -- let me know what you think.

crates/sui-indexer-alt-graphql/src/task/streaming/processed_checkpoint.rs
tpham-mysten: Sounds good. I will fix on another PR

crates/sui-indexer-alt-graphql/src/scope.rs
tpham-mysten: Wanted to check my understanding of "we don't support scopes that are intra-checkpoint" before going further. My read is that the rule is scoped to object visibility — we don't expose object sets narrowed to a particular transaction. The Scope itself can still carry transaction-level information like tx_sequence_number_viewed_at, which 11/n's `Address.asTransactionObject` already relies on. Is that the right reading?

For intra-checkpoint object visibility specifically, I want to talk through whether removing it is actually the right direction. My intuition is that the existing intra-checkpoint visibility may not cause harm, and might even be clearer and less error-prone: under the current shape we don't accidentally leak another transaction's objects into a single transaction's view, while we still retain the option to expose the whole checkpoint's objects to a single transaction when we want to. We already have queries that are intrinsically scoped to a single transaction (`Address.asTransactionObject` in 11/n), so a scope mechanism that supports tx-level object visibility lines up with that direction rather than against it.

I couldn't construct a streaming-only query where removing intra-checkpoint visibility actively breaks correctness, but there is one example about `denyCap`, which is using `Object::with_address`:

For example: in a single checkpoint N, tx_A creates a DenyCap at v5 (empty deny list), tx_B mutates it to v7 (adds an address). Reading coinMetadata.regulated.denyCap.contents from `subscription { events { ... coinMetadata { regulated { denyCap { contents } } } } }`, rendering an event emitted by tx_A:

- Current 9/n (intra-checkpoint visibility): v5 when tx_A mutated the cap, v7 (via DB fallthrough) when it didn't.
- With the simplification (no intra-checkpoint visibility): v7 always.

For reference, v7 is what the Query API returns today (the indexer returns the latest version as of the checkpoint bound and is not scoped by transaction).

WDYT?

tpham-mysten (commented):
Thanks @amnn! For discussion about intra-checkpoint scope, please check my inline comment

crates/sui-indexer-alt-graphql/src/scope.rs
amnn: > The Scope itself can still carry transaction-level information like tx_sequence_number_viewed_at, which 11/n's `Address.asTransactionObject` already relies on

The distinction is that `Address.asTransactionObject` should rely on there being a specific transaction in scope, as opposed to using the scope to apply a bound up-to-and-including a transaction.

In fact, I think that transaction should be represented using its digest, or by caching the whole `Transaction` GraphQL type (effectively the same thing), rather than using the transaction sequence number, to make the distinction clear.

Meanwhile,

> For reference, v7 is what the Query API returns today (the indexer returns the latest version as of the checkpoint bound and is not scoped by transaction).

This, to me, is a problem. It is bad if a request that runs during subscriptions behaves differently when run as a query. This is the main reason why I raised this issue.

tpham-mysten: > In fact, I think that transaction should be represented using its digest, or by caching the whole Transaction GraphQL type (effectively the same thing), rather than using the transaction sequence number, to make the distinction clear.

Sounds good. We can add a field like `active_transaction_digest` to Scope.

> This, to me, is a problem. It is bad if a request that runs during subscriptions behaves differently when run as a query. This is the main reason why I raised this issue.

Yeah, and I'm not sure what is the right behavior, but for now, we can make the Subscription behavior match that of the query by removing the transaction-scoped object set. We can visit later if needed.

I have applied both changes in this [commit](https://github.com/MystenLabs/sui/pull/26487/changes/282d4479437665dac6bb804202f84bf236b9420f#diff-5c6ef774ad773983a6e5bd4e3acff0b0aec0fa78be22b7414135581f35a28ca7) in 10/n (#26487)

PR #26478, opened 2026-05-04
docker/sui-node: correctness fixes + better layer caching

ebmifa:
## Summary

Fixes three latent correctness issues in the sui-node Dockerfile and improves layer cache reuse. No new tooling, no BuildKit-only features — keeps the public Dockerfile portable for external community validators on stock Docker.

## Correctness fixes

- **Workspace rustflags now apply.** `.cargo/config.toml` was never copied into the build context, so workspace rustflags (`force-frame-pointers`, `force-unwind-tables`, etc.) were silently ignored. The Docker-built binary differed from `cargo build` on a checkout of the same SHA. Now copied.
- **Pinned toolchain now takes effect.** `rust-toolchain.toml` was also never copied. Without it, the container used whatever toolchain the base image happened to ship.
- **`PROFILE` arg respected end-to-end.** The runtime stage hardcoded `target/release/sui-node`, so `--build-arg PROFILE=bench` or `=debug` produced an empty image. Now uses `${PROFILE}` consistently.

## Layer cache improvements

- `cargo build --locked` — fails fast on `Cargo.lock` / `Cargo.toml` drift instead of silently re-resolving.
- COPYs reordered least- to most-frequently-changing (`external-crates → sui-execution → consensus → crates`). When a commit only touches `crates/`, layers above hit cache.
- `ARG GIT_REVISION` + `ENV` moved to right before the cargo build line. SHA changes now only invalidate the build layer, not the source COPYs above it.
- `apt --no-install-recommends` + apt list cleanup in both stages.

## Other cleanups

- `WORKDIR "$WORKDIR/sui"` → `WORKDIR /sui`. The old form relied on `$WORKDIR` being unset (which it was), so the path resolved to `/sui` by accident.
- `COPY ... /usr/local/bin` → `COPY ... /usr/local/bin/sui-node`. Explicit destination.

## Compatibility

This is the public Dockerfile that external community validators consume. Constraints kept:

- No BuildKit-only directives (no `--mount=type=cache`, no `# syntax=...`).
- No new build-time or runtime tools.
- No operator-visible behavior change (no `USER`, `ENTRYPOINT`, or `EXPOSE` shifts; same `debian:bullseye-slim` runtime).
- No post-process strip — cargo's per-profile `strip` settings are honored as written. `[profile.release]` keeps `debug = 1` + `split-debuginfo = 'packed'` for stack traces; `[profile.bench]` and `[profile.release-dbgsym]` opt out with `strip = 'none'`. An unconditional `strip` in the Dockerfile would silently undo the `--debug-symbols` flag in `build.sh`, so it's left out.

Internal mp3 builds get an additional cache-mount speedup via the rewriter overlay (MystenLabs/mp3#123), but the speedup measured below is from this Dockerfile **alone** — independent of the rewriter.

### Controlled 3-SHA experiment (Dockerfile-only, prod mp3)

Same 3 prod SHAs built on prod mp3 (no rewriter), once with the old Dockerfile and once with this PR's Dockerfile:

| sha | A: prod + old Dockerfile | B: prod + this PR | speedup |
|---|---|---|---|
| 09cb74d | 1719s | 827s | **52% faster** |
| a5f422e | 985s  | 831s | **16% faster** |
| 3f6d2dd | 1070s | 762s | **29% faster** |
| **mean** | **1258s** | **807s** | **~32% faster** |

Independent of mp3 rewriter — the Dockerfile is the consistent driver.

### Broader 15-SHA staging shadow (cron mirror)

A cron job mirrors every new prod sui-node build onto staging mp3 with this PR's Dockerfile applied (`ebmifa/26478-shadow-<sha>` branches). 15 paired builds aggregated to date:

| Metric | Prod (old Dockerfile, N4 fleet) | Staging (this PR + rewriter, N2D fleet) | Δ |
|---|---|---|---|
| Median | 1287s (21:27) | 632s (10:32) | **−50.9%** |
| Mean | 1592s (26:32) | 753s (12:33) | **−52.7%** |
| Range | 774s – 3414s | 443s – 1083s | −1.3% to −77.5% |

Caveat: not pure A/B for *this PR alone* — staging additionally runs the mp3 rewriter (mold linker + cargo cache mount). The 4-way matrix above isolates the contributions; this Dockerfile alone is the consistent winner across all builds and infrastructure tiers.

The most recent trio (b3902e5 / edbda30 / 5ea3977) finished in 7m45s / 8m21s / 7m23s on a warm cargo-target cache (28–42 GB per buildkitd pod), confirming the cache mount holds across consecutive builds without eviction.

## Considered, not taken

- **cargo-chef multi-stage.** Sui's nested Cargo workspace at `external-crates/move/` breaks `cargo chef cook`'s path-dep resolution: copying source into the deps stage trips on the move workspace's own `Cargo.lock`, and the root workspace's `[patch]` sections don't survive the `recipe.json` round-trip.
- **BuildKit cache mount on `/sui/target`.** Would require `# syntax=docker/dockerfile:1.7` and `DOCKER_BUILDKIT=1`, breaking external operators on legacy Docker.
- **Unconditional binary strip in the Dockerfile.** Cargo profiles already encode the right per-profile behavior, including `strip = 'none'` for the bench / release-dbgsym profiles that `build.sh --debug-symbols` selects. A Dockerfile-level strip would override that.

## Test plan

- [x] Build the new HEAD on mp3 staging — succeeded, sui-node binary produced and pushed to GAR + GCS.
- [x] 3 prod-mp3 builds (no rewriter, this Dockerfile only): builds 31262/31263/31264 — all succeeded; durations match the controlled-experiment table.
- [x] 15-SHA staging shadow via cron mirror (`~/scripts/mp3-shadow-26478.sh`) — all succeeded; pairings tracked in `~/.local/state/mp3-shadow-26478/pairings.csv`. Median 50.9% / mean 52.7% faster than prod.
- [x] Warm-cache repeat (builds 164/165/166 on already-loaded cache) — 7–8 min each, ~50% faster than the same SHAs' first cold-cache runs.
- [x] `cargo build --locked` failure mode — verified at unit test level via the rewriter's `--locked` parsing tests; in practice, drift causes a clear error from cargo at build time rather than silent re-resolve.
- [ ] `--debug-symbols` path: `docker/sui-node/build.sh --debug-symbols` produces an image whose `sui-node` binary still has `.symtab` / `.debug_info` sections (verify with `readelf -S` or `objdump -h`). *(Not yet exercised — strip step removed precisely so this works; needs a manual local check.)*
- [ ] Run sui-node from the new image against testnet for ≥30 min; confirm no regression (CPU, memory, no panics, no log diffs beyond `GIT_REVISION` / `BUILD_DATE`).

### DIFF
--- a/docker/sui-node/Dockerfile
+++ b/docker/sui-node/Dockerfile
@@ -1,32 +1,58 @@
 # Build application
 #
-# Copy in all crates, Cargo.toml and Cargo.lock unmodified,
-# and build the application.
-FROM rust:1.92-bullseye AS builder
-ARG PROFILE=release
-ARG GIT_REVISION
-ENV GIT_REVISION=$GIT_REVISION
-WORKDIR "$WORKDIR/sui"
-RUN apt-get update && apt-get install -y cmake clang
+# Multi-stage with cargo-chef: the deps layer caches on Cargo.lock changes
+# only, so source-only commits reuse compiled deps. No BuildKit-only
+# directives — works on plain Docker.
+#
+FROM rust:1.92-bullseye AS chef
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends cmake clang \
+ && rm -rf /var/lib/apt/lists/*
+RUN cargo install cargo-chef --locked --version 0.1.71
+WORKDIR /sui
 
+FROM chef AS planner
+COPY rust-toolchain.toml ./
+COPY .cargo .cargo
 COPY Cargo.toml Cargo.lock ./
 COPY consensus consensus
 COPY crates crates
 COPY sui-execution sui-execution
 COPY external-crates external-crates
+RUN cargo chef prepare --recipe-path recipe.json --bin sui-node
 
-RUN cargo build --profile ${PROFILE} --bin sui-node
+FROM chef AS deps
+ARG PROFILE=release
+COPY rust-toolchain.toml ./
+COPY .cargo .cargo
+COPY --from=planner /sui/recipe.json recipe.json
+RUN cargo chef cook --locked --profile ${PROFILE} --recipe-path recipe.json --bin sui-node
+
+FROM deps AS builder
+ARG PROFILE=release
+ARG GIT_REVISION
+ENV GIT_REVISION=$GIT_REVISION
+COPY rust-toolchain.toml ./
+COPY .cargo .cargo
+COPY Cargo.toml Cargo.lock ./
+COPY consensus consensus
+COPY crates crates
+COPY sui-execution sui-execution
+COPY external-crates external-crates
+# Force cargo to recompile local crates: stub mtimes from the cook stage can
+# otherwise look "newer" than real source, and cargo skips rebuilds.
+RUN find consensus crates sui-execution external-crates -name "*.rs" -exec touch {} +
+RUN cargo build --locked --profile ${PROFILE} --bin sui-node
 
 # Production Image
 FROM debian:bullseye-slim AS runtime
-# Use jemalloc as memory allocator
-RUN apt-get update && apt-get install -y ca-certificates curl
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends ca-certificates curl \
+ && rm -rf /var/lib/apt/lists/*
+WORKDIR /sui
 ARG PROFILE=release
-WORKDIR "$WORKDIR/sui"
-# Both bench and release profiles copy from release dir
-COPY --from=builder /sui/target/release/sui-node /opt/sui/bin/sui-node
-# Support legacy usages of /usr/local/bin/sui-node
-COPY --from=builder /sui/target/release/sui-node /usr/local/bin
+COPY --from=builder /sui/target/${PROFILE}/sui-node /opt/sui/bin/sui-node
+COPY --from=builder /sui/target/${PROFILE}/sui-node /usr/local/bin/sui-node
 
 ARG BUILD_DATE
 ARG GIT_REVISION
--- a/docker/sui-node/Dockerfile
+++ b/docker/sui-node/Dockerfile
@@ -26,6 +26,9 @@ ARG PROFILE=release
 COPY rust-toolchain.toml ./
 COPY .cargo .cargo
 COPY --from=planner /sui/recipe.json recipe.json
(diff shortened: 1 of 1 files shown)


docker/sui-node/Dockerfile
```diff
@@ -1,32 +1,38 @@
...
+ && rm -rf /var/lib/apt/lists/*
+WORKDIR /sui
 ARG PROFILE=release
-WORKDIR "$WORKDIR/sui"
-# Both bench and release profiles copy from release dir
-COPY --from=builder /sui/target/release/sui-node /opt/sui/bin/sui-node
-# Support legacy usages of /usr/local/bin/sui-node
-COPY --from=builder /sui/target/release/sui-node /usr/local/bin
+COPY --from=builder /sui/target/${PROFILE}/sui-node /opt/sui/bin/sui-node
+COPY --from=builder /sui/target/${PROFILE}/sui-node /usr/local/bin/sui-node
```
stefan-mysten, line 35: Can you override profile when running the command locally? Otherwise do we need another variable?

stefan-mysten (approved):
Stamping, but I am not sure I understand the change though.

docker/sui-node/Dockerfile
ebmifa: Yes — `PROFILE` is overridable end-to-end, no second variable needed. Both stages declare `ARG PROFILE=release` (lines 10 and 33), so a single `--build-arg PROFILE=<name>` propagates to:

- `cargo build --profile ${PROFILE}` in the builder stage (line 25)
- `COPY --from=builder /sui/target/${PROFILE}/sui-node …` in the runtime stage (lines 34, 35)

Local example:
```
docker build --build-arg PROFILE=bench -t sui-node:bench docker/sui-node/
```

The reason both stages need their own `ARG PROFILE=release` is that Docker scopes `ARG`s per stage — but it's still a single source of truth per build (the `--build-arg` value flows to both).

In fact this hardcoded path was the latent bug: pre-PR, the runtime stage said `COPY ... /sui/target/release/sui-node ...` literally, so passing `--build-arg PROFILE=bench` produced an empty image (cargo wrote to `target/bench/`, COPY pulled from `target/release/`).

stefan-mysten: That makes sense, thanks @ebmifa!

PR #26479, opened 2026-05-04
sui-rpc-api: subtract mock gas coin storage cost from simulate budget

bmwill:
Previously, when `simulate_transaction` estimated the gas budget for a transaction whose request omitted both BCS and a budget, the synthetic gas coin that the simulator injects (so it can run when the caller hasn't provided gas payment) bled its storage write cost into the returned `gas_cost_summary.storage_cost`. That cost is phantom — at execution time, real address-balance gas payment writes no coin (the charge becomes an `AccumulatorEvent`) and real coin gas payment writes the user-provided coin instead — so it was inflating every estimate by roughly `mock_size * obj_data_cost_refundable * storage_gas_price` MIST (~1M MIST under default protocol params).

Subtract that cost from the gas cost summary inside `estimate_gas_budget_from_gas_cost`. The simulator already returns the synthetic coin's `ObjectID` and the written object carries its `storage_rebate` field set to its storage cost (see `track_storage_mutation` / `collect_storage_and_rebate`), so the helper just looks the value up in `simulation_result.objects`.

Move the per-coin loading-cost adjustment for actually-selected coins into `select_gas` itself, gated on whether the budget came from the estimator. When `select_gas` picks address balance, no real gas coin is loaded so no loading cost is added; when it picks coins (or prepends a coin reservation), the budget is topped up by the loading cost of the real coins it selected, which is tighter than the previous worst-case assumption of `max_gas_payment_objects` (256) coins.

Also add an e2e regression test (sponsored PTB, sponsor pays gas via address balance, PTB does not use the gas coin) that asserts the simulated budget overshoots the executed net gas usage by less than `1500 * RGP` MIST. Without this fix the overshoot is ~2 * RGP * 1000 MIST due to the leaked mock-gas-coin storage cost.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
+++ b/crates/sui-e2e-tests/tests/grpc_simulate_gas_coin_tests.rs
@@ -1067,3 +1067,165 @@ async fn test_resolve_handles_coin_reservation_in_ptb_input() {
         result.err()
     );
 }
+
+// =============================================================================
+// Regression: when gas selection picks address balance, the estimated budget
+// must not include the storage cost of the synthetic gas coin used internally
+// by the simulator. Real execution charges gas via an accumulator event with
+// no gas-coin write, so any phantom storage cost in the budget is over-billing.
+// =============================================================================
+
+#[sim_test]
+async fn test_estimated_budget_excludes_mock_gas_coin_storage_for_address_balance() {
+    use shared_crypto::intent::Intent;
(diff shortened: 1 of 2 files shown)


manolisliolios (approved):
looks good, thanks!

PR #26481, opened 2026-05-04
kv rpc: configure watermark pipelines

nickvikeras:
## Description

Currently the watermark reading code defaults to the minimum all pipelines when you don't pass in the pipelines you care about. This makes it hard to add new pipelines because they will be lagging while they are backfilled. With this change, the watermark reader errors if you don't explicitly define the pipelines you care about. sui-kv-rpc defaults to the pipelines it currently reads at this moment in time. Once this change lands in mainnet, I will update to pulumi to pass these pipelines as cli args, and remove the default. Then when the new bitmap pipelines are backfilled, I will add them to the watermark check when enabling those rpc APIs in the archival service. 

The BigTableReader which is used by jsonrpc-alt (and formerly by graphql) just defaults to the 3 tables it can read from. This will be deprecated soon, so I won't bother actually updating pulumi for those services.

### DIFF
--- a/crates/sui-indexer-alt-reader/src/bigtable_reader.rs
+++ b/crates/sui-indexer-alt-reader/src/bigtable_reader.rs
@@ -10,16 +10,26 @@ use anyhow::bail;
 use async_graphql::dataloader::DataLoader;
 use prometheus::Registry;
 use sui_kvstore::BigTableClient;
+use sui_kvstore::CHECKPOINTS_PIPELINE;
 use sui_kvstore::CheckpointData;
 use sui_kvstore::KeyValueStoreReader;
+use sui_kvstore::OBJECTS_PIPELINE;
+use sui_kvstore::TRANSACTIONS_PIPELINE;
 use sui_kvstore::TransactionData;
 use sui_kvstore::TransactionEventsData;
 use sui_kvstore::WatermarkV1;
+use sui_kvstore::validate_pipeline_name;
 use sui_types::digests::TransactionDigest;
 use sui_types::messages_checkpoint::CheckpointSequenceNumber;
 use sui_types::object::Object;
 use sui_types::storage::ObjectKey;
 use tracing::warn;
+
+const DEFAULT_BIGTABLE_WATERMARK_PIPELINES: [&str; 3] = [
+    CHECKPOINTS_PIPELINE,
+    TRANSACTIONS_PIPELINE,
+    OBJECTS_PIPELINE,
+];
 
 #[derive(clap::Args, Debug, Clone, Default)]
 pub struct BigtableArgs {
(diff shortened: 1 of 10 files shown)


crates/sui-kv-rpc/src/main.rs
```diff
@@ -62,6 +67,10 @@ struct App {
     app_profile_id: Option<String>,
     #[clap(long = "checkpoint-bucket")]
     checkpoint_bucket: Option<String>,
+    /// Pipeline watermark to include when reporting GetServiceInfo checkpoint height. Repeat to
```
wlmyng, line 70: would be nice if we could do `value_delimiter = ','` so one can repeat, or do `--bigtable a,b,c,d,e`

And I think it's convention to keep the flag singular? So just `watermark_pipeline`

crates/sui-indexer-alt-reader/src/bigtable_reader.rs
```diff
@@ -38,14 +48,22 @@ pub struct BigtableArgs {
     /// Maximum gRPC decoding message size for Bigtable responses, in bytes.
     #[arg(long)]
     pub bigtable_max_decoding_message_size: Option<usize>,
+
+    /// Pipeline watermark to include when reporting the BigTable reader watermark. Repeat to
+    /// include multiple pipelines.
+    #[arg(long, value_name = "PIPELINE", value_parser = validate_pipeline_name)]
+    pub bigtable_watermark_pipeline: Vec<&'static str>,
```
wlmyng: same here, delimiter would be nice

crates/sui-indexer-alt-reader/src/kv_loader.rs
```diff
@@ -65,6 +66,11 @@ pub struct KvArgs {
     #[arg(long, alias = "bigtable-max-decoding-message-size")]
     pub kv_max_decoding_message_size: Option<usize>,
 
+    /// Bigtable pipeline watermark to include when reporting the Bigtable reader watermark.
+    /// Repeat to include multiple pipelines.
+    #[arg(long = "bigtable-watermark-pipeline", value_name = "PIPELINE", value_parser = validate_pipeline_name)]
+    pub bigtable_watermark_pipeline: Vec<&'static str>,
```
wlmyng, line 77: and here (to enable `--pipeline a,b,c`)

wlmyng (approved):
lgtm but would be great if we didn't have to explicitly repeat the flag for every pipeline we care about

amnn (commented):
@nickvikeras, for postgres, we implement this by accepting a list of the indexer configs we care about, see:

https://github.com/MystenLabs/sui/blob/ee41581e0dc6df89a239ac983f7b7d1641185b4b/crates/sui-indexer-alt-graphql/src/args.rs#L71-L74

This was to avoid a situation where we needed to track a list of pipelines in multiple places, because we could just read the indexer config to figure out which pipeline we were running (assuming the backfill was running in a separate instance with its own configuration).

Is it possible to use the same technique here? It would be good to be consistent on how we do this across the various stores.

PR #26482, opened 2026-05-04
kvstore: add tx seq digest pipeline

nickvikeras:
## Description

This adds a pipeline for a new `tx_seq_digest` table that maps a transaction sequence number to it's digest, and event count.

This mapping table is required because the bitmap that is stored in the reverse-index stores transaction sequence numbers, but the transaction contents are keyed by digest in big table.

The event count column is used to avoid over-fetching transaction rows for unfiltered event queries (i.e, if I need to fill a page of size 100 events, I know exactly which transactions to load to get those 100 events). This is just a small optimization to simplify the read path with very little cost/complexity to the write path.

The pipeline is gated behind a new --enable-alpha-pipeline arg so it won't start backfilling until we explicitly enable it in pulumi.

The transaction sequence number key is sequential which was causing hot partition issues when I was backfilling. I was also worried this could surface under heavy chain load, so I "salted" the key across 64 buckets (`seq % 64`) which allows bigtable to rebalance those buckets across nodes. This salting technique enables us to do range queries by performing multiple concurrent reads of each sequential key. We only use this capability for paginated ListEvents/ListTransactions requests with no filters set. I expect that to be a pretty uncommon request pattern, but we have flexibility with the concurrency limits we set on the client side to control big table load for those APIs.

### DIFF
--- a/crates/sui-kvstore/src/tables/tx_seq_digest.rs
+++ b/crates/sui-kvstore/src/tables/tx_seq_digest.rs
@@ -1,0 +1,157 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Direct mapping from `tx_sequence_number → (TransactionDigest, event_count)`.
+//!
+//! Row key layout: `[salt_u8][tx_seq_be_u64]` — 9 bytes, where
+//! `salt_u8 = (tx_seq % SALT_COUNT) as u8`. The salt prefix spreads writes
+//! across `SALT_COUNT` tablet prefixes so an append-only, monotonic
+//! `tx_sequence_number` doesn't funnel every write into a single trailing
+//! tablet. Within a salt bucket rows are still sorted by tx_seq, so point
+//! lookups via `multi_get` and per-shard range scans still work; a full
+//! ascending scan has to fan out across all `SALT_COUNT` buckets and
+//! merge client-side (see `scan_tx_seq_digest_stream`).
+//!
+//! `event_count` lets readers enumerate a transaction's event_seqs without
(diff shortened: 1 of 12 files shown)


crates/sui-kvstore/src/tables/tx_seq_digest.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+    Ok(tx_seq)
+}
+
+pub fn encode(digest: &TransactionDigest, event_count: u32) -> [(&'static str, Bytes); 2] {
+    [
+        (col::DIGEST, Bytes::from(digest.inner().to_vec())),
+        (
+            col::EVENT_COUNT,
+            Bytes::from(bcs::to_bytes(&event_count).unwrap()),
```
bmwill: since this is a u32, maybe a little cheaper to not go through BCS for this and just directly determine how you want it stored, be or le

bmwill (approved):
lgtm with a minor comment

crates/sui-kvstore/src/tables/tx_seq_digest.rs
```diff
@@ -0,0 +1,157 @@
...
+    k.push((tx_seq % SALT_COUNT) as u8);
+    k.extend_from_slice(&tx_seq.to_be_bytes());
+    k
+}
+
+/// Decode a salted row key, validating length and salt-byte consistency.
+pub fn decode_key(key: &[u8]) -> Result<u64> {
+    if key.len() != 9 {
+        anyhow::bail!("tx_seq_digest key not 9 bytes (got {})", key.len());
+    }
```
amnn, line 52: JFYI, [`ensure!`](https://docs.rs/anyhow/latest/anyhow/macro.ensure.html)

PR #26485, opened 2026-05-05
[forking] allow impersonating a sender 17/n

stefan-mysten:
## Description 

This PR allows to execute a transaction as an address for which we do not have the keys.

The Sui CLI now has a `--forking-mode` flag to bypass the signatures check before submitting the transaction execution.

Finally, it renames the crate to `sui-fork`.

## Test plan 

Existing CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [x] CLI: Sui CLI has a new `--forking-mode` flag to be used with the new `sui-fork` tool to allow impersonate a sender. 
- [ ] Rust SDK:
- [ ] Indexing Framework:


jordanjennings-mysten (approved):

crates/sui/src/client_commands.rs
```diff
@@ -713,6 +713,10 @@ pub struct TxProcessingArgs {
     /// private key corresponding to this address is not in keystore.
     #[arg(long, required = false, value_parser)]
     pub sender: Option<SuiAddress>,
+    /// Submit the transaction without signatures for forked networks that support sender
+    /// impersonation. This is only intended for local forked-network testing.
+    #[arg(long)]
+    pub forking_mode: bool,
```
amnn, line 719: nit: I think this should just be called `--no-sign` or something.

stefan-mysten: That makes sense, given that this is not "forking" only related. I will put up a PR to fix it.

PR #26486, opened 2026-05-05
Fix sui-kv-rpc watermark reader code

nickvikeras:
## Description

This code was bad. It was assuming there was only one column and deserializing it as BCS, which makes it impossible to add new columns. This change makes it resilient to schema changes.


ebmifa:
@evan-wall-mysten is it the correct target branch? Shouldn't it be 71?

evan-wall-mysten (approved):

ebmifa (approved):

PR #26487, opened 2026-05-05
[graphql-stream] Resolve Event.transaction in streaming mode [10/n]

tpham-mysten:
## Description 

Streaming subscribers expect `subscription { events { transaction { effects { objectChanges } } } }` to resolve fully in one round trip. Today it does not: `event.transaction` returns a digest-only stub, and chained fields under it (`effects`, `kind`, `gasInput`, `objectChanges`, etc.) come back null. Clients either issue a follow-up query against the indexed API to fill in the data, defeating the point of streaming, or live with incomplete payloads.

Also incorporates review feedback from #26476: 
- renames `ObjectSource` → `DataSource` (the enum now backs in-memory lookups for both objects and transaction contents, not just objects);
-  drops the intra-checkpoint object-visibility narrowing in `Streamed` mode so subscribers see end-of-checkpoint state, matching what the indexed Query path returns;
-  renames `tx_sequence_number_viewed_at` → `active_transaction_digest: Option<TransactionDigest>` to make clear the field identifies a specific transaction in scope, not a "viewed-at" bound.

## Test plan 

- unit tests + e2e tests

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/scope.rs
+++ b/crates/sui-indexer-alt-graphql/src/scope.rs
@@ -34,12 +34,13 @@ use crate::task::watermark::Watermarks;
 pub(crate) type ExecutionObjectMap =
     Arc<BTreeMap<(ObjectID, SequenceNumber), Option<NativeObject>>>;
 
-/// Where object lookups in this scope draw their data from. Encodes the mutually exclusive
-/// modes a [`Scope`] can be in. Indexed mode hits the database/kv_loader (no in-memory payload).
-/// Executed mode is the mutation/simulate path, with a freshly executed transaction's outputs.
-/// Streamed mode is the subscription path, with an in-memory checkpoint payload.
+/// Where in-memory lookups (objects, transaction contents, etc.) in this scope draw their data
+/// from. Encodes the mutually exclusive modes a [`Scope`] can be in. Indexed mode hits the
+/// database/kv_loader (no in-memory payload). Executed mode is the mutation/simulate path, with a
+/// freshly executed transaction's outputs. Streamed mode is the subscription path, with an
+/// in-memory checkpoint payload.
 #[derive(Clone)]
-pub(crate) enum ObjectSource {
+pub(crate) enum DataSource {
     /// Reads go through the indexed-checkpoint path (kv_loader / DB). No in-memory payload.
     Indexed,
     /// A freshly executed transaction's input/output objects.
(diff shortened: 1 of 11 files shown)


crates/sui-indexer-alt-graphql/src/scope.rs
```diff
@@ -274,6 +276,19 @@ impl Scope {
...
+    /// Lookup a transaction by digest within the streamed checkpoint backing this scope.
+    /// Returns `None` if the scope is not in [`ObjectSource::Streamed`] mode, or if no
+    /// transaction with that digest exists in the checkpoint.
+    pub(crate) fn streamed_transaction_by_digest(
+        &self,
+        digest: TransactionDigest,
+    ) -> Option<&ProcessedTransaction> {
+        match &self.object_source {
+            ObjectSource::Streamed { checkpoint } => checkpoint.transaction_by_digest(digest),
+            ObjectSource::Indexed | ObjectSource::Executed { .. } => None,
```
amnn: It's a bit strange that we are conditioning where we fetch transaction information from on something called `ObjectSource` -- I think we may need to rename it to explain its broader scope?

Also, this code path made me think about how we handle this for an executed/simulated transaction. In that case, we create a version of GraphQL's transaction structure that has its transaction and effects contents already populated. This has limitations, like this query:

```graphql
query {
  simulateTransaction(...) {
    effects {
      events {
        nodes {
          transaction {
            digest
          }
        }
      }
    }
  }
}
```

Ther inner `transaction` will be `null` because `events` does not participate in this cache-prefilling behaviour. This is not a big deal -- the example is kind of contrived -- but it is an inconsistency, and it seems like the way we deal with this for streaming may generalise well to also solve this problem. That would be nice for consistency for builders, and for us, because there's then only one pattern we need to maintain for handling this need.

amnn (approved):
Looks good, thanks  @tpham-mysten. I also think there's a further generalisation hiding in this change -- see inline comments for details.

crates/sui-indexer-alt-graphql/src/scope.rs
tpham-mysten: > It's a bit strange that we are conditioning where we fetch transaction information from on something called ObjectSource -- I think we may need to rename it to explain its broader scope?

Agreed! I have renamed to DataSource, which imo, is more general.

> Also, this code path made me think about how we handle this for an executed/simulated transaction. In that case, we create a version of GraphQL's transaction structure that has its transaction and effects contents already populated. This has limitations, like this query:

Sounds good. I have tasked it up in https://linear.app/mysten-labs/issue/DVX-2123/hydrate-transaction-in-executedsimulated-scope-parity-with-streaming, and I will tackle it in a separate PR

PR #26425, opened 2026-04-29
[graphql-stream] SSE transport + SSE-aware GraphiQL [11/n]

tpham-mysten:
## Summary

Replace the WebSocket subscription transport with Server-Sent Events. SSE is the standard transport for GraphQL subscriptions over HTTP. WebSocket was added in 1/n (#26019) as a starting point and is not in production yet, so this PR removes it outright.

## Why

For one-way streaming (server pushes events, client mostly listens — exactly our subscription shape), SSE is closer to current industry best practice than WebSocket: simpler infra (no upgrade dance, plays naturally with HTTP/2/3, standard HTTP caching/proxy semantics), debuggable with curl, native browser support. The same pattern is used by OpenAI/Anthropic completion streaming, Cloudflare Workers AI, GitHub live activity, Vercel AI SDK, etc.

## GraphiQL story

Upstream GraphiQL has no SSE fetcher today. Rather than block on the in-flight upstream PR (graphql/graphiql#4218), this PR ships a small static HTML template (`assets/graphiql.html`) that loads GraphiQL via UMD plus the `graphql-sse` UMD client and wires a custom fetcher: subscriptions go through `graphqlSse.iterate()`, queries through `fetch`. CDN deps pinned; subscription detection uses `parse` + `getOperationAST` so it handles multi-operation documents with an explicit `operationName`. Once #4218 lands, this collapses to a one-line `createGraphiQLFetcher({ url, sseUrl })` call.

## Test plan 

How did you test the new or updated feature?
- unit + e2e tests
- Start local GraphQL Server

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #26495, opened 2026-05-05
[graphql-stream] Add Address.asTransactionObject [12/n]

tpham-mysten:
## Description 

Streaming subscribers want to take an object ID that surfaces inside an event payload and immediately ask "how was this object referenced by the transaction that emitted the event?" without round-tripping to the indexed Query API. The object is also sometimes an unchanged shared input (e.g. a clock or registry), so it never appears in `objectChanges`; the resolver therefore needs to expose both the change case and the read-only consensus input case.

This PR adds:

  ```graphql
  type Address {
    asTransactionObject(transactionDigest: String): TransactionObjectRef
  }

  union TransactionObjectRef = ObjectChange | ConsensusObjectRead
  ```

`ObjectChange` and `ConsensusObjectRead` are reused as-is from the schema; no duplication. The resolver scans `effects.object_changes` first, then `effects.unchanged_consensus_objects` filtered to the ReadOnlyRoot variant. Other unchanged-input markers (cancelled, stream-ended, per-epoch) resolve to null since they do not house an object the user can navigate to.

In an events subscription, the transactionDigest argument may be omitted; the field then resolves against the transaction that emitted the parent event, via a new Scope::tx_digest_viewed_at() helper. Passing an explicit transactionDigest other than the parent event's transaction in subscription context is intentionally not supported (the document states this); use the indexed Query API for arbitrary transaction lookups.

## Test plan 

- unit + e2e tests

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425
   - #26495

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: add asTransactionObject  which allows user to query the status of address that is involved as an object in a particular transaction
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/addressable/as_transaction_object.move
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/addressable/as_transaction_object.move
@@ -1,7 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//# init --protocol-version 108 --accounts A --simulator
+//# init --protocol-version 108 --accounts A --simulator --addresses P=0x0
 
 //# programmable --sender A --inputs 1000 @A
 //> 0: SplitCoins(Gas, [Input(0)]);
@@ -35,6 +35,100 @@
   ) {
     asTransactionObject(transactionDigest: "@{digest_1}") {
       __typename
+    }
+  }
+}
+
+//# publish
+module P::M {
+    use sui::event;
+
+    public struct TestAddressEvent has copy, drop {
+        address_event_id: address,
+    }
+
+    public struct TestObject has key, store {
+        id: UID,
+        value: u64,
+    }
+
+    public fun create_object(value: u64, ctx: &mut TxContext): TestObject {
+        TestObject { id: object::new(ctx), value }
+    }
+
+    public fun mutate_and_emit(obj: &mut TestObject, new_value: u64) {
+        obj.value = new_value;
+        event::emit(TestAddressEvent {
+            address_event_id: object::id(obj).to_address(),
+        });
+    }
+}
+
+//# create-checkpoint
+
+//# programmable --sender A --inputs 7u64 @A
+//> 0: P::M::create_object(Input(0));
+//> 1: TransferObjects([Result(0)], Input(1));
+
+//# create-checkpoint
+
+//# programmable --sender A --inputs object(6,0) 99u64
+//> 0: P::M::mutate_and_emit(Input(0), Input(1));
+
+//# create-checkpoint
+
+//# run-graphql
+{
+  # `asTransactionObject` is called without a `transactionDigest`. Resolution defaults to
+  # the parent event's transaction because `EffectsContents::events` anchors each yielded
+  # `Event`'s scope to the parent transaction (with hydrated contents), and the resolver
+  # short-circuits the fetch via `Scope::active_transaction_contents_for`.
+  transactionEffects(digest: "@{digest_8}") {
+    events {
+      nodes {
+        contents {
+          extract(path: "address_event_id") {
+            asAddress {
+              asTransactionObject {
+                __typename
+                ... on ObjectChange {
+                  inputState { asMoveObject { contents { json } } }
+                  outputState { asMoveObject { contents { json } } }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
+
+//# run-graphql
+{
+  # Same implicit case as above, but entering via `transaction(...).effects` instead of
+  # `transactionEffects(...)`. `From<Transaction> for TransactionEffects` propagates the
+  # hydrated contents into the scope, and `EffectsContents::events` anchors them on each
+  # `Event` so descendants resolve against the parent transaction.
+  transaction(digest: "@{digest_8}") {
+    effects {
+      events {
+        nodes {
+          contents {
+            extract(path: "address_event_id") {
+              asAddress {
+                asTransactionObject {
+                  __typename
+                  ... on ObjectChange {
+                    inputState { asMoveObject { contents { json } } }
+                    outputState { asMoveObject { contents { json } } }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
     }
   }
 }
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/addressable/as_transaction_object.move
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/addressable/as_transaction_object.move
@@ -1,0 +1,40 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//# init --protocol-version 108 --accounts A --simulator
+
+//# programmable --sender A --inputs 1000 @A
+//> 0: SplitCoins(Gas, [Input(0)]);
+//> 1: TransferObjects([Result(0)], Input(1))
+
+//# create-checkpoint
+
+//# run-graphql
+{
+  # Address present in the tx as a newly created object → ObjectChange variant.
+  changed: address(address: "@{obj_1_0}") {
+    asTransactionObject(transactionDigest: "@{digest_1}") {
+      __typename
+      ... on ObjectChange {
+        idCreated
+        outputState { address }
+      }
+    }
+  }
+
+  # No `transactionDigest` argument outside subscription scope → null.
+  noArg: address(address: "@{obj_1_0}") {
+    asTransactionObject {
+      __typename
+    }
+  }
+
+  # Address not referenced by the tx → null.
+  notReferenced: address(
+    address: "0x000000000000000000000000000000000000000000000000000000000000dead"
+  ) {
+    asTransactionObject(transactionDigest: "@{digest_1}") {
+      __typename
+    }
+  }
+}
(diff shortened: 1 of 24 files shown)


crates/sui-indexer-alt-graphql/src/api/types/address.rs
```diff
@@ -213,6 +220,67 @@ impl Address {
         .transpose()
     }
 
+    /// How this address (interpreted as an object ID) was referenced by a specific transaction.
```
amnn, line 223: No hard wrap for doc comments.

crates/sui-indexer-alt-graphql/src/api/types/address.rs
```diff
@@ -213,6 +220,67 @@ impl Address {
...
+    ) -> Option<Result<TransactionObjectRef, RpcError>> {
+        async {
+            let Some(digest) = transaction_digest
+                .map(Into::into)
+                .or_else(|| self.scope.tx_digest_viewed_at())
+            else {
+                return Ok(None);
+            };
+
+            let contents = EffectsContents::empty(self.scope.clone())
```
amnn, line 243: Is it possible to cache the whole `TransactionEffects` in the scope, so that in case you have already lazily fetched the contents, you don't need to refetch it?

crates/sui-indexer-alt-graphql/src/api/types/transaction_object_ref.rs
```diff
@@ -0,0 +1,16 @@
...
+use async_graphql::Union;
+
+use crate::api::types::object_change::ObjectChange;
+use crate::api::types::unchanged_consensus_object::ConsensusObjectRead;
+
+/// Reference to an object as it appears in a specific transaction. The variant discriminates
+/// whether the object was changed by the transaction or read as an unchanged consensus
+/// (shared) input.
+#[derive(Union)]
+pub(crate) enum TransactionObjectRef {
```
amnn: nit: just call this `TransactionObject` as `ObjectRef` usually means just id + version + digest.

amnn (commented):
I was expecting to see the scope update in code related to `Transaction`/`TransactionEffects`/`Event` but I'm not seeing that -- where is the scope being set?

Could you also add tests for the non-subscription case?

crates/sui-indexer-alt-graphql/src/api/types/address.rs
tpham-mysten: Done.

crates/sui-indexer-alt-graphql/src/api/types/address.rs
tpham-mysten: The TransactionContents is already cached on Scope, and that's what streamed_transaction_by_digest returns (see the comment above), backed by the `ProcessedCheckpoint` payload `Scope::for_streamed_checkpoint` puts there.

However, there is actually an inefficiency though: `ProcessedTransaction.contents` was being [deep-cloned ](https://github.com/MystenLabs/sui/blob/2d1e74daec6f16a3988f0a4ee03fc3e5c1fe8d25/crates/sui-indexer-alt-graphql/src/api/types/transaction_effects.rs#L566)per subscriber in . I just updated to wrap it in Arc so the deep clone happens once at ingestion.

tpham-mysten (commented):
Thanks @amnn! For your points

> I was expecting to see the scope update in code related to Transaction/TransactionEffects/Event but I'm not seeing that -- where is the scope being set?

The scope plumbing landed in #26487 (10/n) — that PR added `Scope::streamed_transaction_by_digest` plus the streaming fast paths in `TransactionContents::fetch / EffectsContents::fetch`.

Concretely, the flow in this PR for an event-subscription query that hits asTransactionObject is:

1. as_transaction_object resolves the target digest
2. EffectsContents::empty(scope).fetch(ctx, digest)
3. scope.streamed_transaction_by_digest(digest) (O(1) HashMap lookup in the cached ProcessedCheckpoint) -> contents = Arc::new(tx.contents.clone())   // in-memory only, no DB
4. content.effects() — deserializes TransactionEffects from the cached BCS.

> Could you also add tests for the non-subscription case?

Sure added.

crates/sui-indexer-alt-e2e-tests/tests/graphql/addressable/as_transaction_object.move
amnn, line 1: Also add a test for the implicit case (easiest way is to have an example where the transaction emits an event that references the object).

crates/sui-indexer-alt-graphql/src/api/types/address.rs
amnn: That's true for streaming, but not for queries. 

If instead of `tx_digest_viewed_at` we had `Scope::effects_viewed_at(&self): TransactionEffects`, we could still instantiate that using just a digest to create an instance whose contents were not fetched yet, but if we already have `TransactionEffects` in our hand, we can reuse that, which also means re-using its potentially cached contents.

- In the subscriptions case, we would instantiate the `effects_viewed_at` using just the digest (as you are doing now), and if/when its contents are fetched, it will look in the `Scope`'s data source first (note that this is **not** a cache).
- In the query case, if `asTransactionObject` was nested under some effects, then we would already have the transaction and its contents in our hands, and we would initialize the scope with `self` in the context of `Scope`.
- In either case, if `asTransactionObject` was nested under an event, we would again initialise the effects being viewed with just a digest (if they are not already set), because we may not have fetched the event's transaction.

crates/sui-indexer-alt-graphql/src/api/types/transaction_object.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use async_graphql::Union;
+
+use crate::api::types::object_change::ObjectChange;
+use crate::api::types::unchanged_consensus_object::ConsensusObjectRead;
+
+/// An object as it appears in a specific transaction. The variant discriminates whether the
```
amnn: nit: no hard wrap

crates/sui-indexer-alt-graphql/staging.graphql
```diff
@@ -84,6 +84,16 @@ type Address implements Node & IAddressable {
 	"""
 	asObject: Object
 	"""
+	How this address (interpreted as an object ID) was referenced by a specific transaction.
+	
+	Returns `null` if the address was not referenced, or was present only as a non-object marker variant of unchanged consensus input (e.g. cancelled, stream-ended, per-epoch).
```
amnn: nit: 

```suggestion
Returns `null` if the object was not referenced, or was present only as a non-object marker variant of unchanged consensus input (e.g. cancelled, stream-ended, per-epoch).
```

crates/sui-indexer-alt-graphql/src/api/types/address.rs
tpham-mysten: Yeah makes sense! I anchored Scope at the Transaction level. Every Transaction and TransactionEffects constructor (with_contents, fetch, from_executed_transaction) sets Scope::active_transaction, so descendants like Address.asTransactionObject resolve consistently regardless of entry point.Scope::with_active_transaction_digest preserves an existing same-digest anchor, so re-entering the same transaction (e.g. via Event::transaction) doesn't drop hydrated contents.

crates/sui-indexer-alt-e2e-tests/tests/graphql/addressable/as_transaction_object.move
tpham-mysten: Sounds good. Done

crates/sui-indexer-alt-graphql/staging.graphql
tpham-mysten: Done

tpham-mysten (commented):
Thanks @amnn! I have refactored the Scope a bit to integrate the `effects_viewed_at` ideas, but with different name

crates/sui-indexer-alt-graphql/src/api/types/address.rs
```diff
@@ -213,6 +220,65 @@ impl Address {
         .transpose()
     }
 
+    /// How this address (interpreted as an object ID) was referenced by a specific transaction.
+    ///
+    /// Returns `null` if the object was not referenced, or was present only as a non-object marker variant of unchanged consensus input (e.g. cancelled, stream-ended, per-epoch).
+    ///
+    /// In an `events` subscription, the `transactionDigest` argument may be omitted; the field then resolves against the transaction that emitted the parent event.
```
amnn: nit: this comment needs to be generalised -- `transactionDigest` can be omitted if the query is scoped under any transaction. It could be because of a parent `Event`, but also because of a parent `Transaction` or `TransactionEffects`.

crates/sui-indexer-alt-graphql/src/scope.rs
```diff
@@ -49,12 +49,35 @@ pub(crate) enum DataSource {
...
+/// `Transaction.effects` on a hydrated `Transaction`). `None` for digest-only anchors (e.g.
+/// subscription resolvers, where contents live in the [`DataSource::Streamed`] checkpoint and are
+/// fetched on demand).
+#[derive(Clone)]
+struct ActiveTransaction {
+    digest: TransactionDigest,
+    contents: Option<Arc<NativeTransactionContents>>,
+}
+
+impl Debug for ActiveTransaction {
```
amnn: nit: file order

amnn (approved):
Awesome, thanks @tpham-mysten

PR #26496, opened 2026-05-05
sui-inverted-index: add inverted index crate + dimension extraction code

nickvikeras:
## Description

Adds a new crate that traverse a checkpoint an emits all of the "dimensions" that our query API indexes support. This will be used both by the bigtable and full node grpc indexers.

Splitting this off into it's own PR because it's worth focusing on the correctness of my dimension extraction logic, and considering if there are any other dimensions we want to add for the initial backfill.

Note that using the visitor pattern here to reduce allocation was critical for performance. I started with a more typical looking API that returned a `Vec<u8>` per dimension and landed on the visitor after profiling why it was slow. The visitor allows me to allocate one Vec<u8> per checkpoint instead of per dimension.

### DIFF
--- a/crates/sui-inverted-index/src/lib.rs
+++ b/crates/sui-inverted-index/src/lib.rs
@@ -1,0 +1,511 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use sui_types::accumulator_root::stream_id_from_accumulator_event;
+use sui_types::balance::Balance;
+use sui_types::effects::AccumulatorValue;
+use sui_types::effects::TransactionEffectsAPI;
+use sui_types::full_checkpoint_content::ExecutedTransaction;
+use sui_types::full_checkpoint_content::ObjectSet;
+use sui_types::object::Owner;
+use sui_types::storage::ObjectKey;
+use sui_types::transaction::TransactionDataAPI;
+
+/// A queryable dimension for the checkpoint inverted index.
+///
+/// Each variant has a unique single-byte tag used as a prefix in row keys,
+/// ensuring no two dimensions can produce the same encoded bytes.
+///
+/// Compound dimensions (MoveCall, EmitModule, EventType) use hierarchical
+/// keys: each prefix level is a valid, independently queryable key. For
+/// example, MoveCall encodes `[pkg_32]`, `[pkg_32][module]`, or
+/// `[pkg_32][module\x00function]` depending on the query specificity.
+/// The 32-byte address/package prefix is fixed-width (no separator needed),
+/// and `\x00` separates variable-length components (safe because Move
+/// identifiers cannot contain null bytes).
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum IndexDimension {
+    Sender = 0x01,
+    /// Any address whose state moved as a side effect of the transaction:
+    /// addresses that own an object after the txn (transfers in / new
+    /// owned objects), addresses that owned an object before the txn that
+    /// was mutated, transferred away, deleted, or wrapped, and addresses
+    /// whose address-balance changed via an accumulator `Balance<T>` event.
+    AffectedAddress = 0x02,
+    AffectedObject = 0x03,
+    /// Compound: `[package_32]` | `[package_32][module]` | `[package_32][module\x00function]`
+    MoveCall = 0x04,
+    /// Compound: `[package_id_32]` | `[package_id_32][module]`
+    EmitModule = 0x05,
+    /// Compound: `[type_address_32]` | `[..][module]` | `[..\x00name]` | `[..\x00name\x00instantiation_bcs]`
+    EventType = 0x06,
+    /// Authenticated event stream head id (the `SuiAddress` keying an
+    /// `accumulator_settlement::EventStreamHead` accumulator). Tx-space matches
+    /// any transaction that wrote to the stream; event-space matches the
+    /// individual events committed to the stream.
+    EventStreamHead = 0x07,
+}
+
+impl IndexDimension {
+    pub fn tag_byte(self) -> u8 {
+        self as u8
+    }
+}
+
+const COMPOUND_VALUE_SEPARATOR: u8 = 0x00;
+
+/// Visit all tx-space dimensions for a transaction.
+///
+/// `object_set` is used to resolve owners for the input and output states
+/// referenced by `tx.effects.object_changes()`. Callers typically pass
+/// `&checkpoint.object_set`.
+///
+/// The callback is invoked once per logical tx-space dimension candidate as
+/// `f(dimension, key)`, where `key` is the encoded value bytes for that
+/// dimension (the lookup key, without the dimension's tag byte prefix).
+/// Compound dimensions are emitted at every prefix level so queries at any
+/// specificity remain a single key lookup.
+///
+/// This uses a visitor rather than returning owned dimension values so the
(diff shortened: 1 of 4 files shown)


crates/sui-inverted-index/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+/// The callback is invoked once per logical tx-space dimension candidate.
+/// Compound dimensions are emitted at every prefix level so queries at any
+/// specificity remain a single key lookup.
+///
+/// This uses a visitor rather than returning owned dimension values so the
+/// extractor can reuse one scratch buffer for all compound keys emitted from a
+/// transaction instead of allocating one `Vec<u8>` per dimension.
+pub fn for_each_transaction_dimension(
+    tx: &ExecutedTransaction,
+    mut f: impl FnMut(IndexDimension, &[u8]),
```
bmwill, line 76: can we add some docs that explain that the 2nd param here is the "key"

crates/sui-inverted-index/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+            );
+            f(IndexDimension::EventType, &scratch);
+        }
+    }
+
+    for acc in tx.effects.accumulator_events() {
+        if let Some(stream_id) = stream_id_from_accumulator_event(&acc) {
+            f(IndexDimension::EventStreamHead, stream_id.as_ref());
+        }
+    }
```
bmwill, line 124: I think you're missing address balance events here. Not sure what the correct dimension for that would be though. Maybe `Recipient` should be changed to `AffectedAddress`?

crates/sui-index-dimensions/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+
+            write_move_call_value(&mut scratch, pkg, None, None);
+            f(event_idx, IndexDimension::MoveCall, &scratch);
+
+            write_move_call_value(&mut scratch, pkg, Some(module), None);
+            f(event_idx, IndexDimension::MoveCall, &scratch);
+
+            write_move_call_value(&mut scratch, pkg, Some(module), Some(function));
+            f(event_idx, IndexDimension::MoveCall, &scratch);
+        }
```
bmwill: I'm struggling to understand why these are included for each event. These are already covered by the `tx` dimensions and I don't believe this makes any sense to include here for events. Or am i missing something?

crates/sui-index-dimensions/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+        f(IndexDimension::MoveCall, &scratch);
+
+        write_move_call_value(&mut scratch, pkg, Some(module), None);
+        f(IndexDimension::MoveCall, &scratch);
+
+        write_move_call_value(&mut scratch, pkg, Some(module), Some(function));
+        f(IndexDimension::MoveCall, &scratch);
+    }
+
+    for ev in tx.events.iter().flat_map(|evs| evs.data.iter()) {
```
bmwill: These are duplicated because we can query a txn by if it emitted an event that matches a dimension?

crates/sui-index-dimensions/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+    for acc in tx.effects.accumulator_events() {
+        let AccumulatorValue::EventDigest(event_digests) = &acc.write.value else {
+            continue;
+        };
+        let Some(stream_id) = stream_id_from_accumulator_event(&acc) else {
+            continue;
+        };
+        for (idx, _digest) in event_digests {
+            f(
+                *idx as u32,
```
bmwill: Do we want a sanity check that `idx` is actually in-bounds with the events this txn emitted?

crates/sui-index-dimensions/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+        match owner {
+            Owner::AddressOwner(addr) => f(IndexDimension::Recipient, addr.as_ref()),
+            Owner::ConsensusAddressOwner { owner, .. } => {
+                f(IndexDimension::Recipient, owner.as_ref())
+            }
+            _ => {}
+        }
+    }
+
+    for change in tx.effects.object_changes() {
```
bmwill: I always have to go and relookup these but `object_changes` should include all objects, created, deleted, wrapped, unwrapped, mutated. while above we use `all_changed_objects` to only include objects that are still live after this txn.

crates/sui-index-dimensions/src/lib.rs
nickvikeras: It shouldn't be here. This function started as a copy-paste of the transaction function and this didnt get deleted.

crates/sui-index-dimensions/src/lib.rs
nickvikeras: Yeah, that was the idea.

crates/sui-index-dimensions/src/lib.rs
nickvikeras: I initially just copied this logic from  the original bloom filter code as a starting point without thinking about it much: https://github.com/MystenLabs/sui/blob/2a9d32466535fe536003a8938dd75029daccd7d9/crates/sui-indexer-alt/src/handlers/cp_blooms.rs

But I think it makes sense to use object_changes here. I don't really see any reason not to?

bmwill: Yeah I'm not saying to do anything else, this is correct. I was just saying I always forget what this api actually returns

bmwill (approved):

crates/sui-index-dimensions/src/lib.rs
amnn: This does make me wonder what the point is of iterating over `all_changed_objects`/`modified_at_versions` above? Why not iterate over `object_changes` here, and include the owners of the input and output state of the objects (if they exist) as affected addresses in that loop?

That seems like it would be the same thing, and it avoids us mixing concepts from TransactionEffects V1 (`all_changed_objects` and `modified_at_versions`) with concepts from V2 (`object_changes`).

Generally, I would recommend structuring everything in terms of `object_changes` as much as possible because V1 has a lot of confusing edge cases.

crates/sui-index-dimensions/src/lib.rs
```diff
@@ -0,0 +1,536 @@
...
+        f(IndexDimension::AffectedObject, change.id.as_ref());
+    }
+
+    for (_, package_id, module, function) in tx.transaction.move_calls() {
+        let pkg = package_id.as_ref();
+
+        write_move_call_value(&mut scratch, pkg, None, None);
+        f(IndexDimension::MoveCall, &scratch);
+
+        write_move_call_value(&mut scratch, pkg, Some(module), None);
```
amnn: For these cascading keys, would it make sense to have an API that allowed you to append a dimension? Then you could implement the cascade as:

```rust
scratch.clear()

append_dimension(ID::MoveCall, pkg);
f(ID::MoveCall, &scratch);

append_dimension(ID::Separator, module.bytes());
f(ID::MoveCall, &scratch);

append_dimension(ID::Separator, function.bytes());
f(ID::MoveCall, &scratch);
```

(requires including `0x00` as a dimension variant perhaps).

crates/sui-index-dimensions/src/lib.rs
amnn: Do you actually need to duplicate the implementation, or could you delegate to `for_each_event_dimension` here?

```rust
for_each_event_dimension(tx, |_idx, dim, key| f(dim, key))
```

amnn (approved):
LGTM, just potential clean up comments!

PR #26497, opened 2026-05-05
kvstore: add bitmap index pipelines

nickvikeras:
## Description

Adds new pipelines for generating reverse-indexes for checkpoint, transaction and event queries.

Things that are still not finalized:

- Exact bucket sizes
- I just realized we can store the `checkpoint_seq:transaction_offset`` in a single bitmap value, and then one table can serve both checkpoint and transaction level queries. It will require a schema change to the new tx_seq_digest table as well, so I will send out an additional PR to make that change later. It's still worth getting the core machinery reviewed.

The actual dimension extraction logic is in https://github.com/MystenLabs/sui/pull/26496


crates/sui-kvstore/src/tables/event_bitmap_index.rs
```diff
@@ -0,0 +1,129 @@
...
+
+/// Pack `(tx_seq, event_idx)` into a single globally-ordered u64 event_seq.
+///
+/// `event_idx` must be `< MAX_EVENTS_PER_TX`. Caller is responsible for
+/// respecting the protocol-level limit on events-per-tx.
+#[inline]
+pub fn encode_event_seq(tx_seq: u64, event_idx: u32) -> u64 {
+    debug_assert!(event_idx < MAX_EVENTS_PER_TX);
+    (tx_seq << EVENT_BITS) | (event_idx as u64)
+}
```
bmwill, line 65: we probably want/need an assert that tx_seq doesn't end up having bits shifted off the top when we left shift by EVENT_BITS. In practice it likely can't happen for a long time, but may be worth being defensive against

crates/sui-kvstore/src/store/bitmap_committer/shard.rs
```diff
@@ -0,0 +1,452 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Per-shard accumulated bitmap state, owned exclusively by a single
+//! bitmap committer shard task running on the tokio runtime.
+//!
+//! BigTable cells for the bitmap tables are written with `maxversions=1`, so
+//! each flush writes the full accumulated bitmap (not a delta). That requires
+//! keeping the bits OR'd in over the lifetime of this process in memory.
```
bmwill, line 9: I take it we can store deltas and then consolidate later because "later" we wouldn't actually know which buckets would need to be consolidated without retaining a bunch of info or fully walking a range of key-space

crates/sui-kvstore/src/store/bitmap_committer/shard.rs
```diff
@@ -0,0 +1,452 @@
...
+/// request chunking and retry/backoff.
+const SHARD_ROW_COMMIT_BATCH_SIZE: usize = 1_000;
+
+thread_local! {
+    /// Scratch buffer reused across bitmap serializations on this worker
+    /// thread. Queued row writes still copy into `Bytes` so they own their
+    /// payload, but reusing this Vec avoids one allocation per serialized row.
+    static SER_BUF: std::cell::RefCell<Vec<u8>> =
+        std::cell::RefCell::new(Vec::with_capacity(32 * 1024));
+}
```
bmwill, line 62: probably worth a note that you can't expect anything to be retained cross await points.

bmwill (approved):
lgtm, i have a few nits and idk if we want @amnn to take a look before we land this or not.

crates/sui-kvstore/src/bigtable/client/mod.rs
```diff
@@ -502,6 +525,25 @@ impl BigTableClient {
...
     }
 
+    /// Read the `bucket_start_cp` column for a bitmap-index pipeline, if
+    /// present. Returns `None` for non-bitmap pipelines and for bitmap
+    /// pipelines that haven't yet written the column.
+    pub async fn get_bitmap_bucket_start_cp(&mut self, pipeline: &str) -> Result<Option<u64>> {
+        let pipeline_key = tables::watermarks::encode_key(pipeline);
+
+        let rows = self
+            .multi_get(tables::watermarks::NAME, vec![pipeline_key.clone()], None)
```
wlmyng: Is multi get the only API available?

crates/sui-kvstore/src/store/bitmap_committer/shard.rs
nickvikeras: Yeah. This is much simpler on rocksdb with merge operators.

crates/sui-kvstore/src/store/bitmap_committer/metrics.rs
```diff
@@ -0,0 +1,86 @@
...
+                format!("bitmap_serialized_bitmap_size_bytes_{pipeline}"),
+                "Serialized-size in bytes of a bitmap flushed by the bitmap committer",
+                bitmap_size_buckets,
+                registry,
+            )
+            .unwrap(),
+        })
+    }
+
+    pub(crate) fn noop() -> Arc<Self> {
```
amnn, line 75: Confused by this -- if you are registering the metrics into an empty registry, why do you need the code in `new_with_unique_prefix`?

crates/sui-kvstore/src/store/bitmap_committer/mod.rs
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+/// Configuration and dependencies for the async bitmap committer.
+/// Populate the fields and call [`Self::spawn`] to wire the background
+/// tasks.
+pub(crate) struct BitmapCommitter {
```
amnn, line 216: (nit: file order)

crates/sui-kvstore/src/store/bitmap_committer/mod.rs
```diff
@@ -0,0 +1,443 @@
...
+                merge_rx,
+                seal_rx,
+                write_tx: shard_write_tx,
+                generation_tx,
+                initial_watermarks,
+                is_sealed,
+                startup_min_unsealed_bucket: Default::default(),
+                metrics,
+            };
+            service = service.spawn(async move { worker.run().await });
```
amnn, line 316: This should work as long as `run` accepts the worker by value (which it probably should). Same pattern applies elsewhere in this PR.

```suggestion
            service = service.spawn(worker.run());
```

crates/sui-kvstore/src/store/bitmap_committer/mod.rs
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        drop(write_tx);
+
+        service = service.spawn({
```
amnn, line 321: Should these tasks be using `spawn_aborting`? By default `spawn` does not hook the task up to the graceful shutdown process, and expects to be called with an associated `with_shutdown_signal` that is used to translate a graceful shutdown request for the service as a whole into some signal that the task is listening to, but if it's fine for the task to just abort on any yield point, then `spawn_aborting` hooks it up so that this task just aborts if the service is told to shut down.

For indexers, generally `spawn_aborting` is fine/expected, because they need to be crash-tolerant. Not doing that can cause graceful shutdown to hang for 30 seconds because the task doesn't know it needs to clean itself up, although in this case that may not happen because these workers are connected via channels to the indexing framework which will close those channels when it shuts down.

crates/sui-kvstore/src/store/bitmap_committer/shard.rs
amnn: Given this limitation, would it be easier/less error prone to make this a local variable in the task that does the serialization?

crates/sui-kvstore/src/store/bitmap_committer/shard.rs
```diff
@@ -0,0 +1,457 @@
...
+    /// This shard's slice of the commit's values. Pre-partitioned in
+    /// `Handler::batch`, so the shard can iterate directly — no per-value
+    /// shard check, no cross-shard indirection. `Arc` keeps the outer
+    /// `commit()` fan-out cheap (one refcount bump per shard instead of
+    /// cloning `BitmapIndexValue`s).
+    pub(super) values: Arc<Vec<BitmapIndexValue>>,
+}
+
+/// Generation task → shard (UNBOUNDED per-shard channel).
+pub(super) struct Seal {
```
amnn, line 79: Does this type serve a purpose? Could you just use `BucketId` in the channel type below?

crates/sui-kvstore/src/store/bitmap_committer/shard.rs
```diff
@@ -0,0 +1,457 @@
...
+
+        info!(table, shard_id, "Bitmap shard task exiting");
+        Ok(())
+    }
+
+    async fn handle_shard_merge(&mut self, merge: Merge) -> Result<()> {
+        let checkpoint = merge.checkpoint;
+        let min_bucket_to_accumulate = self.min_bucket_to_accumulate()?;
+        let mut rows_scheduled = 0u64;
+        {
```
amnn, line 132: What's up with this scope?

crates/sui-kvstore/src/store/bitmap_committer/shard.rs
```diff
@@ -0,0 +1,457 @@
...
+            }
+
+            let grew = {
+                let row = self
+                    .rows
+                    .entry(v.bucket_id)
+                    .or_default()
+                    .entry(v.row_key.clone())
+                    .or_insert_with(AccumulatedRow::empty);
+                row.or_in(v)
```
amnn, line 289: Out of interest, does it work to add the call to `or_in` to the end of the chain above?

crates/sui-kvstore/src/store/bitmap_committer/shard.rs
```diff
@@ -0,0 +1,457 @@
...
+            };
+
+            if grew {
+                changed_rows.insert((v.bucket_id, v.row_key.clone()));
+            }
+        }
+        drop(values);
```
amnn, line 296: Why iterate over values by reference and then drop it instead of iterating over values by value?

crates/sui-kvstore/src/store/bitmap_committer/writer.rs
```diff
@@ -0,0 +1,365 @@
...
+            rate_limiter,
+            generation_tx,
+            rows_written,
+            metrics,
+        };
+
+        let chunks = futures::stream::unfold(write_rx, |mut write_rx| async move {
+            let batch = write_rx.recv().await?;
+            Some((batch, write_rx))
+        })
```
amnn, line 121: nit (optional): Note that in other places we use `ReceiverStream` for this (from either `futures` or `tokio_stream`) -- this is fine, just flagging that the helper exists.

crates/sui-kvstore/src/store/mod.rs
```diff
@@ -0,0 +1,714 @@
...
+    /// Bitmap fields recorded during the first `init_watermark` call for
+    /// `pipeline_task`. Returns an error if `init_watermark` has not yet been
+    /// invoked for this pipeline — framework must drive `init_watermark` first.
+    pub(crate) fn get(&self, pipeline_task: &str) -> Result<BitmapInitialWatermark> {
+        self.init_results
+            .lock()
+            .unwrap()
+            .get(pipeline_task)
+            .map(|result| result.bitmap)
+            .ok_or_else(|| {
```
amnn, line 337: `anyhow::Context`

PR #26498, opened 2026-05-05
[Block Streaming] skip commit processing for Observer sync enabled nodes

akichidis:
## Description 

Until we properly support randomness for observer sync enabled nodes, and to allow the temporary testing of the stream via simtest and local deployments, this PR is allowing to skip the commit processing for those nodes.
This is just an intermediary step and will be reverted as soon as we have the missing pieces in place.

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


mwtian (approved):

crates/sui-core/src/consensus_handler.rs
```diff
@@ -3251,17 +3251,23 @@ impl MysticetiConsensusHandler {
...
         tasks.spawn(monitored_future!(async move {
             // TODO: pause when execution is overloaded, so consensus can detect the backpressure.
             while let Some(consensus_commit) = commit_receiver.recv().await {
                 let commit_index = consensus_commit.commit_ref.index;
-                if commit_index <= last_processed_commit_at_startup {
+                if !process_consensus_commits {
+                    debug!(
+                        commit_index,
+                        "Observer skipping consensus commit processing"
+                    );
```
akichidis, line 3269: Tagging a couple of more people just to raise awareness of this new condition. 
This is temporary and will be removed once we have the right bits in place to properly process commits for the observer sync full nodes.

@mlogan @aschran @alex-mysten

PR #26500, opened 2026-05-05
[forking] Fix how the --data-dir stores the files 18/n

stefan-mysten:
## Description 

This PR removes the extra folder creation when `--data-dir` is passed. With this fix, the `objects | checkpoints | transactions | etc` data will live directly in the directory passed through the flag.

In addition, the `seed_manifest.json` will be written always to disk with some metadata; if address/objects seeding is provided at startup, then it will be persisted too, otherwise those data will be empty. This is needed to ensure easy resume of the fork (knows at which checkpoint was forked, and which is the latest available checkpoint locally).

## Test plan 

Existing CI + new tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-fork/README.md
+++ b/crates/sui-fork/README.md
@@ -45,9 +45,9 @@ That means that you have full control over the advancement of checkpoints, time,
     `sui-fork start` with no flags forks from mainnet at the latest checkpoint.
     The fork serves Sui gRPC on `127.0.0.1:9000` by default.
 
-    Local resume state is stored under
-    `{data-dir}/{network}/forked_at_{checkpoint}`. When you restart the same
-    fork, reuse the same `--data-dir`, `--network`, and `--checkpoint`.
+    In this example, local resume state is stored directly under
+    `/tmp/sui-fork-demo`. When you restart the same fork, reuse the same fork
+    data directory and omit seed flags.
 
 3. In terminal 2, confirm the fork is reachable:
 
@@ -112,13 +112,15 @@ Owned-object enumeration can be seeded when starting the fork:
 sui-fork start --checkpoint 12345678 --address 0x... --object 0x...
 ```
 
-- `--data-dir` is the exact filesystem root for local fork data. Objects,
-  checkpoints, transactions, indices, and `seed_manifest.json` are written
-  directly under that directory.
+- `--data-dir <path>` is the exact filesystem root for local fork data.
+  Objects, checkpoints, transactions, indices, and `seed_manifest.json` are
+  written directly under that directory.
 - Without `--data-dir`, the default root is
-  `{base_path}/{network}/forked_at_{checkpoint}`, where the base path is the folder based on XDG_DATA_HOME or 
-  HOME on Unix systems, or AppData on Windows, and creates a `sui-fork-data` (or `.sui-fork-data`) directory.
-    `SUI_FORK_DATA` sets `{base_path}` and the rest will be as above.
+  `{base_path}/{network}/forked_at_{checkpoint}`.
+- `SUI_FORK_DATA` sets `{base_path}` directly.
+- On Unix, the default `{base_path}` is `$XDG_DATA_HOME/sui_fork_data` when
+  `XDG_DATA_HOME` is set, otherwise `$HOME/.sui_fork_data`.
+- On Windows, the default `{base_path}` is `%APPDATA%/sui_fork_data`.
 - `--address` is repeatable and seeds metadata for every object owned by that
   address at the fork checkpoint. Address seeding requires a checkpoint in the
   GraphQL object enumeration range.
--- a/crates/sui-fork/README.md
+++ b/crates/sui-fork/README.md
@@ -116,9 +116,9 @@ sui-fork start --checkpoint 12345678 --address 0x... --object 0x...
   checkpoints, transactions, indices, and `seed_manifest.json` are written
   directly under that directory.
 - Without `--data-dir`, the default root is
-  `{base_path}/{network}/forked_at_{checkpoint}`. `FORKING_DATA_STORE` sets
-  `{base_path}` directly; otherwise the base path uses `.sui_fork_data` under
-  the Sui config directory or home directory.
+  `{base_path}/{network}/forked_at_{checkpoint}`, where the base path is the folder based on XDG_DATA_HOME or 
+  HOME on Unix systems, or AppData on Windows, and creates a `sui-fork-data` (or `.sui-fork-data`) directory.
+    `FORK_DATA_STORE` sets `{base_path}` and the rest as above.
 - `--address` is repeatable and seeds metadata for every object owned by that
   address at the fork checkpoint. Address seeding requires a checkpoint in the
   GraphQL object enumeration range.
--- a/crates/sui-fork/README.md
+++ b/crates/sui-fork/README.md
@@ -112,20 +112,29 @@ Owned-object enumeration can be seeded when starting the fork:
 sui-fork start --checkpoint 12345678 --address 0x... --object 0x...
 ```
 
+- `--data-dir` is the exact filesystem root for local fork data. Objects,
+  checkpoints, transactions, indices, and `seed_manifest.json` are written
+  directly under that directory.
+- Without `--data-dir`, the default root is
+  `{base_path}/{network}/forked_at_{checkpoint}`. `FORKING_DATA_STORE` sets
+  `{base_path}` directly; otherwise the base path uses `.sui_fork_data` under
+  the Sui config directory or home directory.
 - `--address` is repeatable and seeds metadata for every object owned by that
   address at the fork checkpoint. Address seeding requires a checkpoint in the
   GraphQL object enumeration range.
 - `--object` is repeatable and fetches that object at the fork checkpoint. If
   the object is address-owned, it is added to the initial owned-object index.
-- Seed metadata is written once to `seed_manifest.json` under the fork
-  directory. The manifest is immutable.
+- Fork metadata is written once to `seed_manifest.json` under the fork
+  directory. The manifest is immutable and records the source network, original
+  fork checkpoint, and any optional seed object metadata. When no seed inputs
+  are provided, it is still written with an empty seed entry list.
 
-When restarting with the same `--data-dir`, `--network`, and `--checkpoint`,
-omit seed flags. If a seed manifest already exists and any seed flags are
-provided, startup fails instead of overwriting or reinterpreting the local
-state. Resume starts from the highest locally persisted checkpoint and keeps
-the durable owned-object index and deleted-object markers authoritative over
-the original seed manifest.
+When restarting with the same fork data directory, omit seed flags. If a seed
+manifest already exists and any seed flags are provided, startup fails instead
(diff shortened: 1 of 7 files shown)


crates/sui-fork/README.md
```diff
@@ -54,20 +54,29 @@ Owned-object enumeration can be seeded when starting the fork:
 sui-fork start --checkpoint 12345678 --address 0x... --object 0x...
 ```
 
+- `--data-dir` is the exact filesystem root for local fork data. Objects,
+  checkpoints, transactions, indices, and `seed_manifest.json` are written
+  directly under that directory.
+- Without `--data-dir`, the default root is
+  `{base_path}/{network}/forked_at_{checkpoint}`. `FORKING_DATA_STORE` sets
+  `{base_path}` directly; otherwise the base path uses `.sui_fork_data` under
+  the Sui config directory or home directory.
```
stefan-mysten: @mdgeorge4153 this is the updated logic for handling --data-dir. Let's discuss tomorrow!

crates/sui-fork/src/startup.rs
```diff
@@ -38,7 +97,7 @@ pub async fn initialize(
     node: Node,
     forked_at_checkpoint: CheckpointSequenceNumber,
     version: &str,
-    data_dir: Option<std::path::PathBuf>,
+    data_dir: Option<PathBuf>,
```
mdgeorge4153, line 107: I think when you're passing around pathbufs it's very easy to forget what they point to. I'd recommend at least including that in the doc comment. In this case it's pretty clear, because the variable is called data_dir, but I always worry about losing track of what paths are.

crates/sui-fork/src/filesystem.rs
```diff
@@ -147,35 +154,40 @@ pub(crate) struct FilesystemStore {
...
     /// Resolve the default base path for on-disk storage.
     pub(crate) fn base_path() -> Result<PathBuf, Error> {
-        let home_dir = std::env::var("FORKING_DATA_STORE")
-            .or_else(|_| std::env::var("SUI_CONFIG_DIR"))
-            .or_else(|_| std::env::var("HOME"))
-            .or_else(|_| std::env::var("USERPROFILE"))
+        Self::base_path_from_env(|key| env::var(key))
+    }
+
+    fn base_path_from_env(
```
mdgeorge4153, line 193: A comment here would be nice. I think it returns the first of:
 - $FORKING_DATA_STORE
 - $SUI_CONFIG_DIR/.sui_fork_data
 - $HOME/.sui_fork_data
 - $USERPROFILE/.sui_fork_data

mdgeorge4153: I know there's a pretty detailed description of how `jf` computes its data dir which I think @amnn did research into the xdg standards; maybe we should use the same logic here?

crates/sui-fork/src/filesystem.rs
```diff
@@ -147,35 +154,40 @@ pub(crate) struct FilesystemStore {
...
+        mut get_env: impl FnMut(&str) -> Result<String, env::VarError>,
```
mdgeorge4153: I'm curious why `mut` and `FnMut` here, probably some rust I don't fully understand

mdgeorge4153: also, is the closure just for testing?

crates/sui-fork/src/filesystem.rs
```diff
@@ -184,6 +196,26 @@ impl FilesystemStore {
         Ok(PathBuf::from(home_dir).join(DATA_STORE_DIR))
     }
 
+    fn default_root(
```
mdgeorge4153, line 224: a doc comment here showing what this looks like would be helpful. I think it's `$BASE/<network>/forked_at_<checkpoint>`

crates/sui-fork/src/tests/filesystem.rs
```diff
@@ -31,6 +31,56 @@ fn test_store() -> (tempfile::TempDir, FilesystemStore) {
...
+    let base = dir.path().join(DATA_STORE_DIR);
+
+    let root = FilesystemStore::root_from_base(base.clone(), &crate::Node::Testnet, 99);
+
+    assert_eq!(root, base.join("testnet").join("forked_at_99"));
+}
+
+#[test]
+fn forking_data_store_env_is_used_as_default_base_path_directly() {
+    let base = FilesystemStore::base_path_from_env(|key| match key {
```
mdgeorge4153: It might be useful if this get_env included some of the other variables as well, to ensure that the precedence remains unchanged.

crates/sui-fork/src/cli.rs
```diff
@@ -49,8 +49,8 @@ enum Command {
         #[arg(long)]
         checkpoint: Option<u64>,
 
-        /// Base directory for on-disk storage (overrides FORKING_DATA_STORE env var)
-        #[arg(long, env = "FORKING_DATA_STORE")]
+        /// Exact directory for on-disk storage
+        #[arg(long)]
```
mdgeorge4153, line 56: it might be good to have a snapshot test for the help message; I'm not sure if this becomes `--data_dir` or `--data-dir` by default.

crates/sui-fork/src/cli.rs
```diff
@@ -191,18 +200,34 @@ async fn cmd_start(
...
+        current_checkpoint,
+        resuming: resolved_start.resuming,
     };
     print_output(&output, json_output);
 
-    info!(
-        "Starting forked network from {} at checkpoint {} (rpc on {})",
-        network_name, checkpoint, rpc_addr,
-    );
+    if resolved_start.resuming {
```
mdgeorge4153, line 230: can this just be `info!("{resolved_start}")` with the `Display` instance below?

crates/sui-fork/src/startup.rs
```diff
@@ -23,12 +25,69 @@ use sui_types::sui_system_state::{SuiSystemState, SuiSystemStateTrait};
...
+///
+/// Explicit data directories can be inspected directly before the remote
+/// GraphQL client asks for latest. Default roots cannot be inspected without a
+/// checkpoint because their path contains the checkpoint.
+pub(crate) fn resolve_start_checkpoint_from_local(
+    node: &Node,
+    requested_checkpoint: Option<CheckpointSequenceNumber>,
+    data_dir: Option<&Path>,
+) -> Result<ResolvedStartCheckpoint> {
+    let local = match (data_dir, requested_checkpoint) {
```
mdgeorge4153, line 57: If I'm reading this correctly, then running `sui-fork start` with no data dir or checkpoint will cause this to return `{checkpoint: None, resuming: false}`, which will then cause a graphQL query to determine the latest checkpoint; it will then use `<BASE>/<network>/forked_at_<latest>/`. I'm not sure what happens if that fork already exists - reasonable possibilities are clobber the old fork, resume the old fork, or fail. What happens, and is it the right thing?

crates/sui-fork/README.md
mdgeorge4153: lgtm!

crates/sui-fork/src/startup.rs
```diff
@@ -23,12 +25,69 @@ use sui_types::sui_system_state::{SuiSystemState, SuiSystemStateTrait};
...
+    };
+
+    let Some(local) = local else {
+        return Ok(ResolvedStartCheckpoint {
+            checkpoint: requested_checkpoint,
+            resuming: false,
+        });
+    };
+
+    if local.seed_manifest_exists() {
```
mdgeorge4153, line 70: are you concerned with concurrency at all? I wonder if you need some file locking here

mdgeorge4153 (commented):
left a bunch of questions / comments

crates/sui-fork/src/filesystem.rs
stefan-mysten: Fixed by using XDG_DATA_HOME as the root path if no folder is provided.

crates/sui-fork/src/startup.rs
stefan-mysten: > t will then use <BASE>/<network>/forked_at_<latest>/. I'm not sure what happens if that fork already exists - reasonable possibilities are clobber the old fork, resume the old fork, or fail. What happens, and is it the right thing?

If fork data directory exists, it will resume. We expect that people would stop and restart as needed, from the same directory. However, in this case, it's unlikely that it will start from the same checkpoint except:
- somebody starts X processes in parallel and somehow by luck, they both end up at the same checkpoint
- GraphQL indexer has stopped processing checkpoints, so it will always retrieve the latest checkpoint and it will be the same, and thus it will simply restart from that fork data directory.

crates/sui-fork/src/startup.rs
stefan-mysten: Not too concerned just yet, but we need to explain well in the readme that it is not intended to spin up one forked network and submit many parallel transactions (e.g., in testing infra); it works different than a normal local network.

crates/sui-fork/src/filesystem.rs
stefan-mysten: Good catch, we don't need it. A simple `impl FnOnce` is sufficient as we don't need to mutate the closure object while we call it.

crates/sui-fork/src/cli.rs
stefan-mysten: I think it's fine for the help message to change, if needed. Not sure about the second part of the question; clap will transform this field into a kebab-case argument, unless we explicitly tell it what convention style to use.

I updated the doc comment / help message, maybe that helps.

crates/sui-fork/src/tests/filesystem.rs
stefan-mysten: Added more tests here but that required to revert the change from impl FnOnce to the initial mut, to allow us to inject various envs.

crates/sui-fork/src/startup.rs
mdgeorge4153: If it almost always restarts, I might just make it always restart and require --resume to not restart. But I don't feel strongly about it

crates/sui-fork/src/startup.rs
mdgeorge4153: The concern here is more about spinning up lots of forks in parallel, rather than submitting a bunch of txs against one fork. But it makes sense to document the problem and resolve it later.

mdgeorge4153 (approved):

crates/sui-fork/src/startup.rs
stefan-mysten: Oh, by restart I mean that if the folder exists, it will resume - no clobbering, no failing. I used them interchangeably (resume == restart). There is no "overwrite" command now, the folder needs to get deleted or the user needs to specify a different folder if they do not want to restart. I will document this.

crates/sui-fork/src/startup.rs
stefan-mysten: I think I understand now, starting multiple forks from the same seed file. I think that's fine, as long as it's mainly reads, but yeah, I will add this to the list of possible hardenings.

PR #26504, opened 2026-05-05
[protocol-config] Enable address balances and gasless transactions on mainnet (v125)

mystenmark:
## Summary

Bumps `MAX_PROTOCOL_VERSION` to 125, enabling on mainnet the feature flags currently testnet-only:

- `enable_address_balance_gas_payments` (address balances for gas)
- `enable_gasless` + empty `gasless_allowed_token_types` (gasless token list set in a follow-up PR)
- `enable_coin_reservation_obj_refs` (coin reservations)
- `enable_accumulators`, `enable_authenticated_event_streams`, `enable_object_funds_withdraw`, `convert_withdrawal_compatibility_ptb_arguments`
- `split_checkpoints_in_consensus_handler`, `include_checkpoint_artifacts_digest_in_summary`

Verified via `Mainnet_version_125.snap` vs `Testnet_version_125.snap`: only chain-specific data remains (`aliased_addresses`, `gasless_allowed_token_types`, linkage amendments) — no feature-flag differences.

Also drops the obsolete "mainnet override should disable AB / coin reservation" branches in two `grpc_simulate_gas_coin_tests` per the test author's inline TODO, and refreshes the `execution_time_observer` and `populated_genesis` snapshots that always need a refresh after a `MAX_PROTOCOL_VERSION` bump.

## Release notes

- [x] Protocol: Enables address balances, gasless transactions, coin reservations, and the accumulator/withdraw stack on mainnet at protocol version 125. Gasless transactions are gated behind an empty token allow-list until a follow-up configures the mainnet tokens.


tnowacki (approved):
Should we remove the various functions like `enable_address_balance_gas_payments_for_testing` and  `enable_gasless_for_testing`? I'm mostly thinking of making sure that our tests are now actually grabbing the flags from the protocol version, and not some separate test config

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4898,6 +4904,31 @@ impl ProtocolConfig {
...
+                    cfg.feature_flags.enable_address_balance_gas_payments = true;
+                    cfg.feature_flags.enable_authenticated_event_streams = true;
+                    cfg.feature_flags.enable_coin_reservation_obj_refs = true;
+                    cfg.feature_flags.enable_object_funds_withdraw = true;
+                    cfg.feature_flags
+                        .convert_withdrawal_compatibility_ptb_arguments = true;
+                    cfg.feature_flags.split_checkpoints_in_consensus_handler = true;
+                    cfg.feature_flags
+                        .include_checkpoint_artifacts_digest_in_summary = true;
+                    cfg.feature_flags.enable_gasless = true;
```
alex-mysten, line 4922: would it make sense to include all feature flags related to AB here despite redundancy, including ones that were turned on for all chains in previous protocol config versions, just to ensure and make it absolutely clear we have the complete set of functionality we need?

mystenmark: We could - I already checked this by diffing against the testnet snapshots which felt good enough to me.

alex-mysten (approved):

PR #26508, opened 2026-05-06
[move][move-vm] Support pinned / system-level packages in the move runtime

cgswords:
## Description 

Adds support for pinned system-level packages in the Move VM runtime. A host (e.g. sui-adapter) hands the runtime a set of SerializedPackages at construction time. Those packages are validated and JIT'd once, kept alive for the runtime's lifetime, and used by the JIT translator to rewrite cross-package calls into them as direct function pointers instead of the usual virtual-table dispatch — eliminating the per-call vtable lookup for stdlib/framework calls.

- A pinned package must be identity-linked (original_id == version_id) — i.e. a v0 package — and its type_origin_table must only declare types defined by itself. Both checks happen on the serialized form before JIT; rejects are logged and skipped.
- At JIT time, a user package's call into a pinned package is rewritten to a direct pointer only if the user package's linkage table maps OriginalId → that pinned VersionId exactly. So if the host has a non-standard linkage definition for that package (if even possible), we naturally fall back to virtual dispatch.
- Pointer lifetime: pinned Arc<Package>s live in the cache for the cache's lifetime, so any user-package JIT'd output that holds direct pointers into them is automatically outlived by its pointees.

## Test plan 

New tests, plus others all run as expected.
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
```diff
@@ -64,6 +70,27 @@ impl MoveCache {
...
+    }
+
+    /// Register a pinned system package keyed by its `OriginalId`. Returns `true` if newly
+    /// inserted, `false` if a package was already registered at that id (which the system-pkg
+    /// install loop hits naturally when `resolve_packages` returns previously-installed
+    /// siblings as cache hits — caller decides whether to log).
+    pub(crate) fn add_system_package(&mut self, pkg: Arc<Package>) -> bool {
+        use std::collections::btree_map::Entry;
+        let id = pkg.runtime.original_id;
+        match Arc::make_mut(&mut self.system_packages).entry(id) {
```
tzakian: should this be `Arc::get_mut` (and error, don't insert on `None`) instead of `Arc::make_mut`? 

`make_mut` is CoW if multiple strong references exist, and cloning to write seems like it could create weird behavior? IMO better to not cache consistently. 

Another question: would it be better that `system_packages` be an `Arc<RefCell<...>>`? Or I would also be fine with an `Arc<RwLock<...>>` since we should only be hitting read paths in any hot pathways.

external-crates/move/crates/move-vm-runtime/src/jit/execution/translate.rs
```diff
@@ -114,22 +124,51 @@ impl PackageContext<'_> {
...
+            self.system_packages
+                .get(&vtable_entry.package_key())
+                .and_then(|sys_pkg| {
+                    sys_pkg
+                        .runtime
+                        .vtable
+                        .functions
+                        .get(vtable_entry.intra_package_key())
+                })
+        } else {
```
tzakian: nit: if-let instead of contains followed by get

```suggestion
        } else if let Some(sys_pkg) = self
            .system_packages
            .get(&vtable_entry.package_key())
        {
            // System-package call: only direct-resolve if the caller has handed us a pinned
            // version that matches the user's linkage.
           
                    sys_pkg
                        .runtime
                        .vtable
                        .functions
                        .get(vtable_entry.intra_package_key())
        } else {
```

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -42,20 +43,142 @@ pub struct MoveRuntime {
...
+        // Split borrows of `self` so that the `&mut cache` borrow can coexist with the
+        // immutable `&telemetry` / `&natives` borrows below.
+        let Self {
+            cache,
+            telemetry,
+            natives,
+            ..
+        } = self;
+        let Some(cache) = Arc::get_mut(cache) else {
+            error!(
```
tzakian, line 158: debug_assert false here as well plz

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -42,20 +43,142 @@ pub struct MoveRuntime {
...
+    fn install_system_packages(&mut self, system_packages: SystemPackages) {
+        if system_packages.is_empty() {
+            return;
+        }
+
+        let filtered: Vec<SerializedPackage> = system_packages
+            .iter()
+            .filter(|pkg| {
+                if pkg.original_id != pkg.version_id {
+                    error!(
```
tzakian, line 107: debug_assert

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -42,20 +43,142 @@ pub struct MoveRuntime {
...
+                        "System package skipped: version_id != original_id (must be identity-linked)",
+                    );
+                    return false;
+                }
+                if let Some((name, defining_id)) = pkg
+                    .type_origin_table
+                    .iter()
+                    .find(|(_, def_id)| **def_id != pkg.original_id)
+                {
+                    error!(
```
tzakian, line 123: debug_assert as well plz 🙏🏻

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -42,20 +43,142 @@ pub struct MoveRuntime {
...
+                    natives,
+                    std::iter::once(version_id).collect(),
+                )
+            });
+            match result {
+                Ok(resolved) => {
+                    for (_, p) in resolved {
+                        let original_id = p.runtime.original_id;
+                        if !cache.add_system_package(p) {
+                            error!(
```
tzakian, line 194: debug_assert here and below plz

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -42,20 +43,142 @@ pub struct MoveRuntime {
...
+            error!(
+                "install_system_packages: cache Arc is shared; system packages will not be installed",
+            );
+            return;
+        };
+
+        for pkg in resolver.iter() {
+            let version_id = pkg.version_id;
+            let result = telemetry.with_transaction_telemetry(|tx_telemetry| {
+                package_resolution::resolve_packages(
```
tzakian, line 181: Can you add a comment here or somewhere around on why `resolve_packages` shouldn't be called "in bulk" on all of the packages?

external-crates/move/crates/move-vm-runtime/src/shared/system_packages.rs
```diff
@@ -0,0 +1,106 @@
...
+//! consults the pinned set to rewrite cross-package calls into them as direct function pointers;
+//! soundness rests on the `Arc<Package>`s living at least as long as every user package
+//! compiled against them, which is true so long as those Arcs sit in the same cache.
+
+use move_core_types::{
+    account_address::AccountAddress,
+    resolver::{ModuleResolver, SerializedPackage},
+};
+use tracing::error;
+
```
tzakian, line 22: formatting nit: remove blank line

external-crates/move/crates/move-vm-runtime/src/shared/system_packages.rs
```diff
@@ -0,0 +1,106 @@
...
+use std::{collections::BTreeMap, convert::Infallible};
+
+/// Public input type. Mirrors `NativeFunctions`: built by the host, handed to
+/// `MoveRuntime::new_with_system_packages`. Pinned for the lifetime of the runtime.
+#[derive(Debug, Default)]
```
tzakian: Can we remove `Default`? Default for `SystemPackages` just feels wrong to me? 

But I'm a Default hater, so maybe it's just me..

external-crates/move/crates/move-vm-runtime/src/shared/system_packages.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub fn empty() -> Self {
+        Self { packages: vec![] }
+    }
+
+    /// Build a `SystemPackages` from a vector of serialized packages. Inputs sharing a
+    /// `version_id` are deduplicated (first occurrence wins) with a logged error per duplicate;
+    /// a duplicate `version_id` in the host's input list is a host bug, not a runtime error,
+    /// so we surface it loudly rather than aborting construction.
+    pub fn new(packages: Vec<SerializedPackage>) -> Self {
+        let mut seen = std::collections::BTreeSet::new();
```
tzakian: nit: import `BTreeSet` for consistency?

external-crates/move/crates/move-vm-runtime/src/shared/system_packages.rs
```diff
@@ -0,0 +1,106 @@
...
+        let mut out = Vec::with_capacity(packages.len());
+        for pkg in packages {
+            if !seen.insert(pkg.version_id) {
+                error!(
```
tzakian, line 44: debug_assert plz

external-crates/move/crates/move-vm-runtime/src/shared/system_packages.rs
```diff
@@ -0,0 +1,106 @@
...
+                .collect(),
+        }
+    }
+}
+
+/// Internal resolver wrapping the deduplicated input set, keyed by `version_id` for the
+/// `ModuleResolver` callbacks the load/verify/JIT pipeline expects.
+#[derive(Debug)]
+pub(crate) struct SystemPackageResolver {
+    by_version_id: BTreeMap<AccountAddress, SerializedPackage>,
```
tzakian, line 91: Should this be an `IndexMap`? Right now we are somewhat lucky that the init order that we want for system packages follows ordering (0x1 -> 0x2 -> 0x3) so a `BTreeMap` works, but if that ever changes (i.e., og id ordering doesn't follow init ordering) we may run into issues where a dependent module, even if it were provided in the correct order, would be added out-of-order. 

If you'd prefer to keep it as a BTreeMap, tt the very least worth a comment I think.

external-crates/move/crates/move-vm-runtime/src/unit_tests/move_packages/system_packages_chain.move
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Two system packages: `0x3::pinned_b` calls into `0x1::pinned_a`. Used to verify cross-system
+// direct-call rewriting at install time. `pinned_b` lives at `0x3` (not `0x2`) so its address
+// doesn't alias the `0x2` user-pkg fixture used by the basic test.
+module 0x1::pinned_a {
```
tzakian, line 7: if you replicate this test with 0x1/0x3 swapped I think you'll get a different result (because of the BTreeMap in the SystemPackageResolver).

external-crates/move/crates/move-vm-runtime/src/unit_tests/system_packages_tests.rs
```diff
@@ -0,0 +1,398 @@
...
+    let mut adapter = InMemoryTestAdapter::new_with_runtime(runtime);
+    adapter.insert_package_into_storage(lib);
+    adapter
+        .publish_package(USER_ADDR, user.into_serialized_package())
+        .unwrap();
+}
+
+// 7. A user pkg whose linkage maps the system OriginalId to a *different* version ID must
+//    NOT direct-call into our pinned system pkg. This guards against a host that maps
+//    `0x1 -> some_other_version` accidentally pulling our global into their dispatch.
```
tzakian: dios mio 🫨

external-crates/move/crates/move-vm-runtime/src/unit_tests/system_packages_tests.rs
```diff
@@ -0,0 +1,398 @@
...
+    assert_eq!(
+        runtime.cache().system_packages().len(),
+        0,
+        "system pkg with mismatched defining_id should be rejected",
+    );
+}
+
+// 6. Validation failure logs and skips, runtime still constructs and is usable.
+#[test]
+fn validation_failure_logs_and_continues() {
```
tzakian, line 225: Debug asserts will cause this to panic, which is kinda annoying as this is a good test. 🤔

sui-execution/latest/sui-adapter/src/adapter.rs
```diff
@@ -44,12 +44,44 @@ mod checked {
     ) -> Result<MoveRuntime, SuiError> {
         let native_functions =
             NativeFunctions::new(natives).map_err(|_| SuiErrorKind::ExecutionInvariantViolation)?;
-        Ok(MoveRuntime::new(
+        Ok(MoveRuntime::new_with_system_packages(
             native_functions,
             vm_config(protocol_config),
+            system_packages_for_runtime(),
```
tzakian: This isn't quite right. There will need to be some changes in `advance_epoch` and probably some refactoring around that and `process_system_packages`.

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -42,20 +43,142 @@ pub struct MoveRuntime {
...
+                        "System package skipped: type defining_id does not match original_id",
+                    );
+                    return false;
+                }
+                true
+            })
+            .cloned()
+            .collect();
+
+        let resolver = SystemPackages::new(filtered).into_resolver();
```
cgswords: We should in fact retain `load_order_keys` from filtered before creating the resolve, and load in that order.

external-crates/move/crates/move-vm-runtime/src/runtime/mod.rs
```diff
@@ -42,20 +43,142 @@ pub struct MoveRuntime {
...
+            ..
+        } = self;
+        let Some(cache) = Arc::get_mut(cache) else {
+            error!(
+                "install_system_packages: cache Arc is shared; system packages will not be installed",
+            );
+            return;
+        };
+
+        for pkg in resolver.iter() {
```
cgswords: This should iterate the `load_order_keys`; `iter` should be entirely removed from `resolver` as the BTreeMap order is not guaranteed and this is _not safe_.

external-crates/move/crates/move-vm-runtime/src/runtime/package_resolution.rs
```diff
@@ -261,3 +280,26 @@ pub(crate) fn jit_and_cache_package(
...
+
+/// Filter a set of retained system packages down to those that the user package's linkage table
+/// maps `OriginalId -> our pinned VersionId` exactly. Anything else (no entry, wrong version) is
+/// excluded; the JIT translator only direct-resolves into a system pkg when the user has
+/// explicitly linked to that specific pinned version.
+fn effective_system_packages_for(
+    system_packages: &BTreeMap<OriginalId, Arc<Package>>,
+    verified_pkg: &verification::ast::Package,
+) -> BTreeMap<OriginalId, Arc<Package>> {
+    if system_packages.is_empty() {
```
cgswords, line 296: This should exclude the actual package ID for `verified_pkg` so that we never end up with the package we are publishing in the system package table.

external-crates/move/crates/move-vm-runtime/src/jit/execution/translate.rs
```diff
@@ -183,6 +222,7 @@ impl FunctionContext<'_, '_> {
 pub fn package(
     natives: &NativeFunctions,
     interner: &IdentifierInterner,
+    system_packages: &BTreeMap<OriginalId, Arc<CachedPackage>>,
     verified_package: input::Package,
 ) -> PartialVMResult<Package> {
```
cgswords, line 220: Assert that our current package ID is _not_ in the `system_packages`.

external-crates/move/crates/move-vm-runtime/src/cache/move_cache.rs
cgswords: Changed, with an added error path. We really should never need this in a RefCell or RwLock given usage patterns, and I would rather the error on incorrect usage than a lock letting it slide.

external-crates/move/crates/move-vm-runtime/src/shared/system_packages.rs
cgswords: Added a comment

external-crates/move/crates/move-vm-runtime/src/unit_tests/system_packages_tests.rs
cgswords: Right, but we can't really control what people do with linkage, so here we are

sui-execution/latest/sui-adapter/src/adapter.rs
cgswords: Removed, waiting for your upstream PR to enable this

tzakian (approved):
Looks good to me!

PR #26510, opened 2026-05-06
consensus_manager: switch address_overrides to parking_lot::Mutex

aschran:
## Summary

Lock type cleanup on `address_overrides`

## Test plan

CI


bmwill (approved):

akichidis (approved):

mwtian:
Why not use parking lot, which is more common in the codebase?

PR #26514, opened 2026-05-06
[consensus] use a common function to compute fetch blocks limit

mwtian:
## Description 

Share the response size limit logic between validator and observer.

## Test plan 

CI

### DIFF
--- a/consensus/core/src/network/tonic_network.rs
+++ b/consensus/core/src/network/tonic_network.rs
@@ -57,6 +57,26 @@ use crate::{
 // TODO: put max RPC response size in protocol config.
 pub(crate) const MAX_FETCH_RESPONSE_BYTES: usize = 4 * 1024 * 1024;
 
+pub(crate) fn max_fetch_blocks_response_bytes(
+    context: &Context,
+    block_refs: &[BlockRef],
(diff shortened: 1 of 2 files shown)


consensus/core/src/network/tonic_network.rs
```diff
@@ -57,6 +57,26 @@ use crate::{
...
+    context: &Context,
+    block_refs: &[BlockRef],
+    fetch_after_rounds: &[Round],
+) -> usize {
+    let max_response_num_blocks = if fetch_after_rounds.is_empty() {
+        block_refs
+            .len()
+            .min(context.parameters.max_blocks_per_fetch)
+    } else if block_refs.is_empty() {
+        context.parameters.max_blocks_per_fetch
```
akichidis: Looking on the endpoint we do those checks:

```
        let max_response_num_blocks = if !fetch_after_rounds.is_empty() && !block_refs.is_empty() {
            self.context.parameters.max_blocks_per_sync
        } else {
            self.context.parameters.max_blocks_per_fetch
        };
```

maybe here we could just follow the same assumptions just to simplify?

akichidis (approved):

PR #26519, opened 2026-05-06
[move] Cache-package path + external resolver env_name (DVX-2095, DVX-2097)

mdgeorge4153:
[CLAUDE]

## Summary

Two related improvements to `move-package-alt`:

- **DVX-2097**: `cache-package` now returns the on-disk path of the fetched package in `CachedPackageInfo.path`, so downstream tooling can locate it.
- **DVX-2095**: External resolvers now receive the environment **name** (in addition to its chain id) via a new `env_name` field on `ResolveRequest`. This is additive and backward-compatible — existing resolvers ignore the new field, and new ones can use it to call back into `cache-package` with the correct env. `&Environment` is threaded through `PinnedDependencyInfo::pin` and `ResolvedDependency::resolve` rather than passing `&EnvironmentID` alone.

Two commits, one per ticket.

## Test plan

- [x] Unit test pinning the JSON wire format of `ResolveRequest` (env, env_name, data)
- [x] Mock-resolver updated to deserialize `env_name`; all 9 existing `external_*` integration tests still pass (so the field is reaching the resolver process)
- [x] `package::tests::test_cache_package` updated to assert the new `path` matches the input local-dep path
- [x] `crates/sui/tests/shell_tests/cache_package/cache_package.snap` re-snapped for the new `path` field
- [x] `cargo move-clippy` + `cargo fmt -- --check` clean in `move-package-alt`
- [ ] CI passes

### DIFF
--- a/crates/sui/tests/shell_tests/cache_package/cache_package.sh
+++ b/crates/sui/tests/shell_tests/cache_package/cache_package.sh
@@ -1,5 +1,7 @@
 # Copyright (c) Mysten Labs, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-# Run the `cache-package` command
-sui move --client.config $CONFIG cache-package testnet 4c78adac "{ local="\"$(pwd -W 2>/dev/null || pwd)/a\"" }"
+# Run the `cache-package` command. Redact `path` because its value depends on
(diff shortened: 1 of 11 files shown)


manolisliolios (approved):
lgtm!

PR #26521, opened 2026-05-06
[sui cli] Add pass through command `sui move format` for `prettier-move`

tnowacki:
## Description 

- 🤖 Heavily vibe coded
- In an effort to increase awareness and usability of `prettier-move` formatter, this adds a passthrough command to `prettier-move`
  - If it is not in the path, it asks to install it (with the default being N)
  - If `npm` is missing, it informs the user and bails. 

## Test plan 

- Added tests
- Called the command locally 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [X] CLI: Adds `sui move format` passthrough command for `prettier-move` auto-formatter.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-move/src/format.rs
+++ b/crates/sui-move/src/format.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use anyhow::{Context, bail};
+use anyhow::{Context, bail, ensure};
 use clap::Parser;
 use std::{
     io::{self, BufRead, IsTerminal, Write},
@@ -11,15 +11,17 @@ use std::{
 /// Format Move source files using `prettier-move`.
 ///
 /// This is a thin alias around the `prettier-move` shim shipped by the
-/// `@mysten/prettier-plugin-move` npm package — every argument is forwarded
-/// verbatim. If `prettier-move` is not found on `PATH`, the command offers
-/// to install it via `npm i -g`.
+/// `@mysten/prettier-plugin-move` npm package — every argument (including
+/// `--help`) is forwarded verbatim. If `prettier-move` is not found on
+/// `PATH`, the command offers to install it via `npm i -g`.
 #[derive(Parser)]
+#[clap(disable_help_flag = true)]
 #[group(id = "sui-move-format")]
 pub struct Format {
     /// Arguments forwarded verbatim to `prettier-move`. Examples:
     ///   sui move format -c sources/foo.move      # check
     ///   sui move format -w .                     # write the package
+    ///   sui move format --help                   # prettier-move help
     #[clap(trailing_var_arg = true, allow_hyphen_values = true)]
     args: Vec<String>,
 }
@@ -28,7 +30,7 @@ const NPM_INSTALL_ARGS: &[&str] = &["i", "-g", "prettier", "@mysten/prettier-plu
 
 impl Format {
     pub async fn execute(self) -> anyhow::Result<()> {
-        if !is_on_path("prettier-move")? {
+        if which::which("prettier-move").is_err() {
             bootstrap_prettier_move()?;
         }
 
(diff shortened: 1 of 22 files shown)


crates/sui-move/src/format.rs
```diff
@@ -0,0 +1,120 @@
...
+        .status()
+    {
+        Ok(_) => Ok(true),
+        Err(e) if e.kind() == io::ErrorKind::NotFound => Ok(false),
+        Err(e) => Err(e).with_context(|| format!("failed to probe `{bin}`")),
+    }
+}
+
+fn bootstrap_prettier_move() -> anyhow::Result<()> {
+    if !is_on_path("npm")? {
```
mdgeorge4153: Is `prettier-move` available in suiup? Would that be a better choice here?

crates/sui-move/src/format.rs
```diff
@@ -0,0 +1,120 @@
...
+        bail!("`prettier-move` terminated by signal");
+    }
+}
+
+/// Probe `PATH` for `bin` by invoking `bin --version`. Returns `Ok(true)` if
+/// the binary ran (regardless of its exit status — `--version` may be
+/// rejected by some shims), `Ok(false)` if the binary was not found, and
+/// `Err` for other I/O errors.
+fn is_on_path(bin: &str) -> anyhow::Result<bool> {
+    match Command::new(bin)
```
mdgeorge4153: Running the binary seems like a slightly strange way to probe for its existence, but it's probably fine 🤷

crates/sui-move/src/format.rs
tnowacki: Good question, but don't know. @stefan-mysten?

stefan-mysten: `suiup` handles installation of binaries from GitHub release archives or Rust-based programs that can be compiled. To support installing it via `suiup`, we'd still need these npm checks.

I think it's better to have it like this for now; suiup is designed around GitHub releases, so not sure how we'd integrate this (maybe just a custom Rust code for prettier-move). But if you feel strongly about having it there and wish to open a PR in `suiup` @mdgeorge4153, I can have a look at it next week.

crates/sui-move/src/format.rs
amnn: There is a crate -- called [`which`](https://crates.io/crates/which) -- what does this for you, without running the binary, and in a cross-platform way. I would probably recommend using that?

crates/sui-move/src/format.rs
amnn: Agreed with @stefan-mysten here.

crates/sui-move/src/format.rs
```diff
@@ -0,0 +1,120 @@
...
+        .args(NPM_INSTALL_ARGS)
+        .stdin(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .status()
+        .context("failed to spawn `npm`")?;
+
+    if !status.success() {
+        bail!("`{install_cmd}` failed with status {status}");
+    }
```
amnn: `ensure!(status.success(), "...")`

crates/sui/tests/shell_tests/sui_move_format_tests/help/help.snap
amnn, line 1: Should `--help` also forward `--help` to `prettier-move`?

crates/sui/tests/shell_tests/sui_move_format_tests/missing_tool_no_npm/missing_tool_no_npm.snap
amnn, line 1: In this case, when they re-run, they would then be told how to install the package -- should we just give them both instructions in this case (Install `npm`, then install `prettier-move`)?

amnn (approved):
Looks good to me -- I'd recommend using `which` instead of the roll-your-own `is_on_path`, but the core change looks good, and I think it makes sense to just direct to `npm` for now?

Accepting to unblock, with comments.

crates/sui/tests/shell_tests/sui_move_format_tests/help/help.snap
tnowacki: I'll see if I can get that to work

PR #26541, opened 2026-05-07
[address balances] Invariant checks for transaction effects

tnowacki:
## Description 

- Adds invariant checks that are run after conservation checks
  - All net changes must be permissible under the input reservations (as applicable)
  - For any changes outside of the input reservations, they must come from the PTB runtime
  
## Test plan 

- Ran tests 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-adapter/src/temporary_store.rs
+++ b/sui-execution/latest/sui-adapter/src/temporary_store.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::gas_charger::{GasCharger, PaymentLocation};
-use mysten_common::ZipDebugEqIteratorExt;
+use mysten_common::{ZipDebugEqIteratorExt, debug_fatal};
 use mysten_metrics::monitored_scope;
 use parking_lot::RwLock;
 use std::collections::{BTreeMap, BTreeSet, HashSet};
@@ -85,6 +85,16 @@ pub struct TemporaryStore<'backing> {
     /// The set of per-epoch config objects that were loaded during execution, and are not in the
     /// input objects. This allows us to commit them to the effects.
     loaded_per_epoch_config_objects: RwLock<BTreeSet<ObjectID>>,
+
+    /// Index ranges into `execution_results.accumulator_events` for events emitted from PTB
+    /// (Move) execution. Each `record_execution_results` call appends a contiguous range
+    /// bracketing the merge. Any index outside these ranges was emitted by the runtime outside
+    /// of PTB execution (currently only `gas_charger`'s `add_accumulator_event`). Consumed by
+    /// `check_sui_address_balance_changes` inside `run_conservation_checks` to gate
+    /// non-PTB-emitted events behind input reservations. Cleared on `drop_writes` since the
+    /// underlying events are also cleared. A `Vec` is sufficient because real transactions run
+    /// the PTB at most a handful of times.
+    ptb_emitted_accumulator_event_ranges: Vec<std::ops::Range<usize>>,
 }
 
 impl<'backing> TemporaryStore<'backing> {
@@ -138,6 +148,7 @@ impl<'backing> TemporaryStore<'backing> {
             generated_runtime_ids: BTreeSet::new(),
             cur_epoch,
             loaded_per_epoch_config_objects: RwLock::new(BTreeSet::new()),
+            ptb_emitted_accumulator_event_ranges: Vec::new(),
         }
     }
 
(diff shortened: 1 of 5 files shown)


sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -1130,6 +1143,110 @@ impl TemporaryStore<'_> {
...
+        &self,
+        protocol_config: &ProtocolConfig,
+        input_reservations: &BTreeMap<(SuiAddress, TypeTag), u64>,
+    ) -> Result<(), ExecutionError> {
+        let result = self.check_address_balance_changes_impl(input_reservations);
+        if protocol_config.enforce_address_balance_change_invariant() {
+            result
+        } else {
+            if let Err(e) = result {
+                panic!("address-balance-change invariant violated pre-flag: {e}");
```
tzakian, line 1175: `panic!` or `debug_fatal!`? I can see reasons for both, but `panic!` is probably the better option.

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -1130,6 +1143,110 @@ impl TemporaryStore<'_> {
...
+    ) -> Result<(), ExecutionError> {
+        use sui_types::balance::Balance;
+
+        let mut actual_changes: BTreeMap<(SuiAddress, TypeTag), i128> = BTreeMap::new();
+        let mut ptb_changes: BTreeMap<(SuiAddress, TypeTag), i128> = BTreeMap::new();
+        for (idx, event) in self.execution_results.accumulator_events.iter().enumerate() {
+            // Only Balance<T> accumulators are supported today; extend this filter when
+            // additional funds-accumulator types are introduced.
+            if !Balance::is_balance_type(&event.write.address.ty) {
+                continue;
```
tzakian, line 1206: Maybe a `debug_fatal!` log here or debug_assert to break this early if we add a new type of accumulator event and forget it here?

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -85,6 +85,16 @@ pub struct TemporaryStore<'backing> {
...
+
+    /// Index ranges into `execution_results.accumulator_events` for events emitted from PTB
+    /// (Move) execution. Each `record_execution_results` call appends a contiguous range
+    /// bracketing the merge. Any index outside these ranges was emitted by the runtime outside
+    /// of PTB execution (currently only `gas_charger`'s `add_accumulator_event`). Consumed by
+    /// `check_sui_address_balance_changes` inside `run_conservation_checks` to gate
+    /// non-PTB-emitted events behind input reservations. Cleared on `drop_writes` since the
+    /// underlying events are also cleared. A `Vec` is sufficient because real transactions run
+    /// the PTB at most a handful of times.
+    ptb_emitted_accumulator_event_ranges: Vec<std::ops::Range<usize>>,
```
tzakian, line 97: Totally fine as-is, but you nerd-sniped me into a fun algo question: an alternative to this would be a `BTreeMap<usize, bool>` (or similar) to represent a range query with `true = PTB-emitted` `false = runtime-emitted`. But honestly, with the number of events I bet the vector will be faster.

tzakian (approved):
couple small little nits/improvement suggestions but otherwise LGTM and thank you for adding this!

sui-execution/latest/sui-adapter/src/temporary_store.rs
tnowacki: How is that different than just a `BTreeSet<usize>`? Or do you not look up in the map by key but rather by range?

I had set originally but switched to this since
- For all user transactions, this will be of length 1
- For (most?) system transactions, I think it will always be length 1 since we usually run PTBs one after another so they will coalesce into just the one entry

sui-execution/latest/sui-adapter/src/temporary_store.rs
tnowacki: I want a panic here since if this trips with the protocol config rolling out... we got a problem

sui-execution/latest/sui-adapter/src/temporary_store.rs
tzakian: You'd be looking up by range, but unless this vector gets "large" (i.e., > ~256 elems probably) a vector will be faster, so keeping it like this is probably better and the right call.

PR #26545, opened 2026-05-07
[move][move-vm] Improve telemetry speed by swapping to thread-local memory cells and unsafe writes.

cgswords:
## Description 

The title says it all; this drops the Atomic approach for something significantly faster. There are some invariants that must be maintained (basically, any individual thread should only be running a single transaction in the compiler at the same time), but that invariant is literally the deployment plan today.

## Test plan 

Tests still pass and TPS tests demonstrate reduced overhead from 18% of runtime down to 0.6% of runtime.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/runtime/telemetry.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/runtime/telemetry.rs
@@ -274,9 +274,7 @@ impl TelemetryContext {
         let local = self.local();
         macro_rules! add {
             ($field:ident, $delta:expr) => {{
-                local
-                    .$field
-                    .set(local.$field.get().saturating_add($delta));
+                local.$field.set(local.$field.get().saturating_add($delta));
             }};
         }
         macro_rules! update_duration_field {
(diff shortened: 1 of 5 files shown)


external-crates/move/crates/move-vm-runtime/src/runtime/telemetry.rs
```diff
@@ -193,24 +236,29 @@ pub(crate) struct Timer {
...
+    fn local(&self) -> &'static ThreadCounters {
+        match LOCAL_REF.with(|c| c.get()) {
+            Some(tc) if LOCAL_ID.with(|c| c.get()) == self.id => tc,
+            _ => self.register_local(),
+        }
+    }
+
+    #[cold]
+    fn register_local(&self) -> &'static ThreadCounters {
+        let leaked: &'static ThreadCounters = Box::leak(Box::new(ThreadCounters::new()));
```
tzakian, line 255: `Box::leak` makes me worried here. Could this instead return a cheap clone with interior mutability? (e.g., something like an `Rc<RefCell<ThreadCounters>>`?

Otherwise I think this is fine with our threading model as this will causing some leaking per-thread but since our threads are persistent due to tokio it should be pretty small overhead. But, I'd love if @mlogan could double-check to make sure this won't cause issues. 

Another option is we could do the whole package-cache thing and have arenas that we control @cgswords ;)

tzakian (dismissed):
Overall looks good to me. Main concern is the `Box::leak` and seeing if there's some way of not doing that...

PR #26546, opened 2026-05-08
Share transaction deny configs between validators via consensus

aschran:
## Summary

Lets validator operators share `TransactionDenyConfig` rules via consensus. Proposed rules activate once validators holding enough stake agree. Thresholds for activation are fully, individually controlled by each operator.

## How it works

The operator pre-defines, in `NodeConfig`, named **rulesets** plus optional **default** buckets. Each is gated on:

- eligible validator set — an allowlist or denylist of validators whose votes count, and
- stake threshold — required percent of eligible stake.

A validator's broadcast is a `TransactionDenyRules` proposal. On each receiver, the effective config used at signing time is computed as:

- the operator's local `transaction_deny_config`, always applied; plus
- every pre-listed ruleset that that has met its stake threshold; plus
- every individual rule element (the "default" buckets) that has met its stake threshold

A single proposal can satisfy several overlapping or nested rulesets at once.

## What's in the change

- Consensus message `UpdateTransactionDenyConfig` carrying `{ authority, generation, Option<TransactionDenyRules> }`, gated by the `share_transaction_deny_config_in_consensus` protocol flag. `None` withdraws a prior proposal.
- `TransactionDenyRules` — the shareable rule set.
- `PeerDenySyncConfig` in `NodeConfig`: `rulesets` (each a `SharedDenyRuleset`: name, rules, `ValidatorEligibility`, `stake_threshold_percent`), `default_threshold`, `broadcast_on_startup`, `broadcast_on_epoch_change`.
- `TransactionDenyConfigManager` — implements the logic of tracking votes and computing what config to apply.
- Admin endpoints: `POST /broadcast-transaction-deny-config`, `POST /withdraw-transaction-deny-config`, `GET /transaction-deny-config` (operator dump including per-ruleset voting status).
- Per-ruleset Prometheus metrics (active flag, voted stake, eligible stake).

## Test plan

- [x] `cargo nextest run` — unit tests for `TransactionDenyRules`, `PeerDenySyncConfig` (validation + YAML round-trip), the `evaluate_deny_configs` evaluator, and `TransactionDenyConfigManager`
- [x] `cargo simtest -p sui-e2e-tests --test transaction_deny_share_tests` — 8 end-to-end tests on a 4-validator cluster with real consensus messages: threshold activation through real transaction execution, withdrawal, superset / overlapping rulesets, allowlist & denylist eligibility, the default per-element bucket, self-broadcast loopback, and stale-generation rejection
- [x] `cargo xclippy` (sui-core, sui-node, sui-config) and `cargo fmt --all`
- [x] `sui-swarm-config` / `sui-open-rpc` snapshot tests pass


crates/sui-types/src/transaction_deny_rules.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use serde::{Deserialize, Serialize};
+
+use crate::base_types::{ObjectID, SuiAddress};
+
+/// The shareable subset of `TransactionDenyConfig`, rules for transactions that
+/// the node will refuse.
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct TransactionDenyRules {
```
alex-mysten, line 14: This also needs versioning, no?

crates/sui-node/src/admin.rs
```diff
@@ -655,3 +667,109 @@ async fn update_endpoint(
...
+async fn broadcast_transaction_deny_config(
+    State(state): State<Arc<AppState>>,
+) -> (StatusCode, String) {
+    let local_rules = state
+        .node
+        .state()
+        .local_transaction_deny_config()
+        .rules()
+        .clone();
+    submit_transaction_deny_config_update(&state, Some(local_rules)).await
```
alex-mysten, line 731: should we broadcast None when we have an empty ruleset as an explicit clear?

crates/sui-node/src/lib.rs
```diff
@@ -1729,6 +1745,38 @@ impl SuiNode {
...
+                    .unwrap_or(sync_cfg.broadcast_on_epoch_change);
+                if should_broadcast
+                    && cur_epoch_store
+                        .protocol_config()
+                        .share_transaction_deny_config_in_consensus()
+                {
+                    let manager = self.state.transaction_deny_config_manager();
+                    let local_rules = manager.local().rules().clone();
+                    let payload = (!local_rules.is_empty()).then_some(local_rules);
+                    match manager.build_share_consensus_tx(self.state.name, payload) {
```
alex-mysten: should this check the share limit too?

crates/sui-types/src/transaction_deny_rules.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub package_upgrade_disabled: bool,
+
+    /// Whether usage of shared objects is disabled.
+    #[serde(default)]
+    pub shared_object_disabled: bool,
+
+    /// Whether user transactions are disabled (i.e. only system transactions are allowed).
+    /// This is essentially a kill switch for transactions processing to a degree.
+    #[serde(default)]
+    pub user_transaction_disabled: bool,
```
alex-mysten, line 51: maybe we should exclude this one?

crates/sui-core/src/authority.rs
```diff
@@ -962,6 +963,11 @@ pub struct AuthorityState {
     /// Limits the number of concurrent post-processing tasks to avoid overwhelming
     /// the blocking thread pool. Defaults to the number of available CPUs.
     post_processing_semaphore: Arc<tokio::sync::Semaphore>,
+
+    /// Created once per process, then re-attached to each new `AuthorityPerEpochStore`
+    /// at reconfiguration.
+    transaction_deny_config_manager:
+        Arc<crate::transaction_deny_config_manager::TransactionDenyConfigManager>,
```
akichidis: nit: maybe just import the `crate::transaction_deny_config_manager`

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2371,6 +2382,20 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
         }
     }
 
+    fn process_shared_transaction_deny_configs(
+        &self,
+        shared_transaction_deny_configs: Vec<SharedTransactionDenyConfig>,
+    ) {
+        if shared_transaction_deny_configs.is_empty() {
+            return;
+        }
```
akichidis: Probably that's redundant as the loop will not iterate anyways?

crates/sui-core/src/transaction_deny_config_manager.rs
```diff
@@ -0,0 +1,693 @@
...
+        let now_ms = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .expect("system time before unix epoch")
+            .as_millis() as u64;
+        let last = self
+            .perpetual
+            .last_broadcast_deny_generation
+            .get(&())?
+            .unwrap_or(0);
+        let generation = now_ms.max(last + 1);
```
akichidis: Do we leverage somehow the timestamp information? Otherwise could that be a simple monotonically increment counter starting from zero?

crates/sui-types/src/messages_consensus.rs
```diff
@@ -305,8 +315,50 @@ impl Debug for ConsensusTransactionKey {
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Hash, Debug, PartialEq, Eq)]
+pub struct SharedTransactionDenyConfigV1 {
+    /// Originating authority — must match consensus transaction source.
+    pub authority: AuthorityName,
+    /// Generation: ms since epoch (matches `CapabilityNotificationV2`). The receiver
+    /// only accepts updates with strictly greater generation than the last accepted
+    /// for that authority.
+    pub generation: u64,
```
akichidis, line 362: naming is not my strong suit, but `generation` doesn't immediately convey the intention to me. Maybe just `id` could do?

crates/sui-types/src/transaction_deny_rules.rs
aschran: it's already contained in another versioned enum which is itself contained in a versioned enum, so a third level of versioning just seemed gratuitous

crates/sui-node/src/admin.rs
aschran: that's what the withdraw route is for

crates/sui-core/src/transaction_deny_config_manager.rs
aschran: I figured it could be useful for debugging to make these timestamps

crates/sui-types/src/messages_consensus.rs
aschran: we already use the term `generation` for this in multiple other consensus messages, so it's consistent

crates/sui-types/src/transaction_deny_rules.rs
aschran: idk we should talk about it, I could see the case for either

crates/sui-config/src/transaction_deny_config.rs
```diff
@@ -28,6 +35,91 @@ pub struct PeerDenySyncConfig {
...
+    pub threshold: SharedDenyRuleThreshold,
+}
+
+/// Eligibility + stake threshold criteria for activating a proposed deny-rule.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub struct SharedDenyRuleThreshold {
+    pub eligibility: ValidatorEligibility,
+    /// Whole-number percent (0..=100) of eligible stake that must vote to activate.
+    pub stake_threshold_percent: u16,
```
akichidis, line 58: nit: since this is between `0..=100` I would suggest just naming it `stake_threshold` .

crates/sui-core/src/transaction_deny_config_manager.rs
```diff
@@ -91,124 +123,148 @@ impl TransactionDenyConfigManager {
...
-                debug!(
+            let mut accepted = false;
+            for msg in msgs {
+                let authority = msg.authority();
+                let generation = msg.generation();
+                if !committee.authority_exists(&authority) {
+                    debug!(
+                        authority = %authority.concise(),
+                        "Dropping UpdateTransactionDenyConfig: sender not in committee",
+                    );
```
akichidis: Shall this be a `warn` instead? Although I would expect it to appear on epoch change edge cases, every other case would signify something more critical

crates/sui-core/src/transaction_deny_config_manager.rs
```diff
@@ -250,27 +308,199 @@ impl TransactionDenyConfigManager {
...
+    pub prelisted: Vec<PrelistedRulesetStatus>,
+    pub default: Option<DefaultBucketStatus>,
+}
+
+/// Returns true if `voted` is at least `percent`% of `eligible` stake.
+fn meets_threshold(voted: StakeUnit, eligible: StakeUnit, percent: u16) -> bool {
+    if eligible == 0 {
+        // Zero eligible stake never meets a threshold.
+        return false;
+    }
```
akichidis, line 467: Technically though  the `percent` parameter by config can be `zero` right?

crates/sui-core/src/transaction_deny_config_manager.rs
```diff
@@ -91,124 +123,148 @@ impl TransactionDenyConfigManager {
...
 
-    /// Drop entries (in-memory + DB) for any authority that is no longer in the active
-    /// committee. Called at epoch transitions and at startup.
-    pub fn prune_for_committee(&self, committee: &Committee) -> SuiResult<()> {
-        let (new_effective, peers_snapshot, removed) = {
+    /// Update the stored committee and prune proposals (in-memory + DB) from any
+    /// authority that is no longer a member. Always recomputes the effective config,
+    /// since the stake distribution may have shifted even with no departures. Called at
+    /// epoch transitions and at startup.
+    pub fn update_for_committee(&self, committee: Arc<Committee>) -> SuiResult<()> {
```
akichidis, line 283: Do we care about `apply_updates` and `update_for_committee` race conditions during epoch change? I guess chances are super slim, but maybe it would worth guaranteeing that intended changes happen for a specified epoch?

crates/sui-node/src/lib.rs
```diff
@@ -2616,6 +2624,33 @@ fn max_tx_per_checkpoint(_: &ProtocolConfig) -> usize {
...
+///
+/// An oversized local config can't go on the wire (the consensus validator would reject
+/// it), so it is treated as unshareable: on startup a stale prior vote is withdrawn
+/// rather than left to diverge from the local config.
+fn deny_config_reconciliation_payload(
+    local_rules: &TransactionDenyRules,
+    broadcast: bool,
+    is_startup: bool,
+    has_prior_broadcast: bool,
+) -> Option<Option<TransactionDenyRules>> {
```
akichidis: I would suggest using an `enum` for this. I know that is fairly local the scope, but it would probably be easier to read the cases

crates/sui-types/src/transaction_deny_rules.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    /// A list of disabled OAuth providers for zkLogin.
+    #[serde(default)]
+    pub zklogin_disabled_providers: BTreeSet<String>,
+}
+
+impl TransactionDenyRules {
+    /// Maximum total entry count across all set fields permitted in a single
+    /// peer-broadcast deny rules message. Bounds wire and DB cost.
+    pub const MAX_SHARE_ENTRIES: usize = 10_000;
```
akichidis: Well I guess that's a limit that we might not necessarily hit in prod, but it might worth confirming it does not hit the consensus block size limits `max_transactions_in_block_bytes` (currently set to 512KB) , since those deny rules will eventually get shared via consensus , right?

crates/sui-types/src/transaction_deny_rules.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// `other` is a subset of `self`'s, and every boolean set in `other` is set in
+    /// `self`. A peer's proposal "votes for" a pre-listed config only when it is a
+    /// superset of that config's rules.
+    pub fn is_superset_of(&self, other: &Self) -> bool {
+        other.object_deny_list.is_subset(&self.object_deny_list)
+            && other.package_deny_list.is_subset(&self.package_deny_list)
+            && other.address_deny_list.is_subset(&self.address_deny_list)
+            && other
+                .zklogin_disabled_providers
+                .is_subset(&self.zklogin_disabled_providers)
```
akichidis: No errors here, but curious why not using the `BTeeSet::is_superset` which (potentially) matches better the intuition of the method `is_superset_of` ? Ex
```
self.object_deny_list.is_superset(&other.object_deny_list)
```

crates/sui-config/src/transaction_deny_config.rs
```diff
@@ -1,150 +1,210 @@
...
+    pub broadcast_on_startup: bool,
 
-    /// Whether upgrading existing packages is disabled.
     #[serde(default)]
-    package_upgrade_disabled: bool,
-
-    /// Whether usage of shared objects is disabled.
-    #[serde(default)]
-    shared_object_disabled: bool,
+    pub broadcast_on_epoch_change: bool,
```
akichidis, line 36: So this will guarantee that the rules will propagate to next epoch as well right? (especially for newly joined nodes on epoch N+1)

aschran: yes it will

crates/sui-types/src/transaction_deny_rules.rs
aschran: good idea

crates/sui-types/src/transaction_deny_rules.rs
aschran: good flag, reduced the MAX_SHARE_ENTRIES to a lower value which should keep things under the consensus size limit, if it doesn't then consensus will enforce that limit so we should be safe either way

crates/sui-config/src/transaction_deny_config.rs
aschran: percent in the name documents the unit, since it's not documented in the type (otherwise you might erroneously think it's e.g. stake units)

crates/sui-core/src/transaction_deny_config_manager.rs
aschran: I think warn should be reserved for known-problematic conditions, can at least bump to an info though

crates/sui-core/src/transaction_deny_config_manager.rs
aschran: good point, changed it to minimum of 1

crates/sui-core/src/transaction_deny_config_manager.rs
aschran: good point, will fix

crates/sui-node/src/lib.rs
aschran: done

crates/sui-config/src/transaction_deny_config.rs
```diff
@@ -1,150 +1,252 @@
...
+use sui_types::base_types::{AuthorityName, ObjectID, SuiAddress};
+pub use sui_types::transaction_deny_rules::{DenyElementKind, TransactionDenyRules};
 
 use crate::dynamic_transaction_signing_checks::{
     DynamicCheckRunnerContext, DynamicCheckRunnerError,
 };
 
-#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+/// Configuration for activating recommended `TransactionDenyConfig` rules shared by
+/// peers via consensus. The operator pre-defines named rulesets, each gated on a
```
alex-mysten, line 15: How do we keep rulesets up to date between proposers and recievers? It seems that without careful coordination we might fail to activate deny configs, or not properly cover what we need to cover

crates/sui-config/src/transaction_deny_config.rs
```diff
@@ -1,150 +1,252 @@
...
-    #[serde(skip)]
-    object_deny_set: OnceCell<HashSet<ObjectID>>,
+/// Which validators' proposals count toward a deny-rule activation's stake threshold.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub enum ValidatorEligibility {
+    /// Only the listed authorities are eligible.
+    Allowlist(BTreeSet<AuthorityName>),
+    /// All committee members except the listed authorities are eligible.
+    Denylist(BTreeSet<AuthorityName>),
```
alex-mysten, line 83: Do we need a deny list?

crates/sui-config/src/transaction_deny_config.rs
```diff
@@ -1,150 +1,252 @@
...
+    /// `DenyElementKind` and threshold-gates each proposed element of those kinds
+    /// individually.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    address_deny_list: Vec<SuiAddress>,
+    pub default_buckets: Vec<DefaultDenyBucket>,
 
-    /// Whether publishing new packages is disabled.
     #[serde(default)]
-    package_publish_disabled: bool,
+    pub broadcast_on_startup: bool,
```
alex-mysten, line 33: What happens if we want to activate a deny config when the network is still recovering?

crates/sui-node/src/admin.rs
```diff
@@ -662,3 +674,141 @@ async fn update_endpoint(
...
+    }
+}
+
+async fn broadcast_transaction_deny_config(
+    State(state): State<Arc<AppState>>,
+) -> (StatusCode, String) {
+    let local_rules = state
+        .node
+        .state()
+        .local_transaction_deny_config()
```
alex-mysten, line 728: Why have the API? isn't broadcast-on-startup sufficient, since we need to restart the node anyway?

crates/sui-config/src/transaction_deny_config.rs
aschran: It's intentional here that rulesets are fully under local control, to preserve individual validator autonomy. Proposed deny configs are applied to any rule with a matching subset. Call it a sort of "duck typing" for preconfigured rulesets.

crates/sui-config/src/transaction_deny_config.rs
aschran: At the very least, empty denylist is currently how one would indicate "all validators"

crates/sui-node/src/admin.rs
aschran: Just figure a "send it again" button is never a bad idea to have in your emergency toolkit

crates/sui-config/src/transaction_deny_config.rs
aschran: Can you say more? What would be the concern? Just that the sharing depends on consensus at all, or are you referring to something more narrow?

crates/sui-config/src/transaction_deny_config.rs
alex-mysten: Makes sense - i'm just concerned about the possibility of a misfire - wondering if there's any way we can maintain visibility into how much stake has which rulesets to make sure this works when we need it

crates/sui-config/src/transaction_deny_config.rs
alex-mysten: Meaning that any validator in the committee would be trusted to broadcast deny configs? Is that something we would realistically want anyone to enable?

aschran: With a threshold I could see us using that. for example if at least 20% of stake recommends denying X, then maybe I want to go along with it.

crates/sui-config/src/transaction_deny_config.rs
aschran: we have the `tx_deny_default_bucket_applied_elements` and `tx_deny_ruleset_active` metrics, which would report the names of any preconfigured rulesets, though not their full contents (which are probably too big to put inside a prometheus label). presumably that's sufficient for us to determine whether suggested rulesets have been configured on enough validators, unless some validators deliberately mislead us by configuring empty rulesets with the same name, or something? (I don't see an easy way to defend against that; ultimately metrics reporting is all honor-system)

crates/sui-core/src/authority/authority_store_tables.rs
```diff
@@ -161,6 +162,16 @@ pub struct AuthorityPerpetualTables {
...
+
+    /// Most recent `UpdateTransactionDenyConfig` consensus message accepted from each
+    /// allowlisted peer authority. Persisted across epochs so we can rebuild the
+    /// merged effective deny config after a node restart. Pruned at epoch transitions
+    /// when a peer is no longer in the active committee.
+    pub(crate) shared_transaction_deny_configs: DBMap<AuthorityName, SharedTransactionDenyConfig>,
+
+    /// Singleton: the highest generation this node has ever broadcast in an
+    /// `UpdateTransactionDenyConfig` message.
+    pub(crate) last_broadcast_deny_generation: DBMap<(), u64>,
```
lxfind, line 178: How does this work when a validator used to broadcast deny configs but then restored from a formal snapshot?

crates/sui-config/src/transaction_deny_config.rs
alex-mysten: Just for my own understanding - the message would be queued until we recover, and then be activated as one of the first consensus commits where we have quorum?

aschran: It just sends it through consensus adapter like any other message, so yeah I assume the consensus codebase would queue any accepted messages that are pending submission until they can be sequenced

crates/sui-core/src/authority/authority_store_tables.rs
aschran: I guess in that case a validator could accidentally equivocate, if they later broadcast a new shared deny config with a lower or same generation number. If it did, then other validators would keep the submission with the latest generation. However I don't expect this to be a big problem in practice, bc we use timestamps for generation numbers, so it's quite unlikely to accidentally equivocate even if you did forget your last broadcast. I could probably have omitted this altogether, since even if accidental equivocation did somehow occur, nothing terrible really happens. but it felt more "correct" to include it.

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2120,6 +2131,18 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
         }
     }
 
+    fn process_shared_transaction_deny_configs(
+        &self,
+        shared_transaction_deny_configs: Vec<SharedTransactionDenyConfig>,
+    ) {
+        if let Err(e) = self
+            .transaction_deny_config_manager
+            .apply_updates(shared_transaction_deny_configs)
```
mystenmark: i'd suggest moving this to consensus_validator, we don't need determinism for this, and it will reduce latency

crates/sui-core/src/transaction_deny_config_manager.rs
```diff
@@ -0,0 +1,1483 @@
...
+            )
+        };
+        self.apply_evaluation(evaluation, active_proposals);
+        Ok(())
+    }
+
+    /// Allocate the next monotonic generation for an outgoing broadcast. Persists the
+    /// returned value before any submission so a crash between allocate-and-send cannot
+    /// reuse the generation.
+    pub fn allocate_next_broadcast_generation(&self) -> SuiResult<u64> {
```
mystenmark, line 324: is it worth a mutex here to ensure we can't have a race if two threads call this?

crates/sui-core/src/consensus_handler.rs
aschran: That seems kind of confusing and abstraction-breaking, no? do we have any other non-validation/business logic behavior that happens in the validator?

Given that we expect maybe a few of these messages per year (or hopefully none), is the latency impact relevant? I could spawn it in a separate task here if so?

crates/sui-core/src/transaction_deny_config_manager.rs
aschran: good point. it was questionable whether the generation-tracking was even worth bothering to begin with (what are the chances we try to send two in the same millisecond), but if I'm going to do it I should do it right.

crates/sui-core/src/consensus_handler.rs
mystenmark: Yes, we do: we forward CheckpointSignatureV2 messages to the signature aggregator to reduce latency

aschran: done

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2994,6 +2995,10 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
                                 .checkpoint_signature_messages
                                 .push(*checkpoint_signature_message);
                         }
+                        ConsensusTransactionKind::UpdateTransactionDenyConfig(_) => {
+                            // Applied at block verification time by SuiTxValidator;
+                            // nothing to do at commit processing.
```
mystenmark: i should have called this out earlier, but validation does not always happen - if the validator is catching up it will simply execute commits as soon as the commit rule allows without validating every block, so we need to also process updates here

aschran: thanks, fixed

mystenmark (approved):

PR #26552, opened 2026-05-08
[move] On-chain dependency pinning (v2: separate types)

mdgeorge4153:
## Summary

- Alternative approach to #26233 for on-chain dependency pinning
- Uses two separate types instead of one: `OnChainFlag` (`on-chain = true` in `[dependencies]`) and `OnChainAddress` (`on-chain = "0x..."` in `[dep-replacements]` and lockfiles)
- Compile-time guarantees that each context uses the correct form
- Simpler `from_lockfile` (no `Option` unwrapping) and better serde error messages

**Compare with:** #26233 (v1 uses a single `OnChainDepInfo` type with `address_override` on `Pin`)

## Test plan

- [x] All 173 existing tests pass
- [x] New parsing tests for `on-chain = true`, `on-chain = "0x..."`, `on-chain = false`, `on-chain = 42`
- [x] Lockfile round-trip test for on-chain deps with address
- [x] `cargo move-clippy` clean
- [x] `cargo fmt` clean


external-crates/move/crates/move-package-alt/src/compatibility/legacy_parser.rs
```diff
@@ -605,7 +605,7 @@ fn parse_dependency(mut tval: TV, mode: Option<&str>) -> Result<DefaultDependenc
             };
 
             // TODO: Implement once we have the on-chain deps design.
-            ManifestDependencyInfo::OnChain(OnChainDepInfo {
+            ManifestDependencyInfo::OnChain(OnChainFlag {
```
mdgeorge4153: I don't understand why legacy has any onchain code in it - legacy packages don't support on-chain deps.

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
```diff
@@ -985,6 +997,98 @@ mod tests {
...
+            foo = { on-chain = true }
+            "#,
+        )
+        .unwrap();
+
+        assert!(matches!(
+            manifest.get_dep("foo").dependency_info,
+            ManifestDependencyInfo::OnChain(_)
+        ));
+    }
```
mdgeorge4153, line 1021: The more important thing to catch is the converse: you can't have `on-chain = true` in the `dep-replacements` section, and you can't have `on-chain = "0x1234"` in the `dependencies` section.

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
```diff
@@ -193,8 +195,18 @@ impl<'de> Deserialize<'de> for ManifestDependencyInfo {
...
+                let val = &tbl["on-chain"];
+                if val.is_bool() {
+                    let dep = OnChainFlag::deserialize(data).map_err(de::Error::custom)?;
+                    Ok(ManifestDependencyInfo::OnChain(dep))
+                } else if val.is_str() {
+                    let dep = OnChainAddress::deserialize(data).map_err(de::Error::custom)?;
+                    Ok(ManifestDependencyInfo::OnChainAt(dep))
+                } else {
+                    Err(de::Error::custom(
+                        "on-chain must be `true` or a hex address string like \"0x...\"",
```
mdgeorge4153: we could possibly add more detail here - that it must be `true` in the dependency section or a hex string in the dep-replacements section.

external-crates/move/crates/move-package-alt/src/schema/lockfile.rs
```diff
@@ -201,4 +201,59 @@ mod tests {
         debug!("Rendered lockfile:\n---\n{rendered}\n---");
         assert_eq!(rendered, lockfile);
     }
+
+    /// OnChainAddress deserializes from a string value
+    #[test]
+    fn parse_on_chain_address() {
```
mdgeorge4153: Any reason not to fold these into the big example parse/print test above?

mdgeorge4153: I normally like to keep tests separate, but in this case one big test seems maybe simpler?

external-crates/move/crates/move-package-alt/src/package/manifest.rs
```diff
@@ -63,6 +63,18 @@ pub enum ManifestErrorKind {
...
+
+    #[error(
+        "On-chain dependency `{name}` in `[dependencies]` must use `on-chain = true`. \
+         Specify the address in `[dep-replacements]` with `on-chain = \"0x...\"`."
+    )]
+    OnChainDepWithAddress { name: PackageName },
+
+    #[error(
+        "On-chain dependency `{name}` in `[dep-replacements]` must specify an address: \
+         `on-chain = \"0x...\"`. Use `on-chain = true` in `[dependencies]` instead."
```
mdgeorge4153: I find this message a little confusing; I might just drop the second sentence (IIRC it should be fine if they don't actually include the dependency at all in the `[dependencies]` section.

external-crates/move/crates/move-package-alt/src/errors/mod.rs
```diff
@@ -130,6 +130,12 @@ pub enum PackageError {
         "Packages with old-style Move.toml files cannot depend on new-style packages. See https://docs.sui.io/references/package-managers/package-manager-migration for instructions."
     )]
     LegacyDependsOnModern,
+
+    #[error(
+        "On-chain dependency `{name}` requires `published-at` and `original-id` addresses. Add a \
```
mdgeorge4153: This is not accurate in this version. Instead, the dependency needs an `on-chain` field.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -144,9 +158,16 @@ impl CombinedDependency {
         default: DefaultDependency,
         replacement: ReplacementDependency,
     ) -> ManifestResult<Self> {
+        let has_replacement_dep = replacement.dependency.is_some();
         let dep = replacement.dependency.unwrap_or(default);
 
-        // TODO: possibly additional compatibility checks here?
```
mdgeorge4153, line 149: why are you dropping the TODO?

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -144,9 +158,16 @@ impl CombinedDependency {
...
+        // If the replacement explicitly provides an on-chain dep, it must have an address
+        if has_replacement_dep && let ManifestDependencyInfo::OnChain(_) = &dep.dependency_info {
```
mdgeorge4153: It seems like this test should be done when we actually construct the Pinned dependency, which must have the address. It feels premature here.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -48,12 +48,18 @@ pub struct PinnedDependency {
...
+    /// Invariants:
+    /// - On-chain packages never have local transitive dependencies, since their manifests
+    ///   are system-generated.
+    OnChain {
+        published_at: PublishedID,
+    },
     Root(PackagePath),
 }
 
-/// Invariant: if a PinnedDependency has `dep_info` `Root`, then its `containing_file` is either a
```
mdgeorge4153, line 55: IIRC this comment was removed because it was in the wrong place, but I don't see it being readded anywhere.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -93,9 +99,17 @@ impl PinnedDependency {
         // pinning - fix git shas and normalize local deps
         for dep in deps.into_iter() {
             let transformed = match dep.dep_info {
-                Resolved::Local(ref loc) => loc.clone().pin(parent, environment_id)?,
+                Resolved::Local(ref loc) => loc.clone().pin(parent)?,
                 Resolved::Git(ref git) => git.pin().await?,
-                Resolved::OnChain(_) => todo!(),
+                Resolved::OnChain(_) => {
```
mdgeorge4153: I'm not sure we want a `Resolved::OnChain` at all; I don't think external resolvers should return `{on-chain = true}`.

mdgeorge4153: I think there's also a description on the properties that ResolvedDependencies have somewhere (doc comment on `ResolvedDependency` type)? We should update that to include that all on-chain deps have addresses.

mdgeorge4153: In fact, I wonder if this should be done at the combined stage instead of here. I mentioned above that the check felt premature, but maybe the problem is that `Combined` shouldn't have a non-addressed variant?

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -204,25 +219,30 @@ impl Pinned {
     }
 
     /// Return the absolute path to the directory that this package would be fetched into, without
-    /// actually fetching it. `_chain_id` is unused for now but will be needed for on-chain deps.
-    pub(crate) fn unfetched_path(&self, _chain_id: &EnvironmentID) -> PathBuf {
+    /// actually fetching it. `chain_id` is used to determine the cache location for on-chain
+    /// dependencies.
+    pub(crate) fn unfetched_path(&self, chain_id: &EnvironmentID) -> PathBuf {
```
mdgeorge4153, line 219: I wonder if we don't actually need to include the chain ID in the path. I believe object IDs are cryptographic hashes, so collisions even across networks might be extermely unlikely. There's also the issue of forked networks, so maybe we should keep it for now

mdgeorge4153: I've asked the team if there's any danger of overlap.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -48,12 +48,18 @@ pub struct PinnedDependency {
...
     Local(PinnedLocalDependency),
     Git(PinnedGitDependency),
-    OnChain(OnChainDepInfo),
+    /// An on-chain dependency, identified by the address it is published at.
+    ///
+    /// Invariants:
+    /// - On-chain packages never have local transitive dependencies, since their manifests
+    ///   are system-generated.
+    OnChain {
+        published_at: PublishedID,
```
mdgeorge4153: I wonder if a different name would be better, given that published_at is a different field. Maybe published_address or just address (with the fact that it's published coming from the type)

external-crates/move/crates/move-package-alt/src/schema/resolver.rs
```diff
@@ -18,7 +18,8 @@ pub struct ResolverName(String);
 pub enum ResolverDependencyInfo {
     Local(LocalDepInfo),
     Git(ManifestGitDependency),
-    OnChain(OnChainDepInfo),
+    OnChain(OnChainFlag),
```
mdgeorge4153: As mentioned above, I don't think we want the `OnChain` variant, only the `OnChainAt` variant (which maybe we should just call `OnChain`, WDYT?

external-crates/move/crates/move-package-alt/src/compatibility/legacy_parser.rs
mdgeorge4153: [CLAUDE] Agreed — removed. This was carried over from the v1 cherry-pick.

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
mdgeorge4153: [CLAUDE] Done — `from_default` now rejects both `OnChainAt` (address in deps) and `OnChain` (flag with no replacement). `from_replacement` rejects `OnChain` (flag in dep-replacements).

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
mdgeorge4153: [CLAUDE] Updated to mention both contexts: "on-chain must be `true` (in [dependencies]) or a hex address string like \"0x...\" (in [dep-replacements])"

external-crates/move/crates/move-package-alt/src/schema/lockfile.rs
mdgeorge4153: [CLAUDE] Done — folded the on-chain entry into the main `parse_then_render` test.

external-crates/move/crates/move-package-alt/src/package/manifest.rs
mdgeorge4153: [CLAUDE] Dropped the second sentence.

external-crates/move/crates/move-package-alt/src/errors/mod.rs
mdgeorge4153: [CLAUDE] Fixed — now says "needs an address" and points to \`on-chain = "0x..."\` in dep-replacements.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Restored the TODO.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Moved the validation to `from_default` (combine time). `from_default` is called when there's no replacement, so an `OnChain` flag there means no address will ever be provided. Added `OnChainDepMissingReplacement` error variant for this case. Also added an invariant doc on the `Combined` type alias noting that all on-chain deps have addresses after combining.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Readded on the `Root` variant where it belongs.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Agreed — removed `OnChain(OnChainFlag)` from `ResolverDependencyInfo`, renamed `OnChainAt` to `OnChain`. The resolve step now hits `unreachable!()` for the flag variant (validated at combine time). Updated `ResolvedDependency` doc to note all on-chain deps have addresses.

mdgeorge4153: [CLAUDE] Updated the doc comment on `ResolvedDependency`.

mdgeorge4153: [CLAUDE] Done — moved to combine time via `OnChainDepMissingReplacement` error in `from_default`.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Keeping chain_id for now pending the team's answer on overlap.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Renamed to `address`.

external-crates/move/crates/move-package-alt/src/schema/resolver.rs
mdgeorge4153: [CLAUDE] Done — removed `OnChain(OnChainFlag)` variant, renamed `OnChainAt` to `OnChain`.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153, line 200: We're not establishing the new invariant here (I had mentioned early that these checks were premature, but then we moved the invariant to this stage, so we need to enforce it here).

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
mdgeorge4153: The code is there, but I don't see the tests.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: We also need to validate in `from_default_with_replacement`

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Good catch — added the invariant check in `from_default_with_replacement` too. After `unwrap_or`, if the result is still `OnChain(flag)` (neither default nor replacement provided an address), we now error.

mdgeorge4153: This will not give an error if both the default dependency has an OnChainAt I think, unless that's caught elsewhere? This case should have a test.

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
mdgeorge4153: [CLAUDE] Added integration tests using `PackageLoader` that exercise the combine-time validation:
- `on_chain_address_in_deps_rejected`: on-chain = "0x..." in [dependencies]
- `on_chain_flag_without_replacement_rejected`: on-chain = true with no replacement
- `on_chain_flag_in_replacement_rejected`: on-chain = true in [dep-replacements]
- `on_chain_address_in_deps_with_replacement_rejected`: on-chain = "0x..." in [dependencies] even with replacement
- `on_chain_flag_with_address_replacement_ok`: happy path succeeds past combining

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Fixed — now also validates in `from_default_with_replacement` that the default doesn't have `OnChainAt` (address in [dependencies]), and that after `unwrap_or` the result isn't `OnChain` (flag with no address). Added integration tests for all cases.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -182,3 +220,143 @@ impl From<CombinedDependency> for ReplacementDependency {
...
+
+    use crate::{
+        PackageLoader,
+        flavor::Vanilla,
+        schema::Environment,
+        test_utils::{self, basic_manifest},
+    };
+
+    fn test_env() -> Environment {
+        Environment::new("mainnet".to_string(), "35834a8a".to_string())
```
mdgeorge4153: We should avoid referring to `mainnet` in `external-crates I think.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -182,3 +220,143 @@ impl From<CombinedDependency> for ReplacementDependency {
...
+    }
+
+    /// on-chain = "0x..." in [dependencies] is rejected
+    #[test(tokio::test)]
+    async fn on_chain_address_in_deps_rejected() {
+        let project = test_utils::project()
+            .file(
+                "Move.toml",
+                &format!(
```
mdgeorge4153: Use `formatdoc!` to make this file more readable.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -182,3 +220,143 @@ impl From<CombinedDependency> for ReplacementDependency {
...
+                ),
+            )
+            .build();
+
+        let err = PackageLoader::new(&project.root(), test_env(), Vanilla)
+            .load()
+            .await
+            .unwrap_err();
+        let msg = err.to_string();
+        assert!(msg.contains("must use `on-chain = true`"), "got: {msg}");
```
mdgeorge4153: I think I prefer just snapshotting the error message so that I can see what the whole thing looks like during review.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -182,3 +220,143 @@ impl From<CombinedDependency> for ReplacementDependency {
...
+    }
+
+    /// on-chain = true in [dependencies] with no replacement is rejected
+    #[test(tokio::test)]
+    async fn on_chain_flag_without_replacement_rejected() {
+        let project = test_utils::project()
+            .file(
```
mdgeorge4153: I would consider adding a method to the test graph generator to make these tests simpler, but I'm not sure how much benefit there is. The advantage is that when we start to implement this stuff, we could add some fake on-chain packages to `Vanilla`.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Switched to using `TestPackageGraph::add_on_chain_dep` — no fake dep packages needed. Tests use `assert_snapshot!` for error messages. The on-chain value is passed unquoted (e.g. `"0x01"`) and the renderer adds TOML quotes automatically.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Tests now use `TestPackageGraph` with `add_on_chain_dep` instead of raw manifest strings.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Switched to `assert_snapshot!` for all error cases. The happy-path test uses `assert!(err.contains(...))` since the fetch error contains a non-portable path.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Added `add_on_chain_dep` to `TestPackageGraph` and `OnChainSpec` to `PackageSpec`. This could be extended in the future to support fake on-chain packages in `Vanilla`.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
+# Address Identity in the Dependency Graph
+
+## Problem (TODO)
```
stefan-mysten:
```suggestion
## Problem
```

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+
+The dependency system distinguishes between "node" concepts (where a package's source
+lives) and "edge" concepts (how a parent references a dependency). The `Pinned` enum
+represents nodes, and `DependencyContext` represents edges. However, `published_at` and
+`original_id` don't fit cleanly into either category, and the current placement leads to
+design problems. The items below are not yet resolved.
```
stefan-mysten, line 9: This is very confusing to me, maybe I lack some context / up-to-date knowledge about how the code is structured.

What has `published_at` to do with `node`? At least from the text is not clear. Is the `package's source lives` referring to on-chain bytecode (and its address) or github source?

What is a `parent` here? A package that has dependencies?

Leaving that aside, the text should explain why they do not fit cleanly into either category, and what these things are and how they are related to the node and edge concepts.

stefan-mysten: Ah, I see that below. Maybe the doc should start with background and then explain the problem.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+
+- **`original_id`**: The address the package was first published at. This serves as the
+  package's identity in the Move type system — types are identified by `original_id::module::Type`.
+- **`published_at`**: The address where the current version of the package lives on-chain.
+  After an upgrade, `published_at` changes but `original_id` stays the same.
+
+These addresses can come from several sources:
+
+1. **The package's own files** — for legacy packages, this may be in `Move.toml` or
+   `Move.lock`. For modern packages, it is in `Move.published`.
```
stefan-mysten, line 23: not relevant for this PR. In hindsight, maybe we should have "versioned" the package system so that we don't make ourselves restricted on legacy/modern naming. What happens when we have a modern modern package? 😅

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+remains important — two edges claiming the *same* `original_id` with *different*
+`published_at` for what should be the *same* compiled package is a conflict.
+
+## Status
+
+- [x] Put `published_at` into `Pinned::OnChain` (immediate, unblocks on-chain deps)
+- [ ] Decide whether addresses belong on all `Pinned` variants or only `OnChain`
+- [ ] Update ephemeral publication keys to account for address overrides
+- [ ] Implement the address consistency check from DESIGN.md
+- [ ] Consider whether `original_id` should also be in `Pinned::OnChain` or remain in context
```
stefan-mysten: nit: I would keep these in a todo list rather than design doc itself.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+important because it exposes a fundamental question about the design model. We could
+detect this situation and fail, but the fact that we're tripping on these kinds of issues
+suggests we need to think more clearly about the relationship between source identity,
+compiled identity, and address identity.
+
+In the current system:
+- Both edges point to the same fetched source (same git SHA, same local path on disk)
+- But they have different `published_at` / `original_id` via `dep-replacements`
+- The compiled bytecode differs because the self-address differs
+- The type systems are disjoint — `0xA::module::T` and `0xB::module::T` are unrelated types
```
stefan-mysten, line 45: This makes sense

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+**From a compiled/on-chain perspective**: different nodes. They produce different bytecode,
+have different type identities, and are independently upgradeable.
+
+## Design Decision
+
+Addresses are part of the **compiled identity** of a package, but not part of the **source
+identity**. This means:
+
+1. **Source fetching and caching** should be based on the source location alone (git SHA +
+   path, local path, etc.). Two deps with the same source but different addresses should
```
stefan-mysten, line 60: And we know they have the same source because they point to the same git repo + sha? But in the case of `on-chain deps`, we don't know they have the same source. For source fetching it seems reasonable to share the disk location, as long as the SHA matches for both packages.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+   share the same fetched files on disk. The current source cache layout does not need to
+   change.
+
+2. **Compiled identity** (for ephemeral publication, linkage, and type checking) must
```
stefan-mysten, line 64: Would we keep some sort of mapping between source + sha + path etc., to compiled artifacts in the cache?

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+   it determines which bytecode to fetch from the network. So for on-chain deps,
+   `published_at` is unambiguously part of both the source identity and the compiled
+   identity.
+
+## Implications for `Pinned`
+
+If we treat the same source with different addresses as different nodes in the dependency
+graph, then addresses become part of the node identity for **all** `Pinned` variants, not
+just `OnChain`. This would mean `PinnedLocalDependency`, `PinnedGitDependency`, etc. would
+all carry address information.
```
stefan-mysten, line 75: That seems reasonable. Any real downside in having them carry this?

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+
+`unfetched_path` (for source fetching) would ignore addresses, since the source cache is
+keyed by source location alone. But ephemeral publication keys, linkage, and any future
+compiled artifact cache would use addresses as part of the key.
+
+### Immediate step: `Pinned::OnChain` should carry `published_at`
+
+Since `published_at` determines *what to fetch* for on-chain deps (analogous to the git
```
stefan-mysten, line 83: Mostly talking to myself: do we only need `published_at` but not `original_id`? I think so, because that's the package we need to fetch from the network.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+SHA for git deps), it belongs in the `Pinned` enum regardless of the broader design:
+
+```rust
+pub(crate) enum Pinned {
```
stefan-mysten, line 87: Also mostly talking to myself, it may make no sense: would it make sense to have another top level type that has the `published_at` field, and a `pinned_type` field?
```suggestion
pub(crate) struct PinnedData {
   published_at: PublishedID,
   pinned_type: Pinned
}

pub(crate) enum Pinned {
...
}
```

stefan-mysten: Probably this does not work for Local, unless they get published?

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+`EphemeralDependencyInfo`, which wraps `LocalDepInfo` — just the absolute local path to the
+source. The `published_at` and `original_id` are stored as data alongside the key, not as
+part of the key itself.
+
+This means two deps with the same source but different address overrides would collide on
+the same key. If we support the same source at multiple addresses (whether we treat it as
+different nodes or not), the ephemeral publication key should incorporate the address
+override.
+
+### Linkage should validate address consistency
```
stefan-mysten, line 110: This looks to me like it would be blocking the above design + implementation. Not sure I fully understand this section.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -17,7 +17,10 @@ use crate::{
 
 use super::DependencyContext;
 
-/// The dep_info type for the combined stage.
+/// The dep_info type for the combined stage. This is `ManifestDependencyInfo`, but with the
+/// additional invariant that all on-chain dependencies have addresses (i.e. they use the
+/// `OnChainAt` variant, not `OnChain`). This is enforced during combining: `on-chain = true`
```
stefan-mysten: nit: OnChainAt and OnChain are so close to each other that sooner or later they will get people confused. Any way we could be more verbose here?

external-crates/move/crates/move-package-alt/src/dependency/fetch.rs
```diff
@@ -26,20 +25,25 @@ pub enum FetchError {
...
+    OnChainNotSupported,
 }
 
 pub type FetchResult<T> = Result<T, FetchError>;
 
 /// Ensure that the dependency's files are present on the disk and return a path to them.
 /// Assumes that `pinned` is already normalized - paths of any local dependencies are relative
 /// to the current working directory, and local dependencies of git dependencies have been
-/// transformed into git dependencies. `chain_id` is passed through to [Pinned::unfetched_path].
+/// transformed into git dependencies. `chain_id` is used to determine the cache location for
```
stefan-mysten, line 38: Is the cache folder now prefixed by chain id?

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
```diff
@@ -193,8 +195,19 @@ impl<'de> Deserialize<'de> for ManifestDependencyInfo {
...
+                    Ok(ManifestDependencyInfo::OnChain(dep))
+                } else if val.is_str() {
+                    let dep = OnChainAddress::deserialize(data).map_err(de::Error::custom)?;
+                    Ok(ManifestDependencyInfo::OnChainAt(dep))
+                } else {
+                    Err(de::Error::custom(
+                        "on-chain must be `true` (in [dependencies]) or a hex address string \
+                         like \"0x...\" (in [dep-replacements])",
+                    ))
+                }
```
stefan-mysten: Maybe a bit cleaner, but not really important.


```suggestion
                match &tbl["on-chain"] {
                        toml::Value::Boolean(_) => OnChainFlag::deserialize(data)
                            .map(ManifestDependencyInfo::OnChain)
                            .map_err(de::Error::custom),
                        toml::Value::String(_) => OnChainAddress::deserialize(data)
                            .map(ManifestDependencyInfo::OnChainAt)
                            .map_err(de::Error::custom),
                        _ => Err(de::Error::custom(
                            "on-chain must be `true` (in [dependencies]) or a hex address string \
                             like \"0x...\" (in [dep-replacements])",
                        )),
                }
```

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
```diff
@@ -985,6 +998,98 @@ mod tests {
...
+          |
+        7 |             foo = { on-chain = false }
+          |                   ^^^^^^^^^^^^^^^^^^^^
+        Expected the constant `true` for key `on-chain`
+        "###);
+    }
+
+    /// Parsing `on-chain = 42` is rejected
+    #[test]
+    fn parse_on_chain_integer() {
```
stefan-mysten, line 1069: would it make sense to add a non 32 byte hexadecimal test? E.g., one that has more characters, or a 0x000000000000000000000000000000000000000000000000000000000000000q

external-crates/move/crates/move-package-alt/src/schema/shared.rs
```diff
@@ -58,13 +58,26 @@ pub struct LocalDepInfo {
...
 }
 
-/// An on-chain dependency `{on-chain = true}`
+/// An on-chain dependency marker: `{on-chain = true}`.
+///
+/// Used in the `[dependencies]` section of a manifest to declare that a dependency should be
+/// fetched from the chain rather than from source.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
-pub struct OnChainDepInfo {
+pub struct OnChainFlag {
```
stefan-mysten: nit: name them the same, either flag or marker.

external-crates/move/crates/move-package-alt/src/test_utils/graph_builder.rs
```diff
@@ -124,6 +124,16 @@ pub struct PackageSpec {
     /// Custom addresses for legacy packages (name -> Option<address>)
     /// If None is provided, the address is considered to be the legacy `_`.
     legacy_addresses: BTreeMap<String, Option<String>>,
+
+    /// On-chain dependencies (not backed by graph nodes)
+    on_chain_deps: Vec<OnChainSpec>,
+}
+
+/// An on-chain dependency, stored directly on the source node (no target node in the graph).
```
stefan-mysten, line 132: Not sure I follow why this is needed. Just for tests?

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
```diff
@@ -297,27 +317,34 @@ impl TryFrom<LockfileGitDepInfo> for PinnedGitDependency {
...
             }),
-            Pinned::Root(_) => Pinned::Local(PinnedLocalDependency {
-                absolute_path_to_package: parent.unfetched_path(chain_id).join(&self.local).clean(),
+            Pinned::Root(root) => Pinned::Local(PinnedLocalDependency {
+                absolute_path_to_package: root.path().join(&self.local).clean(),
                 relative_path_from_root_package: self.local.clean(),
             }),
-            Pinned::OnChain(_) => todo!(),
+            Pinned::OnChain { .. } => {
+                unreachable!("on-chain packages have system-generated manifests with no local deps")
```
stefan-mysten: Do we want to do some logging here?

stefan-mysten (commented):
Left a few more comments. Overall looks fine, I think. Need to refresh a bit my memory in the pkg-system as I forgot a bunch of types and how all flows together.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
```diff
@@ -0,0 +1,129 @@
...
+   share the same fetched files on disk. The current source cache layout does not need to
+   change.
+
+2. **Compiled identity** (for ephemeral publication, linkage, and type checking) must
+   account for addresses. The same source with different address overrides produces
+   different compilation artifacts. While we don't currently cache compiled artifacts, it
+   seems plausible that we would do so in the future, and such a cache would need to be
+   keyed by (source location + addresses).
+
+3. **For on-chain dependencies**, `published_at` is additionally the **source locator** —
```
mdgeorge4153: This is no longer true in the v2 design of on-chain pinning

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: yes, although this section is also out of date I've noticed. For this version of on-chain pinning, the `on-chain` field is used for the address, rather than reusing the `published-at` field. This sidesteps some of the issues.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: I agree - I think at the point where we need a v3 we should do the refactor so that we have a general-purpose versioned schema.

external-crates/move/crates/move-package-alt/src/dependency/fetch.rs
mdgeorge4153: yes

external-crates/move/crates/move-package-alt/src/test_utils/graph_builder.rs
mdgeorge4153: yes, this whole file is for building tests, and this is how we keep track of the on-chain deps that we're adding to tests.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: for on-chain deps, they will have the same "source", because that's deterministically computed from the on-chain bytecode, which is the same for any dep with the same chain ID / package ID.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: I think if we were to keep compiled artifacts in the cache, we'd just put them inside the source packages (that is, if we have source at `~/.move/https__...__foo_<sha>/packages/bar/{Move.toml, sources, ...}` then we'd compile into `~/.move/https__....__foo_<sha>/packages/bar/build/package.mv`.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: I don't think so! We might want to fold them into a parent struct, but I think that's fine.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: There's two kinds of conflicts here, one of which is fine and one of which is not. If we have two nodes with the same original ID and different published-at, then that means they are two versions of the same package (upgrading changes published-at but not original-id). That's a genuine conflict, because we have two packages that define the same (logical) types.

OTOH, if we have two nodes that have the same source but different original IDs, then there is no conflict - there's just two separate on-chain packages that happen to share the same source code.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: Agreed, will put these in the linear task.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Renamed: `OnChain(OnChainFlag)` → `OnChainPlaceholder(OnChainPlaceholder)` and `OnChainAt(OnChainAddress)` → `OnChain(OnChainAddress)`. The placeholder variant is the `on-chain = true` form; `OnChain` is the one with an address.

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
mdgeorge4153: [CLAUDE] Applied — using `match &tbl["on-chain"]` with `toml::Value::Boolean`/`String` arms.

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
mdgeorge4153: [CLAUDE] Added `parse_on_chain_short_address` test that verifies `on-chain = "0x1"` parses successfully (addresses are zero-padded by `AccountAddress::from_hex`).

external-crates/move/crates/move-package-alt/src/schema/shared.rs
mdgeorge4153: [CLAUDE] Renamed `OnChainFlag` to `OnChainPlaceholder` to match the enum variant name.

external-crates/move/crates/move-package-alt/src/dependency/pin.rs
mdgeorge4153: [CLAUDE] Improved the `unreachable!` message to include the on-chain package address and the local dependency path, so if this invariant ever breaks the error gives enough context to debug.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: [CLAUDE] Fixed — dropped the `(TODO)`.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: [CLAUDE] Replaced the checklist with a pointer to DVX-2115 in Linear.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: [CLAUDE] Agreed — Local deps do not have a published_at, so this wrapper would not work uniformly.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
mdgeorge4153: [CLAUDE] Removed bullet 3 — on-chain deps are now handled separately via `on-chain = "0x..."` in dep-replacements, so this point about published_at as source locator is no longer relevant.

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
mdgeorge4153: [CLAUDE] Updated: added `parse_on_chain_too_long_address` (>32 byte hex) and `parse_on_chain_invalid_hex` (`"0x0000q"`) tests. Both are `#[ignore]` because invalid addresses are currently silently dropped due to `serde(flatten, default)` — tracked in DVX-2143.

external-crates/move/crates/move-package-alt/design/dvx-2115-address-identity.md
stefan-mysten: > with the same original ID and different published-at

But this isn't possible, no? Unless, someone manually mangles the published-at addresses in a manifest.

stefan-mysten (approved):

stefan-mysten:
Looks good, thanks @mdgeorge4153

PR #26559, opened 2026-05-08
[move][move-vm] Improve telemetry speed by swapping to thread-local atomics

cgswords:
## Description 

image

## Test plan 

Everything is green, and testnet testing demonstrates massive speedup for telemetry times.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/runtime/telemetry.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/runtime/telemetry.rs
@@ -204,7 +204,15 @@ struct LocalSlot {
 impl Drop for LocalSlot {
     fn drop(&mut self) {
         if let Some(ctx) = self.ctx.upgrade() {
-            ctx.aggregated.lock().add_from(&self.counters);
+            // Hold `threads` across the Arc drop so the reporter cannot observe `aggregated`
+            // already updated while our `Weak` is still upgradeable. Lock order matches
+            // `to_runtime_telemetry`: threads, then aggregated.
+            let _threads = ctx.threads.lock();
+            // Declaration order is load-bearing: `_counters` drops before `_threads`, so the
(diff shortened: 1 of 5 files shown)


external-crates/move/crates/move-vm-runtime/src/runtime/telemetry.rs
```diff
@@ -80,46 +87,143 @@ pub struct MoveRuntimeTelemetry {
...
+struct LocalSlot {
+    counters: Arc<ThreadCounters>,
+    ctx: Weak<TelemetryContext>,
+}
+
+impl Drop for LocalSlot {
+    fn drop(&mut self) {
+        if let Some(ctx) = self.ctx.upgrade() {
+            ctx.aggregated.lock().add_from(&self.counters);
+        }
```
tzakian, line 216: There's still ostensibly a race condition here I believe where a telemetry report could occur here, but before the `LocalSlot` is fully dropped and the `Arc` downgraded. (worst case I think this could lead to a double-count of the telemetry for the thread but that's it). 

I don't think this will matter in practice, but worth calling out just to make sure it's all fine.

tzakian (approved):
One possible race condition that could lead to double-counting some telemetry but I think in practice this is totally fine (both because it would be a double-count, and also because the `Drop` should very rarely if ever be called unless we are dropping the entire runtime at epoch boundaries).

external-crates/move/crates/move-vm-runtime/src/runtime/telemetry.rs
cgswords: Fixed with more-precise behavior around the mutex protecting that list.

PR #26561, opened 2026-05-11
[onchain-deps 1] Implement on-chain package fetching and manifest generation

mdgeorge4153:
## Summary

Implements on-chain package fetching and manifest generation for move-package-alt,
stacked on #26552 (on-chain pinning v2) and #26562 (Vanilla::new refactor).

- **`OnChainPackageData`** type and **`fetch_onchain_package`** trait method on `MoveFlavor`
  - Vanilla: looks up from pre-populated test data map
  - SuiFlavor: fetches via gRPC `get_object`, extracts `MovePackage` fields
- **`on_chain` module** (parallel to `git`): fetches bytecode, generates `Move.toml` and
  `Published.toml` in `~/.move/on-chain/<chain_id>/<address>/`
- **Fixed `_on_chain` environment**: generated manifests use a single environment name,
  avoiding per-package environment name conflicts. `use_environment` is forced during combining.
- **`OnChainError`** error type, **`PackageSystemLock::new_for_onchain`** for FS locking
- **Skip rename-from** for on-chain deps (synthetic `name = "self"`)
- **`add_on_chain_pkg`** on `TestPackageGraph` with `OnChainPkgSpec` builder
- **`write_source`** stub for future stub generation integration

### Known limitations
- System packages in linkage tables are fetched on-chain instead of using source deps (DVX-2126)
- Manifest serialization uses hybrid approach due to DVX-2125
- On-chain cache uses global `MOVE_HOME`; per-test isolation deferred (DVX-2127)
- Full pipeline requires stub source generation (PR #26555)

## Test plan

- [x] 192 tests pass (8 new + 184 existing), clippy/fmt clean
- [x] Unit tests: bytecode write, manifest generation (with/without deps), Published.toml, round-trip
- [x] Integration: transitive on-chain deps, rename-from skip
- [x] Ignored test documenting expected deduplication behavior (DVX-2126)
- [ ] Shell tests in sui crate (deferred — blocked on system package handling)

### DIFF
--- a/crates/sui-package-alt/src/sui_flavor.rs
+++ b/crates/sui-package-alt/src/sui_flavor.rs
@@ -230,6 +230,14 @@ impl MoveFlavor for SuiFlavor {
     fn is_system_address(&self, address: &move_package_alt::schema::OriginalID) -> bool {
         is_system_package(address.0)
     }
+
+    async fn fetch_onchain_package(
+        &self,
+        _address: &move_package_alt::schema::PublishedID,
+    ) -> anyhow::Result<move_package_alt::OnChainPackageData> {
+        // TODO: implement via gRPC get_object → MovePackage
+        anyhow::bail!("on-chain package fetching is not yet implemented for SuiFlavor")
+    }
 }
 
 /// We validate that a modern manifest cannot define the "legacy" system names.
--- a/crates/sui-package-alt/src/sui_flavor.rs
+++ b/crates/sui-package-alt/src/sui_flavor.rs
@@ -8,10 +8,10 @@ use async_trait::async_trait;
 use indexmap::IndexMap;
 use move_compiler::editions::Edition;
 use move_package_alt::{
-    MoveFlavor,
+    MoveFlavor, OnChainPackageData,
     schema::{
         EnvironmentID, EnvironmentName, GitSha, LockfileDependencyInfo, LockfileGitDepInfo,
-        PackageName, ParsedManifest, ReplacementDependency, SystemDepName,
+        OriginalID, PackageName, ParsedManifest, PublishedID, ReplacementDependency, SystemDepName,
     },
 };
 
@@ -231,10 +231,12 @@ impl MoveFlavor for SuiFlavor {
         is_system_package(address.0)
     }
 
+    /// Fetch an on-chain package via gRPC. Returns an error if there is no network connection
+    /// or if the object at `address` is not a package.
     async fn fetch_onchain_package(
         &self,
-        address: &move_package_alt::schema::PublishedID,
-    ) -> anyhow::Result<move_package_alt::OnChainPackageData> {
+        address: &PublishedID,
+    ) -> anyhow::Result<OnChainPackageData> {
         let Some(ref client) = self.client else {
             anyhow::bail!("cannot fetch on-chain package {address}: no network connection")
         };
@@ -244,7 +246,7 @@ impl MoveFlavor for SuiFlavor {
         let package = object
             .data
             .try_as_package()
-            .ok_or_else(|| anyhow::anyhow!("object {address} is not a package"))?;
+            .with_context(|| format!("object {address} is not a package"))?;
 
         let modules = package.serialized_module_map().clone();
         let dependencies = package
@@ -252,16 +254,15 @@ impl MoveFlavor for SuiFlavor {
             .iter()
             .map(|(original_id, upgrade_info)| {
                 (
-                    move_package_alt::schema::OriginalID((*original_id).into()),
-                    move_package_alt::schema::PublishedID(upgrade_info.upgraded_id.into()),
+                    OriginalID((*original_id).into()),
+                    PublishedID(upgrade_info.upgraded_id.into()),
                 )
             })
             .collect();
-        let original_id =
-            move_package_alt::schema::OriginalID(package.original_package_id().into());
+        let original_id = OriginalID(package.original_package_id().into());
         let version: u64 = package.version().into();
 
-        Ok(move_package_alt::OnChainPackageData {
+        Ok(OnChainPackageData {
             modules,
             dependencies,
             original_id,
--- a/crates/sui-package-alt/src/sui_flavor.rs
+++ b/crates/sui-package-alt/src/sui_flavor.rs
@@ -233,10 +233,43 @@ impl MoveFlavor for SuiFlavor {
 
     async fn fetch_onchain_package(
         &self,
-        _address: &move_package_alt::schema::PublishedID,
+        address: &move_package_alt::schema::PublishedID,
     ) -> anyhow::Result<move_package_alt::OnChainPackageData> {
-        // TODO: implement via gRPC get_object → MovePackage
-        anyhow::bail!("on-chain package fetching is not yet implemented for SuiFlavor")
+        let Some(ref client) = self.client else {
+            anyhow::bail!(
+                "cannot fetch on-chain package {address}: no network connection \
+                 (SuiFlavor created in offline mode)"
+            )
+        };
+
+        let object_id = ObjectID::from(address.0);
+        let object = client.clone().get_object(object_id).await?;
+        let package = object
+            .data
+            .try_as_package()
+            .ok_or_else(|| anyhow::anyhow!("object {address} is not a package"))?;
+
+        let modules = package.serialized_module_map().clone();
+        let dependencies = package
+            .linkage_table()
+            .iter()
+            .map(|(original_id, upgrade_info)| {
+                (
+                    move_package_alt::schema::OriginalID((*original_id).into()),
+                    move_package_alt::schema::PublishedID(upgrade_info.upgraded_id.into()),
+                )
+            })
+            .collect();
+        let original_id =
+            move_package_alt::schema::OriginalID(package.original_package_id().into());
+        let version: u64 = package.version().into();
+
+        Ok(move_package_alt::OnChainPackageData {
+            modules,
+            dependencies,
+            original_id,
+            version,
+        })
     }
 }
 
--- a/crates/sui-package-alt/src/sui_flavor.rs
+++ b/crates/sui-package-alt/src/sui_flavor.rs
@@ -236,9 +236,7 @@ impl MoveFlavor for SuiFlavor {
         address: &move_package_alt::schema::PublishedID,
     ) -> anyhow::Result<move_package_alt::OnChainPackageData> {
         let Some(ref client) = self.client else {
-            anyhow::bail!(
-                "cannot fetch on-chain package {address}: no network connection"
-            )
+            anyhow::bail!("cannot fetch on-chain package {address}: no network connection")
         };
 
         let object_id = ObjectID::from(address.0);
--- a/crates/sui-package-alt/src/sui_flavor.rs
+++ b/crates/sui-package-alt/src/sui_flavor.rs
@@ -237,8 +237,7 @@ impl MoveFlavor for SuiFlavor {
     ) -> anyhow::Result<move_package_alt::OnChainPackageData> {
         let Some(ref client) = self.client else {
             anyhow::bail!(
-                "cannot fetch on-chain package {address}: no network connection \
-                 (SuiFlavor created in offline mode)"
+                "cannot fetch on-chain package {address}: no network connection"
             )
         };
 
--- a/external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
+++ b/external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
@@ -20,15 +20,6 @@ The package system fetches the bytecode and linkage table from the network (via 
 cache, and loads the package like any other dependency. Stub Move source files are
 generated from the bytecode so the compiler can type-check against the dependency.
 
-### Implementation status
-
-- Schema and validation: **done** (#26552)
-- Fetching and manifest generation: **done** (#26561)
-- Stub source generation: DVX-2119
-- System package deduplication: DVX-2126
-- Support for `sui move test-publish`: DVX-2120
-- Support for `sui move test`: DVX-2060
-
 ## User-facing manifest format
 
 On-chain dependencies use two forms of the `on-chain` key:
@@ -91,7 +82,7 @@ The generated `Move.toml` for an on-chain package looks like:
 
 ```toml
 [package]
-name = "self"
+name = "_onchain_package"
 implicit-dependencies = false
 
 [environments]
@@ -104,7 +95,7 @@ onchain_0x0003 = { on-chain = "0x0033", override = true }
 
 Key design choices:
 
-- **`name = "self"`** — a fixed name for all generated packages. Users never need to know
+- **`name = "_onchain_package"`** — a fixed name for all generated packages. Users never need to know
   this name because the rename-from check is skipped for on-chain dependencies.
 - **`implicit-dependencies = false`** — prevents the generated manifest from pulling in the
   flavor's implicit dependencies (e.g. `sui`, `std`), which would conflict with the system
@@ -141,7 +132,7 @@ calls `get_object` via gRPC, extracts the `MovePackage` from the object, and map
 
 After fetching, `Package::load` reads the generated manifest and proceeds through the
 normal loading pipeline — combining deps, checking environments, loading publications.
-The rename-from check is skipped for on-chain deps (the `"self"` package name won't match
+The rename-from check is skipped for on-chain deps (the `"_onchain_package"` name won't match
 the user's dependency name).
 
 ## Environment handling
--- a/external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
+++ b/external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
@@ -1,0 +1,191 @@
+# On-Chain Package Fetching
+
+## Overview
+
+On-chain dependencies allow Move packages to depend on other packages by their on-chain
+address, without requiring source code. This is needed because MVR is currently
+source-only, which breaks when source isn't published or doesn't match on-chain bytecode.
+
+A user declares an on-chain dependency in their `Move.toml`:
+```toml
+[dependencies]
+my_dep = { on-chain = true }
+
+[dep-replacements.mainnet]
+my_dep = { on-chain = "0x1234..." }
+```
+
+The package system fetches the bytecode and linkage table from the network (via the
+`MoveFlavor` trait), generates a synthetic `Move.toml` and `Published.toml` in a local
+cache, and loads the package like any other dependency. Stub Move source files are
+generated from the bytecode so the compiler can type-check against the dependency.
+
+### Implementation status
+
+- Schema and validation: **done** (#26552)
+- Fetching and manifest generation: **done** (#26561)
+- Stub source generation: DVX-2119
+- System package deduplication: DVX-2126
+- Support for `sui move test-publish`: DVX-2120
+- Support for `sui move test`: DVX-2060
+
+## User-facing manifest format
+
+On-chain dependencies use two forms of the `on-chain` key:
+
+- **`on-chain = true`** in `[dependencies]` — declares that a dependency is on-chain,
+  without specifying an address. The address must be provided per-environment in
+  `[dep-replacements]`.
+- **`on-chain = "0x..."`** in `[dep-replacements]` — provides the on-chain address for a
+  specific environment.
+
+```toml
+[package]
+name = "my_package"
+
+[dependencies]
+my_dep = { on-chain = true }
+
+[dep-replacements.mainnet]
+my_dep = { on-chain = "0x1234..." }
+
+[dep-replacements.testnet]
+my_dep = { on-chain = "0x5678..." }
+```
+
+Validation rules enforced during combining:
+- `on-chain = "0x..."` in `[dependencies]` is rejected (addresses belong in dep-replacements)
+- `on-chain = true` in `[dep-replacements]` is rejected (replacements must have an address)
+- `on-chain = true` with no dep-replacement is rejected (no address to fetch)
+- `use-environment` on on-chain deps is rejected (on-chain packages use a fixed environment)
+
+## Cache layout
+
+On-chain packages are cached at `~/.move/on-chain/<chain_id>/<address>/`. The cache is
+populated on first fetch and reused on subsequent builds. A filesystem lock
+(`PackageSystemLock::new_for_onchain`) serializes concurrent access.
+
+```
+~/.move/on-chain/<chain_id>/<address>/
+├── Move.toml          # generated manifest
+├── Published.toml     # generated publication metadata
+├── bytecode/           
+│   ├── module_a.mv    # serialized CompiledModule bytecode
+│   └── module_b.mv
+└── sources/            # (future) generated stub Move source
+    ├── module_a.move
+    └── module_b.move
+```
+
+The `bytecode/` directory uses a different path from the compiler's
+`build/<pkg>/bytecode_modules/` to avoid clobbering the original bytecode if the package
+is recompiled.
+
+The cache check uses manifest existence: if `Move.toml` exists, the fetch is skipped
+entirely. The cache is written atomically — bytecode, source stubs, manifest, and
+Published.toml are all written in a single locked section, with the manifest written last.
+
+## Generated manifest
+
+The generated `Move.toml` for an on-chain package looks like:
+
+```toml
+[package]
+name = "self"
+implicit-dependencies = false
+
+[environments]
+_on_chain = "<chain_id>"
+
+[dep-replacements._on_chain]
+onchain_0x0002 = { on-chain = "0x0002", override = true }
+onchain_0x0003 = { on-chain = "0x0033", override = true }
+```
+
+Key design choices:
+
+- **`name = "self"`** — a fixed name for all generated packages. Users never need to know
+  this name because the rename-from check is skipped for on-chain dependencies.
+- **`implicit-dependencies = false`** — prevents the generated manifest from pulling in the
+  flavor's implicit dependencies (e.g. `sui`, `std`), which would conflict with the system
+  deps already in the graph.
+- **`override = true`** — on-chain dep-replacements are overrides so they take precedence
+  when the same dependency appears in the graph.
+- **Dependency names** — linkage table entries are named `onchain_<original_id>`. These
+  names are arbitrary identifiers; the `on-chain` address is what matters.
+- **`Published.toml`** — records the package's `published-at`, `original-id`, `version`,
+  and `chain-id` under the `_on_chain` environment, following the same schema as
+  user-authored Published.toml files.
+
+## Fetching pipeline
+
+When `fetch()` encounters a `Pinned::OnChain { address }`, it delegates to
+`on_chain::fetch::fetch_onchain`, which:
+
+1. Computes the cache directory from `MOVE_HOME`, chain ID, and address
+2. Acquires an exclusive filesystem lock for the cache directory
+3. Checks if the manifest already exists (cache hit → return immediately)
+4. Calls `config.flavor.fetch_onchain_package(address)` to download
+   `OnChainPackageData` from the network (modules, linkage table, original ID, version)
+5. Writes bytecode to `bytecode/<name>.mv`
+6. Generates stub source files (currently a no-op, DVX-2119)
+7. Generates `Move.toml` via `ParsedManifest` + `RenderToml`
+8. Generates `Published.toml` via `ParsedPublishedFile` + `RenderToml`
+9. Returns the cache directory path as a `PackagePath`
+
+The `MoveFlavor::fetch_onchain_package` trait method is the only network-touching
+operation. `Vanilla` looks up from a pre-populated `BTreeMap` (for testing). `SuiFlavor`
+calls `get_object` via gRPC, extracts the `MovePackage` from the object, and maps
+`module_map`, `linkage_table`, `original_package_id()`, and `version` to
+`OnChainPackageData`.
+
+After fetching, `Package::load` reads the generated manifest and proceeds through the
+normal loading pipeline — combining deps, checking environments, loading publications.
+The rename-from check is skipped for on-chain deps (the `"self"` package name won't match
+the user's dependency name).
+
+## Environment handling
+
+Environment names are per-package: different packages in the dependency graph may use
+different names for the same network. For example, one package might call mainnet
+`"mainnet"` while another calls it `"production"`.
+
+On-chain packages are tied to a specific chain (determined by the chain ID at fetch time),
+so they don't need environment flexibility. Generated manifests use a single fixed
+environment name `_on_chain` mapped to the chain ID.
+
+To connect the caller's environment to the generated package's `_on_chain` environment,
+`use_environment` is forced to `_on_chain` during combining for all `OnChainAt`
+dependencies (in `CombinedDependency::from_replacement` and
+`from_default_with_replacement`). This means the graph builder loads on-chain packages
+with the `_on_chain` environment regardless of what the parent package's environment is
+called. Explicit `use-environment` on on-chain deps is rejected with an error.
+
+This design means the cache is written once per chain ID + address pair and never needs to
+be updated for different environment names.
+
+## Known limitations and future work
+
+- **System package deduplication (DVX-2126):** The linkage table entries are all turned
+  into on-chain dep-replacements, including entries for system packages (e.g. `0x1`, `0x2`
+  on Sui). System packages are mutable and should instead be resolved as system deps. More
+  generally, a package may be reachable both as an on-chain dep (from a linkage table) and
+  as a source dep (from the user's manifest or system deps). These need to be
+  deduplicated, likely in a post-processing step after the full graph is built. This blocks
+  practical end-to-end testing since almost all on-chain packages depend on system
+  packages.
+- **Stub source generation (DVX-2119):** `write_source` is currently a no-op. Stub Move
+  source files need to be generated from bytecode so the compiler can type-check against
+  on-chain dependencies. A draft implementation exists in PR #26555
+  (`mdgeorge/onchain-stubs`) and will be integrated into `on_chain/move_source.rs`.
+- **Manifest serialization mismatch (DVX-2125):** `ManifestDependencyInfo`'s derived
+  `Serialize` doesn't match its custom `Deserialize`. The `RenderToml` impl on
+  `ParsedManifest` works around this by rendering dep-replacements manually. The proper fix
+  is to decouple the digest computation from serialization, then fix the serializer.
+- **`move_home` threading (DVX-2127):** The on-chain cache, git cache, and lock files all
+  hardcode `MOVE_HOME`. Threading a configurable `move_home` through `PackageConfig` would
+  enable per-test cache isolation. Currently, tests use unique addresses to avoid
+  collisions.
+- **Test-publish (DVX-2120):** Support for `sui move test-publish` with on-chain
+  dependencies.
+- **Unit tests (DVX-2060):** Support for `sui move test` with on-chain dependencies.
--- a/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
+++ b/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
@@ -302,6 +302,6 @@ mod tests {
             move_command_line_common::env::MOVE_HOME.as_str(),
             "<MOVE_HOME>",
         );
-        assert_snapshot!(err, @"Error while loading dependency <MOVE_HOME>/on-chain/_test_env_id/0x0000000000000000000000000000000000000000000000000000000000000001: failed to fetch on-chain package at 0x0000000000000000000000000000000000000000000000000000000000000001: on-chain package not found: 0x0000000000000000000000000000000000000000000000000000000000000001");
+        assert_snapshot!(err, @"Error while loading dependency <MOVE_HOME>/on-chain/_test_env_id/0x0000000000000000000000000000000000000000000000000000000000000001: on-chain package not found: 0x0000000000000000000000000000000000000000000000000000000000000001");
     }
 }
--- a/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
+++ b/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
@@ -297,9 +297,11 @@ mod tests {
             .build();
 
         let err = scenario.root_package_err("root").await;
-        assert!(
-            err.contains("on-chain package not found"),
-            "expected fetch error, got: {err}"
+        // Redact the MOVE_HOME path which varies per user
+        let err = err.replace(
+            move_command_line_common::env::MOVE_HOME.as_str(),
+            "<MOVE_HOME>",
         );
+        assert_snapshot!(err, @"Error while loading dependency <MOVE_HOME>/on-chain/_test_env_id/0x0000000000000000000000000000000000000000000000000000000000000001: failed to fetch on-chain package at 0x0000000000000000000000000000000000000000000000000000000000000001: on-chain package not found: 0x0000000000000000000000000000000000000000000000000000000000000001");
     }
 }
--- a/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
+++ b/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
@@ -147,10 +147,23 @@ impl CombinedDependency {
             ));
         }
 
+        // On-chain deps use a fixed environment; reject explicit use-environment
+        let use_environment =
+            if matches!(&dep.dependency_info, ManifestDependencyInfo::OnChainAt(_)) {
+                if replacement.use_environment.is_some() {
+                    return Err(ManifestError::with_file(file)(
+                        ManifestErrorKind::OnChainWithUseEnvironment { name },
+                    ));
+                }
+                crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
+            } else {
+                replacement.use_environment.unwrap_or(source_env_name)
+            };
+
         Ok(Self {
             context: DependencyContext {
                 name,
-                use_environment: replacement.use_environment.unwrap_or(source_env_name),
+                use_environment,
                 is_override: dep.is_override,
                 addresses: replacement.addresses,
                 containing_file: file,
@@ -184,11 +197,24 @@ impl CombinedDependency {
             ));
         }
 
+        // On-chain deps use a fixed environment; reject explicit use-environment
+        let use_environment =
+            if matches!(&dep.dependency_info, ManifestDependencyInfo::OnChainAt(_)) {
+                if replacement.use_environment.is_some() {
+                    return Err(ManifestError::with_file(file)(
+                        ManifestErrorKind::OnChainWithUseEnvironment { name },
+                    ));
+                }
+                crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
+            } else {
+                replacement.use_environment.unwrap_or(source_env_name)
+            };
+
         // TODO: possibly additional compatibility checks here?
         Ok(Self {
             context: DependencyContext {
                 name,
-                use_environment: replacement.use_environment.unwrap_or(source_env_name),
+                use_environment,
                 is_override: dep.is_override,
                 addresses: replacement.addresses,
                 containing_file: file,
--- a/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
+++ b/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
@@ -148,17 +148,17 @@ impl CombinedDependency {
         }
 
         // On-chain deps use a fixed environment; reject explicit use-environment
-        let use_environment =
-            if matches!(&dep.dependency_info, ManifestDependencyInfo::OnChainAt(_)) {
-                if replacement.use_environment.is_some() {
-                    return Err(ManifestError::with_file(file)(
-                        ManifestErrorKind::OnChainWithUseEnvironment { name },
-                    ));
-                }
-                crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
-            } else {
-                replacement.use_environment.unwrap_or(source_env_name)
-            };
+        let use_environment = if matches!(&dep.dependency_info, ManifestDependencyInfo::OnChain(_))
+        {
+            if replacement.use_environment.is_some() {
+                return Err(ManifestError::with_file(file)(
+                    ManifestErrorKind::OnChainWithUseEnvironment { name },
+                ));
+            }
+            crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
+        } else {
+            replacement.use_environment.unwrap_or(source_env_name)
+        };
 
         Ok(Self {
             context: DependencyContext {
@@ -198,17 +198,17 @@ impl CombinedDependency {
         }
 
         // On-chain deps use a fixed environment; reject explicit use-environment
-        let use_environment =
-            if matches!(&dep.dependency_info, ManifestDependencyInfo::OnChainAt(_)) {
-                if replacement.use_environment.is_some() {
-                    return Err(ManifestError::with_file(file)(
-                        ManifestErrorKind::OnChainWithUseEnvironment { name },
-                    ));
-                }
-                crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
-            } else {
-                replacement.use_environment.unwrap_or(source_env_name)
-            };
+        let use_environment = if matches!(&dep.dependency_info, ManifestDependencyInfo::OnChain(_))
+        {
+            if replacement.use_environment.is_some() {
+                return Err(ManifestError::with_file(file)(
+                    ManifestErrorKind::OnChainWithUseEnvironment { name },
+                ));
+            }
+            crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
+        } else {
+            replacement.use_environment.unwrap_or(source_env_name)
+        };
 
         // TODO: possibly additional compatibility checks here?
         Ok(Self {
--- a/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
+++ b/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
@@ -287,8 +287,8 @@ mod tests {
         );
     }
 
-    /// on-chain = true + address replacement passes combining and fails during fetch.
-    /// This error should change when on-chain dep fetching is implemented.
+    /// on-chain = true + address replacement passes combining and fails during fetch
+    /// because Vanilla has no on-chain packages registered.
     #[test(tokio::test)]
     async fn on_chain_with_address_replacement_passes_combining() {
         let scenario = TestPackageGraph::new(["root"])
@@ -298,7 +298,7 @@ mod tests {
 
         let err = scenario.root_package_err("root").await;
         assert!(
-            err.contains("On-chain dependencies are not yet supported"),
+            err.contains("on-chain package not found"),
             "expected fetch error, got: {err}"
         );
     }
--- a/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
+++ b/external-crates/move/crates/move-package-alt/src/dependency/combine.rs
@@ -148,13 +148,15 @@ impl CombinedDependency {
         }
 
         // On-chain deps use a fixed environment; reject explicit use-environment
-        let use_environment = if matches!(&dep.dependency_info, ManifestDependencyInfo::OnChain(_))
-        {
-            if replacement.use_environment.is_some() {
-                return Err(ManifestError::with_file(file)(
-                    ManifestErrorKind::OnChainWithUseEnvironment { name },
-                ));
-            }
+        let is_on_chain = matches!(&dep.dependency_info, ManifestDependencyInfo::OnChain(_));
+
+        if is_on_chain && replacement.use_environment.is_some() {
+            return Err(ManifestError::with_file(file)(
+                ManifestErrorKind::OnChainWithUseEnvironment { name },
+            ));
+        }
+
+        let use_environment = if is_on_chain {
             crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
         } else {
             replacement.use_environment.unwrap_or(source_env_name)
@@ -198,13 +200,15 @@ impl CombinedDependency {
         }
 
         // On-chain deps use a fixed environment; reject explicit use-environment
-        let use_environment = if matches!(&dep.dependency_info, ManifestDependencyInfo::OnChain(_))
-        {
-            if replacement.use_environment.is_some() {
-                return Err(ManifestError::with_file(file)(
-                    ManifestErrorKind::OnChainWithUseEnvironment { name },
-                ));
-            }
+        let is_on_chain = matches!(&dep.dependency_info, ManifestDependencyInfo::OnChain(_));
+
+        if is_on_chain && replacement.use_environment.is_some() {
+            return Err(ManifestError::with_file(file)(
+                ManifestErrorKind::OnChainWithUseEnvironment { name },
+            ));
+        }
+
+        let use_environment = if is_on_chain {
             crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
         } else {
             replacement.use_environment.unwrap_or(source_env_name)
--- a/external-crates/move/crates/move-package-alt/src/flavor/mod.rs
+++ b/external-crates/move/crates/move-package-alt/src/flavor/mod.rs
@@ -13,9 +13,23 @@ use serde::{Serialize, de::DeserializeOwned};
 
 use crate::schema::{
     EnvironmentID, EnvironmentName, LockfileDependencyInfo, OriginalID, PackageName,
-    ParsedManifest, ReplacementDependency, SystemDepName,
+    ParsedManifest, PublishedID, ReplacementDependency, SystemDepName,
 };
 use indexmap::IndexMap;
+
+/// Data returned by [`MoveFlavor::fetch_onchain_package`] representing a package fetched from the
+/// network.
+#[derive(Clone, Debug)]
+pub struct OnChainPackageData {
+    /// Module name → compiled bytecode
+    pub modules: BTreeMap<String, Vec<u8>>,
+    /// Original dependency ID → current linked address
+    pub dependencies: BTreeMap<OriginalID, PublishedID>,
+    /// The original (runtime) ID of this package
+    pub original_id: OriginalID,
+    /// The on-chain version of this package
+    pub version: u64,
+}
 
 /// A [MoveFlavor] is used to parameterize the package management system. It defines the types and
 /// methods for package management that are specific to a particular instantiation of the Move
@@ -66,4 +80,11 @@ pub trait MoveFlavor: Debug + Send + Sync {
     /// addresses are not dropped when substituting ephemeral addresses (they can still be
     /// overridden)
     fn is_system_address(&self, address: &OriginalID) -> bool;
+
+    /// Fetch the on-chain package at `address` from the network. Returns the module bytecode,
+    /// linkage table (as original → linked address), original ID, and version.
+    async fn fetch_onchain_package(
+        &self,
+        address: &PublishedID,
+    ) -> anyhow::Result<OnChainPackageData>;
 }
--- a/external-crates/move/crates/move-package-alt/src/flavor/mod.rs
+++ b/external-crates/move/crates/move-package-alt/src/flavor/mod.rs
@@ -81,8 +81,8 @@ pub trait MoveFlavor: Debug + Send + Sync {
     /// overridden)
     fn is_system_address(&self, address: &OriginalID) -> bool;
 
-    /// Fetch the on-chain package at `address` from the network. Returns the module bytecode,
-    /// linkage table (as original → linked address), original ID, and version.
+    /// Fetch the on-chain package at `address` from the network, returning an
+    /// [`OnChainPackageData`].
     async fn fetch_onchain_package(
         &self,
         address: &PublishedID,
--- a/external-crates/move/crates/move-package-alt/src/flavor/mod.rs
+++ b/external-crates/move/crates/move-package-alt/src/flavor/mod.rs
@@ -21,7 +21,7 @@ use indexmap::IndexMap;
 /// network.
 #[derive(Clone, Debug)]
 pub struct OnChainPackageData {
-    /// Module name → compiled bytecode
+    /// Module name → serialized `CompiledModule` bytecode (the `.mv` format)
     pub modules: BTreeMap<String, Vec<u8>>,
     /// Original dependency ID → current linked address
     pub dependencies: BTreeMap<OriginalID, PublishedID>,
--- a/external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
+++ b/external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
@@ -9,14 +9,16 @@ use std::collections::BTreeMap;
 
 use serde::{Deserialize, Serialize};
 
+use anyhow::anyhow;
+
 use crate::schema::{
     Environment, EnvironmentID, EnvironmentName, LockfileDependencyInfo, OriginalID, PackageName,
-    ParsedManifest, ReplacementDependency,
+    ParsedManifest, PublishedID, ReplacementDependency,
 };
 
 use async_trait::async_trait;
 
-use super::MoveFlavor;
+use super::{MoveFlavor, OnChainPackageData};
 use indexmap::IndexMap;
 
 pub const DEFAULT_ENV_NAME: &str = "_test_env";
@@ -24,9 +26,13 @@ pub const DEFAULT_ENV_ID: &str = "_test_env_id";
 
 /// The [Vanilla] implementation of the [MoveFlavor] trait. This implementation supports no
 /// flavor-specific resolvers and stores no additional metadata in the lockfile.
+///
+/// On-chain packages can be pre-populated via [`Vanilla::with_on_chain_package`] for testing.
 #[derive(Debug, Default)]
 #[non_exhaustive]
-pub struct Vanilla;
+pub struct Vanilla {
+    on_chain_packages: BTreeMap<PublishedID, OnChainPackageData>,
+}
 
 #[derive(Serialize, Deserialize, Clone, Debug, Default)]
 #[serde(rename_all = "kebab-case")]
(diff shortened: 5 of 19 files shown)


crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -230,6 +230,14 @@ impl MoveFlavor for SuiFlavor {
     fn is_system_address(&self, address: &move_package_alt::schema::OriginalID) -> bool {
         is_system_package(address.0)
     }
+
+    async fn fetch_onchain_package(
+        &self,
+        _address: &move_package_alt::schema::PublishedID,
+    ) -> anyhow::Result<move_package_alt::OnChainPackageData> {
+        // TODO: implement via gRPC get_object → MovePackage
+        anyhow::bail!("on-chain package fetching is not yet implemented for SuiFlavor")
```
mdgeorge4153: I think we should implement this as part of this PR.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -298,7 +298,7 @@ mod tests {
 
         let err = scenario.root_package_err("root").await;
         assert!(
-            err.contains("On-chain dependencies are not yet supported"),
+            err.contains("on-chain package not found"),
```
mdgeorge4153: We should change this to snapshot the error message.

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
```diff
@@ -13,10 +13,24 @@ use serde::{Serialize, de::DeserializeOwned};
...
-    ParsedManifest, ReplacementDependency, SystemDepName,
+    ParsedManifest, PublishedID, ReplacementDependency, SystemDepName,
 };
 use indexmap::IndexMap;
 
+/// Data returned by [`MoveFlavor::fetch_onchain_package`] representing a package fetched from the
+/// network.
+#[derive(Clone, Debug)]
+pub struct OnChainPackageData {
+    /// Module name → compiled bytecode
```
mdgeorge4153: I think we may need some more clarity around the format here, unless there's only one obvious format for "compiled bytecode". I'm not familiar enough with the move internals to know.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+/// to the generated package directory in the cache.
+pub(crate) async fn fetch_onchain<F: MoveFlavor>(
+    address: &PublishedID,
+    env: &Environment,
+    config: &PackageConfig<F>,
+) -> anyhow::Result<PathBuf> {
+    let cache_dir = cache_dir_for(&config.chain_id, address);
+
+    let _lock = PackageSystemLock::new_for_onchain(&config.chain_id, address)
+        .map_err(|e| anyhow!("failed to acquire lock for on-chain package {address}: {e}"))?;
```
mdgeorge4153: Most of the package system isn't using anyhow errors, we should be consistent about the error style I think.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+fn cache_dir_for(chain_id: &str, address: &PublishedID) -> PathBuf {
+    PathBuf::from(move_command_line_common::env::MOVE_HOME.as_str())
+        .join("on-chain")
+        .join(chain_id)
+        .join(address.to_string())
+}
+
+/// Write module bytecode files to `<cache_dir>/bytecode/<name>.mv`.
+fn write_bytecode(cache_dir: &Path, modules: &BTreeMap<String, Vec<u8>>) -> anyhow::Result<()> {
+    let bytecode_dir = cache_dir.join("bytecode");
```
mdgeorge4153, line 143: One thing I'm not sure about - does the move compiler already have a standard location where it outputs bytecode? If it does, maybe we should use it. On the other hand, maybe we should actually /avoid/ it, since we could end up accidentally clobbering the original bytecode by recompiling.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+    address: &PublishedID,
+    env: &Environment,
+    config: &PackageConfig<F>,
+) -> anyhow::Result<PathBuf> {
+    let cache_dir = cache_dir_for(&config.chain_id, address);
+
+    let _lock = PackageSystemLock::new_for_onchain(&config.chain_id, address)
+        .map_err(|e| anyhow!("failed to acquire lock for on-chain package {address}: {e}"))?;
+
+    let data = config.flavor.fetch_onchain_package(address).await?;
```
mdgeorge4153: We should only fetch if we don't already have the bytecode, right?

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+    for (name, bytes) in modules {
+        let path = bytecode_dir.join(format!("{name}.mv"));
+        fs::write(&path, bytes)
+            .with_context(|| format!("writing module bytecode: {}", path.display()))?;
+    }
+
+    Ok(())
+}
+
+/// Generate or update `Move.toml` in the cache directory. Adds the environment and
```
mdgeorge4153: We should clarify that we're updating existing manifests to account for multiple environments, likewise for update_pubfile.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+/// dep-replacements for on-chain dependencies from the linkage table.
```
mdgeorge4153: We should double check that doing this update won't cause repinning for other environments (I think it shouldn't, but maybe updating the environments table does?). It's possible we may want to skip repinning for on-chain packages, wdyt?

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+fn update_manifest(
+    cache_dir: &Path,
+    env: &Environment,
+    dependencies: &BTreeMap<OriginalID, PublishedID>,
+) -> anyhow::Result<()> {
+    let manifest_path = cache_dir.join("Move.toml");
```
mdgeorge4153: I think this is fine, but if we have this string as a constant maybe we should reuse the constant. We could even consider using the PackagePath infrastructure for this, but I'm not sure how much it buys us.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+        fs::write(&path, bytes)
+            .with_context(|| format!("writing module bytecode: {}", path.display()))?;
+    }
+
+    Ok(())
+}
+
+/// Generate or update `Move.toml` in the cache directory. Adds the environment and
+/// dep-replacements for on-chain dependencies from the linkage table.
+fn update_manifest(
```
mdgeorge4153: I wonder if a short example would be useful in the doc comment? We should at least call out that we're naming packages `self`, which is a potentially surprising detail.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+        package.insert("name", value("self"));
+        doc.insert("package", Item::Table(package));
+        doc
+    };
+
+    // Add environment entry
+    if !doc.contains_key("environments") {
+        doc.insert("environments", Item::Table(Table::new()));
+    }
+    doc["environments"][env.name()] = value(env.id());
```
mdgeorge4153: This will fail if the environment is already implicit I think.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+            let dep_name = format!("onchain_{original_id}");
+            let mut dep_table = Table::new();
+            dep_table.insert("on-chain", value(linked_address.to_string()));
+            env_table.insert(&dep_name, Item::Table(dep_table));
+        }
+        replacements[env.name()] = Item::Table(env_table);
+    }
+
+    fs::write(&manifest_path, doc.to_string())
+        .with_context(|| format!("writing {}", manifest_path.display()))?;
```
mdgeorge4153: It would be better to use our existing schema data structures and then calling render_toml. instead of manually building the toml here. I think that should help with the fact that this method is getting pretty long.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    Ok(())
+}
+
+/// Generate or update `Published.toml` in the cache directory.
+fn update_published<F: MoveFlavor>(
```
mdgeorge4153: Most of the comments on update_manifest should apply here as well.

external-crates/move/crates/move-package-alt/src/package/root_package.rs
```diff
@@ -1645,7 +1655,7 @@ pkg_b = { local = "../pkg_b" }"#,
         let root = PackageLoader::new(
             scenario.path_for("root"),
             Vanilla::default_environment(),
-            Vanilla,
+            Vanilla::new(),
```
mdgeorge4153, line 1659: Maybe we should pull all the Vanilla::new stuff out into its own base PR to make reviewing a little eaiser.

mdgeorge4153: [CLAUDE] Good idea. I'll extract the Vanilla struct change + all the `Vanilla::new()` call site updates into a separate base PR to reduce noise in this one.

crates/sui-package-alt/src/sui_flavor.rs
mdgeorge4153: [CLAUDE] Agreed, will implement this.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Will switch to `assert_snapshot!`.

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
mdgeorge4153: [CLAUDE] Good point. Each entry in `modules` is a serialized `CompiledModule` — the same format used for `.mv` files. I'll clarify this in the doc comment.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Will switch to a proper error type consistent with the rest of the package system. I'll add variants to `FetchError` or create a dedicated error type in the `on_chain` module.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] The compiler outputs to `build/<pkg>/bytecode_modules/<name>.mv`. I think using a different location (`bytecode/`) is better to avoid accidentally clobbering, as you noted. I'll add a comment explaining the choice.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Done — `fetch_onchain` now checks if the bytecode directory has files and skips the network fetch if so. Still updates manifest/Published.toml if the current environment is missing. See `on_chain/fetch.rs:71-85`.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Updated doc comments on both `update_manifest` and `update_published` to clarify the additive behavior. See module-level doc at `on_chain/fetch.rs:4-13` and function-level docs at lines 120-122 and 228-229.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Verified: the manifest digest is computed per-environment from `CombinedDependency` (see `package_impl.rs:279-300`), so adding a new environment'\''s dep-replacements doesn'\''t change the digest for existing environments. Added a note in the `fetch_onchain` doc comment at `on_chain/fetch.rs:56-59`. A test for this is planned as part of the broader test suite.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Now using `SourcePackageLayout::Manifest.location_str()` for the `"Move.toml"` string. See `on_chain/fetch.rs:129`.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Added a TOML example in the `fetch_onchain` doc comment showing the generated manifest structure, and the `ON_CHAIN_PACKAGE_NAME` constant makes the `"self"` choice explicit. See `on_chain/fetch.rs:44-53` and line 38.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Good catch. Fixed by only adding the environment to the manifest if it'\''s not already present (line 150-157). Since the loading pipeline merges flavor defaults separately in `check_for_environment`, default environments don'\''t need to appear in the generated manifest at all — they'\''re automatically known.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Refactored to use `ParsedManifest` with `Spanned::new(0..1, val)` for generated values, then serialize via `toml_edit::ser::to_document`. No `RenderToml` impl added — `to_document` is sufficient for generated manifests. See `on_chain/fetch.rs:123-205` and the `new_manifest()` helper at lines 208-224.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Applied same improvements: skip if environment already present (line 237), clarified additive behavior in doc comment (line 228-229), proper error types throughout.

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -230,6 +230,47 @@ impl MoveFlavor for SuiFlavor {
...
     }
+
+    async fn fetch_onchain_package(
+        &self,
+        address: &move_package_alt::schema::PublishedID,
+    ) -> anyhow::Result<move_package_alt::OnChainPackageData> {
+        let Some(ref client) = self.client else {
+            anyhow::bail!(
+                "cannot fetch on-chain package {address}: no network connection \
+                 (SuiFlavor created in offline mode)"
```
mdgeorge4153: The `SuiFlavor` should not be included in the user-facing error message.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -297,9 +297,11 @@ mod tests {
...
         let err = scenario.root_package_err("root").await;
-        assert!(
-            err.contains("On-chain dependencies are not yet supported"),
-            "expected fetch error, got: {err}"
+        // Redact the MOVE_HOME path which varies per user
+        let err = err.replace(
+            move_command_line_common::env::MOVE_HOME.as_str(),
+            "<MOVE_HOME>",
         );
+        assert_snapshot!(err, @"Error while loading dependency <MOVE_HOME>/on-chain/_test_env_id/0x0000000000000000000000000000000000000000000000000000000000000001: failed to fetch on-chain package at 0x0000000000000000000000000000000000000000000000000000000000000001: on-chain package not found: 0x0000000000000000000000000000000000000000000000000000000000000001");
```
mdgeorge4153: the address being included 3 times in this message is pretty ugly.

external-crates/move/crates/move-package-alt/src/on_chain/errors.rs
```diff
@@ -0,0 +1,47 @@
...
+    },
+
+    #[error("failed to write on-chain package cache at {path}: {source}")]
+    CacheWriteFailed {
+        path: PathBuf,
+        #[source]
+        source: std::io::Error,
+    },
+
+    #[error("failed to parse cached manifest at {path}: {source}")]
```
mdgeorge4153: I wonder if this should be an internal error; really only we should be writing to the cache (although in principle the user could, but I don't mind them getting a bad error message in that case).

external-crates/move/crates/move-package-alt/src/on_chain/errors.rs
```diff
@@ -0,0 +1,47 @@
...
+    ManifestParseFailed {
+        path: PathBuf,
+        #[source]
+        source: toml_edit::de::Error,
+    },
+
+    #[error("failed to serialize manifest for on-chain package {address}: {source}")]
+    ManifestSerializeFailed {
```
mdgeorge4153: Likewise, I think expects are reasonable for serialization errors, especially for files we're writing

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+
+//! Fetches on-chain packages and generates synthetic manifests and publication files.
+//!
+//! On-chain packages are cached at `~/.move/on-chain/<chain_id>/<address>/`. The cache
+//! contains:
+//! - `bytecode/<module_name>.mv` — serialized `CompiledModule` bytecode (the same format
+//!   the compiler writes to `build/<pkg>/bytecode_modules/`; we use a separate directory
+//!   to avoid clobbering the original bytecode if the package is recompiled)
+//! - `Move.toml` — a generated manifest with `name = "self"`, with environments and
+//!   dep-replacements added incrementally as the package is used from different contexts
```
mdgeorge4153:
```suggestion
//!   dep-replacements added incrementally as the package is used in different environments
```

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+/// in the cache directory. If the bytecode is already cached, skips the network fetch but
+/// still ensures the manifest and Published.toml include the current environment.
+///
+/// The generated `Move.toml` looks like:
+/// ```toml
+/// [package]
+/// name = "self"
+///
+/// [environments]
+/// mainnet = "35834a8a"
```
mdgeorge4153: I'd not include `mainnet` here, since for sui that's a default and would be explicitly _not_ included

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+        let mut env_replacements = BTreeMap::new();
+        for (original_id, linked_address) in &data.dependencies {
+            let dep_name =
+                PackageName::new(format!("onchain_{original_id}")).expect("valid identifier");
+            let replacement = ReplacementDependency {
+                dependency: Some(DefaultDependency {
+                    dependency_info: ManifestDependencyInfo::OnChainAt(OnChainAddress {
+                        on_chain: linked_address.clone(),
+                    }),
+                    is_override: false,
```
mdgeorge4153: I think we actually want is_override: true here

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+                    rename_from: None,
+                    modes: None,
+                }),
+                addresses: None,
+                use_environment: None,
+            };
+            env_replacements.insert(dep_name, Spanned::new(0..1, replacement));
+        }
```
mdgeorge4153: I'm tempted to replace this whole block with parsing
```
format!(r#"onchain_{chain_id} = { on-chain = "{address}", override = true }"#)
```
just because it replaces a dozen lines with like three. it also feels a little ugly. wdyt?

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153, line 1: There's a huge issue to consider here, which is system packages. On sui at least, all non-system packages are immutable, but system packages are mutable. There's a few options to consider: detecting mutable packages and replacing them with system dependencies might be the best workaround.

mdgeorge4153: related, and probably part of the same problem/solution: we need to not generate things that conflict with the implicit dependencies (probably just making implicit deps false is the way to go here)

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+    })?;
+    fs::write(&manifest_path, doc.to_string()).map_err(|source| {
+        OnChainError::CacheWriteFailed {
+            path: manifest_path,
+            source,
+        }
+    })?;
+
+    Ok(())
+}
```
mdgeorge4153, line 208: The above function feels too long. Hopefully internalizing errors will help, but we may also want some helpers for constructing or rendering manifests.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+) -> OnChainResult<()> {
+    let pub_path = cache_dir.join("Published.toml");
+
+    let mut pubfile: ParsedPublishedFile<F> = if pub_path.exists() {
+        let content =
+            fs::read_to_string(&pub_path).map_err(|source| OnChainError::CacheWriteFailed {
+                path: pub_path.clone(),
+                source,
+            })?;
+        toml_edit::de::from_str(&content).map_err(|source| OnChainError::ManifestParseFailed {
```
mdgeorge4153: Another point in favor of `expect` here: we're not parsing a manifest.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+
+/// Generate or update `Published.toml` in the cache directory. Adds a publication entry
+/// for the current environment if not already present.
+fn update_published<F: MoveFlavor>(
+    cache_dir: &Path,
+    address: &PublishedID,
+    env: &Environment,
+    data: Option<&OnChainPackageData>,
```
mdgeorge4153: why an option here?

mdgeorge4153:
PR description needs updating.

crates/sui-package-alt/src/sui_flavor.rs
mdgeorge4153: [CLAUDE] Removed "SuiFlavor created in offline mode" from the error message. Now just says "cannot fetch on-chain package {address}: no network connection". See `sui_flavor.rs:240`.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Simplified `OnChainError::Fetch` to just forward the inner error message (`"{source}"` instead of `"failed to fetch on-chain package at {address}: {source}"`). The address still appears in the cache path and the inner error, but no longer in the middle layer. Updated snapshot at `combine.rs:305`.

external-crates/move/crates/move-package-alt/src/on_chain/errors.rs
mdgeorge4153: [CLAUDE] Agreed — switched to `expect` for all cache write operations. Removed `CacheWrite` from `OnChainError`. See `on_chain/fetch.rs:105,109,144-145,151,154`.

external-crates/move/crates/move-package-alt/src/on_chain/errors.rs
mdgeorge4153: [CLAUDE] Done — using `expect` for serialization and parsing of our own generated files. `OnChainError` now only has `Fetch` and `Lock` variants. See `on_chain/errors.rs:13-23`.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Fixed — "different environments" instead of "different contexts". See `on_chain/fetch.rs:12`.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Good catch — changed to `my_env = "some_chain_id"` since mainnet is a Sui default and would not appear in the generated manifest. See `on_chain/fetch.rs:50`.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Fixed — `is_override: true` in `on_chain_replacement()`. See `on_chain/fetch.rs:183`.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Extracted `on_chain_replacement(address) -> ReplacementDependency` helper and `dep_replacements_for()` to build the full map. Keeps type safety but gets the construction out of the main flow. See `on_chain/fetch.rs:160-189`.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Shortened from ~80 to ~30 lines by: expects instead of error mapping, `load_or_create_manifest()` helper, `dep_replacements_for()` helper. See `on_chain/fetch.rs:116-146`.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Done — using `expect` for both reading and parsing Published.toml. See `on_chain/fetch.rs:221-224`.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] `data` is `None` when bytecode was already cached (we skipped the network fetch). In that case we still check if the environment needs to be added to the manifest/Published.toml, but we don'\''t have linkage table data for dep-replacements or publication info. If the env is already present, we return early; if not and data is None, we add the environment but skip dep-replacements/publication.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Added a doc comment and TODO at `on_chain/fetch.rs:62-73` documenting this limitation. The linkage table entries for system packages currently become on-chain dep-replacements but should be deduplicated with source deps (system or user-provided). This needs a post-processing step after the full graph is built. Landing as-is with the known limitation called out.

mdgeorge4153: [CLAUDE] Already handled — `new_manifest()` sets `implicit_dependencies: false` at `on_chain/fetch.rs:201`. The system package deduplication (previous comment) is the remaining piece.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,266 @@
...
+fn update_manifest(cache_dir: &Path, env: &Environment, data: &Option<OnChainPackageData>) {
+    let manifest_path = cache_dir.join(SourcePackageLayout::Manifest.location_str());
+
+    let mut manifest = load_or_create_manifest(cache_dir, &manifest_path);
+
+    // Skip if this environment is already present
+    if manifest
+        .environments
+        .keys()
+        .any(|k| k.as_ref() == env.name())
```
mdgeorge4153: can this just be a lookup?

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,266 @@
...
+    {
```
mdgeorge4153: Ah, just realized this is not going to work. Environment names are only sensible in the context of a single package, so different packages may use different environment names for the same network, or the same environment name for different environments. We need to rethink how we're handling environments here.

mdgeorge4153: for a given chain_id and address, we don't really care about environments. So there should really only be one entry in the environments table, which we could call `on-chain` or something. We'd need to ensure that we set the on-chain deps to always use that environment. The nice thing is that that means we don't have to keep editing the cache.

mdgeorge4153 (commented):
oops we need to do some more design work.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Reworked — the environment check is now moot since we use a fixed env name and write the cache once. The manifest existence check at `fetch.rs:88` serves as the cache hit test. No iteration needed.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Reworked — generated manifests now use a fixed environment name `_on_chain` with the chain_id as value. `use_environment` is forced to `_on_chain` during combining for `OnChainAt` deps (see `combine.rs:150-161`). No more additive environment updates.

mdgeorge4153: [CLAUDE] Implemented exactly this approach. Generated manifests have `[environments] _on_chain = "<chain_id>"`. The cache is written once and never edited. Explicit `use-environment` on on-chain deps is rejected with a new `OnChainWithUseEnvironment` error. See `on_chain/fetch.rs:49-50` for the constant and `combine.rs:150-161` for the enforcement.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: this is no longer a concern for the new design

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,427 @@
...
+    let data = config
+        .flavor
+        .fetch_onchain_package(address)
+        .await
+        .map_err(|source| OnChainError::Fetch {
+            address: address.clone(),
+            source,
+        })?;
+
+    write_bytecode(&cache_dir, &data.modules);
```
mdgeorge4153: with the new design, saving the bytecode isn't strictly necessary, but probably worth doing anyway.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,427 @@
...
+    write_manifest(&cache_dir, &manifest_path, &config.chain_id, &data);
+    write_published::<F>(&cache_dir, address, &config.chain_id, &data);
+
```
mdgeorge4153, line 113: We may want to add a dummy `write_source` here that we'll fill in when we implement stub generation.

external-crates/move/crates/move-package-alt/src/package/package_loader.rs
```diff
@@ -103,6 +107,7 @@ impl<F: MoveFlavor> PackageLoader<F> {
             ignore_digests: false,
             allow_dirty: false,
             flavor: flavor.into(),
+            move_home: PathBuf::from(move_command_line_common::env::MOVE_HOME.as_str()),
```
mdgeorge4153: we've added `move_home` here but we seem to only be using it for on-chain deps; we should probably also use it for git deps as well (and file locks, and anything else that's currently landing in ~/.move). I would factor these changes out into their own base PR.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,427 @@
...
+            .build();
+
+        let root = scenario.root_package("root").await;
+        assert!(root.packages().len() >= 3, "expected root + A + B");
+    }
+
+    /// When an on-chain dep's linkage table references the same address as a local dep,
+    /// they should be deduplicated to a single package in the graph.
+    #[test(tokio::test)]
+    #[ignore] // TODO: requires deduplication of on-chain and source deps
```
mdgeorge4153: We need to fix this before landing this PR I think

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Correct — the repinning discussion was removed from the doc comment since the fixed `_on_chain` environment means the manifest is written once and never modified.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Acknowledged — keeping bytecode write as-is since it will be useful for future stub generation.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Added `write_source()` stub that will be filled in when stub generation is integrated. See `on_chain/fetch.rs:130-132`.

external-crates/move/crates/move-package-alt/src/package/package_loader.rs
mdgeorge4153: [CLAUDE] Good point — `move_home` should be used consistently for git deps and locks too. I'll extract this into a separate base PR that threads `move_home` through the git cache and lock infrastructure as well.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Fixed — removed `#[ignore]`, test now asserts current behavior (error due to duplicate packages). Updated to use `try_root_package` and assert the error. See `on_chain/fetch.rs:394-414`.

mdgeorge4153: I meant we should fix the behavior rather than modifying the test; but I think actually punting it like we were with `#[ignore]` is maybe fine.

external-crates/move/crates/move-package-alt/src/package/package_loader.rs
mdgeorge4153: we spun off issue dvx-2127 for this, and for this PR just made the addresses in the tests unique

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,440 @@
...
+        assert_snapshot!(fs::read_to_string(&manifest_path).unwrap(), @r#"
+        package = { name = "self", implicit-dependencies = false }
+        environments = { _on_chain = "test_chain" }
+        dependencies = {}
+
+        [dep-replacements]
+
+        [dep-replacements._on_chain]
+
+        [dep-replacements._on_chain.onchain_0x0000000000000000000000000000000000000000000000000000000000000002]
```
mdgeorge4153: Ah, I would prefer a different style here - I would expect
```toml
[dep-replacements._on_chain]
onchain_0x0000....02 = { on-chain = "0x00000....002", override = true }
onchain_0x0000....03 = { on-chain = "0x00000....003", override = true }
```
just because that's how we usually write manifests.

mdgeorge4153 (commented):
one more nit

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Fixed — now using `toml_edit::InlineTable` so dep-replacements render as inline tables: `onchain_0x... = { on-chain = "0x...", override = true }`. See `on_chain/fetch.rs:159-170` and updated snapshot at line 298.

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -230,6 +230,44 @@ impl MoveFlavor for SuiFlavor {
...
+            .try_as_package()
+            .ok_or_else(|| anyhow::anyhow!("object {address} is not a package"))?;
+
+        let modules = package.serialized_module_map().clone();
+        let dependencies = package
+            .linkage_table()
+            .iter()
+            .map(|(original_id, upgrade_info)| {
+                (
+                    move_package_alt::schema::OriginalID((*original_id).into()),
```
stefan-mysten: nit: import these?

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -230,6 +230,44 @@ impl MoveFlavor for SuiFlavor {
     fn is_system_address(&self, address: &move_package_alt::schema::OriginalID) -> bool {
         is_system_package(address.0)
     }
+
+    async fn fetch_onchain_package(
```
stefan-mysten, line 236: Let's add a doc comment with a brief explanation of what's going on here.


```suggestion
    /// Fetches an onchain package by its address.
    ///
    /// Panics if there's no network connection; if the object is not a package, it will error.
    async fn fetch_onchain_package(
```

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -230,6 +230,44 @@ impl MoveFlavor for SuiFlavor {
...
+        let Some(ref client) = self.client else {
+            anyhow::bail!("cannot fetch on-chain package {address}: no network connection")
+        };
+
+        let object_id = ObjectID::from(address.0);
+        let object = client.clone().get_object(object_id).await?;
+        let package = object
+            .data
+            .try_as_package()
+            .ok_or_else(|| anyhow::anyhow!("object {address} is not a package"))?;
```
stefan-mysten:
```suggestion
            .try_as_package()
            .with_context(|| format!("object {address} is not a package"))?;
```

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
```diff
@@ -13,10 +13,24 @@ use serde::{Serialize, de::DeserializeOwned};
...
     EnvironmentID, EnvironmentName, LockfileDependencyInfo, OriginalID, PackageName,
-    ParsedManifest, ReplacementDependency, SystemDepName,
+    ParsedManifest, PublishedID, ReplacementDependency, SystemDepName,
 };
 use indexmap::IndexMap;
 
+/// Data returned by [`MoveFlavor::fetch_onchain_package`] representing a package fetched from the
+/// network.
+#[derive(Clone, Debug)]
+pub struct OnChainPackageData {
```
stefan-mysten, line 23: This looks very much like the MovePackage in `sui-types`. I wonder if we should keep them somewhat similar to make it easier to work with them, if needed. (Probably need to read below to understand how it's used on the sui side, maybe it's not at all).

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
```diff
@@ -66,4 +80,11 @@ pub trait MoveFlavor: Debug + Send + Sync {
     /// addresses are not dropped when substituting ephemeral addresses (they can still be
     /// overridden)
     fn is_system_address(&self, address: &OriginalID) -> bool;
+
+    /// Fetch the on-chain package at `address` from the network. Returns the module bytecode,
+    /// linkage table (as original → linked address), original ID, and version.
```
stefan-mysten: nit: given that this returns a type, I would just reference to that type instead of explaining to avoid this comment becoming stale at some point.

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
```diff
@@ -9,25 +9,28 @@ use std::collections::BTreeMap;
 
 use serde::{Deserialize, Serialize};
```
stefan-mysten, line 12: nit: no need for this space.

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
```diff
@@ -9,25 +9,28 @@ use std::collections::BTreeMap;
...
 use indexmap::IndexMap;
 
 pub const DEFAULT_ENV_NAME: &str = "_test_env";
 pub const DEFAULT_ENV_ID: &str = "_test_env_id";
 
 /// The [Vanilla] implementation of the [MoveFlavor] trait. This implementation supports no
 /// flavor-specific resolvers and stores no additional metadata in the lockfile.
-#[derive(Debug, Default)]
+///
+/// On-chain packages can be pre-populated via [`Vanilla::with_on_chain_package`] for testing.
```
stefan-mysten: Slightly confused here; I thought that Vanilla flavor does not have any registered packages.

stefan-mysten: If it's for testing, then let's name that function `with_on_chain_package_for_testing`, maybe?

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
```diff
@@ -80,14 +83,30 @@ impl MoveFlavor for Vanilla {
...
     }
+
+    async fn fetch_onchain_package(
+        &self,
+        address: &PublishedID,
+    ) -> anyhow::Result<OnChainPackageData> {
+        self.on_chain_packages
+            .get(address)
+            .cloned()
+            .ok_or_else(|| anyhow!("on-chain package not found: {address}"))
```
stefan-mysten:
```suggestion
            .with_context(format!("on-chain package not found: {address}"))
```

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
```diff
@@ -80,14 +83,30 @@ impl MoveFlavor for Vanilla {
...
 
 impl Vanilla {
-    /// Create a new `Vanilla` flavor.
+    /// Create a new `Vanilla` flavor with no on-chain packages.
     pub fn new() -> Self {
         Self::default()
     }
 
+    /// Add an on-chain package for testing. Returns `self` for chaining.
+    pub fn with_on_chain_package(mut self, address: PublishedID, data: OnChainPackageData) -> Self {
```
stefan-mysten, line 105: nit: rename this for testing to indicate it should be used in testing scenarios rather than anytime.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,441 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Fetches on-chain packages and generates synthetic manifests and publication files.
+//!
+//! On-chain packages are cached at `~/.move/on-chain/<chain_id>/<address>/`. The cache
+//! contains:
+//! - `bytecode/<module_name>.mv` — serialized `CompiledModule` bytecode (the same format
+//!   the compiler writes to `build/<pkg>/bytecode_modules/`; we use a separate directory
+//!   to avoid clobbering the original bytecode if the package is recompiled)
+//! - `Move.toml` — a generated manifest with `name = "self"` and a single fixed
```
stefan-mysten: Can this badly interact with a third party package named `self`?

stefan-mysten: Just thinking aloud: how does this naming of this package work with error messages that have nice package naming?

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
```diff
@@ -0,0 +1,191 @@
...
+  user-authored Published.toml files.
+
+## Fetching pipeline
+
+When `fetch()` encounters a `Pinned::OnChain { address }`, it delegates to
+`on_chain::fetch::fetch_onchain`, which:
+
+1. Computes the cache directory from `MOVE_HOME`, chain ID, and address
+2. Acquires an exclusive filesystem lock for the cache directory
+3. Checks if the manifest already exists (cache hit → return immediately)
```
stefan-mysten, line 118: Is there a possibility that the manifest exists but no bytecode?

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Environment handling
+
+Environment names are per-package: different packages in the dependency graph may use
+different names for the same network. For example, one package might call mainnet
+`"mainnet"` while another calls it `"production"`.
+
+On-chain packages are tied to a specific chain (determined by the chain ID at fetch time),
+so they don't need environment flexibility. Generated manifests use a single fixed
+environment name `_on_chain` mapped to the chain ID.
```
stefan-mysten, line 146: What happens in the case of ephemeral networks like devnet. I assume that even if the package does not change, there will always be a new cache with the new chain id when devnet is deployed again.

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
```diff
@@ -0,0 +1,191 @@
...
+entirely. The cache is written atomically — bytecode, source stubs, manifest, and
+Published.toml are all written in a single locked section, with the manifest written last.
+
+## Generated manifest
+
+The generated `Move.toml` for an on-chain package looks like:
+
+```toml
+[package]
+name = "self"
```
stefan-mysten: I don't recall, but if another package is named `self`, the pkg system would be able to avoid name collision, right?

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
```diff
@@ -0,0 +1,191 @@
...
+
+```toml
+[package]
+name = "my_package"
+
+[dependencies]
+my_dep = { on-chain = true }
+
+[dep-replacements.mainnet]
+my_dep = { on-chain = "0x1234..." }
```
stefan-mysten, line 41: I don't fully understand why it's not acceptable to set the address in the dependency section. I assume it's to keep addresses all under dep-replacements to make it easier for implementation?

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
```diff
@@ -0,0 +1,191 @@
...
+generated from the bytecode so the compiler can type-check against the dependency.
+
+### Implementation status
+
+- Schema and validation: **done** (#26552)
+- Fetching and manifest generation: **done** (#26561)
+- Stub source generation: DVX-2119
+- System package deduplication: DVX-2126
+- Support for `sui move test-publish`: DVX-2120
+- Support for `sui move test`: DVX-2060
```
stefan-mysten: As in my previous comment on some other PR, I would keep todos outside of design doc, but up to you.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
```diff
@@ -184,11 +197,24 @@ impl CombinedDependency {
...
+            if matches!(&dep.dependency_info, ManifestDependencyInfo::OnChainAt(_)) {
+                if replacement.use_environment.is_some() {
+                    return Err(ManifestError::with_file(file)(
+                        ManifestErrorKind::OnChainWithUseEnvironment { name },
+                    ));
+                }
+                crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
+            } else {
+                replacement.use_environment.unwrap_or(source_env_name)
+            };
```
stefan-mysten: An alternative that does not nest if statements.

```suggestion
let is_on_chain_at = matches!(&dep.dependency_info, ManifestDependencyInfo::OnChainAt(_));

if is_on_chain_at && replacement.use_environment.is_some() {
    return Err(ManifestError::with_file(file)(
        ManifestErrorKind::OnChainWithUseEnvironment { name },
    ));
}

let use_environment = if is_on_chain_at {
    crate::on_chain::fetch::ON_CHAIN_ENV_NAME.to_string()
} else {
    replacement.use_environment.unwrap_or(source_env_name)
};
```

stefan-mysten (commented):
Left a few comments. It's a big PR so I need to look at it again.

stefan-mysten:
Another question: the compiler does not consume bytecode directory, so the fetched packages does not make it into the compiler. I am not sure if this will go away with the `write_source` stuff which will create stubs, I assume so but wanted to clear my understanding there.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,441 @@
...
+) -> OnChainResult<PathBuf> {
+    let cache_dir = cache_dir_for(&config.chain_id, address);
+    let manifest_path = cache_dir.join(SourcePackageLayout::Manifest.location_str());
+
+    let _lock = PackageSystemLock::new_for_onchain(&config.chain_id, address)?;
+
+    // Skip if the cache is already populated
+    if manifest_path.exists() {
+        return Ok(cache_dir);
+    }
```
stefan-mysten: This might not be enough to determine that the cache is **correctly** populated. Do we need to worry about that here? What happens if the disk is full, or process terminates, but the cache is not fully populated.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+/// Write serialized `CompiledModule` bytecode to `<cache_dir>/bytecode/<name>.mv`.
+fn write_bytecode(cache_dir: &Path, modules: &BTreeMap<String, Vec<u8>>) {
+    let bytecode_dir = cache_dir.join("bytecode");
+    fs::create_dir_all(&bytecode_dir).expect("can create on-chain package cache directory");
+
+    for (name, bytes) in modules {
+        let path = bytecode_dir.join(format!("{name}.mv"));
+        fs::write(&path, bytes).expect("can write module bytecode to cache");
```
stefan-mysten: Maybe the code shouldn't panic here and instead thread these IO errors through to log/print the paths as needed?

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,441 @@
...
+
+/// Generate stub Move source files from bytecode. Currently a no-op; will be implemented
+/// when stub generation (PR #26555) is integrated.
+// TODO: implement stub source generation from bytecode
+fn write_source(_cache_dir: &Path) {}
+
+/// Generate the `Move.toml` for an on-chain package.
+fn write_manifest(
+    cache_dir: &Path,
+    manifest_path: &Path,
```
stefan-mysten: Do we need to pass this argument? Do we ever need to pass custom manifest paths instead of a default one on the root of the package?

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+/// Generate stub Move source files from bytecode. Currently a no-op; will be implemented
+/// when stub generation (PR #26555) is integrated.
+// TODO: implement stub source generation from bytecode
+fn write_source(_cache_dir: &Path) {}
+
+/// Generate the `Move.toml` for an on-chain package.
+fn write_manifest(
+    cache_dir: &Path,
```
stefan-mysten, line 162: is cache_dir the same as package dir for the bytecode?

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,441 @@
...
+                PackageName::new(ON_CHAIN_PACKAGE_NAME).expect("valid identifier"),
+            ),
+            edition: None,
+            implicit_dependencies: false,
+            unrecognized_fields: BTreeMap::new(),
+        },
+        environments: BTreeMap::new(),
+        dependencies: BTreeMap::new(),
+        dep_replacements: BTreeMap::new(),
+        legacy_data: None,
```
stefan-mysten, line 225:
```suggestion
        ..Default::default()
```

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,441 @@
...
+fn new_manifest() -> ParsedManifest {
+    ParsedManifest {
+        package: crate::schema::PackageMetadata {
+            name: Spanned::new(
+                0..1,
+                PackageName::new(ON_CHAIN_PACKAGE_NAME).expect("valid identifier"),
+            ),
+            edition: None,
+            implicit_dependencies: false,
+            unrecognized_fields: BTreeMap::new(),
```
stefan-mysten, line 220:
```suggestion
            ..Default::default()
```

external-crates/move/crates/move-package-alt/src/test_utils/graph_builder.rs
```diff
@@ -164,6 +166,51 @@ pub struct DepSpec {
...
+    }
+
+    /// Add a linkage table entry: this on-chain package depends on `linked` at `original`.
+    pub fn dep(mut self, original: OriginalID, linked: PublishedID) -> Self {
+        self.dependencies.insert(original, linked);
+        self
+    }
+
+    /// Set the original (runtime) ID if it differs from the address.
+    pub fn original_id(mut self, id: OriginalID) -> Self {
```
stefan-mysten, line 202: nit: function name does not indicate that this will be a mutation. Consider renaming this

```suggestion
    pub fn set_original_id(mut self, id: OriginalID) -> Self {
```

external-crates/move/crates/move-package-alt/src/test_utils/graph_builder.rs
```diff
@@ -164,6 +166,51 @@ pub struct DepSpec {
...
+        self.original_id = id;
+        self
+    }
+
+    /// Set the on-chain version.
+    pub fn version(mut self, version: u64) -> Self {
```
stefan-mysten, line 208: ^^

external-crates/move/crates/move-package-alt/src/test_utils/graph_builder.rs
```diff
@@ -164,6 +166,51 @@ pub struct DepSpec {
...
+            original_id: OriginalID(address.0),
+            address,
+            modules: BTreeMap::new(),
+            dependencies: BTreeMap::new(),
+            version: 1,
+        }
+    }
+
+    /// Add a module by loading bytecode from a `.mv` file in `tests/data/`.
+    pub fn module(self, _name: &str, _mv_file: &str) -> Self {
```
stefan-mysten, line 191: nit: api naming
```suggestion
    pub fn add_module(self, _name: &str, _mv_file: &str) -> Self {
```

external-crates/move/crates/move-package-alt/src/test_utils/graph_builder.rs
```diff
@@ -164,6 +166,51 @@ pub struct DepSpec {
...
+        todo!("load .mv bytecode from tests/data/")
+    }
+
+    /// Add a linkage table entry: this on-chain package depends on `linked` at `original`.
+    pub fn dep(mut self, original: OriginalID, linked: PublishedID) -> Self {
```
stefan-mysten, line 196: ^^

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
```diff
@@ -173,6 +176,83 @@ impl ReplacementDependency {
...
+                table.insert("git", git.repo.clone().into());
+                if let Some(ref rev) = git.rev {
+                    table.insert("rev", rev.clone().into());
+                }
+            }
+            ManifestDependencyInfo::System(sys) => {
+                table.insert("system", sys.system.clone().into());
+            }
+            ManifestDependencyInfo::External(_) => {
+                // External deps have complex serialization; skip for now
```
stefan-mysten, line 238: Add a log / a TODO to not forget about this.

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
```diff
@@ -173,6 +176,83 @@ impl ReplacementDependency {
...
+        match &d.dependency_info {
+            ManifestDependencyInfo::OnChainAt(addr) => {
+                table.insert("on-chain", addr.on_chain.to_string().into());
+            }
+            ManifestDependencyInfo::OnChain(flag) => {
+                let val: bool = flag.on_chain.clone().into();
+                table.insert("on-chain", val.into());
+            }
+            ManifestDependencyInfo::Local(loc) => {
+                table.insert("local", loc.local.to_string_lossy().to_string().into());
```
stefan-mysten, line 226: This might work?
```suggestion
                table.insert("local", loc.local.to_string_lossy().into_owned();
```

external-crates/move/crates/move-package-alt/src/schema/manifest.rs
```diff
@@ -173,6 +176,83 @@ impl ReplacementDependency {
...
+    let mut table = toml_edit::InlineTable::new();
+
+    if let Some(ref d) = dep.dependency {
+        match &d.dependency_info {
+            ManifestDependencyInfo::OnChainAt(addr) => {
+                table.insert("on-chain", addr.on_chain.to_string().into());
+            }
+            ManifestDependencyInfo::OnChain(flag) => {
+                let val: bool = flag.on_chain.clone().into();
+                table.insert("on-chain", val.into());
```
stefan-mysten: Is this needed or can be simplified with:
```suggestion
                table.insert("on-chain", true.into());
```

stefan-mysten: Alternatively, does it make sense to add Copy on ConstTrue? Then you do not need the .clone()?

crates/sui-package-alt/src/sui_flavor.rs
amnn: nit (for the suggestion): this function doesn't panic.

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
amnn: My understanding is that it's because addresses are inherently tied to a specific network, so it doesn't make sense to set an address in the `dependencies` section, without being in the context of a specific network.

I had a related question though, which is: "why have the `on-chain = true` entry under `dependencies` at all, given that in order for the dep to be useful, it must have an entry in `dep-replacements.C` for some chain `C`.

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
```diff
@@ -0,0 +1,191 @@
...
+_on_chain = "<chain_id>"
+
+[dep-replacements._on_chain]
+onchain_0x0002 = { on-chain = "0x0002", override = true }
+onchain_0x0003 = { on-chain = "0x0033", override = true }
+```
+
+Key design choices:
+
+- **`name = "self"`** — a fixed name for all generated packages. Users never need to know
```
amnn: nit: could you just leave it empty, or at-least, more noticeably a reserved/system kind of name (like `_on_chain` below). `self` could potentially be read like a real name.

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
amnn, line 118: This caching mechanism is insufficient in the presence of system packages, where you may have cached a system package previously, but then that system package was upgraded -- you need to also perform a version check to understand whether the cached stub is stale or not.

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
```diff
@@ -0,0 +1,441 @@
...
+        .await
+        .map_err(|source| OnChainError::Fetch {
+            address: address.clone(),
+            source,
+        })?;
+
+    write_bytecode(&cache_dir, &data.modules);
+    write_source(&cache_dir);
+    write_manifest(&cache_dir, &manifest_path, &config.chain_id, &data);
+    write_published::<F>(&cache_dir, address, &config.chain_id, &data);
```
amnn: nit: your design mentions that `manifest` will be the last file written, but here `Published.toml` is the last file written. @stefan-mysten's comment regarding incomplete writes is relevant here.

Incidentally, it may be better to use `Published.toml` as the sentinel that the package has been fully written, because it contains the version information that can be used for the system package staleness check.

amnn: The fact that these functions can also panic is potentially also a problem, as noted below. What happens if you write the manifest and then `write_published` panics? Or you partially write the sentinel file, and then panic. These issues should be exposed as errors, so that you can catch them and clean up.

amnn (commented):
Spotted a couple of issues related to caching and atomicity, if you could take another look, thanks!

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
stefan-mysten: Duh, of course they're network specific. Thanks for reminding me that.

mdgeorge4153: the main reason is just to signal intent. It isn't required and is just informational (I can imagine someone writing a manifest and adding `on-chain = true` to the dependencies, and then adding the address later). I'd be open to dropping it

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
mdgeorge4153: yes

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
mdgeorge4153: it can't be empty - it needs to be a move identifier. I think a natural other thing would be something like `on_chain_0x1234` or something. I'm sort of sneaking in the `self` idea here because we've talked about the possibility of using that more broadly as a convention. Whatever we choose, it _is_ a real name in the sense that we can use that name in the generated code.

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
mdgeorge4153: no because we generate the manifest after we fetch the bytecode. The only way we could get into this situation is if the user mucks about in their cache, which I'm not too worried about.

mdgeorge4153: yes, I think there's a note about that somewhere in the doc, but you're absolutely right that we need to do something special for system packages.

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
mdgeorge4153: yes exactly. There'd be a new copy of the bytecode in `<devnet_chain_id>/<package_id>`.

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
mdgeorge4153: interesting point, I can look into it. I think the natural thing would be to reuse this type on the sui side.

crates/sui-package-alt/src/sui_flavor.rs
mdgeorge4153: [CLAUDE] Done — imported `OnChainPackageData`, `OriginalID`, and `PublishedID` from `move_package_alt`.

crates/sui-package-alt/src/sui_flavor.rs
mdgeorge4153: [CLAUDE] Applied — using `.with_context(|| format!(...))`.

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
mdgeorge4153: [CLAUDE] `OnChainPackageData` is intentionally the flavor-independent representation — it uses move-package-alt types (`OriginalID`, `PublishedID`) rather than sui-types (`ObjectID`, `UpgradeInfo`). The conversion happens in `SuiFlavor::fetch_onchain_package`. The fields map naturally: `module_map` → `modules`, `linkage_table` → `dependencies`, etc.

external-crates/move/crates/move-package-alt/src/flavor/mod.rs
mdgeorge4153: [CLAUDE] Updated to reference `OnChainPackageData` instead of listing fields.

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
mdgeorge4153: [CLAUDE] Removed.

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
mdgeorge4153: [CLAUDE] Updated the doc comments — `with_on_chain_package` is not testing-only (Vanilla itself is primarily used for tests, but the method is general-purpose).

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
mdgeorge4153: [CLAUDE] Applied.

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
mdgeorge4153: [CLAUDE] Updated the doc comment to remove the testing-only framing. `Vanilla::new()` is the standard constructor — Vanilla itself is only used in non-Sui contexts.

crates/sui-package-alt/src/sui_flavor.rs
mdgeorge4153: [CLAUDE] Added doc comment. Per amnn's correction, it says "returns an error" rather than "panics".

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Good point about error messages — renamed from `self` to `_onchain_package`.

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
mdgeorge4153: [CLAUDE] Removed — implementation status is tracked in the Linear project.

external-crates/move/crates/move-package-alt/src/dependency/combine.rs
mdgeorge4153: [CLAUDE] Applied (adapted for the `OnChainAt` → `OnChain` rename from #26552).

external-crates/move/crates/move-package-alt/src/on_chain/fetch.rs
mdgeorge4153: [CLAUDE] Fixed: (1) cache is now gated on `Published.toml` which is the last file written, so its presence indicates a complete cache; (2) all write functions now return `Result` with context; (3) on I/O failure, the cache directory is cleaned up via `remove_dir_all` to prevent a partially-populated cache.

external-crates/move/crates/move-package-alt/design/on-chain-fetching.md
mdgeorge4153: [CLAUDE] amnn answered this well — addresses are network-specific so they belong in `[dep-replacements]` which is per-environment, not `[dependencies]` which is shared.

PR #26562, opened 2026-05-11
[move] Make Vanilla a non-unit struct

mdgeorge4153:
## Summary

- Changes `Vanilla` from a unit struct to a struct with a private field
- Adds `Vanilla::new()` constructor
- Updates all call sites from `Vanilla` to `Vanilla::new()`

This is a mechanical refactor to prepare Vanilla for holding test-specific state (e.g. on-chain package data for testing `fetch_onchain_package`). Extracted from #26561 to reduce review noise.

## Test plan

- [x] All 185 move-package-alt tests pass
- [x] Clippy and fmt clean


external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
```diff
@@ -25,7 +25,10 @@ pub const DEFAULT_ENV_ID: &str = "_test_env_id";
 /// The [Vanilla] implementation of the [MoveFlavor] trait. This implementation supports no
 /// flavor-specific resolvers and stores no additional metadata in the lockfile.
 #[derive(Debug, Default)]
-pub struct Vanilla;
+pub struct Vanilla {
+    /// Private field to prevent unit-struct construction; use [`Vanilla::new`] instead.
+    _private: (),
+}
```
amnn: Does this achieve what you want? It makes it so that callers outside this crate don't get to see the fields of `Vanilla` (or depend on them not changing). Not a big issue either way if you plan on adding real fields shortly.

```suggestion
#[non_exhaustive]
pub struct Vanilla;
```

amnn (approved):

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
mdgeorge4153: I was thinking that too. The main thing is not to be constructing `Vanilla` instances outside of the module. I'll switch it to `#[non_exhaustive]` since it looks like I need to fix clippy too.

external-crates/move/crates/move-cli/src/sandbox/commands/test.rs
```diff
@@ -145,7 +145,7 @@ fn copy_pkg_and_deps(tmp_dir: &Path, pkg_dir: &Path) -> anyhow::Result<PathBuf>
 /// will need to be fixed.
 fn package_paths(pkg_dir: &Path) -> anyhow::Result<Vec<PathBuf>> {
     let root_pkg: RootPackage<Vanilla> =
-        PackageLoader::new(pkg_dir, Vanilla::default_environment(), Vanilla)
+        PackageLoader::new(pkg_dir, Vanilla::default_environment(), Vanilla::new())
```
stefan-mysten, line 148: Are there cases where we pass a Flavor with different environments?

external-crates/move/crates/move-decompiler/tests/tests.rs
```diff
@@ -42,8 +42,9 @@ fn run_move_test(file_path: &Path) -> datatest_stable::Result<()> {
 
     let mut writer = Vec::new();
     let env = Vanilla::default_environment();
-    let root_pkg: RootPackage<Vanilla> =
-        config.package_loader(pkg_dir, &env, Vanilla).load_sync()?;
+    let root_pkg: RootPackage<Vanilla> = config
+        .package_loader(pkg_dir, &env, Vanilla::new())
```
stefan-mysten: nit on style:        
Either:
```suggestion
        .package_loader(pkg_dir, &Vanilla::default_environment(), Vanilla::new())
```
or 

```suggestion
    let env = Vanilla::default_environment();
    let vanilla = Vanilla::new();
        .package_loader(pkg_dir, &env, vanilla)
```

stefan-mysten (approved):

external-crates/move/crates/move-package-alt/src/flavor/vanilla.rs
mdgeorge4153: [CLAUDE] Good call — switched to `#[non_exhaustive]` and dropped the `_private` field.

external-crates/move/crates/move-cli/src/sandbox/commands/test.rs
mdgeorge4153: [CLAUDE] No — in vanilla/move-cli there is only ever one environment (`_test_env`). This PR is purely mechanical: replacing `Vanilla` (unit struct) with `Vanilla::new()` to allow fields in a follow-up PR.

external-crates/move/crates/move-decompiler/tests/tests.rs
mdgeorge4153: [CLAUDE] Fixed — inlined both so the call reads `.package_loader(pkg_dir, &Vanilla::default_environment(), Vanilla::new())`.

PR #26564, opened 2026-05-11
sui-tool: add --metrics-port flag to download-formal-snapshot

kklas:
## Summary

`download_formal_snapshot` in `crates/sui-tool/src/lib.rs` currently binds its prometheus metrics server to a hardcoded `127.0.0.1:9184`. This conflicts with:

- Running on a host where a validator is already bound to 9184 (the validator's own prometheus endpoint).
- Running multiple snapshot downloads in parallel on the same host.

This PR adds `--metrics-port <u16>` to `sui-tool download-formal-snapshot`, defaulting to 9184 so existing invocations are unchanged.

## Test plan

- [ ] `cargo check -p sui-tool` passes
- [ ] `sui-tool download-formal-snapshot --help` shows the new flag with default 9184
- [ ] Running the command without the flag continues to bind 9184 as before
- [ ] Running with `--metrics-port 9999` binds 9999 instead

### DIFF
--- a/crates/sui-tool/src/commands.rs
+++ b/crates/sui-tool/src/commands.rs
@@ -343,6 +343,9 @@ pub enum ToolCommand {
         /// Defaults to 3 retries. Set to 0 to disable retries.
         #[clap(long = "max-retries", default_value = "3")]
         max_retries: usize,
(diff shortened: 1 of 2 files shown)


amnn (approved):
LGTM, thanks!

PR #26565, opened 2026-05-11
sui-indexer-alt-consistent-store: support restoring from arbitrary epoch directories

### LINKED ISSUE #26563: sui-indexer-alt-consistent-store restore cannot reach historical snapshots under gs://mysten-mainnet-formal/archive/
## Problem

`gs://mysten-mainnet-formal` (the canonical published formal-snapshot bucket for mainnet) has two distinct snapshot layouts:

```
gs://mysten-mainnet-formal/
├── MANIFEST                    # root manifest covering recent epochs
├── archive/                    # historical snapshots, NO root manifest inside
│   ├── epoch_150/
│   ├── epoch_180/
│   ├── …                       # every 30 epochs
│   └── epoch_1110/
├── epoch_1064/                 # recent snapshots, every epoch
├── epoch_1065/
├── …
└── epoch_1073/
```

`sui-indexer-alt-consistent-store restore formal-snapshot` only knows how to read the *root* layout: it issues `store.get("MANIFEST")` and `store.get("epoch_<N>/_SUCCESS")` against the bucket root (`crates/sui-indexer-alt-consistent-store/src/restore/formal_snapshot.rs`).

That means:

1. The historical archive under `archive/` is **unreachable** — the tool has no way to be told "look under this sub-path."
2. Even if you could point it there, `archive/` ships **no `MANIFEST`** and **no per-epoch `_SUCCESS` markers**, so the current "always validate" logic in `formal_snapshot.rs` would fail anyway.

The historical layout exists only on the GCS bucket. The S3-compatible endpoint that `sui-tool download-formal-snapshot` reads from (`https://formal-snapshot.mainnet.sui.io`, see `crates/sui-tool/src/commands.rs:673` / `:689`) does **not** mirror the `archive/` sub-path — there is no other public route to those historical snapshots.

So for anyone wanting to replay or rebuild consistent-store state from a point in the past (anything older than the last ~10 epochs at the bucket root), there is no path through the official tool.

## What would help

Two small, independent additions to `formal_snapshot.rs`:

1. **GCS sub-path support.** Allow `--gcs bucket/prefix`. Split on the first `/`, wrap the resulting store in `object_store::prefix::PrefixStore`. The crate already has `impl<S: ObjectStore> Storage for S`, so no custom wrapper is needed. Pure additive change.

2. **Tolerant marker checks for `archive/`-style layouts.** When the operator supplies `--epoch <N>` explicitly, treat a missing `MANIFEST` / `_SUCCESS` as acceptable (the operator has vouched for the epoch). When `--epoch` is omitted, keep the current strict behaviour so auto-latest-discovery still works.

Together these would make `--gcs mysten-mainnet-formal/archive --epoch 600` the natural way to restore from historical state.

We're carrying a local patch with these two changes — happy to send a PR if the direction is OK.

## Environment

- Sui main, `c188832a5c` (current as of 2026-05-11)
- Failing flow: pointing the tool at `--gcs mysten-mainnet-formal` for an epoch older than what's at the bucket root.

kklas:
## Summary

Closes #26563.

The `restore formal-snapshot` flow currently assumes a fixed layout at the storage source: a root `MANIFEST` enumerating epochs, an `epoch_<N>/` directory per epoch, and a `_SUCCESS` marker confirming each epoch is complete. That works for the live snapshots at the bucket root of `gs://mysten-mainnet-formal/` but not for the historical archive under `gs://mysten-mainnet-formal/archive/`, which ships none of those markers and isn't at the bucket root. There's no other public source for the historical epochs.

This PR follows the design from #26563 (revised per review feedback):

- Adds a new `--path <prefix>` flag pointing at a path prefix inside the storage source that contains per-epoch subdirectories. The tool then looks for `<path>/epoch_<N>/`, with `<N>` from `--epoch`.
- When `--path` is supplied, the tool skips the root `MANIFEST` and per-epoch `_SUCCESS` lookups — the operator vouches for the snapshot being complete.
- `--path` requires `--epoch` (since the prefix layout may not include a `MANIFEST` to discover the latest epoch).
- When `--path` is not set, behaviour is unchanged.
- The previous `--path` flag (which configured the local filesystem source) is renamed to `--local`.

Example:

```
sui-indexer-alt-consistent-store restore \
  --gcs mysten-mainnet-formal \
  --path archive \
  --epoch 600 \
  --remote-store-url https://checkpoints.mainnet.sui.io \
  --pipeline balances
```

All of the new logic is concentrated in `FormalSnapshot::new`. A new `epoch_dir` field on `FormalSnapshot` carries the resolved directory (`epoch_<N>` or `<path>/epoch_<N>`); `manifest()` and `file()` use it directly instead of reconstructing `epoch_<N>` from the watermark.

## Test plan

- [x] `cargo check -p sui-indexer-alt-consistent-store` passes.
- [x] `cargo build -p sui-indexer-alt-consistent-store --release` succeeds.
- [x] `sui-indexer-alt-consistent-store restore --help` shows `--path` and `--local` with correct descriptions and the `--path` → `--epoch` requirement.
- [x] End-to-end against the live bucket: `restore --gcs mysten-mainnet-formal --path archive --epoch 600 --remote-store-url https://checkpoints.mainnet.sui.io --pipeline balances` — the tool resolves `epoch_dir="archive/epoch_600"`, anchors the watermark, reads the per-epoch `MANIFEST` from `gs://mysten-mainnet-formal/archive/epoch_600/MANIFEST`, broadcasts 726 partitions, and the `balances` pipeline worker starts fetching them. Before this PR no flag combination could reach this layout.
- [x] Canonical flow unchanged: `restore --gcs mysten-mainnet-formal --epoch 1070 --remote-store-url ...` exercises the original MANIFEST + `_SUCCESS` path.

### DIFF
--- a/crates/sui-indexer-alt-consistent-store/src/restore/formal_snapshot.rs
+++ b/crates/sui-indexer-alt-consistent-store/src/restore/formal_snapshot.rs
@@ -51,7 +51,7 @@ pub struct FormalSnapshotArgs {
 
     /// Fetch formal snapshot from local filesystem. Provide the path to the snapshot directory.
     #[arg(long, group = "source")]
-    pub path: Option<PathBuf>,
+    pub local: Option<PathBuf>,
 
     /// URL of the remote checkpoint store. Used to fetch the mapping between epochs and
     /// checkpoints.
@@ -59,8 +59,16 @@ pub struct FormalSnapshotArgs {
     pub remote_store_url: Url,
 
     /// The epoch to restore from. Restores from the latest epoch if not specified.
-    #[arg(long)]
+    #[arg(long, conflicts_with = "path")]
     pub epoch: Option<u64>,
+
+    /// Path within the storage source pointing directly at an epoch directory
+    /// (e.g. `archive/epoch_600`). When supplied, the tool skips reading the
+    /// root `MANIFEST` and the per-epoch `_SUCCESS` marker — the operator vouches
+    /// for the snapshot being complete. The epoch number is parsed from the
+    /// trailing `epoch_<N>` path component.
+    #[arg(long, conflicts_with = "epoch")]
+    pub path: Option<String>,
 }
 
 /// Interface over a formal snapshot at a specific epoch.
@@ -69,14 +77,19 @@ pub(super) struct FormalSnapshot {
     /// The source of files related to the formal snapshot.
     source: Arc<dyn Storage + Send + Sync + 'static>,
 
+    epoch_dir: String,
+
     /// The point in time this snapshot is from.
     watermark: Watermark,
 }
 
 impl FormalSnapshot {
-    /// Establish a connection to a formal snapshot source, and validate that it can serve data for
-    /// a formal snapshot at the specified epoch (or the latest available epoch if none is
-    /// specified).
+    /// Establish a connection to a formal snapshot for a specific epoch.
+    ///
+    /// When `--path` is supplied, the operator names the epoch directory directly.
+    /// Otherwise the source's root `MANIFEST` is consulted to validate the epoch
+    /// and a `_SUCCESS` marker is checked; if no epoch is supplied, the latest in
+    /// the manifest is used.
     pub(super) async fn new(
         snapshot_args: FormalSnapshotArgs,
         connection_args: StorageConnectionArgs,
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-consistent-store/src/restore/formal_snapshot.rs
```diff
@@ -108,42 +121,56 @@ impl FormalSnapshot {
...
         } else {
             bail!("No formal snapshot source provided");
         };
 
-        // Fetch details about all the epochs it has available, and use that to confirm and
-        // validate the epoch to restore from (it is available and complete at the source).
-        let root_manifest = RootManifest::read(
-            store
-                .get("MANIFEST".into())
+        let (epoch, epoch_dir) = if let Some(path) = snapshot_args.path {
```
amnn, line 134: Hmm, I didn't account for the fact that we still need to know the epoch number below in order to fetch the watermark. I think the assumption in this code (that the last compoonent of the path must be of the form `epoch_<num>`) is right, but that then we should expose things slightly differently:

- Use `--path` as a path-prefix, and then expect `--epoch <N>` to tell us which epoch directory to look for inside that directory.
- If you pass `--path` then `--epoch` is required (because we don't have the manifest to figure out what the latest available epoch is in that directory).

So the restore call becomes:

```
--gcs mysten-mainnet-formal --path archive --epoch 600
```

amnn (commented):
Thanks @kklas, looks directionally good -- just one suggested change on the meaning of `--path` and how it interacts with `--epoch`. I hadn't spotted we needed to know the numeric epoch we were restoring from to fetch the watermark.

crates/sui-indexer-alt-consistent-store/src/restore/formal_snapshot.rs
kklas: Good call — applied in b096fd38d7. `--path` is now a path prefix and requires `--epoch`; the tool constructs `<path>/epoch_<N>` internally, no more basename parsing. Verified end-to-end against `--gcs mysten-mainnet-formal --path archive --epoch 600`.

amnn (approved):
LGTM, thanks @kklas !

amnn:
It seems like I can't shake the simtest issue. It is not related to this PR though, I am fairly certain. @kklas, would you mind rebasing on `main` and I can give it another try?

kklas:
Rebased on `main` (`44a2d5dd8c`) — branch is now at `f8772e62d2`. Both commits replayed cleanly, no conflicts.

PR #26572, opened 2026-05-11
docker/sui-tools: build + ship move-analyzer

ebmifa:
## Summary

\`move-analyzer\` is in [\`binary-build-list.json\`'s \`release_binaries\`](../blob/main/binary-build-list.json) — the legacy \`sui-release-binaries.yml\` pipeline builds it and uploads to \`s3://sui-releases/<sha>/move-analyzer\`. But **no mp3 Dockerfile builds it**, so the mp3-driven extraction path (image build pod → \`binary-uploader\` sidecar → S3) has no source artifact for it.

This PR adds \`move-analyzer\` to the \`sui-tools\` image. Natural fit — sui-tools is the developer-CLI image already shipping \`sui\`, \`sui-faucet\`, \`sui-bridge*\`, \`sui-cluster-test\`, \`sui-tool\`. Same build profile, same target dir, same /usr/local/bin/ destination as its siblings.

## Diff

\`docker/sui-tools/Dockerfile\` — two lines:
- \`--bin move-analyzer\` added to the cargo build invocation
- \`COPY --from=builder /sui/target/release/move-analyzer /usr/local/bin/move-analyzer\`

## Test plan

- [x] Dockerfile lints clean (no new flags, same pattern as other binaries)
- [ ] Image build succeeds via mp3 staging (CI will exercise on push)
- [ ] After this lands, MystenLabs/sui-operations#XXXX adds \`move-analyzer\` to mp3's \`binary_upload_config\` for \`sui-tools\` / \`sui-tools-arm64\` so the sidecar extracts it to S3 on the same path the legacy pipeline used

## Cost impact

\`move-analyzer\` is built from the same source tree already in the sui-tools build context. Adding it increases cargo compile time by ~10-30s (one extra binary in the same workspace) and image size by ~50 MB. mp3's sticky-routed sui-tools pod will absorb this on the cache mount, so most builds will see negligible incremental cost.

### DIFF
--- a/docker/sui-tools/Dockerfile
+++ b/docker/sui-tools/Dockerfile
@@ -29,7 +29,8 @@ RUN cargo build --locked --profile ${PROFILE} \
     --bin sui-bridge-cli \
     --bin sui \
     --bin sui-cluster-test \
-    --bin sui-tool
+    --bin sui-tool \
+    --bin move-analyzer
 
 # Production Image
 FROM debian:bullseye-slim AS runtime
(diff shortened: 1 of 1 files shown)


awelc (approved):
LGTM

PR #26581, opened 2026-05-11
[move][move-compiler] Add `let-else`

cgswords:
## Description 

Refutable `let` with a divergent `else` branch. Binders escape into the enclosing scope on the success path; the else must diverge (return / abort / break / continue).

```
  let ABC::C(x) = subject else { return 0 };
  // x is in scope
```

Gated by `FeatureGate::LetElse`; legacy editions parse let as before. The LHS is parsed as a MatchPattern (unifying the grammar with match arms), typing declares the binders in the enclosing scope, then HLIR lowers to a two-arm match (success arm + wildcard catch-all running the else).

Fix: also restored `RestrictedIdentifier` in field patterns (`FIELD_BINDING_START_SET`), fixing a regression in `restricted_ident_positions`.

## Test plan 

23 new tests, plus fixups.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-analyzer/tests/structs.snap
```diff
@@ -163,7 +163,7 @@ Move2024::structs::Positional
 
 -- test 18 -------------------
 use line: 36, use_ndx: 2
-Use: 'v', start: 29, end: 30
+Use: 'val2', start: 29, end: 33
```
cgswords, line 166: These are maybe dubious changes. @awelc

external-crates/move/crates/move-compiler/src/diagnostics/codes.rs
```diff
@@ -309,6 +271,8 @@ codes!(
         InvalidString: { msg: "invalid string after type inference", severity: NonblockingError },
         MissingLiteralType:
             { msg: "unable to determine principal type for literal", severity: Warning },
+        InvalidLetElse:
+            { msg: "'let...else' block must diverge", severity: BlockingError },
```
cgswords, line 275: This could conceivably be a nonblocking error, but I have no idea what that code would even look like / what we would do. Maybe staple on an abort?

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -98,11 +98,17 @@ pub enum DisplayVar {
     Orig(String),
     Tmp,
     MatchTmp(String),
+    /// The subject temporary introduced when lowering a `let ... else` to a
+    /// match. Diagnostics use this to phrase messages in terms of the surface
+    /// `let ... else` form rather than the synthesized `match`.
+    LetElseSubject(String),
 }
 
 pub fn display_var(s: Symbol) -> DisplayVar {
```
cgswords, line 107: We really should do this in a more-principled way some day.

PR #26587, opened 2026-05-12
kv rpc: add bitmap query APIs

nickvikeras:
## Description

This implements the grpc query apis for archival.
This won't build until <https://github.com/MystenLabs/sui-rust-sdk/pull/256> lands.


crates/sui-inverted-index/src/bitmap_query.rs
```diff
@@ -0,0 +1,981 @@
...
+        term.validate()?;
+        Ok(term)
+    }
+
+    pub fn validate(&self) -> Result<()> {
+        if !self.has_include() {
+            bail!("bitmap query term must contain at least one include literal");
+        }
+        Ok(())
+    }
```
bmwill: don't you want to validate each individual Literal here as well?

crates/sui-inverted-index/src/bitmap_query.rs
```diff
@@ -0,0 +1,981 @@
...
+{
+    async_stream::try_stream! {
+        let stream = eval_bitmap_query_bucket_stream(source, query, range.clone(), direction);
+        let flattened = flatten_bucket_stream(stream, range, bucket_size, direction);
+        futures::pin_mut!(flattened);
+        while let Some(item) = flattened.next().await {
+            yield item?;
+        }
+    }
+    .boxed()
```
bmwill: Theres a bit of indirection/boxing happening here. I wonder how much we could compress together to avoid the extra indirection since you're really just chaining iterators together, right?

Eg, why do you even need the extra stream boxing here, can't you just return flatten_bucket_stream directly?

crates/sui-inverted-index/src/bitmap_query.rs
```diff
@@ -0,0 +1,981 @@
...
+    source: S,
+    query: BitmapQuery,
+    range: Range<u64>,
+    direction: ScanDirection,
+) -> BucketStream
+where
+    S: BitmapBucketSource,
+{
+    async_stream::try_stream! {
+        query.validate()?;
```
bmwill: i wonder if we could leverage type safety here to require that a BitmapQuery is correct by construction so avoid needing to do a validate step here at all, you'd know if it was valid well before now.

crates/sui-kv-rpc/src/query_options.rs
```diff
@@ -0,0 +1,761 @@
...
+        None => hasher.update([0]),
+        Some(message) => {
+            hasher.update([1]);
+            let bytes = message.encode_to_vec();
+            let len = u32::try_from(bytes.len()).expect("scan scope part should fit in u32");
+            hasher.update(len.to_be_bytes());
+            hasher.update(bytes);
+        }
+    }
+}
```
bmwill, line 466: Can you provide a bit more context on the hashing here? Generally protobuf encoding isn't going to be canonical so the same message could serialize into different byte streams. Given our current implementation is:
1. deserializing from the wire into our own types
2. dropping unknown fields
3. re-serializing with our own serializer 

this isn't a problem today (since our own serialization will be deterministic) but it is something that isn't a guarantee if things change. So at a minimum having a bit more docs on this would be nice.

crates/sui-kv-rpc/src/object_cache.rs
```diff
@@ -0,0 +1,603 @@
...
+use futures::stream::BoxStream;
+use sui_rpc_api::RpcError;
+use sui_types::object::Object;
+use sui_types::storage::ObjectKey;
+use tokio::sync::oneshot;
+use tokio::task::JoinSet;
+
+use crate::bigtable_client::BigTableClient;
+
+pub(crate) type ObjectMap = Arc<HashMap<ObjectKey, Arc<Object>>>;
```
bmwill: Object is already Arc'd so this is doubly Arc'd.
```suggestion
pub(crate) type ObjectMap = Arc<HashMap<ObjectKey, Object>>;
```

crates/sui-rpc-api/src/metrics.rs
```diff
@@ -177,18 +177,19 @@ fn compute_metric_label(
...
 ) -> Cow<'static, str> {
     match (is_grpc, matched_path) {
-        (true, Some(matched)) => {
-            if grpc_method_allowlist.contains(uri_path) {
-                Cow::Owned(uri_path.to_owned())
-            } else {
-                Cow::Owned(matched.to_owned())
-            }
-        }
+        (true, _) if grpc_method_allowlist.contains(uri_path) => Cow::Owned(uri_path.to_owned()),
```
bmwill, line 180: what was being droped or missing before this change?

nickvikeras: The path label was "unknown" for most of the sui-kv-rpc metrics without this change. [Example](https://metrics.sui.io/goto/s52q4p?orgId=stacks-447897) from the prod dashboard. With this change you get a path label like

```
path="/sui.rpc.v2alpha.LedgerService/ListTransactions"
```

Not sure why archival has this problem and the full node doesn't though, maybe the fix should be something else.

nickvikeras: The difference is apparently that the sui-kv-rpc request is not routed through an Axum router.

bmwill: oh interesting...either way this is ok for now

crates/sui-inverted-index/src/bitmap_query.rs
nickvikeras: I think I removed as much of the redundant boxing as I could.

crates/sui-kv-rpc/src/operation.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    pub(crate) fn event_filter_query(&self, filter: &EventFilter) -> Result<BitmapQuery, RpcError> {
+        event_filter_to_query(filter, self.limits.max_bitmap_filter_literals)
+    }
+}
+
+/// Wrap a server-streaming response with an absolute deadline. When the
+/// deadline fires, the wrapped stream yields a `DeadlineExceeded` Status
+/// once and ends; the inner stream is dropped at that point.
+fn with_deadline<S, T>(
```
bmwill, line 121: Given our conversation in slack we probably want our timeouts to be able to track the last known fully scanned bucket/txid so that we can return a cursor on an `end_response` frame so the user can always make forward progress in their list queries.

I do think this may slightly complicate our cancelation logic though since we don't just want to necessarily drop the whole stream on timeout in all cases.

bmwill: we dont necessarily need to get this completely right before we land this larger PR though

bmwill (approved):

crates/sui-kv-rpc/src/operation.rs
nickvikeras: Yup will knock it out in a smaller PR.

crates/sui-inverted-index/src/bitmap_query.rs
```diff
@@ -0,0 +1,1006 @@
...
+    pub fn new(bytes: Vec<u8>) -> Result<Self> {
+        if bytes.is_empty() {
+            bail!("bitmap dimension key must not be empty");
+        }
+        if bytes.len() == 1 {
+            bail!("bitmap dimension value must not be empty");
+        }
+        if IndexDimension::from_tag_byte(bytes[0]).is_none() {
+            bail!("unknown bitmap dimension tag {}", bytes[0]);
+        }
```
amnn, line 113: nit: `anyhow::ensure!` (here and elsewhere)

crates/sui-inverted-index/src/bitmap_query.rs
```diff
@@ -0,0 +1,1006 @@
...
+                if !bitmap.is_empty() {
+                    yield (target_bucket, bitmap);
+                }
+            } else {
+                // Sparse bitmap rows encode only non-empty buckets. If a child
+                // is behind the current alignment target, then at least one
+                // other include stream is already past that bucket and has an
+                // implicit all-zero bitmap there. The intersection for the
+                // lagging bucket is empty, so consume it and re-peek.
+                for (i, child) in children.iter_mut().enumerate() {
```
amnn, line 301: nit: `zip_eq` with `peeks` instead of `enumerate`?

crates/sui-inverted-index/src/bitmap_query.rs
```diff
@@ -0,0 +1,1006 @@
...
+                yield (next_bucket, bitmap);
+            }
+        }
+    }
+}
+
+/// Merge-join subtraction for an anchored negative literal.
+///
+/// A negative literal in a valid term is evaluated as `a AND NOT b`, where `a`
+/// is the already-anchored positive candidate stream. Bitmap subtraction
```
amnn, line 389: @nickvikeras, could you share more on the rationale for anchoring -- what's an example of a query with shape `NOT a AND NOT b` that exhibits pathological performance characteristics (devolves into a linear scan)? I couldn't think of a case like that which was unique to negation.

PR #26588, opened 2026-05-12
[sui-fork] Fix the way object_by_key is handling deleted/wrapped objects

stefan-mysten:
## Description 

This PR fixes the way `object_by_key` is handling deleted/wrapped objects, as this API is called when a checkpoint is created & published during a transaction which might require an older version of the object that is being deleted.

## Test plan 

New test.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-fork/src/tests/rpc_executor.rs
```diff
@@ -142,6 +144,32 @@ impl TestHarness {
         let tx_data = self.build_transfer_tx_data(amount);
         Transaction::from_data_and_signer(tx_data, vec![&self.sender_key])
     }
+
+    fn build_send_gas_funds_tx(&self, recipient: SuiAddress) -> Transaction {
```
stefan-mysten, line 158: The bug was found when trying to call the `send_funds` API. Because the object is deleted, it failed to create a checkpoint as it needed to read the object version before deletion. So this is required to reconstruct that tx for tests.

jordanjennings-mysten (approved):

PR #26594, opened 2026-05-12
[graphql] Add Query.checkpoint(digest:) lookup

tpham-mysten:
## Summary

- Adds `digest` as a second optional argument to `Query.checkpoint`, closing a parity gap with gRPC's `GetCheckpointRequest.digest`. Pysui (and any client unifying gRPC and GraphQL transports) can now route a digest lookup through GraphQL.
- Specifying both `sequenceNumber` and `digest` returns a `BAD_USER_INPUT` error. Specifying neither continues to return the latest checkpoint. Specifying only `digest` resolves it to a sequence number via the configured KV reader (BigTable, LedgerGrpc, or the new `kv_checkpoints_by_digest` table on PG), then delegates to the existing `Checkpoint::with_sequence_number` constructor so downstream resolvers behave identically.

The PR includes one infrastructure commit that extends the transactional-test runner with a new `@{cp_digest_N}` substitution. This is symmetric with the existing `@{digest_N}` for transactions and lets the snapshot test reference simulator-produced checkpoint digests without hardcoding them.

## Test plan

- `cargo check -p sui-indexer-alt-graphql -p sui-indexer-alt-reader`.
- `cargo nextest run -p sui-indexer-alt-graphql --lib` and `--features staging --lib` (regenerated schema and registry snapshots).

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Support checkpoint lookup by digest
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
```diff
@@ -41,6 +42,12 @@ use crate::task::watermark::Watermarks;
 
 pub(crate) mod filter;
 
+#[derive(thiserror::Error, Debug)]
+pub(crate) enum Error {
```
amnn, line 46: Double check because I may be misremembering, but in other circumstances, I think I've seen us treat having both bounds as a sort of redundant intersection test, so if both are provided, and the sequence number provided happens to match the sequence number of the checkpoint with that digest, then we still return the value, otherwise we return `None`.

It's a pattern to be as permissive as possible with the inputs while still being logically consistent.

crates/sui-indexer-alt-reader/src/checkpoints.rs
```diff
@@ -150,3 +157,105 @@ impl Loader<CheckpointKey> for LedgerGrpcReader {
...
+                None => Ok(None),
+            }
+        });
+
+        Ok(try_join_all(futures).await?.into_iter().flatten().collect())
+    }
+}
+
+#[async_trait::async_trait]
+impl Loader<CheckpointDigestKey> for LedgerGrpcReader {
```
amnn: We may need to be mindful about parallelism here? A data loader can batch up 1000 requests, would archival service be okay with 1000 concurrent requests coming its way?

amnn: Also, if all the data loader is doing is sending all the requests concurrently, then we maybe don't need a data loader at all in this case, and we can just put the individual call directly in `KvLoader::load_one_checkpoint_seq_by_digest`.

amnn (approved):
Only comment that I have is related to the performance implications on the `LedgerGrpcReader` codepath (and whether we need to use a data loader at all there -- I think it might actually make things worse).

It would be good to validate the performance ceiling there before shipping to production, but otherwise the implementation is solid, thanks as always @tpham-mysten .

crates/sui-indexer-alt-reader/src/checkpoints.rs
wlmyng: We have a very similar implementation above in `impl Loader<CheckpointKey> for LedgerGrpcReader {`, the same commentary would apply there as well. I suppose this is because dataloaders are only meaningful if we can batch multiple requests into one - so if we're still fanning out N requests, dataloaders don't meaningfully contribute?

wlmyng (approved):

crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
tpham-mysten: I took a look at how similar resolvers handle this. The codebase pattern is actually returning the error for multiple bounds:
  - [Query.object](crates/sui-indexer-alt-graphql/src/api/types/object.rs:731): errors OneBound if more than one is provided.
  - [Query.package](crates/sui-indexer-alt-graphql/src/api/types/move_package.rs:630) (version / atCheckpoint): same OneBound pattern.
  - [Query.address](crates/sui-indexer-alt-graphql/src/api/types/address.rs:476) (rootVersion / atCheckpoint): same OneBound pattern.

I think  the error is also clear enough for callers: the message tells them exactly what to fix. Hence I will leave thing as-is for now

crates/sui-indexer-alt-reader/src/checkpoints.rs
tpham-mysten: Agreed that DataLoader isn't making a big difference here, so I just dropped it on the BigTable and LedgerGrpc paths for the digest lookup

That said, DataLoader could still buy us key dedup, but it's probably not worth the extra boilerplate in this case.

FWIW, regarding concurrent requests to Archival service, I was having the same questions when doing GraphQL subscription gap recovery, and @nickvikeras mentioned the ideal batch would be ~100 requests, and we are using the [chunk size of 50 ](https://github.com/MystenLabs/sui/blob/7c05585f5b80d24d749b33dbf8caeec9478adfce/crates/sui-indexer-alt-graphql/src/config.rs#L251)there. In this case, if we want to keep DataLoader, and use a small concurrency batch, I think we can do something like:

```
stream::iter(keys.iter())
      .map(|key| async { single_grpc_call(key).await }) 
      .buffer_unordered(100)                        
      .try_collect::<Vec<_>>()
      .await?
```

tpham-mysten (commented):
Thanks both for the review! I was also having some thoughts on the DataLoader in the comment

PR #26595, opened 2026-05-12
[RPC] VM error visibility indexing

jordanjennings-mysten:
## Description 
  Adds JSON-RPC index support for VM execution error visibility.

  This threads richer `ExecutionErrorContext` through fullnode execution, extracts execution error metadata/source for failed txs, and writes it to a new `execution_error_metadata` index by tx digest. Validators continue using the on-chain data `ExecutionFailure` path and drop the extra metadata.

  The new index table is prunable by transaction sequence number and has a round-trip unit test.

requires changes to sui-apis sui-rust-sdk:
https://github.com/MystenLabs/sui-rust-sdk/pull/267
https://github.com/MystenLabs/sui-apis/pull/28

## Test plan 
sui core
- test_execution_error_metadata_round_trip
- test_empty_execution_error_metadata_is_ignored
- test_validator_execution_does_not_store_error_metadata
- execution_error_metadata_table_accepts_future_proto_schema

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -139,7 +139,9 @@ use sui_types::effects::{
     InputConsensusObject, SignedTransactionEffects, TransactionEffects, TransactionEffectsAPI,
     TransactionEvents, VerifiedSignedTransactionEffects,
 };
-use sui_types::error::{ExecutionError, ExecutionErrorContext, SuiErrorKind, UserInputError};
+use sui_types::error::{
+    ExecutionError, ExecutionErrorContext, ExecutionErrorMetadata, SuiErrorKind, UserInputError,
+};
 use sui_types::event::EventID;
 use sui_types::executable_transaction::VerifiedExecutableTransaction;
 use sui_types::execution_status::ExecutionErrorKind;
@@ -1461,7 +1463,7 @@ impl AuthorityState {
         certificate: &VerifiedExecutableTransaction,
         mut execution_env: ExecutionEnv,
         epoch_store: &Arc<AuthorityPerEpochStore>,
-    ) -> ExecutionOutput<(TransactionEffects, Option<ExecutionErrorContext>)> {
+    ) -> ExecutionOutput<(TransactionEffects, Option<ExecutionError>)> {
         let _scope = monitored_scope("Execution::try_execute_immediately");
         let _metrics_guard = self.metrics.internal_execution_latency.start_timer();
 
@@ -1640,10 +1642,7 @@ impl AuthorityState {
         &self,
         certificate: &VerifiedCertificate,
         execution_env: ExecutionEnv,
-    ) -> (
-        VerifiedSignedTransactionEffects,
-        Option<ExecutionErrorContext>,
-    ) {
+    ) -> (VerifiedSignedTransactionEffects, Option<ExecutionError>) {
         let epoch_store = self.epoch_store_for_testing();
         let (effects, execution_error_opt) = self
             .try_execute_immediately(
@@ -1661,10 +1660,7 @@ impl AuthorityState {
         &self,
         executable: &VerifiedExecutableTransaction,
         execution_env: ExecutionEnv,
-    ) -> (
-        VerifiedSignedTransactionEffects,
-        Option<ExecutionErrorContext>,
-    ) {
+    ) -> (VerifiedSignedTransactionEffects, Option<ExecutionError>) {
         let epoch_store = self.epoch_store_for_testing();
         let (effects, execution_error_opt) = self
             .try_execute_immediately(executable, execution_env, &epoch_store)
@@ -1740,7 +1736,7 @@ impl AuthorityState {
     ) -> ExecutionOutput<(
         TransactionOutputs,
         Vec<ExecutionTiming>,
-        Option<ExecutionErrorContext>,
+        Option<ExecutionError>,
     )> {
         let _scope = monitored_scope("Execution::process_certificate");
         let tx_digest = *certificate.digest();
@@ -1917,8 +1913,12 @@ impl AuthorityState {
         SuiGasStatus,
         TransactionEffects,
         Vec<ExecutionTiming>,
-        Result<(), ExecutionErrorContext>,
+        Result<(), ExecutionError>,
+        Option<ExecutionErrorMetadata>,
     ) {
+        // Fullnodes retain source errors and VM metadata as sidecar data for richer user-facing
+        // error reporting. Validators use the lean execution path and intentionally return no
+        // sidecar metadata so effects stay independent of non-consensus data.
         if is_fullnode {
             let (inner_temp_store, gas_status, effects, timings, execution_error) = executor
                 .execute_transaction_to_effects_and_execution_error(
@@ -1939,12 +1939,18 @@ impl AuthorityState {
                     &mut None,
                 );
 
+            let execution_error_metadata = execution_error
+                .as_ref()
+                .err()
+                .and_then(ExecutionErrorContext::metadata_with_source);
+
             (
                 inner_temp_store,
                 gas_status,
                 effects,
                 timings,
-                execution_error,
+                execution_error.map_err(ExecutionError::from),
+                execution_error_metadata,
             )
         } else {
             let (inner_temp_store, gas_status, effects, timings, execution_failure) = executor
@@ -1971,7 +1977,8 @@ impl AuthorityState {
                 gas_status,
                 effects,
                 timings,
-                execution_failure.map_err(ExecutionErrorContext::from),
+                execution_failure.map_err(ExecutionError::from),
+                None,
             )
         }
     }
@@ -1996,7 +2003,7 @@ impl AuthorityState {
     ) -> ExecutionOutput<(
         TransactionOutputs,
         Vec<ExecutionTiming>,
-        Option<ExecutionErrorContext>,
+        Option<ExecutionError>,
     )> {
         let _scope = monitored_scope("Execution::prepare_certificate");
         let _metrics_guard = self.metrics.prepare_certificate_latency.start_timer();
@@ -2060,31 +2067,37 @@ impl AuthorityState {
         let tracking_store = TrackingBackingStore::new(self.get_backing_store().as_ref());
 
         #[allow(unused_mut)]
-        let (inner_temp_store, _, mut effects, timings, execution_error_result) = self
-            .execute_transaction_to_effects(
-                &**epoch_store.executor(),
-                &tracking_store,
-                protocol_config,
-                // TODO: would be nice to pass the whole NodeConfig here, but it creates a
-                // cyclic dependency w/ sui-adapter
-                self.config
-                    .expensive_safety_check_config
-                    .enable_deep_per_tx_sui_conservation_check(),
-                execution_params,
-                &epoch_store.epoch_start_config().epoch_data().epoch_id(),
-                epoch_store
-                    .epoch_start_config()
-                    .epoch_data()
-                    .epoch_start_timestamp(),
-                input_objects,
-                gas_data,
-                gas_status,
-                kind,
-                rewritten_inputs,
-                signer,
-                tx_digest,
-                self.is_fullnode(epoch_store),
-            );
+        let (
+            inner_temp_store,
+            _,
+            mut effects,
+            timings,
+            execution_error_result,
+            execution_error_metadata,
+        ) = self.execute_transaction_to_effects(
+            &**epoch_store.executor(),
+            &tracking_store,
+            protocol_config,
+            // TODO: would be nice to pass the whole NodeConfig here, but it creates a
+            // cyclic dependency w/ sui-adapter
+            self.config
+                .expensive_safety_check_config
+                .enable_deep_per_tx_sui_conservation_check(),
+            execution_params,
+            &epoch_store.epoch_start_config().epoch_data().epoch_id(),
+            epoch_store
+                .epoch_start_config()
+                .epoch_data()
+                .epoch_start_timestamp(),
+            input_objects,
+            gas_data,
+            gas_status,
+            kind,
+            rewritten_inputs,
+            signer,
+            tx_digest,
+            self.is_fullnode(epoch_store),
+        );
 
         let execution_error_opt = execution_error_result.err();
 
@@ -2197,10 +2210,6 @@ impl AuthorityState {
                 &tracking_store.into_read_objects(),
             );
 
-        let execution_error_metadata = execution_error_opt
-            .as_ref()
-            .and_then(ExecutionErrorContext::metadata_with_source);
-
         // index certificate
         let _ = self
             .post_process_one_tx(certificate, &effects, &inner_temp_store, epoch_store)
@@ -2252,10 +2261,7 @@ impl AuthorityState {
                 epoch_store,
             )
             .unwrap();
-        Ok((
-            transaction_outputs,
-            execution_error_opt.map(ExecutionError::from),
-        ))
+        Ok((transaction_outputs, execution_error_opt))
     }
 
     #[instrument(skip_all)]
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -139,9 +139,7 @@ use sui_types::effects::{
     InputConsensusObject, SignedTransactionEffects, TransactionEffects, TransactionEffectsAPI,
     TransactionEvents, VerifiedSignedTransactionEffects,
 };
-use sui_types::error::{
-    ExecutionError, ExecutionErrorContext, ExecutionErrorMetadata, SuiErrorKind, UserInputError,
-};
+use sui_types::error::{ExecutionError, ExecutionErrorMetadata, SuiErrorKind, UserInputError};
 use sui_types::event::EventID;
 use sui_types::executable_transaction::VerifiedExecutableTransaction;
 use sui_types::execution_status::ExecutionErrorKind;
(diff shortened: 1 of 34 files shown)


crates/sui-core/src/authority/authority_store.rs
```diff
@@ -344,6 +344,13 @@ impl AuthorityStore {
             .get(digest)
     }
 
+    pub fn get_execution_error_metadata(
```
stefan-mysten, line 349: Let's add a doc comment here to explain what is this used for and in which cases?

crates/sui-core/src/execution_cache/writeback_cache.rs
```diff
@@ -233,6 +233,8 @@ struct UncommittedData {
 
     unchanged_loaded_runtime_objects: DashMap<TransactionDigest, Vec<ObjectKey>>,
 
+    execution_error_metadata: DashMap<TransactionDigest, ExecutionErrorMetadata>,
```
stefan-mysten, line 238: Let's add a doc comment explaining why this field exists and how it's used.

crates/sui-core/src/execution_cache.rs
```diff
@@ -534,6 +534,11 @@ pub trait TransactionCacheRead: Send + Sync {
         digest: &TransactionDigest,
     ) -> Option<Vec<ObjectKey>>;
 
+    fn get_execution_error_metadata(
```
stefan-mysten, line 537: Doc comment.

crates/sui-core/src/authority.rs
```diff
@@ -2159,6 +2197,10 @@ impl AuthorityState {
                 &tracking_store.into_read_objects(),
             );
 
+        let execution_error_metadata = execution_error_opt
```
stefan-mysten: What's the possible size of this error metadata? If I recall correctly, this will be added onto fullnodes, I assume in a table somewhere. If it's very large, can it cause any disruption/issues during insertion/retrieval?

crates/sui-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
```diff
@@ -551,6 +552,39 @@ async fn test_committed() {
     .await;
 }
 
+#[tokio::test]
```
stefan-mysten, line 556: Great that we're having one test here.

I think we probably need a few more, at least one for when catching the non empty branch of is_empty check. We probably also want to check that the validator produces no metadata and effects are the same. Maybe these tests already exist around here, but I am not familiar with the code.

crates/sui-core/src/authority.rs
```diff
@@ -1905,40 +1911,69 @@ impl AuthorityState {
...
     ) -> (
         InnerTemporaryStore,
         SuiGasStatus,
         TransactionEffects,
         Vec<ExecutionTiming>,
-        Result<(), ExecutionError>,
+        Result<(), ExecutionErrorContext>,
     ) {
-        let (inner_temp_store, gas_status, effects, timings, execution_error) = executor
-            // TODO only run this function on FullNodes, use `execute_transaction_to_effects` on validators.
```
stefan-mysten, line 1916: We used to have a comment here. Maybe it's good to add a new comment explaining the if/else branches differences, to make it more obvious what's going on.

crates/sui-core/src/authority.rs
```diff
@@ -1905,40 +1911,69 @@ impl AuthorityState {
...
-                store,
-                protocol_config,
-                self.metrics.execution_metrics.clone(),
-                enable_expensive_checks,
-                execution_params,
-                epoch_id,
-                epoch_timestamp_ms,
-                input_objects,
-                gas_data,
+        if is_fullnode {
```
stefan-mysten, line 1920: Just thinking aloud: I also wonder if there's a better way to handle this if/else given that most parameters are "duplicated" for both calls; I couldn't come up with anything really, as there are two different functions being called.

crates/sui-core/src/authority.rs
```diff
@@ -2025,7 +2060,7 @@ impl AuthorityState {
         let tracking_store = TrackingBackingStore::new(self.get_backing_store().as_ref());
 
         #[allow(unused_mut)]
-        let (inner_temp_store, _, mut effects, timings, execution_error_opt) = self
+        let (inner_temp_store, _, mut effects, timings, execution_error_result) = self
```
stefan-mysten: Could retain the original naming and just later do:

```rust
    let execution_error_opt = execution_error_opt.err();
```

crates/sui-core/src/execution_cache/writeback_cache.rs
```diff
@@ -2309,6 +2326,21 @@ impl TransactionCacheRead for WritebackCache {
...
+        digest: &TransactionDigest,
+    ) -> Option<ExecutionErrorMetadata> {
+        self.dirty
+            .execution_error_metadata
+            .get(digest)
+            .map(|b| b.clone())
+            .or_else(|| {
+                self.store
+                    .get_execution_error_metadata(digest)
+                    .expect("db error")
```
stefan-mysten, line 2342: This `.expect` leads to a panic, no? Is that required? I'd expect that this should return None if it cannot get execution_error_metadata rather than panicking.

stefan-mysten (commented):
Left a couple of comments. I am not familiar with this codebase so I hope someone else will take a look, but it was interesting to read and learn. Thanks @jordanjennings-mysten.

crates/sui-core/src/execution_cache/unit_tests/writeback_cache_tests.rs
jordanjennings-mysten: added negative case tests

crates/sui-core/src/execution_cache/writeback_cache.rs
jordanjennings-mysten: this was an existing pattern in `get_unchanged_loaded_runtime_objects` if theres a `TypedStoreError` it bails which arguably should panic since that error type is associated serious db errors. I do agree probably something to flag though.

crates/sui-core/src/authority.rs
jordanjennings-mysten: I think at best you can write a macro. at some point you have to decide on what the generic error type is.

crates/sui-core/src/authority.rs
jordanjennings-mysten: my napkin math says about 1 gb of storage per week with 300 bytes per error, 2% error rate on ~20 million txs a day, depending on when the fullnode prunes it could be a bit smaller, maybe more in the 0.33gb range. based on a 5 hour sample window a few weeks back.

stefan-mysten: Thanks for sharing more concrete numbers. I was wondering if there's any limits in the DB of how much data can be inserted in a field in a column?

crates/sui-core/src/authority.rs
```diff
@@ -1905,40 +1911,69 @@ impl AuthorityState {
...
-            effects,
-            timings,
-            execution_error,
-        )
+            (
+                inner_temp_store,
+                gas_status,
+                effects,
+                timings,
+                execution_failure.map_err(ExecutionErrorContext::from),
```
amnn: Would it be worthwhile to split the extra metadata into its own distinct type? I.e. have `Authority::execute_transaction_to_effects` return `ExecutionError` and then a separate, optional `ExecutionErrorContext` or `ExecutionMetadata` (whatever you want to call the extra information only)?

The benefit (for me) of using separate types, is that I would be able to clearly see which part of the richer data is dropped in the validator path, because it would be statically `None` here.

This also fits the onward data flow better because the execution error would need to be indexed with the rest of the base transaction effect data, while the metadata goes into its own table.

sui-execution/src/executor.rs
```diff
@@ -79,7 +79,7 @@ pub trait Executor {
         SuiGasStatus,
         TransactionEffects,
         Vec<ExecutionTiming>,
-        Result<(), ExecutionError>,
+        Result<(), ExecutionErrorContext>,
```
amnn: Should `dev_inspect_transaction` also use this type?

amnn (commented):
This PR looks good to me, just some questions about the types.

I'm also curious what the roll out for this would be, because if you roll out just this change, and then iterate on the error messaging itself, we will end up with different fullnodes in the ecosystem producing different metadata for the same error based on the version of `sui-node` they are running.

The code otherwise looks good, thanks @jordanjennings-mysten

crates/sui-core/src/authority.rs
bmwill: I would also recommend we try to align on the shape of the type using a protobuf message and storing the protobuf message itself in the DB so that we can iterate on the shape if need be.

sui-execution/src/executor.rs
jordanjennings-mysten: yup I was going to split that out into its own PR to keep this focused

crates/sui-core/src/authority.rs
amnn: @bmwill, is it possible to do that today with TideHunter/TypedStore? I was personally less worried about the evolvability because the metadata structure is already quite generic, but agree that if it's possible, then using a more generic protobuf value would be nicer for future evolvability.

crates/sui-core/src/authority.rs
jordanjennings-mysten: I think they are quite large but you would hit performance problems at some point since its rocks db.

crates/sui-core/src/authority.rs
jordanjennings-mysten: put prost and serialize and it seemed to work out, I expect we will add some protos somewhere at some point but for now just used the prost derive.

jordanjennings-mysten:
> I'm also curious what the roll out for this would be, because if you roll out just this change, and then iterate on the error messaging itself, we will end up with different fullnodes in the ecosystem producing different metadata for the same error based on the version of sui-node they are running.

This is an interesting question. I hadn't considered versioning/roll out. protocol version certainly feels overkill, I did have an interest in the past about feature flagging (unclear if that would work here originally for cli) but it seemed there was a preference to not introduce flags. I'll think about it some more.

crates/sui-types/src/error.rs
```diff
@@ -1166,7 +1166,38 @@ impl std::fmt::Debug for SuiError {
...
 pub(crate) type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
-pub type ExecutionErrorMetadata = BTreeMap<String, String>;
+
+#[derive(Clone, Eq, PartialEq, JsonSchema, Serialize, Deserialize, prost::Message)]
+pub struct ExecutionErrorMetadata {
+    #[prost(btree_map = "string, string", tag = "1")]
+    pub attributes: BTreeMap<String, String>,
+}
+
+#[cfg(test)]
```
amnn: Having tests in the middle of the implementation is a bit unconventional, can we move these to the end and just call the module `tests`?

crates/sui-types/src/error.rs
```diff
@@ -1166,7 +1166,38 @@ impl std::fmt::Debug for SuiError {
 }
 
 pub(crate) type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
-pub type ExecutionErrorMetadata = BTreeMap<String, String>;
+
+#[derive(Clone, Eq, PartialEq, JsonSchema, Serialize, Deserialize, prost::Message)]
```
amnn: The fact that this type implements `prost::Message` does not mean that it's being serialized using proto when stored in the authority stores/indices -- I would expect that it's still being BCS encoded (with all the pitfalls around evolvability that this entails) unless you are manually encoding it to proto before writing it, which I didn't see happening in the code above.

This is where my earlier question to @bmwill came from -- whether there is an established pattern for storing proto into tidehunter/`typed-store`.

sui-execution/src/executor.rs
amnn: Now that I understand the rest of this change better, I'm pretty sure you do not want to introduce `ExecutionErrorContext` at this level of abstraction. It bakes in the idea that you have access to a source error on exit from the execution layer when the aim is to abstract that away.

Instead, replicate the change inside authority here -- expose the error and the metadata, and make it the execution layer's responsibility to extract the necessary metadata.

You could package the error and metadata up into a new type, like you have here, or you could introduce the metadata as a new optional field, like in authority (which has smaller knock-on consequences).

amnn (commented):
I don't think the traits on `ExecutionErrorMetadata` are having the desired effect (which also means the tests that were added are not catching the thing you want them to catch -- schema evolvability). Can you take another look?

amnn:
I ended up chatting with @bmwill about this, here's a summary from that discussion:

- The reasons to use proto overlap with the reasons to keep this type generic (which drove the current design where we have a `string -> string` map). It doesn't make much sense to encode the `string -> string` map as proto, instead we should define a structured metadata type in proto, and store it as encoded proto to take advantage of the format's type safety and schema evolvability.
- To do this properly, the metadata type needs to be defined in a `.proto` file, with CI support to check for schema evolution issues, and then we can generate the Rust definition from it. This `.proto` file would not contain a generic string-to-string map, it would contain the exact fields  we want to expose (for now, let's just say a `message: string`)
- This could be folded into `sui-rpc-api`, or it could be a separate thing (although then you would need to rebuild all that CI infra).
- When it comes to roll-out, I would say that we should iterate on this until we are happy with the first E2E use case (a human readable message), and then we can look into backfilling by replaying transactions -- cc @tzakian, this is similar to the stuff we were talking about last Wednesday for the dual-layer execution stuff.

jordanjennings-mysten:
ignore the review request.. I didn't realize you had comments since githubs notifications seem to have failed and didn't tell me you had commented. I'll take a look!

jordanjennings-mysten:
> we can look into backfilling by replaying transactions -- cc @tzakian, this is similar to the stuff we were talking about last Wednesday for the dual-layer execution stuff.

can we not build up the index first then ship the API?

jordanjennings-mysten:
ExecutionErrorContext no longer leaks through the executor interface, and execution now returns ExecutionError plus optional metadata. the DB path now stores encoded sui-rpc proto bytes in Vec<u8> instead of BCS-encoding the local metadata type, with conversions at the read/write functions. 

now relies on:
https://github.com/MystenLabs/sui-apis/pull/28
and
https://github.com/MystenLabs/sui-rust-sdk/pull/267

crates/sui-core/src/authority/authority_store_tables.rs
```diff
@@ -100,6 +100,9 @@ pub struct AuthorityPerpetualTables {
     // Loaded (and unchanged) runtime object references.
     pub(crate) unchanged_loaded_runtime_objects: DBMap<TransactionDigest, Vec<ObjectKey>>,
 
+    // Local execution error metadata, keyed by the digest of the transaction that produced it.
```
amnn, line 103: Let's document what these bytes are here.

crates/sui-core/src/authority/authority_store_tables.rs
```diff
@@ -932,3 +945,61 @@ fn effects_table_config(db_options: DBOptions) -> DBOptions {
...
+    use sui_types::error::ExecutionErrorMetadata;
+    use typed_store::DBMapUtils;
+    use typed_store::rocks::{DBMap, MetricConf};
+    use typed_store::traits::Map;
+
+    #[derive(Clone, PartialEq, prost::Message)]
+    struct FutureExecutionErrorMetadata {
+        #[prost(string, optional, tag = "1")]
+        message: Option<String>,
+        #[prost(string, optional, tag = "2")]
```
amnn: Make this tag very large so you don't have to keep updating it whenever you add future fields.

amnn: Actually looking at this test, I'm not sure it's adding much, because it requires redefining all the stuff that's in the perpetual tables, so if you were to introduce a bug there, this test would not catch it.

You could write this as a test that encoding a newer version of the type to bytes and decoding it as the existing version still works, but that is really testing `prost`, which seems not so useful.

crates/sui-core/src/authority.rs
```diff
@@ -2339,7 +2390,7 @@ impl AuthorityState {
         let execution_error_source = execution_result
             .as_ref()
             .err()
-            .and_then(|e| e.source().as_ref().map(|e| e.to_string()));
+            .and_then(|e| std::error::Error::source(e).map(|e| e.to_string()));
```
amnn: Still necessary?

crates/sui-types/src/rpc_proto_conversions.rs
```diff
@@ -1184,6 +1185,32 @@ impl From<crate::execution_status::ExecutionStatus> for ExecutionStatus {
 // ExecutionError
 //
 
+pub use sui_rpc::proto::sui::rpc::v2::ExecutionErrorMetadata;
```
amnn: nit: Let's move this into its own section? It's currently encroaching on `ExecutionError`. We can put it after.

sui-execution/src/latest.rs
```diff
@@ -14,7 +14,10 @@ use sui_types::{
     committee::EpochId,
     digests::TransactionDigest,
     effects::TransactionEffects,
-    error::{ExecutionError, ExecutionErrorTrait, SuiError, SuiResult},
+    error::{
+        ExecutionError, ExecutionErrorContext, ExecutionErrorMetadata, ExecutionErrorTrait,
```
amnn: Is `ExecutionErrorContext` still providing value now, and if so, does it still belong in `sui-type::error` or is it part of the the execution layer now?

sui-execution/src/v0.rs
```diff
@@ -228,7 +229,19 @@ impl executor::Executor for Executor {
...
+            .as_ref()
+            .err()
+            .and_then(ExecutionErrorContext::metadata_with_source);
+        (
+            inner_temp_store,
+            gas_status,
+            effects,
+            vec![],
+            result.map_err(ExecutionError::from),
+            execution_error_metadata,
```
amnn: Based on the discussion yesterday, shouldn't this be a blanket `None` for errors coming from older execution layers?

amnn (approved):
Looks good to me, but still good to get a review from @bmwill as he is much more familiar with the inner workings on fullnodes.

crates/sui-core/src/authority/authority_store_tables.rs
jordanjennings-mysten: This was originally written as a failing test against the previous commit, I'll take a closer look might have gone wrong along the way.

jordanjennings-mysten: ultimately I think its plainly obvious from the DB type we have forwards compatibility so I think I'll just drop the test.

sui-execution/src/latest.rs
jordanjennings-mysten: - made box error public
- need an into_parts() now on execution error otherwise we would need to pub ExecutionErrorInner which seems the wrong idea here
otherwise moves over cleanly and now seems like the right move since we have pushed context down.

PR #26608, opened 2026-05-12
[move-ide] Generic macro type param processing fix

awelc:
## Description 

This PR fixes a problem related to analyzing macros expanded multiple times in the user code, sometimes with generic and sometimes with non-generic types.

The gist of the problem was that compiler information gathered in the IDE mode for macro expansion was dropped when translated to a data structure on the `move-analyzer` side. The reason for it was that only one macro expansion information was preserved, as it was keyed on original macro function location.

In the process, another bug was fixed - macro body analysis is guarded by the `traverse_only` flag which was not maintained correctly during IR traversals. 

## Test plan 

Added a new test exposing the macro expansion bug (assertion failure in debug mode)

### DIFF
--- a/external-crates/move/crates/move-analyzer/src/analysis/typing_analysis.rs
+++ b/external-crates/move/crates/move-analyzer/src/analysis/typing_analysis.rs
@@ -63,6 +63,9 @@ pub struct TypingAnalysisContext<'a> {
     pub expression_scope: OrdMap<Symbol, LocalDef>,
     /// IDE Annotation Information from the Compiler
     pub compiler_analysis_info: &'a CompilerAnalysisInfo,
+    /// Tracks the root (outermost) macro call location for the current expansion chain.
+    /// Used to disambiguate MacroCallInfo lookups for nested macro calls.
+    pub current_macro_call_loc: Option<Loc>,
 }
 
 /// Definition of a local (or parameter)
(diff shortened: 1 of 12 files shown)


external-crates/move/crates/move-analyzer/src/analysis/typing_analysis.rs
```diff
@@ -1199,49 +1202,129 @@ impl TypingVisitorContext for TypingAnalysisContext<'_> {
...
-                | TE::Annotate(_, _)
-                | TE::UnresolvedError => false,
+                // All remaining expression types are handled here rather than
+                // delegating to the default visit_exp traversal. This is necessary
+                // because the macro and lambda handlers in visit_exp_custom set state
+                // (traverse_only, current_macro_call_loc) that must persist throughout
+                // the entire expression traversal. If visit_exp_inner returns false,
+                // visit_exp_custom also returns false, and the default visit_exp
+                // traversal runs AFTER the state has been restored — defeating the
+                // macro/lambda state management.
```
cgswords, line 1212: I am struggling with this design for this approach; do we want _two_ visitors where the outer one dispatches to the inner one?

cgswords (approved):
Seems fine, but I have some questions about the structure.

external-crates/move/crates/move-analyzer/src/analysis/typing_analysis.rs
awelc: I agree that this looks a bit awkward, but this the gist of a fix for the secondary problem discovered when creating this PR (which was found when trying to implement a fix for the primary problem). The `visit_exp_inner` was is responsible for tracking and recording use-def-s when visiting expressions, but needs some flags set-up and reset before and after a given expression is visited. In the previous version of the code, in `visit_exp_inner` we only visited expressions relevant to creating use-def-s and attempted to use default visitor for everything else:
```
            ... // a fragment of `visit_exp_custom`
            self.traverse_only = true;
            let result = visit_exp_inner(self, exp);
            self.traverse_only = old_traverse_mode;
            result
            ...
```
In the code above, if `visit_exp_inner` returned `false` it meant that we would  delegate to the default visitor (as per value of `result` returned from `visit_exp_custom` which is `visit_exp_inner` caller) as the expression did not require special handling.

The problem with this was that, for example, `Block` did not require special handling, was delegated to default visitor, but then would be processed under `old_traverse_mode` rather  that under `true` for ` self.traverse_mode`  which is not what was intended - if we are in the middle of macro expansion we want to process this block (and the rest of the macro code being expanded) under `true`. 

That's why we no longer delegate to default visitor as I could not figure out another way to do it such that the value of `self.traverse_only` is preserved correctly. Instead, we do indeed re-implement default handling for the remaining expression variants in the "custom" visitor.

PR #26610, opened 2026-05-12
sui-core: drop fastpath guard in object_funds_checker

lxfind:
## Summary

With fastpath execution removed, the `let Some(accumulator_version) = ... else { return false }` guard in `ObjectFundsChecker::should_commit_object_funds_withdraws` was only reachable from a single test that explicitly bypassed scheduling. A tx with object withdraws can only exist when accumulators are enabled for the epoch, and every production path that produces such a tx also assigns an accumulator version, so the guard is no longer load-bearing.

- Replace the guard with `.expect(...)` on `assigned_versions.accumulator_version`
- Remove the now-obsolete `test_object_withdraw_fast_path_abort` integration test
- Clarify the doc on `AssignedVersions::accumulator_version` to enumerate the legitimate `None` cases (accumulator-disabled epoch, end-of-epoch tx) instead of the stale "TODO: make required" note

## Test plan

- [ ] `cargo nextest run -p sui-core` passes
- [ ] `cargo xclippy -p sui-core` clean

### DIFF
--- a/crates/sui-core/src/accumulators/object_funds_checker/mod.rs
+++ b/crates/sui-core/src/accumulators/object_funds_checker/mod.rs
@@ -6,7 +6,7 @@ use std::{
     sync::Arc,
 };
 
-use mysten_common::assert_reachable;
+use mysten_common::{assert_reachable, debug_fatal};
 use parking_lot::RwLock;
 use sui_types::{
     SUI_ACCUMULATOR_ROOT_OBJECT_ID,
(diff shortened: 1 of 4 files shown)


crates/sui-core/src/accumulators/object_funds_checker/mod.rs
```diff
@@ -120,16 +120,14 @@ impl ObjectFundsChecker {
...
-            return false;
-        };
+        // A tx with object withdraws can only exist when accumulators are enabled
+        // for the epoch, and every production path that produces such a tx also
+        // assigns an accumulator version. The `None` paths (accumulator-disabled
+        // epoch, end-of-epoch tx) never produce withdraws and so never reach here.
+        let accumulator_version = execution_env
+            .assigned_versions
+            .accumulator_version
+            .expect("accumulator_version must be set for a tx with object withdraws");
```
mystenmark: can this be a debug_fatal! instead of an expect? Seems odd to give up the ability to tolerate a failure here.

lxfind: Good call - switched to `debug_fatal!` with a `return false` fallback so we keep the tolerance in release builds. The `false` matches the prior None-branch behavior (defer commit and let the tx be retried).

mystenmark (approved):

Issue #7872, opened 2023-01-29
New 5th Struct Ability: Protected


PaulFidika:
What if I wanted to create an asset that has both `key` and `store`; I like both abilities because `key` allows my asset to be a root-level asset that is easily accessible via entry functions, and because `store` gives me the flexibility to wrap my asset inside of another if I want to later on (such as putting it inside of a shared object).

But what if I _also_ want to ensure that (1) only my own module is able transfer ownership (no polymorphic transfer), and (2) only my own module is able to store (wrap) the asset? That is to say, any Sui-ownership transfer and wrapping must be done from within the defining module itself. I think this can be accomplished by preventing this struct from being passed-by-value into functions outside of the defining module (and defining-module friends) itself. Without pass-by-value ability, it would be impossible for other functions to (1) wrap my asset, and (2) use polymorphic transfer with my asset (because sui::transfer::transfer is outside of the defining module).

I recognize that most assets in Move are built around composability, but this is more of a non-composability requirement. We could add a 5th struct ability, called `protected` (temporary suggested name) that enables this sort of protected behavior; it's more of an anti-ability than an ability, in the sense that it removes functionality rather than adding it (like the other abilities do), which makes composability the default setting for assets.

PaulFidika:
Another way of implementing this would be to have `store` and `protected-store` as mutually exclusive abilities. `store` works as usual, but `protected-store` works as if `store` does not exist, _except_ for when the call originates from within the defining module. Something like:

```
struct MyItem has key, protected-store {
   id: UID
}

public entry fun store(my_item: MyItem) {
   external_module::store_item(my_item);
}
```

This call would be valid if made from within the defining module, but invalid if it originated from somewhere else. The defining module can wrap its own assets, or call out to other functions to wrap the asset for it, but the asset cannot be wrapped generally outside of these two cases (as is the case with regular `store`).

amnn:
cc @tnowacki, this seems similar to the `internal` modifier we were discussing earlier.

PaulFidika:
> cc @tnowacki, this seems similar to the `internal` modifier we were discussing earlier.

Will the `internal` modifier be an ability, or how were you thinking of implementing it? Okay so the probability of this actually getting built sometime soon is pretty high? Okay that's awesome.

Because with the Capsules standard I'm working on, the flow would be kind of like:

- Creator creates asset, makes it a key + internal-store asset, gives it to owner
[...]

PR #26611, opened 2026-05-12
docker: cache-friendly Dockerfile rewrites for Tier 2 + Tier 3 images

ebmifa:
## Summary

Apply the same cache-friendly Dockerfile pattern from #26601 (Tier 1) to the remaining mp3-built Rust images.

### Tier 2 — release-tracked binaries (binary_upload_config entries)
- `stress` → `s3://sui-releases/<sha>/stress`
- `sui-proxy` → `s3://sui-releases/<sha>/sui-proxy`

### Tier 3 — auxiliary indexers + tools
- `sui-analytics-indexer`
- `sui-bridge-indexer`
- `sui-bridge-indexer-alt`
- `sui-checkpoint-blob-indexer`
- `sui-kvstore`
- `sui-rpc-benchmark`

## What changes per Dockerfile

- `COPY rust-toolchain.toml ./` + `COPY .cargo .cargo` at the top so the toolchain/config layer is cached separately from sources.
- Workspace COPYs reordered least- → most-frequently-changing: `external-crates → sui-execution → consensus → crates`. Source edits in `crates/` no longer invalidate the heavier upstream layers.
- `ARG GIT_REVISION` + `ENV GIT_REVISION=$GIT_REVISION` moved to **after** the COPYs (right before `RUN cargo build`) for the six binaries that read it at compile time (`sui-proxy`, `bridge-indexer`, `bridge-indexer-alt`, `sui-analytics-indexer`, `sui-checkpoint-blob-indexer`, `sui-rpc-benchmark`).
- For `stress` and `sui-kvstore-alt`, which do **not** read `GIT_REVISION` at compile time, the builder stage doesn't declare it at all — the cargo-build layer cache key stays stable across commits when source is unchanged. `stress`'s **runtime** stage keeps `ARG GIT_REVISION` since it's used to `git checkout` the sui repo at deploy time (a runtime concern that doesn't affect the cargo cache).
- `cargo build` now uses `--locked`.
- apt-get steps consolidated to a single layer with `--no-install-recommends` and apt-list cleanup (builder + runtime).

### Image-specific

- `sui-proxy`, `sui-analytics-indexer`, `sui-checkpoint-blob-indexer`, `sui-kvstore`: drop unused `COPY .git/ .git/`. Per-commit `.git/` churn was invalidating every source layer above it. Verified safe via grep — none of those four crates' `src/` or `build.rs` read `.git` (no `git2`, `vergen`, `env!(\"GIT_REVISION\")`, or `process::Command::new(\"git\")`). Tier 0/1 images already build without `.git/` available, confirming no transitive workspace dependency reads it either.

## Codegen note (same as #26601, worth re-flagging)

Adding `COPY .cargo .cargo` means these images now honor the workspace's `.cargo/config.toml`, which carries `[build] rustflags = [\"-C\", \"force-frame-pointers=yes\", \"-C\", \"force-unwind-tables=yes\", \"-A\", \"unknown_lints\", \"-A\", \"mismatched_lifetime_syntaxes\"]`.

Before this PR, the eight rebuilt images compiled without frame pointers and without unwind tables. After this PR, they match `sui-node` / `sui-tools` / Tier 1 images. Observable effects:

- pprof / perf / flamegraph stack-walking now works on these images (improvement).
- Proper backtraces in panic handlers (improvement).
- ~1-2% runtime cost from frame pointers and a small image-size bump from unwind tables (acceptable; consistent with the rest of the fleet).

Anyone bisecting a future perf change on these images should know this commit aligns codegen, not just layer caching.

## Why

Pairs with mp3's rewriter + stateful_pool — both have been deployed to production (sui-operations #7872). With the canonical preamble in place, the rewriter's injected `--mount=type=cache,target=/sui/target,id=<cache_family>` mount has a stable cache key across SHA-only commits, and the cargo-target PVC bucket can be reused effectively. Tier 1 saw ~11× cold→warm speedup on sui-node; same shape expected here.

## Test plan

- [ ] CI Docker image builds succeed for all 8 images
- [ ] Confirm post-merge mp3 builds for these images use the rewriter and pick a non-empty `cache_family` (visible in `mp3_rewriter_applied_total{cache_family=...}`)
- [ ] Spot-check one build's `sticky_dispatch_total{result=\"hit\"}` increment after a second SHA on the same `cache_family`
- [ ] Sanity-check one of the new binaries still runs (e.g. `sui-proxy --version`, `sui-analytics-indexer --help`) — frame-pointer / unwind-table codegen change shouldn't affect external behavior, but cheap to verify

### DIFF
--- a/docker/stress/Dockerfile
+++ b/docker/stress/Dockerfile
@@ -1,17 +1,23 @@
 FROM rust:1.92-bullseye AS builder
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends cmake clang libpq5 libpq-dev \
+ && rm -rf /var/lib/apt/lists/*
 ARG PROFILE=release
-ARG GIT_REVISION
-ENV GIT_REVISION=$GIT_REVISION
-WORKDIR "$WORKDIR/sui"
-RUN apt-get update && apt-get install -y cmake clang libpq5 libpq-dev
+WORKDIR /sui
 
+# COPYs ordered least- to most-frequently-changing for deeper layer cache hits.
+COPY rust-toolchain.toml ./
+COPY .cargo .cargo
 COPY Cargo.toml Cargo.lock ./
+COPY external-crates external-crates
+COPY sui-execution sui-execution
 COPY consensus consensus
 COPY crates crates
-COPY sui-execution sui-execution
-COPY external-crates external-crates
 
-RUN cargo build --profile ${PROFILE} --bin stress
+# stress binary doesn't use GIT_REVISION at compile time, so we deliberately
+# don't declare ARG/ENV GIT_REVISION in the builder — keeps the cargo-build
+# layer cache key stable across commits when source content is unchanged.
+RUN cargo build --locked --profile ${PROFILE} --bin stress
 
 FROM rust:1.92-bullseye
 ARG PROFILE=release
@@ -27,7 +33,7 @@ RUN git clone https://github.com/MystenLabs/sui.git ; \
         git checkout $GIT_REVISION ; \
         cd .. ; \
         mv sui/* .
-        
+
 RUN mkdir -p /docker/stress
 RUN cp docker/stress/entrypoint.sh /docker/stress/entrypoint.sh
 RUN chmod +x /docker/stress/entrypoint.sh
--- a/docker/sui-analytics-indexer/Dockerfile
+++ b/docker/sui-analytics-indexer/Dockerfile
@@ -14,9 +14,9 @@ COPY sui-execution sui-execution
 COPY consensus consensus
 COPY crates crates
 
-# GIT_REVISION declared late so SHA changes don't invalidate the COPYs above.
-ARG GIT_REVISION
-ENV GIT_REVISION=$GIT_REVISION
+# sui-analytics-indexer doesn't use GIT_REVISION at compile time, so we
+# deliberately don't declare ARG/ENV GIT_REVISION in the builder — keeps the
+# cargo-build layer cache key stable across commits with identical source.
 RUN cargo build --locked --profile ${PROFILE} --bin sui-analytics-indexer
 
 FROM debian:12-slim AS deploy
--- a/docker/sui-analytics-indexer/Dockerfile
+++ b/docker/sui-analytics-indexer/Dockerfile
@@ -1,19 +1,26 @@
-FROM rust:1.92-bullseye as build
-
+FROM rust:1.92-bullseye AS builder
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends cmake clang \
+ && rm -rf /var/lib/apt/lists/*
 ARG PROFILE=release
 WORKDIR /work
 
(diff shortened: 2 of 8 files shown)


ebmifa:
Both findings addressed in \`7f99fd50e6\`:

### P2 — sui-proxy version fallback

You're right — the old Dockerfile's \`COPY .git/\` was the silent fallback path for \`bin_version::bin_version!()\`. Dropping it without enforcing the build arg would silently embed \"\" as the revision in plain \`docker build\` runs.

Added an explicit guard right before \`ENV GIT_REVISION=\$GIT_REVISION\` and \`RUN cargo build\`:

\`\`\`dockerfile
ARG GIT_REVISION
RUN test -n \"\$GIT_REVISION\" || \\
    (echo \"GIT_REVISION build arg is required (e.g. --build-arg GIT_REVISION=\\\$(git rev-parse HEAD))\" >&2 && exit 1)
ENV GIT_REVISION=\$GIT_REVISION
RUN cargo build --locked --profile \${PROFILE} --bin sui-proxy
\`\`\`

Same shape we used on \`MystenLabs/walrus#3372\` for walrus-proxy — fails loudly instead of producing a binary with an empty SHA. mp3 always passes this build arg; CI for sui likewise; only plain local \`docker build\` users would now see the error, which is the right behavior.

### P3 — dropped unused GIT_REVISION from 5 builder stages

You're also right that only sui-proxy actually reads GIT_REVISION at compile time (via \`bin_version::bin_version!()\` in \`crates/sui-proxy/src/main.rs\`). My earlier audit was based on a sloppy grep that matched the Cargo workspace import rather than actual compile-time usage in source. Re-checked each binary's entrypoint:

| Binary | Entrypoint | Reads GIT_REVISION? |
|---|---|---|
| sui-proxy | crates/sui-proxy/src/main.rs | ✅ via \`bin_version!()\` |
| bridge-indexer | crates/sui-bridge-indexer/src/main.rs | ❌ |
| bridge-indexer-alt | crates/sui-bridge-indexer-alt/src/main.rs | ❌ |
| sui-analytics-indexer | crates/sui-analytics-indexer/src/main.rs | ❌ |
| sui-checkpoint-blob-indexer | crates/sui-checkpoint-blob-indexer/src/main.rs | ❌ |
| sui-rpc-benchmark | crates/sui-rpc-benchmark/src/main.rs | ❌ |
| stress | crates/sui-benchmark/src/bin/stress.rs | ❌ |
| sui-kvstore-alt | crates/sui-kvstore/src/bin/sui-kvstore-alt.rs | ❌ |

Removed \`ARG GIT_REVISION\` / \`ENV GIT_REVISION\` from the 5 builder stages that ignore the value. \`stress\` and \`sui-kvstore-alt\` were already declaration-free in the original commit. Runtime stages of bridge-indexer / bridge-indexer-alt / sui-rpc-benchmark keep their own \`ARG GIT_REVISION\` + \`LABEL git-revision=\$GIT_REVISION\` — runtime labels don't affect the cargo-build cache.

Ready for re-review.

johnjmartin (approved):

PR #26612, opened 2026-05-12
use std::process::command in simtest when fetching move package dependency

halfprice:
## Description 

During simtest setup, we need to fetch base sui dependency. However, the new package management switched to use tokio::command, which cannot be run during simtest initialization, where tokio environment is not set up yet.

Walrus repo has been using a hack where we manually fetch sui dependency before starting the test, but this hurts CI
and debugability.

The change here is basically use std::process::command **only in simtest** to fetch dependency.

I have verified that this works by patching this PR in Walrus repo and remove the hack.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-package-alt/src/git/cache.rs
+++ b/external-crates/move/crates/move-package-alt/src/git/cache.rs
@@ -17,7 +17,6 @@ use std::{
     process::Stdio,
     sync::OnceLock,
 };
-#[cfg(not(msim))]
 use tokio::process::Command;
 use tracing::debug;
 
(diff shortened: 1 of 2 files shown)


mdgeorge4153 (commented):
This seems like a reasonable thing to do, but I'd prefer less code duplication. I think you can use `tokio::process::Command::into_std` to convert the constructed tokio command into an std command, so you I think you could just call `cmd.into_std().output()` in msim mode and keep the rest of the function the same.

Separately, I'm wondering if you tried `sui move cache-package` for your workaround or were doing something more invasive.

halfprice:
> This seems like a reasonable thing to do, but I'd prefer less code duplication. I think you can use `tokio::process::Command::into_std` to convert the constructed tokio command into an std command, so you I think you could just call `cmd.into_std().output()` in msim mode and keep the rest of the function the same.
> 
> Separately, I'm wondering if you tried `sui move cache-package` for your workaround or were doing something more invasive.

Thank you @mdgeorge4153 for the comment! After looking closely, I think the change can be much simplified!

stefan-mysten (approved):

mdgeorge4153 (approved):
looks great!

PR #26615, opened 2026-05-12
[decompiler] Total post-dominators over divergent regions

cgswords:
## Description 

`structure_acyclic_region` asks petgraph for the immediate post-dominator of a node and unwraps it. petgraph's `simple_fast` leaves `immediate_dominator` undefined for any node not reachable from the dominator-tree root — and in the reversed CFG rooted at `return_`, that's exactly the input nodes that never reach a sink in the original (infinite loops with no exit, abort-only regions).

This make post-domination total over the input by adding *impossible edges*: after wiring `return_` to original sinks, compute SCCs of the original CFG and, for each terminal SCC (one whose outgoing edges all stay inside the SCC), add `return_` -> highest-rank-input-node-in-SCC. Singleton-sink SCCs slip through the check vacuously and get a redundant edge, but simple_fast handles the duplicate fine.

## Test plan 

Adds three structuring unit tests for the shapes that exercised the fix (lone self-loop; if/else that converges and then falls into a trailing infinite loop; if/else whose two arms each fall into a separate infinite loop, producing two disjoint terminal SCCs).

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


tzakian (approved):
Fancy! Pretty cool though

PR #26622, opened 2026-05-13
Disallow extraneous tx signatures

aschran:
Cleanup of the rules for verifying signatures on transactions that name aliased addresses among their required signers.

Today the verifier requires `tx_signatures.len() == required_signers.len()` and then resolves each required signer through the alias map, greedily taking the first signature whose recovered address is one of the signer's aliases. That greedy assignment is never checked for being one-to-one: two required signers can be satisfied by the same signature, which leaves another provided signature bound to no required signer at all.

Behind a new `fix_aliased_signer_signatures` flag (protocol v125, all chains), the verifier instead requires a one-to-one matching between required signers and provided signatures — each required signer matched to a distinct signature whose recovered address is one of its aliases. The count check is kept, so a successful matching is necessarily perfect (every signature bound to exactly one required signer). The matching is computed with Kuhn's algorithm so a valid assignment is found whenever one exists, even when signers' alias sets overlap. The legacy greedy path is kept for pre-flag behavior.

## Test plan
- [x] `cargo nextest run -p sui-protocol-config -p sui-types -p sui-core --lib` (one unrelated pre-existing flake in `jsonrpc_index::tests`, a `temp_dir()` RocksDB lock race that also fails on `main`)
- [x] `cargo xclippy -D warnings` on every modified crate
- [x] `cargo fmt --all -- --check`
- [x] Affected snapshots regenerated (protocol-config, swarm-config, open-rpc, execution-time-estimator v125, graphql SDL).


crates/sui-types/src/signature_verification.rs
```diff
@@ -197,3 +183,388 @@ pub fn verify_sender_signed_data_message_signatures(
...
+        .map_err(|signer_idx| signer_signature_absent(&alias_sets[signer_idx], sig_mapping))
+}
+
+/// Computes a one-to-one matching of signers to signature indices using Kuhn's
+/// algorithm. `candidates[i]` is the set of signature indices signer `i` may be
+/// matched to. On success returns the matched index for each signer, in input
+/// order; on failure returns the index of the first signer that cannot be matched.
+///
+/// Kuhn's algorithm augments one signer at a time, so a matching is found whenever
+/// one exists. The result is a deterministic function of `candidates`.
```
mystenmark, line 283: i think this is overkill because unequal signatures cannot have intersecting address sets. This algorithm seems to solve the case where:

```
requird_signers = [a, b]
addresses_of(s1) = {a, b}
addresses_of(s2) = {a}
```

in this case, if you start out assigning s1 => a, you will have to backtrack in order find the only possible solution which is s1 => b, s2 => a

But this situation can't happen. Equal sigs have equal addresses, Distinct sigs have disjoint addresses.

aschran: distinct sigs have disjoint addresses, but the the alias sets can still overlap, so  I think we still need the backtracking part for that? or which thing do you think is overkill and can be removed?

ex, imagine:
aliases_of(A) = {K}
aliases_of(B) = {K, B}

tx with:
required_signers = [A, B] 
signatures = [K, B]

we want this to result in mapping [0, 1] but the current algorithm would map it as [0, 0] (which mapping we actually want to reject after this fix)

I thought about hardcoding it to assume max of 2 required signatures, but that didn't actually make it that much simpler, and this way if we add additional signatures for some purpose in the future it doesn't require any further fixing

PR #26629, opened 2026-05-13
[chore] fix cargo-deny issue with `diesel-async`

stefan-mysten:
## Description 

This PR fixes the `diesel-async` cargo deny issue by bumping `diesel-async` to 0.9 and fixing the API call in `sui-pg-db` and `sui-bridge-indexer` crates.

## Test plan 

Existing CI.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -360,7 +360,7 @@ derive-syn-parse = "0.1.5"
 derive_more = "1.0.0"
 diesel = "2.3"
 diesel_migrations = "2.3"
-diesel-async = "0.8"
+diesel-async = "0.9"
 dirs = "4.0.0"
 duration-str = "0.5.0"
 ed25519 = { version = "2", features = ["pkcs8", "alloc", "zeroize"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4785,9 +4785,9 @@ dependencies = [
 
 [[package]]
 name = "diesel"
-version = "2.3.8"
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -360,7 +360,7 @@ derive-syn-parse = "0.1.5"
 derive_more = "1.0.0"
 diesel = "2.3"
 diesel_migrations = "2.3"
-diesel-async = "0.8"
+diesel-async = "0.9"
```
ebmifa, line 363: This workspace-wide bump likely needs one more callsite migration before merge. `crates/sui-bridge-indexer/src/storage.rs` still imports `diesel_async::scoped_futures::ScopedFutureExt` and uses the old `.transaction(|conn| { async move { ... }.scope_boxed() })` shape. With `diesel-async 0.9`, `scoped-futures` is no longer in the lockfile, so `sui-bridge-indexer` will likely fail to compile when that crate is built. Can we migrate that transaction to the new `async |conn| ...` API too?

ebmifa (commented):
One workspace-wide compatibility issue to fix before merge.

amnn (approved):
LGTM, thanks @stefan-mysten

PR #26633, opened 2026-05-13
[chore] fix clippy warnings in tests 

stefan-mysten:
## Description 

This PR gates a number of fields behind staging feature to avoid clippy warnings during tests. It also removes the sequence number argument as it can be derived from summary.

See such a warning here: https://github.com/MystenLabs/sui/actions/runs/25765937534/job/75678453235?pr=26500#step:5:30

## Test plan 

Existing CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
+++ b/crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
@@ -56,6 +56,7 @@
 //! the recovery target, the next live message is itself a gap, and detection
 //! fires again.
 
+#[cfg(feature = "staging")]
 use std::collections::BTreeMap;
 use std::sync::Arc;
 use std::time::Duration;
@@ -74,10 +75,13 @@ use sui_rpc::proto::sui::rpc::v2::Checkpoint as ProtoCheckpoint;
 use sui_rpc::proto::sui::rpc::v2::ExecutedTransaction as ProtoExecutedTransaction;
 use sui_rpc::proto::sui::rpc::v2::SubscribeCheckpointsRequest;
 use sui_rpc::proto::sui::rpc::v2::SubscribeCheckpointsResponse;
+#[cfg(feature = "staging")]
 use sui_rpc::proto::sui::rpc::v2::changed_object::OutputObjectState;
 use sui_rpc::proto::sui::rpc::v2::subscription_service_client::SubscriptionServiceClient;
 use sui_sdk_types::ValidatorAggregatedSignature;
+#[cfg(feature = "staging")]
 use sui_types::base_types::ObjectID;
+#[cfg(feature = "staging")]
 use sui_types::base_types::SequenceNumber;
 use sui_types::crypto::AuthorityStrongQuorumSignInfo;
 use sui_types::crypto::ToFromBytes;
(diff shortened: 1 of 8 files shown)


crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
amnn, line 1: This entire module should be gated by "staging", I think?

amnn: If that's not possible, it would be slightly neater to do something like this, underneath all the other imports, so that we group together the feature gating.

```rust
#[feature(staging)]
mod staging {
    pub use ...;
}

#[feature(staging)]
use staging::*;
```

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -379,6 +383,10 @@ pub(super) fn process_checkpoint(
         .context("Missing summary.bcs")?
         .deserialize()
         .context("Failed to deserialize checkpoint summary")?;
+    anyhow::ensure!(
+        summary.sequence_number == sequence_number,
+        "Checkpoint sequence_number does not match summary sequence_number"
+    );
```
stefan-mysten, line 389: This seems a bit over the top, so happy to remove it.

crates/sui-indexer-alt-graphql/src/task/streaming/mod.rs
amnn, line 1: Again, whole streaming task should be feature gated on `staging`.

crates/sui-indexer-alt-graphql/src/task/streaming/processed_checkpoint.rs
```diff
@@ -37,30 +40,33 @@ pub(crate) struct ProcessedTransaction {
 impl ProcessedCheckpoint {
     /// Construct a checkpoint with the digest index pre-built from `transactions`.
     pub(crate) fn new(
-        sequence_number: u64,
         summary: CheckpointSummary,
         contents: CheckpointContents,
         signature: AuthorityStrongQuorumSignInfo,
         transactions: Vec<ProcessedTransaction>,
-        execution_objects: ExecutionObjectMap,
+        #[cfg(feature = "staging")] execution_objects: ExecutionObjectMap,
```
amnn, line 47: This ...should not be only used in staging.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -379,6 +383,10 @@ pub(super) fn process_checkpoint(
         .context("Missing summary.bcs")?
         .deserialize()
         .context("Failed to deserialize checkpoint summary")?;
+    anyhow::ensure!(
```
amnn, line 386: Why not just ignore the `sequence_number` like you have elsewhere?

amnn (commented):
Thanks for cleaning the warnings up @stefan-mysten -- @tpham-mysten can you take a look? There should be ways to introduce feature gating to silence warnings but that doesn't require introducing annotations in so many different places. E.g.

- Feature gate whole modules out, 
- Keep the additions/generalisations to `Scope` that were added for subscriptions even though subscriptions doesn't exist yet (so that we can test that they work as generalisations for the query APIs.

stefan-mysten:
> * Feature gate whole modules out,
> 
> * Keep the additions/generalisations to `Scope` that were added for subscriptions even though subscriptions doesn't exist yet (so that we can test that they work as generalisations for the query APIs.

@amnn thanks for having a look. I am not familiar with this code but happy to have another look and coordinate with @tpham-mysten to see how we can fix it.

stefan-mysten:
Closing this as @tpham-mysten will take over.

PR #26635, opened 2026-05-14
[sui-fork] add back `ConsensusAddressOwner` support for owned objects and seeding

stefan-mysten:
## Description 

This PR adds back the ConsensusAddressOwner type that we need to support for owned objects and seeding. Initially, I started with it in the design, but in a later PR I removed it. This is needed as we can seed objects that are consensus "owned".

## Test plan 

New tests + existing CI.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-fork/design/owned_objects_design.md
+++ b/crates/sui-fork/design/owned_objects_design.md
@@ -72,8 +72,7 @@ struct OwnedObjectEntry {
 }
 ```
 
-Only Move objects owned by an address with a `StructTag` are indexed. This
-includes `AddressOwner` and `ConsensusAddressOwner`. Shared, immutable,
+Objets with `AddressOwner` and `ConsensusAddressOwner` types are kept in the index. Shared, immutable,
 object-owned, and package objects are excluded.
 
 ## Write Path
--- a/crates/sui-fork/design/owned_objects_design.md
+++ b/crates/sui-fork/design/owned_objects_design.md
@@ -72,8 +72,9 @@ struct OwnedObjectEntry {
 }
 ```
 
-Only address-owned Move objects with a `StructTag` are indexed. Shared,
-immutable, object-owned, consensus-owned, and package objects are excluded.
+Only Move objects owned by an address with a `StructTag` are indexed. This
+includes `AddressOwner` and `ConsensusAddressOwner`. Shared, immutable,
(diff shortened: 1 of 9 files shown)


crates/sui-fork/design/owned_objects_design.md
```diff
@@ -72,8 +72,9 @@ struct OwnedObjectEntry {
 }
 ```
 
-Only address-owned Move objects with a `StructTag` are indexed. Shared,
```
amnn, line 75: I'm not sure these updates to the design are necessary. Objects whose owners are `ConsensusAddressOwner` are still address-owned, they are just also managed by consensus.

The only update is to get rid of "consensus-owned" in this sentence, but that is less to do with this change and more to do with the fact that consensus does not own things -- there is no such thing as a "consensus-owned" object.

crates/sui-fork/src/gql/queries.rs
```diff
@@ -492,6 +498,18 @@ pub(crate) mod address_owned_objects_query {
...
+    fn parse_owner_address(
+        address: Option<OwnerAddress>,
+        object_id: &str,
+    ) -> Result<SuiAddressType, Error> {
+        address
+            .ok_or_else(|| anyhow!("address seed entry is missing owner address"))?
+            .address
+            .0
+            .parse::<SuiAddressType>()
+            .with_context(|| format!("invalid seed owner for object {object_id}"))
```
amnn:
```suggestion
            .context("address seed entry is missing owner address")?
            .address
            .0
            .parse()
            .with_context(|| format!("invalid seed owner for object {object_id}"))
```

amnn (approved):
Looks good -- I personally wouldn't change "address-owned" to "object owned by an address" everywhere, because they mean the same thing and a `ConsensusAddressOwner` object is also "address-owned".

Not a big deal, but it's adding noise to the PR that I think you can avoid.

crates/sui-fork/design/owned_objects_design.md
stefan-mysten: Thanks, fixed!

PR #26640, opened 2026-05-14
[move][move-decompiler] Reconstruct match expressions from variant switches

cgswords:
## Description

Adds a refinement that promotes a `Switch` into a `Match` when each arm starts with an `UnpackVariant` whose fields can be lifted into a pattern. Also emits explicit `let` prefixes for `Unpack` and `UnpackVariant` in the pretty-printer.

## Test plan

Snapshot test updates.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-decompiler/tests/move/enums/colors@full.snap
```diff
@@ -56,16 +54,15 @@ public fun is_green(l0: enums::colors::Color): bool {
 public fun is_red(l0: enums::colors::Color): bool {
     let l3 = l0;
     let l2 = &l3;
-    let l1 = match (l2) {
-        Color::Red => {
-            0x0::colors::Color::Red {} = l2;
-            0x0::colors::Color::Red {} = l3;
+    let l1 = match 0x0::colors::Color (l2) {
```
tzakian: Seems an odd way to pretty print this? Is this what you meant to print here?

external-crates/move/crates/move-decompiler/src/pretty_printer.rs
```diff
@@ -310,11 +310,37 @@ fn exp(context: &Context, exp: &Exp) -> Doc {
...
+                                .concat_space(D::text("{"))
+                                .concat_space(field_doc)
+                                .concat_space(D::text("}"));
+                        }
+                        pat.concat_space(D::text("=>"))
+                            .concat_space(e_block(context, body))
+                    }),
+                    D::text(",").concat(D::line()),
+                );
+                D::text(format!("match {mid}::{enum_}"))
```
tzakian: this seems wrong to me? Should just be `"match "` here I believe.

tzakian (approved):
Overall LGTM, although I think there is a bug in the pretty printer for `match` that should probably be fixed before landing.

PR #26643, opened 2026-05-14
[move][move-decompiler] Eliminate more gotos from conditional restructuring

cgswords:
## Description 

Eliminate the redundant goto-to-post-dom logic that drove most of the residual `goto`s (~77%) in decompiled output. The Condition arm logic now emits None when we own and inline-sequence the post-dom, so the dropped `goto` and the inline post-dom are no longer both emitted. Per-site provenance tags (GotoSource) ride along so we can attribute whatever residue is left.

Also fixes a latent bug where an arm whose target was outside our dominator subtree got silently dropped; it now emits a Jump  marker for the enclosing scope to rewrite. 

Fixes an additional bug with two assertions that overstated invariants; Move's compiler can emit Conditions with no immediate dom-children.

We also not quantify the kind of a Jump that is generated, for easier tracking and fixing the last of them.

## Test plan 

All 52 snapshot tests pass. Two new structuring regressions:
  - inner_alt_shared_with_outer.stt — formerly-dropped continue to outer loop now visible as jump<AOS>.
  - cond_with_no_ichildren.stt — branching node with both arms shared no longer panics in the dispatcher.

Corpus run drops residual-goto packages from ~44k to ~10k. 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-decompiler/src/structuring/ast.rs
+++ b/external-crates/move/crates/move-decompiler/src/structuring/ast.rs
@@ -30,6 +30,47 @@ pub enum Input {
     Code(Label, Code, Option<Label>),
 }
 
+/// Provenance for a surviving `Jump`/`JumpIf`. Each variant names the structurer path that
+/// created the goto; the tag rides through `insert_breaks` and is printed on stderr when a
+/// Jump is lowered to `Unstructured(Goto)` in `generate_output`, letting the corpus driver
+/// attribute residual gotos by source.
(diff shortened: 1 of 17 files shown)


tzakian (approved):
Looks reasonable to me

PR #26644, opened 2026-05-14
[move][move-decompiler] Emit loop bodies in reverse post-order

cgswords:
## Description 

Today, `insert_breaks` has a `LatchKind::InLoop` case that drops a `Jump` on the assumption that its target is the next sibling in the emitted Seq. With sort-by-NodeIndex, that's true only because `move_stackless_bytecode_2` happens to lay blocks out in CFG-flow order. Nothing asserts that invariant. If upstream ever does something different, sort-order could disagree with CFG-flow and the `Jump` drop would silently produce wrong fall-through. Instead, we reverse post-order traverse the `loop_nodes` to ensure the correct order: the IfElse's post-dom is genuinely the next sibling regardless of how blocks are numbered.

## Test plan 

All 52 snapshot tests pass. Zero snapshot drift — confirms RPO and sort-order agree on the current corpus, which is the predicted outcome. The change buys robustness if upstream ever renumbers blocks.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-decompiler/src/structuring/mod.rs
+++ b/external-crates/move/crates/move-decompiler/src/structuring/mod.rs
@@ -116,13 +116,21 @@ fn structure_loop(
         println!("  successor nodes: {succ_nodes:#?}");
     }
 
-    let mut loop_nodes_iter = loop_nodes.clone().into_iter().collect::<Vec<_>>();
-    loop_nodes_iter.sort();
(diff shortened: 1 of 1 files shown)


tzakian (approved):
a little bit of RPO never hurt anything 🎉

PR #26646, opened 2026-05-14
[Block Streaming] support DKG generation as an Observer

akichidis:
## Description 

This is the first in a series of PRs enabling block-streaming full nodes to process randomness round signatures. This will allow them  to execute transactions from local commits without waiting for randomness state update transactions to arrive via checkpoints - which would negate the latency benefit of block streaming.                                                                                   
                                                                                                                                         
This PR updates the fastcrypto library to introduce the Observer construct for the DKG protocol. An Observer (i.e. a block-streaming full node) can follow the DKG process and derive the shared public key without having any voting rights or private key shares. This is possible because all DKG messages and confirmations are propagated through consensus and are therefore durably available to anyone following the DAG. With the public key in hand, the observer can verify randomness round signatures received from its block-streaming  peer - the signatures themselves are the only piece that travels over a side channel rather than through consensus.  

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -617,10 +617,10 @@ move-abstract-stack = { path = "external-crates/move/crates/move-abstract-stack"
 move-analyzer = { path = "external-crates/move/crates/move-analyzer" }
 move-trace-format = { path = "external-crates/move/crates/move-trace-format" }
 
-fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "2ae161a578262a5bbf5c3872054c7f957d744f78" }
-fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", rev = "2ae161a578262a5bbf5c3872054c7f957d744f78" }
-fastcrypto-zkp = { git = "https://github.com/MystenLabs/fastcrypto", rev = "2ae161a578262a5bbf5c3872054c7f957d744f78", package = "fastcrypto-zkp" }
-fastcrypto-vdf = { git = "https://github.com/MystenLabs/fastcrypto", rev = "2ae161a578262a5bbf5c3872054c7f957d744f78", features = [
+fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "239eed70b763f89fdb1e6c6866fe389690945c45" }
+fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", rev = "239eed70b763f89fdb1e6c6866fe389690945c45" }
+fastcrypto-zkp = { git = "https://github.com/MystenLabs/fastcrypto", rev = "239eed70b763f89fdb1e6c6866fe389690945c45", package = "fastcrypto-zkp" }
+fastcrypto-vdf = { git = "https://github.com/MystenLabs/fastcrypto", rev = "239eed70b763f89fdb1e6c6866fe389690945c45", features = [
   "experimental",
 ] }
 passkey-types = { version = "0.4.0" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -617,10 +617,10 @@ move-abstract-stack = { path = "external-crates/move/crates/move-abstract-stack"
 move-analyzer = { path = "external-crates/move/crates/move-analyzer" }
 move-trace-format = { path = "external-crates/move/crates/move-trace-format" }
 
-fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "0eea35ee442a798fb1f1e7c2f39dc021bd21dbc3" }
-fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", rev = "0eea35ee442a798fb1f1e7c2f39dc021bd21dbc3" }
-fastcrypto-zkp = { git = "https://github.com/MystenLabs/fastcrypto", rev = "0eea35ee442a798fb1f1e7c2f39dc021bd21dbc3", package = "fastcrypto-zkp" }
-fastcrypto-vdf = { git = "https://github.com/MystenLabs/fastcrypto", rev = "0eea35ee442a798fb1f1e7c2f39dc021bd21dbc3", features = [
+fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "8e9ae9ba438a3219ad849d3f163c61868e1ce3c9" }
+fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", rev = "8e9ae9ba438a3219ad849d3f163c61868e1ce3c9" }
+fastcrypto-zkp = { git = "https://github.com/MystenLabs/fastcrypto", rev = "8e9ae9ba438a3219ad849d3f163c61868e1ce3c9", package = "fastcrypto-zkp" }
+fastcrypto-vdf = { git = "https://github.com/MystenLabs/fastcrypto", rev = "8e9ae9ba438a3219ad849d3f163c61868e1ce3c9", features = [
   "experimental",
 ] }
 passkey-types = { version = "0.4.0" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -617,10 +617,10 @@ move-abstract-stack = { path = "external-crates/move/crates/move-abstract-stack"
 move-analyzer = { path = "external-crates/move/crates/move-analyzer" }
 move-trace-format = { path = "external-crates/move/crates/move-trace-format" }
 
-fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "8e9ae9ba438a3219ad849d3f163c61868e1ce3c9" }
-fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", rev = "8e9ae9ba438a3219ad849d3f163c61868e1ce3c9" }
-fastcrypto-zkp = { git = "https://github.com/MystenLabs/fastcrypto", rev = "8e9ae9ba438a3219ad849d3f163c61868e1ce3c9", package = "fastcrypto-zkp" }
-fastcrypto-vdf = { git = "https://github.com/MystenLabs/fastcrypto", rev = "8e9ae9ba438a3219ad849d3f163c61868e1ce3c9", features = [
+fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "a2a33f214e67df33e97275b4bc72da84f243f4a4" }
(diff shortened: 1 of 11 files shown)


crates/sui-core/src/epoch/randomness.rs
```diff
@@ -619,20 +788,36 @@ impl RandomnessManager {
...
-                match VersionedProcessedMessage::process(party, msg) {
-                    Ok(processed) => Some(processed),
-                    Err(err) => {
-                        debug!("random beacon: error while processing DKG Message: {err:?}");
-                        None
+        if let Some(observer) = &self.observer {
+            let observer = observer.clone();
+            self.enqueued_messages.insert(
+                msg.sender(),
+                tokio::task::spawn(async move {
```
akichidis: There is no need for async processing for the observer messages as they are lightweight, but this allows me to seamlessly integrate them to the rest of the processing and leverage the `enqueded_messages` without having to introduce separate structs.

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1340,6 +1340,8 @@ impl AuthorityPerEpochStore {
         mut randomness_manager: RandomnessManager,
     ) -> SuiResult<()> {
         let reporter = randomness_manager.reporter();
+        debug_assert!(reporter.is_some() || self.node_role.is_fullnode());
```
aschran: debug assertions should generally be `debug_fatal` unless there is a specific reason not to use it (since that will trigger an alert on mainnet/testnet whereas debug_assert won't)

crates/sui-core/src/epoch/randomness.rs
```diff
@@ -71,7 +71,15 @@ impl VersionedProcessedMessage {
         }
     }
 
-    pub fn process(
+    pub fn unwrap_v1_ref(&self) -> &dkg_v1::ProcessedMessage<PkG, EncG> {
```
aschran: I think conventionally we would call this `as_v1` - `unwrap` suggests a by-value self arg and not by-ref

crates/sui-core/src/epoch/randomness.rs
```diff
@@ -71,7 +71,15 @@ impl VersionedProcessedMessage {
...
+        if let VersionedProcessedMessage::V1(msg) = self {
+            msg
+        } else {
+            panic!("BUG: expected message version is 1")
```
aschran: for an "as_*" function, would be preferable if it's possible to avoid a panic here and instead return an `Option`, so that callers can try not to crash the process if it's avoidable (for example by calling `debug_fatal!` and then maybe somehow retrying, or disabling the randomness subsystem—whatever is sensible in the context)

arguably I should have done something similar above too, but I guess it hasn't broken so far lol

crates/sui-core/src/epoch/randomness.rs
```diff
@@ -96,6 +117,28 @@ impl VersionedProcessedMessage {
             VersionedUsedProcessedMessages::V1(msgs),
         ))
     }
+
+    pub fn merge_observer(
```
aschran: can these be documented? there is an explanation of the overall flow above the `RandomnessManager` struct definition, maybe that could be extended (or wherever else you think makes sense)

crates/sui-core/src/epoch/randomness.rs
```diff
@@ -105,10 +148,18 @@ pub enum VersionedUsedProcessedMessages {
 }
 
 impl VersionedUsedProcessedMessages {
-    fn complete_dkg<'a, Iter: Iterator<Item = &'a VersionedDkgConfirmation>>(
+    pub fn unwrap_v1_ref(&self) -> &dkg_v1::UsedProcessedMessages<PkG, EncG> {
```
aschran: same comments here

crates/sui-core/src/epoch/randomness.rs
```diff
@@ -154,7 +220,8 @@ pub struct RandomnessManager {
 
     // State for DKG.
     dkg_start_time: OnceCell<Instant>,
-    party: Arc<dkg_v1::Party<PkG, EncG>>,
+    party: Option<Arc<dkg_v1::Party<PkG, EncG>>>,
```
aschran: is it ever correct to have both of these or zero of these? if not can they be put in an enum so the type system enforces that we have exactly one of the two? could call it something like `DkgState`

the code from (approx) 327-369 could also then be encapsualted in a constructor of the enum type

and functions like `merge_messages` and `complete_dkg` could become methods on it as well, so that you don't need such lengthy indented if-else blocks for handling the party-vs-observer case

aschran (changes requested):

crates/sui-core/src/authority/authority_per_epoch_store.rs
akichidis: Good point, let me change that

crates/sui-core/src/epoch/randomness.rs
akichidis: Yeah moving to an `as_` will pretty much dictate using an `Option`. That's one of the reasons why I did go with `unwrap` which semantically will know that it _has_ to succeed or panic. But let me go as you suggested - the caller can handle the behaviour.

crates/sui-core/src/epoch/randomness.rs
akichidis: Yep adding documentation here makes sense

crates/sui-core/src/epoch/randomness.rs
akichidis: Yeah I've considered the enum approach when started iterating for all the reasons you outlined, but I've avoided doing it as it would trigger more significant refactoring. That being said though let me have a view on it.

crates/sui-core/src/epoch/randomness.rs
```diff
@@ -148,47 +87,197 @@ pub enum VersionedUsedProcessedMessages {
...
+                    randomness_private_key,
+                ),
+                nodes,
+                t,
+                random_oracle,
+                &mut rand::thread_rng(),
+            ) {
+                Ok(party) => party,
+                Err(err) => {
+                    error!("random beacon: error while initializing Party: {err:?}");
```
aschran: while we're in the neighborhood maybe upgrade this to `debug_fatal!`?

aschran (approved):

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1342,6 +1340,10 @@ impl AuthorityPerEpochStore {
         mut randomness_manager: RandomnessManager,
     ) -> SuiResult<()> {
         let reporter = randomness_manager.reporter();
+        if reporter.is_none() && self.node_role.is_validator() {
```
mwtian: nit: maybe merge this condition with the `if let Some(reporter) = reporter` later.

crates/sui-types/src/node_role.rs
```diff
@@ -79,6 +79,13 @@ impl NodeRole {
     pub fn should_run_rpc_servers(&self) -> bool {
         matches!(self, Self::FullNode(_))
     }
+
+    pub fn is_withdraw_scheduler_enabled(&self) -> bool {
```
mwtian: I think the name and impl of run_consensus() and should_process_consensus_commits() are reversed. Caller of this function should just call should_process_consensus_commits(), which should have the same impl as here?

Cargo.toml
```diff
@@ -617,10 +617,10 @@ move-abstract-stack = { path = "external-crates/move/crates/move-abstract-stack"
 move-analyzer = { path = "external-crates/move/crates/move-analyzer" }
 move-trace-format = { path = "external-crates/move/crates/move-trace-format" }
 
-fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "0eea35ee442a798fb1f1e7c2f39dc021bd21dbc3" }
-fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto", rev = "0eea35ee442a798fb1f1e7c2f39dc021bd21dbc3" }
-fastcrypto-zkp = { git = "https://github.com/MystenLabs/fastcrypto", rev = "0eea35ee442a798fb1f1e7c2f39dc021bd21dbc3", package = "fastcrypto-zkp" }
-fastcrypto-vdf = { git = "https://github.com/MystenLabs/fastcrypto", rev = "0eea35ee442a798fb1f1e7c2f39dc021bd21dbc3", features = [
+fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "2ae161a578262a5bbf5c3872054c7f957d744f78" }
```
mwtian: We may as well take the latest fastcrypto, which is `afa5a34372c95958d79c9dfadfbb5a421878bef5`?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -3251,18 +3252,18 @@ impl MysticetiConsensusHandler {
...
-            process_consensus_commits, "Starting consensus replay"
+            "Starting consensus replay"
         );
         let mut tasks = JoinSet::new();
         tasks.spawn(monitored_future!(async move {
             // TODO: pause when execution is overloaded, so consensus can detect the backpressure.
             while let Some(consensus_commit) = commit_receiver.recv().await {
                 let commit_index = consensus_commit.commit_ref.index;
-                if !process_consensus_commits {
+                if !node_role.should_process_consensus_commits() {
```
mwtian: We are intentionally enabling commit processing here, right?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -3251,18 +3252,18 @@ impl MysticetiConsensusHandler {
...
                     debug!(
                         commit_index,
                         "Observer skipping consensus commit processing"
```
mwtian, line 3269: Should this be a panic with a different message now, since consensus handler should not even be started for full node?

crates/sui-types/src/node_role.rs
akichidis: I am actually going to drop the `is_withdraw_scheduler_enabled` and rename the `should_process_consensus_commits` to `process_consensus_commits` . I'll keep the flow disabled until we do have the randomness round signature propagation.

mwtian: Chatted offline. This is a feature flag for now which I didn't realize before.

mwtian (approved):

PR #26648, opened 2026-05-14
Step-by-step gas-charging pipeline (gas_model v15, protocol v127)

dariorussi:
Replaces the monolithic `charge_gas` with a step-by-step gas-charging pipeline (gas_model v15, protocol v127).

> **On the diff size:** the changed-line count looks large, but the bulk of it is the third commit (`Regenerate snapshots for protocol v127`) — auto-generated test fixtures, not hand-written code. The actual logic change is much smaller and concentrated in the first commit; reviewing the commits individually makes this clear.

> **Follow-up cleanups in this PR (one commit each, on top of the base):** the IFFW short-circuit and the address-balance gas-smash hotfix have been moved into `mod legacy` (commit 4); four further commits remove protocol-flag checks that are dead in the new pipeline — `enable_gasless()`, `gasless_verify_remaining_balance()`, `simple_conservation_checks` plumbing, and the dead `charge_upgrades` arm. All five cleanups touch only `sui-execution/latest` and are runtime-equivalent under v127+. We may squash them into a smaller set of commits before final merge.
>
> Three additional cleanups that originally sat on top of this stack — collapsing `Option<u64>` → `u64` for the abort-tx cap, hardcoding `KeepHalfDigits` and dropping the `GasRoundingMode` enum + `ComputationBucket` helpers, and dropping the non-multiplier arm of `SuiCostTable::new` — were dropped from this PR because they touched the shared `crates/sui-types/src/gas_model/gas_v2.rs`, which is reached from every execution layer (v0..v3 and latest) for old-protocol replay / transactional tests pinned to old versions. They will return in a follow-up PR that introduces `crates/sui-types/src/gas_model/gas_v3.rs` — a v15+-only, clean SuiGasStatus alongside the existing gas_v2 — so the simplifications apply only where they're safe.

## Description

The old `charge_gas` interleaved every concern in one function and had grown messy and hard to follow. This rewrites it as a sequence of explicit, ordered steps, each owning one concern with a clear ok/err outcome so the flow reads top-to-bottom and every stage is testable in isolation:

1. `charge_input_objects` – storage for the input objects
2. `execute_ptb` – run the PTB, bucketize computation
3. `charge_storage` – storage cost + rebate vs the remaining budget
4. `check_effects_limits` – always run, even on the error path
5. `reset_writes` (on err) – drop writes, re-smash gas, recharge storage
6. `charge` – apply net gas to the gas coin / address balance
7. `run_conservation_checks`

The pre-v15 behavior is preserved bit-for-bit in `mod legacy` (gas_model 13/14). It exists only to replay old transactions and is meant to be deleted at the next execution-version cut.

**Why a protocol bump:** gas charging is consensus-critical. The new pipeline changes several effects-visible outcomes — computation derived from the Move meter, the abort price cap, OOG/storage-reset ordering, and reading the gas-charge location at apply time. Gating on a protocol version guarantees old transactions replay identically under their original version. Main shipped v126 and devnet activated it on 2026-06-03, so this lands on **v127**.

## Test plan

- Measured coverage of the new code with `cargo-llvm-cov` over the gas test suites (gas_v2 ~86%, gas_charger ~82% line coverage) — already strong; added one deterministic regression test for the full-budget OOG charge (`test_oog_full_budget_charge_exact_when_budget_indivisible_by_price`), which pins the `force_computation_cost_to_budget` precision fix.
- `gas_tests` (22), transactional `address_balances` (30) + gasless/sponsor suites, and the `p2p_fuzz` proptest all pass under v127.
- **Parity:** replaying mainnet transactions from epoch 1100 onward shows no visible change in effects so far (no fork), indicating the gas results match the legacy pipeline — while still being protocol-version protected.

---

## Release notes


dariorussi:
back to draft given this is changing significantly in light of latest issues and the consistent merge problems

PR #26651, opened 2026-05-14
[move-ide] Added support for smart features in non-inlined macro functions

awelc:
## Description 

This PR adds support for "smart" IDE features for macro functions that have not yet been inlined (e.g., are under development). Prior to this PR, only some "smart" features were available for macros (e.g., on-hover), and only when they already had at least one inlined instance. The reason for this was that macro functions themselves were not available at typing prior to this PR. This was not correct in the first place as for functions inlined more than once information collected from the inlining site that was processed last would "win" and be displayed for macro body (e.g., on-hover for a specific type concretized during inlining).

This PR preserves macro function bodies throughout typing to make them available for analysis to move-analyzer. They are speculatively and optimistically typed (in a sandbox) which is necessary for "smart" IDE features to work, but they are not otherwise processed during typing and dropped during typing-to-HLIR translation.

With this new feature, we also had to finesse how inlined macro function bodies are handled so that information collected for these does not override information collected for macro function body.

All that being said, "smart" features are not going to work as well for macro functions as they do for "regular" functions as, by definition, macro function bodies may not actually type fully correctly by themselves (outside of inlining context). This is an inherent limitation of how macros work in Move rather than limitation of this implementation.

## Test plan 

New tests have been added. All tests must pass

### DIFF
--- a/external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
+++ b/external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
@@ -406,6 +406,18 @@ impl<'env> DiagnosticReporter<'env> {
         }
     }
 
+    /// Creates a reporter for IDE macro-body typing that shares IDE information with the
+    /// original reporter but sends diagnostics to a fresh sink that will be discarded.
+    pub fn clone_for_ide_macro_body_typing(&self) -> Self {
+        Self {
+            flags: self.flags,
+            known_filter_names: self.known_filter_names,
+            diags: Arc::new(RwLock::new(Diagnostics::new())),
+            ide_information: self.ide_information.clone(),
+            filter_stack: self.filter_stack.clone(),
+        }
+    }
+
     pub fn push_warning_filter_scope(&mut self, scope: FilterScope) {
         self.filter_stack.push(scope)
     }
--- a/external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
+++ b/external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
@@ -406,18 +406,6 @@ impl<'env> DiagnosticReporter<'env> {
         }
     }
 
-    /// Creates a reporter for speculative IDE macro-body typing that shares IDE information with
-    /// the original reporter but sends diagnostics to a fresh sink that will be discarded.
-    pub fn clone_for_ide_macro_body_typing(&self) -> Self {
-        Self {
-            flags: self.flags,
-            known_filter_names: self.known_filter_names,
-            diags: Arc::new(RwLock::new(Diagnostics::new())),
-            ide_information: self.ide_information.clone(),
-            filter_stack: self.filter_stack.clone(),
-        }
-    }
-
     pub fn push_warning_filter_scope(&mut self, scope: FilterScope) {
         self.filter_stack.push(scope)
     }
--- a/external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
+++ b/external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
@@ -406,8 +406,8 @@ impl<'env> DiagnosticReporter<'env> {
         }
     }
 
-    /// Creates a reporter for IDE macro-body typing that shares IDE information with the
-    /// original reporter but sends diagnostics to a fresh sink that will be discarded.
+    /// Creates a reporter for speculative IDE macro-body typing that shares IDE information with
+    /// the original reporter but sends diagnostics to a fresh sink that will be discarded.
     pub fn clone_for_ide_macro_body_typing(&self) -> Self {
         Self {
             flags: self.flags,
--- a/external-crates/move/crates/move-compiler/src/shared/ide.rs
+++ b/external-crates/move/crates/move-compiler/src/shared/ide.rs
@@ -32,8 +32,9 @@ use std::{
 #[derive(Debug, Clone, Default)]
 pub struct IDEInfo {
     pub(crate) annotations: Vec<(Loc, IDEAnnotation)>,
-    /// Typed macro function bodies keyed by function definition location.
-    macro_function_bodies: BTreeMap<Loc, T::Sequence>,
+    /// Typed macro function bodies keyed by defining module and function name
+    /// whose location is the location of the function definition.
+    macro_function_bodies: BTreeMap<(E::ModuleIdent, P::FunctionName), T::Sequence>,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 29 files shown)


external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
```diff
@@ -406,6 +406,18 @@ impl<'env> DiagnosticReporter<'env> {
         }
     }
 
+    /// Creates a reporter for speculative IDE macro-body typing that shares IDE information with
+    /// the original reporter but sends diagnostics to a fresh sink that will be discarded.
+    pub fn clone_for_ide_macro_body_typing(&self) -> Self {
```
cgswords: This should live on `CompilationEnv`, not here, next to `dummy_diagnostic_reporter`. I would call it `ide_diagnostic_reporter`.

external-crates/move/crates/move-compiler/src/shared/ide.rs
```diff
@@ -32,6 +32,8 @@ use std::{
 #[derive(Debug, Clone, Default)]
 pub struct IDEInfo {
     pub(crate) annotations: Vec<(Loc, IDEAnnotation)>,
+    /// Typed macro function bodies keyed by function definition location.
+    macro_function_bodies: BTreeMap<Loc, T::Sequence>,
```
cgswords: Would this be better by identifier or some other key? Just asking...

external-crates/move/crates/move-compiler/src/typing/core.rs
```diff
@@ -756,7 +760,14 @@ impl<'env, 'outer> Context<'env, 'outer> {
...
 
     pub fn check_feature(&self, package: Option<Symbol>, feature: FeatureGate, loc: Loc) -> bool {
-        self.outer.check_feature(package, feature, loc)
+        self.env()
+            .check_feature(&self.reporter, package, feature, loc)
+    }
+
+    /// Returns true for regular compiler errors, or while typing an IDE macro body,
+    /// where diagnostics emitted during typing are not recorded in `CompilationEnv`.
+    pub fn has_errors_or_ide_typing_macro_body(&self) -> bool {
```
cgswords: I appreciate this is "abstraction" but inlining this in `macro_expand.rs` seems preferable to me, even just as a macro.

external-crates/move/crates/move-compiler/src/typing/translate.rs
```diff
@@ -522,6 +526,38 @@ fn function_signature(context: &mut Context, macro_: Option<Loc>, sig: &N::Funct
...
+/// (already in scope from function_signature) to type the body, generating IDE annotations
+/// (DotAutocompleteInfo, etc.). Diagnostics from this best-effort pass are discarded.
+/// Speculatively typed macro function body is added to the current context's IDEInfo.
+fn ide_macro_body(context: &mut Context, function_loc: Loc, sp!(_, nb_): N::FunctionBody) {
+    let N::FunctionBody_::Defined(es) = nb_ else {
+        return;
+    };
+
+    let reporter = context.reporter.clone_for_ide_macro_body_typing();
+    let old_reporter = std::mem::replace(&mut context.reporter, reporter);
```
cgswords, line 535: Can we not just call the normal function once we swap the reporters?

external-crates/move/crates/move-compiler/src/diagnostics/mod.rs
awelc: Good point. Done!

external-crates/move/crates/move-compiler/src/shared/ide.rs
awelc: It was just convenient as location is also used on the analyzer side to merge new and cached information. I can still changes to something like function name and module identifier pair and extract location from one of these if need be

external-crates/move/crates/move-compiler/src/typing/core.rs
awelc: It's used in two separate files and I wanted something self-documenting but indeed making this part of `Context`'s implementation is perhaps too much. I turned it into a macro in core.rs, but can also simply inline the condition in both "client" files if that's preferred

external-crates/move/crates/move-compiler/src/typing/translate.rs
awelc: Good call. Done!

external-crates/move/crates/move-compiler/src/shared/ide.rs
cgswords: I think using the qualified path is a much better choice, yes. Location stability seems unreliable. (I know that is not exactly true, but still.)

external-crates/move/crates/move-compiler/src/typing/core.rs
```diff
@@ -756,7 +769,8 @@ impl<'env, 'outer> Context<'env, 'outer> {
     }
 
     pub fn check_feature(&self, package: Option<Symbol>, feature: FeatureGate, loc: Loc) -> bool {
```
cgswords, line 762: This is an interesting change worth flagging. Add a doc comment:

```suggestion
    /// Checked the current feature using the currently-registered diagnostic reporter
    pub fn check_feature(&self, package: Option<Symbol>, feature: FeatureGate, loc: Loc) -> bool {
```

external-crates/move/crates/move-compiler/src/typing/core.rs
```diff
@@ -176,6 +178,16 @@ pub struct Context<'env, 'outer> {
     pub ide_info: IDEInfo,
 }
 
+/// True for regular compiler errors, or while speculatively typing an IDE macro body with
+/// diagnostics thrown away.
+macro_rules! has_errors_or_ide_typing_macro_body {
```
cgswords: Okay, I have gone full-circle on this. I think we should just write a function on the context on context. Something like:

```
fn assert_errors(&self, msg: String) -> bool { 
  if !(self.env().has_errors() || self.ide_typing_macro_body) {
    // report ICE
  }
}
```
There is even some `assert_ice!` or something you can probably use.

Then we get:

```
               Err(None) => {
                context.assert_errors("Macro recoloring failed with error");
                return None;
            }
```

external-crates/move/crates/move-compiler/src/typing/core.rs
awelc: Good idea. Done!

external-crates/move/crates/move-compiler/src/typing/core.rs
awelc: Done! I went for using `assert!` inside the function as this was the original behavior (which this way will stay the same on non-IDE path) plus using `assert_ice!` would require passing location at the call site which would arguably make this change overly churn-y

external-crates/move/crates/move-compiler/src/shared/ide.rs
awelc: Switched to using function name qualified with the module identifier

external-crates/move/crates/move-compiler/src/typing/macro_expand.rs
```diff
@@ -88,7 +88,9 @@ pub(crate) fn call(
         ) {
             Ok(res) => res,
             Err(None) => {
-                assert!(context.env().has_errors());
+                context.assert_errors(
```
cgswords: Tiny nit, but maybe `assert_has_errors` just to be a touch clearer.

cgswords (approved):
LGTM except for one small nit.

external-crates/move/crates/move-compiler/src/typing/macro_expand.rs
awelc: Done!

PR #26652, opened 2026-05-14
[ci] Do not run CI on edits to markdown files only

stefan-mysten:
## Description 

I had a few PRs that would fix README.md files in the `/crates` folder. Unfortunately, that triggers all the tests which I think we shouldn't need to do, except for the `sui-framework` crate and for `external-crates/move/documentation` and `external-crates/move/crates/move-stdlib`.

## Test plan 

Existing CI
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/.github/actions/diffs/action.yml
+++ b/.github/actions/diffs/action.yml
@@ -9,10 +9,10 @@ outputs:
     value: "${{ steps.diff.outputs.isOldDoc }}"    
   isRust:
     description: True when changes happened to the Rust code
-    value: "${{ steps.diff.outputs.isRust }}"
+    value: "${{ steps.diff.outputs.isRust == 'true' && steps.markdown_only.outputs.isNonFrameworkMarkdownOnly != 'true' }}"
   isMove:
     description: True when changes happened to the Move code
-    value: "${{ steps.diff.outputs.isMove }}"
+    value: "${{ steps.diff.outputs.isMove == 'true' && steps.markdown_only.outputs.isNonFrameworkMarkdownOnly != 'true' }}"
   isSolidity:
     description: True when changes happened to the Solidity code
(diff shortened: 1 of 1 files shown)


.github/actions/diffs/action.yml
```diff
@@ -9,10 +9,10 @@ outputs:
     value: "${{ steps.diff.outputs.isOldDoc }}"    
   isRust:
     description: True when changes happened to the Rust code
-    value: "${{ steps.diff.outputs.isRust }}"
+    value: "${{ steps.diff.outputs.isRust == 'true' && steps.markdown_only.outputs.isNonFrameworkMarkdownOnly != 'true' }}"
```
ebmifa, line 12: [P1] This treats a matching Markdown file as if the whole PR is Markdown-only. `dorny/paths-filter` sets a filter output to true when any changed file matches; `predicate-quantifier: every` only changes how an individual file is matched against the include/exclude rules. A mixed PR like `crates/foo/src/lib.rs` + `crates/foo/README.md` can therefore set `isNonFrameworkMarkdownOnly=true` and suppress Rust/Move CI. For this PR, I’d keep the fix surgical, but compute an explicit all-changed-files check and only mask `isRust`/`isMove` when every changed file is a safe README/Markdown path.

ebmifa (commented):
One inline note from the earlier review; the issue still applies on the current diff.

.github/actions/diffs/action.yml
stefan-mysten: Fixed it, I think it should be fine now.

ebmifa (approved):

PR #26661, opened 2026-05-15
[typed-store] Fix schedule_delete_all leaving the largest key behind

wstran:
## Summary
`schedule_delete_all` cleared the table with `schedule_delete_range(first, last)`, which is documented `[from, to)` exclusive-upper (see its docstring and `test_delete_range` at `tests.rs:410`). The largest key therefore always survived; on a single-key map the function deleted nothing.

The sole caller — `AuthorityStore::clear_object_per_epoch_marker_table` — runs at every epoch boundary on every node. Per-epoch markers are keyed with `EpochId` as the leading component, so the residual row is inert at read time (subsequent-epoch reads filter by exact `EpochId` and never observe it), but the function still violates its own contract as a general-purpose `Map` trait method.

## Fix
Clear the table with a single range tombstone whose exclusive upper bound is the lexicographic successor of the encoded largest key: `be_fix_int_ser(&last_key)` with a trailing `0x00` byte appended. Appending `0x00` is the unique lexicographic successor of any byte sequence, so the half-open range `[first_buf, last_buf ++ 0x00)` is exactly the closed range `[first_key, last_key]`. (This mirrors the existing `key_buf.push(0)` pattern in `iterator_bounds_with_range` at `crates/typed-store/src/util.rs`.)

Keeping the whole clear as one range tombstone — rather than a range delete plus a point delete on the largest key — preserves consistent visibility: every entry follows the same path governed by `ignore_range_deletions`, instead of the largest key vanishing immediately via a point delete while the rest of the table stays readable until compaction.

A new `DBBatch::schedule_delete_range_raw` takes pre-encoded byte ranges (`from` inclusive, `to` exclusive) for callers that need bounds not corresponding to any serializable key; `schedule_delete_range` now delegates to it.

## Reproduction (actual test output)
Three regression tests are added. `test_schedule_delete_all_default_read_options` uses the default `ReadWriteOptions` (`ignore_range_deletions = true`): it asserts every key is still visible immediately after `schedule_delete_all` (no leaked point delete), then asserts the table is empty after `flush() + compact_range()`. The two `*_includes_last_key` / `*_single_entry` tests use `ignore_range_deletions = false` to assert immediate emptiness.

**Before the fix** — the default-options test against the previous (range + point delete) implementation reproduces the visibility skew exactly:

```
thread 'rocks::tests::test_schedule_delete_all_default_read_options' panicked at crates/typed-store/src/rocks/tests.rs:481:9:
key 100 disappeared before compaction under default ReadWriteOptions
```

And against the original unpatched `schedule_delete_all`, the `ignore_range_deletions = false` tests fail because the largest key (and, for a single-key map, the whole map) survives:

```
assertion `left == right` failed: schedule_delete_all must remove every entry, including the largest key
  left: []
 right: [(100, "100")]
```

**After the fix** — full typed-store suite (`cargo nextest run -p typed-store`):

```
PASS  rocks::tests::test_schedule_delete_all_single_entry
PASS  rocks::tests::test_schedule_delete_all_includes_last_key
PASS  rocks::tests::test_schedule_delete_all_default_read_options
Summary  38 tests run: 38 passed, 0 skipped
```

`cargo xclippy` on the crate finishes clean.

## Test plan
- [x] 3 regression tests (two with `ignore_range_deletions = false`, one with the default options) fail before the fix, pass with it (output above).
- [x] Full typed-store nextest suite: 38/38.
- [x] `cargo xclippy` clean; `cargo fmt --all -- --check` clean.

### DIFF
--- a/crates/typed-store/src/rocks/mod.rs
+++ b/crates/typed-store/src/rocks/mod.rs
@@ -1783,7 +1783,10 @@ where
             .map(|(k, _v)| k);
         if let Some((first_key, last_key)) = first_key.zip(last_key) {
             let mut batch = self.batch();
+            // `schedule_delete_range` excludes its upper bound, so the largest key
+            // would survive; delete it explicitly to actually clear the map.
             batch.schedule_delete_range(self, &first_key, &last_key)?;
+            batch.delete_batch(self, [last_key])?;
             batch.write()?;
         }
         Ok(())
--- a/crates/typed-store/src/rocks/mod.rs
+++ b/crates/typed-store/src/rocks/mod.rs
@@ -1520,18 +1520,30 @@ impl DBBatch {
         from: &K,
         to: &K,
     ) -> Result<(), TypedStoreError> {
+        self.schedule_delete_range_raw(db, be_fix_int_ser(from), be_fix_int_ser(to))
+    }
+
+    /// Like [`Self::schedule_delete_range`], but takes pre-encoded byte ranges
+    /// directly. Useful when the desired bounds do not correspond to any
+    /// serializable key — for example, when the caller needs an exclusive
+    /// upper bound that is the lexicographic successor of a real key in order
+    /// to express a closed range as a half-open one (see
+    /// `schedule_delete_all`).
+    pub fn schedule_delete_range_raw<K, V>(
+        &mut self,
+        db: &DBMap<K, V>,
+        from: impl AsRef<[u8]>,
+        to: impl AsRef<[u8]>,
+    ) -> Result<(), TypedStoreError> {
         if !Arc::ptr_eq(&db.db, &self.database) {
             return Err(TypedStoreError::CrossDBBatch);
         }
 
-        let from_buf = be_fix_int_ser(from);
-        let to_buf = be_fix_int_ser(to);
-
         if let StorageWriteBatch::Rocks(b) = &mut self.batch {
             b.delete_range_cf(
                 &rocks_cf_from_db(&self.database, db.cf_name())?,
-                from_buf,
-                to_buf,
+                from.as_ref(),
+                to.as_ref(),
             );
         }
         Ok(())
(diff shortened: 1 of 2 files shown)


crates/typed-store/src/rocks/mod.rs
```diff
@@ -1783,7 +1783,10 @@ where
             .map(|(k, _v)| k);
         if let Some((first_key, last_key)) = first_key.zip(last_key) {
             let mut batch = self.batch();
+            // `schedule_delete_range` excludes its upper bound, so the largest key
+            // would survive; delete it explicitly to actually clear the map.
             batch.schedule_delete_range(self, &first_key, &last_key)?;
+            batch.delete_batch(self, [last_key])?;
```
sadhansood: Thanks for catching this! However, can we avoid mixing a range tombstone with a point delete here? `schedule_delete_all` should preserve the same delete visibility but here delete batch will immediately delete the key whereas delete range's behavior depends on `ignore_range_deletions`. Instead of `delete_batch(last_key)`, consider adding a raw range delete helper and passing `last_key`'s lexicographic successor and then replacing the current `batch.schedule_delete_range(self, &first_key, &last_key)?;` call with something like `schedule_delete_range_raw(self, first_key_raw_bytes, last_key_raw_bytes + \0x00)`

That would keep this as a single range tombstone and avoids making the max key disappear
immediately while the rest of the table may remain visible under the default `ignore_range_deletions=true`. Also, could you please add a test with default ReadWriteOptions, not only ignore_range_deletions(false), so the intended visibility behavior is explicit.

wstran: Good call — applied in 10eb9b59. The fix is now a single range tombstone with an exclusive upper bound built from `last_key`'s lexicographic successor, exactly as you suggested.

**What changed**

- Added `DBBatch::schedule_delete_range_raw(db, from_bytes, to_bytes)` (mirrors the existing `compact_range_raw`); `schedule_delete_range` now delegates to it.
- `schedule_delete_all` builds `to_buf = be_fix_int_ser(&last_key); to_buf.push(0u8);` and issues a single `schedule_delete_range_raw(self, &from_buf, &to_buf)`. Appending `0x00` is the unique lex-successor of any byte sequence, so `[first_buf, last_buf ++ 0x00)` is exactly the closed range `[first_key, last_key]`. The whole clear stays under one range tombstone, so visibility is governed end-to-end by `ignore_range_deletions`. (Same `key_buf.push(0)` pattern used at `crates/typed-store/src/util.rs:71` for the analogous "exclusive lower bound at max" case.)

**Test under default `ReadWriteOptions`** (i.e. `ignore_range_deletions = true`): added `test_schedule_delete_all_default_read_options`. It asserts every entry is still readable immediately after `schedule_delete_all` (i.e. no point delete leaked), then asserts the table is empty after `flush() + compact_range()`. The pre-compaction assertion was chosen specifically to catch this kind of visibility skew — re-running it against the previous (range + point) implementation reproduces the asymmetry exactly:

```
thread 'rocks::tests::test_schedule_delete_all_default_read_options' panicked at crates/typed-store/src/rocks/tests.rs:481:9:
key 100 disappeared before compaction under default ReadWriteOptions
```

With the new code, all three `schedule_delete_all` tests pass and the typed-store crate is green end-to-end:

```
PASS  rocks::tests::test_schedule_delete_all_single_entry
PASS  rocks::tests::test_schedule_delete_all_includes_last_key
PASS  rocks::tests::test_schedule_delete_all_default_read_options
Summary  38 tests run: 38 passed, 0 skipped
```

andll:
@wstran thanks for the fix. this is mostly looks good right now, can you update PR description to reflect the changes you've made in response to review by @sadhansood (PR still mentions `Fix: after the range delete, explicitly delete_batch the largest key as well`)
In the doc comment to `schedule_delete_range_raw` can you also explicitly say which bounds are inclusive/exclusive

wstran:
Thanks @andll — both addressed in 00d1c0b:

- Rewrote the PR description to describe the v2 approach (single range tombstone whose exclusive upper bound is the lexicographic successor of the encoded largest key) and the new default-`ReadWriteOptions` regression test, replacing the stale v1 wording.
- Clarified the `schedule_delete_range_raw` doc comment to state explicitly that `from` is inclusive and `to` is exclusive, matching the underlying RocksDB range tombstone.

Suite still green (38/38), `cargo xclippy` and `cargo fmt --all -- --check` clean.

PR #26662, opened 2026-05-15
[move][move-decompiler] Capture types in the AST so the alias renamer rewrites them

cgswords:
## Description 

Revises the `ast` form to include more information extracted from the model so that the rewriter can handle it, and updates the rewriter to do so.

## Test plan 

Test improvements.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-decompiler/src/ast.rs
+++ b/external-crates/move/crates/move-decompiler/src/ast.rs
@@ -1,8 +1,12 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use move_binary_format::normalized::{Constant, ModuleId};
+use move_binary_format::{
+    file_format::{AbilitySet, DatatypeTyParameter, Visibility},
+    normalized::{self, Constant, ModuleId},
+};
 
+use indexmap::IndexMap;
 use move_core_types::{account_address::AccountAddress, runtime_value::MoveValue as Value};
 use move_model_2::{model::Model, source_kind::SourceKind};
 use move_stackless_bytecode_2::ast::{DataOp, PrimitiveOp};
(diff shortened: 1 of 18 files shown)


external-crates/move/crates/move-decompiler/src/ast.rs
```diff
@@ -28,6 +32,11 @@ pub struct Package {
 #[derive(Debug, Clone)]
 pub struct Module {
     pub name: Symbol,
+    /// Structs in declaration order. We use `IndexMap` rather than `BTreeMap` because users
+    /// expect source-position ordering for types, and bytecode preserves the original order in
+    /// its index map. Functions still use `BTreeMap` (alphabetical) to match prior behavior.
+    pub structs: IndexMap<Symbol, Struct>,
+    pub enums: IndexMap<Symbol, Enum>,
     pub functions: BTreeMap<Symbol, Function>,
```
tzakian, line 40: Do we keep source position when compiling to bytecode? If so would be nice to do the same for functions as well eventually, but if we don't I don't see any real reason to use `IndexMap` over `BTreeMap` as the ordering will already have been lost.

external-crates/move/crates/move-decompiler/src/ast.rs
```diff
@@ -43,8 +52,106 @@ pub struct Module {
...
+    /// types as `Type::TypeParameter(index)` and rendered as `T{index}`.
+    pub type_parameters: Vec<AbilitySet>,
+    /// Parameter types, in declaration order. Parameter names are not preserved in bytecode;
+    /// the pretty printer generates `l0..l{N-1}` matching `term_reconstruction::local_name`.
+    pub parameters: Vec<Type>,
+    /// Return types. Empty for `(): ()`, single element for one return, multiple for a tuple.
+    pub returns: Vec<Type>,
     pub code: Exp,
-    // TODO add function args?
+}
```
tzakian, line 66: ^ yes 😉

tzakian (approved):
LGTM. I love it!

external-crates/move/crates/move-decompiler/src/ast.rs
cgswords: I do believe this maintains it, yes.

PR #26664, opened 2026-05-15
[move][move-decompiler] Eliminate more gotos

cgswords:
## Description 

Three things, all serving one goal: make `structure_acyclic_region` produce idiomatic structured output with each scope deciding placement from local data only. 

### This gets us down to about ~200 remaining gotos over all of mainnet.

**1. Local owned-children hoist replaces the global filter and `next_in_outer` threading.** After building the IfElse/Switch, any `ichildren` of `start` that weren't absorbed as arms and aren't an enclosing loop's exit are "orphans".: every CFG path to them goes through `start` — and get appended as siblings of the IfElse/Switch. Walking the IfElse's tail positions then drops `Jump`s targeting the hoisted block. `arm_for` simplifies to five cases: 
- back-edge → `DegenerateJumpIf`;
- loop-succ → `LoopBreak` (the new tag drives `insert_breaks` to rewrite as `Break(loop_head)`);
- post-dom (and not `return_`) → `ArmOutsideSubtree` placeholder for the hoist to elide;
- `target ∈ ichildren` → embed;
- otherwise → `ArmOutsideSubtree`.
This gets rid of the `next_in_outer` threading, the `inside_loop` flag, the absorbed-children prediction, the pre-/post-absorption back-edge predicates, and the post-dom reachability filter.

**2. Code-next singly-entered fusion.** `structure_code_node` fuses `next` if it's our exclusive dom-tree child: `next ∈ ichildren(start)` *and* `is_singly_entered(next)` (no other CFG predecessor outside its own dom subtree) *and* not an enclosing-loop exit. When the test fails, emit an explicit `Jump(CodeBranch, next)` so the owned-children hoist or `insert_breaks` can see and rewrite it.

**3. Make `flatten_sequence` recursive ``.** No longer shallow. More expensive, but better results.

Also fixed:
- SCC impossible-edge logic crashed on packages with abort-only branches (`graph.rs`, `add_edge` on a missing index). Fixed by excluding aborts from the SCC rep selection. `cetus_clmm_worker.mv` added as a regression fixture.
- `find_loop_nodes` iterated HashSet contents non-deterministically, leaking into `refine_loop_nodes`'s greedy fixpoint. Sort both before iterating. `structure_loop`'s `succ_node` is now `.min()` instead of `.next()` for the same reason.

## Test plan 

Added a bunch of new tests and updated snapshots showing the improvements we've made.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-decompiler/src/structuring/graph.rs
```diff
@@ -169,8 +164,13 @@ impl Graph {
             }
         }
 
+        // Iterate `loop_nodes` in sorted order — it's a HashSet so iteration order is
+        // otherwise non-deterministic, and that order leaks into `refine_loop_nodes`'s
+        // greedy fixpoint, which can produce different SCC-boundary refinements run-to-run.
+        let mut loop_nodes_sorted: Vec<NodeIndex> = loop_nodes.iter().copied().collect();
```
tzakian, line 170: Would there be any benefit to using an IndexMap or BTreeMap/Set here to keep things sorted by default instead of needing to pull them out and then sort them?

external-crates/move/crates/move-decompiler/src/structuring/mod.rs
```diff
@@ -314,193 +400,227 @@ fn structure_acyclic_region(
...
+            D::Structured::Jump(GotoSource::LoopBreak, target)
+        } else if target == post_dominator && post_dominator != return_ {
+            D::Structured::Jump(GotoSource::ArmOutsideSubtree, target)
         } else if ichildren.contains(&target) {
-            Some(structured_blocks.remove(&target).unwrap())
+            structured_blocks.remove(&target).unwrap()
         } else {
-            Some(D::Structured::Jump(GotoSource::ArmOutsideSubtree, target))
+            D::Structured::Jump(GotoSource::ArmOutsideSubtree, target)
         }
```
tzakian, line 437: This is when I wish Rust had `cond`....

Although maybe a `match` with guards? Don't know how much nicer that would be though.

external-crates/move/crates/move-decompiler/tests/bytecode/output/0x3001c0d95f0498b8e92fe95878b25e1c2e85ff213f3ff5b1ef088390ed185fc1/cetus_clmm_worker.move
```diff
@@ -0,0 +1,2288 @@
...
+    let l39 = l12;
+    let l38 = l11;
+    let l32 = l4;
+    let l23 = l1;
+    let l49 = cetus_clmm_worker::is_stable(freeze(l0), freeze(l23), freeze(l32), l38, l39, l40);
+    let l21 = &l0.position_operator_cap;
+    let l20 = l1;
+    let l19 = l2;
+    let l18 = &l51;
+    let l17 = l5;
```
tzakian, line 241: Thoughts on alias pushing/removal as a refinement here at some point?

tzakian (approved):
Couple tiny questions/suggestions but fine landing as-is as well.

external-crates/move/crates/move-decompiler/tests/bytecode/output/0x3001c0d95f0498b8e92fe95878b25e1c2e85ff213f3ff5b1ef088390ed185fc1/cetus_clmm_worker.move
cgswords: Yes, binding-inlining is on my list.

external-crates/move/crates/move-decompiler/src/structuring/mod.rs
cgswords: I'm not sure that improves readability, sadly; these arms are long enough to get to:
```
_ if target == start => {
    D::Structured::Jump(GotoSource::DegenerateJumpIf, target)
}
_ if Some(target) == loop_succ {
    D::Structured::Jump(GotoSource::LoopBreak, target)
}
```
Not sure that is better. I agree `cond` would go hard.

external-crates/move/crates/move-decompiler/src/structuring/graph.rs
cgswords: I think the cost of building the index map is probably more work than simply sorting it. In addition, calling out the sorting here has its own friction that makes the algorithm more-insistent on what is going on when reading the code IMO.

PR #26665, opened 2026-05-15
sui-protocol-config: render to any Format (JSON, prost::Value, …)

alex-mysten:
## Summary

- Move \`Format\` + \`Meter\` (plus the \`serde_json::Value\` / \`prost_types::Value\` impls) from \`sui-types::object::rpc_visitor\` into \`mysten-common::rpc_format\`. \`sui-types\` re-exports so existing import paths keep working.
- Add a new \`ToFormat\` trait alongside, with blanket impls for \`bool\`/\`u8\`/\`u16\`/\`u32\`/\`u64\`/\`u128\`/\`usize\`/\`String\`/\`Option<T>\`/\`Vec<T>\`/\`[T]\`/2-, 3-, 4-tuples. Integers wider than 32 bits go through \`Format::string\` so JS clients reading the JSON projection don't lose precision.
- Extend the \`ProtocolConfigAccessors\` macro to emit \`ProtocolConfig::render<F: Format, M: Meter>(&self, &mut M) -> Result<BTreeMap<String, Option<F>>, MeterError>\`. The same call site can target \`serde_json::Value\` or \`prost_types::Value\` (or any other \`Format\` impl) by choosing \`F\`.
- Drop the \`#[skip_accessor]\` on \`gasless_allowed_token_types\` so it flows through \`render\` automatically; remove the hand-written \`set_..._for_testing\` setter (the macro generates it now).

Designed to slot in behind a future \`typed_attributes: map<string, google.protobuf.Value>\` proto field — the consumer just instantiates \`render::<prost_types::Value, _>\` and populates the map directly.

## Test plan

- [x] \`cargo nextest run -p mysten-common -p sui-protocol-config -p sui-protocol-config-macros -p sui-types -p sui-rpc-api -p sui-json-rpc-types -p sui-open-rpc\` — 386 passing
- [x] New \`sui-protocol-config\` unit tests covering \`render::<serde_json::Value>\` and \`render::<prost_types::Value>\` against a populated gasless allowlist, plus the \`None\` value for pre-feature protocol versions
- [x] \`cargo xclippy\` clean on affected crates
- [x] \`cargo fmt --all -- --check\` clean
- [x] \`./scripts/update_all_snapshots.sh\` — no drift


crates/sui-types/src/object/rpc_visitor/mod.rs
```diff
@@ -1,16 +1,17 @@
...
-mod meter;
 pub mod proto;
 
-pub use format::Format;
-pub use meter::LocalMeter;
-pub use meter::Meter;
-pub use meter::MeterError;
-pub use meter::Unmetered;
+// `Format` and `Meter` live in `mysten-common` so that lower-level crates (like
+// `sui-protocol-config`) can render values into any `Format`-implementing wire type without
```
aschran: i'm not sure I'd call protocol config a "lower-level" crate than sui-types?

crates/sui-protocol-config-macros/src/lib.rs
```diff
@@ -39,147 +39,116 @@ use syn::{Data, DeriveInput, Fields, Type, parse_macro_input};
 ///     /// Returns a map of all features to values
 ///     pub fn feature_map(&self) -> std::collections::BTreeMap<String, bool>;
 /// ```
+///
+/// Every field (scalar and non-scalar) also participates in a typed
```
aschran: what does it mean for a field to "participate" in a method?

crates/sui-protocol-config-macros/src/lib.rs
```diff
@@ -39,147 +39,116 @@ use syn::{Data, DeriveInput, Fields, Type, parse_macro_input};
...
+/// `render<F: Format, M: Meter>(&self, meter: &mut M) -> Result<BTreeMap<String, Option<F>>, MeterError>`
+/// method, where each value is produced via `mysten_common::rpc_format::ToFormat`. This is the
+/// path RPC code should use to expose protocol config to clients — integers wider than 32 bits
```
aschran: the integer-width comment seems like a confusing implementation detail in the context of this doc comment?

crates/sui-protocol-config-macros/src/lib.rs
```diff
@@ -39,147 +39,116 @@ use syn::{Data, DeriveInput, Fields, Type, parse_macro_input};
...
                 vec![
-                    #(((#field_names_str).to_owned(), self.lookup_attr((#field_names_str).to_owned())),)*
+                    #(((#scalar_field_names).to_owned(), self.lookup_attr((#scalar_field_names).to_owned())),)*
                     ].into_iter().collect()
             }
 
+            /// Render every protocol-config attribute into the chosen `Format`.
+            ///
+            /// Each field's value flows through `ToFormat`, which keeps integers wider than 32
+            /// bits in `Format::string` form so JS clients reading the JSON projection don't
```
aschran: wow claude REALLY wants to make sure you know about this integer precision thing

crates/sui-protocol-config-macros/src/lib.rs
```diff
@@ -227,6 +198,216 @@ pub fn accessors_macro(input: TokenStream) -> TokenStream {
...
+    lookup_arm: proc_macro2::TokenStream,
+    /// `stringify!(field_name)` — used to assemble `attr_map`.
+    field_name_str: proc_macro2::TokenStream,
+    /// Variant identifier used in `ProtocolConfigValue`.
+    variant_ident: syn::Ident,
+    /// Inner `T` of `Option<T>` — pairs with `variant_ident` when declaring the enum variant.
+    inner_type: syn::Type,
+}
+
+fn expand_field(f: &AccessorField) -> ExpandedField {
```
aschran, line 227: Just noting that I surrender to the vibe-codedness of this lol. proc macros were always black magic to begin with, I suppose

do you think it's worth adding a test for a small/stub example protocol config struct? or if you just manually inspected the output and it looks right to you, then I guess that's good enough for me.

crates/sui-protocol-config-macros/src/lib.rs
```diff
@@ -227,6 +198,216 @@ pub fn accessors_macro(input: TokenStream) -> TokenStream {
...
+/// 2. The `render` arm emits `<_ as ToFormat>::to_format::<F, _>(v, meter)`, requiring a
+///    `ToFormat` impl on `T` (`mysten-common::rpc_format::to_format`).
+///
+/// Without an allowlist, any single-segment identifier — including `String` or a user-defined
+/// struct — would route as a scalar and fail to compile inside macro-generated code with a
+/// confusing error pointing at expansion sites. Anything not in the list below is treated as a
+/// non-scalar and only appears in `render`. Adding a new scalar primitive means adding it here
+/// AND adding (or confirming) a corresponding `ToFormat` impl.
+fn inferred_scalar_variant_ident(ty: &syn::Type) -> Option<syn::Ident> {
+    const SCALAR_PRIMITIVES: &[&str] = &["bool", "u8", "u16", "u32", "u64", "u128", "usize"];
```
aschran, line 372: this seems worth documenting more prominently somewhere in the place where such a field (with new scalar primitive) would be added?

aschran: or do we not care, since everyone should migrate to the `render` path anyway and so it's fine if new scalar fields don't appear in the old path anymore?

crates/sui-protocol-config-macros/src/lib.rs
alex-mysten: 100%. i originally skipped the accessor because i wanted to avoid touching this

we do have UTs for render for scalar and nonscalar in protocol-config/lib.rs

crates/sui-protocol-config-macros/src/lib.rs
alex-mysten: new scalars just won't appear in json-rpc unless they are added here, which I think is probably the right thing to do.

compilation will fail if we don't have a corresponding toFormat for the scalar, so we catch that early

crates/sui-types/src/object/rpc_visitor/mod.rs
alex-mysten: this whole comment is unhelpful - dropped it

crates/sui-protocol-config-macros/src/lib.rs
alex-mysten: rephrased, just means it's included in Format

crates/sui-protocol-config-macros/src/lib.rs
alex-mysten: Agreed - dropped

crates/sui-protocol-config-macros/src/lib.rs
alex-mysten: Dropped

crates/sui-protocol-config-macros/src/lib.rs
bmwill: Do we have any ProtocolConfig values which are of type `f32` or `f64`? If so, do we want to take the opportunity to ensure they are supported via the `Format` type?

bmwill: I did just do a quick grep and it seems we only have things which are commented out, so we probably are good with this list for now.

bmwill (approved):
Thank you for doing this! This should be sufficient for me to build on top to expose these through the rpc layers

crates/sui-protocol-config-macros/src/lib.rs
alex-mysten: No - compile would fail if we had unsupported values

bmwill:
Oh! one more question, how does this new `render` function handle the difference between `attributes` and `flags`?

alex-mysten:
`render` only renders `attributes`, not `feature_flags`. There's a separate map in `get_epoch` for `feature_flags` so the typing is already appropriate.

bmwill:
I did notice that in `feature_flags` there are a number of entries that are not `bool`s. I suppose we'll continue to omit them for the time being?

Specifically:
```
    consensus_transaction_ordering: ConsensusTransactionOrdering,
    zklogin_supported_providers: BTreeSet<String>,
    per_object_congestion_control_mode: PerObjectCongestionControlMode,
    consensus_choice: ConsensusChoice,
    consensus_network: ConsensusNetwork,
    zklogin_max_epoch_upper_bound_delta: Option<u64>,
    mysticeti_num_leaders_per_round: Option<usize>,
```

Should these actually get moved out over to attrs?

mystenmark (approved):
Cool!!

aschran (approved):

alex-mysten:
@bmwill yes, those should move out of feature flags (they aren't flags), and then they would be included in Format. right now they're dropped. can address this as a followup

bmwill:
sounds good! Lets get this merged and i'll get working on the plumbing through the RPC layers

bmwill:
@alex-mysten One more question, from a conversation i had with @amnn, at this point is there any real value in separating feature flags and attributes or would it make sense to just merge these two sets of configs together?

alex-mysten:
@bmwill I don't really have an opinion on feature flags vs attributes, happy to merge them or not. I agree that the way they are interleaved now doesn't make a lot of sense.

Will merge as-is, I think the other items can be a followup.

PR #26669, opened 2026-05-15
[graphql-alt] Add object json test for simulate transaction

tpham-mysten:
## Description 
Adds a regression test for `object.asMoveObject.contents.json` on simulateTransaction output. The test publishes package_resolver_test and asserts contents.json resolves for the gas coin, UpgradeCap, and the newly-published SimpleObject (the last forces Scope::fetch to consult the per-simulation execution-object map for layout resolution). All three populate correctly at HEAD, so the reported gap doesn't reproduce here.

Includes a small graphql_redactions() helper to keep the snapshot stable.

## Test plan 

- e2e tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql_simulate_transaction_tests.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql_simulate_transaction_tests.rs
@@ -241,6 +241,33 @@ impl GraphQlTestCluster {
     }
 }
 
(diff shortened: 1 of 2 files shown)


amnn (approved):
LGTM, thanks @tpham-mysten

PR #26672, opened 2026-05-16
[sui-fork] reuse `latest` marker for object state (deleted, wrapped) 

stefan-mysten:
## Description 

This PR removes the `deleted|wrapped` markers for objects and moves them into the `latest` file, which now contains `version, deleted|wrapped`. The reason for this is to have fewer disk calls than previously.

## Test plan 

Update tests and added new tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-fork/src/filesystem.rs
+++ b/crates/sui-fork/src/filesystem.rs
@@ -11,7 +11,7 @@
 //! Under either root:
 //!     - objects/
 //!         - {object_id}/
-//!            - latest                  (text: <version>[,deleted|,wrapped])
+//!            - latest                  (text: version[,deleted|,wrapped])
 //!            - {version}                (BCS-encoded Object)
 //!     - indices/
 //!         - owned_objects              (BCS-encoded `Vec<OwnedObjectEntry>`)
--- a/crates/sui-fork/src/filesystem.rs
+++ b/crates/sui-fork/src/filesystem.rs
@@ -39,6 +39,7 @@ use std::path::PathBuf;
 use anyhow::Context as _;
 use anyhow::Error;
 use anyhow::bail;
+use anyhow::ensure;
 
 use move_core_types::language_storage::StructTag;
 use sui_types::base_types::ObjectID;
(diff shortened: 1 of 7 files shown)


crates/sui-fork/src/tests/filesystem.rs
```diff
@@ -127,13 +128,66 @@ fn object_dir(store: &FilesystemStore, object_id: &ObjectID) -> std::path::PathB
...
+        parse_object_latest_metadata("5,wrapped\n", path).unwrap(),
+        ObjectLatestMetadata {
+            version: 5,
+            state: ObjectLatestState::Wrapped,
+        },
+    );
+}
+
+#[test]
+fn test_parse_object_latest_metadata_rejects_invalid_formats() {
```
jordanjennings-mysten, line 169: lets say there was a corrupted metadata file? is there a good automatic/manual process for resolving? in theory you could just delete the folder right?

stefan-mysten: Yes, if the metadata is corrupted, we could:
- the user manually fixes it
- the tool could try to fix it -- requires some object analysis from historical transactions, so not part of this PR. We could think about it
- delete the whole folder (💥 option, not the best)
- reuse the seed file and try again in a different folder, but transactions made in the forked network will be lost.

In the current state, the recommended way is to just re-fork from the same checkpoint in a different folder or delete that folder.

crates/sui-fork/src/tests/filesystem.rs
```diff
@@ -187,99 +241,162 @@ fn test_latest_tracks_highest_written_version() {
...
+fn test_wrapped_latest_blocks_latest_preserves_exact_and_clears_on_live_write() {
     let (_dir, store) = test_store();
     let id = ObjectID::random();
     let owner = SuiAddress::random_for_testing_only();
     let obj = make_object_with_owner(id, 5, Owner::AddressOwner(owner));
 
     store.write_object(&obj).unwrap();
-    let object_ref = obj.compute_object_reference();
-    store.mark_object_wrapped(&object_ref).unwrap();
+    store.mark_object_as_wrapped(id, obj.version()).unwrap();
```
jordanjennings-mysten, line 288: nit/just as a broad thought around writing/flushing to disk `ObjectWrite` builder could encode valid transitions and check that transition makes sense against the current disk state. Then you could have a list of object writes that all get batched up together limiting the lock time and pushing all of the logic around consistency into one area of your code.

jordanjennings-mysten (approved):

crates/sui-fork/src/tests/filesystem.rs
stefan-mysten: That's definitely an interesting idea. I wonder what kind of changes are required to be able to do so. I'll have a look, I filed up an issue on linear.

crates/sui-fork/src/tests/filesystem.rs
amnn: The real solution here is to use a durable store -- like RocksDB.

crates/sui-fork/src/filesystem.rs
```diff
@@ -892,6 +913,40 @@ impl FilesystemStore {
...
 }
 
+/// Parse `objects/<id>/latest`, keeping live files numeric-only while accepting deleted/wrapped
+/// state suffixes for current-state removals.
+fn parse_object_latest_metadata(
+    content: &str,
+    path: &Path,
+) -> anyhow::Result<ObjectLatestMetadata> {
+    let content = content.trim();
+    let mut parts = content.split(',');
```
amnn: Try `split_once` for handling the case of "0 or 1" components:

```rust
let (version, state) = if let Some((version, part)) = content.trim().split_once(',') {
    ensure!(!state.is_empty(), "...");

    let state = ObjectLatestState::from_suffix(state)
        .with_context(|| format!("...", ...)?;

    (version, state)
} else {
    (version, ObjectLatestState::Live)
};

ensure!(!version.is_empty(), "...");
let version: u64 = version.parse().with_context(|| format!("...", ...))?;

Ok(ObjectLatestMetadata { version, state })
```

amnn: Also, this pattern of adding context using `with_context` at every fallible point in a function in order to add some information about the inputs is quite common. In those cases, it's better to factor out the context that includes the input information to one point:

```rust
// Before
fn foo(x: &X) -> anyhow::Result<Y> {
  let r = bar(...).with_context(|| format!("bar for {x}"))?;
  let z = baz(...).with_context(|| format!("baz for {x}"))?;
  ...
}

// After
fn foo(x: &X) -> anyhow::Result<Y> {
    foo_(x).with_context(|| format!("foo for {x}"))
}

fn foo_(x: &X) -> anyhow::Result<Y> {
    let r = bar(...).context("bar")?;
    let z = baz(...).context("baz")?;
    ...
}
```

crates/sui-fork/src/tests/filesystem.rs
amnn: You're both slowly reinventing a database, don't do that, just use a database.

crates/sui-fork/src/tests/filesystem.rs
stefan-mysten: Yes, I've looked into it but was not successful on my first attempt. Still planning to switch to that this week.

crates/sui-fork/src/filesystem.rs
stefan-mysten: Thanks for sharing this pattern, I will use it here as well.

PR #26673, opened 2026-05-16
[sui-fork] simplify seeding and owned object index data

stefan-mysten:
## Description 

This PR simplifies the seeding of owned objects to only keep object ref + owner instead of downloading object data at startup. Any owned object query will require to fetch the BCS to extract the needed data, the first time. This keeps lazy loading of objects uniform across normal transaction execution and owned objects query.
For subsequent requests, the owned-object index data keeps data for object type and balance, which make default owned object queries easy (grpc default read mask for owned object queries includes object type), avoiding deserializing object bcs.

The seed data will be used when the owned object index needs to be initialized.

## Test plan 

Existing CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-fork/design/poc.md
+++ b/crates/sui-fork/design/poc.md
@@ -85,7 +85,7 @@ At startup, the user has the choice to seed addresses or objects, recording enou
 
 
 If the user provides both `--address` and `--object`, the tool combines the object refs resolved from both sources and persists them to `seed_manifest.json`.
-The seed manifest is reused on restart and is the source for lazy owned-object index initialization.
+The seed manifest is reused on restart if needed, and is the source for lazy owned-object index initialization.
 
 ```json
 {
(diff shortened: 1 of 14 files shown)


crates/sui-fork/design/poc.md
```diff
@@ -78,37 +78,24 @@ As data is persisted to disk, the user needs to provide a directory where the ne
...
-```json
-{
-    "network": "testnet",
-    "checkpoint": 12345678,
-    "objects": [
-        "0xabcdef1234567890abcdef1234567890abcdef12"
+    "entries": [
+        {
+            "object_ref": ["0xabcdef1234567890abcdef1234567890abcdef12", 42, "..."],
+            "owner": "0x1234567890abcdef1234567890abcdef12345678"
```
amnn: Do you even need to store owner here?

amnn: In fact, do you need anything other than the ID?

amnn (commented):
Looks good, thanks @stefan-mysten -- I think you can simplify further though (see inline comments).

amnn (approved):

crates/sui-fork/design/poc.md
stefan-mysten: Probably not, I can remove.

PR #26674, opened 2026-05-16
[graphql-alt] Allowlist client SDK version metric label

tpham-mysten:
## Description 

- Replace the charset and length sanitization on the `client_sdk_version` Prometheus label with a per-SDK allowlist (`SDK_VERSION_ALLOWLIST`). Versions outside the allowlist appear as `other`, keeping label cardinality bounded by the size of the list rather than by the space of header values an attacker can construct.
 - The allowlist starts empty; adding a version is a one-line change in `logging.rs`. We only add an entry when we explicitly want to track adoption or retention of that version.
## Why

Follow-up to #26437. even with a 32-character cap and a `[A-Za-z0-9._-]` charset, the version label has roughly `65^32` possible values, which is a valid Prometheus cardinality DoS vector (precedent: SUIPR-379, a paid bug bounty on gRPC metrics in January). The allowlist closes that gap.

## Test plan 

- e2e test

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/extensions/logging.rs
+++ b/crates/sui-indexer-alt-graphql/src/extensions/logging.rs
@@ -39,6 +39,7 @@ use tracing::log::log_enabled;
 use tracing::warn;
 use uuid::Uuid;
 
+use crate::config::LoggingConfig;
 use crate::error::code;
 use crate::error::error_codes;
 use crate::error::fill_error_code;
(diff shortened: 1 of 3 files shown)


crates/sui-indexer-alt-graphql/src/extensions/logging.rs
```diff
@@ -53,17 +53,21 @@ pub const REQUEST_ID_HEADER: HeaderName = HeaderName::from_static("x-sui-rpc-req
...
 const SDK_TYPE_WHITELIST: &[&str] = &["rust", "typescript", "python"];
 
-/// Maximum length of a `client-sdk-version` value we accept verbatim.
-const MAX_SDK_VERSION_LEN: usize = 32;
+/// Per-SDK list of versions we emit verbatim as the `client_sdk_version` Prometheus label.
+/// Versions outside this list map to `CLIENT_LABEL_OTHER`, which keeps label cardinality
+/// bounded by the size of this list rather than by the space of possible header values.
+/// Add an entry only when we explicitly want to track adoption or retention of a specific
+/// SDK version.
+const SDK_VERSION_ALLOWLIST: &[(&str, &[&str])] = &[];
```
amnn: Can we make this a configuration option, so that we can update it without having to kick off a new build?

crates/sui-indexer-alt-graphql/src/extensions/logging.rs
```diff
@@ -347,18 +353,15 @@ fn sanitize_sdk_type(value: &str) -> String {
...
-            .bytes()
-            .all(|b| b.is_ascii_alphanumeric() || matches!(b, b'.' | b'-' | b'_'));
-
-    if valid {
+/// Sanitize a `client-sdk-version` header value before it is used as a Prometheus label.
+/// Returns the value verbatim if `(sdk_type, value)` is listed in `SDK_VERSION_ALLOWLIST`,
+/// otherwise `CLIENT_LABEL_OTHER`. Bounding label cardinality this way is what protects the
+/// metric against an adversarially-chosen version string.
+fn sanitize_sdk_version(sdk_type: &str, value: &str) -> String {
+    if SDK_VERSION_ALLOWLIST
```
amnn: When converting it into a configuration operation, let's store it as a set or map-like structure, to make this lookup potentially cheaper.

amnn (approved):
Thanks @tpham-mysten -- let's make this into a configuration option though.

crates/sui-indexer-alt-graphql/src/extensions/logging.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/extensions/logging.rs
tpham-mysten: Sure! Done

PR #26677, opened 2026-05-18
[move][move-decompiler]: add space before `as` in cast expressions

fgfm999:
## Description 

The pretty printer of move-decompiler used .concat() instead of .concat_space() for type cast operations, producing invalid output like `xas u128` instead of `x as u128`. Without the space, `xas` is tokenized as a single identifier, making the decompiled code fail to recompile.

## Test plan 

Unit test
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


fgfm999:
hi, @tzakian please take a look, thanks

PR #26679, opened 2026-05-18
[Block Streaming] broadcast randomness round signatures from Observer peer to client

akichidis:
## Description 

  - **Enables observer nodes to receive and verify randomness round signatures** from validators via the consensus block stream, allowing them to create RandomnessStateUpdate transactions without waiting for checkpoint sync.                                                                     
  - **Introduces RandomnessSignatureHandler trait in consensus-core** as the single interface between consensus and sui-core for both receiving and distributing signatures.                                                                                                                         
  - **Consolidates the randomness signature pipeline into RandomnessRoundReceiver** - one component that handles both locally-generated randomness (from RandomnessManager) and signatures relayed via consensus, with natural buffering during epoch transitions until DKG completes. The code has been constructed in such way so that Full Nodes that block stream, receive and process the randomness signatures, they also forward them to their own connected Observer clients via the same mechanism.

**Note:** the node is not persisting the randomness round signatures in order for the Observer clients (ex full nodes) stream historical signatures. _As with validators_ they'll need to rely on checkpoint state sync. The decision has been mainly made to make the solution simpler and have stronger liveness guarantees. No persistence is needed. No additional mechanics for on-demand fetching of missing signatures. No need to serve the signatures when commit sync. No accounting to correlate signatures to commit rounds.  Even if we were to serve the historical signature the benefit would be marginal, as for Full Nodes checkpoint state sync will be enabled anyways.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-core/src/authority.rs
akichidis, line 6624: Extracted on a separate file and extended further to process the signatures received from consensus

consensus/core/src/network/mod.rs
```diff
@@ -279,6 +298,9 @@ pub(crate) trait ObserverNetworkService: Send + Sync + 'static {
         peer: NodeId,
         commit_range: CommitRange,
     ) -> ConsensusResult<(Vec<TrustedCommit>, Vec<VerifiedBlock>)>;
+
+    /// Returns true if the local commit index is lagging too far behind the quorum.
+    fn is_commit_lagging(&self) -> bool;
```
mwtian: Should this remain in impl instead of adding to trait?

crates/sui-node/src/lib.rs
```diff
@@ -1278,6 +1285,7 @@ impl SuiNode {
         sui_node_metrics: Arc<SuiNodeMetrics>,
         checkpoint_metrics: Arc<CheckpointMetrics>,
         node_role: NodeRole,
+        randomness_receiver_handle: Arc<RandomnessRoundReceiverHandle>,
```
mwtian, line 1289: nit: add this after randomness_handle

crates/sui-node/src/lib.rs
```diff
@@ -1392,6 +1401,7 @@ impl SuiNode {
         sui_tx_validator_metrics: Arc<SuiTxValidatorMetrics>,
         admission_queue: Option<AdmissionQueueContext>,
         node_role: NodeRole,
+        randomness_receiver_handle: Arc<RandomnessRoundReceiverHandle>,
```
mwtian: nit: add this after randomness_handle

consensus/core/src/network/mod.rs
akichidis: We kind of have to do like this as the ObserverSubscriber is expecting an impl of a `ObserverNetworkService` .

mwtian (approved):

consensus/core/src/commit_vote_monitor.rs
```diff
@@ -13,6 +13,21 @@ use crate::{
     context::Context,
 };
 
+// Is used to calculate the threshold for blocking blocks or triggering sync
+// when the local commit index is lagging too far from the quorum commit index.
+pub(crate) const COMMIT_LAG_MULTIPLIER: u32 = 5;
+
+/// When a node is lagging behind peers in commits, we expect commit sync to be
+/// triggered to catch up.
+pub(crate) fn is_commit_lagging(
```
akichidis, line 22: I like the common method here. I believe it could be even part of the `CommitVoteMonitor` as every calling point requires the `CommitVoteMonitor` anyways. The `context` and `quorum_commit_index` are already available on the struct. The `current_index` is the one missing. To be fair, we could event inject the `DagState` and let it calculate everything it self.

mwtian: Feel free to update.

PR #26686, opened 2026-05-18
[move][move-decompiler] Refine `if` and rewrite into `assert` when possible.

cgswords:
## Description 

Two refinements on `IfElse`, stacked so the four assert shapes the structurer can produce all collapse to `assert!(cond, code); rest`. 

**1. `simplify_if` — hoist the non-terminating arm when the other arm always terminates.** Three local rewrites on `IfElse`:

  - then-arm `always_terminates` (`abort`/`return`/`break`/`continue`, or a `Seq`/`IfElse` that recursively does), else non-empty →
 hoist else out: `if (t) { terminator } else alt` → `if (t) { terminator }; alt`.
  - else-arm `always_terminates`, then non-empty → negate, swap, hoist: `if (t) { rest } else { terminator }` → `if (!t) { terminator }; rest`.
  - empty else → drop it.

When both arms terminate and rules 1 and 2 could both fire, the tiebreak prefers keeping an `Abort` in the conditional.

**2. Recover `assert!` from if/abort patterns.** A `recover_asserts` refinement rewrites the two empty-arm shapes:
  - `if (cond) { } else { abort code }`        →  `assert!(cond, code)`
  - `if (cond) { abort code }` (no/empty else) →  `assert!(!cond, code)`

These together recover a ton of the `assert!` patterns in the decompiled modules, looking like realistic `Move` code.

Also fixed:
- The `Exp::Call` Display impl was missing comma separators between arguments — latent until multi-arg `assert!` exercised it.
- Reorder `remove_trailing_continue`/`remove_trailing_return` to run before `simplify_if`. The structurer emits explicit empty `return` markers at function tails that `remove_trailing_return` strips; with the old order, `simplify_if`'s `always_terminates` predicate would see the not-yet-stripped trailing `return` and fire rule 1 on then-arms whose true Move-level shape is just statements falling off the function.

## Test plan

Updated tests to look great

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-decompiler/tests/bytecode/output/0xe1672e093ccbb2f477e460361040e1c7dd62e4b23615caafe05b7b3cf5ce7e25/crank.move
+++ b/external-crates/move/crates/move-decompiler/tests/bytecode/output/0xe1672e093ccbb2f477e460361040e1c7dd62e4b23615caafe05b7b3cf5ce7e25/crank.move
@@ -142,104 +142,70 @@ public(friend) fun update_config(l0: &mut CrankConfig, l1: u64, l2: u64, l3: u64
 
 public fun validate_heartbeat<T0>(l0: &mut CrankConfig, l1: &mut Vault<T0>, l2: &StrategyRegistry, l3: &Clock) {
     let l11 = clock::timestamp_ms(l3);
-    if (l11 - *(&l0.last_heartbeat_ms) >= *(&l0.min_heartbeat_interval_ms)) {
-        let l9 = strategy::borrow_protocol_access_cap(l2);
-        vault::update_heartbeat_tvl(l1, l3, l9);
-        *(&mut l0.last_heartbeat_ms) = l11;
-        let l10 = vault::last_heartbeat_tvl(freeze(l1));
-        let l5 = string::utf8(C14);
-        let l7 = string::utf8(C15);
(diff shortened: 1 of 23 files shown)


external-crates/move/crates/move-decompiler/tests/bytecode/output/0xe1672e093ccbb2f477e460361040e1c7dd62e4b23615caafe05b7b3cf5ce7e25/crank.move
```diff
@@ -142,104 +142,70 @@ public(friend) fun update_config(l0: &mut CrankConfig, l1: u64, l2: u64, l3: u64
...
-                } else {
-                    abort C7
-                }
-            } else {
-                abort C2
-            }
-        } else {
-            abort C1
-        }
+    assert!(strategy::count_available(l2) >= 2u64, C0);
```
tzakian, line 179: niiiiice

tzakian (approved):
Amazing. Really cleans up the output code considerably

PR #26691, opened 2026-05-19
[simulate] Add more tests for simulate tx (1/2)

stefan-mysten:
## Description 

This PR adds a few more simulate transaction related tests before refactoring that code.

Stack:
https://github.com/MystenLabs/sui/pull/26691 -- add more tests
https://github.com/MystenLabs/sui/pull/26692 -- move the simulate code into its own module

## Test plan 

CI & new tests.

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #26692, opened 2026-05-19
[simulate] Refactor the simulate transaction code into its own module (2/2)

stefan-mysten:
## Description 

This PR moves the `simulate_transaction` logic out from `AuthorityState` into its own module to make it easier to reuse. It also moves the `pre_object_load_checks` function out from `AuthorityState` type and into a free function that simulation module can import and call.

Stack:
https://github.com/MystenLabs/sui/pull/26691 -- add more tests
https://github.com/MystenLabs/sui/pull/26692 -- move the simulate code into its own module

## Test plan 

Existing CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -71,6 +71,7 @@ use std::{
 };
 use sui_config::NodeConfig;
 use sui_config::node::{AuthorityOverloadConfig, StateDebugDumpConfig};
+use sui_config::transaction_deny_config::TransactionDenyConfig;
 use sui_execution::Executor;
 use sui_protocol_config::PerObjectCongestionControlMode;
 use sui_types::accumulator_root::AccumulatorObjId;
@@ -133,7 +134,7 @@ use sui_storage::key_value_store_metrics::KeyValueStoreMetrics;
 use sui_types::accumulator_root::AccumulatorValue;
 use sui_types::authenticator_state::get_authenticator_state;
 use sui_types::balance::Balance;
-use sui_types::coin_reservation;
+use sui_types::coin_reservation::{self, CoinReservationResolverTrait};
 use sui_types::committee::{EpochId, ProtocolVersion};
 use sui_types::crypto::{AuthoritySignInfo, Signer};
 use sui_types::deny_list_v1::check_coin_deny_list_v1;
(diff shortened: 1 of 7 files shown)


crates/sui-core/src/simulation.rs
```diff
@@ -0,0 +1,353 @@
...
+        events: effects.events_digest().map(|_| inner_temp_store.events),
+        effects,
+        execution_result,
+        mock_gas_id,
+        unchanged_loaded_runtime_objects,
+        suggested_gas_price: reader.suggested_gas_price(&transaction),
+    })
+}
+
+fn pre_object_load_checks(
```
stefan-mysten: This should stay in authority.rs

crates/sui-core/src/authority.rs
```diff
@@ -963,6 +967,74 @@ pub struct AuthorityState {
...
 
+struct AuthoritySimulateTransactionReader<'a> {
+    state: &'a AuthorityState,
+    epoch_store: &'a AuthorityPerEpochStore,
+}
+
+impl SimulateTransactionReader for AuthoritySimulateTransactionReader<'_> {
+    fn read_inputs_for_simulation(
+        &self,
+        _tx_digest: &TransactionDigest,
```
alex-mysten, line 978: Presumably this is needed for Simulacrum?

crates/sui-core/src/simulation.rs
```diff
@@ -0,0 +1,314 @@
...
+    fn read_inputs_for_simulation(
+        &self,
+        tx_digest: &TransactionDigest,
+        input_object_kinds: &[InputObjectKind],
+        receiving_object_refs: &[ObjectRef],
+    ) -> SuiResult<(InputObjects, ReceivingObjects)>;
+
+    fn suggested_gas_price(&self, _transaction: &TransactionData) -> Option<u64> {
+        None
+    }
```
alex-mysten, line 65: IIRC, this was a point of contention on the original PR - can we completely remove this api if not needed in simulation?

crates/sui-core/src/simulation.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+#[derive(Debug, Copy, Clone)]
+pub struct SimulateTransactionOptions {
+    pub checks: TransactionChecks,
+    pub allow_mock_gas_coin: bool,
+}
+
+pub fn simulate_transaction_with_context<R: SimulateTransactionReader + ?Sized>(
```
alex-mysten, line 74: why not just `simulate_transaction`?

crates/sui-core/src/simulation.rs
stefan-mysten: Yes, I recall that. Simulacrum would need to implement its own gas price suggestion because it has no congestion tracker. For simuation, it gets called at the end: 
https://github.com/MystenLabs/sui/blob/8f0cf1bc0aa090a87f699f4589916b9a2c2f54e3/crates/sui-core/src/authority.rs?plain=1#L2627-L2636

My only worry here is that calling it before doing all the work around checks and object loading, etc., will affect this. Let me look again at it.

PR #26693, opened 2026-05-19
[consensus] v3 Committee

mwtian:
## Description 

Supports v3 API in `Committee`.

## Test plan 

CI

### DIFF
--- a/consensus/config/src/committee.rs
+++ b/consensus/config/src/committee.rs
@@ -25,14 +25,15 @@ pub type Stake = u64;
 pub struct Committee {
     /// The epoch number of this committee
     epoch: Epoch,
-    /// Total stake in the committee.
-    total_stake: Stake,
-    /// The quorum threshold (2f+1).
-    quorum_threshold: Stake,
-    /// The validity threshold (f+1).
-    validity_threshold: Stake,
     /// Protocol and network info of each authority.
     authorities: Vec<Authority>,
+    /// Total stakes in the committee.
+    total_stake: Stake,
+
+    /// Thresholds related to different fault tolerances.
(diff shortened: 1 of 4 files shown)


consensus/config/src/committee.rs
```diff
@@ -58,9 +59,92 @@ impl Committee {
...
+    }
+
+    /// Constructs a committee with thresholds derived from a hybrid fault budget
+    /// (`malicious_stake = f`, `crash_stake = c`).
+    ///
+    /// Nominally, the total stake is `nominal_total_stake = 5f + 3c + 1`;
+    /// and the thresholds evaluate to:
+    ///
+    /// - `validity_threshold                 = f + c + 1`
+    /// - `certification_threshold  = 2f + c + 1`
```
akichidis, line 79: nit: comment alignment

consensus/config/src/committee.rs
```diff
@@ -58,9 +59,92 @@ impl Committee {
...
+                    .checked_div(base_stake)
+                    .unwrap_or_else(|| panic!("Division error: {} {}", nominal, base_stake))
+            };
+            (scale(malicious_stake), scale(crash_stake))
+        } else {
+            // If both fault budgets are zero, there's nothing to scale.
+            (0, 0)
+        };
+
+        let validity_threshold = f + 1;
```
akichidis, line 120: Shouldn't validity be `f + c + 1` ?

mwtian: The comment was incorrect earlier. When we only need validity and not any intersection properties, validity can be `f + 1`.

akichidis (approved):
Discussed offline regarding validity. Approving.

PR #26697, opened 2026-05-19
fix(faucet): make `sui start --with-faucet` shut down on Ctrl+C

hayes-mysten:
## Summary

`start_faucet` calls `axum::serve(...).await` inline with no graceful-shutdown signal wired up, so it blocks forever. As a result, the ctrl-c branch of the health-check loop in `sui_commands::start` is never reached, and `rpc_services.shutdown()` never runs — so `sui start --with-faucet` either appears to ignore SIGINT or exits without cleanly stopping the indexer / GraphQL / consistent-store / faucet.

Verified on `main`: with `--with-faucet`, sending SIGINT to the process never produces the `"Received Ctrl+C, shutting down..."` log line that the health-check loop emits — confirming the loop is unreachable.

## Fix

Refactor `start_faucet` to return a `sui_futures::service::Service` that wraps `axum::serve(...).with_graceful_shutdown(...)`, matching the shape of `start_graphql`, `start_consistent_store`, and the indexer. `sui_commands::start` merges the returned `Service` into `rpc_services`, so the faucet shuts down through the same `rpc_services.shutdown()` path as everything else.

The standalone `sui-faucet` binary now drives the returned `Service` via `Service::main()`, which already handles SIGINT/SIGTERM with a grace period.

## Test plan

- [x] `cargo check -p sui-faucet -p sui` clean
- [x] `cargo xclippy` clean for `sui-faucet` and `sui`
- [x] `cargo nextest run -p sui-faucet` — all 3 tests pass, including the server endpoint test that now uses the new `Service`-based API
- [ ] Reviewer to confirm Ctrl+C cleanly shuts down `sui start --with-faucet` in their environment

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: `sui start --with-faucet` now responds to Ctrl+C / SIGINT and shuts down all subservices gracefully.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14839,6 +14839,7 @@ dependencies = [
  "serde_json",
  "shared-crypto",
  "sui-config",
+ "sui-futures",
  "sui-keys",
  "sui-rpc-api",
  "sui-sdk",
--- a/crates/sui-faucet/Cargo.toml
+++ b/crates/sui-faucet/Cargo.toml
@@ -23,6 +23,7 @@ tracing.workspace = true
 sui-sdk.workspace = true
 sui-rpc-api.workspace = true
 sui-config.workspace = true
+sui-futures.workspace = true
 sui-keys.workspace = true
 shared-crypto.workspace = true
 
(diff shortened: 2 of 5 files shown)


amnn (approved):
LGTM!

stefan-mysten (approved):
Thanks for fixing this @hayes-mysten!

PR #26700, opened 2026-05-19
docs: validate gasless token table against lib.rs at build time

jessiemongeon1:
## Summary
- Add `validate-gasless-tokens.mjs` build script that parses `MAINNET_*` constants and the active `Chain::Mainnet` allowlist from `crates/sui-protocol-config/src/lib.rs` and compares them against the token table in the gasless stablecoin transfers doc page
- Build fails with a side-by-side diff if the table is out of date, preventing docs from going stale when tokens are added or removed from the protocol config
- Table stays inline in the MDX (not a snippet/import) so it renders correctly in markdown exports and LLM ingestion (`llms.txt`)
- Issuer-to-token mapping is maintained in the script since issuer names are not in the Rust source

## Test plan
- [x] `node scripts/validate-gasless-tokens.mjs` passes with current table
- [x] Validation correctly fails (exit 1) when table content is modified
- [ ] `bash scripts/build-and-check.sh` runs validation as part of the pre-build steps

### DIFF
--- a/docs/content/develop/transaction-payment/gasless-stablecoin-transfers.mdx
+++ b/docs/content/develop/transaction-payment/gasless-stablecoin-transfers.mdx
@@ -4,12 +4,6 @@ description: Gasless stablecoin transfers on Sui enable peer-to-peer payments of
 keywords: [ usdc, stablecoins, sui dollar, usdsui, usdy, ausd, fdusd, buck, ysld, gasless, free tier ]
 ---
 
-:::caution 
-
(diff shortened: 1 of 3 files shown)


jon-mysten (approved):

jessiemongeon1:
## Style Guide Audit

### `docs/content/develop/transaction-payment/gasless-stablecoin-transfers.mdx` (1 violation(s))

- **Line 15** — No em dashes in prose
  - Current: `—`
  - Fix: `Rewrite with comma, parentheses, or split sentence`

---

PR #26702, opened 2026-05-19
[decompiler] Inline let-bindings whose RHS is a single-use variable

cgswords:
## Description 

Add a refinement that recognizes `let X = Y;` where Y is itself a variable, Y has exactly one read in the function body (this binding), and X has at most one read per execution path with no re-assignment. In that case the binding is dropped and the use of X is rewritten to Y.

## Test plan 

Snapshot updates

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


cgswords:
Superseded — the two commits from this PR (extract liveness module + inline single-use let-bindings) landed in main as part of #26802. Closing as the stack was merged starting from #26802 onto main directly.

PR #26703, opened 2026-05-19
Redo list api progress propagation

nickvikeras:
## Description

Cargo.toml currently is doing a local override of the rust sdk. I'll fix that once the proto changes land.

The general change in this PR is:

- At the bucket boundaries, a tx_seq watermark is propagated down the stream.
- Each pipeline stage "coalesces" the watermarks and only forwards the newest one.
- In the final stage of the stream, we look up the checkpoint sequence number for the tx_seq watermarks. If any items pass through the stream while that's in flight, we just cancel it (since the items have their own progress marker, no sense it holding up the flow). Otherwise we write the cursor + checkpoint-level watermark to the stream. Newer watermarks don't cancel in-flight watermark lookups, it's better to let them finish to ensure progress is tracked, but they are "debounced."


crates/sui-inverted-index/src/bitmap_query.rs
```diff
@@ -211,60 +482,159 @@ fn term_bucket_stream<S>(
...
+                )
+            } else {
+                (
+                    bucket_end_exclusive.min(range.end),
+                    bucket_start.max(range.start),
+                )
+            };
+            yield Watermarked::Watermark(pre);
+            yield Watermarked::Item(item);
+            yield Watermarked::Watermark(post);
```
bmwill, line 653: So if we yield these i assume they get debounced and folded into an inline watermark when returned to a user?

crates/sui-kv-rpc/src/v2alpha/list_checkpoints.rs
```diff
@@ -243,38 +321,163 @@ pub(crate) async fn list_checkpoints(
...
+fn boundary_watermark(
+    options: &QueryOptions,
+    cp_seq: u64,
+    checkpoint_boundary: Option<u64>,
+) -> Watermark {
+    let mut wm = Watermark::default();
+    wm.cursor = Some(options.cursor_for_boundary(cp_seq, cp_seq));
+    set_checkpoint_bound(&mut wm, options, checkpoint_boundary);
+    wm
+}
```
bmwill: I *think* this logic is duplicated in all the list apis, maybe it would be better to have a shared watermark construction logic?

crates/sui-kv-rpc/src/query_options.rs
```diff
@@ -18,7 +18,6 @@ use sui_rpc::proto::sui::rpc::v2alpha::QueryOptions as ProtoQueryOptions;
 
 const ORDERING_ASCENDING: i32 = ProtoOrdering::Ascending as i32;
 const ORDERING_DESCENDING: i32 = ProtoOrdering::Descending as i32;
-pub(crate) const MAX_CHECKPOINT_SCAN_WIDTH: u64 = 3_000_000;
```
bmwill, line 21: Now that this is gone, do we have a backstop for unfiltered queries? We always have the timeout, and for txns and checkpoints i suppose we have the max page size as a limit, but what about events? I suppose once we have the new event bitmap you could actually use that to handle unfiltered event queries.

crates/sui-kv-rpc/src/v2alpha/list_events.rs
```diff
@@ -348,14 +527,24 @@ async fn fetch_txs_for_refs(
...
+
 async fn render_event(
     event_ref: EventRef,
     tx: TransactionData,
     read_mask: &FieldMaskTree,
     resolver: &PackageResolver,
     wants_json: bool,
-    options: &QueryOptions,
-) -> Result<ListEventsResponse, RpcError> {
+    _options: &QueryOptions,
```
bmwill: remove?

crates/sui-kv-rpc/src/v2alpha/list_checkpoints.rs
```diff
@@ -118,30 +127,44 @@ pub(crate) async fn list_checkpoints(
             elapsed_ms = started.elapsed().as_millis(),
             "list_checkpoints: empty range"
         );
-        return Ok(stream::once(async move { Ok(end_response(end_reason, end_cursor)) }).boxed());
+        return Ok(stream::once(async move { Ok(end_response(end_reason)) }).boxed());
```
bmwill: follow up question to my question on: https://github.com/MystenLabs/sui-rust-sdk/pull/264#discussion_r3269657154

So now `EndFrames` don't have a watermark, i assume that the previous item will generally always have a watermark that is up-to-date with the end of what we scanned?

crates/sui-kv-rpc/src/v2alpha/list_events.rs
```diff
@@ -185,47 +216,204 @@ pub(crate) async fn list_events(
...
+                        checkpoint_boundary,
+                    );
+                    emitted += 1;
+                    yield event_item_response(rendered.item, wm);
+                }
+                Ok(ResolvedWatermarked::Watermark { position, cp }) => {
+                    checkpoint_boundary = advance_checkpoint_boundary(checkpoint_boundary, cp, &options);
+                    let wm = boundary_watermark(&options, cp, position, checkpoint_boundary, direction);
+                    yield watermark_response(wm);
+                }
```
bmwill, line 260: This PR is large so i probably haven't found it yet, but do we smash watermarks together or into Item frames before we send them to the client in the event we have sufficient "ready" values coming out of the stream?

crates/sui-kv-rpc/src/v2alpha/list_transactions.rs
```diff
@@ -190,54 +249,209 @@ pub(crate) async fn list_transactions(
...
+                }
+                Ok(ResolvedWatermarked::Watermark { position, cp }) => {
+                    checkpoint_boundary = advance_checkpoint_boundary(checkpoint_boundary, cp, &options);
+                    let wm = boundary_watermark(&options, cp, position, checkpoint_boundary, direction);
+                    yield watermark_response(wm);
+                }
+                Err(e) => {
+                    if error_contains::<BitmapScanLimitExceeded>(&e).is_some() {
+                        scan_limit_hit = true;
+                        break;
```
bmwill, line 305: If the Scan limit is hit do we have guarantees around the watermarks that have been emitted before? That is do have have an Item or Watermark frame before this that is up to (and maybe including the bucket) where we hit the scan limit? Or should we encode the watermark in the error so that we can pull it out here?

crates/sui-kv-rpc/src/bigtable_client.rs
```diff
@@ -64,13 +63,30 @@ pub struct ConcurrencyConfig {
...
     /// consume downstream request permits.
     pub max_bitmap_filter_literals: usize,
+    /// Per-request evaluated-bucket budget for filtered tx-bitmap scans,
+    /// shared across all DNF dimensions of one query. Caps how many fetched
+    /// buckets the eval evaluates, NOT how many bucket reads BigTable
+    /// receives — at exhaustion each leaf stream may have fetched one
+    /// additional bucket that is discarded rather than evaluated, so
+    /// observed reads can exceed this by up to `max_bitmap_filter_literals`.
+    /// Bounds wall time and resource cost; clients page transparently via
+    /// the returned cursor on SCAN_LIMIT.
```
bmwill, line 73: I guess this may answers my question above about what happens when you hit the scan limit?

crates/sui-kv-rpc/src/pipeline.rs
```diff
@@ -6,57 +6,256 @@ use std::collections::HashMap;
...
+                }
+            };
+
+            // Drain everything else that's immediately ready. Only Items count
+            // toward chunk_size; watermarks are free.
+            let mut next = Some(first);
+            loop {
+                let w = match next.take() {
+                    Some(w) => w,
+                    None => match futures::poll!(upstream.as_mut().next()) {
```
bmwill, line 166: Whats the motivation for doing this vs using something like ready_chunks?

crates/sui-kv-rpc/src/query_options.rs
```diff
@@ -354,31 +349,20 @@ impl CheckpointRange {
...
-                    end_reason: reason,
-                }
-            }
+        // Scan limit is now enforced at runtime by `BitmapScanBudget`
+        // metering bucket fetches across all DNF dimensions; the historical
+        // up-front checkpoint-width clamp would produce a misleading scan
+        // frontier under that model. Whatever range the request asks for is
+        // honored here; handlers terminate filtered scans on
+        // `BitmapScanLimitExceeded` and emit a `SCAN_LIMIT` end with the
+        // latest in-stream `Watermark.cursor` as the resume point.
```
bmwill: idk if it makes sense to leave a comment in about how things "historically" were done.

crates/sui-inverted-index/src/bitmap_query.rs
nickvikeras: Keep in mind the 'item' here is a bucket, not a transaction sequence number.  And this is before we have done any of the boolean bitmap logic to evaluate the query. So these get propagated forward per bucket, and if there are any items left after all the boolean logic, they will be dropped by the debouncing later.

crates/sui-kv-rpc/src/query_options.rs
nickvikeras: Txns/checkpoints are naturally bounded by `limit_items`. I bounded events explicitly in https://github.com/MystenLabs/sui/pull/26710/commits

crates/sui-kv-rpc/src/v2alpha/list_checkpoints.rs
nickvikeras: Yeah. You will always get one of these:

1. The watermark confirming you scanned to the end of the checkpoint bounds/chain history.
2. The last item if you hit `limit_items` (so if it's a tx in Checkpoint N, the watermark is N-1).
3. The watermark from the last bucket scanned if you hit a bucket scan limit before `limit_items`.

And they will all contain the cursor for that point as well. 

You might also get watermarks for intermittently if there are buckets that resolved to no items depending on timing (if items land before we can resolve the watermarks checkpoint, we don't bother to propagate it).

crates/sui-kv-rpc/src/pipeline.rs
nickvikeras: Mainly to avoid counting watermarks in the chunk size since that is what determines the bigtable batch size. But maybe that's an over-optimization.

crates/sui-kv-rpc/src/query_options.rs
nickvikeras: yeah claude started doing this heavily when refactoring Greenfield code. It's infuriating. I missed deleting this one.

crates/sui-kv-rpc/src/v2alpha/list_checkpoints.rs
nickvikeras: yeah cleaned it up

crates/sui-kv-rpc/src/bigtable_client.rs
nickvikeras: This comment made me realize hitting the scan limit could cancel an in-flight checkpoint watermark resolution, so I fixed that, but yeah.

crates/sui-kv-rpc/src/v2alpha/list_events.rs
nickvikeras: Yeah. The process is roughly described in the PR description. But we coalesce them as much as we can and never emit one if there is an item ready in the stream.

crates/sui-kv-rpc/src/pipeline.rs
```diff
@@ -865,7 +865,27 @@ where
...
+                            // lookups.
+                            if let Ok(Some(cp)) = fut.as_mut().await {
+                                yield ResolvedWatermarked::Watermark { position, cp };
+                            }
+                            if let Some(pp) = pending.take() {
+                                if let Ok(Some(cp)) = resolver(pp).await {
+                                    yield ResolvedWatermarked::Watermark { position: pp, cp };
+                                }
+                            }
+                            Err(e)?;
```
bmwill, line 897: This is wild, i had just asked claude about this case and it gave me verbatim this "fix"

bmwill (approved):

PR #26706, opened 2026-05-19
ci(release): switch brew bump from mislav action to direct `brew bump-formula-pr`

ebmifa:
## Summary

Replace the `mislav/bump-homebrew-formula-action` step in `update-homebrew-formula` with a direct `brew bump-formula-pr` invocation. Same secret, same outcome, no third-party action dependency.

## Why

Since ~2026-05-16, every release tag we create on testnet has the homebrew-bump job fail with:

```
Error: unexpected HTTP 303 response
```

The cause is upstream and out of our control: GitHub now returns HTTP **303 See Other** (instead of 302) for codeload tarball redirects when a bearer token is attached. The action's strict `statusCode == 302` check rejects 303 and bails before writing the formula.

- Tracking issue (open, unfixed): [`mislav/bump-homebrew-formula-action#340`](https://github.com/mislav/bump-homebrew-formula-action/issues/340)
- Even the latest release (`v4.1`, March 2026) has the same check — bumping the pin doesn't help.
- No PR is in flight; recent activity on the action is dependabot-only.

The fix is one line in the action, but the maintainer hasn't shipped it. Rather than fork-and-maintain, switch to the canonical Homebrew CLI command. `brew bump-formula-pr` is what humans run, what `cli/cli` and `astral-sh/uv` use in their release workflows, and what Homebrew's own BrewTestBot uses internally.

## Diff highlights

- Replace `uses: mislav/bump-homebrew-formula-action@...` with `uses: Homebrew/actions/setup-homebrew@master` + a `run:` block invoking `brew bump-formula-pr`.
- Use **`--url=` and `--version=`** (not `--tag/--revision`). Sui's formula in `Homebrew/homebrew-core` is URL+sha256 style:
  ```ruby
  url "https://github.com/MystenLabs/sui/archive/refs/tags/testnet-vX.Y.Z.tar.gz"
  sha256 "..."
  ```
  not git tag+revision. `brew bump-formula-pr` cannot switch a formula between styles, so the `--url` form is required.
- **Do not pass `--no-fork`**. The `HOMEBREW_GH_FORMULA_BUMP` PAT is not a homebrew-core maintainer; it cannot push directly. Use the default fork-then-PR flow (fork into the PAT user's GitHub, push, open PR upstream).
- Reuse the existing `secrets.HOMEBREW_GH_FORMULA_BUMP` PAT (same `public_repo,workflow` scopes); just rename the env var from `COMMITTER_TOKEN` to `HOMEBREW_GITHUB_API_TOKEN` (the variable `brew` actually reads).
- `--no-browse` for non-interactive CI behavior.
- `--message=` keeps a link back to the originating release run for audit.
- Configure `git` author as `github-actions[bot]` so the commit attribution matches our usual bot pattern.

## Test plan

- [ ] Verify the next testnet release tag (`testnet-v1.72.3` or `testnet-v1.73.x`) triggers `update-homebrew-formula` and the job succeeds.
- [ ] Confirm a PR appears against `Homebrew/homebrew-core` from the github-actions[bot] / PAT-user account with the correct `url` (`...archive/refs/tags/testnet-vX.Y.Z.tar.gz`) and an accurate `sha256` (auto-computed by `brew` from the tarball).
- [ ] Confirm the PR description includes the release-run link via `--message=`.
- [ ] If `brew bump-formula-pr` exits non-zero (e.g., sha256 cache miss on Homebrew's CDN), retry by re-running the job.

## Out of scope

- No change to `secrets.HOMEBREW_GH_FORMULA_BUMP` or its permissions.
- No change to the trigger condition (still testnet-only by `inputs.sui_tag` / `github.ref_name`).
- No change to where the formula lives (still `Homebrew/homebrew-core`, not a custom MystenLabs tap — bypassing homebrew-core would be worse UX).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -259,10 +259,13 @@ jobs:
           git config --global user.name "github-actions[bot]"
           git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
 
+          # Sui's formula in Homebrew/homebrew-core is URL+sha256 style, not tag+revision,
+          # so pass --url/--version (brew bump-formula-pr cannot switch styles).
+          url="https://github.com/MystenLabs/sui/archive/refs/tags/${{ env.sui_tag }}.tar.gz"
           brew bump-formula-pr \
             --no-browse \
-            --no-fork \
-            --tag="${{ env.sui_tag }}" \
+            --url="${url}" \
+            --version="${{ env.versionless_tag }}" \
             --message="From release: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" \
             sui
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -260,8 +260,15 @@ jobs:
           HOMEBREW_NO_ANALYTICS: "1"
           HOMEBREW_NO_AUTO_UPDATE: "1"
         run: |
-          git config --global user.name "github-actions[bot]"
-          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          # Derive commit author from the PAT owner so the homebrew-core PR
+          # shows the right "Verified" attribution and survives token rotation
+          # without editing this file.
+          user_info=$(curl -fsSL -H "Authorization: Bearer ${HOMEBREW_GITHUB_API_TOKEN}" \
+            -H "Accept: application/vnd.github+json" https://api.github.com/user)
+          gh_login=$(echo "$user_info" | jq -r '.login')
+          gh_id=$(echo "$user_info" | jq -r '.id')
+          git config --global user.name "${gh_login}"
+          git config --global user.email "${gh_id}+${gh_login}@users.noreply.github.com"
 
           # Sui's formula in Homebrew/homebrew-core is URL+sha256 style, not tag+revision,
           # so pass --url/--version (brew bump-formula-pr cannot switch styles).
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -244,20 +244,30 @@ jobs:
...
-          # https://github.com/settings/tokens/new?scopes=public_repo,workflow
-          COMMITTER_TOKEN: ${{ secrets.HOMEBREW_GH_FORMULA_BUMP }}
+          # PAT with public_repo,workflow scopes — same secret previously used as COMMITTER_TOKEN
+          HOMEBREW_GITHUB_API_TOKEN: ${{ secrets.HOMEBREW_GH_FORMULA_BUMP }}
+          HOMEBREW_NO_INSTALL_FROM_API: "1"
+          HOMEBREW_NO_ANALYTICS: "1"
+          HOMEBREW_NO_AUTO_UPDATE: "1"
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
johnjmartin: since it's my PAT does this need to be my email?

ebmifa: Good catch — yes. Pushed a follow-up that derives the git author dynamically from the PAT owner via `GET /user`, so the commit's `user.name` / `user.email` always matches whoever the PAT belongs to (currently you). That gives the right "Verified" attribution on the homebrew-core PR, and the workflow doesn't need editing if the token ever rotates to a different owner.

Commit: 846d6076

johnjmartin (approved):

PR #26707, opened 2026-05-19
[sui-fork] move owned object index to use a dbmap

stefan-mysten:
## Description 

This PR migrates from a file-based owned object index to a DbMap from typed-store. The reason for this is that we maintained manually the ordered owned-object index in the filesystem, and had functions for updates/removals. The `DBMap` from `typed-store` is doing exactly that, so @amnn suggested to use it instead.

## Test plan 

Existing CI.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-fork/src/filesystem.rs
```diff
@@ -202,12 +252,50 @@ impl FilesystemStore {
...
             None => Self::default_root(node, forked_at_checkpoint)?,
         };
-        Ok(Self { root })
+        Ok(Self::new_with_root(root))
     }
 
     /// Create a filesystem store with an explicit root directory.
     pub(crate) fn new_with_root(root: PathBuf) -> Self {
-        Self { root }
+        let owned_index = Self::open_owned_object_index(&root);
```
stefan-mysten: I am unsure of this. We're basically open the DB at startup and keeping it open. Alternatively, we could open/close everytime we need to do something around owned objects. Not sure what's the right way to go here.

This also requires to create a new tokio runtime and block on it for reading the db for file.

crates/sui-fork/src/filesystem.rs
```diff
@@ -184,10 +195,49 @@ impl OwnedObjectEntry {
...
+/// lets RPC pagination seek directly to `(owner, cursor_object_id)`.
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, serde::Deserialize, serde::Serialize)]
+struct OwnedObjectOwnerKey {
+    /// Address whose object list contains `object_id`.
+    owner: SuiAddress,
+    /// Object ID used for deterministic ordering and cursor lower bounds within one owner.
+    object_id: ObjectID,
+}
+
+impl From<&OwnedObjectEntry> for OwnedObjectOwnerKey {
```
stefan-mysten: This should be down.

crates/sui-fork/src/filesystem.rs
```diff
@@ -184,10 +195,49 @@ impl OwnedObjectEntry {
...
+struct OwnedObjectIndexMetadata {
+    /// Schema version of the typed-store owned-object index.
+    version: u64,
+}
+
+/// Typed-store tables for the owned-object index.
+#[derive(DBMapUtils)]
+struct OwnedObjectIndexTables {
+    /// Initialization marker and schema version.
+    meta: DBMap<(), OwnedObjectIndexMetadata>,
```
stefan-mysten: this is used to distinguish between never initialized, and initialized but empty. It is persisted in the db, so on restarts we do not need to worry about re-initializing the owned object index / checking if it was initialized.

crates/sui-fork/src/filesystem.rs
```diff
@@ -202,12 +252,50 @@ impl FilesystemStore {
...
+        let path = root.join(INDICES_DIR).join(OWNED_OBJECTS_INDEX_DB_DIR);
+        if tokio::runtime::Handle::try_current().is_ok() {
+            return Self::open_owned_object_index_at(path);
+        }
+
+        let runtime = tokio::runtime::Builder::new_current_thread()
+            .enable_time()
+            .build()
+            .expect("tokio runtime should build for typed-store open");
+        runtime.block_on(async move { Self::open_owned_object_index_at(path) })
```
stefan-mysten: This is one aspect that I am currently not sure about.

crates/sui-fork/src/filesystem.rs
```diff
@@ -618,51 +723,107 @@ impl FilesystemStore {
...
+    ///
+    /// Removal starts from `by_object` because execution diffs often know only the object ID; the
+    /// stored entry provides the previous owner key needed to clean up `by_owner`.
+    fn delete_owned_entry(&self, batch: &mut DBBatch, object_id: ObjectID) -> anyhow::Result<()> {
+        if let Some(existing) = self.owned_index.by_object.get(&object_id)? {
+            batch.delete_batch(&self.owned_index.by_object, [object_id])?;
+            batch.delete_batch(
+                &self.owned_index.by_owner,
+                [OwnedObjectOwnerKey::from(&existing)],
+            )?;
```
stefan-mysten: What happens if first one succeeds, but the second one fails?

stefan-mysten: Ah, yes. So this is just batching data into memory. The actual write to the DB happens in `batch.write()` calls in a couple of places.

crates/sui-fork/src/filesystem.rs
```diff
@@ -184,10 +195,49 @@ impl OwnedObjectEntry {
     }
 }
 
+/// Ordered owner index key. Scans over this table group rows by owner, then by object ID, which
+/// lets RPC pagination seek directly to `(owner, cursor_object_id)`.
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, serde::Deserialize, serde::Serialize)]
+struct OwnedObjectOwnerKey {
+    /// Address whose object list contains `object_id`.
+    owner: SuiAddress,
```
amnn: You may need to also have the object type here. `(owner, type, id)`

crates/sui-fork/src/filesystem.rs
```diff
@@ -184,10 +195,49 @@ impl OwnedObjectEntry {
...
+    version: u64,
+}
+
+/// Typed-store tables for the owned-object index.
+#[derive(DBMapUtils)]
+struct OwnedObjectIndexTables {
+    /// Initialization marker and schema version.
+    meta: DBMap<(), OwnedObjectIndexMetadata>,
+    /// Canonical row per object ID, used when update paths only know the object ID.
+    by_object: DBMap<ObjectID, OwnedObjectEntry>,
```
amnn: - Rather than storing the entry containing everything, just store the version here, and fetch the rest of the stuff from the object store.
- The easiest way to update these indices is to use the input state to clean up old records, and the output state to write new records.

crates/sui-fork/src/filesystem.rs
```diff
@@ -184,10 +195,49 @@ impl OwnedObjectEntry {
...
+    /// Owner-ordered rows used by `list_owned_objects` and RPC pagination.
+    by_owner: DBMap<OwnedObjectOwnerKey, OwnedObjectEntry>,
+}
+
 /// Local filesystem-backed store for Sui data.
 #[derive(Clone)]
 pub(crate) struct FilesystemStore {
     root: PathBuf,
+    /// DBMap-backed owned-object index. Clones share one table handle for this store root.
+    owned_index: Arc<OwnedObjectIndexTables>,
```
amnn: double check whether you need the Arc here or whether `DbMap` is cheap to clone.

crates/sui-fork/src/filesystem.rs
amnn: yeah, not ideal, but we can live with it for now.

crates/sui-fork/src/filesystem.rs
amnn: Take a look at how these indices are set-up for the consistent store:

- https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt-consistent-store/src/handlers/object_by_owner.rs
- https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt-consistent-store/src/handlers/object_by_type.rs

amnn (commented):
Comments from our pairing session

crates/sui-fork/src/filesystem.rs
jordanjennings-mysten: On windows in particular you probably want to just block out other processes from accessing the db. If someone is creating two processes that are attempting to take over that file we probably want the second process to fail instead. Especially since your concurrency model doesn't take into account multiple processes due to its in memory lock I think? So instead of some undefined behavior just fail faster there, in fact I'd be even more explicit and have something that indicates another process is using the folder.

crates/sui-fork/src/filesystem.rs
jordanjennings-mysten: might be a bit cumbersome but it almost feels like the dependency graph should be
```
store -> owned_index
      -> filesystem

instead of 
store -> filesystem -> owned_index
````
guess it depends on how much crosstalk between filesystem and owned_index there ends up being.

if you split them out they can also more cleanly swap in new backends in the future independently (file, redis, postgres).

Comment (stefan-mysten) [reply crates/sui-fork/src/filesystem.rs]: I would say that this is how it is in this PR, `store -> owned_index`, and the `owned_index` is stored on disk via DB.

Comment (stefan-mysten) [reply crates/sui-fork/src/filesystem.rs]: Or are you referring that this should belong to `DataStore` type than in `FileSystemStore`.

File: crates/sui-fork/src/filesystem.rs (hunk 6 above)
Comment (stefan-mysten) [reply crates/sui-fork/src/filesystem.rs]: Added that and an optional inverted balance.

File: crates/sui-fork/src/filesystem.rs (hunk 7 above)
Comment (stefan-mysten) [reply crates/sui-fork/src/filesystem.rs]: I kept only one table: OwnedObjectIndexKey to version. The key has owner, type, inverted balance, and id. This should be sufficient for most owned objects query through grpc, I think.

File: crates/sui-fork/src/filesystem.rs (hunk 8 above)
Comment (stefan-mysten) [reply crates/sui-fork/src/filesystem.rs]: Removed!

Comment (stefan-mysten) [reply crates/sui-fork/src/filesystem.rs]: There is now a OwnedObjectIndexStore that is part of DataStore and not directly on FileSystemStore.

Comment (jordanjennings-mysten) [reply crates/sui-fork/src/filesystem.rs]: lgtm

File: crates/sui-fork/src/tests/owned_object_index.rs [hunk 9]
```diff
@@ -0,0 +1,244 @@
...
+    OwnedObjectInfo {
+        owner,
+        object_type: object.struct_tag().expect("test object should have a type"),
+        balance: object.as_coin_maybe().map(|coin| coin.value()),
+        object_id: object.id(),
+        version: object.version(),
+    }
+}
+
+fn assert_info_matches_object(info: &OwnedObjectInfo, object: &Object) {
```
amnn: If you want to abstract assertions, use macros rather than functions, because otherwise if the underlying assertion ever fails, the resulting panic will not tell you which test failed.

crates/sui-fork/src/tests/owned_object_index.rs
```diff
@@ -0,0 +1,244 @@
...
+    let infos = store.scan_owner(owner, None, None).unwrap();
+    assert_eq!(infos.len(), 2);
+    assert!(infos[0].object_id < infos[1].object_id);
+    assert!(infos.iter().all(|info| info.owner == owner));
+    assert!(
+        infos
+            .iter()
+            .all(|info| info.object_type == GasCoin::type_())
+    );
+    assert!(infos.iter().all(|info| info.balance == Some(1_000_000)));
```
amnn: We have had a good experience using `insta` snapshot tests for similar assertions when writing E2E tests for RPCs. Typically we would use [JSON snapshots](https://docs.rs/insta/latest/insta/macro.assert_json_snapshot.html) (serialize data as JSON and then take a snapshot of that), and make use of [redactions](https://insta.rs/docs/redactions/) to clean up artifacts that change from run-to-run (specific addresses, digests, etc)

crates/sui-fork/src/owned_object_index.rs
```diff
@@ -0,0 +1,451 @@
...
+        MetricConf {
+            read_sample_interval: SamplingInterval::new(Duration::ZERO, 0),
+            write_sample_interval: SamplingInterval::new(Duration::ZERO, 0),
+            iter_sample_interval: SamplingInterval::new(Duration::ZERO, 0),
+            ..MetricConf::new("sui-fork-owned-index")
+        }
+    }
+
+    /// Read all object infos from the owned-object index.
+    pub(crate) fn get_owned_object_infos(&self) -> anyhow::Result<Vec<OwnedObjectInfo>> {
```
amnn, line 183: What is this used for? Usually it would be a bad idea to offer an API to load everything from RocksDB into memory.

crates/sui-fork/src/owned_object_index.rs
```diff
@@ -0,0 +1,451 @@
...
+    ) -> anyhow::Result<Vec<OwnedObjectInfo>> {
+        let lower = cursor
+            .map(OwnedObjectIndexKey::from_cursor)
+            .unwrap_or_else(|| OwnedObjectIndexKey::lower_bound(owner, object_type));
+
+        self.tables
+            .objects
+            .safe_iter_with_bounds(Some(lower), None)
+            .take_while(|entry| {
+                let Ok((key, _)) = entry else {
```
amnn, line 213: In what circumstance does this produce an error, and why is that a sign to keep going?

crates/sui-fork/src/owned_object_index.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// Apply local execution ownership changes to the owned-object index.
+    pub(crate) fn apply_owned_object_index_updates<'a>(
+        &self,
+        old_objects: impl IntoIterator<Item = &'a Object>,
+        new_objects: impl IntoIterator<Item = &'a Object>,
+    ) -> anyhow::Result<()> {
+        let mut batch = self.tables.objects.batch();
+
+        self.delete_owned_objects(&mut batch, old_objects)?;
+        self.insert_owned_objects(&mut batch, new_objects)?;
```
amnn, line 240: Does this approach work when mutating objects? I have a vague recollection that RocksDB does not like it when a write batch contains conflicting writes for the same key, which would happen if the object was mutated.

crates/sui-fork/src/owned_object_index.rs
```diff
@@ -0,0 +1,451 @@
...
+
+        self.mark_owned_object_index_initialized(&mut batch)?;
```
amnn: How come this is happening on every update rather than at the end of some predefined initialization point?

crates/sui-fork/src/owned_object_index.rs
```diff
@@ -0,0 +1,451 @@
...
+                metadata.version
+            ),
+            None => Ok(false),
+        }
+    }
+
+    /// Mark the owned-object index initialized in the same batch as the index rows.
+    ///
+    /// This marker is written for empty indexes too, so readers can tell an initialized-empty
+    /// index from one that still needs lazy seed initialization.
```
amnn, line 284: Ah I see -- seed initialization is lazy -- why is that?

crates/sui-fork/src/owned_object_index.rs
```diff
@@ -0,0 +1,451 @@
...
+                batch.insert_batch(&self.tables.objects, [(key, object.version())])?;
+            }
+        }
+        Ok(())
+    }
+}
+
+/// `DataStore`-level orchestration of the owned-object index: lazy seed initialization and the
+/// owner-listing reads that back the v2 RPC owned-object iterator.
+impl DataStore {
```
amnn: It's a bit confusing to see this here -- can it not just belong in the module that defines the data store?

Also, I'd like to understand the locking story a bit more -- my intuition is telling me that the locks are situated "too close" to the data at this point, and that a simpler architecture would be to say that the execution layer (simulacrum) enforces serial execution and checkpoint creation, so that everything down stream of it can trust that there's only one actor in the critical path at any time, then it's the service's job to queue requests for the executor.

amnn (commented):
Couple of high level questions:

- Why are seeds lazily initialized? It seems to be adding complexity  and I can't figure out the benefit at the moment.
- Similar for the locking story for the data store -- the fact that it needs locking suggests to me that something upstream of it is not doing a good job serializing access to it, and it will be difficult to maintain good locking discipline because writes to the store usually need to be atomic across multiple different updates (e.g. for a single checkpoint),

Then there were some lower level questions about the API exposed against RocksDB as well, if you could take a look at those, thanks!

crates/sui-fork/src/tests/owned_object_index.rs
stefan-mysten: That's reasonable, moved to using JSON snapshots.

crates/sui-fork/src/owned_object_index.rs
stefan-mysten: It's for tests only, added a `#[cfg(test])`

crates/sui-fork/src/owned_object_index.rs
stefan-mysten: The `safe_iter_with_bounds` can produce an error as it returns this type: https://github.com/MystenLabs/sui/blob/c28e47def5e8523ea1a81cbd96a57ad458d8c077/crates/typed-store/src/traits.rs?plain=1#L8
So it could be a deserialization error or some RocksDB error. At the end, we have a map and collect call, so it should catch the error regardless. This way we avoid returning partial results that might look like all data was collected OK.

crates/sui-fork/src/owned_object_index.rs
stefan-mysten: Great question, and I really do not know, so I had to look it up in multiple sources. From what I can tell, RocksDB is fine with having multiple operations on the same key in a write batch, and the last operation will "win". To be sure, we should probably ask @andll to confirm this is ok.

andll: I don't know what specific rocksdb behavior for this is out of my head, but I assume this is something claude or simple test can figure our. But if I were to guess I would be surprised if it was something other then 'last operation win'

crates/sui-fork/src/owned_object_index.rs
stefan-mysten: Good point, I don't think it needs to be, but we need to ensure that we initialized the index somewhere if it'd be removed. In this current implementation, the operation would be idempotent, but it's an extra operation.

crates/sui-fork/src/owned_object_index.rs
stefan-mysten: I recall we were talking about that and we went with being lazy for no other reason than to keep it lazy as is the object fetching.

It is possible that seed information is not used immediately and the index is not required except for tx execution or owned object listing. I have no strong opinion if we should run this at startup or lazily initialize it. Any preference on your end?

crates/sui-fork/src/owned_object_index.rs
stefan-mysten: > It's a bit confusing to see this here -- can it not just belong in the module that defines the data store?

That store module is getting quite big so was wondering if I could split it up. Perhaps I can keep all the code in the same file, but re-organize it a bit. I'll have a look.

stefan-mysten: Regarding the locking story, we should talk. Perhaps indeed there is a need to change something. 

The RpcReader has a cloned data store, so it can bypass locking that happens when we get access to simulacrum via context. So DataStore then needs some locking mechanism to coordinate read/write access to objects.

stefan-mysten:
> Couple of high level questions:
> 
>     * Why are seeds lazily initialized? It seems to be adding complexity  and I can't figure out the benefit at the moment.
> 
>     * Similar for the locking story for the data store -- the fact that it needs locking suggests to me that something upstream of it is not doing a good job serializing access to it, and it will be difficult to maintain good locking discipline because writes to the store usually need to be atomic across multiple different updates (e.g. for a single checkpoint),
> 
> 
> Then there were some lower level questions about the API exposed against RocksDB as well, if you could take a look at those, thanks!

Thanks for the review. I replied to the comments for the locking and RocksDB stuff, though we should probably chat offline as well.

crates/sui-fork/src/owned_object_index.rs
stefan-mysten: Thanks @andll!

crates/sui-fork/src/filesystem.rs
amnn: Right, you want to open it once and keep it open while the server is running.

amnn (commented):
@stefan-mysten, did you push all changes you want reviewed to this PR? It seems like you've replied to some comments to indicate they've been addressed, but then I don't see the related changes.

stefan-mysten:
> @stefan-mysten, did you push all changes you want reviewed to this PR? It seems like you've replied to some comments to indicate they've been addressed, but then I don't see the related changes.

🤦 yes, forgot to push. Just did that now.

crates/sui-fork/src/owned_object_index.rs
stefan-mysten: Moved the code into `DataStore` in `store.rs`.

PR #26709, opened 2026-05-19
tx offset

nickvikeras:
## Description

Adding the tx offset to the list api responses (populated from the index rather than the checkpoint to save a DB read).

### DIFF
--- a/crates/sui-kv-rpc/src/v2alpha/list_transactions.rs
+++ b/crates/sui-kv-rpc/src/v2alpha/list_transactions.rs
@@ -63,7 +63,7 @@ const CHUNK_MAX: usize = 100;
 
 pub(crate) type ListTransactionsStream =
     BoxStream<'static, Result<ListTransactionsResponse, RpcError>>;
-type TransactionWithObjectsStreamItem = (u64, TransactionData, ObjectMap);
+type TransactionWithObjectsStreamItem = (u64, u32, TransactionData, ObjectMap);
 
 pub(crate) async fn list_transactions(
     ctx: QueryContext,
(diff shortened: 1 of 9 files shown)


crates/sui-kv-rpc/src/v2alpha/list_transactions.rs
```diff
@@ -57,7 +57,7 @@ const CHUNK_MAX: usize = 100;
 
 pub(crate) type ListTransactionsStream =
     BoxStream<'static, Result<ListTransactionsResponse, RpcError>>;
-type TransactionWithObjectsStreamItem = (u64, TransactionData, ObjectMap);
+type TransactionWithObjectsStreamItem = (u64, u32, TransactionData, ObjectMap);
```
bmwill, line 66: Should we have the proto message also use a u32?

crates/sui-kvstore/src/tables/tx_seq_digest.rs
```diff
@@ -45,14 +53,19 @@ pub fn decode_key(key: &[u8]) -> Result<u64> {
...
+) -> [(&'static str, Bytes); 4] {
     [
         (col::DIGEST, Bytes::from(digest.inner().to_vec())),
         (
             col::EVENT_COUNT,
             Bytes::copy_from_slice(&event_count.to_be_bytes()),
         ),
+        (
+            col::TX_OFFSET,
+            Bytes::copy_from_slice(&tx_offset.to_be_bytes()),
```
bmwill: Ok maybe I'm micro optimizing here but should we use a leb encoding here to reduce the number of bytes we have to store here given the vast majority these are going to be small? Same goes for event counts and maybe even checkpoint numbers? (Although checkpoint numbers maybe less so)

bmwill: We don't need to fix this right now but we should maybe start cataloging a set of "to revisit" before we stabilize it

crates/sui-kv-rpc/src/v2alpha/list_transactions.rs
nickvikeras: I'll do this as a followup later before the branch cut after everything else is landed.

crates/sui-kvstore/src/tables/tx_seq_digest.rs
nickvikeras: Oh good catch, I used a varint for the checkpoint seq number. I think that's actually even better than leb for this right?

bmwill (approved):

crates/sui-kvstore/src/tables/tx_seq_digest.rs
bmwill: yeah varint/leb same-diff :D

PR #26710, opened 2026-05-20
nickv/misc kv index

nickvikeras:
## Description

Couple of misc things for the branch cut.

1. Increase event bucket size.
2. Add a "extant" dimension to support negative event queries
3. Bucket the tx_seq_digest table It looks like sequential keys give _way_ better read performance than random keys, and the bucketing should be good enough to spread the load.

### DIFF
--- a/crates/sui-inverted-index/src/dimensions.rs
+++ b/crates/sui-inverted-index/src/dimensions.rs
@@ -45,6 +45,14 @@ pub enum IndexDimension {
     /// any transaction that wrote to the stream; event-space matches the
     /// individual events committed to the stream.
     EventStreamHead = 0x07,
+    /// Internal existence marker (not user-queryable, not exposed via
+    /// `filter.rs`): one bit per real event, recording that an `event_seq`
+    /// position is occupied. Event-space only — it gives the read side a
+    /// concrete universe of real events to subtract from when evaluating
+    /// unanchored negation (`NOT D` == `E \ D`), since the packed `event_seq`
+    /// namespace is far too sparse to synthesize a dense complement. Carries no
+    /// value payload, so its encoded key is just the tag byte.
+    EventExtant = 0x08,
 }
 
 impl IndexDimension {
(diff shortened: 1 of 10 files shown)


crates/sui-inverted-index/src/dimensions.rs
```diff
@@ -45,6 +45,14 @@ pub enum IndexDimension {
     /// any transaction that wrote to the stream; event-space matches the
     /// individual events committed to the stream.
     EventStreamHead = 0x07,
+    /// Internal existence marker (not user-queryable, not exposed via
+    /// `filter.rs`): one bit per real event, recording that an `event_seq`
+    /// position is occupied. Event-space only — it gives the read side a
+    /// concrete universe of real events to subtract from when evaluating
+    /// unanchored negation (`NOT D` == `E \ D`), since the packed `event_seq`
+    /// namespace is far too sparse to synthesize a dense complement. Carries no
+    /// value payload, so its encoded key is just the tag byte.
```
bmwill, line 54: In addition to that it should fully solve the unfiltered event listing as well correct?

crates/sui-kv-rpc/src/v2alpha/list_events.rs
```diff
@@ -609,111 +601,70 @@ struct EventRef {
...
 }
 
-/// Point-lookup `tx_seq_digest` across the tx range covered by `event_range`,
+/// Range-scan `tx_seq_digest` across the tx range covered by `event_range`,
 /// using each row's `event_count` to enumerate real event_seqs per tx without
-/// touching the tx body. Wraps every emitted ref as `Watermarked::Item`; the
-/// unfiltered path doesn't emit watermarks because there's no
-/// bitmap-vs-render race — items come directly from the same row scan.
+/// touching the tx body.
 fn unfiltered_event_refs(
```
bmwill, line 506: Can we use the event extant bitmap for this instead?

crates/sui-kvstore/src/tables/tx_seq_digest.rs
```diff
@@ -21,10 +22,17 @@ use anyhow::bail;
...
 use sui_types::digests::TransactionDigest;
 use sui_types::messages_checkpoint::CheckpointSequenceNumber;
 
 pub const NAME: &str = "tx_seq_digest";
+pub const SCHEMA_VERSION: u32 = 1;
+/// Number of tx_sequence_numbers per forward-scan bucket. Tied to
+/// SCHEMA_VERSION; changing it requires a version bump and backfill.
+pub const FORWARD_BUCKET_SIZE: u64 = 1_000;
+const FORWARD_KEY_NUMBER_WIDTH: usize = 20;
+const FORWARD_BUCKET_PREFIX_WIDTH: usize = 16;
```
bmwill, line 35: Before we stabilize we may want to do a pass on the schema?

crates/sui-inverted-index/src/dimensions.rs
nickvikeras: Yeah I believe so.

crates/sui-kv-rpc/src/v2alpha/list_events.rs
nickvikeras: Yeah I had the same thought, I was gonna test and see if it was any faster.

crates/sui-kvstore/src/tables/tx_seq_digest.rs
nickvikeras: Yeah. I picked 1000 cuz it seemed small enough to not cause hot spots and big enough to satisfy all query patterns with at most one bucket hop per request.

bmwill (approved):

PR #26714, opened 2026-05-20
[graphql-stream] Resumable checkpoints subscription [14/n]

tpham-mysten:
## Description 

- Add `afterCursor` and `afterCheckpoint` to the checkpoints subscription so clients can resume from a known point. On subscribe, a LedgerService scan covers the gap to the live tip in parallel batches, then the live broadcast takes over.
- Gaps or Lagged events during the live phase re-enter the scan, bounded by `resume_max_recovery_attempts` consecutive retries before disconnecting the subscriber.

## Test plan 

  - cargo nextest run --features staging -p sui-indexer-alt-graphql --lib
  - Schema snapshot regenerated and checked in.
  - Manual: connect to a streaming server with afterCheckpoint set

---

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425
   - #26495
   - #26731
   - #26714

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Support resumable checkpoints subscription in features `staging`
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -7,38 +7,80 @@ use async_graphql::Context;
...
+    ///
+    /// Pass `afterCursor` (opaque) or `afterCheckpoint` (sequence number) to resume from a known point. The two arguments are mutually exclusive.
     ///
     /// This subscription is not yet available for use.
     async fn checkpoints(
         &self,
         ctx: &Context<'_>,
-    ) -> Result<impl futures::Stream<Item = Result<Checkpoint, RpcError>>, RpcError> {
+        after_cursor: Option<CCheckpoint>,
+        after_checkpoint: Option<UInt53>,
```
amnn: - Can we call `afterCursor` just `after` to match the pagination API?
- Can we support both and just take the max (again to match the behaviour during pagination)?

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -7,38 +7,80 @@ use async_graphql::Context;
...
+        }
+        let resume_from: Option<u64> = after_cursor
+            .map(|c| *c)
+            .or_else(|| after_checkpoint.map(u64::from));
+
         let package_store = ctx.data::<Arc<StreamingPackageStore>>()?.clone();
         let limits: &Limits = ctx.data()?;
         let resolver_limits = limits.package_resolver();
-        let mut receiver = ctx.data::<CheckpointBroadcaster>()?.resubscribe();
+        let config = ctx.data::<SubscriptionConfig>()?;
```
amnn:
```suggestion
        let config: &SubscriptionConfig = ctx.data()?;
```

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -7,38 +7,80 @@ use async_graphql::Context;
...
+        let resume_batch_size = config.resume_batch_size;
+        let resume_transition_threshold = config.resume_transition_threshold;
+        let resume_max_recovery_attempts = config.resume_max_recovery_attempts;
+        let resources = ctx.data::<SubscriptionResources>()?;
```
amnn: ditto (and in other places)

amnn: It's also nice to group all the `ctx.data` calls together near the top so it's easy to see which context is required at a glance.

crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -444,7 +444,10 @@ pub async fn start_rpc(
         readiness,
     )) = streaming_setup
     {
-        rpc = rpc.data(stream_task.broadcaster()).data(streaming_packages);
+        rpc = rpc
+            .data(stream_task.subscription_resources())
+            .data(streaming_packages)
+            .data(config.subscription.clone());
```
amnn: Why `clone` the subscription config? In other cases above we seem to be able to avoid doing that.

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -241,6 +242,19 @@ pub struct SubscriptionConfig {
...
+    pub resume_batch_size: usize,
+
+    /// Once a resume scan is within this many checkpoints of the live tip, the scan stops
+    /// and the live broadcast takes over.
+    pub resume_transition_threshold: u64,
+
+    /// Maximum consecutive recovery attempts (re-entering the catch-up scan after a Lagged
+    /// event or detecting a gap on the broadcast) before a subscription is disconnected.
+    /// Resets on any successful yield.
+    pub resume_max_recovery_attempts: u32,
```
amnn: Is the feature this configures strictly necessary, as opposed to just dropping the connection if a subscriber that was previously keeping up lags?

I thought when we discussed this feature, we said that it's better not to do it because by making lagging clients reconnect, we get another chance to redistribute load and more control over how many downstream DB reads a subscriber can cause.

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -241,6 +242,19 @@ pub struct SubscriptionConfig {
...
     /// Number of checkpoints fetched concurrently per chunk during upstream gap recovery.
     pub gap_recovery_chunk_size: usize,
+
+    /// Maximum number of checkpoints fetched per LedgerService batch during a resumable
+    /// subscription's catch-up scan.
+    pub resume_batch_size: usize,
+
+    /// Once a resume scan is within this many checkpoints of the live tip, the scan stops
+    /// and the live broadcast takes over.
+    pub resume_transition_threshold: u64,
```
amnn: Is it worth offering a config for this that's separate from `broadcast_buffer`, or should it be derved from `broadcast_buffer` (e.g. once the gap is half the broadcast buffer size, cut over).

Offering it as a separate config runs the risk of someone misconfiguring.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -199,8 +213,12 @@ impl CheckpointStreamTask {
         }
     }
 
-    pub(crate) fn broadcaster(&self) -> CheckpointBroadcaster {
-        self.broadcaster.resubscribe()
+    pub(crate) fn subscription_resources(&self) -> SubscriptionResources {
+        SubscriptionResources {
+            broadcaster: self.broadcaster.resubscribe(),
+            network_tip: self.network_tip.clone(),
+            ledger_grpc_reader: self.ledger_grpc_reader.clone(),
```
amnn: It feels a little off to package up `ledger_grpc_reader` inside `SubscriptionResources` to go in the `Context` -- could we add it directly to the `Context`, like we've done for, e.g. `consistent_reader`?

Then this type can be responsible for just the queue of checkpoints to subscribe to, so you could call it `SubscriptionQueue`

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -7,38 +7,80 @@ use async_graphql::Context;
...
+        let mut receiver = resources.broadcaster.resubscribe();
+        let fetcher = Arc::new(resources.ledger_grpc_reader.clone());
```
amnn: This seems like an anti-pattern -- either `ledger_grpc_reader` is itself cheap enough to clone, in which case, just clone it, or it is not in which case, it should have been wrapped in an `Arc` at an earlier stage.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
amnn: You might also consider encapsulating the logic for generating the checkpoint stream from a `SubscriptionQueue` into `SubscriptionQueue` itself.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
```diff
@@ -0,0 +1,155 @@
...
+/// tip; the caller then takes over by consuming the live broadcast (the bridge between scan
+/// end and the first live item is covered by the caller subscribing before this scan starts).
+pub(crate) fn scan_checkpoints<F: CheckpointFetcher>(
+    fetcher: Arc<F>,
+    network_tip: Arc<AtomicU64>,
+    start_after: u64,
+    max_batch_size: usize,
+    transition_threshold: u64,
+) -> impl Stream<Item = Result<Arc<ProcessedCheckpoint>, RpcError>> {
+    stream! {
```
amnn: Should we apply a rate limit to this request (i.e. only allow N checkpoints per second to come through, even if we can download faster than that) to prevent a single client with a large backfill dominating the resources? `tokio-stream` has a [`throttle` adapter](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.throttle) for this kind of thing -- you set-up a stream of batches, flatten and then throttle it.

amnn: If we're doing that, it may not even be worth explicitly batching checkpoints fetched. Instead, you can use the same `try_for_each_spawned` and adaptive concurrency mechanism to allow the upstream source to dynamically tune concurrency based on how quickly the subscriber is consuming checkpoints.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -324,6 +342,8 @@ impl CheckpointStreamTask {
             let _ = self.package_eviction_tx.send((seq, ids));
         }
         let processed = process_checkpoint(checkpoint)?;
+        self.network_tip
```
amnn: There is technically a tiny race here (because this value is indirectly counting the number of checkpoints sent through `sender` but we are updating it non-atomically).

It's unlikely to be an issue in practice, but it also doesn't need to exist at all because `broadcast::Receiver` tracks its own length in an atomic way, so as long as you maintain a `Receiver` that you never directly receive from, and you know the checkpoint you started broadcasting at, you can always calculate this value without a race and without introducing another shared piece of state.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
```diff
@@ -0,0 +1,155 @@
...
+    let gap = tip_checkpoint.saturating_sub(last_scanned_checkpoint);
+    if gap <= transition_threshold {
+        return None;
+    }
+    let lo = last_scanned_checkpoint + 1;
+    let hi = lo
+        .saturating_add(max_batch_size as u64)
+        .saturating_sub(1)
+        .min(tip_checkpoint);
+    Some(lo..=hi)
```
amnn: It's better to return a half open interval, so that you can represent an empty range here and also bound `hi` from below by `lo`.

I believe certain operations panic if they are provided a range where the upper and lower bounds are inverted, which seems like it could happen if `last_scanned_checkpoint + 1` crosses over `tip_checkpoint` (under some configurations).

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/lib.rs
tpham-mysten: Yes we can avoid that by just borrowing subscription above

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Sounds good. I moved ledger_grpc_reader to the context and refactor `SubscriptionResources` to `SubscriptionBroascaster`, which now includes all the logic to do scanning and subscribe to the tip

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
tpham-mysten: Ack. This logic is no longer there after I use `tokio_stream::StreamExt::throttle` with Stream

crates/sui-indexer-alt-graphql/src/config.rs
tpham-mysten: Agreed! I have removed this and just use half the broadcast buffer size.

crates/sui-indexer-alt-graphql/src/config.rs
tpham-mysten: The retry was added for the scanning case as a delay in LedgerService during the scan-to-live handoff can cause a gap, which is more a server error than a client error. For Lag, that is a client issue and retry probably does not help (the client is already far behind). So I kept the retry only in the resume scan logic and dropped it from the Lag path.

WDYT?

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: I looked into the refactor. The catch is we'd still need a shared variable for the first broadcast sequence (the Receiver gives us a count via len() but not absolute sequence numbers), so the total amount of shared state ends up similar to the atomic today, just with an extra OnceLock<u64>. The logic is also a bit more complex.

The race would be resolved by that refactor, but I agreed it's unlikely to be a problem in practice. Now that resume_transition_threshold is derived as broadcast_buffer / 2, even if the atomic is briefly ahead of the broadcast by one item, the scan still stops well within the buffer's range and the handoff stays smooth.

My preference is to keep things as-is here, but let me know if I'm missing anything.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
tpham-mysten: > to prevent a single client with a large backfill dominating the resources?

Do you mean network bandwidth? I think it would not be a big problem in a subscription-only instance (which is the suggested setup for production), as only resume subscribers call `LedgerService`. With multiple resume subscribers, tokio's scheduler gives each task fair CPU time, so the kv-rpc fetches end up roughly evenly distributed across subscribers.

That said, adding throttling is still useful for predictability. Some RPC providers may prefer co-locating the Query and Subscription APIs on the same instance, and even a forward-only subscription instance still hits the DB sometimes (e.g., for packages). I've added `per_subscriber_scan_max_qps` and pass it into the throttle adapter you mentioned.

>  you can use the same try_for_each_spawned and adaptive concurrency mechanism to allow the upstream source to dynamically tune concurrency based on how quickly the subscriber is consuming checkpoints.

For this, I'm not sure how useful it is in our setup. We already get natural backpressure all the way from the subscriber back to the kv-rpc fetcher: when the GraphQL client over SSE is slow to receive, the yield in the stream blocks, which stops the throttle from releasing items, which stops new fetches from being initiated. So in-flight fetches are already paced to consumer rate without needing adaptive concurrency.

So I went with buffered(N) + throttle, with N exposed via per_subscriber_scan_max_concurrent_fetches

WDYT?

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Ack. but this is obsolete now as ledger_grpc_reader is moved to the context

tpham-mysten (commented):
Thanks @amnn! 

I have applied the implemented the throttle approach for the scanning, but I have some thoughts on whether adaptive concurrency is necessary here.

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -4,59 +4,69 @@
...
+        let package_store: &Arc<StreamingPackageStore> = ctx.data()?;
         let limits: &Limits = ctx.data()?;
+        let config: &SubscriptionConfig = ctx.data()?;
+        let broadcast: &SubscriptionBroadcast = ctx.data()?;
+        let fetcher: &LedgerGrpcReader = ctx.data()?;
+
+        let resume_from: Option<u64> = match (after.map(|c| *c), after_checkpoint.map(u64::from)) {
+            (Some(a), Some(b)) => Some(a.max(b)),
+            (a, b) => a.or(b),
+        };
```
amnn, line 61: nit (optional): there's a helper that we use for other transaction filters that might be useful here, in the `intersect` module:

```suggestion
        let resume_from = intersect::field(
            after.map(|c| *c),
            after_checkpoint.map(u64::from),
            intersect::by_max
        );
```

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
```diff
@@ -0,0 +1,86 @@
...
+                None
+            } else {
+                let next = last + 1;
+                Some((next, (next, tip_handle)))
+            }
+        },
+    );
+
+    let mask = checkpoint_field_mask();
+    tokio_stream::StreamExt::throttle(
```
amnn, line 71: Is it possible to use `throttle` in receiver position if you import both `tokio_stream::StreamExt` and `futures::stream::StreamExt` anonymously

```
use futures::stream::StreamExt as _;
use tokio_stream::StreamExt as _;
```

?

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
```diff
@@ -0,0 +1,86 @@
...
+            .map(move |seq| {
+                let fetcher = fetcher.clone();
+                let mask = mask.clone();
+                async move {
+                    let proto = fetch_one_with_retry(fetcher.as_ref(), &mask, seq).await?;
+                    let processed = process_checkpoint(proto)?;
+                    Ok::<_, RpcError>(Arc::new(processed))
+                }
+            })
+            .buffered(concurrency),
```
amnn, line 82: Note that the `buffered` stream modifier does something slightly different to `for_each_spawned` et al -- it will schedule the io concurrently (the fetching to KV store), but the checkpoint processing after that will always be handled serially on one thread. Maybe that's not a problem here, but wanted to flag the difference.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
amnn: > The catch is we'd still need a shared variable for the first broadcast sequence (the Receiver gives us a count via len() but not absolute sequence numbers), so the total amount of shared state ends up similar to the atomic today, just with an extra OnceLock.

I was assuming that the "first broadcast sequence number" could be fetched once initially, and then stashed in `SubscriptionBroadcast`, so while it is shared state, it's not shared **mutable** state, which makes it strictly better. If that's not possible, and you still need a `OnceLock` to initialize the first checkpoint, then yes, I agree, that's not worth doing.

crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -437,21 +443,22 @@ pub async fn start_rpc(
...
-    } else {
-        None
-    };
+    let streaming_handles =
+        if let Some((stream_task, eviction_task, streaming_packages, readiness, ledger_grpc)) =
+            streaming_setup
+        {
+            rpc = rpc
+                .data(stream_task.subscription_broadcast())
+                .data(ledger_grpc)
```
amnn: nit: I think we should just add this unconditionally with the other stores as it's not specific to subscriptions.

amnn (commented):
Sending half a review for you to look at -- I got called away by all the excitement this week, sorry about that.

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Ack. I tried it but prefer the original a bit more as intersect::field returns Option<Option<T>> and we need to add a flatten() at the end, and the original approach is also clear enough. I will keep it as-is then

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
tpham-mysten: Tried it. `.throttle` resolves fine (only tokio_stream defines it), but .map is on both traits so the compiler can't tell which one we mean, and each .map call would need fully-qualified syntax like futures::StreamExt::map(stream, ...). So I will just keep it as-is

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
tpham-mysten: Ack. I have added a comment to note this

crates/sui-indexer-alt-graphql/src/lib.rs
tpham-mysten: Sure done!

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Yeah actually it's possible. We actually already recorded the first checkpoint from Subscription Readiness, and I have just passed it from SubscriptionReadiness to SubscriptionBroadcast

tpham-mysten (commented):
Sorry for the late resolve and thanks for the review, @amnn!

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
amnn: Hmm, unfortunate -- is it possible to use just one of them though?

crates/sui-indexer-alt-graphql/src/api/subscription.rs
```diff
@@ -4,59 +4,71 @@
...
-                        yield Err(broadcast_error(e));
-                        break;
-                    }
+        let stream = broadcast
+            .clone()
+            .subscribe(resume_from, fetcher.clone(), config);
+
+        Ok(stream.map(move |item| {
+            item.map(|processed| {
+                let scope = Scope::for_streamed_checkpoint(
```
amnn, line 72: The fact that this is the same for all checkpoint streams now makes me wonder whether it's worth lifting this into the stream we are subscribing to rather than redoing it for every stream?

Not a big deal as it's not a huge amount of work, but just an observation.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -163,12 +167,147 @@ pub(super) fn checkpoint_field_mask() -> FieldMask {
...
+        fetcher: LedgerGrpcReader,
+        config: &SubscriptionConfig,
+    ) -> impl Stream<Item = Result<Arc<ProcessedCheckpoint>, RpcError>> + 'static {
+        // Bounded re-entries into Phase 1 to bridge the scan-to-live transition gap (when the
+        // network tip advances between scan end and Phase 2's first broadcast item). Two
+        // leaves headroom since scan throughput dominates live rate. `Lagged` is not retried.
+        const MAX_GAP_RESCANS: u32 = 2;
+
+        let mut receiver = self.broadcaster.resubscribe();
+        let fetcher = Arc::new(fetcher);
```
amnn: Is the `Arc` necessary here? `LedgerGrpcReader` should either already be cheap to clone (e.g. have an inner `Arc`) or come pre-packaged in an `Arc`.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -248,7 +387,7 @@ impl CheckpointStreamTask {
                 .await?;
                 info!("Connected to checkpoint stream at {}", self.uri);
 
-                self.consume_stream(stream, &mut last_broadcast, &mut first_recorded)
+                self.consume_stream(stream, &mut last_broadcast, &mut first_live_recorded)
                     .await?;
                 warn!("Checkpoint stream ended, reconnecting");
```
amnn, line 418: There's an issue with reconnecting to the checkpoint stream, but re-using the same broadcaster and previously recorded `first_live_checkpoint`, if there's a gap (we missed some checkpoints in the gap between disconnecting and reconnecting), because we count network tip as `first_live_checkpoint + broadcaster.len() - 1` -- we will be off by the number of checkpoints we missed.

The safest and simplest thing to do would be to refresh everything if the stream fails:

- Close the old broadcaster, which will cause downstream clients to be kicked off. When they reconnect, they will re-enter phase 1 (backfill/scanning).
- Establish a new upstream connection, peek it to discover its first live checkpoint, and package that up in a new `CheckpointBroadcaster`/`SubscriptionBroadcast`, and update the context with that -- then start consuming the stream.

That also suggests that a `CheckpointBroadcaster`/`SubscriptionBroadcast` needs another level of indirection -- the actual broadcaster and first checkpoint held under a `RwLock`, which each new subscription reads from to get a receiver. The stream task sets this when it has a stream to consume.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -163,12 +167,147 @@ pub(super) fn checkpoint_field_mask() -> FieldMask {
...
+                        start_after,
+                        &config,
+                    ) {
+                        let processed = item?;
+                        last_yielded = Some(processed.summary.sequence_number);
+                        yield Ok(processed);
+                    }
+                }
+
+                // Phase 2: Transition to live.
```
amnn: Reasoning about the hand-off between phase 1 and phase 2, something feels strange/off to me:

- We stop phase 1 when we are within half the receiver queue's capacity away from the tip.
- But then when we start phase 2, we use a `receiver` that was resubscribed before phase 1 started, which means that 
  - if we are backfilling for a long time, it's likely that we'll blow the queue and start phase 2 already lagging (and not retry).
  - if we misconfigure and the queue capacity is too long, then we will end up having to retry a potentially unbounded number of times get the backfill to meet the point where we resubscribed.

What I had expected/wanted was different: 

- While we `scan_checkpoints` we maintain a receiver that is half its capacity away from the tip.
- We exit `scan_checkpoints` when it touches exactly the next checkpoint from the receiver that we are maintaining.
- Then when we start Phase 2, it's exactly at the right spot, and there is enough time for some extra checkpoints to accumulate in the queue between the hand-off between Phase 1 and Phase 2.

This means we don't need retries for the gap, and we don't have the issue with lagging (we should add tests for both these cases).

amnn (commented):
Couple of tricky edge cases when we reconnect to the source checkpoint stream and when we hand-over between backfill and tip-of-chain.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_resume.rs
tpham-mysten: Checking with Claude and they mentioned we can drop throttle by .then(sleep):

```
seq_stream
      .map(move |seq| { ... })
      .buffered(concurrency)
      .then(move |item| async move {
          tokio::time::sleep(throttle_interval).await;
          item
      })
```

Behavior: same emission rate cap as throttle. Subtle difference is that when items arrive slower than the cap, throttle passes them through immediately while .then(sleep) still adds the delay. I personally don't like this approach, and I think the existing `tokio_stream::StreamExt::throttle` should be fine

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Thanks for the catch! The eager-subscribe + threshold combo is the root of both bugs.

For the drain approach you suggested. The concern there is that drained broadcast items are double work: scan already fetches them from kv-rpc, so we'd pay for both and discard the broadcast copy. Also pulls receiver-position math into the scan loop.


I went with a slight twist instead: drop the threshold so scan goes all the way to the tip, and resubscribe partway through Phase 1 when tip - scan_seq <= N (N = buffered concurrency). The receiver then accumulates live items in parallel with the remaining drain, and Phase 2 picks up at the right spot — first few recvs get deduped, then it's just live.

WDYT?

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Walked through the reconnect path, and I think network_tip actually stays accurate.

When the upstream errors, we loop back and reconnect with last_broadcast preserved. The first new checkpoint's seq is checked against last_broadcast + 1; if there's a gap, `recover_gap` broadcasts each missed checkpoint via `self.sender` before the new live one. Each send increments `broadcaster.len()`, so first_live_checkpoint + broadcaster.len() - 1 ends up reflecting the actual tip including the recovered range.
 
Am I missing anything?

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/api/subscription.rs
tpham-mysten: Ack. Agreed it's not a huge of work, and we can refactor later if it became a concern

tpham-mysten (commented):
Thanks @amnn!

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Thinking a bit more while doing the transaction subscription, I think we can improve this further. We no longer scan to the tip. Once the scan gets within half the buffer of the tip, we resubscribe and pin handoff = the current tip, and we stop the scan exactly at handoff. It's basically the "exit when the scan touches the receiver's next checkpoint" you described, with the handoff pinned so we know exactly where that point is.

This fixes both bugs:
  - No unbounded retries: the scan targets a fixed pinned point, not the moving tip, so it always reaches the meet point.
  - No starting Phase 2 already lagging: we resubscribe late, so the receiver is young at handoff, and the other half of the buffer absorbs whatever streams in during the handoff.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
amnn: Ah got it -- no it's me that missed the fact that there is gap recovery logic in this path. As there's no correctness issue here, the following is not a blocker, will leave it with you to consider as you flesh out the other subscription roots, but if we do continue to attempt recovery when subscriptions are dropped then let's add a regression test for that.

I do think there's still a simplicity argument for the structure I proposed previously:

- Start the stream, make it peekable.
- Peek it to find the first checkpoint, make a note of that.
- Make the broadcaster available for new subscribers/clients.
- Keep pushing checkpoints into the subscriber until you hit a failure, at which point, drop everything (which also will drop downstream clients).

It should be possible to implement this lifecycle using straight-line logic (e.g. we don't need to share a slot that will hold the first recorded checkpoint and fill it later; don't have multiple code paths that scan historical checkpoints). 

This approach does mean that downstream clients are not protected from gRPC dropping GraphQL as a subscriber -- they will also be dropped and will have to reconnect, but that may not be a terrible thing:

- If the likely reason for this is high load, it may allow pressure to ease off the system as a whole faster.
- As written, we only recover the gap once a new subscription has been established. This means clients will be left hanging until GraphQL is available to reconnect to gRPC. They won't be able to tell whether there is no data for them, or there is an upstream issue, which means they don't get an opportunity to e.g. try another subscription source.

We should also make sure that the `scan_checkpoints` path utilises a data loader so that when clients reconnect and need common sets of checkpoints, we don't issue redundant fetches for the same checkpoint (regardless of whether the simplification is adopted or not, clients are likely to line up like this when a service with multiple subscriptions spins up) -- this may already happening, I don't fully recall.

crates/sui-indexer-alt-graphql/src/task/streaming/test_utils.rs
amnn, line 1: nit: file order, `super` imports.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -163,12 +168,166 @@ pub(super) fn checkpoint_field_mask() -> FieldMask {
...
+
+                    if handoff.is_some_and(|h| seq >= h) {
+                        break;
+                    }
+                }
+            }
+
+            // Phase 2: follow live from the pinned receiver, or a fresh one if Phase 1 was skipped.
+            let mut receiver = pending_receiver
+                .unwrap_or_else(|| self.broadcaster.resubscribe());
```
amnn: Does this slightly simplify things? (diff is not super clear: I'm combining `pending_receiver` and `receiver` into one non-optional value).

```suggestion
            let mut receiver = self.broadcaster.resubscribe();
            // `resubscribe` is future-only (delivers `handoff + 1`), so Phase 1 stops exactly at
            // `handoff` rather than chasing the tip.
            let mut handoff: Option<u64> = None;

            // Phase 1: scan toward the tip; within `handoff_threshold`, resubscribe + pin, stop at it.
            if let Some(start_after) = last_yielded {
                for await item in scan_checkpoints(fetcher, self.clone(), start_after, &config) {
                    let processed = item?;
                    let seq = processed.summary.sequence_number;
                    last_yielded = Some(seq);
                    yield Ok(processed);

                    if handoff.is_none()
                        && self.network_tip().saturating_sub(seq) <= handoff_threshold
                    {
                        receiver = self.broadcaster.resubscribe();
                        handoff = Some(self.network_tip());
                    }

                    if handoff.is_some_and(|h| seq >= h) {
                        break;
                    }
                }
            }

            // Phase 2: follow live from the pinned receiver, or a fresh one if Phase 1 was skipped.
```

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -163,12 +168,166 @@ pub(super) fn checkpoint_field_mask() -> FieldMask {
...
+        }
+    }
+}
+
+/// General error for a subscription interrupted by a gap or catch-up overflow. The specific reason
+/// is logged at the call site; clients only see that they should reconnect and backfill.
+fn reconnect_error() -> RpcError {
+    anyhow::anyhow!(
+        "Subscription interrupted. Please reconnect and use the query API to backfill \
+         from your last seen sequenceNumber."
```
amnn: Do they need to use the query API? Can't they use resumption? (same question for the error message above).

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
```diff
@@ -163,12 +168,166 @@ pub(super) fn checkpoint_field_mask() -> FieldMask {
...
+                    yield Ok(processed);
+
+                    if pending_receiver.is_none()
+                        && self.network_tip().saturating_sub(seq) <= handoff_threshold
+                    {
+                        pending_receiver = Some(self.broadcaster.resubscribe());
+                        handoff = Some(self.network_tip());
+                    }
+
+                    if handoff.is_some_and(|h| seq >= h) {
```
amnn, line 256: Could we add an extra condition here to check the number of pending checkpoints in the pending receiver? If there are too many (if the buffer is close to full), we can drop the handoff and try again, to further reduce the likelihood we see a Lagged error.

amnn (approved):
Sorry for the long delay on reviewing this @tpham-mysten ! This is looking good now, although I've left some more comments on potential further opportunities to harden some parts and simplify some other parts.

crates/sui-indexer-alt-graphql/src/task/streaming/test_utils.rs
tpham-mysten: Sg. done

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Ah right! I have fixed it to Resumption

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Sg. done

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Yes, we can, but it might contradict the delivery-rate throttle we're planning for the follow-up PR (rate-limiting each subscriber by query cost).

Under that throttle, a heavy subscription consumes slower than chain by design, so its receiver is supposed to fill and eventually lag. That lag is the intended backpressure: the client reconnects and we get a chance to redistribute load. But a buffer-fullness re-pin sees the same "receiver filling" and re-pins to avoid the lag, so instead of lag -> reconnect the subscription would re-pin forever and never converge.

We could tell a transient scan stall apart from deliberate throttling by looking at the consume rate, but I'm not sure it's worth the complexity. I'd leave it as-is for now and add it later if metrics show frequent handoff lags.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
tpham-mysten: Thanks, this is a great framing.

Regarding regression test: that one's already covered. `test_subscription_recovers_from_upstream_disconnect`.

For Recovery vs drop-everything: I was checking all points, and still slighlty leaning towards Recovery:
- The per-subscriber delivery-rate throttle we're planning (rate-limiting by query cost) is the proper load-shedding mechanism, and it partially covers the load-shedding motivation for drop-everything.
- `recover_gap` lives in the source-level `CheckpointStreamTask`, so it's a single shared mechanism that tx/events inherit through the same broadcast, not per root complexity.
- It's also consistent with the indexing framework's gap-recovery pattern, which keeps the mental model uniform.

I do see the simplicity pros (no fill-later slot, one historical-scan path), so if the recovery machinery starts to have problem as we add more methods, we can revisit drop-everything then.

crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs
amnn: Agreed that we shouldn't try and look at consumption rate to tell whether the throttling is artificial (by us) or not. But your counter-example raises an interesting point, because I think it's actually an argument **for** adding the condition to drop the handoff 😅:

- If we don't drop the handoff when we see it slip today, we can still get into a scenario where we're scanning forever, if the scan starts far back enough that it never catches up. In fact if you are backfilling, and throttled below checkpoint rate, then you will never be closer to the tip than when you started.
- The scenario where the condition would be helpful is if you have a subscription that **is** converging, but is doing so slowly, such that it would have taken too long to reach tip after reaching hand-off.

Your argument suggests a couple of different options though: Either we disallow queries that we would have throttled to below the checkpoint rate **or** we have to allow queries to be demoted into scanning and trust that throttling will protect the underlying stores.

I slightly prefer the first option because it is simpler to implement and leaves our ability to shed load under strain open, like you mentioned (the second approach basically commits us to serving every client that successfully connects).

tpham-mysten: > If we don't drop the handoff when we see it slip today, we can still get into a scenario where we're scanning forever,

Yeah it can be a problem with the new throttling logic, and for 2 options, I agree that Option 1 is better!

PR #26717, opened 2026-05-20
Fullnode query apis

nickvikeras:
## Description

fullnode impl of the query apis


crates/sui-core/Cargo.toml
```diff
@@ -45,6 +45,7 @@ rand.workspace = true
 roaring.workspace = true
 rayon.workspace = true
 reqwest.workspace = true
+rustc-hash.workspace = true
```
bmwill, line 48: Do we need to do a similar trick with sharding the indexing pipeline given we don't need to hold bitmaps in memory but rather are just doing Merge op writes?

crates/sui-config/src/rpc_config.rs
```diff
@@ -64,6 +64,16 @@ pub struct RpcConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub authenticated_events_indexing: Option<bool>,
 
+    /// Enable historical checkpoint/transaction indexes for RPC queries.
+    ///
+    /// This flag is folded into the `rpc-index` DB version, so flipping it in
+    /// either direction triggers a full rebuild. While it stays put, forward
+    /// indexing and pruning maintain these indexes normally.
```
bmwill: As discussed we'll want this to be in its own "feature flag" table most likely.

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2574,7 +2574,7 @@ impl AuthorityPerEpochStore {
         // is_consensus_message_processed
         #[cfg(debug_assertions)]
         {
-            let mut seen = HashSet::new();
+            let mut seen = std::collections::HashSet::new();
```
bmwill: The change to this file seems un necessary

crates/sui-inverted-index/src/dimensions.rs
```diff
@@ -3,12 +3,15 @@
 
 use sui_types::accumulator_root::stream_id_from_accumulator_event;
 use sui_types::balance::Balance;
+use sui_types::base_types::SuiAddress;
 use sui_types::effects::AccumulatorValue;
+use sui_types::effects::TransactionEffects;
 use sui_types::effects::TransactionEffectsAPI;
-use sui_types::full_checkpoint_content::ExecutedTransaction;
```
bmwill, line 8: I think that the current rpc indexing pipeline uses the old `CheckpointData` structure vs the new `Checkpoint` one (which houses `ExecutedTransaction`). I think, if its not too much trouble, that we should probably make that change now to plumb through `Checkpoint` and not use `CheckpointData`. That will eliminate you needing to change this code here.

bmwill: Eg all we have to do is change one the one line to not do `.into()` here: https://github.com/mystenlabs/sui/blob/632e55c744d43aa9657c6636690e0eff5b44cf26/crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs#L716-L729

crates/sui-core/src/rpc_index.rs
```diff
@@ -211,15 +280,139 @@ pub struct PackageVersionInfo {
...
 
+/// Row of the `tx_seq_digest` table — direct mapping from `tx_sequence_number`
+/// to `(digest, event_count, checkpoint_number)`. `event_count` lets event
+/// listings enumerate a transaction's event_seqs without rereading the tx row.
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Debug)]
+pub struct TxSeqDigestInfo {
+    pub digest: TransactionDigest,
+    pub event_count: u32,
+    pub checkpoint_number: CheckpointSequenceNumber,
+}
```
bmwill, line 314: offset?

crates/sui-core/src/rpc_index.rs
```diff
@@ -295,6 +488,128 @@ fn balance_table_options() -> typed_store::rocks::DBOptions {
...
+fn decode_bitmap_blob(bcs_bytes: &[u8]) -> Result<RoaringBitmap, anyhow::Error> {
+    let blob: BitmapBlob = bcs::from_bytes(bcs_bytes)?;
+    Ok(RoaringBitmap::deserialize_from(&blob.0[..])?)
+}
+
+fn encode_bitmap_blob(bm: &RoaringBitmap) -> Vec<u8> {
+    let mut buf = Vec::with_capacity(bm.serialized_size());
+    bm.serialize_into(&mut buf)
+        .expect("RoaringBitmap::serialize_into on Vec cannot fail");
+    bcs::to_bytes(&BitmapBlob(buf)).expect("BCS encode of BitmapBlob cannot fail")
```
bmwill, line 526: this is so dumb lol, something we can optimize with a new db wrapper later

crates/sui-core/src/rpc_index.rs
```diff
@@ -546,6 +897,16 @@ impl IndexStoreTables {
...
         }
 
+        if rpc_config.ledger_history_indexing()
+            && let Some(checkpoint_range) = checkpoint_range
+        {
+            self.backfill_ledger_history_indexes(
+                authority_store,
+                checkpoint_store,
+                checkpoint_range,
+            )?;
```
bmwill, line 1007: So for right now are we punting on optimizing and backfilling in the background? So on startup we'll pay the cost of index initialization

crates/sui-core/src/rpc_index.rs
```diff
@@ -627,20 +988,117 @@ impl IndexStoreTables {
...
+            None => Ok(None),
+        }
+    }
+
+    /// Prune data from this Index. `pruned_tx_seq_exclusive` is the
+    /// absolute tx-seq floor after this prune — the caller derives it from
+    /// the last-pruned checkpoint's `network_total_transactions`. Returns
+    /// that floor iff ledger history maintenance ran and advanced it, so the caller
+    /// can update the compaction-filter atomic after the batch commits and
+    /// the atomic never leads disk.
```
bmwill: why don't we just update the atomic in this function directly or rather unconditionally in the caller vs needing to do this dance in this function?

crates/sui-core/src/rpc_index.rs
```diff
@@ -866,6 +1327,140 @@ impl IndexStoreTables {
...
+            StorageError::custom(format!(
+                "checkpoint {cp_seq}: network_total_transactions ({net_total}) \
+                 < tx_count ({tx_count})"
+            ))
+        })?;
+
+        // Build one ObjectSet covering all txs in the cp so the dimension
+        // extractor's object_set.get(ObjectKey) lookups work without per-tx
+        // allocation. Costs O(input_objects + output_objects) clones.
+        let mut object_set = ObjectSet::default();
```
bmwill, line 1464: if we make use of the `Checkpoint` type instead of `CheckpointData` we should already have access to an ObjectSet.

crates/sui-core/src/rpc_index.rs
nickvikeras: it only takes like 15 minutes on my laptop with 2 weeks of history. I can look at moving it to the background before we GA it.

bmwill: sg

crates/sui-core/Cargo.toml
nickvikeras: Doing something like that during the initial backfill might help, but thats not what this is for. We still do a lot of hashing to dedupe dimensions during the dimension-extraction phase of indexing. Although to be honest right now compute isn't a bottleneck.

crates/sui-inverted-index/src/dimensions.rs
nickvikeras: claude refused to do this and said it was a big intrusive refactor, but I will look at it more closely tomorrow.

bmwill: I mean that's fine, we can do it in a separate or later too

crates/sui-core/src/rpc_index.rs
nickvikeras: cleaned this up

crates/sui-core/src/rpc_index.rs
```diff
@@ -92,6 +149,15 @@ struct MetadataInfo {
...
 }
 
+/// Per-DB rpc-index settings, persisted in their own column family so the
+/// schema `version` stays a clean monotonic number rather than carrying packed
+/// feature bits.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
+struct IndexSettings {
+    /// Whether this DB was built with ledger-history indexing enabled.
+    ledger_history_indexing: bool,
+}
```
bmwill, line 168: how do you want to handle the fact that adding a flag is a breaking change to the schema?

bmwill (approved):

crates/sui-e2e-tests/tests/rpc/v2/ledger_service/list_ledger_history.rs
```diff
@@ -0,0 +1,2345 @@
...
+
+async fn new_cluster() -> TestCluster {
+    TestClusterBuilder::new()
+        .with_num_validators(1)
+        .disable_fullnode_pruning()
+        .with_rpc_config(sui_config::RpcConfig {
+            enable_indexing: Some(true),
+            ledger_history_indexing: Some(true),
+            ..Default::default()
+        })
```
bmwill, line 436: at some point we can change the swarm-config builder to enable these indexes by default.

PR #26719, opened 2026-05-20
[chore] gate staging-only streaming items to silence clippy warnings

tpham-mysten:
## Description 

 - `Subscription` and the streaming e2e tests are gated behind the `staging` feature, so any streaming code only consumed from those paths warns as dead under the default build.
 - Date the dead items inline and group otherwise-orphaned imports via the `mod staging { pub use ...; } use staging::*;` pattern

## Test plan 

- cargo check in staging and non-staging - zero warnings

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/scope.rs
+++ b/crates/sui-indexer-alt-graphql/src/scope.rs
@@ -25,9 +25,16 @@ use sui_types::object::Object as NativeObject;
 
 use crate::config::Limits;
 use crate::error::RpcError;
-use crate::task::streaming::ProcessedCheckpoint;
-use crate::task::streaming::StreamingPackageStore;
 use crate::task::watermark::Watermarks;
+
+#[cfg(feature = "staging")]
+mod staging {
+    pub(super) use crate::task::streaming::ProcessedCheckpoint;
+    pub(super) use crate::task::streaming::StreamingPackageStore;
+}
+
+#[cfg(feature = "staging")]
+use staging::*;
 
 /// A map of objects from an executed transaction, keyed by (ObjectID, SequenceNumber).
 /// None values indicate tombstones for deleted/wrapped objects.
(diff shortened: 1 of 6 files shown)


crates/sui-indexer-alt-graphql/src/scope.rs
```diff
@@ -25,10 +25,17 @@ use sui_types::object::Object as NativeObject;
 
 use crate::config::Limits;
 use crate::error::RpcError;
-use crate::task::streaming::ProcessedCheckpoint;
-use crate::task::streaming::StreamingPackageStore;
 use crate::task::watermark::Watermarks;
 
+#[cfg(feature = "staging")]
+mod staging {
+    pub(super) use crate::task::streaming::ProcessedCheckpoint;
```
tpham-mysten, line 32: Applied pattern [mentioned](https://github.com/MystenLabs/sui/pull/26633/changes#r3237945879) in previous PR.

crates/sui-indexer-alt-graphql/src/task/streaming/mod.rs
tpham-mysten, line 1: I tried gating the whole module first, but lib.rs and gap_recovery tests reference streaming types unconditionally, so the cfgs would cascade across most of lib.rs (+ helper extraction to keep it tidy), so I went with the approach to gate dead items inline and group orphaned imports via mod staging { pub use ...; }. The flag will be removed soon, so I think it's fine

stefan-mysten (approved):
I see no warnings, I happy. Thanks @tpham-mysten!

I stamped but as previously, you guys know best this code and what's the right thing to do here. I think this is ok to merge given that you're working in this area anyway.

My only question is if we have tests covering this staging feature.

tpham-mysten:
Thanks @stefan-mysten! Yes we have a good test coverage using staging flag for subscription [here](https://github.com/MystenLabs/sui/tree/c655581b79ba70d3398428330b573927b1e55568/crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription).

amnn (commented):
(LGTM!)

PR #26720, opened 2026-05-20
[consensus] propose at latest possible round after restart

mwtian:
## Description 

Local last proposed block may be higher in round than the synced last known block. So `self.core.new_block(round + 1, true)?;` may not fire even though the validator can propose a block at a higher round.

## Test plan 

CI


akichidis (approved):
Good find @mwtian 💯

PR #26723, opened 2026-05-20
[consensus] add LeaderSlotDecider for v3 commit rule

mwtian:
## Description 

Implements the stateless pure-decision rule for v3, plus the required DagState API.

LeaderSlotDecider has no production callers in this PR. FlexCommitter will wire it up in a follow-up PR. The new items are marked #[allow(dead_code)] to keep the build clean until then.

## Test plan 

CI

### DIFF
--- a/consensus/core/src/leader_slot_decider.rs
+++ b/consensus/core/src/leader_slot_decider.rs
@@ -1,0 +1,210 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Production callers (FlexCommitter) land in a follow-up PR.
+#![allow(dead_code)]
+
+use std::{
+    collections::{BTreeMap, BTreeSet, VecDeque},
+    sync::Arc,
+};
+
+use consensus_types::block::BlockRef;
+use parking_lot::RwLock;
+
+use crate::{
+    BlockAPI, VerifiedBlock,
+    block::Slot,
+    commit::LeaderStatus,
+    context::Context,
+    dag_state::DagState,
+    stake_aggregator::{CertificationThreshold, QuorumThreshold, StakeAggregator},
+};
+
+#[cfg(test)]
+#[path = "tests/leader_slot_decider_tests.rs"]
+mod leader_slot_decider_tests;
+
+/// Stateless decision logic for commit rule v3.
+pub(crate) struct LeaderSlotDecider {
+    context: Arc<Context>,
+    dag_state: Arc<RwLock<DagState>>,
+}
+
+impl LeaderSlotDecider {
+    pub(crate) fn new(context: Arc<Context>, dag_state: Arc<RwLock<DagState>>) -> Self {
+        Self { context, dag_state }
+    }
+
+    /// Evaluates if `slot` should be directly committed or skipped,
+    /// based on voting information from DagState.
+    ///
+    /// A block is directly committed if it has a quorum of commit votes,
+    /// and a block is directly skipped if it has a quorum of skip votes.
+    /// Misbehaviors are assumed to be under the fault tolerance threshold, but we also
+    /// try to detect if this assumption is broken, when it is convenient.
+    pub(crate) fn try_direct_decide(&self, slot: Slot) -> LeaderStatus {
+        let dag_state = self.dag_state.read();
+        let Some(voting_round_info) = dag_state.get_round_info(slot.round + 1) else {
+            // No blocks accepted in the next round yet, so there is no votes to the slot.
+            return LeaderStatus::Undecided(slot);
+        };
+        if !voting_round_info
+            .total_stake
+            .reached_threshold(&self.context.committee)
+        {
+            // Next round has insufficient stake to decide the slot.
+            return LeaderStatus::Undecided(slot);
+        }
+
+        let block_infos = dag_state.get_block_info_at_slot(slot);
+        let num_blocks = block_infos.len();
+        let mut committed = vec![];
+        let mut num_skipped = 0;
+        for block_info in block_infos {
+            // Commit the leader block if it has a quorum of commit votes.
+            if block_info
+                .children_stake
+                .reached_threshold(&self.context.committee)
+            {
+                committed.push(block_info.block.clone());
+                continue;
+            }
+            // Skip the leader block if it has a quorum of skip votes.
+            // Byzantine authorities that produce both vote and non-vote blocks
+            // can count on both sides, but at most once per side against a decision block.
+            let mut reject_votes = StakeAggregator::<QuorumThreshold>::new();
+            for block_ref in voting_round_info.blocks.difference(&block_info.children) {
+                reject_votes.add_unique(block_ref.author, &self.context.committee);
+            }
+            if reject_votes.reached_threshold(&self.context.committee) {
+                num_skipped += 1;
+            }
+        }
+
+        // When there is no block at the slot or all blocks at the slot are skipped,
+        // the whole slot is skipped.
+        // Even if new blocks arrive at the slot later, they are guaranteed to be
+        // skipped directly since a quorum already rejected them.
+        if num_skipped == num_blocks {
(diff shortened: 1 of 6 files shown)


consensus/core/src/stake_aggregator.rs
```diff
@@ -26,6 +29,15 @@ impl CommitteeThreshold for QuorumThreshold {
     }
 }
 
+impl CommitteeThreshold for CertificationThreshold {
+    fn is_threshold(committee: &Committee, amount: Stake) -> bool {
```
arun-koshy, line 33: `exceeds_threshold` or `meets_threshold`?

consensus/core/src/leader_slot_decider.rs
```diff
@@ -0,0 +1,210 @@
...
+            // No blocks accepted in the next round yet, so there is no votes to the slot.
+            return LeaderStatus::Undecided(slot);
+        };
+        if !voting_round_info
+            .total_stake
+            .reached_threshold(&self.context.committee)
+        {
+            // Next round has insufficient stake to decide the slot.
+            return LeaderStatus::Undecided(slot);
+        }
```
arun-koshy, line 61: nit: combine these two conditionals into one? i.e.

```
let Some(voting_round_info) = dag_state
    .get_round_info(slot.round + 1)
    .filter(|info| info.total_stake.reached_threshold(&self.context.committee))
else {
    // Next round either has no blocks accepted yet or has insufficient stake.
    return LeaderStatus::Undecided(slot);
};
```

arun-koshy (approved):

akichidis (changes requested):
Apologies for blocking this - nothing wrong here - but I would like to make sure I've fully reviewed this part before you merge it as it's critical. Will have the review wrapped up by tomorrow. Thanks 🙏

consensus/core/src/stake_aggregator.rs
mwtian: This is from the trait. But I'm thinking of refactoring the StakeAggregator to not be tied a specific threshold.

consensus/core/src/leader_slot_decider.rs
mwtian: I slightly prefer the current version.

consensus/core/src/tests/leader_slot_decider_tests.rs
```diff
@@ -0,0 +1,543 @@
...
+        LeaderStatus::Undecided(s) => assert_eq!(s, slot),
+        other => panic!("Expected Undecided, got {other:?}"),
+    }
+}
+
+/// A fully connected round 2 commits the round-1 leader directly.
+#[tokio::test]
+async fn try_direct_decide_commit() {
+    telemetry_subscribers::init_for_testing();
+    let (context, dag_state, decider) = setup(6, 1, 0);
```
akichidis, line 72: Oh just realised that eventually all the simtests etc (basically any testing that involves consensus) will also need a minimum of 6 validators testing right?

consensus/core/src/leader_slot_decider.rs
```diff
@@ -0,0 +1,210 @@
...
+        // When there is a committed block, assume no other block at the slot can be committed.
+        if committed.len() == 1 {
+            return LeaderStatus::Commit(committed.pop().unwrap());
+        }
+
+        // There are undecided blocks, so the slot is undecided.
+        LeaderStatus::Undecided(slot)
+    }
+
+    /// Evaluates every slot in `slots` (all at `decision_round`) for indirect commit.
```
akichidis: `slots` --> `decision_slots` ?

consensus/core/src/leader_slot_decider.rs
```diff
@@ -0,0 +1,210 @@
...
+                    continue;
+                }
+                // Set up next setup of BFS.
+                if visited.insert(*ancestor) {
+                    let ancestor_block = dag_state
+                        .get_block(ancestor)
+                        .unwrap_or_else(|| panic!("Block {} must exist", ancestor));
+                    to_visit.push_back(ancestor_block);
+                }
+                // Continue to only if this is a vote from the block to a decision block.
```
akichidis: Continue to only?

consensus/core/src/leader_slot_decider.rs
```diff
@@ -0,0 +1,210 @@
...
+                if ancestor.round < decision_round {
+                    // This ancestor and its ancestors are irrelevant for voting.
+                    continue;
+                }
+                // Set up next setup of BFS.
+                if visited.insert(*ancestor) {
+                    let ancestor_block = dag_state
+                        .get_block(ancestor)
+                        .unwrap_or_else(|| panic!("Block {} must exist", ancestor));
+                    to_visit.push_back(ancestor_block);
```
akichidis, line 171: Do we need to add this into `to_visit` if `ancestor_block.round == decision_round` ?

consensus/core/src/leader_slot_decider.rs
```diff
@@ -0,0 +1,210 @@
...
+                // Continue to only if this is a vote from the block to a decision block.
+                if decision_blocks.contains(ancestor) && block.round() == decision_round + 1 {
+                    let commit_stake = votes.entry(*ancestor).or_default();
+                    commit_stake.add_unique(block.author(), &self.context.committee);
+                }
+            }
+        }
+
+        // Collect the blocks with commit certificates.
+        let mut commit_certificates = BTreeMap::<Slot, Vec<VerifiedBlock>>::new();
```
akichidis: The `commit_certificates` at first read looks like this map holds the actual certificates (blocks etc) not the slots that might got certified. 

I would suggest having something such as `certified_slots` ?

consensus/core/src/leader_slot_decider.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// Each slot will be either committed or skipped, never undecided after applying the
+    /// indirect commit rule.
+    pub(crate) fn try_indirect_decide(
+        &self,
+        anchor_block: &VerifiedBlock,
+        decision_slots: &[Slot],
+    ) -> Vec<LeaderStatus> {
+        assert!(!decision_slots.is_empty());
+        let decision_round = decision_slots[0].round;
+        assert!(decision_slots.iter().all(|s| s.round == decision_round));
```
akichidis, line 135: shall we also have an assert like `assert!(decision_round < anchor_block.round())`

consensus/core/src/leader_slot_decider.rs
```diff
@@ -0,0 +1,210 @@
...
+            return LeaderStatus::Skip(slot);
+        }
+
+        // Under safety assumption, at most one block can be committed in a slot.
+        if committed.len() > 1 {
+            panic!(
+                "Multiple committed blocks found at leader slot {}: {:?}",
+                slot, committed
+            );
+        }
```
akichidis: Looking on the offline discussion as well, having a test for this case would be a bonus - just to confirm that indeed it will break and detect the violation.

akichidis (approved):

consensus/core/src/tests/leader_slot_decider_tests.rs
mwtian: I thought about this before. In simtests without explicit misbehavior testing which is the majority, we can use a configuration of f=0 and c=1, which will still be 4 validators.

consensus/core/src/leader_slot_decider.rs
mwtian: Good point.

consensus/core/src/leader_slot_decider.rs
mwtian: The logic has been refactored but a test is added to verify panic on equivocations.

consensus/core/src/tests/leader_slot_decider_tests.rs
akichidis: Good point. It will definitely avoid us triggering a large refactoring. In the same time though very implicit for testing and might break the assumptions that others think when testing.

PR #26724, opened 2026-05-20
Simplify bitmap evaluation

nickvikeras:
## Description

Rewriting the evaluation code in a simpler way that eliminates a bunch of possible problems. I initially designed this to be composable into arbitrarily nested queries, but since we force everything to be DNF at the proto level, that's not really necessary. This new design just scans each leaf up to the floor in lock-step (so branches in a disjunction can't run way ahead of each other).

This reduces the amount of potential re-scanning we have to do on a resumption requestion, and eliminates a lot of the watermark propagation complexity.

### DIFF
--- a/crates/sui-inverted-index/src/bitmap_query/iter.rs
+++ b/crates/sui-inverted-index/src/bitmap_query/iter.rs
@@ -3,43 +3,44 @@
 
 //! Synchronous iterator evaluator for DNF bitmap queries.
 //!
-//! Mirrors the async [`super::stream`] evaluator's merge-joins and watermark
-//! propagation, but stays fully synchronous so request-local backends (e.g.
-//! RocksDB) can drive it from the blocking task that owns their iterators. Like
-//! the stream evaluator it emits matching members as [`Watermarked`] items
-//! interleaved with progress watermarks: each leaf brackets every bucket with
-//! `Watermark(pre), Item, Watermark(post)` and caps natural EOF with a
-//! range-terminus watermark, and the combinators merge per-child watermarks
-//! (min ascending / max descending) so sparse scans still report progress at the
-//! rate the slowest source advances.
+//! A single flat driver merge-joins every leaf scan against one shared *floor*
+//! (the slowest leaf's position). At the floor bucket it evaluates the query —
+//! intersect each term's included dimensions, subtract its excluded ones, then
+//! union across terms — and emits a watermark at the floor. Because leaves only
+//! ever advance at the floor (peeked one bucket ahead), no branch can run ahead
+//! of the others: the resume cursor is always within one sparse read of every
+//! leaf, and there is no windowing/parking machinery to get wrong. Mirrors the
+//! async [`super::stream`] evaluator, which shares the per-bucket evaluation
+//! ([`eval_term_at_bucket`]) and is cross-checked against this one in tests.
+//!
+//! Budget accounting lives in the request layer for the iterator path (each
+//! backend leaf iterator charges its own per-request budget and yields an error
+//! on exhaustion), so this evaluator only propagates those errors.
 
 use std::collections::VecDeque;
 use std::iter::Peekable as IterPeekable;
 use std::ops::Range;
 
-use anyhow::Result;
-use anyhow::bail;
-use itertools::Itertools;
 use roaring::RoaringBitmap;
 
 use super::BitmapBucketIteratorSource;
 use super::BitmapQuery;
-use super::BitmapTerm;
 use super::BucketItem;
+use super::LeafHead;
 use super::MultiError;
 use super::ScanDirection;
+use super::TermSpec;
 use super::Watermarked;
 use super::WatermarkedBucket;
 use super::bound_in_direction;
-use super::complete_peeks;
+use super::bucket_edges;
+use super::eval_term_at_bucket;
 use super::frontier_advanced;
-use super::merge_watermarks;
-use super::merge_with_ceiling;
 use super::split_term_literals;
 
-/// Evaluate a DNF `BitmapQuery` as an ordered iterator of marked bucket
-/// bitmaps. Output emits `Watermarked::Item((bucket_id, bitmap))` interleaved
-/// with `Watermarked::Watermark(p)` merged from the leaf scans.
+/// Evaluate a DNF `BitmapQuery` as an ordered iterator of marked bucket bitmaps.
+/// Output emits `Watermarked::Item((bucket_id, bitmap))` interleaved with
+/// `Watermarked::Watermark(p)` derived from the slowest leaf's progress.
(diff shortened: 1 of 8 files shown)


crates/sui-inverted-index/src/bitmap_query/iter.rs
```diff
@@ -3,43 +3,58 @@
...
-//! RocksDB) can drive it from the blocking task that owns their iterators. Like
-//! the stream evaluator it emits matching members as [`Watermarked`] items
-//! interleaved with progress watermarks: each leaf brackets every bucket with
-//! `Watermark(pre), Item, Watermark(post)` and caps natural EOF with a
-//! range-terminus watermark, and the combinators merge per-child watermarks
-//! (min ascending / max descending) so sparse scans still report progress at the
-//! rate the slowest source advances.
+//! A single flat driver merge-joins every leaf scan against one shared *floor*
+//! (the slowest leaf's position), evaluates the whole DNF — `OR` of
+//! (`AND` includes `MINUS` `OR` excludes) — at the floor bucket, and derives the
```
bmwill: this sentence seems weird to me

crates/sui-inverted-index/src/bitmap_query/iter.rs
```diff
@@ -3,43 +3,58 @@
...
-/// bitmaps. Output emits `Watermarked::Item((bucket_id, bitmap))` interleaved
-/// with `Watermarked::Watermark(p)` merged from the leaf scans.
+/// One DNF term, as index spans into the flat leaf vector.
+struct TermSpec {
+    includes: Vec<usize>,
+    excludes: Vec<usize>,
+    /// Cleared once any include leaf hits EOF: the intersection can never match
+    /// again, so the whole term is retired and its leaves dropped.
+    dead: bool,
+}
```
bmwill: Once you reach the end of the bucket in the forward direct, that doesn't mean there will never be any more hits for that particular bucket right? Does this account for that properly?

crates/sui-inverted-index/src/bitmap_query/iter.rs
```diff
@@ -372,375 +126,181 @@ where
...
+            // An include at EOF kills its term (the intersection is permanently
+            // empty); drop all of that term's leaves so they stop being polled.
+            for term in terms.iter_mut() {
+                if !term.dead
+                    && term
+                        .includes
+                        .iter()
+                        .any(|&i| matches!(class[i], Some(LeafHead::Eof)))
+                {
+                    term.dead = true;
```
bmwill, line 147: Does this also effectively cap the the EOF point in history for all other terms?

crates/sui-inverted-index/src/bitmap_query/stream.rs
```diff
@@ -3,12 +3,15 @@
...
 //!
-//! Emits matching members as [`Watermarked`] items interleaved with progress
-//! watermarks. The merge-join combinators (`intersect_n`, `union_n`,
-//! `subtract_two`) propagate per-source watermarks so sparse scans still report
-//! progress at the rate sources advance. Consumed by streaming backends such as
-//! BigTable; the synchronous [`super::eval_bitmap_query_bucket_iter`] mirrors it
-//! for request-local iterator backends.
+//! A single flat driver merge-joins every leaf scan against one shared *floor*
+//! (the slowest leaf's position), evaluates the whole DNF — `OR` of
+//! (`AND` includes `MINUS` `OR` excludes) — at the floor bucket, and derives the
```
bmwill: same comment about this sentence being confusing

crates/sui-inverted-index/src/bitmap_query/stream.rs
```diff
@@ -178,72 +176,64 @@ where
...
+    /// again, so the whole term is retired and its leaves dropped.
+    dead: bool,
+}
+
+/// A leaf's head this round, from a non-consuming peek.
+enum LeafHead {
+    Bucket(u64),
+    Eof,
+    Error,
+}
```
bmwill: Can we share these types between the stream and iter impls?

bmwill:
Some small other nits:

```
1. stream.rs:407 — the take_side! macro is asymmetric with iter.rs's take_term_side function. The macro captures class/front/leaves/floor_bucket/post from the enclosing scope, presumably because of mutable-borrow + .await ergonomics inside try_stream!. But an
   async fn take_side(indices, floor_bucket, post, class, front, leaves) -> Vec<Option<RoaringBitmap>> would type-check (the borrows are disjoint and not held across await) and would visually mirror iter.rs::take_term_side. Could be a readability win if lifetimes work out.

  2. Per-round Vec allocations. Each loop iteration in both stream.rs:298 and iter.rs:120 allocates peeks, class, errors, and active. For dense scans (many buckets), this is allocator pressure on the hot path. Hoisting these outside the loop and .clear()-ing
  them each round would eliminate it.
```

crates/sui-inverted-index/src/bitmap_query/iter.rs
nickvikeras: Not the terms in other branches of the top-level OR. Those can keep advancing the floor. The cursor that we propagate forward as a watermark continue to track the floor, which is held back by the slowest non-dead branch.

crates/sui-inverted-index/src/bitmap_query/iter.rs
nickvikeras: Yeah that's exactly what this is doing. I think the comment might have been confusing (it said "Cleared" not "Set).

bmwill: Is there a possible race where we reach a dead term but then a new checkpoint is processed and fills in a value for a bucket for that term and so we'd miss it if we marked the bucket as dead? Or is the term processing essentially caped at a known checkpoint/txnid frontier so that we'd end the processing of the whole query at that cap and a subsequent query would pick up at that previous watermark to catch the newly added term at the frontier?

nickvikeras: And keep in mind we clamp the bounds of this scan to the cached bigtable watermark at the start of the request, so this can never hide data from the client.

bmwill (approved):

amnn (commented):
One nice optimisation that you can perform in this simplified model is to support "leap-frogging" between leaves in a conjunction. 

If you have a query of the form `a AND b` and you know that the first instance of `a` is at block `i` and the first instance of `b` is at block `j` where `i << j`, then if `a` is a dense filter, you would need to check each of the populated blocks in the interval (of which there could be many) just to see that they are less than `j`. Instead if your source trait could support a `skip_until` API, then you could use that to tell each `leaf` to skip_until the furthest ahead leaf, repeatedly, until they all line up on a block.

In pathological cases this may not save you anything (where two leaves produce exactly interleaved streams of blocks), but in practical cases where there is low overlap between leaves, this can make a big difference.

PR #26726, opened 2026-05-20
ci: add simtest host disk guardrails

lxfind:
## What

Adds a disk-space guardrail and exit-marker hardening to the nightly simulator-tests workflow, so a full `simtest-01` disk no longer surfaces as a confusing test failure.

## Changes

- **Free-space floor (`SIMTEST_MIN_FREE_GB`, 100G).** A check SSHes into `simtest-01` right before the run starts and fails the job early if `/` has less than the floor free, dumping `df`/`du` for triage.
- **Active-run marker.** Writes `~/simtest_active_run` (workflow / run id / url / start time / floor) when a run starts, and clears it on `Environment clean` and in an `always()` teardown step, so a stale host is easy to spot.
- **Exit-marker hardening.** The wait loop and results step now treat a missing, empty, or non-integer `~/simtest_exit_code` as an infrastructure failure (`BAD_EXIT_MARKER`) instead of silently passing.
- Adds the standard license header; folds in a pre-existing `publically` → `publicly` typo fix flagged by the repo's `typos` hook.

### DIFF
--- a/.github/workflows/simulator-nightly.yml
+++ b/.github/workflows/simulator-nightly.yml
@@ -1,3 +1,6 @@
+# Copyright (c) Mysten Labs, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
 name: Simulator Tests
 
 concurrency:
@@ -28,6 +31,7 @@ on:
 env:
   SUI_REF: "${{ github.event.inputs.sui_ref || 'main' }}"
   TEST_NUM: "${{ github.event.inputs.test_num || '30' }}"
+  SIMTEST_MIN_FREE_GB: "100"
 
 jobs:
   simtest:
@@ -49,7 +53,7 @@ jobs:
         id: auth
         uses: teleport-actions/auth@d48447391aa28b202c98ae3f3358097025c32511 # pin@v2.0.4
         with:
-          # Specify the publically accessible address of your Teleport proxy.
+          # Specify the publicly accessible address of your Teleport proxy.
           proxy: mysten.teleport.sh:443
           # Specify the name of the join token for your bot.
           token: sui-simtest-token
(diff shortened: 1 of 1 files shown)


ebmifa (commented):
Left two inline comments on the simtest workflow guardrail changes.

.github/workflows/simulator-nightly.yml
```diff
@@ -183,6 +211,11 @@ jobs:
           echo "EOF" >> $GITHUB_OUTPUT
           echo "log_dir=$LOG_DIR" >> $GITHUB_OUTPUT
 
+      - name: Clear simtest active marker
+        if: always()
```
ebmifa: This runs on workflow cancellation too. GitHub re-evaluates `always()` steps during cancellation, but the remote simtest was started with `nohup ... &`, so a canceled Actions run can leave the remote process running while this step removes the only active-run marker. Consider `if: always() && !cancelled()` here, or explicitly kill the remote simtest processes before clearing the marker.

.github/workflows/simulator-nightly.yml
```diff
@@ -130,9 +135,22 @@ jobs:
       # Run simulator tests via detached execution to survive SSH session drops.
       # The test script runs independently on the remote machine; we poll for completion
       # with short-lived SSH calls instead of holding a single long-lived session.
+      - name: Check simtest host disk space before run
```
ebmifa: The disk guard currently runs after cleanup, clone/checkout, and `./scripts/simtest/install.sh`. If the host is already below the free-space floor, those setup steps can fail first with a raw clone/install error and skip the new `df`/`du` diagnostics. Consider moving an initial free-space check to just after cleanup but before `git clone`, and optionally keeping this pre-run check too.

.github/workflows/simulator-nightly.yml
lxfind: Good catch — switched this step to `if: always() && !cancelled()` in be3a09c3ca. On cancellation the marker now stays put (correctly reflecting that the nohup'd remote run may still be in flight), and the next nightly's `Environment clean` step pkills the survivor and removes the marker together.

.github/workflows/simulator-nightly.yml
lxfind: Addressed in be3a09c3ca: moved the single floor check to run right after `Environment clean` (the max-free point), before the clone/install/build that actually consume space. To make it truly *before* the clone, the `git clone` moved out of `Environment clean` into the `Checkout sui repo` step. So an already-full host now fails with df/du diagnostics instead of a raw clone/install error. Kept it as a single check rather than two.

ebmifa (approved):

PR #26727, opened 2026-05-20
[consensus] add BlockV3

mwtian:
## Description 

Introduces the BlockV3 block format used by v3 consensus.

No production code constructs BlockV3 yet.

## Test plan 

CI

### DIFF
--- a/consensus/core/src/block.rs
+++ b/consensus/core/src/block.rs
@@ -62,6 +62,7 @@ pub(crate) struct BlockTransactionVotes {
 pub enum Block {
     V1(BlockV1),
     V2(BlockV2),
+    V3(BlockV3),
 }
 
 #[allow(private_interfaces)]
@@ -78,6 +79,9 @@ pub trait BlockAPI {
     fn commit_votes(&self) -> &[CommitVote];
     fn transaction_votes(&self) -> &[BlockTransactionVotes];
     fn misbehavior_reports(&self) -> &[MisbehaviorReport];
+    /// The author's local GC round at the time the block was proposed.
+    /// Only `BlockV3` carries this — earlier variants panic.
+    fn transaction_votes_cutoff_round(&self) -> Round;
 }
 
 #[derive(Clone, Default, Deserialize, Serialize)]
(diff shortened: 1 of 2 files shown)


consensus/core/src/block.rs
```diff
@@ -273,11 +281,124 @@ impl BlockAPI for BlockV2 {
...
+    round: Round,
+    author: AuthorityIndex,
+    timestamp_ms: BlockTimestampMs,
+    ancestors: Vec<BlockRef>,
+    transactions: Vec<Transaction>,
+    transaction_votes: Vec<BlockTransactionVotes>,
+    commit_votes: Vec<CommitVote>,
+    misbehavior_reports: Vec<MisbehaviorReport>,
+    /// The author's local GC round at the time the block was proposed.
+    transaction_votes_cutoff_round: Round,
```
akichidis: Maybe then we could go with something like `local_gc_round` , or just `gc_round`? 
I know that you want to make this domain specific for the transactions, and it totally makes sense, but probably it's one of those properties we could leverage in the future for other purposes as well?

consensus/core/src/block.rs
```diff
@@ -273,11 +281,124 @@ impl BlockAPI for BlockV2 {
...
+
+    fn transaction_votes_cutoff_round(&self) -> Round {
+        self.transaction_votes_cutoff_round
+    }
 }
 
 /// Slot is the position of blocks in the DAG. It can contain 0, 1 or multiple blocks
 /// from the same authority at the same round.
-#[derive(Clone, Copy, PartialEq, PartialOrd, Default, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Default, Hash)]
```
akichidis, line 408: Why `Eq` and `Ord` are needed now?

akichidis (approved):

consensus/core/src/block.rs
mwtian: We may need a `transaction_votes_cutoff_round` above `gc_round` in some cases. We can definitely add a GC round once there is need.

consensus/core/src/block.rs
mwtian: They will be used in later when Slot is a key in BTreeMap / BTreeSet.

consensus/core/src/block.rs
akichidis: If that's the case then I would suggest updating the comment, because it basically points that this would be the same as the gc round.

mwtian: Will be updated in the next PR.

PR #26728, opened 2026-05-21
Let the streams flow!

nickvikeras:
## Description

In my initial prototype, I let items flow through each pipeline stage freely as they were delivered by the previous stage. When I added a request-level big-table concurrency limiter for production, I realized it could deadlock because the previous stage's stream wouldn't be fully drained (and hence the concurreny limit's semaphore permit wouldn't be dropped) when the next stage was backpressuring. If the next stage was backpressuring while trying to get the permit, deadlock. To fix that I added a chunk-level buffer that fully drained the stream before pushing anything to the next stage. That chunk-level buffering seems to have severely harmed latency.

This fixes it while still allowing for request-level concurrency limits by draining the streams on an async background task that is allowed to fully drain the stream to an unbounded channel that the next stage reads from. Unbounded channels are fine because the number of in-flight chunks is limited per stage (the stage-level semaphore doesn't return it's permit until the unbounded channel is fully drained).

### DIFF
--- a/crates/sui-rpc-api/src/grpc/deadline.rs
+++ b/crates/sui-rpc-api/src/grpc/deadline.rs
@@ -116,9 +116,11 @@ where
                             )))?;
                         }
                         Err(_) => {
-                            // Cancellation — only possible if the abort
-                            // guard fired (which only happens on Drop), so
-                            // we shouldn't be polling. Treat as EOF.
+                            // Cancellation — only reachable at runtime shutdown
+                            // (our abort guard fires on Drop, when we wouldn't be
+                            // polling). EOF is harmless then; assert in dev to
+                            // catch any future path that makes this reachable.
+                            debug_assert!(false, "with_deadline producer cancelled unexpectedly");
                             break;
                         }
                     }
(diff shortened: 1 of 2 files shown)


crates/sui-rpc-api/src/grpc/deadline.rs
```diff
@@ -116,9 +116,11 @@ where
...
                         }
                         Err(_) => {
-                            // Cancellation — only possible if the abort
-                            // guard fired (which only happens on Drop), so
-                            // we shouldn't be polling. Treat as EOF.
+                            // Cancellation — only reachable at runtime shutdown
+                            // (our abort guard fires on Drop, when we wouldn't be
+                            // polling). EOF is harmless then; assert in dev to
+                            // catch any future path that makes this reachable.
+                            debug_assert!(false, "with_deadline producer cancelled unexpectedly");
```
bmwill: i'd probably not assert on this as it may cause some issues in tests when they are being torn down.

bmwill (approved):

crates/sui-rpc-api/src/grpc/deadline.rs
```diff
@@ -116,10 +116,9 @@ where
                             )))?;
                         }
                         Err(_) => {
-                            // Cancellation — only possible if the abort
-                            // guard fired (which only happens on Drop), so
-                            // we shouldn't be polling. Treat as EOF.
-                            break;
+                            // Cancellation — only reachable at runtime shutdown
+                            // (our abort guard fires on Drop, when we wouldn't be
+                            // polling), so EOF is harmless.                             break;
```
amnn, line 121: Should this still be a `break;` case?

PR #26731, opened 2026-05-21
[graphql-stream] Add cursor field to Checkpoint [13/n]

tpham-mysten:
## Description 

- Expose cursor: String! on Checkpoint — encoded JsonCursor<u64> of the sequence number, matching the format Query.checkpoints pagination already uses.
- Clients (especially streaming subscribers) read this on each yielded checkpoint and pass it back as afterCursor to resume from this point.

## Test plan 

- e2e + unit tests

---

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425
   - #26495
   - #26731

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Add cursor field to Checkpoint object
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/lib.rs
+++ b/crates/sui-indexer-alt-graphql/src/lib.rs
@@ -80,6 +80,11 @@ const GRAPHQL_SUBSCRIPTIONS_PATH: &str = "/graphql/subscriptions";
 const HEALTH_PATH: &str = "/graphql/health";
 
 mod api;
+pub use crate::api::scalars::cursor::JsonCursor;
+pub use crate::api::types::checkpoint::CCheckpoint;
+pub use crate::api::types::event::CEvent;
+pub use crate::api::types::event::EventCursor;
+pub use crate::api::types::transaction::CTransaction;
 pub mod args;
 pub mod config;
 mod error;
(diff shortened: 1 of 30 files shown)


amnn (commented):
I'm not a fan of introducing a cursor type that is intended for subscription in a type that is used in many other places. Can we use the same approach as we do for pagination and return a `CheckpointEdge`? The edge contains the cursor and then the checkpoint itself as a `node`.

tpham-mysten:
Agreed! I have updated the Schema to use CheckpointEdge. Also update EventEdge and TransactionEdge

crates/sui-indexer-alt-graphql/src/lib.rs
amnn, line 1: nit: crate imports for these?

amnn (approved):
Looks good! Thanks @tpham-mysten

crates/sui-indexer-alt-graphql/src/lib.rs
tpham-mysten: Sg. done

PR #26734, opened 2026-05-21
[cli] Fix tree shake alg by using the pkg id instead of pkg display name

stefan-mysten:
## Description 

This PR fixes the tree shake algorithm by using the pkg id instead of pkg display name. As the build system disambiguates same package names by adding suffixes, the tree shake algorithm did not account for this and thus dropped packages that were required to be in the dependency list.

## Test plan 

New tests + existing CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-move-build/src/lib.rs
+++ b/crates/sui-move-build/src/lib.rs
@@ -363,7 +363,11 @@ impl CompiledPackage {
     /// Return the set of Object IDs corresponding to this package's transitive dependencies'
     /// storage package IDs (where to load those packages on-chain).
     pub fn get_dependency_storage_package_ids(&self) -> Vec<ObjectID> {
-        self.dependency_ids.published.values().cloned().collect()
+        self.dependency_ids
+            .published
+            .values()
+            .map(|dep| dep.published_at)
+            .collect()
     }
 
     /// Return a digest of the bytecode modules in this package.
@@ -564,7 +568,11 @@ impl CompiledPackage {
     }
 
     pub fn get_published_dependencies_ids(&self) -> Vec<ObjectID> {
-        self.dependency_ids.published.values().cloned().collect()
+        self.dependency_ids
+            .published
+            .values()
+            .map(|dep| dep.published_at)
+            .collect()
     }
 }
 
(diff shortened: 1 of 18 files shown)


crates/sui-move-build/src/lib.rs
```diff
@@ -588,35 +596,106 @@ pub enum PublishedAtError {
...
+#[derive(Debug, Clone)]
+pub struct ConflictingDependency {
+    /// Unique package graph ID used by the compiler and build artifacts.
+    ///
+    /// This may differ from `name` when multiple packages have the same declared package name,
+    /// for example `foo` and `foo_1`.
+    pub id: Symbol,
+    /// Human-readable package name declared by the package.
+    pub name: Symbol,
+    pub lock_file_address: ObjectID,
```
stefan-mysten, line 653: This is now published_at / published_address.

stefan-mysten: left it like it is for now.

crates/sui-move-build/src/lib.rs
```diff
@@ -588,35 +596,106 @@ pub enum PublishedAtError {
...
+    pub conflicting: BTreeMap<Symbol, ConflictingDependency>,
+}
+
+#[derive(Debug, Clone)]
+pub struct PublishedDependency {
+    /// Unique package graph ID used by the compiler and build artifacts.
+    ///
+    /// This may differ from `name` when multiple packages have the same declared package name,
+    /// for example `foo` and `foo_1`.
+    pub id: Symbol,
```
mdgeorge4153, line 615: I think at this point the package IDs should really be dropped - I think we're only using them for deduplication, which we should do based on the published_at I think.

mdgeorge4153 (approved):
I think this is reasonable as a short-term fix; it solves the specific problem of using non-unique names for tree shaking. I think longer-term, we shouldn't be using names for tree-shaking at all; the algorithm should operate entirely on addresses and the names should be irrelevant (or maybe used in error messages).

As I was looking through the code, I think there may be bugs in the way we're doing tree shaking, but I'm not sure. I'll try to put together a PR that works differently, but I think we should merge this.

crates/sui-move-build/src/lib.rs
stefan-mysten: We probably need to rethink this tree shake algorithm, like you mentioned offline. It's largely based on the original one before the new pkg system.

PR #26742, opened 2026-05-21
docs: remove outdated package management content from upgrade page

jessiemongeon1:
## Summary
- Remove legacy `[addresses]` section and `published-at` field from the `Move.toml` example on the upgrade docs page
- Remove two tabbed sections ("Automated Addresses" / "Manual Addresses") describing the old `Move.lock`-based and manual address management workflows, both superseded by the v1.63+ `Published.toml` system
- Replace with concise notes that `Published.toml` automatically tracks addresses after publish and upgrade

## Motivation
The outdated content was causing Ask Sui AI / Kappa AI to recommend the old `published-at`/`[addresses]` workflow when users hit upgrade errors, leading to confusion since these fields no longer exist in the current package system.

## Test plan
- [x] Docs site builds without errors
- [ ] Verify the [upgrade page](https://docs.sui.io/develop/publish-upgrade-packages/upgrade) renders correctly
- [ ] Confirm no broken internal links

### DIFF
--- a/docs/content/develop/manage-packages/automated-address-management.mdx
+++ b/docs/content/develop/manage-packages/automated-address-management.mdx
@@ -1,8 +1,15 @@
 ---
-title: Automated Address Management
-description: Packages published across Mainnet, Testnet, and Devnet each have different addresses. Automated address management tracks these addresses for you.
+title: Automated Address Management (Legacy, pre-v1.63)
+sidebar_label: Address Management (Legacy)
+description: "Legacy documentation for the pre-v1.63 Move.lock-based address tracking system. For the current package system, see Move Package Management."
 keywords: [ track address, track package address, package addresses, managing package addresses, managing addresses, tracking addresses, tracking packages, tracking mainnet address, track mainnet address, track testnet address, tracking testnet address, track devnet address, tracking devnet address, conflicting addresses, conflicting package addresses ]
 ---
+
+:::warning Legacy documentation
+
+This page describes the **pre-v1.63** address management system that used `Move.lock` and `published-at` fields in `Move.toml`. This system has been replaced by `Published.toml` and the `[environments]` section. See [Move Package Management](/develop/manage-packages/move-package-management) for the current system, and [the migration guide](/references/package-managers/package-manager-migration) for how to migrate.
+
+:::
 
 When you publish or upgrade a package, its address is tracked in the `Move.lock` file. A package's address is its package ID. This happens automatically, so you do not need to manually record or update hexadecimal addresses.
 
--- a/docs/content/develop/manage-packages/automated-address-management.mdx
+++ b/docs/content/develop/manage-packages/automated-address-management.mdx
@@ -7,7 +7,7 @@ keywords: [ track address, track package address, package addresses, managing pa
 
 :::warning Legacy documentation
 
-This page describes the **pre-v1.63** address management system that used `Move.lock` and `published-at` fields in `Move.toml`. This system has been replaced by `Published.toml` and the `[environments]` section. See [Move Package Management](/develop/manage-packages/move-package-management) for the current system, and [the migration guide](/references/package-managers/package-manager-migration) for how to migrate.
+This page describes the **pre-v1.63** address management system that used `Move.lock` and `published-at` fields in `Move.toml`. See [Move Package Management](/develop/manage-packages/move-package-management) for the current system, and [the migration guide](/references/package-managers/package-manager-migration) for how to migrate.
 
 :::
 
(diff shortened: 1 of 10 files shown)


jessiemongeon1:
## Style Guide Audit

Audited **10** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

### `docs/content/develop/manage-packages/move-package-management.mdx` (1 violation(s))

_1 violation(s) (0 regex, 1 claude)_

- **Line 26** — no-note-at-start
  - Current: `Note that `mainnet` and `testnet` are implicitly available`
  - Fix: `The `mainnet` and `testnet` values are implicitly available`

---

docs/content/develop/manage-packages/automated-address-management.mdx
```diff
@@ -1,9 +1,16 @@
...
+title: Automated Address Management (Legacy, pre-v1.63)
+sidebar_label: Address Management (Legacy)
+description: "Legacy documentation for the pre-v1.63 Move.lock-based address tracking system. For the current package system, see Move Package Management."
 keywords: [ track address, track package address, package addresses, managing package addresses, managing addresses, tracking addresses, tracking packages, tracking mainnet address, track mainnet address, track testnet address, tracking testnet address, track devnet address, tracking devnet address, conflicting addresses, conflicting package addresses ]
 ---
 
+:::warning Legacy documentation
+
+This page describes the **pre-v1.63** address management system that used `Move.lock` and `published-at` fields in `Move.toml`. This system has been replaced by `Published.toml` and the `[environments]` section. See [Move Package Management](/develop/manage-packages/move-package-management) for the current system, and [the migration guide](/references/package-managers/package-manager-migration) for how to migrate.
```
stefan-mysten:
```suggestion
This page describes the **pre-v1.63** address management system that used `Move.lock` and `published-at` fields in `Move.toml`. See [Move Package Management](/develop/manage-packages/move-package-management) for the current system, and [the migration guide](/references/package-managers/package-manager-migration) for how to migrate.
```

stefan-mysten (commented):
Thanks @jessiemongeon1! 

@mdgeorge4153 hope you can have a look as well.

mdgeorge4153 (commented):
I took a look at other mentions of Move.toml in the docs to look for old package management stuff, here is what I found:
 - the Active environment not found warning in move-package-management.mdx is incorrect
 - in testing-debugging/common_errors, we suggest adding an `addresses` section to Move.toml, this is incorrect and will cause the system to treat the package as an old-style Move.toml; we should never recommend that people add `[addresses]`.
 - in common_errors, one of the causes is "your manifest does not exist", I think that it should say "your dependency's manifest does not exist", although even then I'm not sure you'd get this error.
 - The "Your active environment is not present" in common_errors uses a bad example, `mainnet` and `testnet` are both implicitly present in the `environments` section. Moreover, the correct action is more subtle and depends on your needs. The most likely problem is that they are on localnet, which they should _not_ add to their `Move.toml` environments section. Instead, if they want to publish on localnet and devnet the most likely correct thing to do is use `test-publish` instead of `publish`; this avoids polluting their `Published.toml` with their localnet addresses. I *think* we updated the error message that you get to explain that more clearly.
 - Also, the syntax in that note is wrong; it should be `mainnet = "35834a8a"` not `mainnet = { chain-id = "35834a8a" }` (but again, you shouldn't be adding `mainnet`).
 - in `move-best-practices` under "Organization principles" we refer people to the old page, not the new page. 
 - We should probably also include a note about Published.toml in that section
 - in `lootbox-ctf` the sample manifest is old-style, it should not have an `[addresses]` section
 - I didn't review the linked example repos, but we should probably check that they have modern Move.toml files.
 - in `hello-world` it says you must build before you test or publish it, but this is false; both `sui move test` and `sui move publish` automatically build the package first.
 - I'm not really sure what the `wasm-template.mdx` file is about, since I didn't read it carefully, but the fact that it suggests publishing all the dependencies makes me think that it may be a good use case for `sui move test-publish --unpublished-deps`
 - the `deploy-tokenized-asset` file also suggests adding an address to the `[addresses]` section. I think this should probably use the `sui move test-publish --publish-unpublished-deps` workflow. I also suspect that this would replace the benefit of the `npm run publish-template` command, I'd recommend trying to remove that entirely.
 - The references/cli/client.mdx points to the move book for package management, it might be better to point to the docs page on package management. I'm not sure if the move book is up-to-date.
 - the `nautilus-weather-oracle` `Move.toml` is old-style
 - In sui-stack-suins the example uses a git dependency, I'm not sure if suins is in mvr, but if it is we should recommend the mvr dependency instead.

jon-mysten (approved):

PR #26749, opened 2026-05-22
sui-rpc-api: replace alpha proof/event services with v2alpha

bmwill:
Previously, sui-rpc-api shipped its own alpha `ProofService` and `EventService` to back the in-tree authenticated-events light client. With the v2alpha `ProofService` (`GetCheckpointObjectProof`, with both inclusion and non-inclusion proofs) and `LedgerService.ListEvents` (filtered by `EventStreamHeadFilter`) now defined in the `sui-rpc` SDK, the in-tree stubs are obsolete.

With this commit:
- the alpha proof and event services are removed from `sui-rpc-api` and the v2alpha `ProofService` is mounted instead, backed by `sui_sdk_types::merkle::MerkleTree` so the SDK-side proto conversions carry the wire representation;
- the rpc-index column family `events_by_stream` plus its compaction filter, `EventIndexKey`, `event_iter`, and `authenticated_event_iter` are dropped, along with `RpcConfig::authenticated_events_indexing`; authenticated event lookups now go through the existing `ledger_history_indexing` bitmap path, whose backfill already maintains the `EventStreamHead` bitmap dimension;
- the in-tree prototype `sui-light-client` is ported to the new APIs, with cursor-based pagination buffering events across pages until each containing checkpoint is fully covered before folding into the MMR. The on-chain MMR is one fold per checkpoint, so partial-checkpoint folds would never match the on-chain head;
- the e2e tests for authenticated events and object inclusion proofs are rewritten against the v2alpha surface, and the `corrupt_authenticated_event` fail-point moves into the new `render_event_chunk` path so the existing MMR-mismatch simtest keeps working.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-light-client/src/authenticated_events/stream.rs
```diff
@@ -30,95 +54,151 @@ impl EventStreamState {
...
+        // The on-chain MMR is folded one Merkle tree per checkpoint, so a
+        // partial-checkpoint fold would never match the chain head. When
+        // the scan is incomplete (`ItemLimit` / `ScanLimit`) we can only
+        // commit events at checkpoints strictly before the most recent
+        // one — the last checkpoint may still have more events on the next
+        // page. A final end reason guarantees every event at the last
+        // checkpoint has been emitted, so the entire buffer can flush.
+        let scan_incomplete = matches!(
+            end_reason,
+            Some(QueryEndReason::ItemLimit) | Some(QueryEndReason::ScanLimit),
```
nickvikeras, line 135: Do you think it is actually useful to differentiate these in the response or would a singular "ScanExhausted" be easier to deal with?

crates/sui-light-client/src/authenticated_events/stream.rs
```diff
@@ -230,24 +360,120 @@ impl EventStreamState {
...
+    let term = TransactionTerm::default().with_literals(vec![literal]);
+    TransactionFilter::default().with_terms(vec![term])
+}
+
+fn build_event_stream_head_filter(stream_id: SuiAddress) -> EventFilter {
+    let head_filter = EventStreamHeadFilter::default().with_stream_id(stream_id.to_string());
+    let predicate = EventPredicate::default().with_event_stream_head(head_filter);
+    let literal = EventLiteral::default().with_include(predicate);
+    let term = EventTerm::default().with_literals(vec![literal]);
+    EventFilter::default().with_terms(vec![term])
```
nickvikeras, line 476: We should probably add some fluent helpers to make this more intuitive to write. Something like

```
or(
  predicate()
      .event_stream_head(foo)
      .and().sender(bar), 
   not(predicate()...)
)
```

nickvikeras (commented):
Still in the middle of reading this, but need to head out for the night. Just wanted to leave a few thoughts before I log off.

crates/sui-light-client/src/authenticated_events/stream.rs
nickvikeras: I can take a stab at this and send out a separate PR.

crates/sui-core/src/rpc_index.rs
```diff
@@ -2846,7 +2568,7 @@ fn sparse_checkpoint_data_for_epoch_backfill(
         let full_transaction = CheckpointTransaction {
             transaction: tx.into(),
             effects: fx,
-            events: ev,
+            events: None,
```
alex-mysten, line 2571: intentionally removed?

crates/sui-e2e-tests/tests/authenticated_events_api_tests.rs
```diff
@@ -199,72 +170,280 @@ where
...
+        let mut request = ListTransactionsRequest::default()
+            .with_read_mask(read_mask.clone())
+            .with_start_checkpoint(start_checkpoint)
+            .with_filter(filter.clone())
+            .with_options(options);
+        if end_checkpoint_exclusive > start_checkpoint {
+            request = request.with_end_checkpoint(end_checkpoint_exclusive);
+        }
+
+        let mut response = client.list_transactions(request).await?.into_inner();
```
alex-mysten, line 223: Are we concerned about adding yet another RPC to this path (versus I suppose just returning some additional metadata alongside the event)

crates/sui-light-client/src/authenticated_events/mod.rs
```diff
@@ -538,27 +524,46 @@ impl AuthenticatedEventsClient {
...
+                    "Unknown proof variant".to_string(),
+                ));
+            }
         };
 
-        let object_data_bytes = response
-            .object_data
-            .as_ref()
-            .ok_or_else(|| ClientError::InternalError("Missing object data".to_string()))?;
+        let object_data_bytes = inclusion_proof.object_data.as_ref().ok_or_else(|| {
```
alex-mysten, line 561: Should we verify that the object_ref on the inclusion proof and matches the object data retunred by `get_checkpoint_object_proof()`? (this may have been a miss in my original implementation)

crates/sui-e2e-tests/tests/authenticated_events_api_tests.rs
bmwill: No and it's quite cheap in this instance, ideally this wouldn't be needed if we only did merging of events once per checkpoint

crates/sui-light-client/src/authenticated_events/stream.rs
```diff
@@ -230,24 +360,120 @@ impl EventStreamState {
...
+                    if let Some(c) = w.cursor.as_ref() {
+                        last_cursor = Some(c.to_vec());
+                    }
+                }
+                Some(list_transactions_response::Response::End(end)) => {
+                    end_reason = QueryEndReason::try_from(end.reason).ok();
+                }
+                Some(_) | None => {}
+            }
+        }
```
nickvikeras, line 446: Non blocking, but I think we should create helpers in the client that just return a `Stream<EventItem>`, `Stream<TransactionItem>`, etc that handle all the cursor management internally. Then you could write proof verification as a streaming merge pretty easily. That would naturally pipeline the fetches which would minimize lag and eventually we could swap in an implementation that transitions to the live stream.

nickvikeras (approved):

crates/sui-light-client/src/authenticated_events/stream.rs
bmwill: yep agreed, theres a bunch of ergonomic improvements we can make here.

PR #26763, opened 2026-05-22
fix(simulate): accept explicit gas_price=0 for gasless-eligible transactions

mystenmark:
## Summary

The gRPC \`SimulateTransaction\` API had two related bugs where valid gasless transactions were incorrectly routed to the priced flow, which then failed with \`GasPriceUnderRGP\` (gas_price=0 < RGP):

1. **Explicit price=0 (unresolved/proto path):** \`is_gasless_candidate\` checked \`price.is_some()\`, treating \`Some(0)\` the same as any other explicit non-zero price. Callers echoing back the \`gas_price=0\` suggestion from a prior simulate response were rejected.

   Fix: change to \`price.is_some_and(|p| p != 0)\` — explicit \`price=0\` is treated as equivalent to unset.

2. **Pre-built BCS gasless transaction:** A fully-formed BCS transaction with \`price=0, budget=0, payment=[]\` is already in gasless shape, but the BCS path always bypasses \`is_gasless_candidate\` (to avoid second-guessing the caller's explicit choices) and fell into the priced flow unconditionally.

   Fix: skip gas selection entirely when a BCS transaction is already in gasless shape (\`is_gasless_transaction()\` returns true), letting it fall through to direct simulation.

## Test plan

- \`simulate_explicit_gas_price_zero_accepted_for_gasless_eligible_tx\`: verifies the unresolved-path fix (fails before, passes after)
- \`simulate_bcs_gasless_transaction_accepted\`: verifies the BCS path fix (fails before, passes after)
- All existing gasless simulate tests pass unchanged (explicit non-zero price still opts out of gasless as before)

## Release notes

- [ ] Protocol:
- [x] gRPC: \`SimulateTransaction\` now accepts \`gas_price=0\` for gasless-tier-eligible transactions (both via the unresolved proto path and as pre-built BCS transactions) instead of returning an error.
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/grpc_simulate_gasless_tests.rs
+++ b/crates/sui-e2e-tests/tests/grpc_simulate_gasless_tests.rs
@@ -211,6 +211,68 @@ async fn simulate_respects_explicit_price_over_gasless() {
     assert!(returned.effects().status().success());
 }
 
+/// Explicitly setting gas_price=0 on a gasless-eligible tx must be accepted: the API should
+/// treat `price=0` the same as an unset price for purposes of gasless candidacy.
+#[sim_test]
+async fn simulate_explicit_gas_price_zero_accepted_for_gasless_eligible_tx() {
+    let mut test_env = setup_gasless_env().await;
+    let sender = test_env.get_sender(1);
+    let recipient = test_env.get_sender(2);
+    let (coin_type, coin_refs) =
+        setup_mintable_coin_env(&mut test_env, 0, &[(5_000, sender)]).await;
+    let tx = build_coin_send_funds_tx(sender, coin_refs[0], coin_type, recipient);
+
+    let response = connect(&test_env)
+        .await
+        .simulate_transaction(
+            SimulateTransactionRequest::new(to_unresolved_proto(tx, Some(0)))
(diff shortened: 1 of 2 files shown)


bmwill (approved):

aschran (approved):

crates/sui-e2e-tests/tests/grpc_simulate_gasless_tests.rs
```diff
@@ -129,6 +129,13 @@ fn build_balance_withdrawal_send_funds_tx(
     )
 }
 
+/// Turn a `TransactionData` into a BCS proto request. The transaction is already fully resolved
+/// (price=0, budget=0, payment=[]). The `bcs` field is kept so the simulate handler uses the BCS
+/// path rather than the resolver.
+fn to_bcs_proto(tx: TransactionData) -> Transaction {
+    Transaction::from(tx)
+}
+
```
bmwill: seems funny to need a function for this

bmwill (approved):

PR #26766, opened 2026-05-22
ci: restrict heavy CI token permissions

ebmifa:
## Summary

- Add top-level `permissions: contents: read` to the heavy PR CI workflows that use sccache:
  - `rust.yml`
  - `external.yml`
  - `bridge.yml`
- Leave the existing sccache AWS secret behavior unchanged.
- Do not add `id-token: write`; this keeps the OIDC migration as a separate, explicit follow-up.

## Why

These workflows run pull request code and only need repository read access for checkout/local actions. Explicit read-only `GITHUB_TOKEN` permissions reduce the blast radius from the repo default write token while preserving current CI behavior.

## Validation

- `ruby -e 'require "yaml"; ARGV.each { |path| YAML.load_file(path); puts "#{path}: ok" }' .github/workflows/rust.yml .github/workflows/external.yml .github/workflows/bridge.yml`
- `actionlint` was not installed in the local environment, so I used YAML parsing plus diff review for this narrow workflow-only change.

### DIFF
--- a/.github/workflows/bridge.yml
+++ b/.github/workflows/bridge.yml
@@ -18,6 +18,9 @@ on:
         required: false
         default: ''
 
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
--- a/.github/workflows/external.yml
+++ b/.github/workflows/external.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
 
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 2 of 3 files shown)


ebmifa:
Validation update after installing local actionlint:

- Installed `actionlint` 1.7.12 via Homebrew, along with `shellcheck`.
- Ran `actionlint` across the repo. It reports existing workflow/shellcheck findings unrelated to this PR.
- Ran scoped validation on the three edited workflows:

```bash
actionlint -shellcheck= .github/workflows/rust.yml .github/workflows/external.yml .github/workflows/bridge.yml
```

That only reports the pre-existing `rust.yml` `needs.diff.outputs.isSolidity` issue, which is already present on `origin/main`.

- Ran the scoped check again ignoring that known pre-existing finding:

```bash
actionlint -shellcheck= -ignore 'property "issolidity" is not defined' .github/workflows/rust.yml .github/workflows/external.yml .github/workflows/bridge.yml
```

This passes.

johnjmartin (approved):

PR #26768, opened 2026-05-22
[transaction-driver] Reconnect to validator on submit timeout

lxfind:
## Summary
- The transaction driver's per-validator gRPC clients (`AuthorityAggregator` → `NetworkAuthorityClient`) are lazy tonic channels created once per epoch. If a validator's connection silently goes dead (e.g. the peer process restarts mid-epoch without the transport seeing a TCP reset), the channel keeps routing requests onto the dead connection and never re-establishes one — every submit hangs until `SUBMIT_TRANSACTION_TIMEOUT`, retries reuse the same dead channel, and the request never reaches finality (times out after the full finality window → `Unavailable` / "timed-out before finality").
- `NetworkAuthorityClient` now holds a swappable channel (`ArcSwap`) plus a builder closure; added `AuthorityAPI::reconnect()` (no-op default, real impl for the network client) that drops the dead channel and lazily rebuilds it.
- The submitter calls `reconnect()` when a submit RPC times out, so the next attempt opens a fresh connection that routes to the live validator.
- `reconnect()` is rate-limited by a 2s cooldown so concurrent submit timeouts against the same validator don't thrash the channel (per review feedback).

## Context
Surfaced as a flaky failure of `sui-e2e-tests::discovery_tests::test::test_removed_validator_disconnected_after_reconfig` (~1/300 seeds). That test restarts all validators mid-epoch and then submits a `request_remove_validator` transaction through the fullnode; the fullnode's cached connections stayed pinned to the dead pre-restart validator instances, so submission black-holed for the full 90s finality window. Consensus and p2p state-sync remained healthy throughout — only the client→validator submission path was stuck.

## Test plan
- [x] `cargo xclippy` (sui-core) — clean
- [x] Repro: seed `1779465360124` failed reliably before the fix; passes after (19s vs 90s timeout)
- [x] Regression: `scripts/simtest/seed-search.py test_removed_validator_disconnected_after_reconfig` — 150/150 and 500/500 seeds pass
- [x] New unit test `test_reconnect_cooldown`: a reconnect rebuilds the channel, an immediate second reconnect is a no-op, and reconnect works again once the cooldown elapses
- [ ] CI

### DIFF
--- a/crates/sui-core/src/authority_client.rs
+++ b/crates/sui-core/src/authority_client.rs
@@ -3,10 +3,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use anyhow::anyhow;
+use arc_swap::ArcSwap;
 use async_trait::async_trait;
 use mysten_network::config::Config;
 use std::collections::BTreeMap;
 use std::net::SocketAddr;
+use std::sync::Arc;
 use std::time::Duration;
 use sui_network::{api::ValidatorClient, tonic};
 use sui_types::base_types::AuthorityName;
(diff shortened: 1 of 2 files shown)


mwtian (approved):

mwtian (approved):
Maybe add a 1-3s cooldown between reconnections.

lxfind:
Added a 2s cooldown between reconnections (per review feedback): `reconnect()` now no-ops if a reconnect happened within the last `RECONNECT_COOLDOWN`, so concurrent submit timeouts can't thrash the channel. Covered by a new unit test, and re-verified the motivating simtest passes.

PR #26784, opened 2026-05-26
feat(accumulator): add create_for_testing for AccumulatorRoot

mystenmark:
Adds a `#[test_only]` `create_for_testing` function to the `sui::accumulator` module, mirroring the equivalent function in `sui::random`. This allows test code to create an `AccumulatorRoot` object without going through the system-address-gated `create` function.


aschran (approved):

tnowacki (commented):
Could you add tests of using this in some meaningful way? I'm not really sure how it will work for reading accumulators and if we need additional support

PR #26789, opened 2026-05-26
[cli]: add more specific jsonrpc error code for external signers ledger fallback

jordanjennings-mysten:
## Description 

Add a separate error code `CREATE_KEY_UNSUPPORTED_USE_EXISTING_ERROR_CODE = -32013` out of the JSON RPC "server error" range to specify when a device cannot generate keys and should use an existing key. This is common for devices like ledger where key generation is done at device setup and keys are derived from a path (slip 10). Previously this was using a more generic code which may be confused in the future.

## Test plan 

unit tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


amnn (approved):
Seems reasonable!

PR #26794, opened 2026-05-27
Unify validity_check and validity_check_no_gas_check for dry_run

mystenmark:
## Summary

`TransactionDataAPI` previously exposed two validity-check methods:
- `validity_check` — full check including gas budget, price, and payment object requirements
- `validity_check_no_gas_check` — subset used by dry-run because gas payment could be empty

The split existed because dry-run was designed to accept transactions with no gas payment and inject a mock gas coin at runtime. However, empty `gas_data.payment` is now a first-class representation for address-balance gas transactions, not just a dry-run artefact. The old justification no longer applies.

### Changes

**`sui-types`** — `validity_check_no_gas_check` is removed from the `TransactionDataAPI` trait. It becomes a private method on `TransactionDataV1`, still called internally by `validity_check` as its final step.

**`sui-core`** — `simulate_transaction` (the shared path for dry-run and dev-inspect) is updated:
1. Input/receiving object kinds are computed before mock gas injection so the mock coin reference is not included in `input_object_kinds` (it continues to be added directly to `input_objects` after object loading).
2. The mock gas coin is injected as before—now before the validity check rather than after it.
3. `validity_check_no_gas_check` is replaced with `validity_check` using the epoch store's full context, so gas budget and price are validated on the same path as normal transaction submission.

**`sui-e2e-tests`** — Four new regression tests call `dry_run_transaction_block` via JSON-RPC and assert the current error behavior for invalid gas parameters (budget too low/high, price too low/high), ensuring future refactors don't silently change what clients observe.

## Test plan

- [ ] `cargo nextest run -p sui-e2e-tests --test dry_run_tests`
- [ ] `SUI_SKIP_SIMTESTS=1 cargo nextest run -p sui-core -p sui-types --lib`

### DIFF
--- a/crates/sui-e2e-tests/tests/dry_run_tests.rs
+++ b/crates/sui-e2e-tests/tests/dry_run_tests.rs
@@ -1,0 +1,114 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// These tests use the JSON-RPC dry_run_transaction_block endpoint to verify
+// the behavior that external clients observe when submitting transactions with
+// invalid gas parameters.
+#![allow(deprecated)]
+
+use sui_macros::sim_test;
+use sui_protocol_config::ProtocolConfig;
+use sui_types::{
+    base_types::SuiAddress, programmable_transaction_builder::ProgrammableTransactionBuilder,
+    transaction::TransactionData,
+};
+use test_cluster::TestClusterBuilder;
+
+async fn build_transfer_sui_tx(
+    sender: SuiAddress,
+    gas_budget: u64,
+    gas_price: u64,
+) -> TransactionData {
+    let mut builder = ProgrammableTransactionBuilder::new();
+    builder.transfer_sui(SuiAddress::random_for_testing_only(), None);
+    let pt = builder.finish();
+    // Empty gas payment — dry_run injects a mock gas coin.
+    TransactionData::new_programmable(sender, vec![], pt, gas_budget, gas_price)
+}
+
+#[sim_test]
+async fn test_dry_run_gas_budget_too_low() {
+    let cluster = TestClusterBuilder::new().build().await;
+    let client = cluster.sui_client();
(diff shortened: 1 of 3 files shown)


crates/sui-e2e-tests/tests/dry_run_tests.rs
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) Mysten Labs, Inc.
```
dariorussi, line 1: aren't those tests part of https://github.com/MystenLabs/sui/blob/main/crates/sui-core/src/unit_tests/gas_data_tests.rs#L620-L646 or we are testing something different?

mystenmark: these are testing that the errors propagate all the way down the RPC call correctly

aschran (approved):

tzakian (approved):
Looks good to me.

crates/sui-types/src/transaction.rs
```diff
@@ -3450,26 +3448,21 @@ impl TransactionDataAPI for TransactionDataV1 {
...
-                ));
+                )
+                .into());
             };
             pt.validate_gasless_transaction(config)?;
         }
 
-        self.check_sponsorship()
+        self.check_sponsorship()?;
+        Ok(())
```
dariorussi, line 3465: is this change expected? it seems we could do without?

dariorussi (approved):

crates/sui-types/src/transaction.rs
tzakian: I thought the same and did some digging -- it is actually needed since:
* `check_sponsorship` returns a `UserInputResult`
* `validity_check` returns a `SuiResult`

so there needs to be some casting, you could either do that with `?; Ok(())` like here or with `.map_err(|e| e.into())`.

mystenmark: Yes Tim has the answer

PR #26798, opened 2026-05-27
[move-core-types] Abstract underlying layout of type layouts

tzakian:
## Description 

The key things are:
1. The core structure in terms of usage is the ref-based layout
2. After discussions with @cgswords the type layouts are now parametrized by the data "backend" (i.e., a specific representation of the data). The type layouts provide a uniform interface for building/inspecting/using these layouts. There is a default backend (Arc-based) but there is also a box-based backend (but we could likewise go for an Arc-tree, or arena-based ones if we wanted to). 

I'm not 100% happy with the organization of the `backend` dir so would love any suggestions there -- I've tried a bunch of different possible options and disliked them all :'(

## Test plan 

Existing tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


tzakian:
Chatted with Cameron about this, will be making a couple changes before it'll be ready for review probably.

PR #26800, opened 2026-05-27
sui-replay-2: add --skip-artifacts flag

dariorussi:
Make every per-tx artifact write a no-op when `--skip-artifacts` is passed: the `.replay/<digest>/` directory is not created, JSON / zstd serialization is short-circuited, and `print_effects_or_fork` (which reads artifacts back from disk) is skipped in `main.rs`.

When profiling the replay loop with samply, artifact serialization otherwise dominates the CPU trace (~50% of samples in `libsystem_kernel::write`) and drowns out the Move-VM work we actually want to study. The flag's behavior is already documented in the sui-replay skill (see SKILL.md) and is required by the samply recipe there.

## Description 

## Test plan 

👀

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-replay-2/src/artifacts/manager.rs
+++ b/crates/sui-replay-2/src/artifacts/manager.rs
@@ -48,6 +48,11 @@ pub enum EncodingType {
 pub struct ArtifactManager<'a> {
     pub base_path: &'a Path,
     pub overrides_allowed: bool,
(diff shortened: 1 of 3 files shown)


awelc (approved):
LGTM

PR #26803, opened 2026-05-28
[move][move-decompiler] Inline immutable aliases to remove additional lets [1/8]

cgswords:
## Description

Rename the single-use binding inliner to `inline_immutable_alias` and generalize its logic to inline any `let x = y;` whose source `y` is never mutated, not only single-use ones. This eliminates more redundant `let` bindings for readability, while a mut-borrow guard keeps it from inlining a slot that can be written through a `&mut`.

## Test plan

Snapshot updates with new refinements.

---

## Release notes

For each box you select, include information after the relevant heading that describes the impact of your changes that a user
might notice and any actions they must take to implement updates.

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


tzakian (approved):
A.m.a.z.i.n.g!

PR #26807, opened 2026-05-28
[move][move-decompiler] Lift terminating arms out of let-if [5/8]

cgswords:
## Description

Add a refinement that hoists a terminating arm (return/abort/break/continue) out of a `let x = if (...)`, turning it into an early-exit guard followed by a plain binding. This exposes the "guard then compute" shape and feeds downstream cleanups.

## Test plan

Snapshot updates with new refinements.

---

## Release notes

For each box you select, include information after the relevant heading that describes the impact of your changes that a user
might notice and any actions they must take to implement updates.

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-decompiler/src/refinement/lift_terminating_arm.rs
```diff
@@ -0,0 +1,131 @@
...
+                    // the original `false` branch.
+                    Which::Else => {
+                        let mut negated = *cond;
+                        negate(&mut negated);
+                        Box::new(negated)
+                    }
+                };
+                items[i] = Exp::IfElse(guard_cond, Box::new(terminator), Box::new(None));
+                items.insert(i + 1, Exp::LetBind(target, Box::new(surviving)));
+                return true;
```
tzakian, line 67: Should this keep iterating in order to handle further lifting (in both `surviving` and `terminator`?) or is this handled by the refinement runner? 

e.g., to handle cases like

```
let x = if (c) { 
    let y = if (inner) { 
        abort 10 
    } else {
        10 
    }; 
    abort 11 
} else {
    10 
};
```

=> 

```
if (c) {
    let y = if (inner) {
       abort 10
    } else {
       10
    };
    abort 11
}
let x = 10;
```

and then to

```
if (c) { if (inner) { abort 10 }; let y = 10; abort 11}
let x = 10;
```

and then with the simplification (hopefully) to
```
if (c && inner) { abort 10 } else { let y = 10; abort 11 }
let x = 10;
```

tzakian (approved):
One small question which I think the answer is "is handled by the refinements runner" but just want to double check.

external-crates/move/crates/move-decompiler/src/refinement/lift_terminating_arm.rs
cgswords: Today the refinement runner handles this: we re-run refinements until we hit a fixpoint.

PR #26811, opened 2026-05-28
consensus_handler: emergency skip/dump + deny address-balance txns

alex-mysten:
## Summary
Two emergency knobs to recover from a stuck mainnet:

- **Skip user transactions in specific consensus commit rounds** — controlled by `HARDCODED_SKIPPED_CONSENSUS_COMMIT_ROUNDS` const + `SUI_SKIP_CONSENSUS_COMMIT_ROUNDS` env var (CSV of u64). When a commit round matches, every transaction in that commit is logged at info level (full Debug) and any user transaction is set to `ConsensusTxStatus::Dropped` and skipped; system messages still flow through so the prologue + checkpoint are produced and the chain keeps progressing.
- **Skip individual user transactions by digest** — controlled by `HARDCODED_SKIPPED_TRANSACTION_DIGESTS` const + `SUI_SKIP_TRANSACTION_DIGESTS` env var (CSV of base58 `TransactionDigest`). Same drop behavior, scoped to one transaction.
- **`address-balance-disabled` deny config** — new `TransactionDenyConfig` flag that rejects any transaction paying gas from an address balance OR containing a `CallArg::FundsWithdrawal` input. Broader kill switch than `gasless-disabled`.

All quorum validators must coordinate the same skip values or they will fork. A `warn!` is emitted at startup whenever either skip set is non-empty so the active configuration is visible in logs.

## Test plan
- [ ] `cargo nextest run -p sui-core --lib` (703 tests pass)
- [ ] `cargo xclippy` clean on sui-core, sui-config, sui-transaction-checks
- [ ] Manually verify on a testnet: set `SUI_SKIP_CONSENSUS_COMMIT_ROUNDS` and confirm matching commits drop user txns while system txns still produce a checkpoint
- [ ] Manually verify on a testnet: set `SUI_SKIP_TRANSACTION_DIGESTS` and confirm only the targeted digest is dropped
- [ ] Manually verify `address-balance-disabled: true` rejects both gas-paid-from-balance and FundsWithdrawal-input PTBs at signing time

### DIFF
--- a/crates/sui-config/src/transaction_deny_config.rs
+++ b/crates/sui-config/src/transaction_deny_config.rs
@@ -58,6 +58,19 @@ pub struct TransactionDenyConfig {
     #[serde(default)]
     gasless_disabled: bool,
 
+    /// Whether transactions that use address balances in any way are disabled.
+    /// Matches transactions that pay gas from an address balance, contain any
+    /// `FundsWithdrawal` input, or contain any coin reservation object ref.
+    #[serde(default)]
+    address_balance_disabled: bool,
+
+    /// Whether transactions with a non-target coin reservation in gas payment
+    /// are disabled. Matches transactions with any coin reservation at a
+    /// non-zero position in `gas_data.payment` (position 0 is the only target
+    /// slot; reservations elsewhere are the smash-into-Coin pattern).
+    #[serde(default)]
+    non_target_coin_reservation_disabled: bool,
+
     /// In-memory maps for faster lookup of various lists.
     #[serde(skip)]
     object_deny_set: OnceCell<HashSet<ObjectID>>,
@@ -130,6 +143,14 @@ impl TransactionDenyConfig {
         self.gasless_disabled
     }
 
+    pub fn address_balance_disabled(&self) -> bool {
+        self.address_balance_disabled
+    }
+
+    pub fn non_target_coin_reservation_disabled(&self) -> bool {
+        self.non_target_coin_reservation_disabled
+    }
+
     pub fn receiving_objects_disabled(&self) -> bool {
         self.receiving_objects_disabled
     }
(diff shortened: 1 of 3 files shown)


crates/sui-core/src/consensus_handler.rs
```diff
@@ -101,6 +103,104 @@ struct FilteredConsensusOutput {
     owned_object_locks: HashMap<ObjectRef, TransactionDigest>,
 }
 
+/// Hardcoded commit rounds for which all user transactions should be dropped.
+/// Merged with the contents of `SUI_SKIP_CONSENSUS_COMMIT_ROUNDS`. Edit this
+/// list when shipping an emergency binary so operators do not need to set the
+/// env var on every validator.
+const HARDCODED_SKIPPED_CONSENSUS_COMMIT_ROUNDS: &[u64] = &[824333];
```
akichidis: Let's make sure we confirm this before shipping

crates/sui-core/src/consensus_handler.rs
```diff
@@ -101,6 +103,104 @@ struct FilteredConsensusOutput {
...
+                        );
+                    }
+                }
+            }
+        }
+        if !digests.is_empty() {
+            warn!(
+                "Skipped transaction digests active: user transactions with digests {digests:?} will be dropped (hardcoded: {:?}, env: {:?})",
+                HARDCODED_SKIPPED_TRANSACTION_DIGESTS,
+                env::var("SUI_SKIP_TRANSACTION_DIGESTS").ok(),
```
akichidis, line 204: Maybe print here the `digests` instead (or too?)

crates/sui-config/src/transaction_deny_config.rs
```diff
@@ -171,6 +181,11 @@ impl TransactionDenyConfigBuilder {
         self
     }
 
+    pub fn disable_address_balance(mut self) -> Self {
+        self.config.address_balance_disabled = true;
```
mwtian, line 196: Default disable address balance?

alex-mysten: When omitted from node.yaml it defaults to false, this is a helper to turn on the disabling

PR #26822, opened 2026-05-28
[traffic_controller] Avoid cloning PolicyConfig on every check

wstran:
## Summary
`TrafficController::check` runs on the inbound-request critical path (every validator RPC goes through the traffic check). It clones the entire `PolicyConfig` — including its `allow_list: Option<Vec<String>>` and the `PolicyType` threshold configs — solely to read the single `dry_run: bool`. This reads the bool directly under the read lock instead, dropping a per-request heap clone and shortening the lock hold to a single field read.

Behavior is unchanged: `dry_run` is the only field of the cloned config that `check` used; the allow/block decision reads `self.acl`, not the config.

```diff
-        let policy_config = { self.policy_config.read().await.clone() };
+        let dry_run = self.policy_config.read().await.dry_run;

         let allowed = match &self.acl {
             ...
         };

-        match (allowed, policy_config.dry_run) {
+        match (allowed, dry_run) {
```

## Test plan
- `cargo nextest run -p sui-core traffic` — pass
- `cargo xclippy` clean
- `cargo fmt --all -- --check` clean

### DIFF
--- a/crates/sui-core/src/traffic_controller/mod.rs
+++ b/crates/sui-core/src/traffic_controller/mod.rs
@@ -342,7 +342,7 @@ impl TrafficController {
 
     /// Handle check with dry-run mode considered
     pub async fn check(&self, client: &Option<IpAddr>, proxied_client: &Option<IpAddr>) -> bool {
-        let policy_config = { self.policy_config.read().await.clone() };
+        let dry_run = self.policy_config.read().await.dry_run;
 
         let allowed = match &self.acl {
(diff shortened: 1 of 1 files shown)


wstran:
Closing this. On a closer look, in the common (blocklist/default) configuration `PolicyConfig` is a cheap stack copy of scalars/enums — it only heap-allocates when `allow_list = Some` (allowlist mode), and `check()` doesn't read that field there anyway. So removing the clone isn't a meaningful optimization on the hot path. Not worth the churn.

PR #26828, opened 2026-05-29
Land #26816 gas-underflow fix in main

mystenmark:
## Summary

Lands the address-balance gas-underflow fix from #26816 into `main`. #26816 went directly to `releases/sui-v1.72.0` as an out-of-band emergency fix; this brings it to `main`.

On an `InsufficientFundsForWithdraw` early abort, the gas payment's address-balance entries are pruned before smashing (real coins are kept). This avoids underflowing the already-drained address balance at settlement, which otherwise aborts the settlement transaction.

**This is the unconditional fix only.** The follow-up stacked PR adds the protocol-version + mainnet accumulator-version gating so the rollout is safe and mainnet replay stays bit-for-bit correct — kept separate so that gating diff reads cleanly against this fix.

## Test plan

- [x] `cargo check` / `xclippy` on `sui-adapter-latest`
- [x] e2e regression test `test_gas_smash_no_ab_underflow_on_iffw`
- [ ] CI (incl. simtests)

### DIFF
--- a/crates/sui-e2e-tests/tests/address_balance_compatibility_tests.rs
+++ b/crates/sui-e2e-tests/tests/address_balance_compatibility_tests.rs
@@ -1368,6 +1368,94 @@ async fn test_mix_coin_reservations_real_coins_and_shared_object() {
     test_env.cluster.trigger_reconfiguration().await;
 }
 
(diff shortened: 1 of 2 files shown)


tzakian (approved):

Aranmobinarann:
Don't be tired🙏

PR #26829, opened 2026-05-29
Gate #26816 gas-underflow fix for safe rollout (mainnet replay correctness)

mystenmark:
## Summary

Stacked on #26828 (the fix). This PR adds gating so the fix can land in `main` under proper protocol-versioning discipline while reproducing the out-of-band mainnet hotfix bit-for-bit. **The diff here shows exactly how the gating was applied to the pre-existing fix.**

> Review **#26828 first**; this PR's base is `mlogan-26816-fix`, so the diff below is gating-only.

## Activation logic

```
prune = IFFW && (
    prune_address_balance_gas_payment_on_iffw()      // protocol v126, all chains — forward rollout
    || accumulator_version >= 692949576              // mainnet-only backfill (None elsewhere)
)
```

- **All chains, protocol v126+:** the new feature flag enables the fix (clean, versioned rollout).
- **Mainnet only:** an accumulator-version backfill (ported from #26817) additionally applies the fix at/above the recovery accumulator root version, reproducing the mid-epoch hotfix that shipped *before* the flag existed. This is what makes mainnet replay correct — those transactions ran under a protocol version where the flag is off.
- **Testnet / devnet / fresh networks:** `accumulator_version` is `None`, so the backfill is inert and behavior is governed purely by the v126 flag.

`ExecutionOrEarlyError` becomes a struct carrying `(early_error, accumulator_version)`. The accumulator version is populated **only for mainnet committed execution** (`authority.rs`); `None` on every other chain and on non-committed paths (dev-inspect, simulation, genesis, replay tools).

## Notes for reviewers

- The constant `692949576` is the mainnet recovery accumulator root version, copied verbatim from #26817 — worth confirming it matches the actual recovery point.
- Offline replay tools (`sui-replay`, `sui-replay-2`) pass `None` (matching #26817), so they reproduce the fix only for v126+ transactions via the flag, not the pre-126 incident era. The bit-for-bit mainnet replay guarantee covers fullnode/validator **committed re-execution**.
- Assumes the incident was mainnet-only (no committed testnet history depends on the fix).

## Test plan

- [x] `cargo check` / `xclippy` clean across all touched crates
- [x] New unit tests for the mainnet accumulator gate (`address_balance_smash_gate_tests`)
- [x] e2e regression test runs at v126, exercising the flag path
- [ ] CI (incl. simtests)


sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -88,6 +88,41 @@ mod checked {
...
+        execution_params: &ExecutionOrEarlyError,
+        protocol_config: &ProtocolConfig,
+    ) -> bool {
+        if !matches!(
+            execution_params.early_error(),
+            Some(ExecutionErrorKind::InsufficientFundsForWithdraw)
+        ) {
+            return false;
+        }
+        protocol_config.prune_address_balance_gas_payment_on_iffw()
```
lxfind, line 122: I don't see why we actually need the protocol config gate here for this fix, assuming we are not shipping any new release to mainnet or testnet until this is fully resolved, we can just do the proper fix which will be gated by protocol config.

mystenmark: I just didn't want to check something in that isn't shippable without some other PR landing as well

lxfind: Well, this is shippable without the protocol config check on prune_address_balance_gas_payment_on_iffw, it won't cause any issues. It's just that it also won't fix anything for testnet. But I don't think that's a concern right now since we know we will merge another fix soon.

mystenmark: yes the protocol flag is mainly for testnet's benefit

lxfind (approved):

PR #26835, opened 2026-05-29
docs: Security-related docs updates and lagging tasks

jon-mysten:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -1,18 +1,32 @@
 ---
 title: Party Objects
-description: Party objects are owned by a specified party at the time of transfer and versioned by consensus. Learn how to create and access these objects.
+description: Party objects are owned by a specified party at the time of transfer and versioned by consensus. Learn how to create, transfer, and access these objects.
```
jessiemongeon1, line 3: Ask @aschran for review of this page

docs/content/develop/transaction-payment/gas-smashing.mdx
```diff
@@ -22,6 +22,16 @@ This design makes gas smashing a practical way to both manage coins and pay gas
 
 You can smash at most 256 coins in a single PTB. Transactions with more than 256 gas coins fail.
 
+## Gas smashing and address balance gas payment
```
jessiemongeon1, line 25: Ask @mystenmark for review on this page

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
```diff
@@ -0,0 +1,172 @@
+---
+title: TypeScript SDK PTB Template
+sidebar_label: PTB Template
+description: A minimal starter template for building a programmable transaction block with the Sui TypeScript SDK, with separate backend keypair and frontend wallet examples.
+keywords: [ typescript sdk, ptb, programmable transaction block, transaction, sui dapp kit ]
+---
+
+A [Programmable Transaction Block](/develop/transactions/ptbs/prog-txn-blocks) (PTB) groups several onchain actions into a single transaction that succeeds or fails as a unit. This page gives you a minimal template to start from, and points you to the full guide. The Sui TypeScript SDK documentation already covers transaction building in depth, so use this page to get moving, then follow the links for the complete reference.
```
jessiemongeon1, line 8: Ask @hayes-mysten for review on this page

docs/content/develop/transactions/soft-bundles.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+Because no first-party SDK exposes the method today, most teams integrate through an MEV service instead.
+
+:::info
+
+Sui is migrating data access from JSON-RPC to gRPC and GraphQL RPC by July 2026. The public full-node gRPC API does not include soft bundle execution. See [Sui and Community SDKs](/references/sui-sdks) for current SDK support.
+
+:::
+
+## Security considerations
```
jessiemongeon1, line 81: Verify this with security

docs/content/onchain-finance/asset-custody/address-balances/migrate-address-balances.mdx
```diff
@@ -127,9 +128,13 @@ commands:
 1: 0x2::transfer::public_transfer(result(0))
 ```
 
-### Legacy clients
+### Compatibility coin reservations for legacy clients
 
-A JSON-RPC compatibility layer presents synthetic coin objects representing address balance reservations. This preserves basic functionality for clients that cannot upgrade but should not be relied upon for new development.
+A JSON-RPC compatibility layer presents compatibility coin reservations to clients that cannot upgrade. A compatibility coin reservation is a synthetic object reference that reserves address balance funds for backward compatibility with clients that still expect coin object references. It is not a persisted `Coin<T>` object.
```
jessiemongeon1, line 133: Also confirm with @mystenmark

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+## Set up your project
+
+The Sui TypeScript SDK is published as an ESM only package, so a few project settings keep your imports working. Set `"type": "module"` in your `package.json`, and set `module` and `moduleResolution` in your `tsconfig.json` to a value that supports ESM, such as `NodeNext`. Avoid `commonjs`.
+
+When `tsconfig.json` uses `"module": "commonjs"`, VS Code and other editors auto-import from an internal build path instead of the package entry point. That path is not a supported entry point and can break at runtime. Compare the two:
+
+```ts
+// Wrong: auto-imported from an internal build path
+import { getFullnodeUrl, SuiClient } from '@mysten/sui/dist/cjs/client';
```
hayes-mysten: this seems odd thing to include in our most basic docs (and is a path that doesn't even exist anymore).  Is this an llm over-correction to stale training data or a common issue we actually see and want to cover?

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+// Correct: import from the package entry point
+import { getFullnodeUrl, SuiClient } from '@mysten/sui/client';
+```
+
+:::caution
+
+If an import path includes `dist/cjs`, your editor generated it from a `commonjs` configuration. Fix the path, and set `module` to `NodeNext` in your `tsconfig.json` so future imports resolve correctly.
```
hayes-mysten: dist/cjs does not exist, and this explanation doesn't really make sense

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+This page is a starting point, not the full guide. For complete coverage of transaction building, including inputs, commands, results, and execution, see [Sui Programmable Transaction Basics](https://sdk.mystenlabs.com/typescript/transaction-building/basics) in the TypeScript SDK documentation.
+
+:::
+
+## Set up your project
+
+The Sui TypeScript SDK is published as an ESM only package, so a few project settings keep your imports working. Set `"type": "module"` in your `package.json`, and set `module` and `moduleResolution` in your `tsconfig.json` to a value that supports ESM, such as `NodeNext`. Avoid `commonjs`.
+
+When `tsconfig.json` uses `"module": "commonjs"`, VS Code and other editors auto-import from an internal build path instead of the package entry point. That path is not a supported entry point and can break at runtime. Compare the two:
```
hayes-mysten: I don't think this is really accurate.  It can be true in some situations, but when using build systems or newer versions of node this can still work, but its a lot more nuanced.  Where are these docs coming from?

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+```ts title='buildTransfer.ts'
+import { coinWithBalance, Transaction } from '@mysten/sui/transactions';
+import { MIST_PER_SUI } from '@mysten/sui/utils';
+
+// Build a transaction that sends 0.1 SUI to a recipient.
+export function buildTransfer(recipient: string) {
+  const tx = new Transaction();
+  tx.transferObjects(
+    [coinWithBalance({ balance: BigInt(0.1 * Number(MIST_PER_SUI)) })],
```
hayes-mysten: double casting here is weird.   coinWithBalance can take a number and cast internally, or we can just provide the value in MIST directly as a BigInt

hayes-mysten: we should probably also start using tx.coin or tx.balance instead of coinWithBalance (these are newer methods that internally work the same, but don't require an extra import)

hayes-mysten: We should also figure out if we want to have our default examples continue to send coins instead of address balances

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+For a frontend app, scaffold a new project with the Sui dApp Kit starter. It sets up a working `tsconfig.json` and project structure for you:
+
+```bash
+$ npm create @mysten/dapp
+```
+
+## Build the transaction
+
+Build the PTB once in a reusable function, then execute it from either a backend script or a frontend wallet. This pattern keeps transaction construction separate from execution, which matters for wallet compatibility because the construction step must stay synchronous when a wallet popup is involved.
```
hayes-mysten: I am not sure that splitting this out in our most basic example really makes sense.  This is over-complicating something very simple and feels unnecessarily prescriptive

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
```diff
@@ -0,0 +1,172 @@
...
+Always check the `$kind` field on the result. A transaction can be submitted and still fail onchain, so treat a `FailedTransaction` result as an error rather than assuming success.
+
+## Execute from a frontend wallet
+
+A frontend app does not hold the user's key. The user's wallet signs the transaction through Sui dApp Kit, so your code only builds the transaction and hands it to the wallet.
+
+Install Sui dApp Kit alongside the core SDK:
+
+```bash
+$ npm install @mysten/dapp-kit
```
hayes-mysten: this is the old deprecated package, we should not be telling people to use this anymore

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
```diff
@@ -0,0 +1,172 @@
...
+  return (
+    <div>
+      <ConnectButton />
+      {currentAccount && (
+        <button
+          onClick={() => {
+            signAndExecuteTransaction(
+              {
+                transaction: buildTransfer(currentAccount.address),
+                chain: 'sui:devnet',
```
hayes-mysten: we should probably not be passing chain here 99% of the time

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
hayes-mysten: I guess it makes it easier to write the rest of the docs, but I think it is still cleaner to add the 3 lines to construct the transaction in each example

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
zhu-jon: definitely an over-correction, thanks for catching !

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
zhu-jon: makes sense, fixed !

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
zhu-jon: fixed to tx.coin. I guess going forward do you have a preference?

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -1,18 +1,32 @@
...
 ---
 
 A party object is a consensus-address-owned object: a single address owns it, but consensus sequences and versions it. APIs expose this ownership with a `ConsensusAddressOwner` owner variant. Like shared objects, party objects use consensus sequencing. Unlike shared objects, they retain single-address ownership and can be transferred to and from other ownership types and wrapped.
```
aschran: I think we shouldn't explain it like this because this is just a current limitation

the intention was always for party objects to support multiple forms of ownership/permissions

and we are already working on adding another `Party` ownership type in addition to `ConsensusAddressOwner` to implement this, so soon two owner variants will both correpsond to party objects

aschran: (the "info" block below is correct and I think good to keep, but for the actual main-text description of party objects I think we should describe the actual mental model and not just the current implementation)

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,131 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
+
+The `sui::party` module also provides two convenience macros that wrap the `transfer` functions. The `sui::party::transfer` macro calls `sui::transfer::party_transfer`, and the `sui::party::public_transfer` macro calls `sui::transfer::public_party_transfer`. Use whichever form reads more clearly in your module.
+
+The following example creates an object and transfers it to a single-owner party in one function:
 
 <ImportContent source="examples/move/basics/sources/object_basics.move" mode="code" fun="create_party" noComments />
 
+A party object's ownership can change over its lifetime, either by adding it as a [dynamic object field](/develop/objects/dynamic-fields), transferring it to a different address or ownership type, or making it [immutable](/develop/objects/object-ownership/immutable). However, after you create an object and set its ownership, you cannot [share](/develop/objects/object-ownership/shared) it.
+
+### Party permission model
```
aschran: I think this is no longer current 

@tnowacki can you advise?

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,131 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
+
+Keep the following constraints in mind when you design with party objects:
+
+- **Single-owner only:** The public transfer functions accept only single-owner parties. Multi-member parties are modeled in the type but are rejected at transfer time with `EInvalidPartyPermissions`.
+- **No transfer to object for object-ID owners:** A party object whose owning address corresponds to an object ID cannot be received through the [transfer to object](/develop/objects/transfers/transfer-to-object) mechanism.
+- **No gas payment with party `Coin<SUI>`:** You cannot pay gas with a party object `Coin<SUI>`. Transfer it back to address-owned first.
+- **Cannot be shared after creation:** Once you set an object's ownership, you cannot [share](/develop/objects/object-ownership/shared) it. You can still transfer it to another ownership type, wrap it, or make it [immutable](/develop/objects/object-ownership/immutable).
+
+:::caution
+
```
aschran: I think they are now supported on every network? unless we need to worry about non-public networks (but in that case tons of features could theoretically be re-disabled that are already enabled on mainnet, so I'd expect we don't worry about describing that possibility in our docs)

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
hayes-mysten: I don't really know.  I think long term most transfers should be address balances.  Thats what will work with free tier and should reduce gas costs, short term I am not sure if we are ready to push people that way since some wallets still have issues with address balances

jessiemongeon1:
## Style Guide Audit

Audited **7** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

**2** violation(s) found. All must be fixed before merge.

### `docs/content/develop/transactions/soft-bundles.mdx` (1 violation(s))

_1 violation(s) (0 regex, 1 claude)_

- **Line 69** — passive voice
  - Current: `the transaction was submitted to consensus, was already executed, or was rejected`
  - Fix: `the validator submitted the transaction to consensus, already executed it, or rejected it`

### `docs/content/onchain-finance/asset-custody/address-balances/migrate-address-balances.mdx` (1 violation(s))

- **Line 34** — Use present tense, not future
  - Current: `will`
  - Fix: `Use present tense verb`

---

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -1,18 +1,32 @@
...
-description: Party objects are owned by a specified party at the time of transfer and versioned by consensus. Learn how to create and access these objects.
+description: Party objects are owned by a specified party at the time of transfer and versioned by consensus. Learn how to create, transfer, and access these objects.
 ---
 
-A party object is a consensus-address-owned object: a single address owns it, but consensus sequences and versions it. APIs expose this ownership with a `ConsensusAddressOwner` owner variant. Like shared objects, party objects use consensus sequencing. Unlike shared objects, they retain single-address ownership and can be transferred to and from other ownership types and wrapped.
+A party object is owned by a `Party`: a permission set that determines which addresses can access the object and how. Like [shared](/develop/objects/object-ownership/shared) objects, party objects are sequenced and versioned by [consensus](/develop/sui-architecture/consensus), so several transactions can use them concurrently. Unlike shared objects, only the owning party can access a party object, and you can transfer it to and from other ownership types and wrap it.
```
tnowacki:
```suggestion
A party object is owned by a `Party`: a permission set that determines which addresses can access the object and how. Like [shared](/develop/objects/object-ownership/shared) objects, party objects are sequenced and versioned by [consensus](/develop/sui-architecture/consensus), so several transactions can submitted concurrently. Unlike shared objects, only the addresses with the appropriate permissions can access a party object. These permissions govern all aspects of an object, including reads, writes, and transfers.
```

On the first change, the consensus sequenced objects can have multiple transactions submitted concurrently, but for transactions that mutate the object, they are executed sequentially. Though immutable usage can be done concurrently.

On the second change, I would not say that the "owning party" is really an entity in this way. We have a mapping from address -> permission. And then a default permission set for if the address is not in that map. My point being you could not be in the set, and actually have more permissions than an address explicitly specified.

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -1,18 +1,32 @@
...
 
-Party objects are transferred using the [`sui::transfer::party_transfer`](/references/framework/sui_sui/transfer#sui_transfer_party_transfer) or [`sui::transfer::public_party_transfer`](/references/framework/sui_sui/transfer#sui_transfer_public_party_transfer) function. It is accessible to the `Party` to which it is transferred.
+Party objects are transferred using the [`sui::transfer::party_transfer`](/references/framework/sui_sui/transfer#sui_transfer_party_transfer) or [`sui::transfer::public_party_transfer`](/references/framework/sui_sui/transfer#sui_transfer_public_party_transfer) function. The object is accessible to the `Party` to which it is transferred.
```
tnowacki:
```suggestion
Party objects are transferred using the [`sui::transfer::party_transfer`](/references/framework/sui_sui/transfer#sui_transfer_party_transfer) or [`sui::transfer::public_party_transfer`](/references/framework/sui_sui/transfer#sui_transfer_public_party_transfer) function. The object's permissions are then defined by the `Party` to which it is transferred.
```

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -1,18 +1,32 @@
...
 
 :::info
 
-Party objects currently support only single-address ownership. The `sui::party::Party` type already models permissions for multiple members, but the public ownership mode is restricted to `party::single_owner`.
+Party objects currently support only single-address ownership. The `sui::party::Party` type already models permissions for multiple members, but the public transfer functions accept only a single-owner party. A transfer of a party with more than one member aborts with `EInvalidPartyPermissions`.
```
tnowacki: I would revert this change. Only a single owner is constructible at this point, so there is really no need to talk about about error states.

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -1,18 +1,32 @@
...
 
 :::
 
+## Party, address-owned, and shared objects compared
```
tnowacki, line 16: I do not fully understand the goal of this section so it is a bit hard for me to give feed back on. Why just these properties here? I don't feel all of them important to the end user, or at least the same set of end users.

Additionally I don't think it really highlights the permission granularity of party objects, which is sort of half the reason we built them (with the consensus versioning being the other half)

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,124 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
 public fun public_party_transfer<T: key + store>(obj: T, party: sui::party::Party)
 ```
 
-- Use the [`sui::transfer::party_transfer`](/references/framework/sui_sui/transfer#sui_transfer_party_transfer) function if you are defining a [custom transfer policy](/develop/objects/transfers/custom-rules) for the object. 
+- Use the [`sui::transfer::party_transfer`](/references/framework/sui_sui/transfer#sui_transfer_party_transfer) function if you are defining a [custom transfer policy](/develop/objects/transfers/custom-rules) for the object.
```
tnowacki: This function should not be used at all today since you can only use the single owner party, which has public transfer. 

In either case for both this and the public_ variant, the behavior will instead be governed by a permission.

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,124 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
 
 - Use the [`sui::transfer::public_party_transfer`](/references/framework/sui_sui/transfer#sui_transfer_public_party_transfer) function if the object has the `store` capability.
 
-A party object's ownership can change over its lifetime, either by adding it as a [dynamic object field](/develop/objects/dynamic-fields), transferring it to a different address or ownership type, or making it [immutable](/develop/objects/object-ownership/immutable). However, after you create an object and set its ownership, it cannot be [shared](/develop/objects/object-ownership/shared).
+Both functions take a `sui::party::Party` value. Construct that value with [`sui::party::single_owner`](https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/packages/sui-framework/sources/party.move), which grants all permissions to one address and no permissions to anyone else:
```
tnowacki:
```suggestion
Both functions take a `sui::party::Party` value. Currently, that value can be constructed only with [`sui::party::single_owner`](https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/packages/sui-framework/sources/party.move), which grants all permissions to one address and no permissions to anyone else:
```

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,124 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
+
+```
+public fun single_owner(owner: address): sui::party::Party
+```
+
+The `sui::party` module also provides two convenience macros that wrap the `transfer` functions. The `sui::party::transfer` macro calls `sui::transfer::party_transfer`, and the `sui::party::public_transfer` macro calls `sui::transfer::public_party_transfer`. Use whichever form reads more clearly in your module.
```
tnowacki:
```suggestion
The `sui::party` module also provides wrapper macros for the `transfer` functions that can be used with method syntax, e.g. `my_party.public_transfer()`. The `sui::party::transfer` macro calls `sui::transfer::party_transfer`, and the `sui::party::public_transfer` macro calls `sui::transfer::public_party_transfer`.
```

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,124 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
+
+The following example creates an object and transfers it to a single-owner party in one function:
 
 <ImportContent source="examples/move/basics/sources/object_basics.move" mode="code" fun="create_party" noComments />
 
+A party object's ownership can change over its lifetime, either by adding it as a [dynamic object field](/develop/objects/dynamic-fields), transferring it to a different address or ownership type, or making it [immutable](/develop/objects/object-ownership/immutable). However, after you create an object and set its ownership, you cannot [share](/develop/objects/object-ownership/shared) it.
```
tnowacki: This list is not exhaustive/complete, so I'm not sure it is worth adding. Party objects are just like owned objects in terms of restrictions. And trying to enumerate that separately here feels like it will only cause confusion.

docs/content/develop/objects/object-ownership/party.mdx
tnowacki: Yeah this is not current. But it is also not accessible so I'm not sure it is worth adding.
There is a good amount of detail documenting how it will work here:
https://github.com/MystenLabs/sui/blob/main/crates/sui-types/src/object.rs#L497

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,124 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
 
-Party objects can be used by multiple inflight [transactions](/develop/transactions/txn-overview) at the same time. This contrasts with [address-owned](/develop/objects/object-ownership/address-owned) objects, which only allow a single inflight transaction. Many applications can benefit from the ability to pipeline multiple transactions on the same party object.
+Party objects can be used by multiple inflight [transactions](/develop/transactions/txn-overview) at the same time. This contrasts with [address-owned](/develop/objects/object-ownership/address-owned) objects, which allow only a single inflight transaction because they take a fast path object lock. Because consensus assigns each version, many applications can pipeline several transactions against the same party object without waiting for one to finalize before submitting the next.
```
tnowacki: I'm not sure the description of locks is really accurate anymore. I might just say that that the versioning is managed by the owner and leave it that.

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,124 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
 
-<ImportContent source="examples/move/basics/sources/object_basics.move" mode="code" fun="party_transfer_single_owner" noComments />
\ No newline at end of file
+<ImportContent source="examples/move/basics/sources/object_basics.move" mode="code" fun="party_transfer_single_owner" noComments />
+
+## Limitations
+
+Keep the following constraints in mind when you design with party objects:
+
+- **Single-owner only:** The public transfer functions accept only single-owner parties. Multi-member parties are modeled in the type but are rejected at transfer time with `EInvalidPartyPermissions`.
```
tnowacki: - The abort is not reachable at this time
- We will lift this restriction, which I think is worth noting

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,124 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
+- **No transfer to object for object-ID owners:** A party object whose owning address corresponds to an object ID cannot be received through the [transfer to object](/develop/objects/transfers/transfer-to-object) mechanism.
```
tnowacki: This does not make a ton of sense to me as written, but I think I understand what we are trying to get at? 
Assuming I do, I might say something more like 
Even if the party is a single owner, and that address corresponds to an object's ID, the party object cannot currently be received through `sui::transfer::Receiving`. 

And we might find some way of working through this in the future... maybe... I don't know.

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,124 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
+- **No gas payment with party `Coin<SUI>`:** You cannot pay gas with a party object `Coin<SUI>`. Transfer it back to address-owned first.
+- **Cannot be shared after creation:** Once you set an object's ownership, you cannot [share](/develop/objects/object-ownership/shared) it. You can still transfer it to another ownership type, wrap it, or make it [immutable](/develop/objects/object-ownership/immutable).
```
tnowacki: I would cut this. Shared objects by definition have to be newly created. This restriction is completely orthogonal to party objects.

docs/content/develop/objects/object-ownership/party.mdx
```diff
@@ -22,30 +36,124 @@ public fun party_transfer<T: key>(obj: T, party: sui::party::Party)
...
+
+## Troubleshooting
+
+Use the following guidance when a transaction that uses party objects fails.
+
+- **Invalid party permissions:** A transfer aborts with `EInvalidPartyPermissions` when the party passed to `party_transfer` or `public_party_transfer` has more than one member or does not grant all permissions to a single owner. Build the party with `sui::party::single_owner` so it has exactly one owner with full permissions.
```
tnowacki: This is not reachable today

docs/content/develop/transaction-payment/gas-smashing.mdx
```diff
@@ -22,6 +22,16 @@ This design makes gas smashing a practical way to both manage coins and pay gas
...
+
+Gas smashing applies only to real `Coin<SUI>` objects that you supply as gas payment. When you provide multiple `Coin<SUI>` objects, the network merges them into a single coin and deducts the gas fee from that coin.
+
+Address balance gas payment is different. When you pay gas from an address balance, you do not select or provide any coin objects. The network resolves the gas fee as a withdrawal from your reserved address balance funds, so there is no gas coin selection or merge workflow, and gas smashing is not needed.
+
+The compatibility layer can present compatibility coin reservations to older clients. A compatibility coin reservation is a synthetic object reference that reserves address balance funds, not a persisted `Coin<SUI>` object. Gas smashing does not merge these reservations, because the network resolves an address balance reservation as a withdrawal instead of smashing it into a coin.
```
tnowacki: This is not true. The synthetic coins are smashed. And it is all based on which payment method is listed first.
- If it is a real coin, they are smashed into the real coin (coin reservations included, where the reservation is fully withdrawn)
- If it is a coin reservation, all real coins are smashed and sent to the address balance.

docs/content/develop/transaction-payment/gas-smashing.mdx
```diff
@@ -22,6 +22,16 @@ This design makes gas smashing a practical way to both manage coins and pay gas
...
+
+To pay gas from an address balance, see [Paying for gas from address balances](/onchain-finance/asset-custody/address-balances/using-address-balances#paying-for-gas-from-address-balances). For details on the compatibility layer, see [Compatibility coin reservations for legacy clients](/onchain-finance/asset-custody/address-balances/migrate-address-balances#compatibility-coin-reservations-for-legacy-clients).
```
tnowacki, line 41: It is worth noting that you can use the compatibility layer for non-compatibility layer reasons. For example, when using address balances and if you want to have some extra SUI to spend within the PTB you have two options:
1. Add additional funds withdrawals as input arguments
2. Use a coin reservation with a reservation larger than the gas budget. The excess amount will be available with the `GasCoin` in the PTB, just like is the case with `Coin` based gas payments. (This can be helpful with PTB templates and similar situations)

docs/content/develop/transactions/ptbs/ts-sdk-ptb-template.mdx
tnowacki: cc @manolisliolios @damirka

docs/content/onchain-finance/asset-custody/address-balances/migrate-address-balances.mdx
```diff
@@ -127,9 +128,13 @@ commands:
...
+
+These reservations exist only to preserve basic functionality for older clients that expect coin object references. Do not treat them as normal persisted coin objects, and do not use them as a model for new integrations. For new development, use address balances directly, as described in [Using Address Balances](/onchain-finance/asset-custody/address-balances/using-address-balances).
```
tnowacki:
```suggestion
These reservations primarily to preserve basic functionality for older clients that expect coin object references. Do not treat them as normal persisted coin objects. For new development, use address balances directly, as described in [Using Address Balances](/onchain-finance/asset-custody/address-balances/using-address-balances).
```

docs/content/onchain-finance/asset-custody/address-balances/migrate-address-balances.mdx
```diff
@@ -127,9 +128,13 @@ commands:
...
+
+When a legacy client supplies a compatibility coin reservation as gas payment, the network resolves it as a withdrawal from the address balance rather than smashing it into a coin. Gas smashing applies only to real `Coin<SUI>` objects supplied as gas payment. For more information, see [Gas Smashing](/develop/transaction-payment/gas-smashing).
```
tnowacki:
```suggestion
When a legacy client supplies a compatibility coin reservation as gas payment, the network resolves it as a withdrawal from the address balance rather than smashing it into a coin. And gas smashing applies depending on the position of the coin reservation in the list of gas payments. For more information, see [Gas Smashing](/develop/transaction-payment/gas-smashing).
```

jessiemongeon1 (approved):

PR #26838, opened 2026-05-29
feat(genesis): stamp genesis coins with storage rebates (protocol v127)

lxfind:
## Problem

Genesis runs the executor **unmetered**, so every genesis gas coin is minted with `storage_rebate = 0` and the storage fund starts empty — unlike coins written by ordinary transactions, which always carry the storage rebate they were charged. This is a long-standing test foot-gun: because virtually all tests are built on genesis coins, they **never exercise gas/settlement paths that depend on a coin having a non-zero storage rebate**, which hides bugs in those paths (the recent address-balance gas-smash conservation hole is one example — it only bites rebate-bearing coins, so zero-rebate genesis coins masked it).

## Change

Behind a new `genesis_objects_have_storage_rebate` protocol feature flag (added at **v127**, enabled on **all chains**):

- The genesis executor stamps each genesis gas coin with the storage rebate it would have been charged under normal metered execution (`size * storage_per_byte_cost * storage_gas_price`, matching `gas_v2`), and
- seeds the storage fund's `total_object_storage_rebates` with the matching SUI,

so the conservation imbalance (`storage_fund - sum(storage_rebate)`) stays **zero**. Coin balances are unchanged — only the rebate attribution moves into the fund.

### Why these specific seams

- **It has to be in the executor, not the genesis builder.** Nodes re-execute the genesis transaction at startup (`sui-node/src/lib.rs`), and its effects must match the genesis blob. The builder can't diverge from what the executor produces.
- **Stamping runs *after* gas charging.** The unmetered `collect_storage_and_rebate` (inside `charge_gas`) resets every object's rebate to 0, so stamping in the `Genesis` match arm gets wiped. Genesis is exempt from the per-tx conservation check, so doing it post-charge is safe. The fund seeding reuses `get_sui_system_state_wrapper` + `get_dynamic_field_object_from_store` (the same path as `advance_epoch_safe_mode`).

### Why it's automatic and safe

- **Automatic**: gated purely on a protocol *version*, not a per-test flag, so tests and local networks get it with zero action.
- **Safe for production**: existing mainnet/testnet pin their genesis to their **launch protocol version** (far below v127), and genesis re-execution always uses that pinned version, so their genesis is **never** stamped. The flag only ever applies to a network whose genesis is *built* at v127+ (devnet/local/tests today).
- **Why not chain-gated**: it deliberately is *not* restricted to non-production chains. Genesis is built *before* the chain ID is known (the chain ID is derived from the genesis digest), so the builder runs at `Chain::Unknown` while a node's genesis re-execution runs at the real chain. A chain-*dependent* genesis option would therefore make the two disagree and **fork** at the genesis checkpoint. Gating on version alone keeps build and re-execution identical.

## Tests

- New `genesis_coins_have_storage_rebate` test asserts genesis gas coins carry a rebate and the fund equals the sum of object rebates (zero imbalance).
- Protocol-config v127 snapshots (flag present for all chains) + swarm-config genesis snapshots regenerated.
- Indexer-alt e2e snapshots regenerated (effects/balance changes now carry the genesis rebate).
- `/protocol-config` verified: new version (released max is 125), version-history comment added, version-only gate (no chain guard).

## Release notes

- [x] Protocol: Adds protocol version 127 with the `genesis_objects_have_storage_rebate` feature flag (enabled on all chains). At this version, genesis stamps gas coins with a realistic storage rebate and seeds the storage fund to match. Existing mainnet/testnet are unaffected — their genesis is pinned to their launch protocol version, far below 127, so it is never stamped.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -670,6 +670,15 @@ mod checked {
         // to the 0x5 object so that it's not lost.
         temporary_store.conserve_unmetered_storage_rebate(gas_charger.unmetered_storage_rebate());
 
+        // Genesis runs unmetered, so `collect_storage_and_rebate` (above, in `charge_gas`) just set
+        // every created object's storage rebate to 0. When enabled, stamp genesis gas coins with a
+        // realistic storage rebate and seed the storage fund to match, so genesis coins behave like
+        // coins written by ordinary transactions. Done here (after gas charging, and genesis is
+        // exempt from the conservation checks below) so the rebates aren't immediately overwritten.
+        if is_genesis_tx && protocol_config.genesis_objects_have_storage_rebate() {
```
alex-mysten, line 678: Should we make the gating more evident by excluding mainnet and testnet genesis tx digests?

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -694,6 +695,69 @@ impl<'backing> TemporaryStore<'backing> {
...
+    /// Stamp each genesis gas coin with the storage rebate it would have been charged under normal
+    /// metered execution (`size * storage_per_byte_cost * storage_gas_price`, matching `gas_v2`),
+    /// and seed the storage fund's `total_object_storage_rebates` with the matching SUI so the
+    /// conservation imbalance (`storage_fund - sum(storage_rebate)`) stays zero. Genesis otherwise
+    /// runs unmetered, leaving every genesis coin with a zero storage rebate -- unlike coins written
+    /// by ordinary transactions -- which prevents tests from ever exercising rebate-dependent
+    /// gas/settlement paths. Must be called after gas charging (`collect_storage_and_rebate` would
+    /// otherwise overwrite the rebates back to 0). Genesis always uses the V1 system state layout,
+    /// so any failure here is a genesis bug and panics rather than silently breaking conservation.
+    pub fn seed_genesis_storage_rebates(&mut self) {
```
alex-mysten, line 707: A TestCluster-only implementation seems less intrusive. Could we start with that and then adjust other test infra as needed?

lxfind: We could, but how would we adjust other test infra to achieve this? From what I can tell, doing it during genesis has the broadest lift

mystenmark:
Are there still genesis coins live on production networks? If so, we can enter either the storage_rebate == 0 and the storage_rebase != 0 path. If we introduce a bug on the == 0 path, it could lay dormant for a long time until a genesis coin is used.

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4993,6 +5013,16 @@ impl ProtocolConfig {
...
                 }
+                127 => {
+                    // Genesis stamps gas coins with a realistic storage rebate and seeds the
+                    // storage fund to match, so tests exercise rebate-dependent gas/settlement
+                    // paths. Enabled on all chains, not chain-gated: genesis is built before the
+                    // chain ID is known (it is derived from the genesis digest), so a chain-
+                    // dependent genesis option forks (the builder uses Chain::Unknown while node
+                    // re-execution uses the real chain). Existing mainnet/testnet are untouched --
+                    // their genesis is pinned to their launch protocol version, far below 127.
+                    cfg.feature_flags.genesis_objects_have_storage_rebate = true;
```
mystenmark, line 5024: one way to address the concern I raised earlier is to only enable this for unknown chains. Then when we run tests with the mainnet config override, tests will run with 0 storage rebate, and when they run without, they'll use non-zero rebate

PR #26842, opened 2026-05-29
[Block Streaming] consensus sync full nodes to verify local checkpoints

akichidis:
## Description 

Currently full nodes by default are executing transactions from synced (verified) checkpoints exclusively https://github.com/MystenLabs/sui/blob/027024d7f6a91a4cf5701b7e2f17d837cd6d09c3/crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs#L384

For full nodes that have block streaming enabled (consensus sync) we should attempt to use the code path that validators follow and verify locally (verified) built checkpoints instead.

This PR is:
* allowing the consensus sync full nodes (that ones that run consensus) to attempt verify locally built checkpoints
* matching the existing logic that populates that legacy json indexes and the rpc indexes
* populating the full data to the checkpoint state (so the subscribers can successfully receive them)

## Test plan 

CI/Simtests/PT

**Note:** building an Antithesis setup to use an consensus sync full node instead of a state sync only. Will let it run for a few hours to confirm the behaviour.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/tests/simtest.rs
+++ b/crates/sui-benchmark/tests/simtest.rs
@@ -2156,6 +2156,7 @@ mod test {
     #[sim_test(config = "test_config()")]
     async fn test_network_with_observer_node() {
         use consensus_config::{NetworkPublicKey, ObserverParameters, PeerRecord};
+        use sui_benchmark::workloads::composite::*;
         use sui_types::crypto::KeypairTraits;
 
         sui_protocol_config::ProtocolConfig::poison_get_for_min_version();
@@ -2258,26 +2259,44 @@ mod test {
         // Let the Observer node stabilize
         tokio::time::sleep(Duration::from_secs(5)).await;
 
-        info!("Running network for a period to verify Observer node doesn't crash");
+        info!("Running object-funds workload to verify Observer node doesn't crash");
 
-        // Let the network run for a while to ensure the Observer node is stable
-        // Submit a few transactions to generate some network activity
-        let sender = test_cluster.get_address_0();
-        let rgp = test_cluster.get_reference_gas_price().await;
-
-        for i in 0..5 {
-            info!("Submitting transaction {}", i);
-            // Fund an address to generate some transaction activity
-            let _ = test_cluster
-                .fund_address_and_return_gas(rgp, None, sender)
-                .await;
-            tokio::time::sleep(Duration::from_secs(2)).await;
+        let composite_metrics = Arc::new(Mutex::new(CompositionMetrics::new()));
+        let composite_config = CompositeWorkloadConfig {
+            num_shared_counters: 1,
+            address_balance_amount: 1000,
+            address_balance_gas_probability: 0.0,
+            mixed_gas_payment_probability: 0.0,
+            conflicting_transaction_probability: 0.0,
+            alias_tx_probability: 0.0,
+            metrics: Some(composite_metrics.clone()),
+            ..Default::default()
         }
+        .with_probability(ObjectBalanceDeposit::NAME, 0.1)
+        .with_probability(ObjectBalanceWithdraw::NAME, 0.3)
+        .with_probability(ObjectBalanceOverdraw::NAME, 1.0)
+        .with_probability(AccumulatorBalanceRead::NAME, 0.2);
 
-        info!("Waiting for the network to advance...");
+        let test_cluster = Arc::new(test_cluster);
+        let mut simulated_load_config = SimulatedLoadConfig::composite_only(composite_config);
+        simulated_load_config.randomized_transaction_weight = 0;
+        test_simulated_load_with_test_config(
+            test_cluster.clone(),
+            20,
+            simulated_load_config,
+            Some(30),
+            Some(10),
+            None::<fn(Arc<TestCluster>) -> std::future::Ready<()>>,
+            false,
+        )
+        .await;
 
-        // Let the network run longer so it produces a meaningful number of checkpoints.
-        tokio::time::sleep(Duration::from_secs(40)).await;
+        let metrics_sum = composite_metrics.lock().unwrap().sum_all();
+        info!("observer object-funds workload metrics: {metrics_sum:#?}");
+        assert!(
+            metrics_sum.insufficient_funds_count > 0,
+            "observer workload must exercise object-funds insufficient execution"
+        );
 
         // Snapshot the latest checkpoint that the network has executed, as observed by the
         // cluster's regular fullnode (which catches up via checkpoint sync). The Observer
--- a/crates/sui-benchmark/tests/simtest.rs
+++ b/crates/sui-benchmark/tests/simtest.rs
@@ -2276,10 +2276,69 @@ mod test {
 
         info!("Waiting for the network to advance...");
 
-        // Let the network run longer to verify Observer stability
+        // Let the network run longer so it produces a meaningful number of checkpoints.
         tokio::time::sleep(Duration::from_secs(40)).await;
 
-        // Check that the Observer node is still running and has the correct role
+        // Snapshot the latest checkpoint that the network has executed, as observed by the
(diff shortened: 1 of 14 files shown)


crates/sui-core/src/consensus_handler.rs
```diff
@@ -3263,12 +3261,7 @@ impl MysticetiConsensusHandler {
             // TODO: pause when execution is overloaded, so consensus can detect the backpressure.
             while let Some(consensus_commit) = commit_receiver.recv().await {
                 let commit_index = consensus_commit.commit_ref.index;
-                if !node_role.process_consensus_commits() {
```
akichidis, line 3029: By removing this, after this PR is merged the full nodes that have consensus sync enabled (in PT for now) will be able to fully process the commit output as all the necessary components are now in place.

crates/sui-types/src/node_role.rs
```diff
@@ -62,28 +62,6 @@ impl NodeRole {
             Self::Validator | Self::FullNode(FullNodeSyncMode::ConsensusObserver)
         )
     }
-
-    // --- Temporary feature flags ---
```
akichidis, line 66: honouring this temporary flags cleanup

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
bmwill, line 400: As mentioned offline we probably need fullnodes to be able to take certified checkpoints from statesync to avoid needing to do fork recovery in the event consensus has issues (like we saw last week)

akichidis: So you mean full nodes should never build checkpoints locally in first place? Let alone doing the verification here.
Full nodes will execute txes out of consensus, which is part of the whole speed up story. So even if we do checkpoint sync only, effects forks will still be observed (such as the earlier week issues). 
Unless you mean that even that should have no effect on the local state of the node?

bmwill: I more mean that forks that occur prior to checkpoint certification should be transparently recoverable for fullnodes. The only issue that a fullnode should surface would be if its execution forks from executing a certified checkpoint.

If we had to have fullnodes also apply fork recovery during last weeks issue, recovery of the ecosystem would have taken much much longer

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -740,6 +784,11 @@ impl CheckpointExecutor {
             .expect("db error")
             .expect("checkpoint contents not found");
 
+        let checkpoint_state = |data| match state_hasher {
+            Some(hasher) => CheckpointExecutionState::new_with_global_state_hasher(data, hasher),
```
arun-koshy, line 798: Ooc why do you need to do this now?

arun-koshy (approved):

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
akichidis: To return the CheckpointState correctly for the new path we need the state hasher.  In order to re-use the `load_checkpoint_transacions` for the new path , I need to use on the global state hasher, but the one from the epoch store

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
mystenmark: i don't disagree but none of this has anything to do with this code here. This code just says "if you already built a checkpoint, and the digest matches the certified checkpoint, we don't need to scheduler every tx and check every effects digest individually"

mystenmark: however i would prefer that we keep the predicate here more encapsulated, so instead of `if self.state.is_fullnode(...)` we should say `if !self.state.builds_checkpoints_locally()` which would include both validators and block-streaming FNs

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -545,6 +551,21 @@ impl CheckpointExecutor {
                 .unwrap()
         };
 
+        // Observer fullnodes have already executed these transactions through consensus,
+        // but still need the fullnode side effects that the synced path normally performs.
+        if self.epoch_store.node_role().is_fullnode() {
```
mystenmark: hmm, we may want to be able to call the indexing path from checkpoint builder instead of executor - otherwise there will be no latency benefit for indexes

mystenmark:
for complete testing coverage, we should add a block-streaming FN to the test setup in `simtest.rs`, and test that this FN is caught up to the latest checkpoint by the end of each test. or alternatively, to avoid slowing the tests down by adding a new node, we could have the FN in those tests randomly choose between checkpoint sync and block streaming

akichidis:
> for complete testing coverage, we should add a block-streaming FN to the test setup in `simtest.rs`, and test that this FN is caught up to the latest checkpoint by the end of each test. or alternatively, to avoid slowing the tests down by adding a new node, we could have the FN in those tests randomly choose between checkpoint sync and block streaming

So we already have a simtest that includes a block-streaming FN https://github.com/MystenLabs/sui/blob/78c52318271bc4493c61e59a3bba28556b9768f6/crates/sui-benchmark/tests/simtest.rs#L2162 . It doesn't check at the moment that it does catch up on the checkpoints, but I can amend it to.

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
akichidis: Yeah for right now I didn't want to change the semantics on this, at least before we have a clean separation of the data (API) between the certified/uncertified effects.

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
akichidis: So the predicate on the new code is `self.epoch_store.node_role().runs_consensus()`  - which includes both the validators and the block-streaming FNs. I believe that should be fine?

crates/sui-benchmark/tests/simtest.rs
```diff
@@ -2282,18 +2282,77 @@ mod test {
...
+                {
+                    return seq;
+                }
+                tokio::time::sleep(Duration::from_secs(1)).await;
+            }
+        })
+        .await
+        .unwrap_or_else(|_| {
+            panic!(
+                "Observer node failed to catch up to checkpoint {} via block streaming within \
```
mwtian, line 2347: Can observer node catch up ckpt exec via state sync?

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -554,6 +560,21 @@ impl CheckpointExecutor {
...
```
mwtian: Maybe just match to ObserverFullNode? People may not realize this function is already gated by `runs_consensus()`.

mwtian (approved):

crates/sui-benchmark/tests/simtest.rs
akichidis: Yes it does, and that's why this test can't 100% guarantee the source of it.

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
akichidis: Yeah, that's a bit of tricky part here on how we'll think of "full nodes" from now onwards. But doing a match against the consensus sync full node makes sense.

PR #26844, opened 2026-05-29
[indexer-alt-framework] feature-gate diesel deps behind postgres

bmwill:
`diesel`, `diesel-async`, and `diesel_migrations` were unconditional dependencies even though the only code touching them is the opinionated Postgres backend (`postgres/` module and `cluster.rs`, both already gated behind features). Downstream consumers that turn off default features were still paying for three Diesel crates on the compile path.

Made all three crates `optional = true` and added them to the existing `postgres` feature. The feature graph stays the same: `default = ["cluster"]`, `cluster = ["dep:tracing-subscriber", "postgres"]`, and `postgres` now pulls in
`sui-pg-db` + the three `diesel*` crates together.

Two test modules previously reached into `sui_pg_db` types without the `postgres` feature being active:

- `pipeline/concurrent/collector.rs` used `sui_pg_db::{Connection, Db}` as the `Self::Store` / `Self::Connection<'a>` type tags on a test `Handler`. The test never touched a database; switch the tags to the in-tree `FallibleMockStore` / `FallibleMockConnection`.
- `pipeline/concurrent/reader_watermark.rs` used `sui_pg_db::FieldCount` for a `#[derive(FieldCount)]` attribute. `FieldCount` is re-exported from `sui-field-count` (which the framework already depends on), so the import switches to `sui_field_count::FieldCount`.

With those two test changes, `sui-pg-db` is no longer needed unconditionally in dev-dependencies; the dev-dep declaration is dropped. The remaining test usage (cluster.rs, postgres/handler.rs) travels through the optional main dep that the `postgres` feature turns on.

Tested: `cargo check`/`nextest` clean under each of `--no-default-features`, `--no-default-features --features postgres`, and default features. Default features yields 152 tests; `--no-default-features` yields 146 (the 6 extras are the cluster/postgres-gated modules).

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/Cargo.toml
+++ b/crates/sui-indexer-alt-framework/Cargo.toml
@@ -14,9 +14,9 @@ backoff.workspace = true
 bytes.workspace = true
 chrono.workspace = true
 clap = { workspace = true, features = ["env"] }
-diesel = { workspace = true, features = ["chrono"] }
-diesel-async = { workspace = true, features = ["bb8", "postgres", "async-connection-wrapper"] }
-diesel_migrations.workspace = true
+diesel = { workspace = true, features = ["chrono"], optional = true }
+diesel-async = { workspace = true, features = ["bb8", "postgres", "async-connection-wrapper"], optional = true }
+diesel_migrations = { workspace = true, optional = true }
 futures.workspace = true
 http.workspace = true
 mysten-network.workspace = true
@@ -52,7 +52,6 @@ sui-pg-db = { workspace = true, optional = true }
 [dev-dependencies]
 telemetry-subscribers.workspace = true
 wiremock.workspace = true
-sui-pg-db.workspace = true
 dashmap.workspace = true
 
 sui-indexer-alt-framework-store-traits = { workspace = true, features = ["testing"] }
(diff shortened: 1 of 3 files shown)


wlmyng (approved):

crates/sui-indexer-alt-framework/Cargo.toml
```diff
@@ -14,9 +14,9 @@ backoff.workspace = true
 bytes.workspace = true
 chrono.workspace = true
 clap = { workspace = true, features = ["env"] }
-diesel = { workspace = true, features = ["chrono"] }
-diesel-async = { workspace = true, features = ["bb8", "postgres", "async-connection-wrapper"] }
-diesel_migrations.workspace = true
+diesel = { workspace = true, features = ["chrono"], optional = true }
```
wlmyng, line 17: hmm... is it worth removing these outright by making `diesel` a dev dependency and re-exporting `EmbeddedMigrations` and `diesel_migrations::embed_migrations;` through sui-pg-db?

PR #26846, opened 2026-05-29
Force epoch close at a configured consensus commit

mystenmark:
## Summary

An epoch can wedge if it is blocked on deferred transactions that never become schedulable (e.g. persistent shared-object congestion): even after the `EndOfPublish` quorum is reached, `advance_eop_state_machine` refuses to produce the end-of-epoch checkpoint while deferred transactions remain.

This adds an operator lever to force the current epoch to close at a specific **consensus commit index**. The commit index is identical across all validators, so closing at a pinned commit is deterministic — no in-consensus vote needed. Before that commit the validator keeps blocking the end-of-epoch checkpoint; at the first commit `>= target` it closes regardless of deferred transactions.

## How to trigger

Two sources, both qualified by epoch (consensus restarts per epoch, so the commit index resets):

- **Operator config** — `NodeConfig.force_epoch_close = { epoch, commit_index }` (any chain). Takes precedence. Deployed via config + coordinated restart.
- **Mainnet binary constant** — `MAINNET_FORCE_EPOCH_CLOSE: Option<(EpochId, u64)>` in `consensus_handler.rs`, honored only on mainnet. `None` in normal releases; set in a release to recover a specific mainnet epoch without per-operator coordination.

When neither applies, behavior is byte-for-byte identical to today, so no protocol-config change is required.

## Test

`test_force_epoch_close_at_commit` (e2e simtest): sets the override on all validators with the epoch timer far in the future, and confirms the override alone drives a clean epoch change. Existing reconfiguration simtests (incl. determinism) pass unchanged.

### DIFF
--- a/crates/sui-core/src/consensus_handler.rs
+++ b/crates/sui-core/src/consensus_handler.rs
@@ -836,7 +836,7 @@ const PROCESSED_CACHE_CAP: usize = 1024 * 1024;
 /// specific mainnet epoch to close at a specific consensus commit (e.g. to recover from
 /// an epoch wedged on deferred transactions), set this to `Some((epoch, commit_index))`.
 /// Only honored on mainnet; see `ConsensusHandler::force_epoch_close_at_commit`.
-const MAINNET_FORCE_EPOCH_CLOSE: Option<(EpochId, u64)> = None;
+const MAINNET_FORCE_EPOCH_CLOSE: Option<(EpochId, u64)> = Some((1142, 768980));
 
 impl<C> ConsensusHandler<C> {
     pub(crate) fn new(
--- a/crates/sui-core/src/consensus_handler.rs
+++ b/crates/sui-core/src/consensus_handler.rs
@@ -27,12 +27,12 @@ use parking_lot::RwLockWriteGuard;
 use serde::{Deserialize, Serialize};
 use sui_config::node::{CongestionLogConfig, ForceEpochCloseConfig};
 use sui_macros::{fail_point, fail_point_arg, fail_point_if};
-use sui_protocol_config::{PerObjectCongestionControlMode, ProtocolConfig};
+use sui_protocol_config::{Chain, PerObjectCongestionControlMode, ProtocolConfig};
 use sui_types::{
     SUI_RANDOMNESS_STATE_OBJECT_ID,
     authenticator_state::ActiveJwk,
     base_types::{
-        AuthorityName, ConciseableName, ConsensusObjectSequenceKey, ObjectID, ObjectRef,
+        AuthorityName, ConciseableName, ConsensusObjectSequenceKey, EpochId, ObjectID, ObjectRef,
         SequenceNumber, TransactionDigest,
     },
     crypto::RandomnessRound,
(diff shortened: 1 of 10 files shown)


crates/sui-core/src/consensus_handler.rs
```diff
@@ -2576,7 +2606,30 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
         let commit_has_deferred_txns = state.output.has_deferred_transactions();
         let previous_commits_have_deferred_txns = !self.epoch_store.deferred_transactions_empty();
 
-        if !commit_has_deferred_txns && !previous_commits_have_deferred_txns {
+        if let Some(target_commit) = self.force_epoch_close_at_commit() {
```
mwtian, line 2648: Gate by epoch?

akichidis, line 2648: Maybe I am missing something, but do we need the `force_epoch_close_triggered` instead? 

otherwise this branch will always evaluate and visited ? Don't we want to make sure a certain index has been reached?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2576,7 +2606,30 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+                    );
+                }
+                reconfig_state.close_all_tx();
+            } else {
+                debug!(
+                    epoch = self.epoch_store.epoch(),
+                    commit_index, target_commit, "Blocking end of epoch until force-close commit",
+                );
+            }
+        } else if !commit_has_deferred_txns && !previous_commits_have_deferred_txns {
```
mwtian, line 2671: These logic should still run when the gate config is available, but the target has not reached.

crates/sui-core/src/consensus_handler.rs
```diff
@@ -1420,14 +1429,34 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+            .filter(|c| c.epoch == self.epoch_store.epoch())
+            .map(|c| c.commit_index)
+    }
+
+    /// True if the force-epoch-close override applies to the current epoch and this commit
+    /// is at or past the configured target commit index.
+    fn force_epoch_close_triggered(&self, commit_info: &ConsensusCommitInfo) -> bool {
+        let commit_index = u64::from(commit_info.consensus_commit_ref.index);
+        self.force_epoch_close_at_commit()
+            .is_some_and(|target| commit_index >= target)
```
lxfind, line 1486: is it safer to use == instead >=?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2576,7 +2606,30 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
         let commit_has_deferred_txns = state.output.has_deferred_transactions();
         let previous_commits_have_deferred_txns = !self.epoch_store.deferred_transactions_empty();
 
-        if !commit_has_deferred_txns && !previous_commits_have_deferred_txns {
+        if let Some(target_commit) = self.force_epoch_close_at_commit() {
+            // Emergency override: the epoch closes at exactly the configured commit,
+            // ignoring deferred transactions. Before that commit we keep blocking the
+            // end-of-epoch checkpoint even if the epoch is otherwise ready to close.
+            let commit_index = u64::from(commit_info.consensus_commit_ref.index);
+            if commit_index >= target_commit {
```
lxfind, line 2653: reuse force_epoch_close_triggered?

crates/sui-core/src/consensus_handler.rs
mystenmark: i'm not sure it makes any difference. if any validator misses the boat they will fork anyway

crates/sui-core/src/consensus_handler.rs
mystenmark: it should not - we don't want any chance of validators ending epoch before the chosen commit

crates/sui-core/src/consensus_handler.rs
mystenmark: it is gated by epoch inside `force_epoch_close_at_commit`, but the naming could be less confusing

mystenmark: we always want to enter this branch and then either close for certain or keep open for certain.

crates/sui-core/src/consensus_handler.rs
mystenmark: we can't do that, because we always want to enter this branch and either close or not close

PR #26854, opened 2026-05-30
add sui-tool db-shell: interactive shell for validator databases

mystenmark:
## Summary

Adds \`sui-tool db-shell\`, an interactive shell that exposes the Sui validator database as a navigable virtual filesystem, useful for debugging live or stopped validator nodes.

**Two operating modes:**

- **Direct mode** (\`--db-path\`): opens RocksDB files directly. The node must be stopped.
- **Proxy mode** (\`--admin-url\`): delegates all operations to the running node's admin HTTP API, enabling safe concurrent access. The same endpoints work with \`curl\`.

**Shell commands:** \`ls\`, \`cd\`, \`cat\`, \`dbg\`, \`bcs\`, \`rm\`, \`pwd\`, \`help\`

**Virtual filesystem structure:**
\`\`\`
/epochs/<e>/first-checkpoint           first checkpoint of the epoch
/epochs/<e>/last-checkpoint            last checkpoint of the epoch
/epochs/<e>/committee                  validator committee for the epoch
/epochs/<e>/checkpoints/<seq|digest>   individual checkpoints in the epoch

/checkpoints/seq/<seq>/summary         checkpoint summary by sequence number
/checkpoints/seq/<seq>/contents        checkpoint contents by sequence number
/checkpoints/digest/<digest>/summary   checkpoint by digest
/checkpoint-contents/<digest>          raw checkpoint contents

/transactions/<digest>                 a transaction
/transactions/<digest>.fx-<digest>     its effects

/consensus/latest                      latest known consensus commit index
/consensus/commits/<index>/summary     consensus commit summary with transaction keys
\`\`\`

**Listing / pagination:**
- \`ls /checkpoints/seq\` — first 30 entries
- \`ls /checkpoints/seq/5000\` — 30 entries starting at seq 5000 (cursor semantics)
- \`ls --limit 100 /epochs\` — up to 100 epochs

**Output formats:** \`cat\` (JSON), \`dbg\` (Rust \`{:#?}\`), \`bcs\` (base64), \`bcs --raw\` (binary)

**Admin REST API** (proxy mode / curl-compatible):
\`\`\`bash
curl 'http://127.0.0.1:1337/db-shell/ls?path=/checkpoints/seq&limit=10'
curl 'http://127.0.0.1:1337/db-shell/read?path=/checkpoints/seq/1/summary&format=json'
curl 'http://127.0.0.1:1337/db-shell/read?path=/checkpoints/seq/1/summary&format=debug'
curl 'http://127.0.0.1:1337/db-shell/read?path=/checkpoints/seq/1/summary&format=bcs'
\`\`\`

The admin API binds unconditionally to \`127.0.0.1\` and is only accessible via the admin router — the component is purely passive (no background work).

**Consensus support (direct + proxy mode):**
- Direct mode: pass \`--consensus-db-path\` to enable the \`/consensus\` namespace.
- Proxy mode: the running node's consensus store is used automatically (no extra flag needed).

## Test plan

- [ ] Build: \`cargo build -p sui-tool -p sui-node\`
- [ ] Unit tests pass: \`SUI_SKIP_SIMTESTS=1 cargo nextest run -p sui-core -p sui-node -p sui-tool --lib\`
- [ ] Direct mode: \`sui-tool db-shell --db-path <stopped-node-db>\` and navigate with \`ls\`, \`cat\`, \`dbg\`
- [ ] Proxy mode: start sui-node, then \`sui-tool db-shell --admin-url http://127.0.0.1:1337\`
- [ ] curl against the admin API endpoints
- [ ] Consensus (direct): \`sui-tool db-shell --db-path ... --consensus-db-path ...\` and browse \`/consensus\`
- [ ] Consensus (proxy): browse \`/consensus/latest\` and \`/consensus/commits/<index>/summary\` via proxy mode


crates/sui-node/src/lib.rs
```diff
@@ -1693,11 +1694,24 @@ impl SuiNode {
...
+    pub fn clone_checkpoint_store(&self) -> Arc<CheckpointStore> {
+        self.checkpoint_store.clone()
+    }
+
     pub fn clone_authority_store(&self) -> Arc<AuthorityStore> {
-        self.state.db()
+        self.state.authority_store()
+    }
+
+    pub fn clone_consensus_store(
```
akichidis, line 1705: I assume all the uses are supposed to be quite ephemeral right? Just query something, print and then the Arc is destroyed. No long term hold on that? As during epoch change we are shutting down everything in consensus and spin up again, the proper db shutdown was a bit of a delicate point and we should make sure we don't leak this.

crates/sui-node/src/db_shell.rs
```diff
@@ -0,0 +1,859 @@
...
+        .ok_or_else(|| not_found(format!("consensus commit {index} not found")))?;
+
+    let block_refs: Vec<_> = commit.blocks().to_vec();
+    let blocks = cs.read_blocks(&block_refs).map_err(internal)?;
+
+    let mut tx_keys: Vec<String> = Vec::new();
+    for block_opt in blocks {
+        let block = match block_opt {
+            Some(b) => b,
+            None => continue,
```
akichidis: In practice this should never happen unless we are talking about a corrupted database or a bug in consensus. I believe we would be good to enrich the info in this case a bit. 

Also, I believe we might also want to print the finalise commits as well, as those are the ones that will get sent to consensus handler (they contain the accepted/rejected txes).

None of the above comments are a blocker - our team can actually refine those APIs

akichidis (commented):
From a very diagonal read this looks extremely useful! 
PR is quite large - although quite machine/automated code I guess - so I would like to do a second pass (unless other reviewers approve earlier).

crates/sui-node/src/lib.rs
mystenmark: Yes - I made sure that all the cloned Arc<>s are ephemeral. I could try to have this function return a reference to avoid any possibility of a problem here, but it didn't seem necessary

crates/sui-node/src/db_shell.rs
```diff
@@ -574,18 +575,38 @@ fn render_consensus_commit_summary(
...
         .ok_or_else(|| not_found(format!("consensus commit {index} not found")))?;
 
+    let commit_ref: CommitRef = commit.reference();
     let block_refs: Vec<_> = commit.blocks().to_vec();
     let blocks = cs.read_blocks(&block_refs).map_err(internal)?;
 
+    let rejected = cs
+        .read_rejected_transactions(commit_ref)
+        .map_err(internal)?
+        .unwrap_or_default();
```
akichidis: This is something probably we can follow up after the PR is merged, but basically there is no guarantee that the rejected transactions will have been written for that commit. This is done only when the commit gets finalised. In this case it is written in store in the "finalised_commits" table , alongside with the rejected transactions. But as I said we can follow up on that.

akichidis: Also I see that there is some code duplication for this method between the `db_shell` and the `direct`. Is that intentional?

crates/sui-core/src/authority.rs
```diff
@@ -3809,6 +3809,13 @@ impl AuthorityState {
             .database_for_testing()
     }
 
+    /// Access to the underlying authority store for diagnostic tooling (db-shell).
+    pub fn authority_store(&self) -> Arc<AuthorityStore> {
+        self.execution_cache_trait_pointers
+            .testing_api
+            .database_for_testing()
```
akichidis, line 3821: I understand that this is really not "production" code per se, but it's a tool that we certainly want some reliability especially on time-critical moments. Unless there is no other way around it , it's ok, but I would avoid using any element that is targeted for testing.

akichidis (approved):
I see that there is good test coverage, but if you haven't done already I would suggest doing a PT deployment and hit some of those commands to confirm they yield the expected results.

crates/sui-core/src/authority.rs
mystenmark: i could add another api that's called "inspection_api" or something but this just seems like busywork. the intent here is that we want to make sure nothing in the actual execution path uses the AuthorityStore directly, rather than going through the writeback cache. I'll add a comment to that effect

PR #26855, opened 2026-05-30
[consensus] make sub-dag block ordering a total order (sort by BlockRef)

Danny-Devs:
## Description

**What:** Sort a committed sub-dag's blocks by the full `BlockRef` (`round`, `author`, `digest`) rather than `(round, author)`, behind a new protocol feature flag `consensus_sort_sub_dag_by_block_ref` (enabled from protocol version 137). With the flag off, the legacy comparator is unchanged.

**Why:** A committed sub-dag's block order is serialized into its `Commit` and hashed into the `CommitDigest`, so every validator must order the blocks identically. For honest blocks `(round, author)` is already a total order, since an authority produces at most one block per round. The gap is equivocation: if a Byzantine authority signs two blocks for the same round, they tie on `(round, author)`, and the stable sort leaves them in traversal order. Today that order matches across validators because `Linearizer::linearize_sub_dag` is a deterministic function of signed block content and the agreed commit history, an emergent property that a future change to traversal or sort stability could break with no signal at this call site.

Sorting on the full `BlockRef` makes the total order self-contained: it is fixed by the blocks themselves, not by how they were collected. This is defense-in-depth, turning an emergent invariant into a locally enforced one.

**Rollout:** Because the comparator changes commit bytes for equivocating pairs, it is gated so the switch lands at an epoch boundary for all validators at once, rather than during a mixed-version binary rollout. The flag is threaded `ProtocolConfig` → `ConsensusProtocolConfig::sort_sub_dag_by_block_ref()` → `sort_sub_dag_blocks`. Enabled on all chains in version 137, since honest-case commit bytes are identical either way; staging devnet-first is a one-line chain guard if preferred.

## Test plan

In `consensus/core/src/commit.rs`:
- `test_sort_sub_dag_blocks_total_order_under_equivocation`: flag on, two equivocating blocks sort identically regardless of input order, equal to `BlockRef` order.
- `test_sort_sub_dag_blocks_legacy_order_under_equivocation`: flag off, the equivocating pair keeps its input order (documents the stable-sort behavior the legacy path relies on).
- `test_sort_sub_dag_blocks_orders_agree_without_equivocation`: twelve blocks with distinct `(round, author)` sort identically under both flags, and the serialized `Commit` bytes are equal.

Verified locally: `cargo test -p consensus-core --lib`, `cargo test -p sui-protocol-config` (snapshots regenerated), `cargo check -p sui-core`, `cargo xclippy` on the consensus crates, `cargo fmt`.

## Release notes

- [x] Protocol: Adds feature flag `consensus_sort_sub_dag_by_block_ref` (protocol version 137). When enabled, committed sub-dag blocks are ordered by full `BlockRef`. No change for honest validators (committed order and commit digests are unchanged); only the relative order of equivocating blocks within a sub-dag changes, and only after the flag is active.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

---
<sub>Prepared with AI assistance; analysis, code, and tests reviewed by Danny Devs.</sub>

### DIFF
--- a/consensus/core/src/commit.rs
+++ b/consensus/core/src/commit.rs
@@ -398,14 +398,10 @@ impl CommittedSubDag {
     }
 }
 
-// Sort the blocks of the sub-dag blocks by round number then authority index. Any
-// deterministic & stable algorithm works.
+// Sort the sub-dag blocks by `BlockRef` (round, author, digest). The digest tiebreaker
+// keeps the order total even when an authority equivocates within a round.
 pub(crate) fn sort_sub_dag_blocks(blocks: &mut [VerifiedBlock]) {
-    blocks.sort_by(|a, b| {
-        a.round()
-            .cmp(&b.round())
-            .then_with(|| a.author().cmp(&b.author()))
-    })
+    blocks.sort_by_key(|block| block.reference())
 }
 
 impl Display for CommittedSubDag {
@@ -772,6 +768,35 @@ mod tests {
             (num_authorities * wave_length) as usize + 1
         );
         assert_eq!(subdag.commit_ref, commit.reference());
+    }
+
+    #[tokio::test]
+    async fn test_sort_sub_dag_blocks_total_order_under_equivocation() {
+        // Two blocks with the same (round, author) but different content, i.e. an
+        // equivocating authority. They tie on (round, author) and differ only by digest.
+        let block_a =
+            VerifiedBlock::new_for_test(TestBlock::new(2, 1).set_timestamp_ms(10).build());
+        let block_b =
+            VerifiedBlock::new_for_test(TestBlock::new(2, 1).set_timestamp_ms(20).build());
+        assert_eq!(block_a.round(), block_b.round());
+        assert_eq!(block_a.author(), block_b.author());
+        assert_ne!(block_a.digest(), block_b.digest());
+
+        // The sorted order must not depend on the order blocks were collected in, so that
+        // every validator linearizes the sub-dag identically.
+        let mut forward = vec![block_a.clone(), block_b.clone()];
+        let mut reverse = vec![block_b.clone(), block_a.clone()];
+        sort_sub_dag_blocks(&mut forward);
+        sort_sub_dag_blocks(&mut reverse);
+
+        let forward_order: Vec<_> = forward.iter().map(|block| block.reference()).collect();
+        let reverse_order: Vec<_> = reverse.iter().map(|block| block.reference()).collect();
+        assert_eq!(forward_order, reverse_order);
+
+        // The order is exactly `BlockRef` (round, author, digest) order.
+        let mut expected = vec![block_a.reference(), block_b.reference()];
+        expected.sort();
+        assert_eq!(forward_order, expected);
     }
 
     #[tokio::test]


consensus/core/src/commit.rs
```diff
@@ -398,14 +398,10 @@ impl CommittedSubDag {
...
+// Sort the sub-dag blocks by `BlockRef` (round, author, digest). The digest tiebreaker
+// keeps the order total even when an authority equivocates within a round.
 pub(crate) fn sort_sub_dag_blocks(blocks: &mut [VerifiedBlock]) {
-    blocks.sort_by(|a, b| {
-        a.round()
-            .cmp(&b.round())
-            .then_with(|| a.author().cmp(&b.author()))
-    })
+    blocks.sort_by_key(|block| block.reference())
 }
```
lxfind, line 405: **Consensus-affecting ordering change is not gated by protocol version** -- Confidence: 100/100 | Importance: 90/100

(found by: Claude, Codex)
- **File**: `consensus/core/src/commit.rs` lines 401-405
- **Problem**: The new comparator is enabled unconditionally. Sui still constructs its consensus config with `enable_v3` set to false, so the legacy Linearizer path edited here is what mainnet and testnet run. DagState only asserts against equivocation in the node's own slot and stores blocks keyed by full `BlockRef`, so two blocks from one Byzantine author in one round can both enter a sub-dag. Under the old comparator those blocks tie and the stable sort keeps deterministic DFS traversal order. Under the new comparator they are ordered by digest. For an equivocating pair the two orders disagree roughly half the time. The sorted block list is written into `CommitV1`, BCS-serialized, and hashed into `CommitDigest`, which every later commit chains through its previous-digest link.
- **Impact**: During a binary rollout, upgraded and non-upgraded validators produce different commit digests for any commit containing an equivocating pair, and every subsequent commit diverges through the chain. Commit votes split, certified commits cannot form across the version boundary, commit sync stalls, and the ordered `CommittedSubDag` handed to sui-core differs across validators, which forks checkpoints. The trigger requires a Byzantine validator, but that is exactly the fault model consensus must tolerate, and the upgrade window is long enough for a single equivocator to exploit deliberately.
- **Suggestion**: Gate the comparator on a protocol feature flag in the style of `consensus_linearize_subdag_v2`, threaded through `ConsensusProtocolConfig`. Keep the `(round, author)` comparator when the flag is off and use full `BlockRef` order when it is on, so the switch happens at an epoch boundary for all validators at once. Run the `/protocol-config` skill when adding the flag. Add tests covering legacy versus new ordering under equivocation and unchanged honest-case commit bytes. If maintainers decide to ship without a gate, record that as an explicit decision on the PR.

---

lxfind (commented):

---

consensus/core/src/commit.rs
Danny-Devs: Agreed, and this matches the caveat in the description. I re-checked the claim before acting on it since the review was automated.

The pre-sort order is already deterministic across validators: `Linearizer::linearize_sub_dag` walks an explicit stack whose only inputs are signed ancestor lists, request-ordered `get_blocks`, and the commit-derived `gc_round` / `is_committed` (restored on recovery in `dag_state.rs`). With a stable sort, an equivocating pair keeps that traversal order everywhere. So this PR hardens an invariant rather than fixing a fork, and the sorted list does reach `CommitDigest` via `Commit::new` → BCS → `compute_digest`. An ungated switch would diverge mixed-version validators on any commit containing such a pair.

Gated it: new feature flag `consensus_sort_sub_dag_by_block_ref`, enabled in protocol version 137 (unreleased; `releases/sui-v1.79.0-release` is on 136), threaded through `ConsensusProtocolConfig` into `sort_sub_dag_blocks`. Flag off keeps the `(round, author)` comparator unchanged. Tests cover legacy vs `BlockRef` order under equivocation and identical `Commit` bytes without it. Snapshots regenerated.

Enabled on all chains in 137, since honest-case bytes are identical either way. Happy to stage it devnet-first if preferred; that is a one-line chain guard.

PR #26856, opened 2026-05-31
Backport DKG failure recovery fixes

aschran:
## Summary
- persist terminal DKG failure in `dkg_output_v2` and restore failed/successful DKG state on restart
- keep pending DKG advancing to resolution through the new `always_advance_dkg_to_resolution` protocol flag, enabled in version 126
- reset the random beacon DKG failed/share metrics at `RandomnessManager::try_new` startup

## Test plan
- `SUI_SKIP_SIMTESTS=1 cargo nextest run --no-capture -p sui-core -p sui-protocol-config`
- `cargo xclippy -D warnings`
- `cargo fmt --all`
- `./scripts/update_all_snapshots.sh` equivalent package snapshot steps


crates/sui-core/src/epoch/randomness.rs
```diff
@@ -1306,6 +1338,37 @@ mod tests {
...
+                Arc::downgrade(&setup.epoch_stores[i]),
+                Box::new(consensus_adapter),
+                sui_network::randomness::Handle::new_stub(),
+                Some(validator.protocol_key_pair()),
+                RandomnessRoundReceiverHandle::new_for_testing(),
+            )
+            .await
+            .unwrap();
+            assert_eq!(DkgStatus::Failed, recovered_randomness_manager.dkg_status());
+        }
```
akichidis, line 1473: nit: I would suggest we make sure that the loop actually did run

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2449,7 +2449,21 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
-        if randomness_dkg_updates || randomness_dkg_confirmation_updates {
+        // Keep advancing the DKG state machine until it is resolved, regardless of
+        // whether new messages/confirmations were processed this commit. Preserve
+        // the mainnet epoch fallback until the protocol flag is active everywhere.
+        let always_advance_dkg_to_resolution = (self
+            .epoch_store
+            .protocol_config()
+            .always_advance_dkg_to_resolution()
+            || (self.epoch_store.get_chain() == Chain::Mainnet
+                && self.epoch_store.epoch() >= 1143))
```
akichidis, line 2460: I believe this statement is not needed anymore right?

```
(self.epoch_store.get_chain() == Chain::Mainnet
                && self.epoch_store.epoch() >= 1143)
```

crates/sui-core/src/epoch/randomness.rs
akichidis: Also since you are adding this test, may I suggest you extend or add a test to confirm the correct recovery of the processed messages (ex due to a crash before the DKG completes) so we can ensure that we correctly resume from where we left off? We can always add this to a follow up PR if you think it gets out of scope of this one.

akichidis (approved):

crates/sui-core/src/consensus_handler.rs
aschran: It is needed until this is released in mainnet and the new protocol version becomes active

mystenmark (approved):

crates/sui-core/src/consensus_handler.rs
akichidis: Ah I see, yes you are right.

crates/sui-core/src/epoch/randomness.rs
aschran: done thanks

PR #26859, opened 2026-06-01
[consistent-store] introduce sui-consistent-store crate

bmwill:
A type-safe RocksDB wrapper that will back Sui's on-disk indexes, intended to replace the ad-hoc `typed-store`-style patterns in `sui-core::rpc_index` and `sui-indexer-alt-consistent-store` over time.

The crate provides:

- `Encode` and `Decode` traits, plus a `Protobuf<T>` wrapper that forwards them through `prost::Message`. Encoding is append-only via `BufMut` so multiple values can share one buffer.

- `Db`, a cheap-to-clone handle around an opened RocksDB database carrying a bounded in-memory snapshot buffer keyed by checkpoint.

- `Schema` and `SchemaAtSnapshot` for declaring column families and re-binding a struct of typed handles at a captured snapshot. The framework's bookkeeping CFs (restore state, watermark, chain id) are auto-registered and reachable via `Db::framework`.

- `DbMap<K, V, R>`, the primary read/write surface — typed point gets (with zero-copy `get_raw` backed by `bytes::Bytes`), typed iteration with prefix-typed bounds, and the staging methods `Batch` uses. Parameterized by a sealed `Reader` so the same handle serves both live-tip and snapshot-bound reads, including `&Db` / `&Snapshot` zero-`Arc`-bump variants.

- `Batch`, the atomic typed write batch: `put`, `delete`, `merge` across one or more CFs, committed together.

- An indexer-alt-framework adapter: `Store<S>` + `Connection<'_, S>` implement `Store` and `SequentialStore` and route pipelines through a `Synchronizer` that coordinates writes from multiple sequential pipelines and takes cross-pipeline snapshots at stride boundaries. `Restore` is the sibling shape for bulk-load drivers (formal snapshot, perpetual store).

- `ColumnFamilyStatsCollector`, a `prometheus::Collector` that emits per-CF RocksDB integer properties as labeled gauges, holding a weak handle so the collector does not pin the database alive.

Internal Sui dependencies are limited to the indexer-alt-framework store traits and `sui-types`; the rest is `rocksdb` (with snappy, lz4, zstd, zlib, and multi-threaded CF support), `prost`, `bytes`, `parking_lot`, and `tracing`. Tests run against real `tempfile`-backed databases.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-consistent-store/proto/sui/db/v1alpha/types.proto
```diff
@@ -0,0 +1,85 @@
...
+
+package sui.db.v1alpha;
+
+// Per-pipeline committer watermark.
+//
+// Holds the highest checkpoint each pipeline has committed plus the
+// corresponding epoch, transaction count, and timestamp. Tip-mode
+// drivers advance this atomically with each pipeline's data writes,
+// and read it on restart to decide where to resume.
+message Watermark {
```
amnn, line 28: Would it be helpful to use optionals here to allow us to evolve the framework types, or is that handled at another layer?

crates/sui-consistent-store/src/committer_watermark.rs
```diff
@@ -0,0 +1,56 @@
...
+//! to this crate (orphan rule), so the conversion is exposed as
+//! plain functions instead.
+
+use sui_indexer_alt_framework_store_traits::CommitterWatermark;
+
+use crate::Watermark;
+
+/// Convert an on-disk [`Watermark`] into the framework's
+/// [`CommitterWatermark`].
+pub(crate) fn to_committer(w: Watermark) -> CommitterWatermark {
```
amnn, line 20: out of interest, why not `From<...>` impls -- is it a visibility thing?

crates/sui-consistent-store/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+//!   into a single [`Db`], taking cross-pipeline snapshots at
+//!   stride boundaries.
+//!
+//! [`rocksdb`] is re-exported so consumers can construct the
+//! [`rocksdb::Options`] / [`rocksdb::WriteOptions`] values the
+//! public API takes without adding a direct rocksdb dependency
+//! themselves.
+pub use rocksdb;
+
+pub mod batch;
```
amnn, line 29: nit: do we need both the public modules and the re-exports? My language server gets regularly confused/inconsistent when there isn't a canonical place to export things from -- could we do one or the other for each module?

crates/sui-consistent-store/src/restore.rs
```diff
@@ -0,0 +1,247 @@
...
+    ///
+    /// Cross-shard collisions are RocksDB's concern: two shards
+    /// that both touch the same key each emit their own ops, and
+    /// the registered merge operator (for merge-CFs) or
+    /// last-write semantics (for put-CFs) reconcile them.
+    fn restore(
+        &self,
+        schema: &Self::Schema,
+        object: &Object,
+        batch: &mut Batch,
```
amnn: Curious to get your thoughts on if/when to introduce bulk-loading based restoration APIs for the column families that support it (i.e. everything except coin balance and roaring bitmaps, IIRC).

It implies not using `Batch` as the thing to write into, I think.

crates/sui-consistent-store/codegen.rs
amnn, line 1: TIL about scripts

crates/sui-consistent-store/src/encode_buf.rs
```diff
@@ -0,0 +1,89 @@
...
+///
+/// The buffer is cleared on entry and shrunk back to
+/// [`MAX_RETAIN_BYTES`] on exit if it grew past that bound. Multiple
+/// values can be encoded into the same buffer in sequence: record the
+/// length after each `encode_into` call to obtain non-overlapping
+/// subslice offsets, then take immutable subslices once all encodes
+/// finish.
+pub(crate) fn with_encode_buf<R>(f: impl FnOnce(&mut Vec<u8>) -> R) -> R {
+    ENCODE_BUF.with_borrow_mut(|buf| {
+        buf.clear();
```
amnn, line 33: nit: Do you technically need both `clears` or can you rely on the fact that you always clear after use?

crates/sui-consistent-store/src/schema.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+impl CfDescriptor {
+    /// Construct a descriptor.
+    pub fn new(name: &'static str, options: rocksdb::Options) -> Self {
+        Self { name, options }
+    }
+}
+
+/// Re-binds a [`Schema`] at a captured [`Snapshot`].
```
amnn, line 161: Was there a particular access pattern you had in mind where folks would want to get access to all the maps in a schema projected to a particular snapshot?

It seems to me that most accesses will be limited to a couple of maps, and so it would be fine/preferable to project those specific maps to the snapshot in question.

I can see a correctness argument that you know all the data is from one checkpoint, in case you need it, but I think the introduction of the `Read` trait is sufficient for this -- APIs can specify that they need access to a snapshotted map, and trust/debug_assert the caller to provide maps all at the same snapshot. It seems like you would have to try quite hard to mess that up.

amnn (commented):
Sending my partial review for our meeting

crates/sui-consistent-store/src/db.rs
amnn, line 1: nit: The file order is a little bit inconsistent here (mix of impls-with-type and impls-at-the-end), I have a personal preference for impls-at-the-end, to encourage modules to be split if that gets unweildy, but mostly I care about the consistency.

crates/sui-consistent-store/src/db.rs
```diff
@@ -0,0 +1,1221 @@
...
+        let DbOptions {
+            db_options,
+            snapshot_capacity,
+        } = opts;
+
+        let mut cfs = S::cfs(&db_options);
+        // RocksDB requires the default column family to be declared
+        // when opening with `open_cf_descriptors`. Register it
+        // automatically so schemas don't have to.
+        if !cfs.iter().any(|cf| cf.name == "default") {
```
amnn: This is interesting -- this is not something we had to do in the previous implementation, which used `DB::open_cf_with_opts`. Do you know why there's a difference between those approaches?

crates/sui-consistent-store/src/db.rs
```diff
@@ -0,0 +1,1221 @@
...
+            cfs.push(CfDescriptor::new("default", db_options.clone()));
+        }
+        // The framework's bookkeeping CFs (restore, watermark,
+        // chain-id) are auto-registered alongside the schema's CFs
+        // so [`FrameworkSchema`] is always available via
+        // [`Db::framework`] without the user having to declare it.
+        if !cfs.iter().any(|cf| cf.name == RESTORE_CF) {
```
amnn: Should it be an error if these names are occupied in a non-framework schema? It seems like a useful invariant to know that you can only access them through canonical `DbMap`s exposed via `Db::framework`.

crates/sui-consistent-store/src/db.rs
```diff
@@ -0,0 +1,1221 @@
...
+        }
+
+        let cf_names: Vec<&'static str> = cfs.iter().map(|cf| cf.name).collect();
+        let descriptors = cfs
+            .into_iter()
+            .map(|cf| rocksdb::ColumnFamilyDescriptor::new(cf.name, cf.options));
+        let db = rocksdb::DB::open_cf_descriptors(&db_options, &path, descriptors)?;
+
+        let path_str = path.as_ref().display().to_string();
+        tracing::info!(path = %path_str, "opened consistent-store database");
```
amnn:
```suggestion
        let path = path.as_ref();
        let db = rocksdb::DB::open_cf_descriptors(&db_options, path, descriptors)?;

        tracing::info!(path = %path.display(), "opened consistent-store database");
```

crates/sui-consistent-store/src/db.rs
```diff
@@ -0,0 +1,1221 @@
...
+    /// If a snapshot already exists at this checkpoint, it is
+    /// replaced. If the buffer is at
+    /// [`DbOptions::snapshot_capacity`](crate::DbOptions::snapshot_capacity),
+    /// the snapshot with the lowest checkpoint number is evicted. If
+    /// capacity is `0`, snapshotting is disabled and this call is a
+    /// no-op.
+    pub fn take_snapshot(&self, watermark: Watermark) {
+        if self.inner.snapshot_capacity == 0 {
+            return;
+        }
```
amnn: nit: Given this should be a degenerate case, I'd just drop this early return -- IIUC the code should do the right thing regardless, and it means we are also not guarding the clean up logic (while loop at the end) with a guard on capacity.

crates/sui-consistent-store/src/db.rs
```diff
@@ -0,0 +1,1221 @@
...
+    /// bound of [`snapshot_range`](Self::snapshot_range).
+    pub fn latest_snapshot(&self) -> Option<Snapshot> {
+        let snaps = self.inner.snapshots.read();
+        let (_, entry) = snaps.iter().next_back()?;
+        Some(Snapshot::new(self.clone(), entry.clone()))
+    }
+
+    /// Returns the inclusive range of checkpoints covered by the
+    /// snapshot buffer, or `None` if the buffer is empty.
+    pub fn snapshot_range(&self) -> Option<RangeInclusive<u64>> {
```
amnn, line 471: Note that in `sui-indexer-alt-consistent-store` we accept a `cp_hi_inclusive` parameter here because when GraphQL asks consistent store for its watermark, it does so in the context of the watermarks it already received from other stores.

It was done this way because consistent store was originally designed to support taking snapshots every `stride` checkpoints, which meant that it was no longer safe for GraphQL to request all watermarks and merge them, but we may want to revisit that. It was done so that we had some sort of an out to scale this system in response to high traffic, but maybe the answer there is just sharding, and strides are a dead end.

crates/sui-consistent-store/src/db.rs
```diff
@@ -0,0 +1,1221 @@
...
+        Ok(())
+    }
+
+    /// Drop the snapshot at `checkpoint`. Returns `true` if a
+    /// snapshot was removed.
+    ///
+    /// Outstanding [`Snapshot`](crate::Snapshot) values for this
+    /// checkpoint remain usable until they themselves drop; only the
+    /// buffer's reference is released.
+    pub fn drop_snapshot(&self, checkpoint: u64) -> bool {
```
amnn, line 711: What would the intended use case be for this? It has potential to mess up the guarantees provided by `snapshot_range`.

crates/sui-consistent-store/src/restore.rs
bmwill: I did start to look at how to do that but decided that was extra complexity that could be done later

crates/sui-consistent-store/src/batch.rs
```diff
@@ -0,0 +1,607 @@
...
+        Ok(self)
+    }
+
+    /// Commit the staged operations atomically.
+    ///
+    /// Consumes `self`. On success, all staged operations are
+    /// visible to subsequent reads. On failure, the database is
+    /// left in the state it was in before the commit was attempted.
+    pub fn commit(self) -> Result<(), Error> {
+        self.db.rocksdb().write(self.inner)?;
```
amnn, line 272: IIUC, `self.db` is only used during commit, and could be taken by reference here -- could we therefore keep `Batch` as a wrapper over `WriteBatch` and accept a `db: &Db` in `commit` and `commit_opt`?

The other APIs are already doing something similar by taking a `DbMap` so it seems incongruous that `commit` and `commit_opt` do not.

crates/sui-consistent-store/src/iter.rs
```diff
@@ -0,0 +1,476 @@
...
+    let mut next = prefix.to_vec();
+    while let Some(last) = next.last_mut() {
+        if *last < 0xFF {
+            *last += 1;
+            return Some(next);
+        }
+        next.pop();
+    }
+    None
+}
```
amnn, line 247: I had found the in-place version of this quite convenient in many cases, and avoided an allocation:

https://github.com/MystenLabs/sui/blob/286f8a7454a6af620e963dacefb833f3cd8118cb/crates/sui-indexer-alt-consistent-store/src/db/key.rs#L38-L47

crates/sui-consistent-store/src/map.rs
```diff
@@ -0,0 +1,1740 @@
...
+    /// a `MissingColumnFamily` error one read later) would mask the
+    /// underlying bug.
+    pub fn at(&self, snap: &Snapshot) -> DbMap<K, V, Snapshot> {
+        assert!(
+            self.reader.db().ptr_eq(snap.db()),
+            "snapshot was taken on a different Db than this DbMap is bound to",
+        );
+        DbMap {
+            reader: snap.clone(),
+            cf_name: self.cf_name,
```
amnn, line 238: nit: seeing as this function unconditionally takes ownership of the snapshot, could we accept it by value so that a caller that can move the snapshot in can avoid an extra clone?

```suggestion
    pub fn at(&self, snap: Snapshot) -> DbMap<K, V, Snapshot> {
        assert!(
            self.reader.db().ptr_eq(snap.db()),
            "snapshot was taken on a different Db than this DbMap is bound to",
        );
        DbMap {
            reader: snap,
            cf_name: self.cf_name,
```

crates/sui-consistent-store/src/map.rs
```diff
@@ -0,0 +1,1740 @@
...
+    /// Useful for callers that need to mix typed writes into a
+    /// pre-existing [`rocksdb::WriteBatch`] supplied from outside
+    /// this crate. The write does not become visible until the
+    /// caller commits `write_batch` themselves; for the routine
+    /// "stage and commit" path use [`Batch::put`](crate::Batch::put)
+    /// instead.
+    ///
+    /// Constrained to a [`Db`]-bound handle: writes always land at
+    /// the live tip.
+    pub fn stage_put(
```
amnn: There seems to be some duplication of logic between here and the `batch` module -- do we need both?

crates/sui-consistent-store/src/map.rs
amnn, line 1: The use of multiple impl blocks is quite difficult to follow in this file -- there's so many `K`s and `V`s with different implicit constraints that when I'm looking at any given function, it's hard to keep track.

I would find it easier to follow if we had two blocks: One for the generic case, with whatever bounds were truly common declared at the top, and constraints on K and V declared on each function; then another for the specialisation on `R = Db`.

crates/sui-consistent-store/src/snapshot.rs
```diff
@@ -0,0 +1,734 @@
...
+    }
+
+    /// Borrowed handle to the auto-registered
+    /// [`FrameworkSchema`](crate::FrameworkSchema), bound to this
+    /// snapshot.
+    ///
+    /// Returns a `FrameworkSchema<&Snapshot>` borrowing `self`.
+    /// Zero `Arc` bumps; the returned schema reads the framework
+    /// CFs at the snapshot's captured state.
+    pub fn framework(&self) -> crate::FrameworkSchema<&Snapshot> {
```
amnn, line 161: What does this get used for? Usually the framework is the one thing you don't want to fetch against a snapshot.

crates/sui-consistent-store/src/store.rs
```diff
@@ -0,0 +1,595 @@
...
+            committer_watermark::from_committer(watermark),
+        ));
+        Ok(true)
+    }
+}
+
+#[async_trait]
+impl<S: Send + Sync> SequentialConnection for Connection<'_, S> {}
+
+#[cfg(test)]
```
amnn, line 350: @evan-wall-mysten introduced some macros to automate conformance tests for stores. Could we use that here?

crates/sui-consistent-store/src/synchronizer.rs
```diff
@@ -0,0 +1,532 @@
...
+        let init_checkpoint = self
+            .last_watermarks
+            .values()
+            .map(|w| w.map_or(self.first_checkpoint, |w| w.checkpoint_hi_inclusive))
+            .max()
+            .expect("non-empty by ensure! above");
+        let next_snapshot_checkpoint = ((init_checkpoint / stride) + 1) * stride;
+
+        let mut queue: Queue = HashMap::new();
+        let mut join_set = JoinSet::new();
```
amnn, line 249: Out of interest, was there a reason to avoid using `Service` here? It should behave largely like a `JoinSet` most of the time, but is easier to compose with join sets from multiple other services

amnn (approved):
Looks great! Comments are a mix of consistency nits, and questions based on my understanding of the diff between this implementation and the `-alt` implementation.

crates/sui-consistent-store/src/db.rs
bmwill: Let me check on this

crates/sui-consistent-store/src/map.rs
bmwill: yeah i have this removed locally

crates/sui-consistent-store/src/encode_buf.rs
bmwill: Its mostly to just be extra sure, a clear is just a single usize write anyway so its cheap

crates/sui-consistent-store/src/db.rs
bmwill: yeah this isn't needed, i'll remove.

crates/sui-consistent-store/src/db.rs
bmwill: yeah i'll make this an error

PR #26864, opened 2026-06-01
[move-prompt] Initial sui-prompt implementation

awelc:
## Description 

The idea behind `sui prompt` is for it to serve as an endpoint to access expert Sui-related and Move-related knowledge. Currently it contains an initial set of (self-describing) skills that are divided into (self-describing categories), though multiple skills can be utilized by multiple categories. The idea behind is that an agent cannot simply ingest ALL this knowledge due to context size limitations. Instead it should ask for help in some particular aspect of working with Sui and Move (writing Move contracts, testing, debugging, auditing, etc.) and receive guidance on how to proceed with their desired workload

This is a PR adding `sui prompt` command to Sui CLI with initial set of bytecode analysis skills and auditing skills (the latter derived from Mysten Labs Agent skills at https://github.com/MystenLabs/skills), with the intention of additional categories and skills being added when needed.

## Test plan 

Testing is currently focusing on auditing and this version is capable of discovering some simple vulnerabilities that nevertheless resulted in loss of funds in the past


external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/reading-disassembly.md
```diff
@@ -0,0 +1,91 @@
...
+Disassembly is faithful but low-level: a stack machine with numbered locals and basic blocks.
+
+```sh
+sui move disassemble path/to/module.mv
+```
+
+## Worked example (real output)
+
+```
+// Move bytecode v6
```
damirka: I don't think we need output, it pollutes context.

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/reading-decompiled.md
```diff
@@ -0,0 +1,77 @@
...
+findings to humans.
+
+```sh
+"$MOVE_BIN" decompile --input <dir-or-file.mv> --output <outdir>
+# -> <outdir>/<package>/<module>.move, each starting with "// Auto-generated by Move decompiler"
+```
+
+## Worked example (real output)
+
+```move
```
damirka: Not sure this output is needed here.

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/reading-decompiled.md
```diff
@@ -0,0 +1,77 @@
+# Reading `move decompile` output
```
damirka: I don't think it should be "reading", more "decompiling" with step by step + caveats

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/reading-disassembly.md
```diff
@@ -0,0 +1,91 @@
...
+| `CastU8..CastU256` | integer width cast — **silent truncation site (SM-F1)** |
+| `Add/Sub/Mul/Div/Mod`, `Eq/Lt/...` | arithmetic / comparison (over/underflow aborts natively) |
+| `VecPack(tok, n)` / `VecPushBack` / `VecMutBorrow` | vector build / mutate |
+| `Branch/BrTrue/BrFalse b` | jump to block `b` |
+| `Abort` | pop abort code and abort (the code is the constant just loaded) |
+| `Ret` | return |
+
+## Audit moves from disassembly
+
+- **Transfer variant:** `Call transfer::transfer<ClaimCap>` (module-internal, no `store` needed) vs
```
damirka: Perhaps, it should say that any function in the `0x2::transfer` that has a public variant falls here. It's not just `transfer`, but also `share_object` and `freeze_object`, and `receive`

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/SKILL.md
```diff
@@ -0,0 +1,75 @@
+---
+name: move-bytecode-comprehension
+description: >
+  Use when reading or reasoning about compiled Move bytecode, `sui move disassemble` output, or
```
damirka: Too long for agents, should be shorter.

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/SKILL.md
```diff
@@ -0,0 +1,75 @@
...
+| `format.md` | The binary format: tables, abilities, visibility, signatures, opcode categories. Read once. |
+| `reading-disassembly.md` | Interpreting `sui move disassemble` (stack machine, locals, the opcodes you'll actually see). |
+| `reading-decompiled.md` | Use to translate **already-confirmed** findings into readable Move for human explanation; understand the decompiler's artifacts so you don't misread them when presenting. |
+
+## Practical stance
+
+- **Analyze on disassembly** — it is faithful to the executed bytecode (1:1 with what the validator
+  runs). Every `SM-*` finding's evidence must be an assembly excerpt; cite as
+  `<module>.asm:B<block>@i<index>`.
+- **Use decompiled source only to *explain* confirmed findings to humans**, never to derive them.
```
damirka: I'm not sure about this part. While we cannot be sure that decompiled version matches exact source, decompiled output consumes less tokens than deserialised.

crates/sui-prompt/src/skills/sui-and-move-tools/fetch-and-decompile.md
```diff
@@ -0,0 +1,93 @@
...
+> for human readers.** Do not derive findings from `.move`.
+
+## 1. Get the `.mv` modules
+
+### Route A — fetch on-chain bytecode by package id
+
+Make sure the CLI is pointed at the right network first (`sui client active-env`; switch with
+`sui client switch --env mainnet` etc. — the package id is network-specific).
+
+```sh
```
damirka: There's much better option to fetch with GraphQL. Let me attach a gist that my agents use.

Here's a gist. Perhaps, can be extended slightly and converted into a page on graphql: https://gist.github.com/damirka/5d47e1cf385323339c808741761d68f7

external-crates/move/crates/move-cli/src/prompt/skills/sui-and-move-tools/fetch-and-decompile.md
```diff
@@ -0,0 +1,93 @@
...
+
+```sh
+"$MOVE_BIN" decompile --input "$OUT/mv" --output "$OUT/decompiled"
+# Recurses the input dir for *.mv; deserializes each module; tolerates missing dependencies
+# (allow_missing_dependencies = true), so a single package decompiles without its deps.
+# Output: $OUT/decompiled/<package>/<module>.move
+ls -R "$OUT/decompiled"
+```
+
+The decompiled `.move` is reached for **only after a finding is confirmed on the assembly**, to
```
damirka: I think language didn't change after generalisation. I'm not sure what findings here mean.

external-crates/move/crates/move-cli/src/prompt/skills/sui-and-move-tools/setup.md
```diff
@@ -0,0 +1,115 @@
+# Setup — clean toolchain via `suiup` + fresh `move` build
+
+Run once per session, then reuse `$MOVE_BIN` IF it was built in *this* `$WORK_DIR` from the
+currently-confirmed `SUI_REF`. See `SKILL.md` for the **Isolation** rule, `SUI_REF`, and
+`WORK_DIR`. **Never** reuse a binary or checkout from anywhere else.
+
+## 1. Install / verify a clean `sui` via `suiup`
```
damirka: somewhere here we want to suggest to install formatter or link to it, it's a part of basic setup

external-crates/move/crates/move-cli/src/prompt/skills/sui-and-move-tools/SKILL.md
```diff
@@ -0,0 +1,102 @@
+---
+name: sui-and-move-tools
+description: >
+  Tools for working with Sui Move: the **`suiup`-managed `sui` CLI** for fetching on-chain
```
damirka: long description, potentially annoying behaviour from an llm

crates/sui-prompt/src/skills/sui-and-move-tools/SKILL.md
```diff
@@ -0,0 +1,102 @@
+---
+name: sui-and-move-tools
```
damirka, line 2: what does this skill do?

damirka: Given the contents, I think there can be some changes here

external-crates/move/crates/move-cli/src/prompt/skills/sui-move-security-review/evals/evals.json
```diff
@@ -0,0 +1,116 @@
+{
```
damirka: What is this file?

external-crates/move/crates/move-cli/src/prompt/categories/audit/CATEGORY.md
```diff
@@ -0,0 +1,81 @@
+---
+name: audit
+description: Reviewing a deployed Sui Move package for vulnerabilities, exploitable bugs, suspicious behavior, or anything suspected wrong on chain. Reach for this on "is this package safe?", "what's wrong with this?", "I suspect there's a bug in X", "find vulnerabilities", "security review", or a direct "audit this".
```
damirka: We don't use "Sui Move", it's just Move

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/reading-disassembly.md
awelc: Good point!

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/reading-decompiled.md
awelc: Good point. Initial design point was that this could be used by humans (where it could be useful) and agents (where it would not while increasing context window size) but realistically, no human will read these skill files directly from `move prompt`

external-crates/move/crates/move-cli/src/prompt/skills/sui-move-security-review/evals/evals.json
awelc: This was added when deriving auditing skills from https://github.com/MystenLabs/skills, inspired by the other repository structure, but it's indeed unnecessary here

external-crates/move/crates/move-cli/src/prompt/categories/audit/CATEGORY.md
awelc: Great point. Applied this more broadly as there we more occurrences...

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/SKILL.md
awelc: Tightened it up

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/SKILL.md
awelc: The goal was to operated on the most reliable source of truth, but considering that our decompiler output is pretty reliable itself, and the token-related gains, it is indeed the right thing to switch to decompilation-first approach

crates/sui-prompt/src/skills/sui-and-move-tools/fetch-and-decompile.md
awelc: True. Since this is not yet covered by official docs/skills I included what's necessary for this in here rather than linking to external source

external-crates/move/crates/move-cli/src/prompt/skills/sui-and-move-tools/setup.md
awelc: The setup file is now completely gone. With addition of GraphQL to obtain disassembled bytecode, we no longer even need the `sui` binary. The idea is to have only tools that are needed by current skills/categories, and add additional ones when they are actually needed

external-crates/move/crates/move-cli/src/prompt/skills/sui-and-move-tools/fetch-and-decompile.md
awelc: This whole skill has been seriously re-written and this was fixed in the process

crates/sui-prompt/src/skills/sui-and-move-tools/SKILL.md
awelc: Agreed! This skill evolved into this shape by being split from audit in the attempt to generalize it. It is also to verbose in the sense of overlap with existing docs. It has been significantly rewritten to be much shorter (especially after dropping `sui` CLI) and more of a placeholder for future tools addition

external-crates/move/crates/move-cli/src/prompt/skills/sui-and-move-tools/SKILL.md
awelc: As per the other [comment](https://github.com/MystenLabs/sui/pull/26864/changes#r3342909315) for this file, this skill is now super short and this particular part is dropped

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/reading-disassembly.md
awelc: Good catch! However, this section breaks the skills independence principle and should really be part of audit skill. In fact, audit skill already contains most of this info making it redundant. The suggested change is still applicable there, though, where it has been indeed applied

external-crates/move/crates/move-cli/src/prompt/skills/move-bytecode-comprehension/reading-decompiled.md
awelc: Renamed to (to "decompile", though, to better match also renamed "disassembly") and finessed the content

damirka:
@awelc few notes after running it locally (had to modify it to work as a part of `sui` binary):

1. Shall we add it to debug builds of `sui`? I don't see it working as a separate tool, we need full bundle of `sui` to have agents build dapps + use `prompt`
2. I think that we should stick to more or less standard CLI format in the `move prompt` command. Right now it outputs md with some information not making much sense, eg: `## Adding categories and skills` which references local paths in `sui` repo while the tool can be used in any directory
3. I think we can omit notes like "agent agnostic" and so on, these seem to be selling points for humans
4. formatting in `move prompt categories` feels too much, I believe it should be a concise list with shorter descriptions than we have today
5. Overall seems like we're overly verbose at every step of the way

Still trying to get the agent to use it, not clear how they would discover it on their own.

crates/sui-move/src/decompile.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Mysten Labs, Inc.
```
damirka: Should this be a separate PR? Perhaps, easier to land.

crates/sui/src/sui_commands.rs
```diff
@@ -377,6 +378,10 @@ pub enum SuiCommand {
         cmd: sui_move::Command,
     },
 
+    /// Expert Move knowledge for AI agents (agent-agnostic).
+    #[clap(name = "prompt")]
+    Prompt(sui_prompt::Prompt),
```
damirka, line 385: Yay!

crates/sui-move/src/decompile.rs
awelc: Good point. I thought it's such a small change that we can sneak it in here but you are right in that there are more stakeholders for this change than for the sui-prompt PR landing as sui-prompt is pretty much standalone (and gated)

awelc:
> @awelc few notes after running it locally (had to modify it to work as a part of `sui` binary):
> 
> 1. Shall we add it to debug builds of `sui`? I don't see it working as a separate tool, we need full bundle of `sui` to have agents build dapps + use `prompt`
> 2. I think that we should stick to more or less standard CLI format in the `move prompt` command. Right now it outputs md with some information not making much sense, eg: `## Adding categories and skills` which references local paths in `sui` repo while the tool can be used in any directory
> 3. I think we can omit notes like "agent agnostic" and so on, these seem to be selling points for humans
> 4. formatting in `move prompt categories` feels too much, I believe it should be a concise list with shorter descriptions than we have today
> 5. Overall seems like we're overly verbose at every step of the way
> 
> Still trying to get the agent to use it, not clear how they would discover it on their own.

Moved the whole thing to be `sui move` and it's now gated to work only in debug builds. Also fixed and tightened up both category files (pretty dramatically) and skills files, as well the README files.

damirka:
To reiterate on the approach:

- let's make it `sui prompt`
- let's open it up in release builds
- let's omit decompiler for the time being

The sooner we get it out, the sooner we can start fine tuning it.

crates/sui-move/src/decompile.rs
damirka: Ping: should we remove decompiler from this PR?

awelc:
> To reiterate on the approach:
> 
> * let's make it `sui prompt`
> * let's open it up in release builds
> * let's omit decompiler for the time being
> 
> The sooner we get it out, the sooner we can start fine tuning it.

On it!

crates/sui-move/src/decompile.rs
awelc: Indeed. Also on it

crates/sui-prompt/build.rs
```diff
@@ -0,0 +1,257 @@
+// Copyright (c) Mysten Labs, Inc.
```
tzakian, line 1: Thoughts on using `include_dir` for this? At least from as far as I can till it would work here possibly?

crates/sui-prompt/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+use clap::{Parser, Subcommand};
+use std::collections::BTreeSet;
+
+// build.rs generates this file with `SKILL_FILES: &[PromptSkillFile]` and
+// `CATEGORIES: &[PromptCategory]`.
+include!(concat!(env!("OUT_DIR"), "/embedded.rs"));
+
+/// The `sui prompt` no-args overview rendered to the agent. Source is
+/// `prompt-output.md` next to this file, embedded at build time via `include_str!`.
+const OVERVIEW: &str = include_str!("prompt-output.md");
```
tzakian, line 29: This is where I wonder if you could just use `include_dir` I think.

crates/sui-prompt/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+}
+
+impl Prompt {
+    /// Dispatch the parsed subcommand. Prints to stdout; never writes to the filesystem.
+    /// Returns an error only for user-input failures (unknown bundle / unknown reference
+    /// file); successful renders return `Ok(())`.
+    pub fn execute(self) -> anyhow::Result<()> {
+        match self.cmd {
+            None => {
+                print!("{}", OVERVIEW);
```
tzakian: nit: should this be a `println!`? or where is the newline coming from otherwise?

crates/sui-prompt/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+    let target_file = match file {
+        Some(name) => normalize_skill_file(name),
+        None => "SKILL.md".to_string(),
+    };
+    match SKILL_FILES
+        .iter()
+        .find(|s| s.bundle == bundle && s.file == target_file)
+    {
+        Some(s) => {
+            print!("{}", s.content);
```
tzakian: Similar question as above re: `println!` here

crates/sui-prompt/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+            .find(|s| s.bundle == bundle && s.file == file)
+        else {
+            // build.rs guarantees the file exists; defensive in case the embedded
+            // table ever drifts. Skip silently rather than panicking mid-stream.
+            continue;
+        };
+        print!("{}", s.content);
+        // Normalize to a blank line separating the file from the next `#` heading,
+        // regardless of whether the source ends with `\n`, `\n\n`, or no newline.
+        if !s.content.ends_with("\n\n") {
```
tzakian: would it be better to `trim_end_matches('\n')` and then just always output a blank line? This then fixes the case where the file may have a large number of newlines at the end of it too (i.e., 3+).

crates/sui-prompt/src/lib.rs
```diff
@@ -0,0 +1,658 @@
...
+        "Bundles in category '{}' ({} bundles, {} chars):",
+        c.name,
+        c.skills.len(),
+        grand_total
+    );
+    for (bundle, files) in &bundles {
+        let bundle_total: usize = files.iter().map(|(_, n)| *n).sum();
+        let max_name = files
+            .iter()
+            .map(|(f, _)| f.strip_suffix(".md").unwrap_or(f).len())
```
tzakian: This strip + unwrap_or is called enough that lifting it into a helper may be nice.

tzakian (commented):
Overall looks good. I left a couple technical questions and suggestions.

crates/sui-prompt/src/lib.rs
awelc: Makes sense. Done!

crates/sui-prompt/src/lib.rs
awelc: It would be! Applied here and in another place where we embed files (via a newly introduced helper)

crates/sui-prompt/src/lib.rs
awelc: The assumption was that the file will end with newline but it indeed does not have to. Generalized (via a newly inrroduced helper function) to do the right thing regardless (here and in other places)

crates/sui-prompt/src/lib.rs
awelc: Fixed!

crates/sui-prompt/build.rs
awelc: Are you suggesting replacing build.rs altogether? I really like the idea of doing some validation at build time rather than suck the whole directory tree and do it at runtime. Even if using `include_dir` and keeping build.rs I like the ability to filter unneeded content (e.g., I left some Emacs temps during development and with the current validation at build time they will never get included). Of course we can re-do this kind of filtering on content ingested via `include_dir` but I am just not seeing that much benefit from switching to `include_dir`...

crates/sui-prompt/src/lib.rs
awelc: It's just one file... Unless this is additional explanation on the other [comment](https://github.com/MystenLabs/sui/pull/26864#discussion_r3480371467), and it suggests ditching build.rs altogether. I tried to explain my reasoning for not doing this in said comment

tzakian: ah sorry sorry. I meant this as additional explanation on the comment above about include_dir

tzakian (approved):

PR #26865, opened 2026-06-01
composite workload: regression test for IFFW gas-smashing underflow

mystenmark:
## Summary

Adds a randomized mixed-payment branch to the composite workload so that the IFFW (`InsufficientFundsForWithdraw`) gas-smashing settlement path is exercised continuously by simtest. With some probability, the gas builder constructs `gas_data.payment` as a `[real_coin, coin_reservation]` pair in randomized order (50/50 swap), using a reservation amount roughly an order of magnitude above the per-account address balance.

This is the path where the recent settlement-underflow bug lived: when a tx with mixed payment was cancelled with IFFW, the AB leg of gas smashing emitted a Split against an already-depleted accumulator, and `0x2::accumulator::update_u128` would abort during settlement — a settlement-tx-cannot-fail invariant violation.

## Verification

- Default `mixed_gas_payment_probability = 0.1` keeps composite throughput in range (~97s vs 67s baseline, still passing the workload's signed-tx-count assertion).
- With both IFFW gates locally short-circuited (simulating the pre-fix state), the same workload trips the `settlement transaction cannot fail` panic in `settlement_scheduler.rs:380` — the underflow inside `0x2::accumulator::update_u128`. This was reproduced once during development and the panic shape matches the original incident.
- With the upstream IFFW handling intact, the workload passes.

## Test plan

- [x] `cargo simtest -p sui-benchmark --test simtest test_composite_workload`
- [x] `cargo xclippy -D warnings -p sui-benchmark`
- [x] `cargo fmt`

### DIFF
--- a/crates/sui-benchmark/src/workloads/composite/mod.rs
+++ b/crates/sui-benchmark/src/workloads/composite/mod.rs
@@ -36,7 +36,10 @@ use std::time::Duration;
 use sui_protocol_config::ProtocolConfig;
 use sui_test_transaction_builder::TestTransactionBuilder;
 use sui_types::TypeTag;
+use sui_types::accumulator_root::AccumulatorValue;
+use sui_types::balance::Balance;
 use sui_types::base_types::{ObjectID, ObjectRef, SequenceNumber, SuiAddress};
+use sui_types::coin_reservation::ParsedObjectRefWithdrawal;
 use sui_types::crypto::{AccountKeyPair, get_key_pair};
 use sui_types::digests::TransactionDigest;
 use sui_types::gas_coin::GAS;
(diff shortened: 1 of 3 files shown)


crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -958,6 +977,42 @@ impl Payload for CompositePayload {
                     current_epoch,
                     nonce,
                 )
+            } else if !address_balance_gas_disabled
+                && rng.gen_bool(self.config.mixed_gas_payment_probability as f64)
+            {
+                used_gas.push(i);
+                // Build a gas payment mixing 1-2 real coins with 1-2 coin-reservation digest
+                // entries, then shuffle the whole vector. Exercises the gas-smashing path
```
aschran: do we want to be referencing specific historical bugs in comments here (as opposed to just talking about what systems are being exercised)? I think there would be a LOT more comments in this workload if we were doing this consistently

(doesn't necessarily matter much in this particular spot, but I do worry about the proliferation of Claude-slop comments across the codebase)

aschran (approved):

crates/sui-benchmark/src/workloads/composite/mod.rs
mystenmark: yeah this is just more claude comment slop that i failed to eradicate. i really can't stand the way it writes comments these days, thanks for calling it out

mystenmark: Good call — dropped the bug narrative from the code comments (kept it in the commit message, which is where the history belongs) and tightened them to describe the mechanism under test for future readers.

PR #26867, opened 2026-06-01
Bound total gas payment to i64::MAX in check_gas

mystenmark:
Enforces that the combined total of all gas-payment values (coin balances plus reservation amounts) fits in `i64`, rejecting transactions that exceed it during `check_gas`. This guarantees the `i64` conversions in `smash_gas` cannot overflow.

Also moves the SUI-coin-type and ownership checks inline into the gas-payment loop so they apply consistently across execution, dry-run, and simulate paths, and unifies `validity_check` / `validity_check_no_gas_check` with mock-gas injection reordered so gas budget and price are validated consistently in the simulate/dev-inspect path.

Adds tests for the total-bound rejection, dry-run gas budget/price validation, and rejection of non-SUI gas reservations.

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -2417,9 +2417,9 @@ impl AuthorityState {
             .into());
         }
 
-        // Compute input/receiving object kinds before mock gas injection so the mock
-        // gas reference is not included in input_object_kinds (it is added to
-        // input_objects directly after object loading).
+        // Full validity check including gas budget and price.
+        transaction.validity_check(&epoch_store.tx_validity_check_context())?;
+
         let input_object_kinds = transaction.input_objects()?;
         let receiving_object_refs = transaction.receiving_objects();
 
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -2417,9 +2417,9 @@ impl AuthorityState {
             .into());
         }
 
-        // Full validity check including gas budget and price.
-        transaction.validity_check(&epoch_store.tx_validity_check_context())?;
-
+        // Compute input/receiving object kinds before mock gas injection so the mock
+        // gas reference is not included in input_object_kinds (it is added to
+        // input_objects directly after object loading).
         let input_object_kinds = transaction.input_objects()?;
         let receiving_object_refs = transaction.receiving_objects();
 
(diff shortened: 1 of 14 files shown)


crates/sui-transaction-checks/src/lib.rs
```diff
@@ -432,21 +439,45 @@ mod checked {
...
+                        }
+                        .into());
+                    }
+                    if gas_ownership_checks && object.owner != Owner::AddressOwner(gas_owner) {
+                        return Err(UserInputError::GasObjectNotOwnedObject {
+                            owner: object.owner.clone(),
+                        }
+                        .into());
+                    }
+                    total += sui_types::gas::get_gas_balance(object)? as u128;
```
alex-mysten, line 461: Does this need to be gated? It looks like check_gas is called both during admission & execution.

crates/sui-e2e-tests/tests/address_balance_compatibility_tests.rs
```diff
@@ -2316,36 +2283,481 @@ async fn test_fake_coin_reservation_dev_inspect_safe_when_flag_disabled() {
...
+    assert!(
+        executed.effects.status().is_ok(),
+        "tx should succeed, got: {:?}",
+        executed.effects.status()
+    );
+}
+
+// Gas-payment coin reservations must be for SUI. A reservation for a custom coin type
+// (even one with an oversized balance) must be rejected before reaching smash_gas.
+// validity_check_no_gas_check now enforces the SUI-type constraint on all paths
```
alex-mysten: stale comment?

crates/sui-transaction-checks/src/lib.rs
mystenmark: added gating - i think it probably wasn't necessary in practice but better to just Do It Right

crates/sui-core/src/authority.rs
```diff
@@ -2452,6 +2452,7 @@ impl AuthorityState {
         // Full validity check including gas budget and price.
         transaction.validity_check(&epoch_store.tx_validity_check_context())?;
 
+
```
tnowacki, line 2455: ?

crates/sui-types/src/transaction.rs
```diff
@@ -3465,6 +3465,7 @@ impl TransactionDataAPI for TransactionDataV1 {
         Ok(())
     }
 
+
```
tnowacki, line 3468: ?

PR #26870, opened 2026-06-02
[trace-view] Hardening against source/bytecode mismatches

awelc:
## Description 

Currently, when debugging at the source level, it is assumed that the debug artifacts generated from source (source map aka debug info) matches bytecode on chain (and in the trace) exactly. In case of a mismatch (e.g., a function missing in the source by available in bytecode) the debugger throws an error and does not even start debugging session (as these types of errors are typically detected and the trace ingestion time).

This is suboptimal as the user may want to continue debugging even if not all debug metadata is correct. Furthermore, they may even only need to debug their issue up to a point before the first mismatch manifests itself (in which case they wouldn't even notice any debugging-related problems despite mismatches existing).

The goal of this PR is to harden the implementation of the debugger to be more permissive in terms of handling mismatches. The basic strategy is to revert to bytecode level debugging when source-level information is absent or unreliable and issue a warning to the user informing them about the situation.

In terms of implementation, existing error at trace ingestion time is removed, and a "diagnostic" is inserted into translated event stream to be sent to the client when a given portion of the trace is encountered during debugging session. For all cases covered in this PR, the "diagnostic" is at the level of `OpenFrame` event and that's where we revert to bytecode as well.

At the low level, the way it's handled is that upon encountering a mismatch (sometimes in the middle of a given function's trace being parsed, we do the following:
- find "real" (as opposed to "virtual" for inlined macros) `OpenFrame` for this function 
- mark "real" `OpenFrame` as "force disassembly"
- keep going with parsing the trace until encountering `CloseFrame for this function
- at `CloseFrame`, if its corresponding `OpenFrame` is "force disassembly" walk events of the whole function and mark its vritual frames as "force disassembly" as well (but only virtual frames of this function and not, for example, of functions it calls)

The idea here is that the whole function where the mismatch was encountered should be conservatively debugged at the bytecode level, but the functions it calls don't have to be.

The two explicit mismatches of on-chain bytecode vs source code are missing function and missing module. The other two cases are more of a hardening against inconsistent source code debugging artifacts. Both of these types of problems can happen when it's the user's responsibility to provide source code debugging artifacts to the debugger.

Of course there are many more errors that the debugger can throw, but as far as I can tell these are either unrecoverable or they shouldn't actually happen due to how the internal representation is constructed (in that sense,  they are more like assertions).

## Test plan 

New tests have been added. These are synthetic tests (generated with the help of AI) but their behavior is verified both in terms of analyzing test harness output and the behavior of each one of them in the actual debgugger


external-crates/move/crates/move-analyzer/trace-adapter/src/trace_utils.ts
```diff
@@ -893,6 +897,140 @@ export async function readTrace(
...
+            bcodeFilePath = currentBCodeFile.path;
+        }
+    }
+
+    let forceDisassembly = false;
+    let diagnostics = undefined;
+    if (!srcFunEntry) {
+        if (!bcodeMap || !bcodeFunEntry) {
+            // if we have neither source nor bytecode debug info for the function,
+            // then we have a real problem
```
tzakian, line 997: Indeed we do! :)

tzakian (commented):
👍🏻

tzakian (approved):
Looks good!

PR #26873, opened 2026-06-03
[grpc] VM Error Visibility GRPC: simulate_transaction, get_transaction, get_checkpoint

jordanjennings-mysten:
## Description 

Add execution error metadata to simulate_transaction, get_transaction, get_checkpoint responses

## Test plan 

crates/sui-e2e-tests/tests/rpc/v2:
- get_checkpoint_exposes_execution_error_metadata
- get_transaction_exposes_execution_error_metadata
- simulate_transaction_exposes_execution_error_metadata
---

## Release notes

- [ ] Protocol: 
- [x] gRPC: include error metadata in simulate, get_transaction, get_checkpoint starting with "source" field
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


sui-execution/src/lib.rs
```diff
@@ -20,9 +20,9 @@ mod v1;
 mod v2;
 mod v3;
 
+mod error_context;
```
amnn: nit: this belongs in its own block before the version modules but after `pub mod executor` and `pub mod verifier`.

crates/sui-rpc-api/src/client/mod.rs
```diff
@@ -663,6 +664,8 @@ pub struct ExecutedTransaction {
     pub signatures: Vec<GenericSignature>,
     pub effects: TransactionEffects,
     pub clever_error: Option<proto::CleverError>,
+    #[serde(skip_serializing_if = "execution_error_metadata_is_empty")]
```
amnn: Nothing else seems to be skipped if it's empty -- why do we do that with `execution_error_metadata`?

amnn (commented):
This looks reasonable to me but I can only give a surface level review. @bmwill and @nickvikeras are the experts over here, so I'll defer to them for the final review and stamp.

PR #26874, opened 2026-06-03
[typed-store] Add DBMap::snapshot_iterator for point-in-time iteration

andll:
## Summary

Adds `DBMap::snapshot_iterator()` and `DBMap::snapshot_iterator_with_bounds(lower, upper, reverse)` — iterators over a **consistent point-in-time view** of a table, with optional bounds and direction.

- **RocksDB / in-memory**: their iterators are already snapshot-consistent from the moment they're created, so this just delegates to the regular `safe_iter_with_bounds` / `reversed_safe_iter_with_bounds`.
- **tidehunter**: the live iterator is *not* stable under concurrent writes, so this opens a short-lived `Db::checkpoint()` and iterates that frontier instead. The returned iterator owns the checkpoint, pinning the snapshot for its lifetime.
- Bound semantics match the existing iterators: forward `[lower, upper)` (upper exclusive), reverse `[lower, upper]` (both inclusive).

All tidehunter-specific code is `#[cfg(tidehunter)]`-gated; the rocksdb build is unchanged.

Bumps the tidehunter pin to `7dc059d`, which provides the `Db::checkpoint()` API (used only on the tidehunter path).

## Test plan

- `cargo check -p typed-store` (default/rocksdb) — clean
- `cargo clippy -p typed-store --all-targets` (default) — clean
- `USE_TIDEHUNTER=1 cargo check/clippy -p typed-store --lib` — clean
- `USE_TIDEHUNTER=1 cargo check -p sui-core --lib` — compiles against the new rev

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -180,7 +180,7 @@ jobs:
           swap-size-gb: 256
       - name: cargo test with tidehunter
         run: |
-          RUST_BACKTRACE=1 USE_TIDEHUNTER=1 cargo nextest run --profile ci --cargo-quiet -p sui-core --features typed-store/tidehunter
+          RUST_BACKTRACE=1 USE_TIDEHUNTER=1 cargo nextest run --profile ci --cargo-quiet -p sui-core -p typed-store --features typed-store/tidehunter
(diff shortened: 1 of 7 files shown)


bmwill (approved):
lgtm

PR #26876, opened 2026-06-03
[consensus] add v3 FlexCommitter

mwtian:
## Description 

Add `FlexCommitter`, which calls `LeaderSlotDecider` to produce commits over multiple leaders per-round.

## Test plan 

CI

### DIFF
--- a/consensus/core/src/flex_committer.rs
+++ b/consensus/core/src/flex_committer.rs
@@ -1,0 +1,494 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// FlexCommitter is wired into Core in a follow-up PR.
+#![allow(dead_code)]
+
+use std::{collections::BTreeSet, sync::Arc};
+
+use consensus_config::{DIGEST_LENGTH, DefaultHashFunction};
+use consensus_types::block::{BlockRef, BlockTimestampMs, Round};
+use fastcrypto::hash::HashFunction as _;
+use itertools::Itertools as _;
+use parking_lot::RwLock;
+use rand::{SeedableRng as _, rngs::StdRng, seq::SliceRandom as _};
+
+use crate::{
+    BlockAPI, VerifiedBlock,
+    block::Slot,
+    commit::{Commit, CommitAPI, CommittedSubDag, LeaderStatus, TrustedCommit},
+    context::Context,
+    dag_state::DagState,
+    leader_schedule_v3::NextCommitLeaderSchedule,
+    leader_slot_decider::{INDIRECT_COMMIT_DEPTH, LeaderSlotDecider},
+};
+
+#[cfg(test)]
+#[path = "tests/flex_committer_tests.rs"]
+mod flex_committer_tests;
+
+/// FlexCommitter supports committing multiple and varying number of leaders per round,
+/// based on DagState and the schedule for next commit leaders.
+pub(crate) struct FlexCommitter {
+    context: Arc<Context>,
+    dag_state: Arc<RwLock<DagState>>,
+    slot_decider: LeaderSlotDecider,
+    pending_commit_state: PendingCommitState,
+}
+
+impl FlexCommitter {
+    pub(crate) fn new(context: Arc<Context>, dag_state: Arc<RwLock<DagState>>) -> Self {
+        let slot_decider = LeaderSlotDecider::new(context.clone(), dag_state.clone());
+        Self {
+            context,
+            dag_state,
+            slot_decider,
+            pending_commit_state: PendingCommitState::default(),
+        }
+    }
+
+    /// Attempts to create a commit based on the current DagState.
+    pub(crate) fn try_commit(
+        &mut self,
+        next_commit_leaders: NextCommitLeaderSchedule,
+    ) -> Option<(TrustedCommit, CommittedSubDag)> {
+        self.maybe_refresh_pending_commit_state(next_commit_leaders);
+
+        self.try_direct_commit();
+        if let Some(commit_leader_round) = self.find_commit_leader_round() {
+            return self.build_commit(commit_leader_round);
+        }
+
+        self.try_indirect_commit();
+        if let Some(commit_leader_round) = self.find_commit_leader_round() {
+            return self.build_commit(commit_leader_round);
+        }
+
+        None
+    }
+
+    fn maybe_refresh_pending_commit_state(
+        &mut self,
+        next_commit_leaders: NextCommitLeaderSchedule,
+    ) {
+        if self
+            .pending_commit_state
+            .next_commit_leaders
+            .next_commit_index
+            == next_commit_leaders.next_commit_index
+        {
+            // Use cached pending commit state for the same commit index.
+            return;
+        }
+        assert!(
+            self.pending_commit_state
+                .next_commit_leaders
+                .next_commit_index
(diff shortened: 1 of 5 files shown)


consensus/core/src/flex_committer.rs
```diff
@@ -0,0 +1,494 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// FlexCommitter is wired into Core in a follow-up PR.
```
akichidis: Let's make sure this comment is amended later

consensus/core/src/flex_committer.rs
```diff
@@ -0,0 +1,494 @@
...
+            .pending_commit_state
+            .next_commit_leaders
+            .min_next_leader_round;
+        let highest_accepted_round = self.dag_state.read().highest_accepted_round();
+
+        for leader_round in min_next_leader_round..highest_accepted_round {
+            let round_state = self
+                .pending_commit_state
+                .get_or_create_round_state(leader_round);
+            let slots_to_decide = round_state
```
akichidis, line 126: Looking on the code overall I see that we don't do any use of `debug` or `trace` statements, although we might benefit from that as we currently do on the current codebase. Maybe we could add some debug statements e.g. like the `slots_to_decide` per round etc?

consensus/core/src/flex_committer.rs
```diff
@@ -0,0 +1,494 @@
...
+            commit.reference(),
+        );
+        subdag.decided_with_local_blocks = false;
+        subdag
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+struct PendingCommitState {
+    next_commit_leaders: NextCommitLeaderSchedule,
```
akichidis, line 414: So the `NextCommitLeaderSchedule` will be used as a schedule for multiple commits , right?

consensus/core/src/flex_committer.rs
```diff
@@ -0,0 +1,494 @@
...
+        for r in add_from_round..=round {
+            let mut seed_bytes = [0u8; 32];
+            seed_bytes[28..].copy_from_slice(&r.to_le_bytes());
+            let mut rng = StdRng::from_seed(seed_bytes);
+            let mut leaders = self
+                .next_commit_leaders
+                .allowed_leaders
+                .choose_multiple(&mut rng, self.next_commit_leaders.num_leaders())
+                .cloned()
+                .collect::<Vec<_>>();
```
akichidis: Why is this needed? isn't the num of leaders the same as the set of the allowed leaders?

consensus/core/src/flex_committer.rs
```diff
@@ -0,0 +1,494 @@
...
+                parent_refs.insert(*ancestor);
+            }
+        }
+    }
+    let parent_refs: Vec<BlockRef> = parent_refs.into_iter().collect();
+    let blocks = dag_state
+        .get_blocks(&parent_refs)
+        .into_iter()
+        .map(|b| b.expect("Parent block must be in dag state"));
+    let ts = crate::linearizer::median_timestamp_by_stake(context, blocks)
```
akichidis, line 587: what about block equivocations? Since we have multiple leaders now it's possible to have equivocated ancestors here right? Maybe we want to make sure that we deterministically keep only one block per authority?

consensus/core/src/flex_committer.rs
```diff
@@ -0,0 +1,494 @@
...
+        // higher rounds to be decided first.
+        for round in (min_next_leader_round
+            ..=(highest_accepted_round.saturating_sub(INDIRECT_COMMIT_DEPTH)))
+            .rev()
+        {
+            let Some(anchor_block) = self.find_anchor_block(round + INDIRECT_COMMIT_DEPTH) else {
+                // Skip when there is no anchor block for this round yet.
+                continue;
+            };
+            self.decide_with_anchor_block(anchor_block.clone(), round);
```
akichidis, line 163: Maybe let's have a comment why we are relying on the first committed leader to decide with an anchor block.

consensus/core/src/flex_committer.rs
```diff
@@ -0,0 +1,494 @@
...
+    /// Attempts to create a commit based on the current DagState.
+    pub(crate) fn try_commit(
+        &mut self,
+        next_commit_leaders: NextCommitLeaderSchedule,
+    ) -> Option<(TrustedCommit, CommittedSubDag)> {
+        self.maybe_refresh_pending_commit_state(next_commit_leaders);
+
+        self.try_direct_commit();
+        if let Some(commit_leader_round) = self.find_commit_leader_round() {
+            return self.build_commit(commit_leader_round);
```
akichidis, line 62: Do I understand correctly that even though we might decide for slots of multiple rounds, in the end we'll only build the commit for the first decided round? So in practice the commit rule will re-run again?

consensus/core/src/flex_committer.rs
```diff
@@ -0,0 +1,494 @@
...
+        {
+            // Use cached pending commit state for the same commit index.
+            return;
+        }
+        assert!(
+            self.pending_commit_state
+                .next_commit_leaders
+                .next_commit_index
+                < next_commit_leaders.next_commit_index,
+            "next_commit_index should only move forward: {} vs {}",
```
akichidis, line 93: IIUC it is still possible to have a `pending_commit_state` with some `round_state` where the `round_state.num_committed > 0` and we us pushing forward the commit index right? But always the first one should have produced a committed sub dag? Just trying to see if that check here is enough on its own or some additional safeguards could be used.

consensus/core/src/flex_committer.rs
mwtian: Yes. `try_commit()` only takes in next commit's schedule, and will only produce next commit or None. When `try_commit()` receives the 2nd next commit schedule, it will reset the internal state of `FlexCommitter`. The commit rule will rerun with potentially different schedule.

consensus/core/src/flex_committer.rs
mwtian: I updated comment here and above RoundState. commit index should not get pushed forward until the commit with next_commit_index gets created. Then commit index will not get pushed further.
Also as a side, slots voted to be committed are different from blocks actually becoming leaders in commits, just like changing leader schedule right now can make "committed" leaders with round after leader schedule change no longer committed in the new schedule.

consensus/core/src/flex_committer.rs
mwtian: No, it will only be used for next commit. When a leader schedule for 2nd next commit is received, the schedule gets updated and round state resets.
The caller of `FlexCommitter::try_commit()` can choose to use the same set of `allowed_leaders` for a number of subsequent commits.

consensus/core/src/flex_committer.rs
mwtian: Added a few more logs and metrics.

consensus/core/src/flex_committer.rs
mwtian: Updated comment inside find_anchor_block().

consensus/core/src/flex_committer.rs
mwtian: Yeah this can be simplified now.

consensus/core/src/flex_committer.rs
mwtian: Good catch!

akichidis (approved):

PR #26884, opened 2026-06-04
fix(sui): `sui start` resumes the embedded fullnode (index over the fullnode gRPC)

hayes-mysten:
## Problem

`sui start` against an **existing persisted config dir** (created by `sui genesis -f`) loads `network.yaml` so validators resume from their `authorities_db` — but the embedded RPC fullnode was rebuilt from scratch every invocation and the `fullnode.yaml` genesis saved was never read back, so the fullnode re-synced the whole (growing) chain from genesis on every restart while the validator resumed:

```
fullnode:  executed=0    ... Creating new epoch start config from genesis ... execute_checkpoint{seq=0}
validator: executed=4433 ... Loading epoch start config from DB
```

On disk it leaked one orphaned `full_node_db/<key>/` per boot (~11M → ~173M over 16 boots, ≈ 1.4 GB), because each restart minted a new fullnode identity → new db dir.

## Root cause

- `crates/sui-swarm-config/src/node_config_builder.rs` — `FullnodeConfigBuilder::build()` mints a fresh random keypair every call and derives `db_path` from it; no way to give the fullnode a stable identity.
- `crates/sui/src/sui_commands.rs` (start) loaded `network.yaml` but never read back `fullnode.yaml` — `SwarmBuilder::build()` always rebuilt the fullnode fresh.
- Doubled db path: genesis built the fullnode with `with_config_directory(FULL_NODE_DB_PATH.into())`, so `build()` joined `full_node_db` again → relative `full_node_db/full_node_db/<key>` (validators pass the real config dir → single absolute level, which is why validators resumed and the fullnode didn't).

## The fix (scoped to `sui start` + `sui-swarm`; no shared-crate change)

- **Resume the embedded fullnode** — when `<config_dir>/fullnode.yaml` exists, load and run it via the new `SwarmBuilder::with_fullnode_config(NodeConfig)` instead of generating a fresh fullnode; the persisted identity + db path are preserved so it resumes from its db. The field defaults to `None`, so existing `SwarmBuilder` callers (tests, `test-cluster`, e2e) are unchanged.
- **Index over the fullnode's gRPC** — the embedded indexer + consistent-store now ingest from the fullnode's gRPC (`rpc_api_url = socket_addr_to_url(fullnode_rpc_address)`) instead of a local ingestion directory. A resumed fullnode only writes *new* checkpoints to a local dir, so local-file ingestion would strand a consumer that needs already-executed checkpoints; the fullnode serves them all over gRPC. **Enabling an indexer no longer implies/sets up a data-ingestion directory** (an explicit `--data-ingestion-dir` is still honored).
- **Fix the doubled genesis db path** — build the persisted fullnode with the real config dir (like validators) → absolute, single-level `<config_dir>/full_node_db/<key>`.
- **Regenerate non-RPC bind ports** (metrics/admin/network/p2p) on the persisted path so genesis-time ports that may now be occupied don't fail bind.

## Compatibility

- Cold boot / `--force-regenesis`: no persisted `fullnode.yaml` to reuse → existing fresh-build path, behavior unchanged.
- **No change to any shared crate** — `git diff upstream/main -- crates/sui-indexer-alt-framework` is empty. The diff is `crates/sui/src/sui_commands.rs` + `crates/sui-swarm/src/memory/swarm.rs` only.

## Notes

Three ingestion strategies were explored — resume→gRPC, hybrid (local primary + gRPC fallback), and always-gRPC. Per review, **always-gRPC** was chosen (simplest, no local-file ingestion lifecycle, no shared-framework change). The alternatives (#26887 hybrid, #26888 always-gRPC sibling) are closed in favor of this PR.

## Testing

`cargo check -p sui` / `-p sui-indexer-alt-framework`, `cargo clippy`, `cargo fmt --check`, and the framework ingestion tests pass; full CI is green on this commit (verified green on the identical #26888 before consolidation).

### DIFF
--- a/crates/sui/src/sui_commands.rs
+++ b/crates/sui/src/sui_commands.rs
@@ -33,7 +33,7 @@ use sui_bridge::sui_transaction_builder::build_committee_register_transaction;
 use sui_config::node::Genesis;
 use sui_config::p2p::SeedPeer;
 use sui_config::{
-    Config, FULL_NODE_DB_PATH, PersistedConfig, SUI_CLIENT_CONFIG, SUI_FULLNODE_CONFIG,
+    Config, FULL_NODE_DB_PATH, NodeConfig, PersistedConfig, SUI_CLIENT_CONFIG, SUI_FULLNODE_CONFIG,
     SUI_NETWORK_CONFIG, genesis_blob_exists, sui_config_dir,
 };
 use sui_config::{
@@ -48,7 +48,9 @@ use sui_indexer_alt_consistent_store::{
 };
 use sui_indexer_alt_framework::{
     IndexerArgs,
-    ingestion::{ClientArgs, ingestion_client::IngestionClientArgs},
+    ingestion::{
+        ClientArgs, ingestion_client::IngestionClientArgs, streaming_client::StreamingClientArgs,
+    },
 };
 use sui_indexer_alt_graphql::{
     RpcArgs as GraphQlArgs, args::SubscriptionArgs, config::RpcConfig as GraphQlConfig,
@@ -857,7 +859,7 @@ async fn start(
     force_regenesis: bool,
     epoch_duration_ms: Option<u64>,
     fullnode_rpc_port: u16,
-    mut data_ingestion_dir: Option<PathBuf>,
+    data_ingestion_dir: Option<PathBuf>,
     no_full_node: bool,
     committee_size: Option<usize>,
 ) -> Result<(), anyhow::Error> {
@@ -1021,13 +1023,6 @@ async fn start(
         sui_config_path
     };
 
-    // the indexer and consistent store require the fullnode's data ingestion directory
-    // note that this overrides the default configuration that is set when running the genesis
-    // command, which sets data_ingestion_dir to None.
-    if (with_indexer.is_some() || with_consistent_store.is_some()) && data_ingestion_dir.is_none() {
-        data_ingestion_dir = Some(mysten_common::tempdir()?.keep())
-    }
-
     if let Some(ref dir) = data_ingestion_dir {
         swarm_builder = swarm_builder.with_data_ingestion_dir(dir.clone());
     }
@@ -1046,7 +1041,37 @@ async fn start(
         swarm_builder = swarm_builder
             .with_fullnode_count(1)
             .with_fullnode_rpc_addr(fullnode_rpc_address)
-            .with_fullnode_rpc_config(rpc_config);
+            .with_fullnode_rpc_config(rpc_config.clone());
+
+        let fullnode_config_path = config_dir.join(SUI_FULLNODE_CONFIG);
+        if fullnode_config_path.exists() {
+            let mut fullnode_config: NodeConfig = PersistedConfig::read(&fullnode_config_path)
+                .map_err(|err| {
+                    err.context(format!(
+                        "Cannot open Sui fullnode config file at {:?}",
+                        fullnode_config_path
+                    ))
+                })?;
+            fullnode_config.json_rpc_address = fullnode_rpc_address;
+            fullnode_config.rpc = Some(rpc_config);
+            let localhost = sui_config::local_ip_utils::localhost_for_testing();
+            fullnode_config.metrics_address =
+                sui_config::local_ip_utils::new_local_tcp_socket_for_testing();
+            fullnode_config.admin_interface_port =
+                sui_config::local_ip_utils::get_available_port(&localhost);
+            fullnode_config.network_address =
+                sui_config::local_ip_utils::new_tcp_address_for_testing(&localhost);
+            fullnode_config.p2p_config.listen_address =
+                sui_config::local_ip_utils::new_udp_address_for_testing(&localhost)
+                    .udp_multiaddr_to_listen_address()
+                    .unwrap();
+            if let Some(ref dir) = data_ingestion_dir {
+                fullnode_config
+                    .checkpoint_executor_config
+                    .data_ingestion_dir = Some(dir.clone());
+            }
+            swarm_builder = swarm_builder.with_fullnode_config(fullnode_config);
+        }
     }
 
     let mut swarm = swarm_builder.build();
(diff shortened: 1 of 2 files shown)


hayes:
@clud-bot review

hayes-mysten:
@clud-bot thanks — both addressed in `852e2d8a73`, please re-review.

**Critical (indexer/consistent-store deadlock):** `data_ingestion_dir` now defaults to a stable `config_dir/ingestion` (created with `create_dir_all`) instead of a per-run tempdir. Ingestion checkpoint files now survive across restarts, so a persisted indexer/consistent-store at watermark `I <= H` always finds `I..H` even though the resumed fullnode only writes `H+1..` — no more `NotFound` retry-forever stall. Under `--force-regenesis`, `config_dir` is itself a tempdir, so this stays ephemeral with no special-casing.

**Minor (sticky bind ports):** on the persisted-fullnode path I now regenerate the four non-RPC bind addresses from free ports (`metrics_address`, `admin_interface_port`, `network_address`, `p2p_config.listen_address`) via `local_ip_utils`, mirroring `FullnodeConfigBuilder::build`. `json_rpc_address`, `rpc`, `db_path`, key-pairs, and `genesis` stay as persisted.

`cargo check -p sui` and `cargo clippy -p sui` are clean (zero warnings).

amnn (approved):
LGTM!

hayes-mysten:
@clud-bot re-review please — addressed in `0074b45e3f`, and I reverted the framework approach entirely.

Both findings are resolved **without any change to the shared `sui-indexer-alt-framework` crate** (`git diff upstream/main -- crates/sui-indexer-alt-framework` is empty — the FallbackIngestionClient is fully reverted). The diff is now just 2 files: `crates/sui/src/sui_commands.rs` + `crates/sui-swarm/src/memory/swarm.rs`.

- **Residual hang (first-time-indexer-on-advanced-dir):** scoped to `sui start`. On a **resuming** start, the embedded indexer + consistent-store now ingest from the fullnode's gRPC (`socket_addr_to_url(fullnode_rpc_address)`), which serves all of `0..H`, so a resumed fullnode (which only re-writes new checkpoints to the local dir) can't strand the indexer on `NotFound`. A **cold/fresh-genesis** boot keeps local-file ingestion unchanged. `resuming` is derived from whether a persisted network config pre-existed this invocation.
- **Unbounded ingestion-dir growth:** also gone — cold boot uses a tempdir (upstream behavior), resume uses gRPC, so there's no persisted local ingestion dir to accumulate. (FIX-1's `config_dir/ingestion` persistence reverted.)

`cargo check -p sui`/`-p sui-indexer-alt-framework`, `cargo clippy`, `cargo fmt --check`, and the framework ingestion tests (incl. `test_args_multiple_ingestion_sources_are_rejected`) are green.

hayes-mysten:
Opened #26887 as the **hybrid alternative**: keeps local-file ingestion as the primary everywhere (preserving the perf default) and adds a gRPC *fallback* to backfill the resume gap — at the cost of a small additive change to the shared `sui-indexer-alt-framework` crate. This PR (#26884) is the minimal, framework-untouched version (on resume, ingest from the fullnode's gRPC). Same core resume fix in both; pick whichever ingestion approach you prefer.

hayes-mysten:
Third alternative opened: **#26888** (always-gRPC — simplest, no local-file ingestion, no framework change). The full set of same-core-fix alternatives: **#26884** (this — resume→gRPC, cold boot local), **#26887** (hybrid local-file primary + gRPC fallback), **#26888** (always gRPC). Pick whichever ingestion strategy fits.

hayes-mysten:
@clud-bot please review — consolidated to the **always-gRPC** approach per the maintainer's direction (closed the hybrid #26887 and the sibling #26888). The embedded indexer + consistent-store now ingest from the fullnode's gRPC; enabling an indexer no longer requires a local data-ingestion directory. No change to the shared `sui-indexer-alt-framework` crate (its diff vs upstream is empty) — the change is just `sui_commands.rs` + `swarm.rs`. This is the same code that was green on #26888.

hayes-mysten:
@clud-bot — re-review request on the **settled** commit `22490c8a24` (CI is green). Note this is now the **always-gRPC** consolidation, which differs from the `0074b45e3f` you previously approved:
- The embedded indexer + consistent-store ingest from the fullnode's gRPC (`rpc_api_url`); enabling an indexer no longer requires/sets up a local data-ingestion directory (an explicit `--data-ingestion-dir` is still honored).
- The `resuming` conditional was removed (always-gRPC, simplest form).
- The hybrid (#26887) and always-gRPC sibling (#26888) are closed in favor of this.
- No change to the shared `sui-indexer-alt-framework` crate — the diff is `crates/sui/src/sui_commands.rs` + `crates/sui-swarm/src/memory/swarm.rs` only.

Could you approve the current commit?

crates/sui/src/sui_commands.rs
```diff
@@ -1059,6 +1082,13 @@ async fn start(
         .to_string();
     info!("Fullnode RPC URL: {fullnode_rpc_url}");
 
+    let ingestion_client_args = || -> anyhow::Result<IngestionClientArgs> {
```
amnn: nit: I would just create this once here, and then clone it, rather than create a closure to build it each time.

amnn: Also is it possible to set-up streaming as well? I.e. have downstream indexers use both streaming and ingestion (polling) to catch up with the fullnode. It should be possible by building up `ClientArgs` here, rather than just `IngestionClientArgs`, and filling out the `streaming` section with a URI that points to the fullnode as well.

hayes-mysten: Done in `20f78ebc53` — `client_args` is built once and `.clone()`d for both the indexer and the consistent-store; the closure is gone.

hayes-mysten: Added in `20f78ebc53`. `client_args` now fills both sections pointed at the fullnode: `ingestion.rpc_api_url` (polling / catch-up) and `streaming.streaming_url` (the gRPC `SubscriptionService`). I confirmed the embedded fullnode actually serves it: `sui-node/src/lib.rs:2593-2606` builds the `SubscriptionService` unconditionally and wires it into the gRPC `RpcService`, it's served on `json_rpc_address` (the same addr the ingestion URL uses), and `:1744` feeds the checkpoint sender into the `CheckpointExecutor` — so executed checkpoints are actually streamed out, not just connectable. So downstream indexers now stream live + poll to catch up.

hayes-mysten:
@amnn addressed both comments in `20f78ebc53` (streaming + ingestion via a single `ClientArgs` built once and cloned) — replies inline. @clud-bot please re-review the new commit. (CI re-running; the framework crate is still untouched.)

hayes-mysten:
Tested end-to-end on `20f78ebc53` — built `sui` and ran `sui start --with-faucet --with-graphql` through a clean stop + restart (persistent config dir):

**Cold boot** — indexer + consistent-store + GraphQL come up; the indexer tracks the chain tip within 1 checkpoint via the gRPC stream (GraphQL high-watermark 111→166→223→329 over ~36s).

**Restart (resume)** — the fullnode resumes from its DB: `executed=378 ... Loading epoch start config from DB` (not `executed=0` / genesis re-sync), with a single stable `full_node_db/<key>` across both boots. The indexer resumes from its persisted watermark (377) and logs `Within catchup threshold, starting streaming network_latest_cp=381 checkpoint_hi=377`, then catches up: GraphQL high-watermark 489→532→589, staying within 1 of the tip — no `NotFound`/stall loops. So streaming + polling both work against the embedded fullnode, and resume is clean.

(Env note: the embedded indexer needs a `postgres` server binary on PATH — on macOS I ran it with `postgresql@16`. Pre-existing requirement, not from this change.)

PR #26885, opened 2026-06-04
Add DST argument to Ristretto255 range proof verification

jonas-lj:
Adds `verify_bulletproofs_with_dst_ristretto255`, a range proof verification native that takes a domain separation tag. The existing `verify_bulletproofs_ristretto255` is disabled (aborts on call) and deprecated. The new function is only enabled on devnet.

Snapshots (genesis, compiled framework, docs) intentionally not included yet, and the fastcrypto pointer will be updated after the PR is approved.


crates/sui-framework/packages/sui-framework/sources/crypto/rangeproofs.move
```diff
@@ -28,25 +28,43 @@ const EUnsupportedVersion: u64 = 4;
...
             bits,
             &commitments.map_ref!(|c| *c.bytes()),
+            dst,
         ),
         _ => abort EUnsupportedVersion,
     }
 }
 
-native fun verify_bulletproofs_ristretto255_internal(
+#[deprecated(note = b"Use `verify_bulletproofs_with_dst_ristretto255` instead.")]
```
benr-ml, line 56: should this function ever be enabled on testnet/mainnet? what about just call abort 1 and document that it's disabled? devnet validators are deployed together anyhow
if we are to do that gradually because of contra, then we can switch to abort in a follow up pr after we migrate contra

Cargo.lock
```diff
@@ -5488,7 +5488,7 @@ dependencies = [
 [[package]]
 name = "fastcrypto"
 version = "0.1.9"
-source = "git+https://github.com/MystenLabs/fastcrypto?rev=239eed70b763f89fdb1e6c6866fe389690945c45#239eed70b763f89fdb1e6c6866fe389690945c45"
+source = "git+https://github.com/MystenLabs/fastcrypto?rev=e798bd3e33f45880c01c069dee2590812758e4b2#e798bd3e33f45880c01c069dee2590812758e4b2"
```
benr-ml: let's update the pointer in its own pr

crates/sui-framework/packages/sui-framework/sources/crypto/rangeproofs.move
jonas-lj: Sure, disabling is fine too

Cargo.lock
jonas-lj: Sure. This refers to a PR on fastcrypto that I'll keep open until this PR is approved. Then I'll merge the PR in fastcrypto and update the pointer here.

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
benr-ml, line 47: we don't need this function anymore, no?

sui-execution/latest/sui-move-natives/src/lib.rs
```diff
@@ -1306,6 +1306,11 @@ pub fn all_natives(silent: bool, protocol_config: &ProtocolConfig) -> NativeFunc
             "verify_bulletproofs_ristretto255_internal",
```
benr-ml, line 1306: do we need this one?

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -51,7 +51,33 @@ pub fn verify_bulletproofs_ristretto255(
...
+    let proof = pop_arg!(args, VectorRef);
+    verify_bulletproofs_ristretto255_impl(context, commitments, range_bits, proof, dst)
+}
 
+fn verify_bulletproofs_ristretto255_impl(
+    context: &mut NativeContext,
+    commitments: VectorRef,
+    range_bits: u8,
+    proof: VectorRef,
+    dst: Vec<u8>,
```
benr-ml: - do we need to limit the length of dst?
- do we need to charge per byte?

sui-execution/latest/sui-move-natives/src/lib.rs
jonas-lj: No, but I think we may need to keep it to allow old versions to link with the native functions. I've talked to @tzakian about our options.

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: I did some benchmarks and the cost of the length of a DST is barely measurable, so I think an upper limit is fine here. What is reasonable? We could set it to, say 32 or 64 bits, and then users should hash it if they need anything larger?

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
jonas-lj: It's needed for linking until next execution version cut.

benr-ml (approved):
The interfaces look good to me, maybe worth another review by @tzakian

crates/sui-framework/packages/sui-framework/sources/crypto/rangeproofs.move
```diff
@@ -28,23 +31,50 @@ const EUnsupportedVersion: u64 = 4;
...
+}
+
+native fun verify_bulletproofs_with_dst_ristretto255_internal(
+    proof: &vector<u8>,
+    bits: u8,
+    commitments: &vector<vector<u8>>,
+    dst: &vector<u8>,
+): bool;
+
+// TODO: this should be removed in the next execution version cut.
```
tzakian: This function can be removed now I believe. It's only the Rust-side native that needs to stay until the next execution version cut

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -45,12 +45,29 @@ fn is_supported(context: &NativeContext) -> PartialVMResult<bool> {
...
+    _context: &mut NativeContext,
+    _ty_args: Vec<Type>,
+    _args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    // Disabled: no longer called by the framework. The binding is retained only until the
+    // next execution version cut (see the Move-side TODO).
+    Err(partial_vm_error!(
+        UNKNOWN_INVARIANT_VIOLATION_ERROR,
+        "verify_bulletproofs_ristretto255 is disabled",
+    ))
```
tzakian: This needs to return `NOT_SUPPORTED` still since this will replicate the protocol-config'd behavior on non-devnet networks. Because of this, this should just always return `NOT_SUPPORTED` here so existing behavior is preserved outside of devnet:


```suggestion
    Ok(NativeResult::err(context.gas_used(), NOT_SUPPORTED))
```

sui-execution/latest/sui-move-natives/src/crypto/rangeproofs.rs
```diff
@@ -45,12 +45,29 @@ fn is_supported(context: &NativeContext) -> PartialVMResult<bool> {
...
     ty_args: Vec<Type>,
     mut args: VecDeque<Value>,
 ) -> PartialVMResult<NativeResult> {
     debug_assert!(ty_args.is_empty());
-    debug_assert!(args.len() == 3);
+    debug_assert!(args.len() == 4);
+    let dst = pop_arg!(args, VectorRef).as_bytes_ref()?.to_vec();
+    let commitments = pop_arg!(args, VectorRef);
+    let range_bits = pop_arg!(args, u8);
+    let proof = pop_arg!(args, VectorRef);
```
tzakian: Nit: Can you move these below the `is_supported` check?

sui-execution/latest/sui-move-natives/src/lib.rs
tzakian: Yea, need to keep it for linking purposes to old versions of the framework (sadly).

tzakian (approved):
Changes look good to me!

Cargo.lock
jonas-lj: Actually, since the new version of fastcrypto changes the signature of BP verification, we need to update the pointer here

PR #26886, opened 2026-06-04
Defer owned object double spending transactions

akichidis:
## Description 

Today conflicting transactions on owned object locks result to the drop of transactions from the commit output - when processing on consensus handler - and only the transaction that managed to acquire the locks eventually execute.

This PR will detect, in the same commit, the presence of conflicting transactions and will enforce on execution deferral , by 1 round, the transaction that acquire successfully the locks.

## Release plan
- Initially release without enabling the flag to gather metrics
- Once satisfied with the observability the feature will get enabled

## Test plan 

CI/PT/Antithesis


crates/sui-core/src/consensus_handler.rs
```diff
@@ -2111,6 +2120,43 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+                DeferralKey::new_for_consensus_round(commit_info.round + 1, deferred_from_round);
+
+            if transaction_deferral_within_limit(
+                &deferral_key,
+                protocol_config.max_deferral_rounds_for_congestion_control(),
+            ) {
+                debug!(
+                    "Deferring transaction {:?} due to owned object double-spend contention",
+                    transaction.tx().digest(),
+                );
```
mystenmark, line 1848: please add an assert_reachable! here and verify that it can be reached by running seed-search.py on `test_composite_workload`

akichidis: Done

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2111,6 +2129,57 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
             }
         }
 
+        // Check for owned object double-spend: if this transaction won an owned object
+        // lock while another transaction in the same commit tried to lock the same object,
+        // defer it as a penalty.
+        if let Some(conflict_info) = state
+            .contested_transaction_digests
+            .get(transaction.tx().digest())
```
mystenmark: every call to tx().digest() is a fresh blake2b hash, lets pull the digest out into a local variable once at the top of this function

mystenmark (approved):

PR #26887, opened 2026-06-04
fix(sui): `sui start` resumes the embedded fullnode — hybrid local+RPC ingestion (alt to #26884)

hayes-mysten:
**This is an alternative to #26884** — same core fix (make `sui start` resume the embedded fullnode from its persisted config + DB instead of re-syncing from genesis every boot), but a different way to handle the embedded indexer/consistent-store ingestion on resume. Opening both so maintainers can pick the preferred approach; happy to close whichever isn't wanted.

## The shared problem
On resume the fullnode only writes *new* checkpoints to the local ingestion dir (`store_checkpoint_locally` is a side-effect of *executing* a checkpoint, and the executor resumes at `H+1`), so a freshly-enabled embedded indexer can need `0..H` that aren't on disk → `IngestionClient::wait_for` retries `NotFound` forever. The data is still in the fullnode's DB / served over gRPC.

## This PR's approach — hybrid: local-file primary + gRPC fallback
- Keeps **local-file ingestion as the primary everywhere** (preserving what looks like an intentional perf choice — per-checkpoint gRPC is heavier), and persists the local ingestion dir under `config_dir/ingestion` so it survives restarts.
- Adds a `FallbackIngestionClient` to `sui-indexer-alt-framework`: tries the primary source, and on `NotFound` falls back to a secondary. `sui start` wires the embedded indexer/consistent-store with **local-file primary + the fullnode's gRPC as fallback**, so checkpoints a resumed fullnode didn't re-write locally are backfilled from the fullnode's RPC — no hang, and the fast local path is used whenever the file exists.
- Cost: a small **additive** change to the shared `sui-indexer-alt-framework` crate (`FallbackIngestionClient` + `IngestionClient::new` combining two sources when both are set). The CLI `ArgGroup` (single-source) is unchanged — the fallback is only reachable via programmatic construction (only `sui start` sets both). Existing single-source behavior is preserved; covered by 2 new unit tests.

## vs #26884
#26884 takes the simpler, fully `sui start`-scoped route: on a **resuming** start it switches the indexer ingestion entirely to the fullnode's gRPC (cold boot stays local-file), with **no change to the shared framework crate**. The trade-off: a resuming start ingests over gRPC even when the persisted local files would suffice.

| | #26884 (resume → gRPC) | this PR (hybrid) |
|---|---|---|
| local-file as primary | cold boot only | always (perf preserved) |
| gRPC used | all resume ingestion | only to backfill missing checkpoints |
| `sui-indexer-alt-framework` change | none | small additive (`FallbackIngestionClient`) |
| blast radius | `sui start` only | + the shared indexer framework |

If local-file ingestion was chosen for perf, this hybrid preserves it; if minimal blast radius is preferred, #26884 is tighter.

`cargo check -p sui`/`-p sui-indexer-alt-framework`, `cargo clippy`, `cargo fmt --check`, and the framework ingestion tests (incl. the new fallback tests + `test_args_multiple_ingestion_sources_are_rejected`) are green.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
@@ -178,6 +178,38 @@ pub enum CheckpointError {
 
 pub type CheckpointResult = Result<Checkpoint, CheckpointError>;
 
+/// Tries `primary` first and falls back to `secondary` for checkpoints the primary doesn't have.
+/// Used to back a local-file ingestion source with a fullnode RPC source so a resumed fullnode
+/// (which only re-writes new checkpoint files) can still serve already-executed checkpoints.
+struct FallbackIngestionClient {
+    primary: Arc<dyn IngestionClientTrait>,
+    secondary: Arc<dyn IngestionClientTrait>,
+}
+
+#[async_trait]
+impl IngestionClientTrait for FallbackIngestionClient {
+    async fn chain_id(&self) -> anyhow::Result<ChainIdentifier> {
+        match self.primary.chain_id().await {
+            Ok(chain_id) => Ok(chain_id),
+            Err(_) => self.secondary.chain_id().await,
+        }
+    }
+
+    async fn checkpoint(&self, checkpoint: u64) -> CheckpointResult {
+        match self.primary.checkpoint(checkpoint).await {
+            Err(CheckpointError::NotFound) => self.secondary.checkpoint(checkpoint).await,
+            result => result,
+        }
+    }
+
+    async fn latest_checkpoint_number(&self) -> anyhow::Result<u64> {
+        match self.primary.latest_checkpoint_number().await {
+            Ok(number) => Ok(number),
+            Err(_) => self.secondary.latest_checkpoint_number().await,
+        }
+    }
+}
+
 #[derive(Clone)]
 pub struct IngestionClient {
     client: Arc<dyn IngestionClientTrait>,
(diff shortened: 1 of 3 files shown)


hayes-mysten:
@clud-bot please review — this is the **hybrid alternative** to #26884 (which you already reviewed). Same core fix (resume the embedded fullnode from persisted config + db). The difference: instead of switching resume-path ingestion to gRPC, this keeps **local-file ingestion as the primary** and adds a `FallbackIngestionClient` that falls back to the fullnode's gRPC on `NotFound` — so the residual you flagged (first-time-indexer-on-advanced-dir) is resolved by backfilling `0..H` from the fullnode's RPC while the fast local path is used whenever the file exists. Trade-off vs #26884: this touches the shared `sui-indexer-alt-framework` crate (additive: the fallback client + `new()` combining two sources; CLI ArgGroup unchanged, 2 new unit tests).

hayes-mysten:
Third alternative opened: **#26888** (always-gRPC — simplest, no local-file ingestion, no framework change). The full set: **#26884** (resume→gRPC, cold boot local), **#26887** (this — hybrid local-file primary + gRPC fallback), **#26888** (always gRPC).

hayes-mysten:
Closing in favor of #26884. Per review the decision is the **always-gRPC** ingestion approach (simplest, no change to the shared `sui-indexer-alt-framework` crate); the resume fix + always-gRPC are consolidated into #26884. The hybrid (local-file primary + gRPC fallback) remains in this branch's history if we ever want the local-file fast path back. Thanks for the review!

PR #26888, opened 2026-06-04
fix(sui): `sui start` resumes the embedded fullnode — always-gRPC ingestion (alt to #26884, #26887)

hayes-mysten:
## Summary

This is the **simplest** of three alternatives that all share the same core fix: make `sui start` **resume** the embedded fullnode from its persisted config + db instead of rebuilding it from scratch (which minted a fresh keypair/db each boot and re-synced the whole chain from genesis). See #26884 for the full diagnosis of the resume bug and the doubled genesis db-path.

The three PRs differ **only** in how the embedded indexer + consistent-store get their checkpoints:

- **#26884** — resume → fullnode gRPC, cold boot → local checkpoint files (a `resuming` conditional + an auto-defaulted local data-ingestion dir).
- **#26887** — hybrid: local-file primary with gRPC fallback (touches the shared `sui-indexer-alt-framework` crate).
- **this PR** — **always gRPC**: the embedded indexer / consistent-store ingest from the fullnode's gRPC (which serves *all* checkpoints) on **both** cold boot and resume. No `resuming` conditional, no local-file ingestion lifecycle, and **no change to the shared framework crate at all**.

Because the fullnode's gRPC serves every checkpoint, a resumed fullnode that didn't re-write its already-executed checkpoints to a local dir is simply a non-issue — there is nothing to special-case. An explicitly-provided `--data-ingestion-dir` is still honored so an operator can feed an external consumer; we just no longer auto-create one for the embedded indexer.

## What's in this PR

Same core resume fix as #26884:
- Resume the embedded fullnode from the persisted `fullnode.yaml` + db (via `with_fullnode_config`).
- Fix the doubled genesis fullnode db-path (`full_node_db/full_node_db/<key>` → single, absolute level).
- Regenerate the non-RPC bind ports (metrics/admin/network/p2p) on the resume path, since the genesis-time ports may now be occupied.

Plus the always-gRPC simplification:
- Ingestion `IngestionClientArgs` for both the indexer and consistent-store unconditionally use `rpc_api_url: Some(socket_addr_to_url(fullnode_rpc_address))`.
- Removed the `resuming` variable and its derivation entirely (now dead).
- Removed the auto-default local `data_ingestion_dir` block.

All scoped to `sui start`; `crates/sui-indexer-alt-framework` is untouched.

## Trade-offs vs the alternatives

| | local-file used | gRPC used | shared-framework change | blast radius | complexity |
|---|---|---|---|---|---|
| #26884 (resume→gRPC, cold→local) | cold boot | resume | none | `sui start` only | medium (`resuming` conditional + auto-default dir) |
| #26887 (hybrid local+gRPC fallback) | primary | fallback | **yes** (framework) | `sui start` + shared crate | highest (dual-source fallback logic) |
| **this PR (always gRPC)** | only if `--data-ingestion-dir` given | always | **none** | `sui start` only | **lowest** (no conditional, no dir lifecycle) |

## Downside

Cold-boot initial sync for the embedded indexer/consistent-store now also goes over gRPC instead of reading local checkpoint files. For a local `sui start` cluster this is negligible, but it's the one behavior that #26884 retains (local-file on cold boot) and this PR trades away for simplicity.

## Alternatives

- #26884 — resume → gRPC, cold boot → local-file.
- #26887 — hybrid local-file primary + gRPC fallback.

### DIFF
--- a/crates/sui-swarm/src/memory/swarm.rs
+++ b/crates/sui-swarm/src/memory/swarm.rs
@@ -49,6 +49,7 @@ pub struct SwarmBuilder<R = OsRng> {
     fullnode_rpc_port: Option<u16>,
     fullnode_rpc_addr: Option<SocketAddr>,
     fullnode_rpc_config: Option<sui_config::RpcConfig>,
+    fullnode_config: Option<NodeConfig>,
     supported_protocol_versions_config: ProtocolVersionsConfig,
     // Default to supported_protocol_versions_config, but can be overridden.
     fullnode_supported_protocol_versions_config: Option<ProtocolVersionsConfig>,
@@ -85,6 +86,7 @@ impl SwarmBuilder {
             fullnode_rpc_port: None,
             fullnode_rpc_addr: None,
             fullnode_rpc_config: None,
+            fullnode_config: None,
             supported_protocol_versions_config: ProtocolVersionsConfig::Default,
             fullnode_supported_protocol_versions_config: None,
             db_checkpoint_config: DBCheckpointConfig::default(),
@@ -121,6 +123,7 @@ impl<R> SwarmBuilder<R> {
             fullnode_rpc_port: self.fullnode_rpc_port,
             fullnode_rpc_addr: self.fullnode_rpc_addr,
             fullnode_rpc_config: self.fullnode_rpc_config.clone(),
+            fullnode_config: self.fullnode_config,
             supported_protocol_versions_config: self.supported_protocol_versions_config,
             fullnode_supported_protocol_versions_config: self
                 .fullnode_supported_protocol_versions_config,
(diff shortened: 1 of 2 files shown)


hayes-mysten:
@clud-bot please review — this is the **simplest** of three alternatives to the same core resume fix. The embedded indexer + consistent-store **always** ingest from the fullnode's gRPC (cold boot and resume alike), so there's no local-file ingestion lifecycle and **no change to the shared `sui-indexer-alt-framework` crate** (`git diff upstream/main` on that crate is empty). The residual you flagged on #26884 (first-time-indexer-on-advanced-dir) is moot here since the fullnode's gRPC always serves `0..H`. Trade-off vs the others: cold-boot initial sync also goes over gRPC instead of local files. Siblings: #26884 (resume→gRPC), #26887 (hybrid local+gRPC fallback).

hayes-mysten:
Closing — consolidated into #26884, which now carries this exact always-gRPC code as the canonical PR.

PR #26891, opened 2026-06-04
[forwarding-address] Add ForwardingAddressRegistry system object (step 1)

lxfind:
## Description

First step of the **forwarding-address** feature (Sui-native deposit aliases that forward funds to a registered master at resolution time — the design analog of Tempo TIP-1022, deliberately named "forwarding address" rather than "virtual address" since these are pure forwarding aliases that can't sign or own state).

This PR adds **only** the minimal foundation: an empty registry object and the machinery to create it. No registration or resolution APIs yet.

- **Move:** new `0x2::forwarding_address` module with `ForwardingAddressRegistry { id: UID }` and a system-only `create()`. Static object ID `0xfa` added in `object.move` and `sui-types`.
- **Protocol flag** `create_forwarding_address_registry`, enabled in **v132** for non-mainnet/testnet chains only. v130 is cut in `sui-v1.76.0` (with its framework bytecode snapshot frozen), and v131 was claimed by other features on `main`, so the flag lives in v132's existing devnet-only block.
- **Epoch-change / genesis creation:** new `EndOfEpochTransactionKind::ForwardingAddressRegistryCreate`, wired through the authority's end-of-epoch tx assembly, the epoch-start configuration (registered in `SYSTEM_SHARED_OBJECT_IDS`, so no new configuration version is needed), the existence check, and the genesis builder. The real executor impl lives only in `latest` (execution_version 4); v0–v3 are frozen pre-exec-4 versions that reject the variant.
- **Allowlisting the hardcoded UID for sharing:** the move-compiler ID-leak "fresh UID" list + symbol pool, and the runtime `object_runtime::is_framework_obj` set, so `share_object` accepts `0xfa` at genesis / epoch change.
- Regenerated framework artifacts, protocol-config / BCS-format / GraphQL / OpenRPC snapshots.

The `sui-types` conversions to the externally-pinned `sui-sdk-types` / `sui-rpc` types are real arms: MystenLabs/sui-rust-sdk#288 added the `ForwardingAddressRegistryCreate` variant upstream, and this PR bumps the pinned `rev` in `Cargo.toml` to it.

(Also includes a small, separate commit tidying `.typos.toml`: relax the Base58 ignore threshold 44→43 and exclude compiled Move bytecode.)

## Test plan

- Registry creation at genesis is exercised by every test network started with default (devnet) protocol config, and is pinned in the populated-genesis snapshot.
- The framework build test verifies the new `forwarding_address` module compiles and that the ID-leak allowlist accepts sharing the hardcoded `0xfa` UID.
- Tests that pin execution version 3 (`move_package_upgrade_tests`, `address_balance_tests`, `coin_registry_tests`) disable the flag, covering the constraint that the registry object cannot be created pre-exec-v4.
- Protocol-config, BCS format, GraphQL schema, and OpenRPC snapshots pin the new flag, transaction kind, and API surface.
- New `test_create_forwarding_address_registry_object_at_upgrade` e2e simtest covers the epoch-change creation path (the one existing chains like devnet will actually take, since only fresh networks get the registry at genesis): it starts a cluster at protocol 130 with the frozen v130 framework snapshot (which predates the `forwarding_address` module), upgrades to 132, and verifies every node creates the shared registry object at the end of the epoch.

---

## Release notes

- [x] Protocol: Adds a `ForwardingAddressRegistry` system object (`0x2::forwarding_address`, object ID `0xfa`), created during epoch change when the `create_forwarding_address_registry` flag is enabled (devnet only for now). No user-facing API yet.
- [x] Nodes (Validators and Full nodes): New end-of-epoch system transaction `ForwardingAddressRegistryCreate`.
- [x] gRPC: New `FORWARDING_ADDRESS_REGISTRY_CREATE` kind in the `EndOfEpochTransactionKind` proto enum.
- [ ] JSON-RPC:
- [x] GraphQL: New `ForwardingAddressRegistryCreateTransaction` member of the `EndOfEpochTransactionKind` union.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


mystenmark (approved):

admin-aftermath:
@lxfind Can we read about forwarding addresses somewhere?

PR #26893, opened 2026-06-04
Introduce gas_v3.rs for v15+ SuiGasStatus

dariorussi:
## Summary
- New `crates/sui-types/src/gas_model/gas_v3.rs` as a clean alternative to `gas_v2.rs`, dispatched to whenever `gas_model_version >= 15` via the version-aware `SuiGasStatus` enum (`enum_dispatch(SuiGasStatusAPI)`). Old executors keep `gas_v2` untouched, so old-protocol replay is unaffected.
- `per_object_storage()` lifted onto `SuiGasStatusAPI` so `sui-replay`'s display can drive either variant via the trait object. `gas_v2.rs` change is a single additive trait-method body; semantics unchanged.

## Cleanups introduced in V3 only
V3 drops the dead branches V2 still carries for backwards compatibility:
- `ComputationBucket` struct + helpers — V3 hardcodes `half_digits_rounding`.
- `GasRoundingMode` enum, field, and per-config selection.
- `max_gas_price_rgp_factor_for_aborted_transactions: Option<u64>` → `u64` (always `Some(_)` at v127+).
- `SuiCostTable::new`'s non-multiplier base-cost arm (`txn_base_cost_as_multiplier` is true at every v127+ version).

`PerObjectStorage` is shared (`gas_v3` re-exports `gas_v2`'s definition; pure data type, zero behavior).

## Why two SuiGasStatus types
The cross-executor `sui_types::gas::SuiGasStatus` is reached from every execution layer (v0..v3 *and* latest) for old-protocol replay. We can't change `gas_v2.rs` semantics without regressing those layers. `gas_v3.rs` lets us land all the cleanups for v15+ semantics behind the existing version-aware dispatch — no per-executor duplication needed.

## Test plan
- New regression test `test_oog_full_budget_charge_exact_when_budget_indivisible_by_price`: pins the `force_computation_cost_to_budget` precision fix using a budget deliberately not a multiple of the gas price — the pre-existing OOG test used a clean multiple and never tripped the rounding loss.
- Version-dispatch coverage falls out of the existing suites: old-protocol-pinned transactional tests exercise gas_v2 + the legacy path unchanged; v127+ tests exercise gas_v3 + the step pipeline.
- Replaying mainnet transactions from epoch 1100 onward shows no change in effects.

### DIFF
--- a/sui-execution/latest/sui-adapter/src/execution_engine.rs
+++ b/sui-execution/latest/sui-adapter/src/execution_engine.rs
@@ -14,8 +14,8 @@ mod checked {
     use move_trace_format::format::MoveTraceBuilder;
     use move_vm_runtime::runtime::MoveRuntime;
     use mysten_common::{assert_reachable, debug_fatal, in_test_configuration};
-    use std::collections::BTreeMap;
-    use std::{cell::RefCell, collections::HashSet, rc::Rc, sync::Arc};
+    use std::collections::{BTreeMap, BTreeSet};
+    use std::{cell::RefCell, rc::Rc, sync::Arc};
     use sui_types::accumulator_root::{ACCUMULATOR_ROOT_CREATE_FUNC, ACCUMULATOR_ROOT_MODULE};
     use sui_types::balance::{
         BALANCE_CREATE_REWARDS_FUNCTION_NAME, BALANCE_DESTROY_REBATES_FUNCTION_NAME,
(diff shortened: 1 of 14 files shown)


sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -88,97 +89,107 @@ mod checked {
...
         execution_params: &ExecutionOrEarlyError,
     ) -> bool {
         execution_params.early_errors().is_some_and(|errors| {
-            matches!(
-                errors.head,
-                ExecutionErrorKind::InsufficientFundsForWithdraw
-            )
+            errors
+                .iter()
+                .any(|e| matches!(e, ExecutionErrorKind::InsufficientFundsForWithdraw))
```
lxfind, line 116: After some thought, I believe this is incorrect change, i.e. it needs to be gated otherwise can fork.
When adding errors to the list of early return errors, the iffw is always added last, which means it being the head is very different semantics than in any location.

PR #26894, opened 2026-06-04
[graphql] Thread `alpha_ledger_grpc_reader` to optionally support bitmap stream apis with backward compatibility 1/n

wlmyng:
## Description 
New ledger grpc reader for alpha bitmap streaming apis to be used by graphql. Supports draining a stream from a single grpc request. 

Setup is through KvArgs controlled by a new `Option<bool>` flag, and reuses the same config for ledger service. This alpha ledger service gets added to graphql context if instantiated.

## Test plan 

Tests in new `alpha_ledger_grpc_reader.rs` file

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
+++ b/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
@@ -8,7 +8,7 @@ use futures::Stream;
 use futures::StreamExt;
 use prometheus::Registry;
 use sui_rpc::proto::sui::rpc::v2alpha as grpc_alpha;
-use sui_rpc::proto::sui::rpc::v2alpha::ledger_service_client::LedgerServiceClient as V2alphaLedgerServiceClient;
+use sui_rpc::proto::sui::rpc::v2alpha::ledger_service_client::LedgerServiceClient;
 use tonic::transport::Channel;
 use tonic::transport::ClientTlsConfig;
 use tonic::transport::Uri;
@@ -24,7 +24,7 @@ const DEFAULT_MAX_DECODING_MESSAGE_SIZE: usize = 32 * 1024 * 1024;
 /// A reader backed by the gRPC LedgerService's v2alpha experimental query APIs.
 #[derive(Clone)]
 pub struct AlphaLedgerGrpcReader {
-    client: V2alphaLedgerServiceClient<GrpcMetricsService<Channel>>,
+    client: LedgerServiceClient<GrpcMetricsService<Channel>>,
     timeout: Option<Duration>,
 }
 
@@ -77,7 +77,7 @@ impl AlphaLedgerGrpcReader {
         let max_decoding_message_size = args
             .ledger_grpc_max_decoding_message_size
             .unwrap_or(DEFAULT_MAX_DECODING_MESSAGE_SIZE);
-        let client = V2alphaLedgerServiceClient::new(layered.clone())
+        let client = LedgerServiceClient::new(layered.clone())
             .max_decoding_message_size(max_decoding_message_size);
 
         Ok(Self { client, timeout })
--- a/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
+++ b/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
@@ -144,7 +144,7 @@ impl<I> StreamPage<I> {
     /// Fold one frame into the page. `first_cursor` is set once to the first cursor-bearing frame.
     /// `last_cursor` continuously tracks the latest cursor-bearing frame.
     ///
-    /// Returns `true` when the frame is the terminal `QueryEnd`.
+    /// Returns `true` when the frame is `QueryEnd`.
     fn apply(&mut self, frame: FrameKind<I>) -> bool {
         match frame {
             FrameKind::Item { item, cursor } => {
@@ -350,13 +350,13 @@ mod tests {
     }
 
     #[test]
-    fn apply_signals_terminal_only_on_end_frame() {
+    fn apply_signals_stop_only_on_end_frame() {
         // The bool returned by `apply` is the drain loop's stop signal: `true` means "stop
-        // draining," `false` means "keep going." Only the terminal `End` frame should signal stop.
+        // draining," `false` means "keep going." Only the `End` frame should signal stop.
         let mut page: StreamPage<grpc_alpha::TransactionItem> = StreamPage::default();
         assert!(!page.apply(item_response(b"c1").into()));
         assert!(!page.apply(watermark_response(b"w1").into()));
-        // Outer message with no oneof set → `FrameKind::Unknown` → also non-terminal.
+        // Outer message with no oneof set → `FrameKind::Unknown` → continue.
         assert!(!page.apply(grpc_alpha::ListTransactionsResponse::default().into()));
         assert!(page.apply(end_response(grpc_alpha::QueryEndReason::LedgerTip).into()));
     }
@@ -387,7 +387,7 @@ mod tests {
             assert!(page.has_more(), "expected has_more for {reason:?}");
         }
 
-        // `end_reason == None` covers both the deadline cut-short case (no terminal frame
+        // `end_reason == None` covers both the deadline cut-short case (no end frame
         // received) and any unrecognized / future-added variant — defaulting to "may have more"
         // avoids silent truncation.
         let page: StreamPage<grpc_alpha::TransactionItem> = StreamPage::default();
--- a/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
+++ b/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
@@ -71,9 +71,8 @@ impl AlphaLedgerGrpcReader {
             GrpcMetricsLayer::new(prefix.unwrap_or("ledger_grpc"), registry).layer(channel);
 
         let timeout = args.statement_timeout();
-        let max_decoding_message_size = args.max_decoding_message_size();
         let client = LedgerServiceClient::new(layered.clone())
-            .max_decoding_message_size(max_decoding_message_size);
+            .max_decoding_message_size(args.ledger_grpc_max_decoding_message_size);
 
         Ok(Self { client, timeout })
     }
--- a/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
+++ b/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
@@ -10,6 +10,7 @@ use bytes::Bytes;
 use futures::Stream;
 use futures::StreamExt;
 use prometheus::Registry;
+use sui_rpc::proto::sui::rpc::v2::ExecutedTransaction;
 use sui_rpc::proto::sui::rpc::v2alpha as proto;
 use sui_rpc::proto::sui::rpc::v2alpha::ledger_service_client::LedgerServiceClient;
 use tonic::transport::Channel;
@@ -29,21 +30,26 @@ pub struct AlphaLedgerGrpcReader {
     timeout: Option<Duration>,
 }
 
+/// A single item from a list stream and the resume cursor the server emitted alongside it.
+#[derive(Debug, Clone)]
+pub struct PageItem<T> {
+    pub payload: T,
+    pub cursor: Bytes,
+}
+
 /// A page drained from a single gRPC list stream.
 #[derive(Debug, Clone)]
-pub struct StreamPage<I> {
+pub struct StreamPage<T> {
     /// Items that matched the filters, in stream order.
-    pub items: Vec<I>,
-    /// First cursor value observed across item and standalone watermark frames.
-    pub first_cursor: Option<Bytes>,
-    /// Latest cursor value observed, advances as new cursor-bearing frames arrive.
-    pub last_cursor: Option<Bytes>,
+    pub items: Vec<PageItem<T>>,
+    first_wm_cursor: Option<Bytes>,
+    last_wm_cursor: Option<Bytes>,
     pub end_reason: Option<proto::QueryEndReason>,
 }
 
 #[derive(Debug)]
-enum FrameKind<I> {
-    Item { item: I, cursor: Option<Bytes> },
+enum FrameKind<T> {
+    Item { payload: T, cursor: Bytes },
     Watermark { cursor: Option<Bytes> },
     End { reason: i32 },
     Unknown,
@@ -80,7 +86,7 @@ impl AlphaLedgerGrpcReader {
     pub async fn list_transactions(
         &self,
         request: proto::ListTransactionsRequest,
-    ) -> anyhow::Result<StreamPage<proto::TransactionItem>> {
+    ) -> anyhow::Result<StreamPage<ExecutedTransaction>> {
         let mut client = self.client.clone();
         let stream = client
             .list_transactions(self.request(request))
@@ -101,7 +107,7 @@ impl AlphaLedgerGrpcReader {
     }
 }
 
-impl<I> StreamPage<I> {
+impl<T> StreamPage<T> {
     /// Whether further data may exist in the direction of pagination.
     ///
     /// `false` iff one of:
@@ -115,51 +121,52 @@ impl<I> StreamPage<I> {
             None => true,
             Some(R::Unspecified | R::ItemLimit | R::ScanLimit) => true,
             Some(R::LedgerTip | R::CheckpointBound) => false,
-            Some(R::CursorBound) => self.last_cursor.is_some(),
+            Some(R::CursorBound) => self.last_cursor().is_some(),
             // `QueryEndReason` is non exhaustive — conservatively `true` if a
             // future variant slips past `apply()`'s `unwrap_or(Unspecified)`.
             Some(_) => true,
         }
     }
 
-    /// The latest cursor observed in the direction of pagination.
-    pub fn next_cursor(&self) -> Option<&Bytes> {
-        self.last_cursor.as_ref()
+    /// The page's starting cursor: the standalone-watermark cursor if one preceded any items,
+    /// otherwise the first item's own cursor.
+    pub fn first_cursor(&self) -> Option<&Bytes> {
+        self.first_wm_cursor
+            .as_ref()
+            .or_else(|| self.items.first().map(|item| &item.cursor))
     }
 
-    /// Fold one frame into the page. `first_cursor` is set once to the first cursor-bearing frame.
-    /// `last_cursor` continuously tracks the latest cursor-bearing frame.
+    /// The page's resume cursor: a standalone-watermark cursor emitted after the last item if one
+    /// exists, otherwise the last item's own cursor.
+    pub fn last_cursor(&self) -> Option<&Bytes> {
+        self.last_wm_cursor
+            .as_ref()
+            .or_else(|| self.items.last().map(|item| &item.cursor))
+    }
+
+    /// Fold one frame into the page.
     ///
     /// Returns `true` when the frame is `QueryEnd`.
-    fn apply(&mut self, frame: FrameKind<I>) -> bool {
+    fn apply(&mut self, frame: FrameKind<T>) -> bool {
         match frame {
-            FrameKind::Item { item, cursor } => {
-                if cursor.is_some() {
-                    if self.first_cursor.is_none() {
-                        self.first_cursor = cursor.clone();
-                    }
-                    self.last_cursor = cursor;
-                }
-                self.items.push(item);
+            FrameKind::Item { payload, cursor } => {
+                self.last_wm_cursor = None;
+                self.items.push(PageItem { payload, cursor });
             }
             FrameKind::Watermark { cursor } => {
-                if cursor.is_some() {
-                    if self.first_cursor.is_none() {
-                        self.first_cursor = cursor.clone();
-                    }
-                    self.last_cursor = cursor;
+                self.last_wm_cursor = cursor.clone();
+                if self.items.is_empty() && self.first_wm_cursor.is_none() {
+                    self.first_wm_cursor = cursor.clone();
                 }
             }
             FrameKind::End { reason } => {
                 // Fold an unknown reason into `Unspecified` so `None` remains unambiguous shorthand
                 // for "no End frame received" (i.e. the deadline cut the stream short).
-                self.end_reason = match proto::QueryEndReason::try_from(reason) {
-                    Ok(decoded) => Some(decoded),
-                    Err(_) => {
-                        warn!(reason, "unknown QueryEndReason",);
-                        Some(proto::QueryEndReason::Unspecified)
-                    }
-                };
+                self.end_reason =
+                    Some(proto::QueryEndReason::try_from(reason).unwrap_or_else(|_| {
+                        warn!(reason, "unknown QueryEndReason");
+                        proto::QueryEndReason::Unspecified
+                    }));
                 return true;
             }
             FrameKind::Unknown => {
@@ -170,44 +177,52 @@ impl<I> StreamPage<I> {
     }
 }
 
-impl<I> Default for StreamPage<I> {
+impl<T> Default for StreamPage<T> {
     fn default() -> Self {
         Self {
             items: Vec::new(),
-            first_cursor: None,
-            last_cursor: None,
+            first_wm_cursor: None,
+            last_wm_cursor: None,
             end_reason: None,
         }
     }
 }
 
-impl From<proto::ListTransactionsResponse> for FrameKind<proto::TransactionItem> {
-    fn from(response: proto::ListTransactionsResponse) -> Self {
+impl TryFrom<proto::ListTransactionsResponse> for FrameKind<ExecutedTransaction> {
+    type Error = anyhow::Error;
+
+    fn try_from(response: proto::ListTransactionsResponse) -> anyhow::Result<Self> {
         use proto::list_transactions_response::Response;
 
         let Some(response) = response.response else {
-            return FrameKind::Unknown;
+            return Ok(FrameKind::Unknown);
         };
-        match response {
+        Ok(match response {
             Response::Item(item) => {
-                let cursor = item.watermark.as_ref().and_then(|w| w.cursor.clone());
-                FrameKind::Item { item, cursor }
+                let cursor = item
+                    .watermark
+                    .and_then(|w| w.cursor)
+                    .context("Item frame missing watermark.cursor")?;
+                let payload = item
+                    .transaction
+                    .context("Item frame missing transaction payload")?;
+                FrameKind::Item { payload, cursor }
             }
             Response::Watermark(watermark) => FrameKind::Watermark {
                 cursor: watermark.cursor,
             },
             Response::End(end) => FrameKind::End { reason: end.reason },
             _ => FrameKind::Unknown,
-        }
+        })
     }
 }
 
-async fn drain_list_stream<R, I, S>(
+async fn drain_list_stream<R, T, S>(
     rpc_name: &'static str,
     stream: S,
-) -> anyhow::Result<StreamPage<I>>
+) -> anyhow::Result<StreamPage<T>>
 where
-    R: Into<FrameKind<I>>,
+    R: TryInto<FrameKind<T>, Error = anyhow::Error>,
     S: Stream<Item = Result<R, tonic::Status>>,
 {
     futures::pin_mut!(stream);
@@ -215,8 +230,11 @@ where
     while let Some(result) = stream.next().await {
         match result {
             Ok(response) => {
+                let frame = response
+                    .try_into()
+                    .with_context(|| format!("{rpc_name}: malformed frame"))?;
                 // Process and break on receiving `QueryEnd`.
-                if page.apply(response.into()) {
+                if page.apply(frame) {
                     break;
                 }
             }
@@ -242,10 +260,11 @@ where
         }
     }
 
-    // Exited via `break` or via `None`. If `has_more() promises further data, the cursor to resume
-    // from must be present.`
+    // Exited via `break` or via `None`. If `has_more()` promises further data, the cursor to resume
+    // from must be present — either a standalone watermark advanced past the last item, or the last
+    // item's own cursor.
     ensure!(
-        !page.has_more() || page.last_cursor.is_some(),
+        !page.has_more() || page.last_cursor().is_some(),
         "{rpc_name}: server reported more results but did not provide resume cursor — cannot continue",
     );
 
@@ -256,11 +275,13 @@ where
 mod tests {
     use super::*;
 
+    /// Well-formed `Item` frame with both `transaction` payload and a cursor-bearing watermark.
     fn item_response(cursor: &[u8]) -> proto::ListTransactionsResponse {
         let mut watermark = proto::Watermark::default();
         watermark.cursor = Some(Bytes::copy_from_slice(cursor));
         let mut item = proto::TransactionItem::default();
         item.watermark = Some(watermark);
+        item.transaction = Some(ExecutedTransaction::default());
         let mut response = proto::ListTransactionsResponse::default();
         response.response = Some(proto::list_transactions_response::Response::Item(item));
         response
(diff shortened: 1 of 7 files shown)


crates/sui-indexer-alt-reader/src/kv_loader.rs
```diff
@@ -82,6 +83,11 @@ pub struct KvArgs {
     #[arg(long, group = "kv_source")]
     pub ledger_grpc_url: Option<Uri>,
 
+    /// Whether the configured ledger gRPC service also has alpha experimental query APIs enabled
+    /// (e.g. bitmap-backed transaction pagination). When unset, treated as `false`.
+    #[arg(long)]
+    pub enable_experimental_query_apis: Option<bool>,
```
amnn: (nit: For the sake of the flag -- flags are implicitly "enablers")

```suggestion
    pub experimental_query_apis: Option<bool>,
```

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,496 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::time::Duration;
+
+use bytes::Bytes;
+use futures::Stream;
+use futures::StreamExt;
+use prometheus::Registry;
+use sui_rpc::proto::sui::rpc::v2alpha as grpc_alpha;
+use sui_rpc::proto::sui::rpc::v2alpha::ledger_service_client::LedgerServiceClient as V2alphaLedgerServiceClient;
```
amnn: I think this alias is no longer necessary because it's in its own file? Can just call this `LedgerServiceClient` now.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,496 @@
...
+    /// the `end_cursor` if there are more results to yield after the current page.
+    pub async fn list_transactions(
+        &self,
+        request: grpc_alpha::ListTransactionsRequest,
+    ) -> anyhow::Result<StreamPage<grpc_alpha::TransactionItem>> {
+        let mut client = self.client.clone();
+        let stream = client
+            .list_transactions(self.request(request))
+            .await
+            .map_err(|s| anyhow::anyhow!("ListTransactions stream open failed: {}", s.message()))?
```
amnn: `anyhow::Context`?

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,496 @@
...
+
+    /// The cursor to continue paginating, or `None` if the requested range has been exhausted and
+    /// no further pagination is possible.
+    ///
+    /// Invariant: `has_more()` ⇒ `end_cursor.is_some()`. Enforced at the page boundary by
+    /// `drain_list_stream` (returns `data_loss` on violation) and preserved by any caller that
+    /// re-synthesizes a `StreamPage` from a previously-validated one's `end_reason` +
+    /// `end_cursor` fields together.
+    pub fn next_cursor(&self) -> Option<&Bytes> {
+        self.has_more().then(|| {
```
amnn: Should we force this value to `None` if `has_more` is `false`? A query pattern that is trying to keep up with the tip of the network would then not be able to see the cursor that points to the last known tip, to resume from later.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,496 @@
...
+    /// True while the server has not exhausted the requested range.
+    pub fn has_more(&self) -> bool {
+        !matches!(
+            self.end_reason,
+            Some(
+                grpc_alpha::QueryEndReason::CheckpointBound
+                    | grpc_alpha::QueryEndReason::CursorBound
+                    | grpc_alpha::QueryEndReason::LedgerTip
+            )
+        )
```
amnn: I found myself a bit confused by the double negative here, could we use an explicit `match` statement? Also if more reasons are added in future we would get a compile time error for exhaustiveness:

```rust
pub fn has_more(&self) -> bool {
    use grpc_alpha::QueryEndReason as R;
    match self.end_reason {
        None => true,
        Some(R::Unspecified) => true,
        Some(R::ItemLimit | R::ScanLimit) => true,
        Some(R::CheckpointBound | R::CursorBound | R::LedgerTip) => false,
    }
}
```

amnn: Also, should hitting a `CursorBound` be an indication that you don't have more? That's not how things work today in GraphQL.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,496 @@
...
+            // Both are timeout-shaped — preserve partial work if any progress was made;
+            // propagate as error only if zero progress, so the caller can reshape.
+            Some(Err(status))
+                if matches!(
+                    status.code(),
+                    tonic::Code::DeadlineExceeded | tonic::Code::Cancelled
+                ) =>
+            {
+                if page.items.is_empty() && page.end_cursor.is_none() {
+                    return Err(anyhow::anyhow!(
```
amnn: Use `ensure!`

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,496 @@
...
+                        "{rpc_name} stream {:?} with no progress: {}",
+                        status.code(),
+                        status.message()
+                    ));
+                }
+                break;
+            }
+            Some(Err(status)) => {
+                return Err(anyhow::anyhow!(
```
amnn: Use `bail!`

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,496 @@
...
+    loop {
+        match stream.next().await {
+            Some(Ok(response)) => {
+                // Process and break on receiving `QueryEnd`.
+                if page.apply(response.into()) {
+                    break;
+                }
+            }
+            // We expect the server to yield an `End` frame before reaching this branch.
+            None => break,
```
amnn: Should we be also checking that we at least one item or an end cursor here?

amnn: In fact should that be a check that happens outside of the loop, at the end, as a final consistency check? It seems like it's possible for `page.apply(...)` to return `true` and for that property not to hold.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,496 @@
...
+                    "{rpc_name} stream error: {}",
+                    status.message()
+                ));
+            }
+        }
+    }
+
+    // Pagination is considered unresumable if there is more server-side work, but no valid cursor
+    // was yielded (either from a standalone `Watermark` or the last `Item`'s watermark.)
+    if page.has_more() && page.end_cursor.is_none() {
```
amnn: `ensure!`

amnn (commented):
Thanks @wlmyng -- broad structure looks good, but I had some questions about edge cases related to stream draining.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: Ah good catch, basically only the range bounds should indicate no more items

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: Yeah that makes sense. I was preoccupied with the case where `has_more` but `!end_cursor` and didn't account for `!has_more`, in which we should still return a cursor if we do have one

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: There is a check at L275, but if we get an `End` framew and no item or watermark, it will slip through it ... Yeah i think it makes sense to have this check, as we expect the server to yield at least one item or watermark that represents some progress the server has done in scanning through

wlmyng: Revisiting, if the server did some work, we expect at least one cursor-bearing frame. However, it's possible for the server to not have done any work, such as hitting `End::CursorBound` in a way that exceeds tip (or `LedgerTip`). In these cases though `has_more()` would be false

https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt-e2e-tests/tests/kv_rpc_tests.rs#L3650

wlmyng: Had to sit with it some more, iterating through the end reasons, `items.is_empty() && end_cursor.is_none()` is potentially and only reachable in the case `End::CursorBound`, where ledger service short-circuits because the cursor has reached range end

At `LedgerTip` or `CheckpointBound`, we do expect some terminal watermark

and otherwise we expect the server to report watermark progress or items if work has been done

I'll do some tidying to make this clearer in the code, but the check `if page.has_more() && page.end_cursor.is_none()` should be correct...

Well, a related point is whether we consider `CursorBound` terminal or not, I suppose. On one hand the request has been fulfilled, on the other, more matching data may exist beyond the cursor bound set.

wlmyng: So I think from Graphql/ Relay's perspective, `CursorBound` or `CheckpointBound` does not necessarily indicate `has_more() == false`, there could be more data past the current page returned.

But those query end reasons are terminal when request-scoped

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: It's true that `CursorBound` is not treated the same way here as in graphql, but I think this should stay request-scoped, and graphql layer can then determine whether to set `hasNext/PrevPage` true or false

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: if we keep the `has_more` logic request-scoped, `End::CursorBound` is a terminal reason, and it is thus possible for `page.apply()` to return `true` without any item or watermark emitted. Adding a check for at least one item or end cursor would reject this valid scenario

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: Unlike in Postgres, I'm not sure if we can efficiently determine that elements exist beyond `before`, in which case the specs allow us to return `false` for `hasNextPage` for the `CursorBound` case
```
HasNextPage(allEdges, before, after, first, last)
If first is set:
Let edges be the result of calling ApplyCursorsToEdges(allEdges, before, after).
If edges contains more than first elements return true, otherwise false.
If before is set:
If the server can efficiently determine that elements exist following before, return true.
Return false.
```

wlmyng: I got my `after` and `before`s mixed up, let me sit on this a bit more

wlmyng: Hitting `End::CursorBound` in forward pagination happens when the scan reaches the `before` bound, or `after` goes past the ledger tip

The Ledger Service doesn't allow us to peek across the bound the same way we could do with Postgres today. i suppose we could issue a second call though. But from draining a single stream from `list_transactions` we can't directly tell whether matching items exist past `before`

So i think it's fine to treat `CursorBound` as a terminal condition setting `has_more = false`

amnn: Why can we not do the same trick we do for postgres, which is to turn the exclusive cursor bound into an inclusive one, and check if we get one more element?

wlmyng: For posterity, in the forward pagination case, the trick with postgres is that, if the `after` cursor appears in the result set, then there is a previous page. If we successfully overfetched, or the `before` cursor appears in the result set, then there is a next page.

We could apply this to grpc by modifying the cursors at grpc boundary so that they are included in the result set. However, if the cursor came from a progress watermark, there would be no matched item. Instead, we need to additionally look for the single match past the direction of the provided cursors. 

The possible scenarios that arise from this probe are conditioned on whether any dimension in the filter have additional matches beyond the cursor.
- If at least one dimension is exhausted past the cursor, we definitively know that no more matches are possible in the range, and we can determine this quite efficiently, through prefix scans on the table before getting to the merge-join mechanism (and thus before spending `ScanLimit` budget)
- Otherwise, we either find that 1 matching item, or exhaust `ScanLimit` in doing so, and eventually `CheckpointBound` or `LedgerTip`.

In other words, we can efficiently determine if `hasNextPage/hasPrevPage` is false, but need to do some probing otherwise. This probing is the same work callers have to do anyways to paginate through the partial results we intend to return.

Consequently, I think we should have the ledger reader always report `hasPrev/NextPage` true unless it hits a `CheckpointBound`, `LedgerTip`, or `CursorBound` short-circuit (when the client's cursor bound collapses the scanning range to empty, e.g when beyond ledger tip). Because until that point, we cannot say for sure that we've exhausted the possible list of matching items to return.

The extra cost manifests as at most one extra paginate request from the caller to discover that they've reached the end.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: phew, `ensure!` took some time to wrap my head around haha

amnn (commented):
Leaving a review to move this off my review queue -- feel free to re-request once the design doc has been finalized.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: Updated to be an explicit match statement.

To separate graphql concerns from the ledger reader's concerns, I had the latter return a page as follows:
- items in stream order
- `first_cursor` from the first cursor-bearing frame seen
- `last_cursor` from the last cursor-bearing frame seen

Reader continues draining until it encounters an end frame.

`has_more` is true as long as its not one of the three terminal conditions:
- LedgerTip
- CheckpointBound
- CursorBound + no emitted cursors

`next_cursor` always returns the `last_cursor`, and expects `Some` if `has_more` is also true


So then it's up to graphql to determine which cursor is the `startCursor` and `endCursor` respectively, and re-order items so that it is consistent across `first` and `last` queries

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: 1. At the `None` branch, we did not receive an `End` frame, so we'll treat that as `has_more`
2. At the end, outside the loop, if `page.has_more()` then we expect a `next_cursor`
The `CursorBound` expectation is handled in `fn has_more`

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,510 @@
...
+    stream: S,
+) -> anyhow::Result<StreamPage<I>>
+where
+    R: Into<FrameKind<I>>,
+    S: Stream<Item = Result<R, tonic::Status>>,
+{
+    futures::pin_mut!(stream);
+    let mut page = StreamPage::default();
+    loop {
+        match stream.next().await {
```
tpham-mysten: Can stream.next().await hang indefinitely here? If the server stalls or the peer dies mid-stream, is there an enforced timeout that bounds this await, or could the drain block forever? I see statement_timeout is applied conditionally on the channel above; does that bound the whole streaming drain, and what happens when it's None?

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,510 @@
...
+                ensure!(
+                    !page.items.is_empty() || page.last_cursor.is_some(),
+                    "{rpc_name} stream {:?} with no progress: {}",
+                    status.code(),
+                    status.message(),
+                );
+                break;
+            }
+            Some(Err(status)) => {
+                bail!("{rpc_name} stream error: {}", status.message());
```
tpham-mysten: nit: Should we include status.code() here too?

tpham-mysten (approved):
Look great! Just a question on a potential hanging problem in stream.next()

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,510 @@
...
+        }
+
+        let channel = endpoint.connect_lazy();
+        let layered =
+            GrpcMetricsLayer::new(prefix.unwrap_or("ledger_grpc"), registry).layer(channel);
+
+        let timeout = args.statement_timeout();
+        let max_decoding_message_size = args
+            .ledger_grpc_max_decoding_message_size
+            .unwrap_or(DEFAULT_MAX_DECODING_MESSAGE_SIZE);
```
amnn: Rather than duplicating the default logic in a bunch of places:

```rust
#[derive(clap::Args, Debug, Clone)]
pub struct LedgerGrpcArgs {
    /// Timeout for gRPC statements to the ledger service, in milliseconds.
    #[arg(long)]
    pub ledger_grpc_statement_timeout_ms: Option<u64>,

    /// Maximum gRPC decoding message size for Ledger service responses, in bytes.
    #[arg(long, default_value_t = Self::default().ledger_grpc_max_decoding_message_size)]
    pub ledger_grpc_max_decoding_message_size: usize,
}

impl Default for LedgerGrpcArgs {
    fn default() -> Self {
        Self {
            ledger_grpc_statement_timeout_ms: None,
            ledger_grpc_max_decoding_message_size: 32 * 1024 * 1024
        }
    }
}
```

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,510 @@
...
+    pub fn next_cursor(&self) -> Option<&Bytes> {
+        if self.has_more() {
+            Some(
+                self.last_cursor
+                    .as_ref()
+                    .expect("invariant: has_more implies last_cursor is Some"),
+            )
+        } else {
+            self.last_cursor.as_ref()
+        }
```
amnn: If you need this assertion, then I'd write it like this:

```suggestion
        assert!(!self.has_more() || self.last_cursor.is_some());
        self.last_cursor.as_ref()
```

But it seems like you're potentially setting up a powder keg, because what happens if the defensive measure in `has_more` triggers (where we default to returning `true` because we get fed `Unspecified` or a variant we don't know about) and for that variant, there is no cursor.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,510 @@
...
+    fn apply(&mut self, frame: FrameKind<I>) -> bool {
+        match frame {
+            FrameKind::Item { item, cursor } => {
+                if cursor.is_some() {
+                    if self.first_cursor.is_none() {
+                        self.first_cursor = cursor.clone();
+                    }
+                    self.last_cursor = cursor;
+                }
+                self.items.push(item);
```
amnn: When we turn this into a GraphQL response, we're going to need to construct a cursor for each item, for its edge. Given that, does it make sense to associate each item with its cursor here too? 

If you do that, would that potentially simplify the `first_cursor`/`last_cursor` logic? Rather than setting each item you see as the first or last cursor, the implementation here could look something like this:

```rust
match frame {
    FrameKind::Item { item, cursor } => {
        self.last_cursor = None;
        self.items.push((item, cursor));
    }

    FrameKind::Watermark { cursor } => {
        if self.items.is_empty() {
            self.first_cursor = cursor.clone();
        } else {
            self.last_cursor = cursor.clone();
        }
    }

    FrameKind::End { reason } => {
        self.end_reason = Some(
            proto::QueryEndReason::try_from(reason)
                .unwrap_or_else(|_| {
                    warn!(reason, "unknown QueryEndReason");
                    proto::QueryEndReason::Unspecified;
                })
        );
        
        return true;
    }

    FrameKind::Unknown => {
        warn!("unknown Frame");
    }
}
```

And the cursor implementations would look like so:

```rust
    fn first_cursor(&self) -> Option<&Bytes> {
        if let Some(first) = &self.first_cursor {
            return Some(&first);
        }

        if let Some((_, first)) = self.items.first() {
            return Some(&first);
        } 

        None
    }

    fn last_cursor(&self) -> Option<&Bytes> {
        if let Some((_, last)) = self.items.last() {
            return Some(&last);
        }

        if let Some(last) = &self.last_cursor {
            return Some(&last);
        }

        None
    }
```

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,510 @@
...
+                        );
+                        Some(grpc_alpha::QueryEndReason::Unspecified)
+                    }
+                };
+                return true;
+            }
+            FrameKind::Unknown => {
+                warn!("list stream: server sent empty or unrecognized Frame");
+            }
+        }
```
amnn: Don't introduce labels into tracing messages -- let tracing deal with that for you.


```suggestion
                    Err(_) => {
                        warn!(reason, "unknown QueryEndReason");
                        Some(grpc_alpha::QueryEndReason::Unspecified)
                    }
                };
                return true;
            }
            FrameKind::Unknown => {
                warn!("unknown Frame");
            }
        }
```

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,510 @@
...
+use std::time::Duration;
+
+use anyhow::Context;
+use anyhow::bail;
+use anyhow::ensure;
+use bytes::Bytes;
+use futures::Stream;
+use futures::StreamExt;
+use prometheus::Registry;
+use sui_rpc::proto::sui::rpc::v2alpha as grpc_alpha;
```
amnn: nit: I think in other places we use this convention.

```suggestion
use sui_rpc::proto::sui::rpc::v2alpha as proto;
```

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,510 @@
...
+impl From<grpc_alpha::ListTransactionsResponse> for FrameKind<grpc_alpha::TransactionItem> {
+    fn from(response: grpc_alpha::ListTransactionsResponse) -> Self {
+        use grpc_alpha::list_transactions_response::Response;
+
+        let Some(response) = response.response else {
+            return FrameKind::Unknown;
+        };
+        match response {
+            Response::Item(item) => {
+                let cursor = item.watermark.as_ref().and_then(|w| w.cursor.clone());
```
amnn: Why take the watermark as a reference and then clone its cursor? Can't you move out out of `item`? You own it.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
amnn: Note that this is now a `while let Some(result) = stream.next().await {` loop.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -0,0 +1,510 @@
...
+            // propagate as error only if zero progress, so the caller can reshape.
+            Some(Err(status))
+                if matches!(
+                    status.code(),
+                    tonic::Code::DeadlineExceeded | tonic::Code::Cancelled
+                ) =>
+            {
+                ensure!(
+                    !page.items.is_empty() || page.last_cursor.is_some(),
+                    "{rpc_name} stream {:?} with no progress: {}",
```
amnn: What does the `:?` format for `status.code` produce here? Is it an intelligible message we'd be happy to pass on to the user via an internal server error?

amnn (commented):
Couple more comments from me -- some simplifications, but main thing is the various defensive measures for schema compatibility seem to contradict each other.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
amnn: This also seems to be a duplicate of the invariant that you're checking at the end of `drain_list_stream`, that's a better place for an invariant like this (closer to the source of the data), and the bad interaction between defensive measures is less problematic there because it won't cause a panic.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: Ah, interesting ... yeah, that makes sense. Each item already has a cursor, as it's retrieved from `let cursor = item.watermark.as_ref().and_then(|w| w.cursor.clone());`, so we don't need to associate each item with its cursor

My intention was to let the first instance of Item or Watermark be responsible for setting `first_cursor` once. But there should be a way to simplify

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: ah, true, updated accordingly

wlmyng: When we set the request timeout, that applies to the stream as well. E.g if we set 30 seconds, we'll time out at 30 seconds. Graphql additionally has a timeout middleware, so we don't hang indefinitely. There is an enforced timeout that bounds the await

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: That makes sense, the assertion applies at the end of a drained stream

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: It would print the variant name: DeadlineExceeded, Cancelled, Internal, etc.

I think this is safe to do as any gRPC caller would see the same information

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: I'm going to relax `fn next_cursor` to just return a ref to `last_cursor`, and consolidate the invariant check to `drain_list_stream`

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: I think the simplification as proposed pushes the book-end cursors in a different direction

Stream: [watermark@A, watermark@B, watermark@C] (empty page, scan walked through it).

  - Each watermark hits the items.is_empty() branch → self.first_cursor = cursor.clone().
  - After 3 watermarks: self.first_cursor = C. The earlier ones are overwritten.
  - self.last_cursor was never set (items always empty, so the else branch never fired).
  - Accessors: first_cursor() = C, last_cursor() = None.

what we want instead is first_cursor = A, last_cursor = C. Paginating backwards on the floor of the scanned range, continue paginating in the same direction on the ceiling of the scanned range

Graphql can still access items, as the ledger grpc reader just passes through the TransactionItem  - to access the item cursor, we do TransactionItem.watermark.cursor

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: True, but we store the gRPC transactionItem as is on the stream page, as I wanted the reader to just drain the stream, pass the items through, and capture scan cursors

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
amnn: > Each item already has a cursor, as it's retrieved from `let cursor = item.watermark.as_ref().and_then(|w| w.cursor.clone());`, so we don't need to associate each item with its cursor

My question was referring to `StreamPage` -- we parse the cursor into `FrameKind` from the streaming response (that's where the `let cursor = ...` line you mentioned comes from), but we discard it during draining, and later when we present it in GraphQL we're going to need that cursor (for each item), so it doesn't make sense to discard it.

The simplification to draining logic then follows from the fact that we would need to retain those cursors in `StreamPage<I>`, but yes you're right that my code snippet doesn't quite work correctly -- maybe this would be better?

```rust
match frame {
    FrameKind::Item { item, cursor } => {
        self.last_cursor = None;
        self.items.push((item, cursor));
    }

    FrameKind::Watermark { cursor } => {
        self.last_cursor = cursor.clone();
        if self.items.is_empty() && self.first_cursor.is_none() {
            self.first_cursor = cursor.clone();
        }
    }

    FrameKind::End { reason } => {
        self.end_reason = Some(
            proto::QueryEndReason::try_from(reason)
                .unwrap_or_else(|_| {
                    warn!(reason, "unknown QueryEndReason");
                    proto::QueryEndReason::Unspecified;
                })
        );
        
        return true;
    }

    FrameKind::Unknown => {
        warn!("unknown Frame");
    }
}
```

And the new cursor implementation (should flip the order of checks for `last_cursor` -- if there is cursor from the watermark, always prefer it):

```rust
    fn first_cursor(&self) -> Option<&Bytes> {
        if let Some(first) = &self.first_cursor {
            return Some(&first);
        }

        if let Some((_, first)) = self.items.first() {
            return Some(&first);
        } 

        None
    }

    fn last_cursor(&self) -> Option<&Bytes> {
        if let Some(last) = &self.last_cursor {
            return Some(&last);
        }

        if let Some((_, last)) = self.items.last() {
            return Some(&last);
        }

        None
    }
```

amnn: Okay, it finally clicked when I read your reply to the cursor cloning comment below -- the cursor **is** currently available, it's just not known to the `StreamPage`. So my question changes -- should we make it so that it is known to `StreamPage`? I.e. when you drain/interpret a `ListTransactionsResponse`, you pull apart the `TransactionItem` into its `ExecutedTransaction`, and separately its cursor, so that the generic pagination logic can take on the responsibility of building up GraphQL's connections for this data?

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
amnn: Ah yep, penny dropped -- this is also related to the question I had above (and I've followed up there).

crates/sui-indexer-alt-reader/src/ledger_grpc_reader.rs
```diff
@@ -23,7 +23,7 @@ use crate::metrics::GrpcMetricsService;
 
 const DEFAULT_MAX_DECODING_MESSAGE_SIZE: usize = 32 * 1024 * 1024;
```
amnn: nit: would suggest getting rid of this constant now, to discourage proliferating the defaulting logic instead of just relying on the `Default` impl (just inline it there).

amnn (approved):
Thanks @wlmyng, all makes sense now. I think we may still want to consider making `StreamPage` aware of the per-item cursor instead of stashing it in its generic parameter, but that is not a blocker.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
wlmyng: Yeah I think that's reasonable, changing `FrameKind` and `StreamPage` as follows:

```

/// A single item from a list stream, paired with the cursor reported alongside it. The cursor is
/// `None` when the server emitted the item without an attached watermark.
#[derive(Debug, Clone)]
pub struct PageItem<T> {
    pub payload: T,
    pub cursor: Bytes,
}

/// A page drained from a single gRPC list stream.
#[derive(Debug, Clone)]
pub struct StreamPage<T> {
    /// Items that matched the filters, in stream order, each paired with its own cursor.
    pub items: Vec<PageItem<T>>,
    first_wm_cursor: Option<Bytes>,
    last_wm_cursor: Option<Bytes>,
    pub end_reason: Option<proto::QueryEndReason>,
}

#[derive(Debug)]
enum FrameKind<T> {
    Item { payload: T, cursor: Bytes },
    Watermark { cursor: Option<Bytes> },
    End { reason: i32 },
    Unknown,
}

```

I'll adopt the implementation proposed too

PR #26897, opened 2026-06-04
Custom `TransactionConnection` to support partially-filled results from bitmap streaming 2/n

wlmyng:
## Description 

Once we switch the backend from postgres to bitmap streams, we can return partially-filled pages (where previously we would timeout.) In order to do so, we need to define custom connections instead of using async-graphql's stock `Connection` (which derives cursors from `items`).

PR stack:
https://github.com/MystenLabs/sui/pull/26894

## Test plan 

Existing tests should pass

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
@@ -75,24 +75,6 @@ pub(crate) struct TransactionConnection {
     pub page_info: PageInfo,
 }
 
-#[Object]
-impl TransactionConnection {
-    /// Information to aid in pagination.
-    async fn page_info(&self) -> &PageInfo {
-        &self.page_info
-    }
-
-    /// A list of edges.
-    async fn edges(&self) -> &[Edge<String, Transaction, EmptyFields>] {
-        &self.edges
-    }
-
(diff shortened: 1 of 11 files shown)


crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -67,6 +70,29 @@ pub(crate) struct TransactionContents {
 
 pub(crate) type CTransaction = JsonCursor<u64>;
 
+pub(crate) struct TransactionConnection {
+    pub edges: Vec<Edge<String, Transaction, EmptyFields>>,
+    pub page_info: PageInfo,
+}
+
+#[Object]
```
amnn: nit: file order -- move below the `#[Object]` block for `Transaction`.

amnn (approved):
LGTM! We should maybe consider dropping the `From<Connection<...>>` impl in favour of constructors on `TransactionConnection` (like `TransactionConnection::empty`) but that can be a follow-up.

PR #26900, opened 2026-06-05
gasless rosetta

nickvikeras:
## Description

Enable free tier stablecoin payments in rosetta.

### DIFF
--- a/crates/sui-rosetta/src/types/internal_operation/pay_coin.rs
+++ b/crates/sui-rosetta/src/types/internal_operation/pay_coin.rs
@@ -5,9 +5,11 @@ use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 
+use move_core_types::identifier::Identifier;
 use sui_rpc::client::Client;
 use sui_rpc::proto::sui::rpc::v2::{GetBalanceRequest, Object, owner::OwnerKind};
 use sui_sdk_types::{Address, TypeTag as SdkTypeTag};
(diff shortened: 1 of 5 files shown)


crates/sui-rosetta/src/types/internal_operation/pay_coin.rs
```diff
@@ -169,11 +218,114 @@ pub fn pay_coin_pt(
...
+        .map(|&v| builder.pure(v))
+        .collect::<Result<Vec<_>, _>>()?;
+    let split_result = builder.command(Command::SplitCoins(source, amount_args));
+    let Argument::Result(split_idx) = split_result else {
+        anyhow::bail!("Expected Result argument from SplitCoins");
+    };
+
+    // Deposit each split piece into the recipient's address balance.
+    for (i, recipient) in recipients.into_iter().enumerate() {
+        coin_send_funds(
```
bmwill, line 296: Did we decide to default all transfers to ABs right?

bmwill (approved):

crates/sui-rosetta/src/types/internal_operation/pay_coin.rs
nickvikeras: No that's not what this does. We simulate all non-SUI transfers as AB, and if they not are eligible for free tier (i.e I believe coinable supports WAL/DEEP/NS now), we fall back to coin transfers.

At some point in the near future we should IMHO just default everything to AB for simplicity.

PR #26912, opened 2026-06-07
Fix rpc-index coin-balance double-count after an unclean restart

bmwill:
## Description 

Summary

  A fullnode's rpc-index could report a coin balance at twice its true value after an unclean restart. The root cause is a watermark
  consistency gap in checkpoint execution; the fix records a crash-consistent watermark and uses it to gate the rpc-index restore and
  forward indexing. This is intermittently reproducible across an unclean fullnode restart, and is consensus-safe (no protocol or
  on-disk-format change beyond an additive, self-populating column family).

  Background

  The checkpoint executor commits a checkpoint's transaction outputs (objects, effects) to the perpetual store and then bumps the
  checkpoint store's highest_executed watermark in a separate write (there is even a fail_point!("crash") between the two). An abrupt
  stop in that window leaves the object writes durable while highest_executed still lags — so the perpetual store's live object set is
  momentarily ahead of the executed watermark.

  The rpc-index rebuild reads the live object set directly and stamps its watermark at a "restore target." When that target was
  highest_executed but the live set already reflected a later in-flight checkpoint, the restore captured that checkpoint's objects, and
  then the checkpoint executor — which also resumes from highest_executed — re-applied the same checkpoint through forward indexing.
  Additive indexes (coin balances) got counted twice.

  What's in this PR

  1. core: record a highest-committed-checkpoint watermark atomic with outputs
  Adds a highest_committed_checkpoint singleton column family to AuthorityPerpetualTables, written in the same atomic batch as the
  checkpoint's outputs via a new ExecutionCacheCommit::set_highest_committed_checkpoint_in_batch hook the executor calls between
  building and committing the batch. The watermark is therefore always consistent with the durable object set, even after a crash. It is
  always >= highest_executed.

  2. core: stop rpc-index double-counting checkpoints after a rebuild
  - RpcIndexStore::init uses highest_committed as the restore target (falling back to highest_executed for databases written before the
  watermark existed), so the stamped watermark matches the checkpoint the live set actually reflects. A fresh node stamps no watermark,
  so genesis is still indexed by forward indexing rather than dropped.
  - commit_update_for_checkpoint now gates forward updates on Watermark::Indexed: it commits only the next contiguous checkpoint
  (watermark + 1), drops a batch for a checkpoint at or below the watermark (already covered), and panics on a gap. This is what
  actually prevents the re-application, and it also closes the normal-operation crash window where a running indexed node stops between
  the forward commit and the highest_executed bump.
  - Adds the enabling_legacy_index_rebuilds_without_double_counting e2e test.

  ▎ Note: the watermark change alone is necessary but not sufficient for the legacy rpc-index. Unlike the embedded store (whose tip
  ▎ indexer gates on its own cohort watermark), legacy forward indexing is driven by the executor from highest_executed; the
  ▎ commit_update_for_checkpoint drop is what stops the re-application.

  Testing

  The new e2e test runs a transfer with indexing off, restarts the node with indexing on, advances the chain past the rebuild point so
  forward indexing re-processes the restored checkpoints before the test reads (reading too early masks the bug), and asserts the
  recipient's balance is reported exactly once.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/rpc/rpc_index_restore.rs
+++ b/crates/sui-e2e-tests/tests/rpc/rpc_index_restore.rs
@@ -1,0 +1,347 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Rebuild behavior of the legacy `rpc-index` index backend (`enable_indexing`).
+//!
+//! A node that ran without indexing and is restarted with it on rebuilds the
+//! `rpc-index` from the perpetual store: `RpcIndexStore::init` bulk-loads the
+//! live object set (the `balance` index) and backfills the historical
+//! checkpoint indexes (the `transaction_bitmap`), then the checkpoint executor
+//! resumes forward indexing from `highest_executed`.
+//!
+//! The bulk restore reads the live object set, which the perpetual store writes
+//! atomically with the `highest_committed` watermark. The checkpoint store's
+//! `highest_executed` watermark is bumped in a *separate* write afterward, so an
+//! unclean stop can leave the live set reflecting a checkpoint that
+//! `highest_executed` does not yet count. If the restore stamps its watermark at
+//! the lagging `highest_executed`, the executor re-applies the checkpoints in
+//! `(highest_executed, highest_committed]` that the live set already captured,
+//! double-counting additive indexes -- a recipient's coin balance shows up at
+//! twice its value. This test exercises that path: it asserts the recipient's
+//! balance is reported exactly once after the rebuild.
+//!
+//! The unit under test is a fullnode spawned into the swarm separately from the
+//! cluster's primary fullnode. The cluster's wallet executes transactions
+//! against the primary (which stays up the whole time); the dedicated node
+//! follows by state sync and is restarted with a mutated `NodeConfig.rpc`
+//! between runs. Its `db_path` is stable across restarts, so each restart sees
+//! the previous run's perpetual store and exercises the real rebuild path.
+//!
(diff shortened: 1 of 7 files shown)


amnn (approved):
Nice find -- what happens between when the write batch containing the new atomic checkpoint update lands and when the separate write to `last_executed_checkpoint` lands? (Could we eventually move to just having the one watermark that gets updated atomically?)

bmwill:
So this new watermark is just a way to tell us concretely the set of data this db has and marks the live object set. It doesn't change the semantics of how core executes checkpoints. 

I think in hindsight it probably would make more sense to have one db vs 3 for this data but that would be a much larger change right now.

crates/sui-e2e-tests/tests/rpc/rpc_index_restore.rs
```diff
@@ -0,0 +1,347 @@
...
+    let name = config.protocol_public_key();
+    let rpc_url = format!("http://{}", config.json_rpc_address);
+    cluster.swarm.spawn_new_node(config).await;
+    (name, rpc_url)
+}
+
+/// Stop the fullnode `name`, swap in `rpc` (the `db_path` is unchanged), and
+/// restart it. The stop releases the previous run's RocksDB locks because no
+/// strong `SuiNodeHandle` is held across the call.
+async fn restart_fullnode(cluster: &TestCluster, name: &AuthorityName, rpc: RpcConfig) {
```
bmwill, line 111: @mystenmark

PR #26913, opened 2026-06-07
fix(sui-benchmark): prevent duplicate tx digests in benchmark workloads

mystenmark:
## Summary

- Adds `TestTransactionBuilder::ensure_unique() -> Self` — a chainable method that injects a random `u64` unused pure input (via `force_separate_pure`) so every subsequent `build()` / `build_and_sign()` produces a unique digest even when all logical inputs are identical
- `build()` itself is unchanged, so tests that intentionally build duplicate transactions are unaffected
- All 38 sui-benchmark call sites now call `.ensure_unique()` before building

## Root cause

In the alias-revocation benchmark workload (`composite/mod.rs`), both the `Active` and `Revoked` states build structurally identical transactions — `transfer_sui(None, sender)` signed by the alias keypair. When a bundle submission fails with a transport error during rolling restarts, the workload marks the transaction failed and does not update gas tracking. However the transaction may have already reached consensus and committed.

Later, the `Revoked` state fires with the same stale gas object version and produces the exact same digest. The validator returns "already executed" with the cached success result, tripping the `debug_fatal!` assertion that the post-revocation tx must fail.

## Test plan

- Verified the specific failing seed passes: `MSIM_TEST_SEED=1780854396117 cargo simtest --test simtest -E 'test(=test::test_simulated_load_rolling_restarts_all_validators)'`
- Ran post-fix seed search across 200 seeds with no failures

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17093,6 +17093,7 @@ version = "0.1.0"
 dependencies = [
  "bcs",
  "move-core-types",
+ "rand 0.8.5",
  "shared-crypto",
  "sui-genesis-builder",
  "sui-move-build",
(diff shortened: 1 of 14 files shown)


alex-mysten (approved):
LGTM!

PR #26919, opened 2026-06-08
fix(sui-core): allow epoch to advance by more than one in submit_tx retry

mystenmark:
## Summary

When a transaction receives `ValidatorHaltedAtEpochEnd`, `authority_server.rs` records `next_epoch = current_epoch + 1`, then calls `wait_for_epoch(next_epoch)` and retries. The previous code panicked (via `debug_fatal!`) if the returned epoch wasn't exactly `next_epoch`.

This was a false invariant. `wait_for_epoch` returns the current epoch as soon as it is `>= target_epoch`, so the returned value can be higher than `next_epoch` — this happens legitimately under crashes and delays, where the network can reconfigure multiple times before the retry loop wakes up. The panic surfaced as a flaky `test_simulated_load_reconfig_with_crashes_and_delays` failure.

Fix: drop the strict equality check and always retry after a successful `wait_for_epoch` return. Any epoch `>= next_epoch` is a valid signal that reconfiguration completed.

Also includes a small improvement to `scripts/simtest/seed-search.py`: new `--error-regex` flag that matches test output against a regex and exits early with the repro command when a matching failing seed is found, making targeted seed searches much faster.

## Test plan
- [ ] `SUI_SKIP_SIMTESTS=1 cargo nextest run -p sui-core` passes
- [ ] `cargo xclippy -D warnings` on `sui-core` passes
- [ ] Run `scripts/simtest/seed-search.py --test simtest test::test_simulated_load_reconfig_with_crashes_and_delays --exact --error-regex "expected epoch.*after reconfiguration"` — should find no matching seeds with this fix applied

### DIFF
--- a/crates/sui-core/src/authority_server.rs
+++ b/crates/sui-core/src/authority_server.rs
@@ -8,7 +8,7 @@ use fastcrypto::traits::KeyPair;
 use futures::{TryFutureExt, future};
 use itertools::Itertools as _;
 use mysten_common::ZipDebugEqIteratorExt;
-use mysten_common::{assert_reachable, debug_fatal};
+use mysten_common::assert_reachable;
 use mysten_metrics::spawn_monitored_task;
 use prometheus::{
     Gauge, Histogram, HistogramVec, IntCounter, IntCounterVec, Registry,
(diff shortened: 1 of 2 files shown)


mystenmark:
Splitting into two independent PRs.

PR #26929, opened 2026-06-08
Resilience against crashes during execution

mystenmark:
## Description

Certain transactions can deterministically panic a node during execution; without recovery the node crash-loops on every restart. This adds crash-recovery (test/antithesis-only, opt-in via a content-addressed poison marker; inert in release): a panic hook records the executing transaction's digest, and on restart the node re-executes it with a new terminal `InternalExecutionError` early error instead of running the logic that crashes.

A crashed transaction is **not** dropped — it is still assigned versions, executed (as a deterministic failure that advances its shared-object versions), and checkpointed like any other transaction. This keeps version assignment, the `ConsensusCommitPrologue` version-assignment commitment, and checkpoint contents a pure function of the consensus commit, so a node's pre-crash build and post-restart rebuild agree. (An earlier drop-based approach forked checkpoints and stalled the builder because the node-local crashed set leaked into consensus-committed state.) Fullnodes and lagging nodes reproduce the effect from the certified checkpoint without re-running the crashing logic.

## Test plan

Covered by `test_simulated_load_reconfig_with_crashes_and_delays` (sui-benchmark simtest), which injects deterministic crashes under load with reconfiguration. Verified on the seeds that previously forked / hung the checkpoint builder / hit the accumulator assertion, plus a 60-seed search with no failures.

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6060,9 +6060,9 @@ checksum = "f90f7dce0722e95104fcb095585910c0977252f286e354b5e3bd38902cd99988"
 
 [[package]]
 name = "futures-timer"
-version = "3.0.3"
+version = "3.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f288b0a4f20f9a56b5d1da57e2227c661b7b16168e2f72365f57b63326e29b24"
+checksum = "af43fadb8a98512d547e37b4e92e0ced13e205c061b87b4623eff01d918d6968"
 
 [[package]]
 name = "futures-util"
--- a/crates/sui-benchmark/src/drivers/bench_driver.rs
+++ b/crates/sui-benchmark/src/drivers/bench_driver.rs
@@ -878,6 +878,14 @@ async fn run_bench_worker(
                     err,
                     transaction.digest()
                 );
+                #[cfg(msim)]
+                if sui_core::crash_recovery::crash_recovery_probability().is_some()
+                    && sui_core::crash_recovery::should_poison_transaction(
+                        transaction.data().transaction_data(),
+                    )
+                {
+                    return NextOp::Failure { payload };
+                }
                 match payload.get_failure_type() {
                     Some(ExpectedFailureType::NoFailure) => {
                         panic!(
--- a/crates/sui-benchmark/src/workloads/addr_bal_deposit.rs
+++ b/crates/sui-benchmark/src/workloads/addr_bal_deposit.rs
@@ -204,7 +204,13 @@ impl Workload<dyn Payload> for AddrBalDepositWorkload {
                     vec![coin_balance, recipient_arg],
                 );
             }
-            let tx = tx_builder.ensure_unique().build_and_sign(keypair.as_ref());
+            // Seeding transactions do not opt in to crash injection (they carry no crash-opt-in
+            // marker), so they can never be poisoned.
+            let tx_data = tx_builder.ensure_unique().build();
+            let tx = sui_types::transaction::Transaction::from_data_and_signer(
+                tx_data,
+                vec![keypair.as_ref()],
+            );
             let proxy_ref = execution_proxy.clone();
             futures.push(async move {
                 let result = proxy_ref.execute_transaction_block(tx).await;
--- a/crates/sui-benchmark/src/workloads/composite/mod.rs
+++ b/crates/sui-benchmark/src/workloads/composite/mod.rs
@@ -11,8 +11,8 @@ pub use operations::{
     AddressBalanceDeposit, AddressBalanceOverdraw, AddressBalanceWithdraw, AuthenticatedEventEmit,
     CoinReservationWithdraw, INVALID_ALIAS_TX, ImmutableObjectRead, ObjectBalanceDeposit,
     ObjectBalanceOverdraw, ObjectBalanceWithdraw, OperationDescriptor, RandomnessRead,
-    SharedCounterIncrement, SharedCounterRead, TestCoinAddressDeposit, TestCoinAddressWithdraw,
-    TestCoinMint, TestCoinObjectWithdraw,
+    RequestCrash, SharedCounterIncrement, SharedCounterRead, TestCoinAddressDeposit,
+    TestCoinAddressWithdraw, TestCoinMint, TestCoinObjectWithdraw,
 };
 use rand::seq::SliceRandom;
 
@@ -319,6 +319,9 @@ impl CompositeWorkloadConfig {
         probabilities.insert(AuthenticatedEventEmit::NAME, 0.1);
         probabilities.insert(ImmutableObjectRead::NAME, 0.2);
         probabilities.insert(CoinReservationWithdraw::NAME, 0.1);
+        // Every composite transaction opts in to crash injection; whether it is actually poisoned
+        // is then decided (content-addressed) at the configured crash probability.
+        probabilities.insert(RequestCrash::NAME, 1.0);
         Self {
             probabilities,
             alias_tx_probability: 0.3,
@@ -343,8 +346,13 @@ impl CompositeWorkloadConfig {
             .map(|desc| (desc.factory)())
             .collect();
 
-        if ops.is_empty()
-            && let Some(desc) = ALL_OPERATIONS.first()
+        // Ensure at least one command-producing operation. RequestCrash only attaches an unused
+        // marker argument and produces no command, so a transaction containing only it would be an
+        // empty (command-less) PTB.
+        if ops.iter().all(|op| op.name() == RequestCrash::NAME)
+            && let Some(desc) = ALL_OPERATIONS
+                .iter()
+                .find(|desc| desc.name != RequestCrash::NAME)
         {
             ops.push((desc.factory)());
         }
--- a/crates/sui-benchmark/tests/simtest.rs
+++ b/crates/sui-benchmark/tests/simtest.rs
@@ -509,7 +509,25 @@ mod test {
         // quorum, empirically ~6% per-validator skip rate produces ~5% DKG failure per epoch.
         register_fail_point_if("rb-dkg", || thread_rng().gen_bool(0.06));
 
-        test_simulated_load(test_cluster, 120).await;
+        // Only transactions that opt in (by carrying the crash-opt-in marker arg, attached by the
+        // composite workload) are eligible to be poisoned, so setup/gas transactions are never
+        // crashed and a transaction's poison status is consistent for the whole run. The crash
+        // probability can therefore be enabled before load generation begins.
+        const CRASH_PROB: f32 = 0.02;
+        test_simulated_load_with_test_config(
+            test_cluster,
+            120,
+            SimulatedLoadConfig::default(),
+            None,
+            None,
+            Some(|_cluster: Arc<TestCluster>| {
+                register_fail_point_if("crash-with-tx-logging", || true);
+                sui_core::crash_recovery::set_crash_recovery_probability(CRASH_PROB as f64);
+                std::future::ready(())
+            }),
+            true,
+        )
+        .await;
     }
 
     #[sim_test(config = "test_config()")]
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -805,6 +805,11 @@ pub struct ExecutionEnv {
     /// Transactions that must finish before this transaction can be executed.
     /// Used to schedule barrier transactions after non-exclusive writes.
     pub barrier_dependencies: Vec<TransactionDigest>,
+    /// Set when executing from a source (e.g. a certified checkpoint) whose effects show the
+    /// transaction failed with `InternalExecutionError` (a crash-recovered poison transaction).
+    /// Forces the same early error on re-execution so the node reproduces the effects without
+    /// running the real logic that would crash the process again.
+    pub internal_execution_error: bool,
 }
 
 impl Default for ExecutionEnv {
@@ -814,6 +819,7 @@ impl Default for ExecutionEnv {
             expected_effects_digest: None,
             funds_withdraw_status: FundsWithdrawStatus::MaybeSufficient,
             barrier_dependencies: Default::default(),
+            internal_execution_error: false,
         }
     }
 }
@@ -838,6 +844,11 @@ impl ExecutionEnv {
 
     pub fn with_insufficient_funds(mut self) -> Self {
         self.funds_withdraw_status = FundsWithdrawStatus::Insufficient;
+        self
+    }
+
+    pub fn with_internal_execution_error(mut self) -> Self {
+        self.internal_execution_error = true;
         self
     }
 
@@ -1472,6 +1483,8 @@ impl AuthorityState {
         let _metrics_guard = self.metrics.internal_execution_latency.start_timer();
 
         let tx_digest = certificate.digest();
+        let _crash_guard =
+            crate::crash_recovery::register_executing_transaction(*tx_digest, self.name);
 
         if let Some(fork_recovery) = &self.fork_recovery_state
             && let Some(override_digest) = fork_recovery.get_transaction_override(tx_digest)
@@ -1987,11 +2000,38 @@ impl AuthorityState {
             self.config.certificate_deny_config.certificate_deny_set(),
             &execution_env.funds_withdraw_status,
         );
+        // Crash-recovery: a transaction that previously crashed the node is re-executed with an
+        // early error instead of running its real (crashing) logic. It still assigns versions,
+        // executes (as a deterministic failure), advances shared-object versions, and is
+        // checkpointed like any other transaction, so consensus-committed state stays a pure
+        // function of the commit. Two triggers: the local crashed-set (validator re-executing after
+        // its own crash) and the ExecutionEnv flag (executing from a certified checkpoint whose
+        // effects are InternalExecutionError, so the node never runs the crashing logic). Prepended
+        // so the effect status is InternalExecutionError regardless of any other early error.
+        let early_execution_error = if epoch_store.is_crashed_transaction(&tx_digest)
+            || execution_env.internal_execution_error
+        {
+            let mut errors = vec![ExecutionErrorKind::InternalExecutionError];
+            if let Some(existing) = early_execution_error {
+                errors.extend(existing);
+            }
+            NonEmpty::from_vec(errors)
+        } else {
+            early_execution_error
+        };
         let accumulator_version = execution_env.assigned_versions.accumulator_version;
         let execution_params = match early_execution_error {
             None => ExecutionOrEarlyError::ok(accumulator_version),
             Some(errors) => ExecutionOrEarlyError::failed(errors, accumulator_version),
         };
+
+        // Inject the simulated crash only when the transaction actually executes (no early error),
+        // mirroring the execution engine's early-exit-on-early-error. A crash-recovered tx takes the
+        // early-error path above (InternalExecutionError) and must not re-crash. The upstream
+        // effects-cache check ensures this only fires on the first execution attempt.
+        if execution_params.is_ok() && !transaction_data.kind().is_system_tx() {
+            crate::crash_recovery::maybe_crash_for_testing(transaction_data);
+        }
 
         // Skip on early error: the tx will fail anyway and rewriting may fail if the accumulator
         // was deleted.
--- a/crates/sui-core/src/crash_recovery.rs
+++ b/crates/sui-core/src/crash_recovery.rs
@@ -1,0 +1,470 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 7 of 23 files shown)


crates/sui-benchmark/src/drivers/bench_driver.rs
```diff
@@ -1,6 +1,24 @@
...
+/// passed to `should_poison_transaction` in the fail point.
+#[cfg(msim)]
+pub fn set_crash_recovery_probability(prob: f32) {
+    sui_core::crash_recovery::set_crash_recovery_probability(prob as f64);
+}
+
+#[cfg(msim)]
+pub fn crash_recovery_probability() -> Option<f32> {
+    sui_core::crash_recovery::crash_recovery_probability().map(|p| p as f32)
+}
```
mystenmark: lets move these functions into the crash recovery crate, and then `should_poison_transaction()` should take only the transaction digest. this way there is no possibility of different callsites passing the wrong probability

crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -644,6 +644,10 @@ impl CompositePayload {
             }
         }
 
+        // ensure_unique() randomises the transaction digest on every submission so that
+        // repeated calls for the same logical operation produce distinct digests.  This
+        // also means that if a submission happens to fall in the crash-simulation poison
+        // set, the next retry will almost certainly have a different (clean) digest.
```
mystenmark: this comment doesn't make much sense - where do we have a retry loop that rebuilds the transaction every time? probably best to remove the comment

crates/sui-benchmark/src/workloads/addr_bal_deposit.rs
```diff
@@ -204,7 +204,34 @@ impl Workload<dyn Payload> for AddrBalDepositWorkload {
...
+                    );
+                    if !sui_core::crash_recovery::should_poison_transaction(
+                        candidate.digest(),
+                        prob as f64,
+                    ) {
+                        break;
+                    }
+                    tx_data.gas_data_mut().budget += 1;
+                }
+            }
```
mystenmark: lets encapsulte this logic in a function in the crash recovery crate that takes a &mut TransactionData

Also, although its a bit more code, lets do the mining by adding an unused PureArg, since its possible that altering the budget can actually change other behavior

crates/sui-core/src/authority.rs
```diff
@@ -1465,6 +1465,11 @@ impl AuthorityState {
         let _metrics_guard = self.metrics.internal_execution_latency.start_timer();
 
         let tx_digest = certificate.digest();
+        // Register the digest in TLS so that a panic hook can identify which transaction caused
+        // the crash. The guard clears the slot on drop and is !Sync to prevent accidental use
+        // across threads in an async context.
```
mystenmark: this comment is unnecessary and overly verbose, a curious reader can read the doc comments for `register_executing_trasnaction`

crates/sui-core/src/authority.rs
```diff
@@ -1495,6 +1500,46 @@ impl AuthorityState {
             return ExecutionOutput::Success((effects, None));
         }
 
+        // Simtest-only: simulate a deterministic transaction-induced crash.
+        //
+        // Must be placed AFTER the effects-cache check above. If placed before, the fail point
+        // would fire on every call to try_execute_immediately including re-executions where the
+        // effects are already committed (e.g. during checkpoint rebuild). Crashing on a
+        // re-execution records the tx in the crash log but the tx is already in a committed
+        // checkpoint, so dropping it on restart produces a checkpoint mismatch.
```
mystenmark: this comment is incoherent, poison transactions cannot be committed because their effects cannot be obtained

crates/sui-core/src/authority.rs
```diff
@@ -1495,6 +1500,46 @@ impl AuthorityState {
...
+        //
+        // A process-global seed is used so that all validators — regardless of which OS thread
+        // they run on — reach the same crash decision for a given digest. This mirrors a real
+        // production bug where the same transaction causes every validator that executes it to
+        // panic. Without crash-recovery the node would crash again on every restart.
```
mystenmark: unneeded comment, perhaps it can go in `should_poison_trnasaction`

crates/sui-core/src/authority.rs
```diff
@@ -1495,6 +1500,46 @@ impl AuthorityState {
...
+                    // as an intentional crash and write the crash log. The hook distinguishes this
+                    // from random node kills (which panic with a PanicWrapper, not a &str).
+                    // catch_unwind lets the hook fire and write the log before we call
+                    // kill_current_node, whose own PanicWrapper panic would not be identified as
+                    // a crash-simulation by the hook.
+                    let _ = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+                        std::panic::panic_any(crate::crash_recovery::CRASH_SIM_PANIC_MSG);
+                    }));
+                    sui_simulator::task::kill_current_node(Some(std::time::Duration::from_secs(
+                        20,
```
mystenmark: this is a long time for the node to be down and will just slow down tests. an immediate restart is fine (100ms)

crates/sui-core/src/authority.rs
```diff
@@ -1495,6 +1500,46 @@ impl AuthorityState {
...
+        #[cfg(msim)]
+        if !certificate.data().transaction_data().kind().is_system_tx() {
+            sui_macros::fail_point_if!("crash-with-tx-logging", || {
+                if crate::crash_recovery::should_poison_transaction(&tx_digest, 0.002) {
+                    // Panic with the crash-simulation marker so the panic hook can identify this
+                    // as an intentional crash and write the crash log. The hook distinguishes this
+                    // from random node kills (which panic with a PanicWrapper, not a &str).
+                    // catch_unwind lets the hook fire and write the log before we call
+                    // kill_current_node, whose own PanicWrapper panic would not be identified as
+                    // a crash-simulation by the hook.
```
mystenmark: Comment can be more concise: " use catch_on_wind to trigger the panic_hook without crashing the test process. "

crates/sui-core/src/authority.rs
```diff
@@ -1495,6 +1500,46 @@ impl AuthorityState {
...
+        // production bug where the same transaction causes every validator that executes it to
+        // panic. Without crash-recovery the node would crash again on every restart.
+        //
+        // We use catch_unwind to fire a plain panic (triggering the crash-recovery hook, which
+        // writes the crash log via TLS), then call kill_current_node once we are back in normal
+        // execution. We cannot call kill_current_node from inside the panic hook because
+        // kill_current_node itself panics (with PanicWrapper), which would be a double-panic.
+        //
+        // Only poison user (programmable) transactions. System transactions (epoch change,
+        // randomness, consensus prologue, etc.) must always succeed or the cluster stalls.
```
mystenmark: also unnecessary comments

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2167,6 +2192,15 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
                             deferral_key,
                             congested_objects
                         );
+                        // Crash-recovered transactions must not reach execution even via
+                        // the congestion-cancel path: their effects are gone and the fail
+                        // point would fire again, causing a second crash.
```
mystenmark: comments shouldn't discuss fail points, thats a testing detail

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2177,6 +2211,21 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+            // The cost observation above keeps IndirectStateObserver consistent with
+            // the pre-crash run, so the ConsensusCommitPrologue's additional_state_digest
+            // is unchanged and locally_computed_checkpoints remains valid across restarts.
+            //
+            // Also bump the congestion tracker so that the shared objects this transaction
+            // was using are no longer marked as congested.  In the pre-crash run the
+            // transaction was eventually scheduled and the bump happened at that point;
+            // without it here the objects stay congested indefinitely, slowing down
+            // other transactions and preventing checkpoint progress.
+            shared_object_congestion_tracker.bump_object_execution_cost(tx_cost, transaction.tx());
```
mystenmark: Remove the giant comment. instead of using the else if construct, fold this into the `else` branch below so that only `scheduled_txns.push` is skipped

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2760,6 +2809,31 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
 
+                // Transactions that caused a panic in a previous run are flagged here.
+                // We set Dropped so callers waiting on effects get a definitive response,
+                // but we do NOT continue — the transaction flows through the rest of the
+                // scheduling pipeline so that get_tx_cost still updates IndirectStateObserver.
+                // That keeps the ConsensusCommitPrologue's additional_state_digest identical
+                // to the pre-crash run, which is required for locally_computed_checkpoints to
+                // remain consistent across restarts.  Lock acquisition is skipped below to
+                // avoid creating a persistent owned-object lock that can never be released.
+                let mut is_crash_recovered = false;
```
mystenmark: don't use `mut` unless necessary, the if stmt can be an expression

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2760,6 +2809,31 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
                 }
 
+                // Transactions that caused a panic in a previous run are flagged here.
+                // We set Dropped so callers waiting on effects get a definitive response,
+                // but we do NOT continue — the transaction flows through the rest of the
+                // scheduling pipeline so that get_tx_cost still updates IndirectStateObserver.
+                // That keeps the ConsensusCommitPrologue's additional_state_digest identical
+                // to the pre-crash run, which is required for locally_computed_checkpoints to
+                // remain consistent across restarts.  Lock acquisition is skipped below to
+                // avoid creating a persistent owned-object lock that can never be released.
```
mystenmark: comment is mostly unnecessary. move the bits about conensus status and object locking to the relevent code

crates/sui-core/src/crash_recovery.rs
```diff
@@ -0,0 +1,353 @@
...
+        Err(e) => {
+            error!(
+                "Failed to open crash-recovery log {}: {e}",
+                log_path.display()
+            );
+            return HashSet::new();
+        }
+        Ok(file) => {
+            let mut digests = HashSet::new();
+            for line in BufReader::new(file).lines() {
```
mystenmark: limit the number of lines we are willing to read to 1024*1024 to prevent OOMing here

crates/sui-core/src/crash_recovery.rs
```diff
@@ -0,0 +1,353 @@
...
+        }
+
+        prev_hook(info);
+    }));
+}
+
+fn append_digest_to_log(path: &Path, digest: TransactionDigest) -> std::io::Result<()> {
+    let mut file = OpenOptions::new().create(true).append(true).open(path)?;
+    writeln!(file, "{}", digest)?;
+    file.flush()
```
mystenmark, line 363: panic hooks can run concurrently if multiple threads panic. acquire a global lock before writing to the file

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2760,6 +2809,31 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+                let mut is_crash_recovered = false;
+                if let Some(tx) = parsed.transaction.kind.as_user_transaction() {
+                    let digest = tx.digest();
+                    if self.crashed_transactions.contains(digest) {
+                        assert_reachable!(
+                            "crash-recovery: dropping previously-crashing transaction"
+                        );
+                        warn!(
+                            ?digest,
+                            "Dropping transaction that caused a crash in a previous run"
```
mystenmark: log the entire transaction as base64 encoded BCS so we can recover it

crates/sui-core/src/crash_recovery.rs
```diff
@@ -0,0 +1,353 @@
...
+                #[cfg(msim)]
+                if info.payload().downcast_ref::<&str>().copied() != Some(CRASH_SIM_PANIC_MSG) {
+                    prev_hook(info);
+                    return;
+                }
+
+                EXECUTING_TX.with(|slot| slot.set(None));
+                if let Some(db_path) = get_db_path(&authority_name) {
+                    let log_path = db_path.join(PANIC_TX_LOG_FILE);
+                    match append_digest_to_log(&log_path, digest) {
```
mystenmark: we need to log the transaction digest and the git commit of the currently running binary - once the git commit changes we will assume the crashing bug has been fixed

crates/sui-core/src/crash_recovery.rs
```diff
@@ -0,0 +1,353 @@
...
+// ---------------------------------------------------------------------------
+// Startup: read crashed transactions
+// ---------------------------------------------------------------------------
+
+/// Read the panic log and return the set of transaction digests that were active during a past
+/// crash.  Returns an empty set if the log file does not exist.
+///
+/// `db_path` should be the same path passed to `install_panic_hook`.
+pub fn load_crashed_transactions(db_path: &Path) -> HashSet<TransactionDigest> {
+    let log_path = db_path.join(PANIC_TX_LOG_FILE);
```
mystenmark, line 381: this function needs to ignore transactions that crashed in a different version of the software (by checking git sha, see above). 

all digests originating from a different git sha binary should be removed from the log (the log can be filtered and re-written).

Cargo.lock
```diff
@@ -6012,9 +6012,9 @@ checksum = "f90f7dce0722e95104fcb095585910c0977252f286e354b5e3bd38902cd99988"
 
 [[package]]
 name = "futures-timer"
-version = "3.0.3"
+version = "3.0.4"
```
mystenmark, line 6063: we probably don't need/want to update this dependency in this pr

mystenmark:
Addressed all review comments in beeb75a7ee:

- **crash_recovery.rs**: `should_poison_transaction` now takes only the digest (reads probability from the global atomic internally); added `mine_non_poison_transaction` helper using an unused PureArg nonce to change the digest; log format changed to `<git_rev> <digest>` per line so entries from a previous binary version are automatically filtered on load (and the log is rewritten to remove stale entries); added `FILE_WRITE_LOCK` to guard concurrent panics writing to the log; `load_crashed_transactions` limited to 1M lines
- **authority.rs**: removed verbose/redundant comments; reduced crash-sim kill delay from 20 s to 100 ms; simplified catch_unwind comment to "use catch_unwind to trigger the panic hook without crashing the test process"
- **consensus_handler.rs**: removed comment about fail points (testing detail); folded crash-recovered check into the `else` branch so only `scheduled_txns.push` is skipped; fixed `let mut` → `let` expression; drop warning now logs full base64-BCS transaction; moved object-lock and Dropped-status notes to the relevant code sites
- **addr_bal_deposit.rs**: mining loop replaced with `mine_non_poison_transaction` call
- **composite/mod.rs**: removed comment that mentioned retry loops
- **Cargo.lock**: reverted unrelated `futures-timer` bump

mystenmark:
Addressed the remaining comments in 01ddfe76:

- **bench_driver.rs**: Removed the wrapper functions entirely. `set_crash_recovery_probability` / `crash_recovery_probability` / `should_poison_transaction` are now called via `sui_core::crash_recovery` directly at all call sites; no duplicate functions.
- **authority.rs**: Removed the incoherent comment about effects being committed (they can't be for poison txns). Collapsed the catch_unwind explanation to one line: "use catch_unwind to trigger the panic hook without crashing the test process."
- **consensus_handler.rs**: Folded the crash-recovered check into the `else` branch so only `scheduled_txns.push` is skipped (no more `else if`). Removed the fail-point mention from the congestion-cancel guard. Trimmed the block comment in the filter path — lock-skip rationale moved inline to the lock-acquisition guard.

(Previous push beeb75a7 addressed the other 15 comments.)

mystenmark:
200-seed sweep (seeds 1–200) passes cleanly with `seed-search.py`. The `assert_reachable!("crash-recovery: dropping previously-crashing transaction")` assertion fired during the sweep, confirming crash recovery triggered in practice.

mystenmark:
Latest commits add Antithesis / non-msim crash-recovery support:

- **Add non-msim crash-recovery path** (9ec0d0872c): `maybe_crash_for_testing` in `crash_recovery.rs` encapsulates both the msim (fail-point + catch_unwind + kill_current_node) and non-msim (`in_test_configuration()` + plain panic) paths; `authority.rs` is a single unconditional call.
- **Default 0.2% probability in Antithesis** (1023baba7a): `crash_recovery_probability()` returns `Some(0.002)` when `in_antithesis()` is true and no explicit value has been set, so crash-recovery cycles fire without any test-harness setup call.

crates/sui-core/src/crash_recovery.rs
```diff
@@ -0,0 +1,497 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Crash-recovery support: detect which transaction was executing when the process panicked.
+//!
+//! # How it works
+//!
+//! 1. At process startup, `install_panic_hook` chains a closure into the existing panic hook.
+//!    When a panic fires, the hook reads a thread-local slot.  If that slot holds a transaction
```
alex-mysten, line 9: I presume the thread-local is how we ensure that the panic can be attributed to a particular user tx?

crates/sui-core/src/crash_recovery.rs
```diff
@@ -0,0 +1,497 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Crash-recovery support: detect which transaction was executing when the process panicked.
+//!
+//! # How it works
```
alex-mysten, line 6: Are we concerned that a non-deterministic panic impacting only one or a small set of validators becomes more difficult to recover from? i.e. that validator is effectively forked until manual recovery steps are taken, when today it would restart and reprocess the same consensus commit successfully?

crates/sui-core/src/crash_recovery.rs
alex-mysten: nvm answered below

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2760,6 +2795,29 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+                    if let Some(tx) = parsed.transaction.kind.as_user_transaction() {
+                        let digest = tx.digest();
+                        if self.crashed_transactions.contains(digest) {
+                            assert_reachable!(
+                                "crash-recovery: dropping previously-crashing transaction"
+                            );
+                            warn!(
+                                ?digest,
+                                tx = ?bcs::to_bytes(&parsed.transaction).unwrap_or_default(),
+                                "Dropping transaction that caused a crash in a previous run"
```
alex-mysten: What happens to descendants of a dropped tx?

crates/sui-benchmark/tests/simtest.rs
```diff
@@ -511,7 +511,24 @@ mod test {
         // quorum, empirically ~6% per-validator skip rate produces ~5% DKG failure per epoch.
         register_fail_point_if("rb-dkg", || thread_rng().gen_bool(0.06));
 
-        test_simulated_load(test_cluster, 120).await;
+        // crash-with-tx-logging is deferred to pre_load_setup so it doesn't fire during
+        // workload initialization. Setup transactions that crash and get logged would
+        // otherwise stall initialization because the init code retries the same digest.
+        const CRASH_PROB: f32 = 0.002;
+        test_simulated_load_with_test_config(
```
alex-mysten, line 517: do we / should we have a test that exercises multiple subsequent crashes?
or also a crash that impacts one validator?

crates/sui-core/src/crash_recovery.rs
```diff
@@ -0,0 +1,497 @@
...
+/// This chains onto whatever panic hook is already installed (typically the tracing hook set up by
+/// `telemetry-subscribers`), so existing behaviour is preserved.
+///
+/// `authority_name` is this node's public key identity; it is used as the key in
+/// `NODE_DB_PATHS` and to gate the hook so that each validator only claims panics that
+/// originated in its own execution context.
+///
+/// `db_path` is the node's base database directory; the log file is written at
+/// `{db_path}/panic-tx.log`.
+pub fn install_panic_hook(authority_name: AuthorityName, db_path: PathBuf) {
```
alex-mysten, line 280: Presume we will alert whenever we hit this path?

PR #26933, opened 2026-06-09
[feat] Add sui-execution-backtest crate

tzakian:
## Description 

This creates an indexing pipeline that allows for rapid transaction backtesting of the execution layer. Can backtest at ~175 checkpoints/s. 

## Test plan 

Local testing/usage.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-execution-backtest/README.md
```diff
@@ -0,0 +1,240 @@
...
+  --remote-store-url https://checkpoints.mainnet.sui.io \
+  --fullnode-url https://mysten-rpc.mainnet.sui.io:443 \
+  --start-epoch 1152 --end-epoch 1152 --status all \
+  --cache ./.package-cache \
+  --store postgres --database-url postgres://localhost/backtest \
+  --task my-linkage-change
+```
+
+- **Checkpoint source.** Prefer a remote object store (`--remote-store-url
+  https://checkpoints.<network>.sui.io`) — the fast archival path. A fullnode for epoch + package
```
amnn: The HTTP-based remote store now only contains 30 days retention. The real best option is to access the GCS bucket directly -- More info [here](https://docs.sui.io/develop/accessing-data/custom-indexer/indexer-data-integration#remote-store). Even better if you can do this colocated with the GCS bucket as then you won't pay the network or $ cost for egress.

crates/sui-execution-backtest/src/ndjson_store.rs
```diff
@@ -0,0 +1,315 @@
...
+    reader_lo: u64,
+    pruner_timestamp: u64,
+    pruner_hi: u64,
+    chain_id: Option<[u8; 32]>,
+}
+
+/// Appends rows to a newline-delimited JSON file; watermarks live in memory.
+#[derive(Clone)]
+pub struct NdjsonStore {
+    writer: Arc<Mutex<BufWriter<File>>>,
```
amnn, line 49: I would be tempted to implement this by having a tokio task that's responsible for writing to the file, that draws from a channel, and having connections take a clone of that channel, but I'm guessing this is not a bottleneck for you, so all good.

crates/sui-execution-backtest/src/ndjson_store.rs
```diff
@@ -0,0 +1,315 @@
...
+        entry.epoch_hi_inclusive = watermark.epoch_hi_inclusive;
+        entry.checkpoint_hi_inclusive = Some(watermark.checkpoint_hi_inclusive);
+        entry.tx_hi = watermark.tx_hi;
+        entry.timestamp_ms_hi_inclusive = watermark.timestamp_ms_hi_inclusive;
+        Ok(true)
+    }
+}
+
+#[async_trait]
+impl ConcurrentConnection for NdjsonConnection<'_> {
```
amnn, line 184: You may be able to just leave these `unimplemented!` because you don't use pruning.

crates/sui-execution-backtest/src/main.rs
```diff
@@ -0,0 +1,407 @@
...
+        chain_id,
+        status: args.status,
+        task: row_task,
+        indexer_task,
+        stats_enabled: !args.no_stats,
+        first_checkpoint,
+        last_checkpoint,
+        fanout: args
+            .execute_concurrency
+            .map(|value| ConcurrencyConfig::Fixed { value }),
```
amnn, line 238: Why use a fixed concurrency here?

crates/sui-execution-backtest/src/main.rs
```diff
@@ -0,0 +1,407 @@
...
+        None,
+        IngestionConfig::default(),
+        ingestion_metrics,
+    );
+
+    match args.store {
+        StoreKind::Ndjson => {
+            let output = args
+                .output
+                .context("--output is required for --store ndjson")?;
```
amnn: Clap can enforce this for you (has a `requires` attribute).

crates/sui-execution-backtest/src/main.rs
```diff
@@ -0,0 +1,407 @@
...
+        fanout: plan.fanout,
+        ..Default::default()
+    };
+    indexer.concurrent_pipeline(handler, config).await?;
+
+    let service = indexer.run().await?;
+    service
+        .main()
+        .await
+        .map_err(|e| anyhow::anyhow!("indexer terminated: {e:?}"))?;
```
amnn:
```suggestion
        .context("indexer terminated")?;
```

crates/sui-execution-backtest/src/ingestion.rs
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! A remote-object-store ingestion client whose chain identifier is supplied up front instead of
+//! derived by fetching genesis.
+//!
+//! The framework's store client derives the chain id with `checkpoint(0)` — fetching and decoding
+//! mainnet genesis, a large, slow (tens of seconds), one-time startup cost on the critical path
```
amnn: Fetching the genesis checkpoint takes 10s of seconds...?

crates/sui-execution-backtest/src/grpc.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Batch-fetch objects by id with a chunked multi-get, returning the decoded objects that were
+    /// found (missing ids and per-object errors are silently omitted — used to warm the package
+    /// cache, where a miss just falls through to the lazy single fetch). Retries transient errors
+    /// per chunk like [`Self::fetch_object`].
+    pub async fn fetch_objects(&self, ids: &[ObjectID]) -> Result<Vec<Object>> {
+        // Each result carries the full BCS object, and packages can be ~MBs, so keep chunks well
+        // under the 128 MiB gRPC message cap.
+        const CHUNK: usize = 50;
+        let mut out = Vec::new();
+        for chunk in ids.chunks(CHUNK) {
```
amnn: You could run these concurrently?

crates/sui-execution-backtest/src/grpc.rs
```diff
@@ -0,0 +1,201 @@
...
+                for id in chunk {
+                    request
+                        .requests_mut()
+                        .push(rpc::GetObjectRequest::new(&id.into_bytes().into()));
+                }
+                match ledger.batch_get_objects(request).await {
+                    Ok(response) => break response.into_inner(),
+                    Err(status) if attempt < MAX_RETRIES && is_transient(&status) => {
+                        let seed = chunk.first().map_or(0, |id| id.into_bytes()[0]);
+                        tokio::time::sleep(backoff_delay(attempt, seed)).await;
```
amnn: And use the `backoff` crate for this?

amnn (approved):
Apologies for the long delay on reviewing! I think this looks good.

I do think you can simplify the package caching pieces a bit so there isn't such a difference between prefetching and incidental fetches, and you can batch fetches between transactions that are executing concurrently, but this is just a small potential win.

crates/sui-execution-backtest/README.md
tzakian: 100%! this is going to be my next thing I work on getting go when I have the cycles.

crates/sui-execution-backtest/src/ndjson_store.rs
tzakian: Yea this hasn't shown up as a bottleneck (all network bound). I will add a comment though -- especially if the info that is being written out is quite large or a lot (which hopefully it isn't)...it might be useful to add (or breadcrumbs for someones LLM to make an ad-hoc local change).

crates/sui-execution-backtest/src/ingestion.rs
tzakian: Yea, it takes a while, at least from local benchmarking, there was a decently noticeable difference in perf with this..

crates/sui-execution-backtest/src/main.rs
tzakian: We only use it if supplied which seems reasonable to me. Happy to iterate on it and/or we can remove it later if we want though.

crates/sui-execution-backtest/src/ingestion.rs
tzakian: Without `FixedChainId`, when I benchmarked things it was ~220 cp/s with vs ~55–60 cp/s without, and this was consistent across rounds (for a 40k run: 155s vs 700s). 

From looking at the logs, the `chain_id` slow-op recurs for ~5 min in the logs, so it seems to be recurrent as well possibly? At this point I got to the limits of my knowledge and this seemed reasonable enough to do, but just in case this is useful to you :)

amnn: I think there's multiple interesting things going on here:

1. Why is a seemingly simple operation slow (triggering the slow op warning).
2. Why is it happening over and over? (It's supposed to be cached).
3. Why is it affecting your indexer and not others?

I have some theories on this -- I think (1) and (3) are because you are running outside of GCP, so the cost is exaggerated and (2) is because of a bug/misunderstanding on how `OnceCell` cloning works. I'll dig in!

amnn: Yep, that was probably it! Fixed here: #27390, so you should be able to remove your caching layer now.

PR #26936, opened 2026-06-10
[move][move-decompiler] Revise loop refinement to handle multi-exit loops [0/8]

cgswords:
## Description

Reworks how the structurer handles loops with multiple exit targets. Previously, multi-succ loops would surface as raw gotos because the structurer couldn't express them with a single labeled break. This PR introduces a dispatch-table mechanism: at every owned-successor exit site in the loop body, the structurer rewrites the Jump as `Assign(__dispatch_<N>, k); Break(loop_head)`, and emits a `match (__dispatch_<N>) { 0 => arm_0, ..., K-1 => arm_{K-1} }` after the loop. Each arm carries the cascade tail starting at its owned succ (the succ's body, then `cascade_next(succ)`'s body, etc.) so the downstream refinements can recover the original branchy shape. Also includes preparatory work that landed in the same series: embedding multi-pred sink arms at loop heads, a convergence check on the orphan-hoist that refuses to fuse loop exits into Code blocks, recovering loop exits and hoisting orphans in CFG order, plus mechanical clean-ups (rename `Structured::Match` → `Structured::SelectorMatch`, add `pub type DispatchTag = u32` alias threaded through every dispatch site, density debug-assert on the synthesized tag range, drop the unused `_loop_header` parameter, rename `succ_node` / `loop_succ` / `succ_for_insert_breaks` → `loop_successor`, switch `cascade_next`'s subtree from `HashSet` to `Vec`) and a `convergence_ok_reject.stt` regression fixture pinning the orphan-hoist's rejection behavior.

## Test Plan

`cargo nextest run` from `external-crates/move/crates/move-decompiler` (108 tests pass after snapshot regen). `cargo move-clippy` clean.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-decompiler/tests/bytecode/almm_bin_tree.mv.snap
```diff
@@ -0,0 +1,204 @@
...
+
+public fun add(l0: &mut BinTree, l1: u32): bool {
+    assert!(l1 >> 24u8 == 0u32, C2);
+    let l5 = l1 >> 8u8as u256;
+    let l6 = if (vec_map::contains(&l0.level2, &l5)) {
+        *(vec_map::get(&l0.level2, &l5))
+    } else {
+        0u256
+    };
+    let l8 = l6 | 1u256 << l1 & C1as u32as u8;
```
tzakian: formatting/PP seems to be dropping spacing around `as`.

external-crates/move/crates/move-decompiler/src/structuring/mod.rs
```diff
@@ -369,25 +599,142 @@ fn structure_acyclic(
...
+/// `target`. Returns `None` for any tail that isn't a `Jump` and for multi-tail forms
+/// (IfElse/Switch) whose tails disagree on the target.
+fn unique_tail_jump_target(s: &D::Structured) -> Option<NodeIndex> {
+    use D::Structured as DS;
+    match s {
+        DS::Jump(_, target) => Some(*target),
+        DS::Seq(items) => items.last().and_then(unique_tail_jump_target),
+        DS::IfElse(_, conseq, alt) => {
+            let c = unique_tail_jump_target(conseq)?;
+            let a = unique_tail_jump_target(alt.as_ref().as_ref()?)?;
```
tzakian: `as_ref().as_ref()`? :psyduck:

tzakian (approved):
👍🏻

PR #26937, opened 2026-06-10
[move][move-decompiler] Add reaching-condition acyclic structurer [1/8]

cgswords:
## Description

Replaces the dom-tree structurer with the No-More-Gotos structurer at every level: the dominator-tree post-order walk dispatches loops to `structure_loop` and acyclic SESE slices to `acyclic::structure_region`. The acyclic structurer now uses a reaching-condition structurer (a la NMG IV-B) to project the region into a refined, nested control flow structure via Formula classification, reduction, and optimization.

We also improve performance significantly, both in terms of wall-clock time and eliminating most of the remaining gotos by folding `&&` and `||` short-circuit tests back up.

We also flag a pathological case in our test suite (which can be accounted for in future work, but which the Move compiler should never generate today).

## Test plan

`cargo insta test` from `external-crates/move/crates/move-decompiler` (`--release` is no longer necessary via `cargo nextest run`).

`output/` dir is also removed, since it was an artifact and should not have been added.

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


tzakian (approved):
Had a look over and everything looks sensible to me.

PR #26950, opened 2026-06-11
[sui-adapter][easy] Skip collecting runtime telemetry for dry-run/dev-inspect

tzakian:
## Description 

A new runtime is spun up for dry-run/dev-inspect calls, so any VM telemetry from those calls shouldn't be reported as it will have pollute metrics (esp. cache metrics).

Skip reporting for these cases by wrapping it in an `if !Mode::TRACK_EXECUTION` which is set only for these two. Diff looks larger than it is because of the formatting change. 

## Test plan 

👀 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-adapter/src/execution_engine.rs
+++ b/sui-execution/latest/sui-adapter/src/execution_engine.rs
@@ -460,62 +460,66 @@ mod checked {
             *epoch_id,
         );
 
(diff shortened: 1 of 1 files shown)


dariorussi (approved):
LGTM

PR #26954, opened 2026-06-12
Remove old checkpoint splitting path

alex-mysten:
## Description

Checkpoint splitting in the consensus handler (`split_checkpoints_in_consensus_handler`) is enabled on all networks including mainnet (since protocol v124). This removes the old, disabled code path and asserts the flag is enabled.

## Test plan

- `cargo nextest run -p sui-core`: 704/704 pass
- `cargo simtest -p sui-core` (checkpoint/quarantine/consensus filters): pass
- `cargo simtest -p sui-e2e-tests`: checkpoint (22), reconfiguration (8), protocol_version (7), coin_registry, display_registry — all pass
- `cargo check --workspace --all-targets`, `cargo xclippy -D warnings`, `cargo fmt --check`: clean

### DIFF
--- a/crates/sui-core/src/checkpoints/mod.rs
+++ b/crates/sui-core/src/checkpoints/mod.rs
@@ -19,7 +19,6 @@ pub use crate::checkpoints::metrics::CheckpointMetrics;
 use crate::consensus_manager::ReplayWaiter;
 use crate::execution_cache::TransactionCacheRead;
 
-use crate::execution_scheduler::funds_withdraw_scheduler::FundsSettlement;
 use crate::global_state_hasher::GlobalStateHasher;
 use crate::stake_aggregator::{InsertResult, MultiStakeAggregator};
 use consensus_core::CommitRef;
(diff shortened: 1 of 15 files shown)


crates/sui-core/src/jsonrpc_index.rs
```diff
@@ -2301,8 +2300,13 @@ mod tests {
         // and verified from both db and cache.
         // This tests make sure we are invalidating entries in the cache and always reading latest
         // balance.
-        let index_store =
-            IndexStore::new_without_init(temp_dir(), &Registry::default(), Some(128), false);
+        let dir = tempfile::tempdir().unwrap();
+        let index_store = IndexStore::new_without_init(
```
alex-mysten, line 2304: fix for concurrency issue between the two tests

crates/sui-core/src/checkpoints/mod.rs
```diff
@@ -132,14 +120,12 @@ pub struct CheckpointRoots {
     pub height: CheckpointHeight,
 }
 
-/// PendingCheckpointV2 is merged and split in ConsensusHandler
+/// PendingCheckpoint is merged and split in ConsensusHandler
 /// instead of CheckpointBuilder. It contains 1 or more
```
aschran: "instead of CheckpointBuilder" is no longer relevant right?

aschran (approved):

PR #26955, opened 2026-06-12
Bound accumulator amounts to representable limits before charging gas

mystenmark:
## Summary

- Reject transactions whose per-key accumulator totals are not representable, before charging gas. SUI balance keys are bounded by the total SUI supply; other balances are bounded by `u64::MAX`.
- Also bound the cross-key total SUI withdrawn (gross Split) by the total supply, since withdrawn SUI can be recombined outside the accumulator (e.g. merged into the PTB gas coin).
- Accumulate the SUI-conservation input/output totals in `u128` so intermediate sums stay exact regardless of individual event amounts.
- Add e2e regression tests covering per-key SUI/custom-coin merges, single-withdrawal conservation, the gas-refund path, and gas-coin recombination via `MergeCoins`.

The new guard turns a would-be arithmetic failure into a deterministic `CoinBalanceOverflow` abort. It is ungated and applies uniformly across protocol versions.

### DIFF
--- a/crates/sui-e2e-tests/tests/address_balance_tests.rs
+++ b/crates/sui-e2e-tests/tests/address_balance_tests.rs
@@ -2157,6 +2157,445 @@ async fn test_multiple_deposits_merged_in_effects() {
     test_env.trigger_reconfiguration().await;
 }
 
+/// A single transaction produces two Merge accumulator events to the same `(sender, Balance<SUI>)`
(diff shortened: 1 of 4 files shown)


dariorussi (approved):
If everybody agrees, I am fine getting this in and talk about how to do it otherwise.

tzakian (approved):

PR #26962, opened 2026-06-15
ci(sccache): same-repo PRs assume the read-only OIDC role (Phase 4, option A)

ebmifa:
## Summary

Implements **Phase 4, decision A** (recorded in #26960). Now that `sui-sccache-ro-github` is provisioned, each `SCCACHE_ROLE` gate in `rust.yml` / `external.yml` / `bridge.yml` resolves the **read-only** role for same-repo PRs, and the static-key inputs are removed from these workflows entirely.

Role selection after this PR:

| Context | Role | Cache |
|---|---|---|
| trusted push / dispatch-from-main (no override) | `sui-sccache-rw-github` | read/write |
| **same-repo `pull_request`** | **`sui-sccache-ro-github`** | **read-only (new)** |
| fork `pull_request` | — (empty) | local |
| override dispatch (`sui_repo_ref` set) | — (empty) | local |

The RO role grants `s3:GetObject` + `s3:ListBucket` only — **no `PutObject`** — so a PR can read the shared cache but cannot poison it.

## The fork nuance (why same-repo only)

GitHub caps fork-PR `id-token` to `none`, so a fork cannot mint the OIDC token — and `setup-sccache`'s `Configure AWS credentials (OIDC)` step has no `continue-on-error`, so setting `role-to-assume` on a fork would **hard-fail the job**. The gate therefore resolves the RO ARN only when `github.event.pull_request.head.repo.fork == false`; fork PRs stay empty → local cache, exactly as today.

**This fork skip is a fail-safe accommodation, not a security boundary.** The RO role still trusts `pull_request`, so if GitHub settings ever permit fork `id-token`, a fork assuming the RO role remains acceptable under Decision A. We are not relying on fork-local as a durable guarantee (that would require withholding `id-token: write` from fork-reachable jobs — the option-C complexity we declined).

**Static-key inputs are removed from these three workflows** (not deferred to Phase 7). Keeping them would have been an active hole, not dead weight: `setup-sccache` is a **local** action loaded from the checked-out workspace, so a same-repo PR could edit it to consume the RW static keys directly — defeating the RO boundary — and an override dispatch (`SCCACHE_ROLE` empty) would have fed RW static keys to the untrusted override build. With them gone, forks and override dispatches get a local cache (no creds). `release.yml` still passes static keys (no `pull_request` trigger → not PR-exposed); the repo secrets are deleted in Phase 7.

## Test plan

- [x] Expression traced for every event/ref: trusted push/dispatch → RW; same-repo PR → RO; fork PR → empty; override dispatch → empty. The `github.event_name == 'pull_request'` guard short-circuits before the `fork == false` comparison on non-PR events (avoids GitHub's null→0 coercion making `fork == false` spuriously true).
- [x] `actionlint`: identical finding count to main on all three files (11/0/5, all pre-existing shellcheck/isSolidity); zero new.
- [ ] **Post-merge:** confirm a same-repo PR's sccache step shows `Assuming role with OIDC` for `sui-sccache-ro-github`; confirm a trusted push still assumes the RW role; confirm a fork PR (or simulated empty role) degrades to local cache without failing.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/.github/workflows/bridge.yml
+++ b/.github/workflows/bridge.yml
@@ -101,11 +101,10 @@ jobs:
           ref: ${{ github.event.inputs.sui_repo_ref || github.ref }}
       - uses: ./.github/actions/setup-sccache
         with:
-          # Dual-pass: OIDC on trusted refs (SCCACHE_ROLE non-empty); static-key
-          # fallback on same-repo PRs / untrusted refs until Phase 4 sets PR cache.
+          # OIDC only: RW on trusted refs, RO on same-repo PRs, empty (→ local
+          # cache) on forks / override dispatches. No static-key fallback — those
+          # were removed so a PR-controlled local action cannot reach the RW keys.
           role-to-assume: ${{ env.SCCACHE_ROLE }}
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           key-prefix: ubuntu-ghcloud
       - run: rustup component add clippy
       - name: cargo clippy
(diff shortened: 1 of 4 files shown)


.github/workflows/bridge.yml
```diff
@@ -129,11 +132,10 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # pin@v6.0.2
       - uses: ./.github/actions/setup-sccache
         with:
-          # Dual-pass: OIDC on trusted refs (SCCACHE_ROLE non-empty); static-key
-          # fallback on same-repo PRs / untrusted refs until Phase 4 sets PR cache.
+          # OIDC only: RW on trusted refs, RO on same-repo PRs, empty (→ local
+          # cache) on forks / override dispatches. No static-key fallback — those
+          # were removed so a PR-controlled local action cannot reach the RW keys.
```
johnjmartin: do we need this comment duplicated?

johnjmartin (approved):

.github/workflows/bridge.yml
ebmifa: Good catch — removed both copies in 111496580e. The role-selection rationale already lives on the workflow-level `SCCACHE_ROLE` env, so the per-call-site comments were redundant.

PR #26966, opened 2026-06-15
[sui-fork] Fix child object reads to support bounded versions

stefan-mysten:
## Description 

This PR adds support for reading child object bounded versions. Without this, a bunch of transactions do not work correctly and the backend complains with an invariant violation, because we're passing the latest object rather than the correct bounded version of that child object.

## Test plan 

Several tests added that cover loading the object at the right version.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-fork/src/filesystem.rs
+++ b/crates/sui-fork/src/filesystem.rs
@@ -141,6 +141,16 @@ struct ObjectLatestMetadata {
     state: ObjectLatestState,
 }
 
+/// Result of a bounded local object lookup.
+pub(crate) enum BoundedObjectLookup {
+    /// A live object version at or below the requested bound was found locally.
+    Hit(Object),
+    /// A local tombstone at or below the requested bound proves the object is absent.
+    NegativeHit,
+    /// Local storage has no authoritative answer for the requested bound.
+    Miss,
+}
+
 impl ObjectLatestMetadata {
     /// Construct numeric-only latest metadata for a live object version.
     fn live(version: u64) -> Self {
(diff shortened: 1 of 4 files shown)


crates/sui-fork/src/store.rs
```diff
@@ -342,6 +343,36 @@ impl DataStore {
         Ok(object)
     }
 
+    /// Get the latest object version at or below the given root version.
+    fn get_object_lt_or_eq_version(
```
jordanjennings-mysten, line 347: does something like `get_object_at_or_before_version` make more sense? also curious about the trade off between a query like:
```
  enum ObjectVersionQuery {
      Latest,
      Exact(SequenceNumber),
      AtOrBefore(SequenceNumber),
      AtCheckpoint(u64),
  }
```

Im guessing you don't need too much flexibility here though.

crates/sui-fork/src/filesystem.rs
```diff
@@ -445,6 +455,40 @@ impl FilesystemStore {
...
+        &self,
+        object_id: &ObjectID,
+        version_bound: u64,
+    ) -> anyhow::Result<BoundedObjectLookup> {
+        let object_dir = self.objects_dir().join(object_id.to_string());
+        if !object_dir.exists() {
+            return Ok(BoundedObjectLookup::Miss);
+        }
+
+        if let Some(latest) = self.read_object_latest_metadata_if_exists(&object_dir)?
```
jordanjennings-mysten, line 469: not sure if I'd do anything in particular here but it does feel strange that this can return `None` since wouldnt a lack of latest file mean something was corrupted? Maybe better overall for UX to not be failing fast on "corrupted" object folders.

jordanjennings-mysten (approved):

crates/sui-fork/src/store.rs
stefan-mysten: We have a similar query to that, here
https://github.com/MystenLabs/sui/blob/c28e47def5e8523ea1a81cbd96a57ad458d8c077/crates/sui-fork/src/lib.rs?plain=1#L74-L85

amnn (approved):
Seems reasonable!

I do get the feeling that even storing the point lookup data in RocksDB may be beneficial to support queries like this, but it's not a blocker.

PR #26973, opened 2026-06-16
Reliably report consistently-failing state-sync peers

aschran:
## Description

`state_sync_resilience_tests::test_disconnect_consistently_failing_peers` is flaky: under sustained latency the syncing node sometimes never reports any peer as failing. State-sync only scored a peer while actively fetching checkpoints from it, so a node that had already caught up scored nothing; and the "continuously failing" timer was reset by any single successful response, so a peer that failed most-but-not-all requests (e.g. flapping connections) never crossed the disconnect threshold.

Failed liveness/discovery queries now count as failures against the peer's score (distinguishing a genuine request failure from a peer that simply has nothing newer to report), and the "continuously failing" duration is derived from the windowed failure rate rather than reset on every success. A consistently-unreachable peer is now reported whether or not sync traffic is flowing.

## Test plan

The `failing_since` unit test in `sui-network` is updated for the rate-based semantics. Reproduced the flake locally with a fast-checkpoint-interval harness that fails 51/400 seeds without this change and passes 400/400 with it.

---

## Release notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): A node now reports and disconnects state-sync peers that are consistently unreachable even when it is not actively syncing from them; previously this could fail to happen. No operator action required.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-network/src/state_sync/mod.rs
+++ b/crates/sui-network/src/state_sync/mod.rs
@@ -181,7 +181,11 @@ impl PeerScore {
         while self.successes.len() > Self::MAX_SAMPLES {
             self.successes.pop_front();
         }
-        self.failing_since = None;
+        // Note: a single success does NOT clear `failing_since`. Whether the peer is
+        // "consistently failing" is driven by the windowed failure RATE (see
+        // `update_failing_state`), so an occasional success while the rate stays high (e.g.
+        // a brief reconnect during connection churn) must not keep resetting the clock and
+        // prevent the peer from ever being reported.
     }
 
     pub(super) fn record_failure(&mut self) {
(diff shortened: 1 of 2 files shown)


crates/sui-network/src/state_sync/mod.rs
```diff
@@ -181,7 +181,11 @@ impl PeerScore {
         while self.successes.len() > Self::MAX_SAMPLES {
             self.successes.pop_front();
         }
-        self.failing_since = None;
+        // Note: a single success does NOT clear `failing_since`. Whether the peer is
+        // "consistently failing" is driven by the windowed failure RATE (see
+        // `update_failing_state`), so an occasional success while the rate stays high (e.g.
+        // a brief reconnect during connection churn) must not keep resetting the clock and
+        // prevent the peer from ever being reported.
```
alex-mysten, line 188: Comments referencing previous iterations of the code are potentially confusing

crates/sui-network/src/state_sync/mod.rs
```diff
@@ -1268,14 +1290,36 @@ async fn query_peers_for_their_latest_checkpoint(
...
+                            .unwrap()
+                            .record_success(peer_id, size, elapsed);
+                        peer_heights
+                            .write()
+                            .unwrap()
+                            .update_peer_info(peer_id, highest_checkpoint.clone(), low_watermark)
+                            .then_some(highest_checkpoint)
+                    }
+                    // Peer responded but had nothing useful; neutral, do not score.
+                    Ok(None) => None,
```
alex-mysten, line 1308: If a peer responded, even without useful information, doesn't this mean that it's healthy?

alex-mysten (approved):

PR #26975, opened 2026-06-16
sui-proxy: add on-chain hashi committee resolver

tharbert:
## Description 

Hashi requires every Sui validator to run a hashi-server instance (~100+ third-party operators at launch). This adds a third peer-discovery path to sui-proxy alongside sui_nodes and bridge_nodes so those operators can push Prometheus metrics to Mimir using the same mTLS protocol sui-node and sui-bridge already use.

Authentication piggybacks on hashi's existing on-chain operator registry: `MemberInfo.tls_public_key` in `hashi::committee_set::CommitteeSet`. The resolver fetches the active Committee (and pending Committee during reconfig) and pulls each member's live tls_public_key from the `members` Bag, so on-chain key rotations are picked up within one poll cycle without operator-side coordination.

Config lives on `dynamic-peers.hashi-object-id` — hashi is conceptually another dynamic-discovery source against the same Sui chain we're already polling. Bridge already works this way (no separate config block; uses `dynamic-peers.url` + `dynamic-peers.interval` with a hardcoded RPC method). Hashi only needs the object id because its discovery isn't a fixed RPC method.

Labels: hashi peers get `host=hashi-<sui-validator-name>` so a single Mimir tenant can host sui-node and hashi-server pushes without metric-name collision (the `host` label disambiguates shared metrics like `process_cpu_seconds_total`).

Tolerates the pre-genesis state (empty `committees` Bag, epoch=0) by producing an empty allowlist rather than erroring — sui-proxy can be deployed before hashi runs its first reconfig.

Telemetry: new `hashi_proxy_observed_committee_epoch` and `hashi_proxy_allowed_members` gauges, both labeled by `hashi_object_id`, make resolver health observable from the proxy's self-metrics endpoint.

## Test plan 

Verified end-to-end against private-testnet with hashi pushing real metrics; resolver unit tests cover the empty/invalid tls_public_key filter, dual-format Move Option<u64> parsing, Bag id extraction, and name-fallback when validator metadata is missing.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-proxy/src/peers.rs
```diff
@@ -537,6 +635,399 @@ fn append_path_segment(mut url: Url, segment: &str) -> Option<Url> {
...
+
+async fn get_hashi_member_info(
+    rpc_url: &str,
+    members_bag_id: &str,
+    validator_address: SuiAddress,
+) -> Result<ResolvedHashiMember> {
+    let rpc_method = "suix_getDynamicFieldObject";
+    let _timer = JSON_RPC_DURATION
+        .with_label_values(&[rpc_method])
+        .start_timer();
```
bmwill, line 999: semi-unrelated to this pr, but who can we have to look at migrating sui-proxy off of jsonrpc?

crates/sui-proxy/src/peers.rs
```diff
@@ -537,6 +635,399 @@ fn append_path_segment(mut url: Url, segment: &str) -> Option<Url> {
...
+    if let Some(s) = value.as_str() {
+        return Ok(Some(s.parse().context("parsing Option<u64> bare value")?));
+    }
+    bail!("unrecognized Option<u64> JSON shape: {value}")
+}
+
+/// JSON-RPC helper that returns the raw response `Value` so callers can
+/// pointer-walk into nested fields without imposing a fixed schema (the Move
+/// display layout differs per type).
+async fn sui_rpc_call(rpc_url: &str, method: &str, params: Value) -> Result<Value> {
```
bmwill: in order to avoid needing to migrate this could we look at using grpc instead?

crates/sui-proxy/src/peers.rs
tharbert: yes we can, i will take that on in a subsequent change soon / prior to the deprecation date

crates/sui-proxy/src/peers.rs
tharbert: yes we can, ill make the change now and re-test

crates/sui-proxy/src/peers.rs
```diff
@@ -635,30 +639,307 @@ fn append_path_segment(mut url: Url, segment: &str) -> Option<Url> {
     Some(url)
 }
 
-// On-chain shape we depend on (see hashi/packages/hashi/sources/core/committee/
-// committee_set.move and hashi.move):
```
bmwill: It would also be fine to request the JSON form of the objects so that you can continue to do the path-based traversal into the JSON to pull out the values you want, then you'll be slightly less susceptible to things changing, or you can maybe depend directly on `hashi-types` to just pull in the types directly and not need to redefine them here again.

tharbert: moved to requesting json and just pulling out the fields we need

https://github.com/MystenLabs/sui/pull/26975/changes/1b5c05a6a25df150d65fd8cb22614f5171a84a0e

repeated end to end testing

bmwill (approved):

PR #26978, opened 2026-06-16
rpc-store: introduce sui-rpc-store crate

bmwill:
A storage backend for `sui-rpc-api`, built on `sui-consistent-store`, that hosts the column families backing every read the RPC service performs. It consolidates raw chain data previously served by the validator's perpetual / checkpoint / committee stores together with the indexes previously served by `sui-core::rpc_index` and `sui-indexer-alt-consistent-store`.

The crate is organized into:

- `schema` — the column-family layout. Each CF declares its on-disk key and value types plus per-CF RocksDB options. Covers raw chain data (objects, transactions, effects, events, checkpoints) and the indexes (owner, dynamic field, coin balance, package versions, epochs, ledger history, transaction / event bitmaps).

- `indexer` — sequential `sui-indexer-alt-framework` pipelines that populate the schema from observed checkpoints, the `Indexer` orchestrator that wires them against a shared `Synchronizer`, plus the `pruner` and `restore` drivers for history pruning and bulk-load from a formal snapshot.

- `reader` — `RpcStoreReader`, which exposes the schema through the trait stack `sui-rpc-api` consumes (`ObjectStore`, `ReadStore`, `ChildObjectResolver`, `RpcStateReader`, and `RpcIndexes`). Generic over a `Reader` so one type serves both tip reads and reads bound to a captured snapshot.

- `proto` — bespoke protobuf value messages defined under `proto/sui/rpc_store/`, generated via `codegen.rs`.

- `config` — `ServiceConfig` and the layered pipeline / committer / pruner / restore configuration.

`start_indexer` is the standalone-binary entry point; the embedded fullnode path constructs `Indexer::from_store` directly with its own ingestion and streaming clients.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


nickvikeras (approved):

crates/sui-rpc-store/src/indexer/checkpoint_broadcast.rs
```diff
@@ -0,0 +1,198 @@
...
+        Ok(1)
+    }
+}
+
+/// Seed the broadcast pipeline's resume watermark to `checkpoint` when
+/// it has none yet, so it starts at the live tip rather than genesis.
+///
+/// The broadcast pipeline is registered like any other (it persists a
+/// watermark and resumes from it), but it is not covered by the formal
+/// snapshot or [`floor_unrestored_pipelines`](crate::floor_unrestored_pipelines),
```
amnn, line 113: Out of interest, why not cover this by `floor_unrestored_pipelines`? (Maybe will be clear when I reach that function).

crates/sui-rpc-store/src/indexer/epochs.rs
```diff
@@ -0,0 +1,140 @@
...
+        else {
+            return Ok(vec![]);
+        };
+
+        let mut rows = Vec::with_capacity(2);
+
+        // Start record for the new epoch. `EpochInfo` from
+        // `epoch_info()` populates every field except the end-of-
+        // epoch ones; pull them out into our `Row::Start` shape.
+        let system_state_bcs = epoch_info
```
amnn, line 70: One of the reasons we split epoch information across multiple tables in GraphQL was that a full epoch record is quite large (lots of fields, system state alone is large), and it's rare to need all of it in one go.

That may still apply here? In fact, it may even be worth splitting it up further than what we do today in postgres (just start and end) so that system state is in its own table as well.

GraphQL currently takes advantage of this data organisation to avoid overfetching to resolve its fields. If fullnodes/RPC node also do the same thing, it can pass the benefit on.

crates/sui-rpc-store/src/indexer/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+        // Load the persisted pruning watermarks into the
+        // in-memory bitmap floor so the bitmap CFs' compaction
+        // filters resume against the on-disk watermark instead of
+        // the process-default zero (which would prune nothing).
+        store
+            .schema()
+            .refresh_pruning_atomics()
+            .context("Failed to refresh pruning watermarks")?;
+
+        let stride = NonZero::new(consistency_config.stride)
```
amnn, line 312: Note that jury's still out on whether `stride` justifies its own complexity.

In theory a higher stride will remove some bottlenecks so should improve max throughput, but we've never hit a practical limit for scale, and if we did, maybe horizontal scaling via sharding is a better way to deal with it.

crates/sui-rpc-store/src/indexer/object_by_owner.rs
```diff
@@ -0,0 +1,147 @@
...
+use crate::indexer::checkpoint_output_objects;
+use crate::schema::keys::U64Varint;
+use crate::schema::object_by_owner;
+
+/// Pipeline marker for `object_by_owner`.
+pub struct ObjectByOwner;
+
+pub enum Row {
+    Delete(object_by_owner::Key),
+    Put(object_by_owner::Key, U64Varint),
```
amnn, line 40: Today, GraphQL treats `ObjectRef` as the identity of an object (i.e. including digest), so if we want to represent an `Object` we need to produce at least its ID, version, and digest.

To make use of this index efficiently (and avoid over-fetching content to calculate a digest), we'll likely need to drop the digest from the key.

crates/sui-rpc-store/src/indexer/objects.rs
```diff
@@ -0,0 +1,262 @@
...
+        for tx in &checkpoint.transactions {
+            for object in tx.output_objects(&checkpoint.object_set) {
+                rows.push(Row {
+                    id: object.id(),
+                    version: object.version(),
+                    value: objects::store(object),
+                });
+            }
+            let lamport_version = tx.effects.lamport_version();
+            for (id, _) in tx
```
amnn, line 67: nit: it would be nicer to use the `object_changes` API for this.

crates/sui-rpc-store/src/indexer/package_versions.rs
```diff
@@ -0,0 +1,154 @@
...
+use crate::RpcStoreSchema;
+use crate::indexer::Schema;
+use crate::indexer::Store;
+use crate::indexer::checkpoint_output_objects;
+use crate::schema::package_versions;
+
+/// Pipeline marker for `package_versions`.
+pub struct PackageVersions;
+
+pub struct Row {
```
amnn, line 35: Couple of things missing here compared to `kv_packages`:

- `cp_sequence_number` (used for paginating packages over time): this is partially covered by the package publish dimension of the transaction bitmap (this will not surface packages that were published before the formal snapshot the node was restored from, but maybe that's an OK trade-off).
- `is_system_package` (used in GraphQL to implement `Epoch.systemPackages` and for evicting the system packages from the package resolver's cache on epoch boundaries). The alternative is to bake in a list of the system packages into the RPC binary, but I'd recommend avoiding that if you can, because it relies on operators being prompt about keeping their deployments up-to-date with releases, otherwise the cache eviction logic can fail to evict everything it should, causing problems.  This is calculated by checking whether the sender for the transaction that published/upgraded the package is `0x0` -- a bit unclear how you would restore this from a formal snapshot though...

crates/sui-rpc-store/src/indexer/pruner.rs
```diff
@@ -0,0 +1,1261 @@
...
+        db.compact_range_cf(event_bitmap::NAME, None, None)
+            .context("Compacting event_bitmap after prune")?;
+    }
+
+    Ok(())
+}
+
+/// Prune one chunk of checkpoints `[cursor.checkpoint_lo,
+/// chunk_ckpt_hi)` and their transactions, returning the new floor.
+fn prune_chunk(
```
amnn, line 286: Just flagging that it's an important property that batches commiitted by `prune_chunk` respect checkpoint boundaries, because this service does not coordinate with the indexer's synchronizer in any other way to make sure that snapshots don't see a ragged trailing edge (by witnessing a half-complete pruning job).

crates/sui-rpc-store/src/reader/layout.rs
```diff
@@ -0,0 +1,237 @@
...
+        let executor = sui_execution::executor(&protocol_config, /* silent */ true)
+            .map_err(StorageError::custom)?;
+
+        let backing = PackageStoreOverObjects {
+            schema: self.schema(),
+        };
+        let overlay_store = OverlayBackingPackageStore::new(overlay, &backing);
+
+        let layout = executor
+            .type_layout_resolver(&protocol_config, Box::new(overlay_store))
```
amnn, line 80: It would be better to use `sui-package-resolver` here -- I believe @tzakian is in the process of fully removing `type_layout_resolver` and the associated traits from the authority codebase.

crates/sui-rpc-store/src/schema/epochs.rs
```diff
@@ -0,0 +1,374 @@
...
+    })
+}
+
+/// Build a partial `StoredEpoch` carrying the end-of-epoch fields.
+/// Indexer pipelines that observe an epoch's final checkpoint
+/// stage this as a merge operand against the epoch's key.
+pub fn end(end_timestamp_ms: u64, end_checkpoint: u64) -> Value {
+    Protobuf(StoredEpoch {
+        end_timestamp_ms: Some(end_timestamp_ms),
+        end_checkpoint: Some(end_checkpoint),
```
amnn, line 73: In postgres we also index commitments from the end-of-epoch info, and data from the `SystemEpochInfoEvent` emitted during epoch change (bunch of gas-related information, and an indication of whether we entered safe mode or not):

https://github.com/MystenLabs/sui/blob/985f7c2781faea7bda15db5914ec426b1abcd2f0/crates/sui-indexer-alt/src/handlers/kv_epoch_ends.rs#L34-L125

crates/sui-rpc-store/src/schema/keys.rs
```diff
@@ -0,0 +1,381 @@
...
+}
+
+/// A `u64` encoded as a protobuf-style varint, suitable for
+/// *value* positions where on-disk size matters more than sort
+/// order. Compared to [`U64Be`]: smaller for typical values (1–5
+/// bytes for anything below `2^35`), slightly larger only near the
+/// `u64::MAX` corner (up to 10 bytes), and never sort-stable —
+/// hence values only.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct U64Varint(pub u64);
```
amnn, line 54: nit: `keys` is probably a misnomer for this module then, given that this  is meant to be for value positions? Maybe "primitives"?

crates/sui-rpc-store/src/schema/object_by_owner.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+/// Prefix encoder for "all address-owned objects of `owner`".
+///
+/// Encodes as `kind_tag(1) || owner(32)` — exactly the first 33
+/// bytes of every `Key` whose `kind` is `AddressOwner(owner)`. The
+/// schema's `Encode` impl matches this layout, so passing this
+/// type to [`DbMap::iter_prefix`](sui_consistent_store::DbMap::iter_prefix)
+/// walks every row owned by the given address.
+pub struct AddressOwnerPrefix(pub SuiAddress);
```
amnn, line 227: Could you use the various variants of `OwnerKind` for this directly, instead of introducing a dedicated type for each variant?

crates/sui-rpc-store/src/schema/object_by_owner.rs
```diff
@@ -0,0 +1,473 @@
...
+
+/// Prefix encoder for "all objects owned by `owner` that match
+/// `type_filter`". Composes [`AddressOwnerPrefix`] with a
+/// [`TypeFilter`](super::type_filter::TypeFilter).
+pub struct AddressOwnerTypePrefix {
+    pub owner: SuiAddress,
+    pub type_filter: super::type_filter::TypeFilter,
+}
+
+impl Encode for AddressOwnerTypePrefix {
```
amnn, line 262: Would it make sense to offer generic impls of `Encode` for tuples of size 2, 3, 4, etc so that key fragments could be combined at usage?

crates/sui-rpc-store/src/schema/mod.rs
amnn, line 1: I think the only gap here relative to postgres is `obj_versions`, which has the following schema:

https://github.com/MystenLabs/sui/blob/985f7c2781faea7bda15db5914ec426b1abcd2f0/crates/sui-indexer-alt-schema/src/schema.rs#L149-L156

We use this to answer the following kinds of query:

- multi-get object by ID and checkpoint bound (latest version of the object at that checkpoint)
- multi-get object by ID and version bound
- paginate versions of an object over time (can be replaced by affected object dimension of transaction bitmaps).

Once you have this, you may not need a "live objects" table because a live object query is just a "multi-get object by ID and checkpoint bound" query where the checkpoint bound is the latest checkpoint.

amnn (commented):
I made it! Tried to flag anything I noticed that would require some work to interface with from GraphQL (split of data for epochs, extra package version info, fields that comprise keys for objects, and the object versions table).

crates/sui-rpc-store/src/indexer/mod.rs
bmwill: Yeah I mean we can always just nix stride for now and then worry about it later if we need to. Atm it is always 1

crates/sui-rpc-store/src/schema/mod.rs
bmwill: Yeah this looked like the main gap when I took a look. I just put up https://github.com/MystenLabs/sui/pull/27055 to introduce a table to do the bounds by checkpoint.

crates/sui-rpc-store/src/indexer/object_by_owner.rs
bmwill: This is an embedded db so we're much more sensitive, imo, to db footprint itself. We're already adding a ton of extra CFs and i don't think it makes sense to include Digest and cost us and additional 32 bytes for every entry, assuming you mean you think the digest should be included in the key, not excluded as your comment says (its already not part of the key)

crates/sui-rpc-store/src/schema/object_by_owner.rs
bmwill: yeah i think i missed cleaning this up

crates/sui-rpc-store/src/indexer/object_by_owner.rs
amnn: I agree -- this is just a note that we need to make the change to GraphQL's key.

crates/sui-rpc-store/src/indexer/package_versions.rs
bmwill: I mean the list of system package IDs does not change that often, and we already have a dep on where this is defined...

bmwill: https://github.com/MystenLabs/sui/pull/27084 should fill the gap that still remains wrt checkpoint based reads

amnn: > we already have a dep on where this is defined...

The main issue was never about taking the code dependency (GraphQL also already takes a dependency on that crate). It was that if we relied on the code dependency, then once validators upgraded to a new version that introduced a new system package RPCs had less than a day to upgrade themselves or risk seeing subtly incorrect behaviour.

When we first took these defensive steps, it was motivated by real changes to system packages causing problems for us because GraphQL was not part of the normal release train. These days the new RPC stack is part of the release train, and we also update system packages less, but on the flip-side, external parties also operate these services and we have much less control over their deployments.

PR #26979, opened 2026-06-16
sui-node: embed the rpc-store into the fullnode behind a config flag

bmwill:
Add `RpcConfig::use_experimental_rpc_store`, an opt-in flag that serves a fullnode's rpc-api index surface from an embedded `sui-rpc-store` instance instead of the built-in `rpc-index`. The two index backends are mutually exclusive; raw chain data (objects, transactions, effects, events, and checkpoints) is still served from the perpetual and checkpoint stores either way, so the embedded store only owns the derived-index and ledger-history column families.

`EmbeddedRpcStore::bootstrap` owns the store's lifecycle. At startup it opens the rpc-store under the node's `db_path`, compares its persisted per-pipeline watermarks against the perpetual store's available range, and decides whether to resume the on-disk indexes as-is, re-seed the history cohort, or rebuild the live cohort. A rebuild bulk-loads the live cohort from the perpetual store (sharded by `ObjectID` prefix) and seeds the history cohort before the node starts executing, so the perpetual store's range stays stable for the duration of the blocking restore.

Once running, a tip-following indexer backfills the history cohort and follows the chain forward. It is fed by two new clients:

- `PerpetualStoreIngestionClient` assembles historical checkpoints from the node's own local stores, for backfill and gap-fill.
- `BroadcastStreamingClient` follows the tip over the checkpoint executor's broadcast stream, seeding from the local tip so an idle chain does not block the framework's initial `peek`.

To feed both the indexer and the subscription service, the checkpoint executor's `subscription_service_checkpoint_sender` changes from an `mpsc::Sender<Checkpoint>` to a `broadcast::Sender<Arc<Checkpoint>>`. Publishing is now non-blocking and best-effort: a send error just means there are no live subscribers, and subscribers recover any missed checkpoints from the local stores. The subscription service consumes the broadcast, tears down all in-progress subscriptions on `Lagged`, and -- when the embedded store is active -- holds a checkpoint back until the live cohort has committed it, preserving the synchronously-committed-index semantics a client relies on after observing a checkpoint.

The `AuthorityStorePruner` threads an optional `RpcStore` through and prunes its history cohort on the same floor as the legacy index, in lockstep.

`RpcStoreReadStore` composes the two backends behind the `sui-rpc-api` trait stack: raw chain data from `RocksDbStore`, the index surface from `RpcStoreReader`, with the available range taken as the intersection of the two.

Also add `SuiNode::embedded_rpc_store` for introspection, unit tests for the bootstrap decision and the two clients, and an e2e `restore` test that drives a dedicated fullnode through restarts -- toggling its index config between runs -- to exercise the resume/rebuild paths and confirm the index-backed `GetBalance` and `ListTransactions` RPCs answer correctly.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-core/src/rpc_store_ingestion_client.rs
```diff
@@ -0,0 +1,119 @@
...
+{
+    async fn chain_id(&self) -> anyhow::Result<ChainIdentifier> {
+        Ok(self.chain_id)
+    }
+
+    async fn checkpoint(&self, checkpoint: u64) -> CheckpointResult {
+        // A checkpoint below the perpetual store's pruning watermark, or above
+        // the highest executed one, is simply absent. Report it as NotFound so
+        // the framework's `wait_for` can retry checkpoints that have not been
+        // executed yet, while pruned checkpoints stay permanently unavailable.
```
nickvikeras, line 56: Treating "indexer fell behind the pruning watermark" and "indexer ran ahead of the tip of chain" as the same thing here is unlikely to be a problem in normal operations, but if the indexer did somehow fall behind, we'd just get stuck retrying fetching the pruned checkpoint indefinitely and would have to reason through why the system is stalled. 

I think we should at least log a very loud ERROR if we hit this case, if not trigger the `Bootstrap::Restore { clear: true }` path to automatically recover.

nickvikeras (approved):

crates/sui-core/src/rpc_store_ingestion_client.rs
bmwill: I'll go ahead and add an error log for now.

PR #26980, opened 2026-06-16
rpc-node: introduce sui-rpc-node crate

bmwill:
Add `sui-rpc-node`, a standalone, non-executing fullnode-replacement that exercises the entire rpc-store read stack end to end: the `sui-rpc-store` indexer (raw chain data, derived indexes, and the ledger-history bitmap column families) driven by the `sui-indexer-alt-framework` ingestion service, the `sui-consistent-store` RocksDB layout with cross-pipeline snapshot consistency, and the `sui-rpc-api` gRPC / HTTP read surface mounted over an `RpcStoreReader`.

It is a proof-of-concept that gates the embedded-fullnode integration: anything we want `sui-node` to host has to work here first. There is no execution path -- the binary only serves reads from indexed state.

The binary has four subcommands:

- `Run` opens (or creates) the database, builds the indexer with every pipeline enabled, and mounts the read server, resuming each pipeline from its persisted watermark (genesis on a fresh database).
- `Restore` rebuilds the derived-index and `objects` column families from a Sui formal snapshot, then floors the watermarks of the pipelines that cannot be sourced from a snapshot (raw chain data, bitmaps) to the snapshot anchor so a later `Run` resumes at `target_checkpoint + 1`.
- `Serve` mounts the read server over an existing database with no indexer, for inspecting a restored or previously indexed database as it is on disk.
- `GenerateConfig` prints the default TOML `ServiceConfig`.

`consistent_service` is a bespoke gRPC implementation of the `sui.rpc.consistent.v1alpha` service over the rpc-store schema, mirroring the surface served by `sui-indexer-alt-consistent-store` so its clients keep working unchanged when pointed at a `sui-rpc-node`.

Also add the TOML `ServiceConfig`, CLI args, and an extensive integration test suite covering the v2 state / ledger / move-package services, the v1alpha consistent service, the subscription service, and restore/seed behavior, with a shared cluster harness and Move test packages.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-rpc-node/src/consistent_service/objects.rs
+++ b/crates/sui-rpc-node/src/consistent_service/objects.rs
@@ -1,0 +1,338 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! `list_owned_objects` + `list_objects_by_type` handlers.
+//!
+//! `list_owned_objects` paginates over the `object_by_owner` CF
+//! at a snapshot, narrowed by the owner kind (address / object /
+//! shared / immutable) and, optionally, a `TypeFilter`. The
+//! `!` exclusion prefix is honoured: a request for
+//! `!0x2::coin::Coin<SUI>` returns all owned objects *except*
+//! that type. Internally that's the alt-consistent-store's
+//! `paginate_exclude` shape — for v1 here we implement the
+//! simpler "filter on the post-decode key" variant and leave
+//! a TODO to push that into the byte iterator if it becomes a
+//! hot path.
+//!
+//! `list_objects_by_type` is structurally the same but paginates
+//! over the `object_by_type` CF — every object of a given Move
+//! type, regardless of owner.
+
+use sui_consistent_store::Schema as _;
+use sui_consistent_store::SchemaAtSnapshot as _;
+use sui_consistent_store::Snapshot;
+use sui_indexer_alt_consistent_api::proto::rpc::consistent::v1alpha as grpc;
+use sui_indexer_alt_consistent_api::proto::rpc::consistent::v1alpha::owner::OwnerKind as ProtoOwnerKind;
+use sui_rpc_store::RpcStoreSchema;
+use sui_rpc_store::schema::object_by_owner;
+use sui_rpc_store::schema::type_filter::TypeFilter;
+use sui_types::base_types::ObjectID;
+use sui_types::base_types::SequenceNumber;
+use sui_types::base_types::SuiAddress;
+use sui_types::parse_sui_address;
+use sui_types::parse_sui_module_id;
+use sui_types::parse_sui_struct_tag;
+
+use crate::consistent_service::State;
+use crate::consistent_service::pagination::End;
+use crate::consistent_service::pagination::Page;
+use crate::consistent_service::state::Error as StateError;
+
+#[derive(Debug, thiserror::Error)]
+pub(super) enum Error {
+    #[error("missing owner")]
+    MissingOwner,
+    #[error("missing object_type")]
+    MissingType,
+    #[error("invalid owner kind")]
+    InvalidOwnerKind,
+    #[error("invalid owner: {0:?}")]
+    InvalidOwner(String),
+    #[error("invalid object_type: {0:?}")]
+    InvalidType(String),
+    #[error(transparent)]
+    Db(#[from] sui_consistent_store::error::Error),
+    #[error("failed to open schema: {0}")]
+    Open(#[from] sui_consistent_store::error::OpenError),
+    #[error(transparent)]
+    State(#[from] StateError),
+}
+
+impl From<Error> for tonic::Status {
+    fn from(e: Error) -> Self {
+        match e {
+            Error::MissingOwner
+            | Error::MissingType
+            | Error::InvalidOwnerKind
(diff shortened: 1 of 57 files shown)


crates/sui-rpc-node/src/consistent_service/state.rs
```diff
@@ -0,0 +1,169 @@
...
+use std::sync::Arc;
+
+use sui_consistent_store::Db;
+use sui_consistent_store::Snapshot;
+use sui_indexer_alt_consistent_api::proto::rpc::consistent::v1alpha::CHECKPOINT_HEIGHT_METADATA;
+use sui_indexer_alt_consistent_api::proto::rpc::consistent::v1alpha::LOWEST_AVAILABLE_CHECKPOINT_METADATA;
+use tonic::metadata::AsciiMetadataValue;
+
+use crate::config::PaginationConfig;
+use sui_rpc_store::ConsistencyConfig;
```
wlmyng, line 35: nit:

```
use tonic::metadata::AsciiMetadataValue;
use sui_rpc_store::ConsistencyConfig;

use crate::config::PaginationConfig;
```

crates/sui-rpc-node/src/consistent_service/state.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> Self {
+        Self {
+            db,
+            pagination: Arc::new(pagination),
+            consistency: Arc::new(consistency),
+        }
+    }
+}
+
+/// Errors raised by [`State`] helpers that map to specific
```
wlmyng, line 65: nit: I think these comments are shorter than the typical comment across the repo

crates/sui-rpc-node/src/consistent_service/objects.rs
```diff
@@ -0,0 +1,338 @@
...
+    // We model the "single owner address" listings via the
+    // existing prefix encoders. Shared / Immutable scans iterate
+    // the entire owner-kind range using a tiny ad-hoc encoder so
+    // we share the same pagination plumbing.
+    let resp = match kind {
+        OwnerSelector::Address(addr) => {
+            let prefix = object_by_owner::AddressOwnerPrefix(addr);
+            page.paginate_filtered(&schema.object_by_owner, &prefix, |_, key, _| {
+                matches_type_filter(filter.as_ref(), negated, &key.type_)
+            })?
```
wlmyng, line 103: could we use `AddressOwnerTypePrefix` if the filter is inclusive? Here and for `Object` arm (which would need an `ObjectOwnerTypePrefix`)

crates/sui-rpc-node/src/consistent_service/objects.rs
```diff
@@ -0,0 +1,338 @@
...
+        OwnerSelector::Shared => {
+            let prefix = KindPrefix(2);
+            page.paginate_filtered(&schema.object_by_owner, &prefix, |_, key, _| {
+                matches_type_filter(filter.as_ref(), negated, &key.type_)
+            })?
+        }
+        OwnerSelector::Immutable => {
+            let prefix = KindPrefix(3);
+            page.paginate_filtered(&schema.object_by_owner, &prefix, |_, key, _| {
+                matches_type_filter(filter.as_ref(), negated, &key.type_)
```
wlmyng, line 120: as is, all filter states (empty, inclusive, exclusion) become a full-range scan on the `-byte kind prefix with post-decode filtering

on inclusive: could we not leverage prefix narrowing instead of predicate-filtering?

on empty/ exclusive: in alt-consistent-store, we reject this combination (even with the byte iterator skip), because we only wanted to answer "owned objects of address X that are not Y". do we want a similar limitation here or is there a meaningful query that justifies this combination?

crates/sui-rpc-node/src/consistent_service/objects.rs
```diff
@@ -0,0 +1,338 @@
...
+//!
+//! `list_owned_objects` paginates over the `object_by_owner` CF
+//! at a snapshot, narrowed by the owner kind (address / object /
+//! shared / immutable) and, optionally, a `TypeFilter`. The
+//! `!` exclusion prefix is honoured: a request for
+//! `!0x2::coin::Coin<SUI>` returns all owned objects *except*
+//! that type. Internally that's the alt-consistent-store's
+//! `paginate_exclude` shape — for v1 here we implement the
+//! simpler "filter on the post-decode key" variant and leave
+//! a TODO to push that into the byte iterator if it becomes a
```
wlmyng, line 14: Is this a significant burden to include right now? At least on `alt-consistent-store` this was not a significant undertaking. How do we track and when do we say that negated queries are on the hot path? `RpcMetrics` ias per-method latency but do we distinguish an exclusion filter request from a cheap one?

wlmyng: `skip_past` already exists in `sui-consistent-store` in both directions, actually, so it seems we just need to extend the engine + handler to leverage that. Adding this ability would allow us to treat exclusion as a contiguous, skippable range

wlmyng (commented):
Making my way through the tests, leaving an initial review around object + pagination

PR #26982, opened 2026-06-16
[tx-timestamp 1/4] Add commit timestamp to ExecutionEnv and source it from consensus

lxfind:
## Description

Part 1 of a 4-PR stack (below) that feeds the consensus commit timestamp through execution so it can eventually be read from Move.

Adds `tx_timestamp_ms: Option<u64>` to `ExecutionEnv` (`None` when a transaction is not scheduled from a consensus commit) and populates it on the consensus execution path from the commit timestamp that ordered the transaction (threaded from `ConsensusHandler` through the execution-scheduler message). Also adds a debug assertion that, when a transaction reads the `Clock` as an immutable input and a commit timestamp is present, the two agree — guarding the invariant the rest of the stack relies on.

Nothing reads the value yet, so this is behavior-preserving.

### Stack
Feeds the consensus commit timestamp through execution so Move can read it (merge in order):
1. #26982 — `tx_timestamp_ms` on `ExecutionEnv`, sourced from consensus
2. #26994 — record / recover it in transaction effects
3. #26984 — thread it into `TxContext`
4. #27000 — write path + gated `tx_context::timestamp_ms` native

## Test plan

Covered by existing `sui-core` consensus/scheduler tests, which exercise the scheduler message and the test helpers updated here. Nothing downstream to assert yet because the value is unused until later PRs.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -166,8 +166,9 @@ use sui_types::sui_system_state::epoch_start_sui_system_state::EpochStartSystemS
 use sui_types::sui_system_state::{SuiSystemState, get_sui_system_state};
 use sui_types::supported_protocol_versions::{ProtocolConfig, SupportedProtocolVersions};
 use sui_types::{
-    SUI_SYSTEM_ADDRESS,
+    SUI_CLOCK_OBJECT_ID, SUI_SYSTEM_ADDRESS,
     base_types::*,
+    clock::Clock,
     committee::Committee,
     crypto::AuthoritySignature,
     error::{SuiError, SuiResult},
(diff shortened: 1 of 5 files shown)


crates/sui-core/src/authority.rs
```diff
@@ -791,6 +792,15 @@ pub struct ExecutionEnv {
...
     pub barrier_dependencies: Vec<TransactionDigest>,
+    /// Timestamp in milliseconds of the consensus commit that included the transaction.
+    /// This value will be used for reading timestamp in Move directly from TxContext without involving
+    /// Clock object.
+    ///
+    /// For almost all transactions, this value will be the same as if the transaction reads
+    /// the Clock object according to the consensus ordering. The only exception is the prologue
+    /// transaction itself, which reads the old Clock object that still contains the previous
+    /// commit's timestamp.
+    pub tx_timestamp_ms: Option<u64>,
```
mystenmark, line 803: I think we should consider refactoring this to avoid Option + unwrap, but we can do that in a different PR

mystenmark (approved):

crates/sui-core/src/authority.rs
lxfind: We cannot do that for transactions that do not have the timestamp in their effects today.

PR #26983, opened 2026-06-16
[tx-timestamp 2/3] Source commit timestamp on the checkpoint-execution path

lxfind:
## Description

Part 2 of the `tx_timestamp_ms` stack (depends on #26982).

When executing transactions from a checkpoint, derive each transaction's `ExecutionEnv.tx_timestamp_ms` from the `ConsensusCommitPrologue` that ordered it, so checkpoint-replaying fullnodes reproduce exactly the value validators fed in from consensus.

A checkpoint can contain several whole commits (each led by a prologue) and can also begin mid-commit when a large commit is split across checkpoints, so the prologue timestamp is tracked while walking the checkpoint in order and carried across checkpoints (the `ExecuteTransactions` pipeline stage runs strictly in sequence order). On cold start / restart the carry is seeded from the on-chain `Clock`, which the prior commit's prologue already set. Still behavior-preserving — nothing reads the value yet.

## Test plan

Covered by existing checkpoint-executor tests in `sui-core`. No behavior change to assert downstream (value unused until part 3+); the logic is structured so validator and checkpoint paths populate the identical timestamp.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
+++ b/crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
@@ -23,7 +23,9 @@ use mysten_common::{ZipDebugEqIteratorExt, debug_fatal, fatal, izip_debug_eq};
 use parking_lot::Mutex;
 use std::{sync::Arc, time::Instant};
 use sui_types::SUI_ACCUMULATOR_ROOT_OBJECT_ID;
+use sui_types::SUI_CLOCK_OBJECT_ID;
 use sui_types::base_types::SequenceNumber;
+use sui_types::clock::Clock;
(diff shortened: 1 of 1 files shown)


lxfind:
Closing — we're dropping the checkpoint-side timestamp derivation entirely. The commit timestamp will instead be recorded into transaction effects (a new `UnchangedConsensusKind` variant) at authoritative execution, and checkpoint execution + replay will read it back from the effects rather than re-deriving it. That removes the carry/seed/prologue-walk complexity this PR was wrestling with.

PR #26988, opened 2026-06-17
Remove deadcode from authority logic

mwtian:
## Description

Remove dead code that accumulated in the authority subsystem — unused methods, functions, types, fields, and enum variants across `AuthorityState`, `AuthorityStore`, `AuthorityPerEpochStore`, and the perpetual/epoch table holders. Every removed item had zero references repo-wide.

The one non-obvious removal is the deprecated `transaction_cert_signatures` column family: it has no accessors, and dropping the declaration is safe because typed_store reopens any on-disk column families the schema no longer declares.

## Test plan

Covered by existing tests. Verified the full workspace builds and lints clean (`cargo clippy --workspace --all-targets`) and that the `tidehunter` configuration still compiles.

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -105,7 +105,7 @@ use tracing::{debug, error, info, instrument, warn};
 
 use self::authority_store::ExecutionLockWriteGuard;
 use self::authority_store_pruner::{AuthorityStorePruningMetrics, PrunerWatermarks};
-pub use authority_store::{AuthorityStore, ResolverWrapper, UpdateType};
+pub use authority_store::{AuthorityStore, ResolverWrapper};
 use mysten_metrics::{monitored_scope, spawn_monitored_task};
 
 use crate::jsonrpc_index::IndexStore;
@@ -1280,12 +1280,6 @@ impl AuthorityState {
             .check_system_overload_at_signing
     }
 
-    pub fn check_system_overload_at_execution(&self) -> bool {
-        self.config
-            .authority_overload_config
-            .check_system_overload_at_execution
-    }
-
     pub(crate) fn check_system_overload(
         &self,
         tx_data: &SenderSignedData,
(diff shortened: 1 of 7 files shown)


aschran (approved):

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -496,11 +468,6 @@ pub struct AuthorityEpochTables {
     /// to disk.
     signed_effects_digests: DBMap<TransactionDigest, TransactionEffectsDigest>,
 
-    /// No longer used.
-    #[allow(dead_code)]
-    #[deprecated(note = "column family retained only for backward DB compatibility")]
-    transaction_cert_signatures: DBMap<TransactionDigest, AuthorityStrongQuorumSignInfo>,
```
aschran, line 502: just triple checking you're sure this is safe to remove?

akichidis: I believe for rocksdb it's safe now to do so. Not so sure about Tidehunter?

crates/sui-core/src/authority.rs
```diff
@@ -1280,12 +1280,6 @@ impl AuthorityState {
             .check_system_overload_at_signing
     }
 
-    pub fn check_system_overload_at_execution(&self) -> bool {
-        self.config
-            .authority_overload_config
-            .check_system_overload_at_execution
```
akichidis, line 1286: Is the `check_system_overload_at_execution` used anymore? If not maybe we can clean up that as well?

crates/sui-core/src/authority/authority_per_epoch_store.rs
lxfind: Adding @andll to confirm, likely not safe for tidehunter

andll: I will add feature to remove tables ~soon, but for now please keep it as dead_code

mwtian: Yes removing column family is now safe on rocksdb. But I will keep this for tidehunter.

PR #26990, opened 2026-06-17
[gql] VM Error Visibility: simulateTransaction and transaction(digest)

jordanjennings-mysten:
## Description 

Optional field "metadata" on ExecutionError type in graphql schema provides improved error information for simulate and get transaction.

## Test plan 

sui-indexer-alt-e2e-tests
- test_transaction_query_insufficient_funds_error_metadata
- test_simulate_transaction_insufficient_funds_error_metadata

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: simulateTransaction and transaction(digest) now include error metadata
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/src/graphql.rs
+++ b/crates/sui-indexer-alt-e2e-tests/src/graphql.rs
@@ -1,0 +1,141 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::net::IpAddr;
+use std::net::Ipv4Addr;
+use std::net::SocketAddr;
+
+use anyhow::Context;
+use prometheus::Registry;
+use reqwest::Client;
+use serde_json::Value;
+use serde_json::json;
+use sui_futures::service::Service;
+use sui_indexer_alt::config::IndexerConfig;
+use sui_indexer_alt::setup_indexer;
+use sui_indexer_alt_framework::IndexerArgs;
+use sui_indexer_alt_framework::ingestion::ClientArgs;
+use sui_indexer_alt_framework::ingestion::ingestion_client::IngestionClientArgs;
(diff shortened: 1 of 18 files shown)


crates/sui-indexer-alt-e2e-tests/src/graphql.rs
amnn, line 1: Could we just use `OffchainCluster` for this? Arguably `graphql_simulate_transaction_tests` should do the same.

amnn (changes requested):
@jordanjennings-mysten, please organise a discussion with folks who work on Indexing and RPCs to explain the aims of your work and get them to help design the best way to expose that information in the RPC layer.

This integration seems to be missing a number of important steps that a conversation like that would likely highlight:

- It only exposes information via fullnodes -- what about the indexing pathways, dealing with fullnode retention, query performance?
- It does not abstract the ExecutionMetadata -- now there are two places in the schema where we are exposing error message information, which one is canonical, how should they be used together? What does this mean for schema evolution at the RPC-layer in future?

crates/sui-indexer-alt-e2e-tests/src/graphql.rs
jordanjennings-mysten: It seemed the options were:
- make a smaller version of OffchainCluster to avoid spinning up all of the services
- modify OffchainCluster to disable some of the parts unnecessary 
- use the whole OffchainCluster which was a bit heavier than what this test needed
open to whichever here.

amnn: I would opt for the last one mostly because the cost of these extra services is not that high/noticeable, whereas the cognitive overhead of having different cluster variants is noticeable. All our transactional tests run with all these services enabled.

(But also this is not a big deal and we should sort out the core design questions first, so feel free to ignore this comment!)

PR #26991, opened 2026-06-17
[cli] upstream move lint

damirka:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [x] CLI: adds `sui move lint` command
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-move/src/lib.rs
+++ b/crates/sui-move/src/lib.rs
@@ -13,6 +13,7 @@ pub mod cache_package;
 pub mod coverage;
 pub mod disassemble;
 pub mod format;
+pub mod lint;
 pub mod migrate;
 pub mod new;
 pub mod summary;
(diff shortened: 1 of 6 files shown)


stefan-mysten (approved):
This looks good, from my end. Would it make sense to add a more complete set of test-suites or it doesn't really matter? (a.k.a. will these existing lints possibly break over time or we're confident they're stable?)

damirka:
> Would it make sense to add a more complete set of test-suites or it doesn't really matter? (a.k.a. will these existing lints possibly break over time or we're confident they're stable?)

We've enough test cases in the move-compiler / move-cli, so I think we should be good for now!

PR #26994, opened 2026-06-17
[tx-timestamp 2/4] Add commit timestamp to effects

lxfind:
## Description

Part 2 of the `tx_timestamp_ms` stack (below); stacked on #26982.

PR1 sources the commit timestamp into `ExecutionEnv` on the live consensus path, but checkpoint execution and replay rebuild `ExecutionEnv` from a transaction's effects, where it isn't recorded. This persists it in effects (v2) so it survives into re-execution.

A trailing `UnchangedConsensusKind::ConsensusCommitTimestamp(u64)` variant carries the value, exposed via a new `consensus_commit_timestamp()` accessor on the effects API; the checkpoint executor reads it back into `ExecutionEnv`. No transaction produces the variant yet, so this is behavior-preserving. RPC/SDK/GraphQL rendering of the new variant is left as `todo!()` pending the corresponding variants in the external rust-sdk repo and a dedicated timestamp surface.

### Stack
Feeds the consensus commit timestamp through execution so Move can read it (merge in order):
1. #26982 — `tx_timestamp_ms` on `ExecutionEnv`, sourced from consensus
2. #26994 — record / recover it in transaction effects
3. #26984 — thread it into `TxContext`
4. #27000 — write path + gated `tx_context::timestamp_ms` native

## Test plan

Behavior-preserving plumbing with no producer yet, so it's exercised by existing effects and checkpoint-execution tests. The new effects variant is captured in the BCS format snapshot (`format__sui.yaml.snap`); it's appended in last position and therefore wire-compatible.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
+++ b/crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
@@ -916,6 +916,10 @@ impl CheckpointExecutor {
                             env = env.with_insufficient_funds();
                         }
 
+                        if let Some(timestamp) = effects.consensus_commit_timestamp() {
+                            env = env.with_tx_timestamp_ms(timestamp);
+                        }
+
                         Some((tx_digest, (txn.clone(), env)))
                     }
                 },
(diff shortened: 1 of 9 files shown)


mystenmark:
This should be unnecessary - the commit timestamps are recorded in the ConsensusCommitPrologue transactions, so checkpoint execution can rebuild ExecutionEnv without new information in effects.

PR #26999, opened 2026-06-17
[cli] Update the sui client balance to use gRPC list_balances

stefan-mysten:
## Description 

I realized that the CLI `balance` command does not use the gRPC `list_balances` so it does not properly support address balances. This change moves to using `list_balances` for these two cases.
```
sui client balance
sui client balance --coin-type
```

When `--with-coins` is passed, we still need to fetch all the owned objects and then display those. For this case, it also makes sure that the address balance is taken into account.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/src/client_commands.rs
+++ b/crates/sui/src/client_commands.rs
@@ -16,6 +16,7 @@ use std::{
     path::{Path, PathBuf},
     str::FromStr,
     sync::Arc,
+    time::Duration,
 };
 use sui_rpc::proto::sui::rpc::v2::{self as proto};
 
@@ -118,6 +119,8 @@ use tracing::{debug, info};
 
 /// Concurrency level for fetching coin metadata for balances.
 const NUM_CONCURRENCY_REQS: usize = 8;
+/// Rate limit for RPC calls to avoid being throttled by the server. This is equivalent to 20rps.
+const RATE_LIMIT_MILLIS: u64 = 50;
 
 pub(crate) static USER_AGENT: &str =
     concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
@@ -3004,18 +3007,21 @@ async fn balance_outputs_for_address(
         client.list_balances(address).try_collect().await?
     };
 
-    futures::stream::iter(balances)
-        .map(|balance| async move {
-            let metadata = coin_metadata_for_balance(client, &balance).await?;
-            Ok(BalanceOutput {
-                metadata,
-                balance,
-                coins: Vec::new(),
-            })
+    tokio_stream::StreamExt::throttle(
+        futures::stream::iter(balances),
+        Duration::from_millis(RATE_LIMIT_MILLIS),
+    )
+    .map(|balance| async move {
+        let metadata = coin_metadata_for_balance(client, &balance).await?;
+        Ok(BalanceOutput {
+            metadata,
+            balance,
+            coins: Vec::new(),
         })
-        .buffered(NUM_CONCURRENCY_REQS)
-        .try_collect()
-        .await
+    })
+    .buffered(NUM_CONCURRENCY_REQS)
+    .try_collect()
+    .await
 }
 
 /// Best-effort metadata lookup for a balance returned by the balance API.
--- a/crates/sui/src/client_commands.rs
+++ b/crates/sui/src/client_commands.rs
@@ -10,7 +10,7 @@ use crate::{
 };
 use futures::TryStreamExt;
 use std::{
-    collections::{BTreeMap, BTreeSet, btree_map::Entry},
+    collections::{BTreeMap, BTreeSet},
     fmt::{Debug, Display, Formatter, Write},
     fs,
     path::{Path, PathBuf},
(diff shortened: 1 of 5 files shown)


crates/sui/src/client_commands.rs
```diff
@@ -3029,18 +2987,166 @@ pub async fn request_tokens_from_faucet(
...
+/// Normalize equivalent type strings before grouping and ordering balances.
+fn canonicalize_type(type_: &str) -> Result<String, anyhow::Error> {
+    Ok(TypeTag::from_str(type_)
+        .context("Cannot parse coin type")?
+        .to_canonical_string(/* with_prefix */ true))
+}
+
+/// Keep SUI first while preserving the balance API's order for other coin types.
+fn order_balance_outputs_sui_first(balances: &mut Vec<BalanceOutput>) -> Result<(), anyhow::Error> {
+    let sui_type_tag = canonicalize_type(SUI_COIN_TYPE)?;
```
amnn: nit: I think we have `GasCoin::type_()` that we can use here instead of re-parsing the type to canonicalize it.

crates/sui/src/client_commands.rs
```diff
@@ -3029,18 +2987,166 @@ pub async fn request_tokens_from_faucet(
...
-    builder: &mut TableBuilder,
-    with_coins: bool,
-) {
-    let format_decmials = 2;
+/// Parse the CLI coin type argument as the inner type `T`, not as `Coin<T>`.
+fn parse_coin_type_arg(coin_type: String) -> Result<StructTag, anyhow::Error> {
+    match coin_type.parse::<TypeTag>()? {
+        TypeTag::Struct(coin_type) => Ok(*coin_type),
+        coin_type => bail!("coin type must be a struct type, got {coin_type}"),
+    }
```
amnn: Why not parse a `StructTag` directly?

crates/sui/src/client_commands.rs
```diff
@@ -3029,18 +2987,166 @@ pub async fn request_tokens_from_faucet(
...
+) -> Result<Vec<BalanceOutput>, anyhow::Error> {
+    let balances = if let Some(coin_type) = coin_type {
+        vec![client.get_balance(address, coin_type).await?]
+    } else {
+        client.list_balances(address).try_collect().await?
+    };
+
+    let mut outputs = Vec::with_capacity(balances.len());
+    for balance in balances {
+        let metadata = coin_metadata_for_balance(client, &balance).await?;
```
amnn: This is going to issue a sequential RPC call for each balance/coin type. We should issue these concurrently instead, potentially with some throttling to be kind for rate limit purposes (async `Stream`s can be a good way of implementing rate limited concurrent requests, especially with the helper traits from `tokio_stream` and `future::stream`).

crates/sui/src/client_commands.rs
```diff
@@ -3029,18 +2987,166 @@ pub async fn request_tokens_from_faucet(
...
+                previous_transaction: o.previous_transaction,
+            }))
+        })
+        .try_collect()
+        .await?;
+
+    let mut coins_by_type: BTreeMap<String, Vec<Coin>> = BTreeMap::new();
+    for coin in coins {
+        coins_by_type
+            .entry(canonicalize_type(&coin.coin_type)?)
```
amnn: There's multiple levels of canonicalization going on here -- above you convert the coin_type to a canonical string, and then you parse it here to canonicalize it.

crates/sui/src/client_commands.rs
```diff
@@ -3029,18 +2987,166 @@ pub async fn request_tokens_from_faucet(
...
+        });
+    }
+
+    Ok(())
+}
+
+/// Build the object type filter expected by `list_owned_objects`.
+fn coin_object_type_filter(coin_type: Option<&StructTag>) -> StructTag {
+    if let Some(coin_type) = coin_type {
+        sui_types::coin::Coin::type_(TypeTag::Struct(Box::new(coin_type.clone())))
```
amnn: nit: import `Coin`, and then use the `From` impl (which I **think** exists).

```suggestion
        Coin::type_(coin_type.clone().into())
```

crates/sui/src/client_commands.rs
```diff
@@ -3029,18 +2987,166 @@ pub async fn request_tokens_from_faucet(
...
+    let coin_balance = coins
+        .iter()
+        .fold(0u64, |total, coin| total.saturating_add(coin.balance));
+    let mut balance = proto::Balance::default()
+        .with_coin_type(coin_type)
+        .with_balance(coin_balance);
+    if coin_balance != 0 {
+        balance.set_coin_balance(coin_balance);
+    }
+    balance
```
amnn: Couple of things strike me as odd about this:

- Why do we only conditionally set the coin balance, but not the overall (total balance)?
- When would list balances not return a balance for an address, such that we need to recover the coin balance by summing over the listed coins?

amnn (approved):
Accepting to unblock, but there were a couple of strange things to look into before landing. Thanks @stefan-mysten

crates/sui/src/client_commands.rs
stefan-mysten: Yep, changed.

crates/sui/src/client_commands.rs
stefan-mysten: Good point, I'll try that.

crates/sui/src/client_commands.rs
stefan-mysten: Thanks for spotting this, removed it and instead I just clone the `coin_type`.

crates/sui/src/client_commands.rs
stefan-mysten: You're right, we should not try to rebuild the balance. `list_balances` returns all known coin balances. I deleted this function and removed the calls to it.

crates/sui/src/client_commands.rs
stefan-mysten: Ok, tried to do something simple though not sure it's the right way to do it.

stefan-mysten:
> Accepting to unblock, but there were a couple of strange things to look into before landing. Thanks @stefan-mysten

Thanks so much for the review. I tried to address all the feedback but still a bit unsure about the streaming part as I did not use `tokio_stream`, and only used `futures::stream`.

crates/sui/src/client_commands.rs
amnn: What you have here limits the number of concurrent in flight requests, but won't limit the number of requests over a period of time. The [`throttle`](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.throttle) stream adapter  can be used for this though -- it introduces a delay between elements in the stream, so if you put it in front of your map, and set it to a rate limit, you know that you won't exceed a time-based rate limit either:

```rust
    futures::stream::iter(balances)
        .throttle(delay)
        .map(|balance| async move {
            let metadata = coin_metadata_for_balance(client, &balance).await?;
            Ok(BalanceOutput {
                metadata,
                balance,
                coins: Vec::new(),
            })
        })
        .buffered(NUM_CONCURRENCY_REQS)
        .try_collect()
        .await
```

(the parameter is given as a delay, so if the rate limit is 20rps, then you need to provide a rate limit of 50ms).

Note that `StreamExt` from `futures` and from `tokio_stream` can conflict with each other in kind of annoying ways, unfortunately.

stefan-mysten: That makes sense, thanks. I'll look into using `stream` from `tokio_stream` instead.

amnn: You may need to use both, sadly, because `tokio_stream` doesn't have `buffered` 😢

stefan-mysten: Yes, I've been looking at it for the past hour and trying to understand if I can go around it. There's a way to avoid futures, but then we have to clone the Client to pass to the spawned task. It's not the prettiest but might work.

```rust
  let mut outputs = Vec::with_capacity(balances.len());
  let balance_stream = tokio_stream::iter(balances.into_iter().enumerate())
      .throttle(COIN_METADATA_REQUEST_INTERVAL);
  tokio::pin!(balance_stream);

  let mut tasks = tokio::task::JoinSet::new();

  while let Some((idx, balance)) = balance_stream.next().await {
      while tasks.len() >= NUM_CONCURRENCY_REQS {
          if let Some(result) = tasks.join_next().await {
              outputs.push(result??);
          }
      }

      let client = Client::clone(client);
      tasks.spawn(async move {
          let metadata = coin_metadata_for_balance(&client, &balance).await?;
          Ok::<_, anyhow::Error>((
              idx,
              BalanceOutput {
                  metadata,
                  balance,
                  coins: Vec::new(),
              },
          ))
      });
  }

  while let Some(result) = tasks.join_next().await {
      outputs.push(result??);
  }

  outputs.sort_by_key(|(idx, _)| *idx);
  Ok(outputs.into_iter().map(|(_, output)| output).collect())
 ```

stefan-mysten: From what I can tell, cloning the `sui_rpc::Client` might not be too bad. It does not recreate the connection. Definitely a bit heavier than the alternative, I think.

stefan-mysten: There's also the question of rate limit values. By default, apparently the public gRPC is limited to 100req/30 sec.

stefan-mysten: Hmm, I ended using the initial version and adding throttle to it. As this is hopefully not a critical command, I hope that things will work out ok.

PR #27003, opened 2026-06-18
[1/5] Add ExecutionRetryError plumbing from execution to authority

lxfind:
## Description

Base of the stack. Threads a transaction-retry signal out of execution to the authority, so a native can later request that a transaction be retried later rather than committed.

Plumbing only — nothing requests a retry on this branch. The temporary store owns an interior-mutable retry flag (`ExecutionRetryError::SystemObjectUnavailable { object_id }`); execution checks it before any gas or effects, so the transaction bails without side effects, and the error is threaded out of the executor to the authority (which `expect()`s it for now — generic handling arrives in 3/5).

## Stack

1/5 #27003 — `ExecutionRetryError` plumbing from execution to authority
2/5 #27019 — rename `ChildObjectResolver` → `RuntimeObjectResolver`
3/5 #27036 — waitable system object framework
4/5 #27037 — record read system objects in effects
5/5 #27038 — check object funds withdraw sufficiency in the Move VM

## Test plan

No behavior change: the retry flag is never set on this branch, so the early bail-out is unreachable. The executor signature change is exercised by existing execution paths across `sui-core`, `sui-replay`, `simulacrum`, and genesis.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/simulacrum/src/epoch_state.rs
+++ b/crates/simulacrum/src/epoch_state.rs
@@ -169,7 +169,8 @@ impl EpochState {
                 signer,
                 tx_digest,
                 &mut None,
-            );
+            )
+            .expect("simulacrum execution does not produce retry errors");
         Ok((inner_temp_store, gas_status, effects, result))
     }
 }
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -69,6 +69,7 @@ use std::{
 use sui_config::NodeConfig;
 use sui_config::node::{AuthorityOverloadConfig, StateDebugDumpConfig};
 use sui_execution::Executor;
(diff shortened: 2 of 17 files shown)


lxfind:
Superseded by the reorganized 3-PR stack: #27071 (retry infra) → #27072 (sufficiency API) → #27073 (in-VM check). The retry mechanism was reworked to a post-execution signal (no early-exit), and the generic waitable-framework, effects-recording, and ChildObjectResolver rename were dropped. Closing.

PR #27004, opened 2026-06-18
Log GraphQL requests at trace level for replay

evan-wall-mysten:
## Description

Captures GraphQL requests so they can be replayed later. Each request's verbatim payload (query, variables, operationName) is logged on the `graphql_request` tracing target at `trace` level, with the operation kind recorded on the enclosing span.

Disabled by default. Capture all kinds with `RUST_LOG=graphql_request=trace`, or a single kind with an `EnvFilter` span-field directive, e.g. `RUST_LOG="graphql_request[{kind=mutation}]=trace"` (add `RUST_LOG_JSON=1` / `RUST_LOG_FILE` for newline-delimited JSON). Capture reuses the document parsed during normal request execution, so it never re-parses the query and stays a no-op when disabled.

## Test plan

New unit tests assert that a span-field directive captures only the configured operation kind, that `graphql_request=trace` captures all kinds, and that capture is off by default.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [x] GraphQL: GraphQL request payloads can be captured for replay by enabling the `graphql_request` tracing target at trace level — all kinds via `RUST_LOG=graphql_request=trace`, or a single operation kind via `RUST_LOG="graphql_request[{kind=mutation}]=trace"`. Disabled by default.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/extensions/logging.rs
+++ b/crates/sui-indexer-alt-graphql/src/extensions/logging.rs
@@ -25,7 +25,9 @@ use async_graphql::extensions::NextRequest;
 use async_graphql::extensions::NextResolve;
 use async_graphql::extensions::NextValidation;
 use async_graphql::extensions::ResolveInfo;
+use async_graphql::parser::types::DocumentOperations;
 use async_graphql::parser::types::ExecutableDocument;
+use async_graphql::parser::types::OperationType;
 use axum::http::HeaderMap;
 use axum::http::HeaderName;
 use pin_project::pin_project;
@@ -36,6 +38,8 @@ use tracing::debug;
 use tracing::info;
 use tracing::log::Level;
 use tracing::log::log_enabled;
+use tracing::trace;
+use tracing::trace_span;
 use tracing::warn;
 use uuid::Uuid;
 
@@ -92,6 +96,9 @@ pub(crate) struct Logging(pub Arc<RpcMetrics>);
 struct LoggingExt {
     session: Arc<OnceLock<Session>>,
     query: Arc<OnceLock<String>>,
+    /// The client-selected operation name, stashed by `prepare_request` so `parse_query` can
+    /// classify the operation that will execute (it is not reachable from `parse_query` otherwise).
+    operation_name: Arc<OnceLock<Option<String>>>,
     metrics: Arc<RpcMetrics>,
 }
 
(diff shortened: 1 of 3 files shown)


amnn (commented):
I would personally not use configs for this. Instead I would use the `trace!` log level so that we could enable this using the `RUST_LOG` environment variable (you can even enable the logging for a particular operation type if the type is included in the label).

Just less boilerplate for passing this kind of config through, and more consistent with how we manage telemetry configuration in other cases. Thoughts?

evan-wall-mysten:
> I would personally not use configs for this. Instead I would use the `trace!` log level so that we could enable this using the `RUST_LOG` environment variable (you can even enable the logging for a particular operation type if the type is included in the label).
> 
> Just less boilerplate for passing this kind of config through, and more consistent with how we manage telemetry configuration in other cases. Thoughts?

I'm not sure it is possible to filter on a log label using `RUST_LOG`? The closest thing I could figure out how to do is to wrap the log in a span and add a label to the span which still allowed me to do it without adding an extra config field.

Let me know if this is what you were looking for or if there is a way to filter on a log label itself.

crates/sui-indexer-alt-graphql/src/extensions/logging.rs
```diff
@@ -364,17 +369,80 @@ fn sanitize_sdk_version(sdk_type: &str, value: &str, config: &LoggingConfig) ->
...
+        }),
+        "Captured request",
+    );
+}
+
+/// Parse `request`'s query and return the kind of the operation that will execute (`query`,
+/// `mutation`, or `subscription`), selecting by operation name for multi-operation documents.
+/// Returns `None` if the query fails to parse or the operation cannot be resolved.
+fn operation_kind(request: &Request) -> Option<&'static str> {
+    let doc = parse_query(&request.query).ok()?;
```
amnn: It's a bit unfortunate that we are parsing the whole request again just for this information -- can we move where the request is captured so that we can take advantage of the already parsed request?

amnn (approved):
Yep, this looks good, thanks @evan-wall-mysten . I had a question about potentially avoiding the re-parse but seeing as this is not a cost we will pay often, it's fine if that's not easily avoidable.

crates/sui-indexer-alt-graphql/src/extensions/logging.rs
evan-wall-mysten: Done. This also prevents `capure_request` from needing to be called by each GraphQL handler.

PR #27008, opened 2026-06-18
[adapter] Simplify and remove stale gas-related gating

tzakian:
## Description 

Remove some gating and flagging that is no longer needed in the current execution version as it was gating behavior that pre-existed the current execution version.

## Test plan 

CI + 👀 and looking at protocol config values.
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-adapter/src/gas_charger.rs
+++ b/sui-execution/latest/sui-adapter/src/gas_charger.rs
@@ -17,9 +17,7 @@ pub mod checked {
     use sui_types::digests::TransactionDigest;
     use sui_types::error::ExecutionErrorTrait;
     use sui_types::gas::{GasCostSummary, SuiGasStatus, deduct_gas};
-    use sui_types::gas_model::gas_predicates::{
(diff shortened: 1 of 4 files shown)


tnowacki (approved):

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -482,42 +472,13 @@ pub mod checked {
         /// If we exceed gas_budget, we set execution_result to InsufficientGas, failing the tx.
```
cgswords, line 472: Could we add a one-line breadcrumb here that older execution versions retain a v1 fallback? Something like `// (older execution versions retain a v1 fallback; see vN/sui-adapter)`.

cgswords (commented):
Would it be worth adding a snapshot test that exercises a publish *and* an upgrade with the new charging path, to pin the behavior?

PR #27009, opened 2026-06-18
[adapter] Add `PackagePayload` enum

tzakian:
## Description 
Add a `PackagePayload` from the loaded SPTB AST on-down, but do not create the `PackagePayload::Deserialized` variant yet.

Plumbing for future work in the PR stacked on this that adds unified linkage.

## Test plan 

CI
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
+++ b/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
@@ -12,7 +12,10 @@ use crate::{
     static_programmable_transactions::{
         execution::context::subst_signature,
         linkage::{analysis::LinkageAnalyzer, resolved_linkage::ExecutableLinkage},
-        loading::ast::{self as L, Datatype, LoadedFunction, LoadedFunctionInstantiation, Type},
+        loading::ast::{
+            self as L, Datatype, DeserializedPackage, LoadedFunction, LoadedFunctionInstantiation,
+            Type,
+        },
     },
 };
 use move_binary_format::{
@@ -602,11 +605,11 @@ where
         self.load_vm_type_from_type_tag(Some(type_arg_idx), &tag)
     }
 
-    pub fn deserialize_modules(
+    pub fn deserialize_package(
         &self,
         module_bytes: &[Vec<u8>],
         dep_ids: &[ObjectID],
-    ) -> Result<(Vec<CompiledModule>, usize, [u8; 32]), Mode::Error> {
+    ) -> Result<DeserializedPackage, Mode::Error> {
         assert_invariant!(
             !module_bytes.is_empty(),
             "empty package is checked in transaction input checker"
@@ -615,7 +618,7 @@ where
         let total_bytes = module_bytes.iter().map(|v| v.len()).sum();
 
         let binary_config = self.protocol_config.binary_config(None);
-        let modules = module_bytes
+        let deserialized_modules = module_bytes
             .iter()
             .map(|b| {
                 CompiledModule::deserialize_with_config(b, &binary_config)
@@ -628,7 +631,11 @@ where
             dep_ids,
             /* hash_modules */ true,
         );
-        Ok((modules, total_bytes, computed_digest))
+        Ok(DeserializedPackage {
+            deserialized_modules,
+            total_bytes,
+            computed_digest,
+        })
     }
 }
 
--- a/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
+++ b/sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
@@ -16,7 +16,8 @@ use crate::{
     },
 };
 use move_binary_format::{
-    errors::VMError,
+    CompiledModule,
+    errors::{Location, VMError, VMResult},
     file_format::{Ability, AbilitySet, TypeParameterIndex},
 };
 use move_core_types::{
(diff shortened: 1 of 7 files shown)


sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
```diff
@@ -127,16 +128,27 @@ pub enum Command {
...
     ),
 }
 
+#[derive(Debug, Clone)]
+pub enum PackagePayload {
+    Serialized(Vec<Vec<u8>>),
+    // NB: deserialized, but not yet verified.
+    Deserialized {
+        modules: Vec<CompiledModule>,
+        total_bytes: usize,
```
tnowacki: maybe `total_bcs_bytes`? (I'm assuming the bytes were from bcs)

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
```diff
@@ -127,16 +128,27 @@ pub enum Command {
...
+        computed_digest: [u8; 32],
```
tnowacki: nit, maybe a small comment of how the digest is computed

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -1022,27 +1022,33 @@ where
     //
     // Publish and Upgrade
     //
-    pub fn deserialize_modules(
+
+    pub fn get_modules(
```
tnowacki: I tend to dislike `get_` prefixes, especially since it is not rooted in `self` here.
Maybe `package_payload_modules`?  `package_payload_to_compiled_modules`? `get_or_deserialize_modules`?

tnowacki (approved):

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
tzakian: I like `package_payload_modules`

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
tzakian: They're the sum length of the compiled module bytes, so not BCS necessarily...

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
```diff
@@ -127,16 +128,30 @@ pub enum Command {
...
         ResolvedLinkage,
     ),
 }
 
+#[derive(Debug, Clone)]
+pub enum PackagePayload {
+    Serialized(Vec<Vec<u8>>),
+    Deserialized {
+        // NB: Modules are deserialized but not yet verified.
+        modules: Vec<CompiledModule>,
```
cgswords: Maybe we should name these `deserialized_modules`, to make it clear these have not undergone verification yet.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
```diff
@@ -127,16 +128,30 @@ pub enum Command {
...
         ObjectID,
         Argument,
         ResolvedLinkage,
     ),
 }
 
+#[derive(Debug, Clone)]
+pub enum PackagePayload {
+    Serialized(Vec<Vec<u8>>),
+    Deserialized {
```
cgswords: Should this also carry `dep_ids` for sanity? If even just a hash?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
```diff
@@ -600,6 +601,35 @@ where
         let tag = to_type_tag_internal(self, type_arg_idx, ty)?;
         self.load_vm_type_from_type_tag(Some(type_arg_idx), &tag)
     }
+
+    pub fn deserialize_modules(
```
cgswords: Can we make this return a struct instead of a tuple-by-convention? Named fields would make this a little easier to read.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -1022,27 +1022,33 @@ where
...
+        dep_ids: &[ObjectID],
+    ) -> Result<(Vec<CompiledModule>, [u8; 32]), Mode::Error> {
+        Ok(match package_payload {
+            PackagePayload::Deserialized {
+                modules,
+                computed_digest,
+                ..
+            } => (modules, computed_digest),
+            PackagePayload::Serialized(module_bytes) => {
+                // This assertion is also checked in the call to `deserialize_modules`, but we
```
cgswords, line 1034: Call out what behavior is being retained?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -1022,27 +1022,33 @@ where
     //
     // Publish and Upgrade
     //
-    pub fn deserialize_modules(
+
+    pub fn package_payload_modules(
```
cgswords: Nit: Unify names `package_payload_modules` on context and `deserialize_modules` on env? So we can tell they are the same data?

cgswords (commented):
Could we add a small test that constructs a `Deserialized` variant directly and runs it through the interpreter, so the new arm gets coverage before LUMPY lands?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
tzakian: dep_ids is held in the `Upgrade` or `Publish` command that is wrapping this, so no need to hold them here too IMO.

tzakian:
> Could we add a small test that constructs a Deserialized variant directly and runs it through the interpreter, so the new arm gets coverage before LUMPY lands?

These are all going to land at the same time, so no real need as the LUMPY PR switches over to always using this path immediately (IMO).

PR #27013, opened 2026-06-19
Add object funds sufficiency check to RuntimeObjectResolver

lxfind:
## Description

Adds `RuntimeObjectResolver::check_object_funds_sufficiency`, implemented by `TemporaryStore`. It gates on the accumulator root object being available at the version this transaction requires (via `is_system_object_available`, #27020): if the root has no assigned version the availability check errors (a withdrawal is always sequenced against the accumulator root, so its absence is an invariant violation); if it has not caught up, it records a retry request and returns `Unknown`. Otherwise it loads the owner's balance at the required version and compares it against the cumulative amount withdrawn from that `(owner, type)` so far in the transaction — so multiple withdrawals from the same account in one PTB are summed — returning `Sufficient` or `Insufficient`.

A `TODO` marks the one known gap: the accumulator root reaching its required version does not by itself guarantee an individual account has settled to it, so the readiness condition needs a per-account settlement signal.

## Stack
3. #27020 — generic `is_system_object_available`
4. #27013 — object funds sufficiency check
5. #27014 — native + protocol flag that requests the retry

## Test plan

The check is inert in this PR: the consumer native and the protocol flag arrive in #27014, so `check_object_funds_sufficiency` is not reached during execution yet. Behavioral coverage — sufficient/insufficient/retry outcomes and cumulative multi-withdrawal accounting — lands with the native that calls it in #27014.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-types/src/storage/mod.rs
+++ b/crates/sui-types/src/storage/mod.rs
@@ -30,6 +30,7 @@ use itertools::Itertools;
 use move_binary_format::CompiledModule;
 use move_core_types::language_storage::{ModuleId, TypeTag};
 use move_core_types::resolver::SerializedPackage;
+use move_core_types::u256::U256;
 pub use object_store_trait::ObjectStore;
 pub use read_store::BalanceInfo;
 pub use read_store::BalanceIterator;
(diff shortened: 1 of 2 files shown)


lxfind:
Merged into #27014 — the funds-sufficiency resolver method and the native that exercises it are now a single PR.

PR #27014, opened 2026-06-19
[5/7] Check object funds withdraw sufficiency inside the Move VM

lxfind:
## Description

Top of the stack. Checks object-funds-withdraw sufficiency inside the Move VM instead of post-execution, behind a protocol flag, and wires the in-execution retry signal through to the authority. With this PR the object-funds path works end-to-end under the new flow (verified by the object-balance stress test and the checker integration tests, enabled at protocol version 128 on devnet/testnet).

Adds:
- `RuntimeObjectResolver::check_object_funds_sufficiency`, implemented by `TemporaryStore`. It gates on the accumulator root being available at the version this transaction requires (via `is_system_object_available`, #27020) — erroring if it has no assigned version, recording a retry (`ExecutionRetryError::ObjectFundsNotSettled`) and returning `Unknown` if it has not caught up — then compares the **net** withdrawn from the account so far in this transaction (running withdraw limits minus deposits back to the same account, #27022's unsettled reader discounting other in-flight transactions in the commit) against the settled balance.
- a `check_sufficient_object_funds` native called by `funds_accumulator::withdraw_from_object`: `Sufficient` proceeds, `Insufficient` aborts the transaction inside the VM (`E_OBJECT_FUNDS_INSUFFICIENT`), `Unknown` unwinds the VM so the transaction is retried (the temporary store has already recorded the request).
- the `check_object_funds_withdraw_in_execution` protocol feature flag (enabled on devnet and testnet at protocol version 128).

Authority wiring:
- `AuthorityState::execute_transaction_to_effects` propagates the executor's `Result<_, ExecutionRetryError>`. On `ObjectFundsNotSettled` no effects were produced; the authority asks `ObjectFundsChecker::reenqueue_after_settlement` to wait for the accumulator root to settle and then re-enqueue.
- On a successful in-execution withdrawal the authority records the withdrawals as unsettled (`record_object_funds_withdraws`) so later transactions in the same consensus commit account for them.
- The `ObjectFundsChecker` is now initialized on every node that runs the in-execution check (not only validators), so checkpoint execution / fullnodes can wait-and-reschedule on a retry rather than stalling.
- When the flag is on, the post-execution `should_commit_object_funds_withdraws` path is bypassed (the VM is authoritative).

### Known gap (TODO)

The availability gate (root reaching the required version) does not by itself guarantee an individual account has settled to it; revisit with a per-account settlement signal.

## Stack
3. #27020 — generic `is_system_object_available` + version-map wiring
4. #27022 — unsettled object funds reader threading
5. #27014 — object funds sufficiency check + native + protocol flag + retry/record wiring (this PR)

## Test plan

- `object_balance_tests::test_object_balance_withdraw_stress` (e2e simtest) is updated to enable `check_object_funds_withdraw_in_execution`. It stresses 7 concurrent threads withdrawing from owned/party/shared vault objects with random amounts (some expected to fail on insufficient balance), exercising the in-VM sufficiency check, the unsettled-withdrawal accounting across a consensus commit, and the retry/settle/re-enqueue path on both validators and the fullnode (checkpoint execution). Passes across multiple simulator seeds.
- `object_funds_checker` integration tests (`test_object_withdraw_multiple_withdraws`, `test_object_withdraw_and_deposit_same_transaction`, `test_object_withdraw_basic_flow`) run at v128 with the flag on and now assert the new in-execution behaviour: insufficient object funds abort inside the VM (`MoveAbort` `E_OBJECT_FUNDS_INSUFFICIENT`) rather than retry→`InsufficientFundsForWithdraw`, and a withdraw-then-deposit of equal amount nets out and succeeds.

## Release notes

- [x] Protocol: adds feature flag `check_object_funds_withdraw_in_execution`, enabled at protocol version 128 on devnet and testnet. No mainnet impact and no user action required.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


lxfind:

PR #27015, opened 2026-06-19
Drop vestigial ParentSync impl from latest TemporaryStore

lxfind:
## Description

The latest `TemporaryStore`'s `ParentSync` impl was dead weight — its only method was `unreachable!()`, since `get_latest_parent_entry_ref_deprecated` is only ever called by the v0 adapter. It was forced onto `TemporaryStore` purely via the `ExecutionState: StorageView` supertrait chain (`StorageView: Storage + ParentSync + ChildObjectResolver`).

This decouples the latest adapter's `ExecutionState` from `StorageView`, requiring `Storage + ChildObjectResolver` directly instead, and drops the `ParentSync` impl. The shared `StorageView` in `sui-types` is untouched, so v0–v3 are unaffected.

## Test plan

No behavioral change — the removed impl was unreachable and the new trait bound is equivalent minus `ParentSync`. Covered by existing execution tests.

## Release notes

Check each box that your change affects. If none of the boxes relate to your changes, release notes aren't required. For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates.

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/sui-execution/latest/sui-adapter/src/execution_value.rs
+++ b/sui-execution/latest/sui-adapter/src/execution_value.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use sui_types::storage::{BackingPackageStore, ChildObjectResolver, StorageView};
+use sui_types::storage::{BackingPackageStore, ChildObjectResolver, Storage};
 
 pub trait SuiResolver: BackingPackageStore {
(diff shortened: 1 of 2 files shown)


dariorussi (approved):
good cleanup

PR #27018, opened 2026-06-19
graphql: generalize Address.asTransactionObject

amnn:
Promote this field to the `IAddressable` interface, so that it can be called directly on any entity that has an address.

## Test plan

Updated E2E test:

```
$ cargo nextest run -p sui-indexer-alt-e2e-tests -- as_transaction_object
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Introduce `IAddressable.asTransactionObject` as a generalization of `Address.asTransactionObject`.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/move_object.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/move_object.rs
@@ -15,6 +15,7 @@ use tokio::sync::OnceCell;
 
 use crate::api::scalars::base64::Base64;
 use crate::api::scalars::big_int::BigInt;
+use crate::api::scalars::digest::Digest;
 use crate::api::scalars::id::Id;
 use crate::api::scalars::sui_address::SuiAddress;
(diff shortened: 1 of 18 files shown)


crates/sui-indexer-alt-graphql/src/api/types/move_object.rs
```diff
@@ -174,6 +176,24 @@ impl MoveObject {
         DynamicField::from_move_object(self, ctx).await
     }
 
+    /// How this address (interpreted as an object ID) was referenced by a specific transaction.
```
tpham-mysten: nit: MoveObject is an object itself, so "this address" is a bit confusing. Maybe "How this object..." ?

Same for CoinMetadata, DynamicField, MovePackage, and Object?

tpham-mysten (approved):
Looking great! Just a nit in docs

PR #27019, opened 2026-06-19
[2/5] Rename ChildObjectResolver to RuntimeObjectResolver (and resolver vars)

lxfind:
## Description

Mechanical rename of the `ChildObjectResolver` trait (and every reference, including `child_object_resolver` variables/parameters → `runtime_object_resolver`) to `RuntimeObjectResolver`. The trait is about to gain checks that resolve more than child objects during execution — system-object availability and object-funds sufficiency — so "child object" no longer describes it. No behavior change.

## Stack

1/5 #27003 — `ExecutionRetryError` plumbing from execution to authority
2/5 #27019 — rename `ChildObjectResolver` → `RuntimeObjectResolver`
3/5 #27036 — waitable system object framework
4/5 #27037 — record read system objects in effects
5/5 #27038 — check object funds withdraw sufficiency in the Move VM

## Test plan

Rename only — exercised by the existing test suites across the renamed crates.

## Release notes

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #27020, opened 2026-06-19
[3/7] Add is_system_object_available to RuntimeObjectResolver

lxfind:
## Description

Adds `RuntimeObjectResolver::is_system_object_available(object_id)` and threads a `system_object_versions: BTreeMap<ObjectID, SequenceNumber>` (system object → the version this transaction may read) into `TemporaryStore`, wired from `authority.rs` through the `Executor` trait down to `TemporaryStore::new`. The map is built in `prepare_certificate` from `execution_env.assigned_versions.accumulator_version`, so it currently holds the accumulator root at the version the transaction was sequenced against (when present).

`is_system_object_available` reports whether the object's latest committed version has reached the required version. Every system object read during execution must have an assigned version in the map: querying one that is absent is an invariant violation (the transaction was not sequenced against it) and the check **errors** rather than treating the object as unconstrained.

Generic infrastructure for the in-execution object-funds check (consumed in #27013). The frozen execution versions (v0–v3) accept and ignore the new parameter; `dev_inspect` and genesis pass an empty map.

## Stack
3. #27020 — generic `is_system_object_available` + version-map wiring
4. #27013 — object funds sufficiency check
5. #27014 — native + protocol flag that requests the retry

## Test plan

The availability check has no execution consumer in this PR (that arrives in #27013/#27014), but the version map is now populated for every certificate, so the plumbing is exercised by existing execution tests across `sui-core` and the adapter. The invariant-error path is reached only when a resolver queries a system object absent from the map; behavioral coverage of the funds-check outcomes lands with the consumer.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


lxfind:

PR #27022, opened 2026-06-20
[4/7] Thread unsettled object funds reader into the temporary store

lxfind:
## Description

Inserts the read-side plumbing for unsettled object-funds withdrawals ahead of the in-execution funds check (#27014).

Adds a `UnsettledObjectFundsRead` trait (one method: total amount withdrawn from an account at an accumulator version that has executed but not yet settled), implemented by `ObjectFundsChecker`, and threads an `Option<&dyn UnsettledObjectFundsRead>` through the execution layer (`Executor` trait → `execute_transaction_to_effects` → `TemporaryStore::new`) so execution can consult it. The authority passes the loaded `ObjectFundsChecker`; dev-inspect, genesis, and replay pass `None`. Frozen execution versions v0–v3 accept and ignore the parameter.

No behavior change on its own: the threaded reader is exposed via `TemporaryStore::unsettled_object_funds()` but not consulted until #27014 subtracts unsettled withdrawals in the object-funds sufficiency check.

## Stack
1. #27003 — signal out-of-band transaction retry from a native (base)
2. #27019 — rename `ChildObjectResolver` → `RuntimeObjectResolver`
3. #27020 — generic `is_system_object_available` + version-map wiring
4. #27022 — unsettled object funds reader threading (this PR)
5. #27014 — object funds sufficiency check + native + protocol flag + retry wiring

## Test plan

Pure plumbing; exercised structurally by existing execution tests across the threaded crates. The reader's effect on the sufficiency decision is covered end-to-end by the object-balance stress test enabled in #27014.

## Release notes

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #27023, opened 2026-06-21
[6/7] Generalize accumulator version to a system-object version map

lxfind:
## Description

Generalizes the version-passing framework that feeds system-object versions into execution. `AssignedVersions` carried a dedicated `accumulator_version: Option<SequenceNumber>` field; this replaces it with a generic `system_object_versions: BTreeMap<ObjectID, SequenceNumber>` so more system objects can be threaded over time (and so checkpoint replay can carry their versions, in the follow-up). Accumulator-specific consumers read the accumulator root version through an `accumulator_version()` accessor.

Pure refactor — no behavior change. The map currently holds at most the accumulator root version, exactly as before. A `BTreeMap` (not `HashMap`) is used to match the downstream temporary-store map and keep iteration deterministic.

## Stack
1. #27003, 2. #27019, 3. #27020, 4. #27022, 5. #27014 — object-funds-in-VM stack
6. this PR — generic system-object version map
7. (next) — record read system objects in effects for checkpoint replay

## Test plan

Covered by existing `shared_object_version_manager` assignment tests and the `object_funds_checker` suite (22 tests), which exercise version assignment and the accumulator-version reads through the new accessor.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/accumulators/object_funds_checker/mod.rs
+++ b/crates/sui-core/src/accumulators/object_funds_checker/mod.rs
@@ -188,7 +188,8 @@ impl ObjectFundsChecker {
         // for the epoch, and every production path that produces such a tx also
         // assigns an accumulator version. The `None` paths (accumulator-disabled
         // epoch, end-of-epoch tx) never produce withdraws and so never reach here.
-        let Some(accumulator_version) = execution_env.assigned_versions.accumulator_version else {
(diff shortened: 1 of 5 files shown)


lxfind:

PR #27024, opened 2026-06-21
[7/7] Record read system objects in effects

lxfind:
## Description

Records the system objects a transaction reads during execution (currently the accumulator root, read by the in-execution object-funds check) into its effects, so the read can be reproduced when re-executing from a checkpoint.

`TemporaryStore` tracks each system object it reads, at the version (and digest) it required — taken at `required_version`, not the latest, so the recorded value is deterministic across nodes. These are emitted into `unchanged_consensus_objects` as `ReadOnlyRoot` entries (reusing the existing variant — a read system object is, externally, a read-only consensus object), deduped against objects already present as changed or unchanged.

Because a read system object is not a declared/sequenced input (like `PerEpochConfig`), the replay-time version assignment (`assign_versions_from_effects`) is adjusted to skip consensus objects that have no entry in the transaction's declared input map, rather than asserting every input-consensus object is declared. The accumulator root version continues to flow into `AssignedVersions` separately, so replay reproduces the read.

## Stack
1. #27003, 2. #27019, 3. #27020, 4. #27022, 5. #27014 — object-funds-in-VM stack
6. #27023 — generic system-object version map
7. this PR — record read system objects in effects

## Test plan

Covered end-to-end by `object_balance_tests::test_object_balance_withdraw_stress` (e2e simtest): it executes object-funds withdrawals (which read the accumulator root), so every such transaction now records the root in its effects, and the recorded effects are re-produced bit-for-bit during checkpoint execution on the fullnode and validators (the simulator's fork detection would fail on any nondeterminism). Also exercised by the `effects` and `shared_object_version_manager` unit suites.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


lxfind:

PR #27026, opened 2026-06-22
[telemetry-subscribers] add RUST_LOG_TAILS for per-target log file tails

evan-wall-mysten:
## Description

Adds `RUST_LOG_TAILS` to `telemetry-subscribers` (with a `with_log_tail` builder / `log_tails` config) that *additionally* writes individual tracing targets to dedicated files, on top of the normal log output. A "tail" is an extra file sink: matching events are **duplicated** to the file, not moved off the console/JSON stream.

**Syntax** — comma-separated `target=file` entries (mirroring `RUST_LOG`'s `target=level`, but assigning a file instead of a level):

- `=file` is a daily-rolling file (like `RUST_LOG_FILE`) and is **mandatory** — `RUST_LOG_TAILS` only adds file sinks, it never routes to the stream.
- The output **format is inferred from the file extension**: `.jsonl` / `.ndjson` (case-insensitive) → newline-delimited JSON; everything else (including `.json` and `.log`) → text.

| Entry | Result |
| --- | --- |
| `graphql_request=/var/log/sui/gql.log` | `graphql_request` → text in that file |
| `graphql_request=/var/log/sui/gql.jsonl` | `graphql_request` → NDJSON in that file (replay corpus) |
| `a=/x.jsonl,b=/y.log` | two tails, each to its own file |

Behavior:

- Tails are **additional** sinks — the events still appear on the normal stream as well, **including when `RUST_LOG_JSON` (global JSON output) is set**.
- Each tail honors **`RUST_LOG`'s per-target level** (its layer filter is the target restriction ANDed with an `EnvFilter`): e.g. with `RUST_LOG=warn` a `bar` tail captures only `warn`+, while `RUST_LOG=warn,bar=debug` also captures `bar` debug. Targets are matched by prefix (like `RUST_LOG`).
- A malformed entry (missing target or `=file`) or a file whose directory cannot be created **hard-fails `init()`** rather than silently degrading.

Example — capture `graphql_request` as a JSON replay corpus in its own file while normal logs stay human-readable:

```
RUST_LOG="info,graphql_request=trace" RUST_LOG_TAILS="graphql_request=/var/log/sui/graphql_request.jsonl"
```

## Test plan

Added unit tests (entry parsing, extension→format inference, malformed-entry panics, builder). `cargo test -p telemetry-subscribers`, `cargo xclippy`, and `cargo fmt -- --check` all pass.

---

## Release notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): New optional `RUST_LOG_TAILS` env var additionally writes individual tracing targets to dedicated files (format inferred from the file extension — `.jsonl`/`.ndjson` for JSON, otherwise text), leaving normal log output unchanged. Backward-compatible — unset by default, so log output is unchanged and no action is required.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [x] Indexing Framework: The same `RUST_LOG_TAILS` env var is available to indexer binaries (e.g. `sui-indexer-alt-graphql`). Set it alongside `RUST_LOG` (e.g. `RUST_LOG="info,graphql_request=trace"` with `RUST_LOG_TAILS="graphql_request=/var/log/sui/graphql_request.jsonl"`) to capture that target as NDJSON for replay while normal logs stay readable. Opt-in; no action required otherwise.

### DIFF
--- a/crates/telemetry-subscribers/src/lib.rs
+++ b/crates/telemetry-subscribers/src/lib.rs
@@ -15,7 +15,8 @@ use opentelemetry_sdk::{
     trace::{BatchSpanProcessor, ShouldSample, TracerProvider},
 };
 use span_latency_prom::PrometheusSpanLatencyLayer;
-use std::path::PathBuf;
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
 use std::time::Duration;
 use std::{
     env,
@@ -29,7 +30,8 @@ use tracing::{Level, error, info};
 use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
 use tracing_error::ErrorLayer;
 use tracing_subscriber::{
-    EnvFilter, Layer, Registry, filter,
+    EnvFilter, Layer, Registry,
+    filter::{self, FilterExt},
     fmt::{self, format::Pretty},
     layer::SubscriberExt,
     reload,
@@ -123,9 +125,44 @@ impl<S: tracing::Subscriber> Layer<S> for GlobalLevelFilter {
 /// Alias for a type-erased error type.
 pub type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
+/// Output format for a per-target log route (see [`TelemetryConfig::log_tails`]). Inferred
+/// from the route's file extension.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
+pub enum LogFormat {
+    Json,
+    #[default]
+    Text,
+}
+
+impl LogFormat {
+    /// Infer the output format from a log file's extension: the common newline-delimited
+    /// JSON extensions `jsonl` and `ndjson` (case-insensitive) select [`LogFormat::Json`];
+    /// every other extension (including `.json` and `.log`) and none default to
+    /// [`LogFormat::Text`].
+    fn from_path(path: &Path) -> Self {
+        if let Some(ext) = path.extension().and_then(|ext| ext.to_str())
+            && (ext.eq_ignore_ascii_case("jsonl") || ext.eq_ignore_ascii_case("ndjson"))
+        {
+            LogFormat::Json
+        } else {
+            LogFormat::Text
+        }
+    }
+}
+
+/// A per-target output route: *additionally* write `target`'s logs to `file`. Configured via
+/// the `RUST_LOG_TAILS` env var, e.g. `graphql_request=/var/log/sui/graphql_request.jsonl`.
+#[derive(Clone, Debug)]
+pub struct LogTail {
+    pub target: String,
+    pub format: LogFormat,
+    pub file: PathBuf,
+}
+
 /// Configuration for different logging/tracing options
 /// ===
 /// - json_log_output: Output JSON logs to stdout only.
+/// - log_tails: Additional per-target log files via RUST_LOG_TAILS.
 /// - log_file: If defined, write output to a file starting with this name, ex app.log
 /// - log_level: error/warn/info/debug/trace, defaults to info
 #[derive(Default, Clone, Debug)]
@@ -155,6 +192,10 @@ pub struct TelemetryConfig {
     pub trace_target: Option<Vec<String>>,
     /// Print unadorned logs from the target on standard error if this is a tty
     pub user_info_target: Vec<String>,
+    /// Per-target output routing — one [`LogTail`] per entry — populated from the
+    /// `RUST_LOG_TAILS` env var. See [`TelemetryConfig::with_log_tail`] for how each tail
+    /// is emitted.
+    pub log_tails: Vec<LogTail>,
     // Sets the subscriber created by this config to be the global default. Defaults to true if not set.
     pub set_global_default: bool,
     // Add error layer to capture trace spans. Defaults to false if not set.
@@ -167,6 +208,9 @@ pub struct TelemetryConfig {
 #[allow(dead_code)]
 pub struct TelemetryGuards {
     worker_guard: WorkerGuard,
+    /// Flush guards for any per-target log files (`RUST_LOG_TAILS`). Held so the
+    /// background writer threads keep flushing until the guards are dropped.
+    file_guards: Vec<WorkerGuard>,
     provider: Option<TracerProvider>,
     subscriber: Option<DefaultGuard>,
 }
@@ -175,6 +219,7 @@ impl TelemetryGuards {
     fn new(
         config: TelemetryConfig,
         worker_guard: WorkerGuard,
+        file_guards: Vec<WorkerGuard>,
         provider: Option<TracerProvider>,
         subscriber: Option<DefaultGuard>,
     ) -> Self {
@@ -184,6 +229,7 @@ impl TelemetryGuards {
         }
         Self {
             worker_guard,
+            file_guards,
             provider,
             subscriber,
         }
@@ -302,6 +348,24 @@ fn get_output(log_file: Option<String>) -> (NonBlocking, WorkerGuard) {
     }
 }
 
+/// Open a per-target log file (`RUST_LOG_TAILS`), pre-creating its parent directory.
+/// Panics `init()` if the directory cannot be created, rather than silently dropping the
+/// target's file sink.
+fn open_tail_file(path: &Path) -> (NonBlocking, WorkerGuard) {
+    if let Some(parent) = path.parent()
+        && !parent.as_os_str().is_empty()
+    {
+        std::fs::create_dir_all(parent).unwrap_or_else(|e| {
+            panic!(
+                "telemetry: cannot create directory for RUST_LOG_TAILS file {}: {e}",
+                path.display()
+            )
+        });
+    }
+    let file_appender = tracing_appender::rolling::daily("", path);
+    tracing_appender::non_blocking(file_appender)
+}
+
 // NOTE: this function is copied from tracing's panic_hook example
 fn set_panic_hook(crash_on_panic: bool) {
     let default_panic_handler = std::panic::take_hook();
@@ -361,6 +425,33 @@ pub fn get_global_telemetry_config() -> Option<TelemetryConfig> {
     global_config.clone()
 }
 
+fn parse_log_tails(spec: &str) -> Vec<LogTail> {
+    spec.split(',')
+        .map(str::trim)
+        .filter(|entry| !entry.is_empty())
+        .map(parse_log_tail_entry)
+        .collect()
+}
+
+fn parse_log_tail_entry(entry: &str) -> LogTail {
+    if let Some((target, file)) = entry.split_once('=') {
+        let (target, file) = (target.trim(), file.trim());
+        if !target.is_empty() && !file.is_empty() {
+            let file = PathBuf::from(file);
+            let format = LogFormat::from_path(&file);
+            return LogTail {
+                target: target.to_owned(),
+                format,
+                file,
+            };
+        }
+    }
+    panic!(
+        "telemetry: invalid RUST_LOG_TAILS entry {entry:?}: a `=file` path is \
+         required, e.g. `graphql_request=/var/log/sui/graphql_request.jsonl`"
+    )
+}
+
 impl TelemetryConfig {
     pub fn new() -> Self {
         Self {
@@ -377,6 +468,7 @@ impl TelemetryConfig {
             sample_rate: 1.0,
             trace_target: None,
             user_info_target: Vec::new(),
+            log_tails: Vec::new(),
             set_global_default: true,
             enable_error_layer: false,
             enable_test_layer: false,
@@ -434,6 +526,19 @@ impl TelemetryConfig {
         self
     }
 
+    /// Adds a `RUST_LOG_TAILS`-style routing rule: *additionally* write `target`'s logs to
+    /// `file`, in a format inferred from its extension.
+    pub fn with_log_tail(mut self, target: &str, file: impl Into<PathBuf>) -> Self {
+        let file = file.into();
+        let format = LogFormat::from_path(&file);
+        self.log_tails.push(LogTail {
+            target: target.to_owned(),
+            format,
+            file,
+        });
+        self
+    }
+
     pub fn with_set_global_default(mut self, set_global_default: bool) -> Self {
         self.set_global_default = set_global_default;
         self
@@ -460,6 +565,10 @@ impl TelemetryConfig {
 
         if env::var("RUST_LOG_JSON").is_ok() {
             self.json_log_output = true;
+        }
+
+        if let Ok(spec) = env::var("RUST_LOG_TAILS") {
+            self.log_tails = parse_log_tails(&spec);
         }
 
         if env::var("TOKIO_CONSOLE").is_ok() {
@@ -502,8 +611,8 @@ impl TelemetryConfig {
                 directives.push_str(&format!(",{}=trace", target));
             }
         }
-        let env_filter =
-            EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new(directives));
+        let env_filter = EnvFilter::try_from_default_env()
+            .unwrap_or_else(|_| EnvFilter::new(directives.clone()));
         // When the test layer is enabled, it accepts events at any level without a
         // per-layer filter, so global event-level filtering would hide events the
         // test wants to capture. Disable it by pinning max_event_level to TRACE
@@ -629,6 +738,8 @@ impl TelemetryConfig {
         }
 
         let (nb_output, worker_guard) = get_output(config.log_file.clone());
+        // Flush guards for any per-target files opened below; kept alive in TelemetryGuards.
+        let mut file_guards: Vec<WorkerGuard> = Vec::new();
         if config.json_log_output {
             // Output to file or to stderr in a newline-delimited JSON format
             let json_layer = fmt::layer()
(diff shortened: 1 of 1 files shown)


amnn (commented):
This is currently feeling quite ad-hoc to me, especially the way it requires json log output to be disabled in order for json log target to be applicable, and the fact that enabling this causes logs to disappear from the normal format (coupled with the fact that targets are prefix matched, which means you may end up inadvertently capturing logs you didn't mean to).

Instead, can we offer some kind of orthogonal control surface for this? I.e. I'd like to be able to emit logs matching certain targets (separate from the normal filter) to a file (separate from log_file) in addition to whatever normal logging configuration there is.

evan-wall-mysten:
> This is currently feeling quite ad-hoc to me, especially the way it requires json log output to be disabled in order for json log target to be applicable, and the fact that enabling this causes logs to disappear from the normal format (coupled with the fact that targets are prefix matched, which means you may end up inadvertently capturing logs you didn't mean to).
> 
> Instead, can we offer some kind of orthogonal control surface for this? I.e. I'd like to be able to emit logs matching certain targets (separate from the normal filter) to a file (separate from log_file) in addition to whatever normal logging configuration there is.

I changed the env var to be less focused on JSON by allowing it to set either the format, the file, or both per target.

Should the logs in this new file be accessed through Grafana like the existing logs or copied out of the k8s pod? I was testing doing it either way in https://github.com/MystenLabs/sui-operations/pull/8197/, but logging to a new file would require either ingesting that file into Grafana (would be a separate change) or getting it from the k8s pod.

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -123,9 +125,31 @@ impl<S: tracing::Subscriber> Layer<S> for GlobalLevelFilter {
...
+}
+
+/// A per-target output route: emit `target`'s logs in `format`, to `file` (or the
+/// normal output stream when `file` is `None`). Configured via the `RUST_LOG_TARGETS`
+/// env var, e.g. `graphql_request=json@/var/log/sui/graphql_request.log`.
+#[derive(Clone, Debug)]
+pub struct LogTarget {
+    pub target: String,
+    pub format: LogFormat,
+    /// Daily-rolling file (like `RUST_LOG_FILE`); `None` routes to the normal stream.
```
wlmyng: this seems like an implementation detail?

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -155,6 +179,12 @@ pub struct TelemetryConfig {
     pub trace_target: Option<Vec<String>>,
     /// Print unadorned logs from the target on standard error if this is a tty
     pub user_info_target: Vec<String>,
+    /// Per-target output routing (the `RUST_LOG_TARGETS` env var). Each entry sends a
+    /// tracing target's logs to a chosen format (JSON or text) and/or a dedicated file,
+    /// while everything else stays on the normal human-readable stream. Emission is
+    /// still governed by the env filter (`RUST_LOG`); this only changes where/how the
+    /// matching targets are written. Ignored when `json_log_output` is set.
```
wlmyng: for backward compatibility? would it not work if we allow json_log_output, and then still the log_targets carve outs?

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -302,6 +337,20 @@ fn get_output(log_file: Option<String>) -> (NonBlocking, WorkerGuard) {
     }
 }
 
+/// Open a per-target log file (`RUST_LOG_TARGETS`), pre-creating its parent directory.
+/// Panics `init()` if the directory cannot be created, rather than silently routing the
+/// target to the normal output stream.
+fn open_target_file(path: &str) -> (NonBlocking, WorkerGuard) {
```
wlmyng: the code reads straightforward enough that the doc comment may not be needed

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -361,6 +410,59 @@ pub fn get_global_telemetry_config() -> Option<TelemetryConfig> {
     global_config.clone()
 }
 
+/// True if `target` starts with one of `prefixes`. This is the raw-prefix rule that
+/// `Targets`/`EnvFilter` (and therefore `RUST_LOG`) use to match target directives, so
+/// e.g. `graphql` matches `graphql_request`. It must stay identical to the per-target
+/// include `Targets` filters: a routed event is then excluded from the human-readable
+/// layer exactly when it is emitted by its own layer, so each event appears once.
+fn target_matches(prefixes: &[String], target: &str) -> bool {
```
wlmyng: "Must stay identical..." seems like an invariant that is better off at the callsite

How about just

```
  /// Prefix match using the same rule as `Targets`/`EnvFilter`, so `graphql` matches
  /// `graphql_request`.
```

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -361,6 +410,59 @@ pub fn get_global_telemetry_config() -> Option<TelemetryConfig> {
...
+        .any(|prefix| target.starts_with(prefix.as_str()))
+}
+
+/// Parse the `RUST_LOG_TARGETS` value: comma-separated `target[=format][@file]` entries.
+/// `format` is `text` (the default) or `json`; `@file` routes the target to that file
+/// (daily-rolling, like `RUST_LOG_FILE`) instead of the normal output stream.
+///
+/// Targets are matched by prefix (like `RUST_LOG`), so list disjoint targets — if two
+/// entries share a prefix, the more general one's format/sink wins for the overlap. A
+/// `@file`'s directory is created if missing; an uncreatable path panics `init()`.
```
wlmyng: >  A `@file`'s directory is created if missing; an uncreatable path panics `init()`.

Seems irrelevant to `parse_log_targets`?

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -361,6 +410,59 @@ pub fn get_global_telemetry_config() -> Option<TelemetryConfig> {
...
+fn parse_log_targets(spec: &str) -> Vec<LogTarget> {
+    spec.split(',')
+        .map(str::trim)
+        .filter(|entry| !entry.is_empty())
+        .map(parse_log_target_entry)
+        .collect()
+}
+
+fn parse_log_target_entry(entry: &str) -> LogTarget {
```
wlmyng:
```
  /// Parse one `RUST_LOG_TARGETS` entry: `target[=format][@file]`. `format` is `text`
  /// (default) or `json`; unknown formats silently fall back to text rather than
  /// failing init. `@file` routes the target to that path instead of the default sink.
```

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -434,6 +537,20 @@ impl TelemetryConfig {
         self
     }
 
+    /// Adds a `RUST_LOG_TARGETS`-style routing rule: emit `target`'s logs in `format`
+    /// (JSON or text), to `file` (daily-rolling, like `RUST_LOG_FILE`) or the normal
```
wlmyng: `file` is just an arbitrary file path, no? I think the method's code is descriptive but the doc comment makes it confusing, as it includes functionality beyond what the method itself does

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -640,14 +768,92 @@ impl TelemetryConfig {
...
+                // and/or file. Targets reformatted as JSON on the stream, or sent to a
+                // file, are excluded from the human-readable layer so each event is
+                // emitted exactly once. The env filter above still governs whether these
+                // events fire at all (so a target must also be enabled in `RUST_LOG`,
+                // since the global level filter gates callsites by the env filter's max
+                // level); this only changes where/how matching events are written.
+                //
+                // Text-on-stream targets need no special layer — they already appear on
+                // the human-readable layer — so they are not excluded and get no layer.
+                let excluded: Vec<String> = config
```
wlmyng: so `RUST_LOG` decides what fires, and `RUST_LOG_TARGETS` decide where they get written to

Could we not treat the 2 + 1 incoming branch today as just the same primitive?

```
  - RUST_LOG_JSON → implicit LogTarget { target: "" /* matches all */, format: Json, file: None }
  - default text-on-stream → implicit LogTarget { target: "", format: Text, file: None }
  - log_file: Some(p) → that implicit entry's file = Some(p)
```

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -640,14 +768,92 @@ impl TelemetryConfig {
...
+                        // Every file path was opened (or panicked) in the loop above, so the
+                        // writer is guaranteed present; the default is unreachable.
+                        Some(path) => file_writers
+                            .get(path)
+                            .cloned()
+                            .unwrap_or_else(|| nb_output.clone()),
+                        None => nb_output.clone(),
+                    };
+                    let target_filter = filter::Targets::new()
+                        .with_target(log_target.target.clone(), LevelFilter::TRACE);
```
wlmyng: should we not add `log_filter` as well?  so the per-target layer also honors  RUST_LOG's level rules for that target. For example if you specify `RUST_LOG=trace,graphql_request=warn`

amnn (commented):
The telemetry subscriber is already pretty complex, with lots of flags interacting with each other. If we can avoid introducing new complexity (interactions between flags or multiple ways to do things) then we should.

In that vein, I'd suggest introducing an environment variable for tailing logs, filtered by target, to a file, optionally modifying the format. Key differences from what we have here:

- Just duplicate the logs (don't eliminate logs from the normal stream because they are being tailed elsewhere). Given filters are prefix matches, it can be difficult to predict what they pick up and it's incorrect to assume that the audience for the human-readable logs and the audience for the tailed logs are the same (in a way where they would care that the logs are duplicated).
- The file/path becomes mandatory -- we are always writing to a file, never to the default stream -- but the format is optional. This reduces overlap with other features (like `json_log_output`, so don't ignore this config if that flag is set), and also the potential to duplicate logs in an obvious way.
- I would also suggest parsing the path as a `PathBuf` rather than a `String`, and keying the format on the extension of that path (e.g. if it's `.jsonl` then you assume json formatting), to avoid having to invent new syntax.
- Name it something other than `RUST_LOG_TARGETS` -- a "target" is the label for the log and lots of configs we care about accept a target in this file. Perhaps `RUST_LOG_TAILS` or something?

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: Removed

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: Yeah I changed the target code to only affect its files. `json_log_output` still affects the main log and does not disable the target files.

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: Removed this.

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: Removed this.

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: Removed the parsing comments because the logic is simpler.

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: Removed.

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: Changed to `RUST_LOG_TAILS` which now does not affect the main log, but allows optionally duplicating targets to files in either format depending on the file extension.

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: Removed this comment.

crates/telemetry-subscribers/src/lib.rs
evan-wall-mysten: This is now handled with the `and`.

evan-wall-mysten:
> The telemetry subscriber is already pretty complex, with lots of flags interacting with each other. If we can avoid introducing new complexity (interactions between flags or multiple ways to do things) then we should.
> 
> In that vein, I'd suggest introducing an environment variable for tailing logs, filtered by target, to a file, optionally modifying the format. Key differences from what we have here:
> 
> * Just duplicate the logs (don't eliminate logs from the normal stream because they are being tailed elsewhere). Given filters are prefix matches, it can be difficult to predict what they pick up and it's incorrect to assume that the audience for the human-readable logs and the audience for the tailed logs are the same (in a way where they would care that the logs are duplicated).
> * The file/path becomes mandatory -- we are always writing to a file, never to the default stream -- but the format is optional. This reduces overlap with other features (like `json_log_output`, so don't ignore this config if that flag is set), and also the potential to duplicate logs in an obvious way.
> * I would also suggest parsing the path as a `PathBuf` rather than a `String`, and keying the format on the extension of that path (e.g. if it's `.jsonl` then you assume json formatting), to avoid having to invent new syntax.
> * Name it something other than `RUST_LOG_TARGETS` -- a "target" is the label for the log and lots of configs we care about accept a target in this file. Perhaps `RUST_LOG_TAILS` or something?

Updated the PR

> The telemetry subscriber is already pretty complex, with lots of flags interacting with each other. If we can avoid introducing new complexity (interactions between flags or multiple ways to do things) then we should.
> 
> In that vein, I'd suggest introducing an environment variable for tailing logs, filtered by target, to a file, optionally modifying the format. Key differences from what we have here:
> 
> * Just duplicate the logs (don't eliminate logs from the normal stream because they are being tailed elsewhere). Given filters are prefix matches, it can be difficult to predict what they pick up and it's incorrect to assume that the audience for the human-readable logs and the audience for the tailed logs are the same (in a way where they would care that the logs are duplicated).
> * The file/path becomes mandatory -- we are always writing to a file, never to the default stream -- but the format is optional. This reduces overlap with other features (like `json_log_output`, so don't ignore this config if that flag is set), and also the potential to duplicate logs in an obvious way.
> * I would also suggest parsing the path as a `PathBuf` rather than a `String`, and keying the format on the extension of that path (e.g. if it's `.jsonl` then you assume json formatting), to avoid having to invent new syntax.
> * Name it something other than `RUST_LOG_TARGETS` -- a "target" is the label for the log and lots of configs we care about accept a target in this file. Perhaps `RUST_LOG_TAILS` or something?

Updated the PR with these changes which ended up simplifying the code a bit.

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -123,9 +125,45 @@ impl<S: tracing::Subscriber> Layer<S> for GlobalLevelFilter {
...
+    fn from_path(path: &Path) -> Self {
+        match path.extension().and_then(|ext| ext.to_str()) {
+            Some(ext)
+                if ext.eq_ignore_ascii_case("jsonl") || ext.eq_ignore_ascii_case("ndjson") =>
+            {
+                LogFormat::Json
+            }
+            _ => LogFormat::Text,
+        }
+    }
```
amnn: nit (optional): `if let ... && ...` may be slightly nicer here? (less nesting, no need for an underscore pattern).

```suggestion
        if let Some(ext) = path.extension.and_then(|ext ext.to_str())
            && (ext.eq_ignore_ascii_case("jsonl") || ext.eq_ignore_ascii_case("ndjson"))
        {
            LogFormat::Json
        } else {
            LogFormat::Text
        }
    }
```

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -302,6 +349,21 @@ fn get_output(log_file: Option<String>) -> (NonBlocking, WorkerGuard) {
...
 
+/// Open a per-target log file (`RUST_LOG_TAILS`), pre-creating its parent directory.
+/// Panics `init()` if the directory cannot be created, rather than silently dropping the
+/// target's file sink.
+fn open_tail_file(path: &Path) -> (NonBlocking, WorkerGuard) {
+    if let Some(parent) = path.parent()
+        && !parent.as_os_str().is_empty()
+    {
+        std::fs::create_dir_all(parent).unwrap_or_else(|e| {
+            panic!("telemetry: cannot create directory for RUST_LOG_TAILS file {path:?}: {e}")
```
amnn:
```suggestion
            panic!("telemetry: cannot create directory for RUST_LOG_TAILS file {}: {e}", path.display())
```

crates/telemetry-subscribers/src/lib.rs
```diff
@@ -361,6 +423,33 @@ pub fn get_global_telemetry_config() -> Option<TelemetryConfig> {
...
+fn parse_log_tails(spec: &str) -> Vec<LogTail> {
+    spec.split(',')
+        .map(str::trim)
+        .filter(|entry| !entry.is_empty())
+        .map(parse_log_tail_entry)
+        .collect()
+}
+
+fn parse_log_tail_entry(entry: &str) -> LogTail {
+    if let Some((target, file)) = entry.split_once('@') {
```
amnn: Personally would just use `=` for this, as it reads similarly to the `RUST_LOG` config which assigns a level to each target filter (in this case you're assigning a file to each target filter).

```suggestion
    if let Some((target, file)) = entry.split_once('=') {
```

amnn (approved):
Looks good, thanks @evan-wall-mysten !

PR #27032, opened 2026-06-22
[adapter] Core unified linkage changes

tzakian:
## Description 

Adds unified (aka LUMPY) linkage as discussed (single linkage, publication linkages do not contribute unless there are a `init` functions present, upgrades do not currently contribute but will when we add support for `init` functions).

Currently this is just piggy-backing on protocol version 127, but that is incorrect. Once we are happy with the changes I'll create a protocol version, update snapshots (etc.) but until then figured it's best not to to keep this as clear as possible. 

## Test plan 

Added a number of additional tests (see new adapter transactional tests) around this and also ported over a couple tests from `sui-core` `move_package_upgrade_tests` that were affected to adapter transactional tests, and captured previous behavior as well as new expected behavior for them.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-adapter-transactional-tests/tests/entry_points/entry_fun_upgrade_removal.move
```diff
@@ -15,6 +15,13 @@ module test1::m;
 
 public fun bar() {}
 
+// Cannot be resolved under unified linkage
 //# programmable --sender A
```
tnowacki: Nit, if you flip these lines the comment reads better in the snap file

sui-execution/latest/sui-adapter/src/static_programmable_transactions/mod.rs
```diff
@@ -88,6 +88,9 @@ pub fn execute<Mode: ExecutionMode>(
         )
         .map_err(|e| (e, vec![]))?
     };
+    let txn =
+        linkage::refine_linkage::<Mode>(txn, &linkage_analysis, &package_store, protocol_config)
+            .map_err(|e| (e, vec![]))?;
```
tnowacki: This is `loading::Program -> loading::Program`? 
If so, let's move this into `loading::translate::transaction`

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -0,0 +1,212 @@
...
+        Command::Upgrade(_, _, _, _, _) => (),
+        Command::TransferObjects(_, _) | Command::SplitCoins(_, _) | Command::MergeCoins(_, _) => {}
+    };
+    Ok(())
+}
+
+/// Overwrite each `MoveCall`'s per-call linkage with the unified transaction-wide linkage (pass 2).
+/// Only `MoveCall`s carry an executable linkage; the other commands need no write-back.
+fn write_back_linkage<E: ExecutionErrorTrait>(
+    command: &mut Command,
```
tnowacki: nit, I'd move this to the end to just separate and group all the linkage steps and the write back step

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -0,0 +1,212 @@
...
+                    .all(|k| ptb_linkage.0.linkage.contains_key(k)),
+                "single linkage drops a package that the per-call linkage of MoveCall had resolved"
+            );
+            move_call.function.linkage = ptb_linkage.clone();
+        }
+        Command::TransferObjects(_, _)
+        | Command::SplitCoins(_, _)
+        | Command::MergeCoins(_, _)
+        | Command::MakeMoveVec(_, _)
+        | Command::Publish(_, _, _)
```
tnowacki: What about the linkage for publish and it's init function?

tnowacki (commented):
We should have some tests that verify we end up with the versions we  expect. At least some simple ones, especially with init. 

We could just have them write fields with a specific value per-version, and then view the object so that we know we got the right version at the end? (very worried about init)

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -0,0 +1,212 @@
...
+///
+/// Done in two passes:
+///   1. Fold every command's package and type-argument constraints into one `ResolutionTable`,
+///      unifying as we go (an error here means the commands cannot agree on a single set of
+///      package versions).
+///   2. Write the resulting unified linkage back into every `MoveCall`.
+///
+/// Because all calls end up sharing one linkage, every package version selection is consistent
+/// across the transaction.
+pub fn refine_to_single_linkage<E: ExecutionErrorTrait>(
```
cgswords, line 38: Do we want to meter linkage refinement? It's a non-trivial amount of work, and the existing metering doesn't cover it.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -0,0 +1,212 @@
...
+/// Replace each command's per-call linkage with a single linkage shared by the whole transaction.
+///
+/// Done in two passes:
+///   1. Fold every command's package and type-argument constraints into one `ResolutionTable`,
+///      unifying as we go (an error here means the commands cannot agree on a single set of
+///      package versions).
+///   2. Write the resulting unified linkage back into every `MoveCall`.
+///
+/// Because all calls end up sharing one linkage, every package version selection is consistent
+/// across the transaction.
```
cgswords, line 37: Can we add a doc-comment somewhere on `refine_to_single_linkage` explaining the callee package pinning invariant? Something like: "the callee package is pinned `exact`, so unification can only converge on the loaded version; this is what makes the key-subset assertion in write-back sufficient."

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -0,0 +1,212 @@
...
+            // bug in how component constraints were folded together.
+            assert_invariant!(
+                previous_linkage
+                    .0
+                    .linkage
+                    .keys()
+                    .all(|k| ptb_linkage.0.linkage.contains_key(k)),
+                "single linkage drops a package that the per-call linkage of MoveCall had resolved"
+            );
+            move_call.function.linkage = ptb_linkage.clone();
```
cgswords: The key-subset check implies the length check because it's a set, right?

Also, maybe scope a borrow to avoid the clone:
```rust
{
    let previous = &move_call.function.linkage;
    assert_invariant!(previous.0.linkage.keys().all(|k| ptb_linkage.0.linkage.contains_key(k)), ...);
}
move_call.function.linkage = ptb_linkage.clone();
```

sui-execution/latest/sui-adapter/src/static_programmable_transactions/metering/translation_meter.rs
```diff
@@ -92,6 +93,22 @@ impl<'pc, 'gas> TranslationMeter<'pc, 'gas> {
         self.charge(amount)
     }
 
+    pub fn charge_package_load<E: ExecutionErrorTrait>(
+        &mut self,
+        payload: &PackagePayload,
+    ) -> Result<(), E> {
+        match payload {
+            PackagePayload::Serialized(_) => {
+                // Serialized during execution, and will be charged at that point
```
cgswords: Suggest:
```rust
// Payload stays serialized; deserialization (and charge) happen at execution time.
```

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -0,0 +1,212 @@
...
+    self_resolution_fn: fn(&VerifiedPackage) -> Option<VersionConstraint>,
+    dep_resolution_fn: fn(&VerifiedPackage) -> Option<VersionConstraint>,
+) -> Result<(), E> {
+    let pkg = get_package(object_id, store)?;
+    let transitive_deps = resolution_table
+        .config
+        .linkage_table(&pkg)
+        .into_values()
+        .map(ObjectID::from);
+    add_and_unify(object_id, store, resolution_table, self_resolution_fn)?;
```
cgswords, line 184: `add_package` calls `get_package` just to walk the linkage table, and then `add_and_unify` does the same lookup again. Do we care to change the interface to avoid the double-fetch?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/mod.rs
```diff
@@ -5,3 +5,34 @@ pub mod analysis;
...
+    static_programmable_transactions::{
+        linkage::analysis::LinkageAnalyzer, loading::ast as loading,
+    },
+};
+use sui_protocol_config::ProtocolConfig;
+
+/// Refine the transaction's per-call linkages into a single, unified linkage for the whole
+/// transaction (when enabled by the protocol config).
+pub fn refine_linkage<Mode: ExecutionMode>(
+    mut txn: loading::Transaction,
```
cgswords, line 22: `&mut` here?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -0,0 +1,212 @@
...
+            //
+            // NB: We presuppose here that if there is a function with the name "init" in the
+            // modules being published, then it is the init function for the package. This holds
+            // because we presuppose that the sui-verifier `entry_points_verifier` has already run
+            // and verified that there is at most one function named "init" in the package and that
+            // it has the correct signature. Additionally, even if it has not, it will eventually
+            // run after this and raise an error if the "init" function does not have the correct
+            // signature.
+            //
+            // `modules` is guaranteed to be non-empty by the `deserialize_modules` function.
```
cgswords, line 95: Can we tighten this comment up?

crates/sui-protocol-config/src/lib.rs
```diff
@@ -5027,6 +5035,9 @@ impl ProtocolConfig {
                     }
 
                     cfg.feature_flags.timestamp_based_epoch_close = true;
+                    // TODO: Place this in the correct protocol version once we have a better idea
+                    // on rollout date.
+                    cfg.feature_flags.enable_unified_linkage = true;
```
cgswords: Add a TODO for the protocol version bump and snapshot regeneration plan? This should probably get done before landing, right?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
tzakian: Probably not at least at the moment, we charge before this (and after this) on things that contribute to it though. But once/if we add metering as more first-class part of PTB translation it would make sense.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
tzakian: > The key-subset check implies the length check because it's a set, right?


correct. Moved the len check to a debug assert below this to document this (+ amended the comment).

Good call on being able to remove the `clone` -- will fix :)

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
tzakian: Not really -- I think it's fine. These are all cached behind the scene and will be quite fast.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/mod.rs
tzakian: Thought about it, but I'd like to/like the interface of having it be a transaction -> transaction function (and is more in keeping with the rest of the passes).

crates/sui-protocol-config/src/lib.rs
tzakian: Will do that in this PR :)

tnowacki (approved):

PR #27035, opened 2026-06-23
[graphql] Migrate Transaction type to use gRPC cursor format

wlmyng:
## Description 

Split out `CursorToken` from sui-rpc-api into separate sui-rpc-cursor crate.

(And migrate `CursorToken` to proto from bcs: https://github.com/MystenLabs/sui/pull/27102)

Migrate `Transaction` cursor wire format to gRPC's CursorToken, wrapped in a new `ByteCursor`.

For now, Postgres path decodes the bytes to extract the position for SQL bounds. After the migration to bitmap scanning apis, the cursors will be pure pass through. Subscriptions will mint the same format directly.

## Test plan 

Migrated cursor-related tests on transactional test runner to e2e in Rust.

---

## Release notes

- [ ] Protocol: 
- [x] gRPC: Opaque pagination cursors returned by the scanning apis `list_transactions`, `list_events`, and `list_checkpoints` have a new internal encoding. Cursors remain opaque and clients should continue treating them as such. Users should restart pagination from the beginning if they encounter a cursor-related error.
- [ ] JSON-RPC: 
- [x] GraphQL: The opaque pagination cursor returned for a `Transaction` edge of the `TransactionConnection` have a new internal encoding. Cursors remain opaque and clients should continue treating them as such. Users should restart pagination from the beginning if they encounter a cursor-related error.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-e2e-tests/tests/graphql/checkpoints/transactions.move
```diff
@@ -74,7 +74,9 @@
   }
 }
 
-//# run-graphql --cursors 1
+//# run-graphql --cursors bcs(bcs(1u8,0u8,0u64,1u64))
+# BCS-encoded `CursorToken { query_type: Transactions(1), kind: Item(0), checkpoint: 0, position: 1 }`,
+# then wrapped as a `Vec<u8>` (outer `bcs(...)` adds the length prefix BcsCursor<Vec<u8>> expects).
```
amnn: Let's avoid the double-encoding here -- it will involve introducing a new cursor type, like `ByteCursor`, but it seems worth it.

This is also another reminder that having to manually write out cursors in tests is painful. Some of the testing work to support jellyfish could help here, I can demo that soon.

crates/sui-rpc-cursor/src/lib.rs
amnn, line 1: More doc comments would not go amiss here. Now that this is in its own module, it's going to be harder to infer from context.

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -68,7 +71,7 @@ pub(crate) struct TransactionContents {
     pub(crate) contents: Option<Arc<NativeTransactionContents>>,
 }
 
-pub(crate) type CTransaction = JsonCursor<u64>;
+pub(crate) type CTransaction = BcsCursor<Vec<u8>>;
```
amnn: Related to my question about double-encoding -- why not make this a `BcsCursor<CursorToken>`?

crates/sui-rpc-cursor/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+pub fn encode_cursor_for_item(query_type: QueryType, checkpoint: u64, position: u64) -> Bytes {
+    encode_cursor(CursorToken {
+        query_type,
+        kind: CursorKind::Item,
+        checkpoint,
+        position,
+    })
+}
```
amnn: Could we expose a `CursorToken::item` constructor instead?

amnn (commented):
Some questions around the double-encoding. I know we need to leak the fact that the cursor is encoded as BCS, but that seems like an okay concession to make to simplify everything else.

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
wlmyng: When we move off Postgres graphql will not need to extract `position` from the cursor, just create the gRPC cursor on subscriptions, and on reads, pass it directly to gRPC. I wanted to eliminate one hop and go straight to that eventuality, hence not going with `BcsCursor<CursorToken>`. Added a new `ByteCursor`

crates/sui-indexer-alt-graphql/src/api/scalars/cursor.rs
```diff
@@ -50,6 +59,12 @@ impl<C> BcsCursor<C> {
     }
 }
 
+impl ByteCursor {
+    pub(crate) fn new(bytes: impl Into<Vec<u8>>) -> Self {
```
amnn: nit (optional): I personally struggle with when it's the right move to use `impl Into<T>` and when to just accept the `T` directly -- it's really tempting to use `impl Into<T>` everywhere, so I err on the side of doing the explicit thing 😅:

```suggestion
    pub(crate) fn new(bytes: Vec<u8>) -> Self {
```

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -280,8 +283,18 @@ impl Transaction {
         page: &Page<CTransaction>,
         filter: TransactionFilter,
     ) -> Result<TransactionConnection, RpcError> {
-        let after = page.after().map(|c| **c);
-        let before = page.before().map(|c| **c);
+        let after = page
+            .after()
+            .map(|c| bcs::from_bytes::<CursorToken>(c.deref()))
```
amnn: I think `Deref::deref` is intended to be called automatically in response to `&`, does this work?

```suggestion
            .map(|c| bcs::from_bytes::<CursorToken>(&*c))
```

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -280,8 +283,18 @@ impl Transaction {
...
+            .transpose()
+            .context("Invalid `after` cursor")?
```
amnn: Currently this is going to be expressed as an internal error which is not right -- this is a bad user input error. You need to introduce a custom `Error` type which you can parametrize `RpcError` with (`RpcError<Error>`) and this can carry the `BadCursor` error variant.

crates/sui-rpc-cursor/src/lib.rs
amnn: Still think this would be worth it, but not blocking.

amnn (approved):
Looks good, thanks @wlmyng !

amnn:
P.S. don't forget to update release notes.

crates/sui-rpc-cursor/src/lib.rs
wlmyng: Added!

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
wlmyng: Makes sense, updated

PR #27036, opened 2026-06-23
[3/5] Add waitable system object framework

lxfind:
## Description

Adds a generic framework for **waitable system objects** — system objects (beyond a transaction's declared inputs) that execution may read at a specific version and must wait for if the local node hasn't caught up.

- `AssignedVersions` now carries a generic `system_object_versions: BTreeMap<ObjectID, SequenceNumber>` (replacing the dedicated `accumulator_version` field, read via an `accumulator_version()` accessor), threaded from consensus through the execution layer into `TemporaryStore`.
- `RuntimeObjectResolver::is_system_object_available` checks whether a required system object has reached its version. On `TemporaryStore` it errors if the object has no assigned version, and when the object hasn't caught up it records `ExecutionRetryError::SystemObjectUnavailable { object_id }` (out-of-band, since the signal must cross the Move VM boundary) and returns `false`.
- The authority propagates that retry: on `SystemObjectUnavailable` it produces no effects, waits via the new `ObjectCacheRead::notify_read_system_object_at_version` cache primitive for the object to reach the required version, then re-enqueues the transaction.

Inert until a consumer calls the availability check (object funds, 5/5).

## Stack

1/5 #27003 — `ExecutionRetryError` plumbing from execution to authority
2/5 #27019 — rename `ChildObjectResolver` → `RuntimeObjectResolver`
3/5 #27036 — waitable system object framework
4/5 #27037 — record read system objects in effects
5/5 #27038 — check object funds withdraw sufficiency in the Move VM

## Test plan

Framework infrastructure with no consumer in this PR, so no behavior change: validated by the existing `shared_object_version_manager` assignment tests (version map round-trips) and, end to end, by the object-balance stress test in 5/5, which exercises the availability check, the `SystemObjectUnavailable` retry, the cache-notify wait, and re-enqueue.

## Release notes

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #27037, opened 2026-06-23
[4/5] Record read system objects in effects

lxfind:
## Description

Records the system objects a transaction reads during execution into its effects, and reproduces them on replay — completing the round-trip for the waitable-system-object framework (3/5).

- `TemporaryStore` records each system object it reads (version + digest, taken at the required version so it is deterministic across nodes) and emits it into the transaction effects as a read-only consensus object (`UnchangedConsensusKind::ReadOnlyRoot`, reusing the existing variant — externally a read system object is a read-only consensus object), deduped against objects already present.
- On replay, `assign_versions_from_effects` collects those recorded reads (consensus objects that are not declared inputs) into the transaction's `system_object_versions`, alongside the accumulator back-fill that legacy accumulator-version consumers still rely on; for the accumulator root the two agree.

Inert until a consumer records a read (object funds, 5/5): with no recorder, effects are byte-identical and replay falls back to the accumulator back-fill.

## Stack

1/5 #27003 — `ExecutionRetryError` plumbing from execution to authority
2/5 #27019 — rename `ChildObjectResolver` → `RuntimeObjectResolver`
3/5 #27036 — waitable system object framework
4/5 #27037 — record read system objects in effects
5/5 #27038 — check object funds withdraw sufficiency in the Move VM

## Test plan

No behavior change without a recorder; covered by the existing `effects` and `shared_object_version_manager` suites, and exercised end to end by the object-balance stress test in 5/5 — where every object-funds transaction records the accumulator root and the recorded effects reproduce bit-for-bit during checkpoint execution and replay (the simulator's fork detection would catch any nondeterminism).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


lxfind:
Superseded by the reorganized 3-PR stack: #27071 (retry infra) → #27072 (sufficiency API) → #27073 (in-VM check). The retry mechanism was reworked to a post-execution signal (no early-exit), and the generic waitable-framework, effects-recording, and ChildObjectResolver rename were dropped. Closing.

PR #27039, opened 2026-06-23
retire signed_effects_digests write post-MFP

alex-mysten:
`signed_effects_digests` was introduced in #18426 to prevent effects equivocation: with the writeback cache, a validator could lose uncommitted execution outputs on restart, re-execute on an upgraded binary, and produce effects contradicting ones it had already signed and returned to a user (whose signature fed the pre-MFP Quorum-Driver effects certificate). That precondition no longer holds, finality is established via WaitForEffects digest votes, not aggregated per-validator effects signatures, and there is a single post-consensus execution path.

This simplifies fork detection and recovery. After this change, a transaction fork can only occur when CheckpointExecutor executes a certified checkpoint transaction and the effects differ from what the certified checkpoint contains.

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -5251,31 +5251,16 @@ impl AuthorityState {
         effects: TransactionEffects,
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> SuiResult<VerifiedSignedTransactionEffects> {
-        let tx_digest = *effects.transaction_digest();
-        let signed_effects = match epoch_store.get_effects_signature(&tx_digest)? {
-            Some(sig) => {
-                debug_assert!(sig.epoch == epoch_store.epoch());
-                SignedTransactionEffects::new_from_data_and_sig(effects, sig)
-            }
-            _ => {
-                let sig = AuthoritySignInfo::new(
-                    epoch_store.epoch(),
-                    &effects,
-                    Intent::sui_app(IntentScope::TransactionEffects),
-                    self.name,
-                    &*self.secret,
-                );
-
-                let effects = SignedTransactionEffects::new_from_data_and_sig(effects, sig.clone());
-
-                epoch_store.insert_effects_signature(&tx_digest, &sig)?;
-
-                effects
-            }
-        };
+        let sig = AuthoritySignInfo::new(
+            epoch_store.epoch(),
+            &effects,
+            Intent::sui_app(IntentScope::TransactionEffects),
+            self.name,
+            &*self.secret,
+        );
 
         Ok(VerifiedSignedTransactionEffects::new_unchecked(
-            signed_effects,
+            SignedTransactionEffects::new_from_data_and_sig(effects, sig),
         ))
     }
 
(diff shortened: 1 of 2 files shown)


crates/sui-core/src/authority.rs
```diff
@@ -1722,21 +1721,8 @@ impl AuthorityState {
             Err(e) => return ExecutionOutput::Fatal(e),
         };
 
-        let expected_effects_digest = match execution_env.expected_effects_digest {
```
mwtian, line 1725: Is this used by the ckpt exec path?

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -461,11 +455,8 @@ pub struct AuthorityEpochTables {
     /// signature is still epoch-specific and so is stored in the epoch store.
     effects_signatures: DBMap<TransactionDigest, AuthoritySignInfo>,
```
mwtian, line 453: Can this be removed too? It doesn't have to be in this PR.

mwtian (commented):
In the new MFP path, it is still possible for a validator to equivocate by acknowledging different effects digests. The arguments to not add logic to prevent equivocation are:
(1) If the equivocation is due to deterministic execution change, the equivocation is intentional.
(2) If the equivocation is due to non-deterministic execution (after restarting and dropping non-finalized effects), the stored effects digests may not match the eventually checkpointed version anyway.

alex-mysten:
> In the new MFP path, it is still possible for a validator to equivocate by acknowledging different effects digests. The arguments to not add logic to prevent equivocation are: (1) If the equivocation is due to deterministic execution change, the equivocation is intentional. (2) If the equivocation is due to non-deterministic execution (after restarting and dropping non-finalized effects), the stored effects digests may not match the eventually checkpointed version anyway.

Also want to note that this is just removing the stored_effects_digests path, NOT the equivocation prevention on checkpoint executor, which verifies that transaction effects match the certified checkpoint. 

Today, if clients are not currently calling `handle_transaction_info_request`, (which, as I understand it, they should not be on the write path post-MFP), then transactions executed from consensus handler already do not check for fx forks. We rely on checkpoint fork detection to determine if any tx has forked.

crates/sui-core/src/authority.rs
alex-mysten: Yes - moved to ln 2050

mwtian (approved):

crates/sui-core/src/authority/authority_per_epoch_store.rs
andll: Yes tables can now be simply removed

crates/sui-core/src/authority.rs
```diff
@@ -1722,21 +1721,8 @@ impl AuthorityState {
...
-            Some(expected_effects_digest) => Some(expected_effects_digest),
-            None => {
-                // We could be re-executing a previously executed but uncommitted transaction, perhaps after
-                // restarting with a new binary. In this situation, if we have published an effects signature,
-                // we must be sure not to equivocate.
```
mystenmark, line 1730: while we don't produce effects certificates per se any more, validators still give effects to clients, and should not equivocate if asked again. Just as with effects certificates, without equivocation prevention, it could be possible after an incident for a client to observe two contradictory quorums.

this is quite unlikely (probably an understatement), and we could say that if this ever happens, the certified checkpoint controls. but i do want to make it clear that the underlying issue didn't just disappear because we got rid of effects certs

mystenmark (approved):

PR #27043, opened 2026-06-23
Fork execute_transaction_to_effects to eliminate a bunch of obsolete checks

dariorussi:
## Description
Introduce the alternative or forked `execute_transaction_to_effects` path. All the code is in but the protocol flag is not enabled yet (#27049)
One of the main reason for the refactor is to eliminate all the protocol gated checks that have polluted the code base over time, and have caused massive confusion.
The code is written on the old code and does not have divergences with the old behavior except in edge conditions. So semantics is very much the same. The few differences are protocol gated (by the feature itself) so small changes in edge condition are expected and welcome.
In time when we cut a new execution version we will be able to remove the old code entirely and apply even more "intrusive" changes/abstractions in the "new code".

Best way to review is probably go to `execute_transaction_to_effects` and walk the code. Side by side with the old code would be ideal.

## Test plan
Existing tests. Also looking in test coverage for transaction execution here (#27050)

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-types/src/gas_model/gas_common.rs
+++ b/crates/sui-types/src/gas_model/gas_common.rs
@@ -1,0 +1,226 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+pub use checked::*;
+
+mod checked {
+    use crate::error::{UserInputError, UserInputResult};
+    use crate::transaction::ObjectReadResult;
+    use crate::{ObjectID, gas};
+    use serde::{Deserialize, Serialize};
+
+    pub fn check_gas_objects(gas_objs: &[&ObjectReadResult]) -> UserInputResult {
+        // All gas objects have an address owner
+        // Note: because of address balance payments, gas_objs may be empty.
+        for gas_object in gas_objs {
+            // if as_object() returns None, it means the object has been deleted (and therefore
+            // must be a shared object).
+            if let Some(obj) = gas_object.as_object() {
+                if !obj.is_address_owned() {
+                    return Err(UserInputError::GasObjectNotOwnedObject {
+                        owner: obj.owner.clone(),
+                    });
+                }
+            } else {
+                // This case should never happen (because gas can't be a shared object), but we
+                // handle this case for future-proofing
+                return Err(UserInputError::MissingGasPayment);
+            }
+        }
+        Ok(())
+    }
+
+    pub fn check_gas_data(
+        gas_objs: &[&ObjectReadResult],
+        gas_budget: u64,
+        available_address_balance_gas: u64,
+        min_transaction_cost: u64,
+        max_gas_budget: u64,
+    ) -> UserInputResult {
+        // Gas budget is between min and max budget allowed
+        if gas_budget > max_gas_budget {
+            return Err(UserInputError::GasBudgetTooHigh {
+                gas_budget,
+                max_budget: max_gas_budget,
+            });
+        }
+        if gas_budget < min_transaction_cost {
+            return Err(UserInputError::GasBudgetTooLow {
+                gas_budget,
+                min_budget: min_transaction_cost,
+            });
+        }
+
+        // Gas balance (all gas coins + address balance together) is bigger or equal to budget
+        let mut gas_balance = available_address_balance_gas as u128;
+        for gas_obj in gas_objs {
+            gas_balance += gas::get_gas_balance(gas_obj.as_object().ok_or(
+                UserInputError::InvalidGasObject {
+                    object_id: gas_obj.id(),
+                },
+            )?)? as u128;
+        }
+        if gas_balance < gas_budget as u128 {
+            Err(UserInputError::GasBalanceTooLow {
+                gas_balance,
+                needed_gas_amount: gas_budget as u128,
+            })
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Portion of the storage rebate that gets passed on to the transaction sender. The remainder
+    /// will be burned, then re-minted + added to the storage fund at the next epoch change
+    pub fn sender_rebate(storage_rebate: u64, storage_rebate_rate: u64) -> u64 {
+        // we round storage rebate such that `>= x.5` goes to x+1 (rounds up) and
+        // `< x.5` goes to x (truncates). We replicate `f32/64::round()`
+        const BASIS_POINTS: u128 = 10000;
+        (((storage_rebate as u128 * storage_rebate_rate as u128)
+        + (BASIS_POINTS / 2)) // integer rounding adds half of the BASIS_POINTS (denominator)
+        / BASIS_POINTS) as u64
+    }
+
+    pub fn half_digits_rounding(n: u64) -> u64 {
+        if n < 1000 {
+            return 1000;
+        }
+        let digits = n.ilog10();
+        let drop = digits / 2;
+        let base = 10u64.pow(drop);
+        n.div_ceil(base).saturating_mul(base)
+    }
+
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct PerObjectStorage {
+        /// The new "value" for this object storage. Computed
+        /// at the end of execution while determining storage charges.
+        /// This will be the new storage rebate.
+        pub storage_cost: u64,
+        /// storage_rebate is the value of this object.
+        /// This is computed at the end of execution while determining storage charges.
+        /// The value is in Sui.
+        pub storage_rebate: u64,
+        /// The object size post-transaction in bytes
+        pub new_size: u64,
+    }
+
+    /// Per-object storage-gas accumulator, shared by both gas models. Pure data + arithmetic; the Move
+    /// meter stays on the outer `SuiGasStatus`, which passes the `unmetered` flag into `track_mutation`.
+    #[derive(Debug)]
+    pub struct StorageGas {
+        /// Per-object storage cost + rebate, accumulated during execution.
+        per_object_storage: Vec<(ObjectID, PerObjectStorage)>,
+        /// Running total of per-object storage cost. Metered path only.
+        total_storage_cost: u64,
+        /// Running total of per-object storage rebate. Metered path only.
+        total_storage_rebate: u64,
+        /// Storage rebate accrued while running unmetered (system transactions), retained in effects
+        /// and parked onto 0x5. Kept separate from `total_storage_rebate`: it must read 0 on metered
+        /// txns (its consumer `conserve_unmetered_storage_rebate` runs unconditionally).
+        unmetered_storage_rebate: u64,
+        /// Multiplier applied to the storage byte cost (`ProtocolConfig::storage_gas_price`).
+        pub storage_gas_price: u64,
+        /// Refundable per-byte storage cost (`ProtocolConfig::obj_data_cost_refundable`).
+        storage_per_byte_cost: u64,
+    }
+
+    impl StorageGas {
+        pub fn new(storage_gas_price: u64, storage_per_byte_cost: u64) -> Self {
+            Self {
+                per_object_storage: Vec::new(),
+                total_storage_cost: 0,
+                total_storage_rebate: 0,
+                unmetered_storage_rebate: 0,
+                storage_gas_price,
+                storage_per_byte_cost,
+            }
+        }
+
+        pub fn storage_gas_units(&self) -> u64 {
+            self.total_storage_cost
+        }
+
+        pub fn storage_rebate(&self) -> u64 {
+            self.total_storage_rebate
+        }
+
+        pub fn unmetered_storage_rebate(&self) -> u64 {
+            self.unmetered_storage_rebate
+        }
+
+        pub fn per_object_storage(&self) -> &Vec<(ObjectID, PerObjectStorage)> {
+            &self.per_object_storage
+        }
+
+        pub fn reset(&mut self) {
+            self.per_object_storage = Vec::new();
+            self.total_storage_cost = 0;
+            self.total_storage_rebate = 0;
+            self.unmetered_storage_rebate = 0;
+        }
+
+        /// Update the running storage cost/rebate totals for the object.
+        /// Returns the new object storage cost (based on `new_size`), or `None` on overflow.
+        pub fn track_mutation(
+            &mut self,
+            object_id: ObjectID,
+            new_size: usize,
+            storage_rebate: u64,
+            unmetered: bool,
+        ) -> Option<u64> {
+            if unmetered {
+                return self
+                    .unmetered_storage_rebate
+                    .checked_add(storage_rebate)
+                    .map(|total| {
+                        self.unmetered_storage_rebate = total;
+                        0
+                    });
+            }
+
+            let new_size = new_size as u64;
+            let storage_cost = new_size
+                .checked_mul(self.storage_per_byte_cost)?
+                .checked_mul(self.storage_gas_price)?;
+            self.total_storage_cost = self.total_storage_cost.checked_add(storage_cost)?;
+            self.total_storage_rebate = self.total_storage_rebate.checked_add(storage_rebate)?;
+            self.per_object_storage.push((
+                object_id,
+                PerObjectStorage {
+                    storage_cost,
+                    storage_rebate,
+                    new_size,
+                },
+            ));
+            Some(storage_cost)
+        }
+    }
+
+    #[test]
+    fn test_half_digits_rounding() {
+        assert_eq!(half_digits_rounding(0), 1000);
+        assert_eq!(half_digits_rounding(1), 1000);
+        assert_eq!(half_digits_rounding(999), 1000);
+        assert_eq!(half_digits_rounding(1000), 1000);
+        assert_eq!(half_digits_rounding(1001), 1010);
+        assert_eq!(half_digits_rounding(1050), 1050);
+        assert_eq!(half_digits_rounding(1999), 2000);
+        assert_eq!(half_digits_rounding(20_000), 20_000);
+        assert_eq!(half_digits_rounding(20_001), 20_100);
+        assert_eq!(half_digits_rounding(20_500), 20_500);
+        assert_eq!(half_digits_rounding(29_999), 30_000);
+        assert_eq!(half_digits_rounding(300_000), 300_000);
+        assert_eq!(half_digits_rounding(300_001), 300_100);
+        assert_eq!(half_digits_rounding(305_500), 305_500);
+        assert_eq!(half_digits_rounding(305_501), 305_600);
+        assert_eq!(half_digits_rounding(999_999), 1_000_000);
+        assert_eq!(half_digits_rounding(1_000_000), 1_000_000);
+        assert_eq!(half_digits_rounding(1_000_001), 1_001_000);
+        assert_eq!(half_digits_rounding(1_005_000), 1_005_000);
+        assert_eq!(half_digits_rounding(1_005_001), 1_006_000);
+        assert_eq!(half_digits_rounding(1_999_999), 2_000_000);
+        assert_eq!(half_digits_rounding(10_000_001), 10_001_000);
+        assert_eq!(half_digits_rounding(100_000_001), 100_010_000);
+    }
+}
--- a/crates/sui-types/src/gas_model/gas_common.rs
+++ b/crates/sui-types/src/gas_model/gas_common.rs
@@ -1,0 +1,232 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#![deny(clippy::arithmetic_side_effects)]
+
+pub use checked::*;
+
+mod checked {
+    use crate::error::{UserInputError, UserInputResult};
+    use crate::transaction::ObjectReadResult;
+    use crate::{ObjectID, gas};
+    use serde::{Deserialize, Serialize};
+
+    pub fn check_gas_objects(gas_objs: &[&ObjectReadResult]) -> UserInputResult {
+        // All gas objects have an address owner
+        // Note: because of address balance payments, gas_objs may be empty.
+        for gas_object in gas_objs {
+            // if as_object() returns None, it means the object has been deleted (and therefore
+            // must be a shared object).
+            if let Some(obj) = gas_object.as_object() {
+                if !obj.is_address_owned() {
+                    return Err(UserInputError::GasObjectNotOwnedObject {
+                        owner: obj.owner.clone(),
+                    });
+                }
+            } else {
+                // This case should never happen (because gas can't be a shared object), but we
+                // handle this case for future-proofing
+                return Err(UserInputError::MissingGasPayment);
+            }
+        }
+        Ok(())
+    }
+
+    pub fn check_gas_data(
+        gas_objs: &[&ObjectReadResult],
+        gas_budget: u64,
+        available_address_balance_gas: u64,
+        min_transaction_cost: u64,
+        max_gas_budget: u64,
+    ) -> UserInputResult {
+        // Gas budget is between min and max budget allowed
+        if gas_budget > max_gas_budget {
+            return Err(UserInputError::GasBudgetTooHigh {
+                gas_budget,
+                max_budget: max_gas_budget,
+            });
+        }
+        if gas_budget < min_transaction_cost {
+            return Err(UserInputError::GasBudgetTooLow {
+                gas_budget,
+                min_budget: min_transaction_cost,
+            });
+        }
+
+        // Gas balance (all gas coins + address balance together) is bigger or equal to budget
+        let mut gas_balance = available_address_balance_gas as u128;
+        for gas_obj in gas_objs {
+            // Saturation is unreachable: a sum of u64 coin balances cannot overflow u128.
+            gas_balance =
+                gas_balance.saturating_add(gas::get_gas_balance(gas_obj.as_object().ok_or(
+                    UserInputError::InvalidGasObject {
+                        object_id: gas_obj.id(),
+                    },
+                )?)? as u128);
+        }
+        if gas_balance < gas_budget as u128 {
+            Err(UserInputError::GasBalanceTooLow {
+                gas_balance,
+                needed_gas_amount: gas_budget as u128,
+            })
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Portion of the storage rebate that gets passed on to the transaction sender. The remainder
+    /// will be burned, then re-minted + added to the storage fund at the next epoch change
+    // Cannot overflow: u64::MAX * u64::MAX leaves ~2^65 of u128 headroom for the `+ 5000`.
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn sender_rebate(storage_rebate: u64, storage_rebate_rate: u64) -> u64 {
+        // we round storage rebate such that `>= x.5` goes to x+1 (rounds up) and
+        // `< x.5` goes to x (truncates). We replicate `f32/64::round()`
+        const BASIS_POINTS: u128 = 10000;
+        (((storage_rebate as u128 * storage_rebate_rate as u128)
+        + (BASIS_POINTS / 2)) // integer rounding adds half of the BASIS_POINTS (denominator)
+        / BASIS_POINTS) as u64
+    }
+
+    pub fn half_digits_rounding(n: u64) -> u64 {
+        if n < 1000 {
+            return 1000;
+        }
+        let digits = n.ilog10();
+        let drop = digits / 2;
+        let base = 10u64.pow(drop);
+        n.div_ceil(base).saturating_mul(base)
+    }
+
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct PerObjectStorage {
+        /// The new "value" for this object storage. Computed
+        /// at the end of execution while determining storage charges.
+        /// This will be the new storage rebate.
+        pub storage_cost: u64,
+        /// storage_rebate is the value of this object.
+        /// This is computed at the end of execution while determining storage charges.
+        /// The value is in Sui.
+        pub storage_rebate: u64,
+        /// The object size post-transaction in bytes
+        pub new_size: u64,
+    }
+
+    /// Per-object storage-gas accumulator, shared by both gas models. Pure data + arithmetic; the Move
+    /// meter stays on the outer `SuiGasStatus`, which passes the `unmetered` flag into `track_mutation`.
+    #[derive(Debug)]
+    pub struct StorageGas {
+        /// Per-object storage cost + rebate, accumulated during execution.
+        per_object_storage: Vec<(ObjectID, PerObjectStorage)>,
+        /// Running total of per-object storage cost. Metered path only.
+        total_storage_cost: u64,
+        /// Running total of per-object storage rebate. Metered path only.
+        total_storage_rebate: u64,
+        /// Storage rebate accrued while running unmetered (system transactions), retained in effects
+        /// and parked onto 0x5. Kept separate from `total_storage_rebate`: it must read 0 on metered
+        /// txns (its consumer `conserve_unmetered_storage_rebate` runs unconditionally).
+        unmetered_storage_rebate: u64,
+        /// Multiplier applied to the storage byte cost (`ProtocolConfig::storage_gas_price`).
+        pub storage_gas_price: u64,
+        /// Refundable per-byte storage cost (`ProtocolConfig::obj_data_cost_refundable`).
+        storage_per_byte_cost: u64,
+    }
+
+    impl StorageGas {
+        pub fn new(storage_gas_price: u64, storage_per_byte_cost: u64) -> Self {
+            Self {
+                per_object_storage: Vec::new(),
+                total_storage_cost: 0,
+                total_storage_rebate: 0,
+                unmetered_storage_rebate: 0,
+                storage_gas_price,
+                storage_per_byte_cost,
+            }
+        }
+
+        pub fn storage_gas_units(&self) -> u64 {
+            self.total_storage_cost
+        }
+
+        pub fn storage_rebate(&self) -> u64 {
+            self.total_storage_rebate
+        }
+
+        pub fn unmetered_storage_rebate(&self) -> u64 {
+            self.unmetered_storage_rebate
+        }
+
+        pub fn per_object_storage(&self) -> &Vec<(ObjectID, PerObjectStorage)> {
+            &self.per_object_storage
+        }
+
+        pub fn reset(&mut self) {
+            self.per_object_storage = Vec::new();
+            self.total_storage_cost = 0;
+            self.total_storage_rebate = 0;
+            self.unmetered_storage_rebate = 0;
+        }
+
+        /// Update the running storage cost/rebate totals for the object.
+        /// Returns the new object storage cost (based on `new_size`), or `None` on overflow.
+        pub fn track_mutation(
+            &mut self,
+            object_id: ObjectID,
+            new_size: usize,
+            storage_rebate: u64,
+            unmetered: bool,
+        ) -> Option<u64> {
+            if unmetered {
+                return self
+                    .unmetered_storage_rebate
+                    .checked_add(storage_rebate)
+                    .map(|total| {
+                        self.unmetered_storage_rebate = total;
+                        0
+                    });
+            }
+
+            let new_size = new_size as u64;
+            let storage_cost = new_size
+                .checked_mul(self.storage_per_byte_cost)?
+                .checked_mul(self.storage_gas_price)?;
+            self.total_storage_cost = self.total_storage_cost.checked_add(storage_cost)?;
+            self.total_storage_rebate = self.total_storage_rebate.checked_add(storage_rebate)?;
+            self.per_object_storage.push((
+                object_id,
+                PerObjectStorage {
+                    storage_cost,
+                    storage_rebate,
+                    new_size,
+                },
+            ));
+            Some(storage_cost)
+        }
+    }
+
+    #[test]
+    fn test_half_digits_rounding() {
+        assert_eq!(half_digits_rounding(0), 1000);
+        assert_eq!(half_digits_rounding(1), 1000);
+        assert_eq!(half_digits_rounding(999), 1000);
+        assert_eq!(half_digits_rounding(1000), 1000);
+        assert_eq!(half_digits_rounding(1001), 1010);
+        assert_eq!(half_digits_rounding(1050), 1050);
+        assert_eq!(half_digits_rounding(1999), 2000);
+        assert_eq!(half_digits_rounding(20_000), 20_000);
+        assert_eq!(half_digits_rounding(20_001), 20_100);
+        assert_eq!(half_digits_rounding(20_500), 20_500);
+        assert_eq!(half_digits_rounding(29_999), 30_000);
+        assert_eq!(half_digits_rounding(300_000), 300_000);
+        assert_eq!(half_digits_rounding(300_001), 300_100);
+        assert_eq!(half_digits_rounding(305_500), 305_500);
+        assert_eq!(half_digits_rounding(305_501), 305_600);
+        assert_eq!(half_digits_rounding(999_999), 1_000_000);
+        assert_eq!(half_digits_rounding(1_000_000), 1_000_000);
+        assert_eq!(half_digits_rounding(1_000_001), 1_001_000);
+        assert_eq!(half_digits_rounding(1_005_000), 1_005_000);
+        assert_eq!(half_digits_rounding(1_005_001), 1_006_000);
+        assert_eq!(half_digits_rounding(1_999_999), 2_000_000);
+        assert_eq!(half_digits_rounding(10_000_001), 10_001_000);
+        assert_eq!(half_digits_rounding(100_000_001), 100_010_000);
+    }
+}
--- a/crates/sui-types/src/gas_model/gas_common.rs
+++ b/crates/sui-types/src/gas_model/gas_common.rs
@@ -4,6 +4,8 @@
 #![deny(clippy::arithmetic_side_effects)]
 #![deny(clippy::cast_possible_truncation)]
 #![deny(clippy::indexing_slicing)]
+#![deny(clippy::cast_possible_wrap)]
+#![deny(clippy::cast_sign_loss)]
 
 use crate::error::{UserInputError, UserInputResult};
 use crate::transaction::ObjectReadResult;
--- a/crates/sui-types/src/gas_model/gas_common.rs
+++ b/crates/sui-types/src/gas_model/gas_common.rs
@@ -2,231 +2,224 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #![deny(clippy::arithmetic_side_effects)]
+#![deny(clippy::cast_possible_truncation)]
+#![deny(clippy::indexing_slicing)]
 
-pub use checked::*;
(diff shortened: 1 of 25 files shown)


sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -88,21 +89,83 @@ mod checked {
         sui_system_state::{ADVANCE_EPOCH_FUNCTION_NAME, SUI_SYSTEM_MODULE_NAME},
     };
 
-    /// Whether the *head* early error is `InsufficientFundsForWithdraw`. Used to gate the first
-    /// address-balance gas-payment pruning hotfix: the head error is the one surfaced as the
-    /// failure status, so keying off it (rather than any occurrence) keeps the pruning bit-for-bit
-    /// with the original single-error hotfix.
-    fn head_error_is_insufficient_funds_for_withdraw(
+    /// Whether `InsufficientFundsForWithdraw` appears anywhere in the early-error list.
+    /// `get_early_execution_error` always appends IFFW last (priority 5), so when a
```
lxfind: This is not a reliable assumption.

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -390,7 +1004,7 @@ mod checked {
             execution_params: &ExecutionOrEarlyError,
             protocol_config: &ProtocolConfig,
         ) -> bool {
-            if !head_error_is_insufficient_funds_for_withdraw(execution_params) {
+            if !any_error_is_insufficient_funds_for_withdraw(execution_params) {
```
lxfind: Can you explain why is this change safe?

dariorussi: we talked about this 1:1

sui-execution/latest/sui-adapter/src/execution_engine.rs
dariorussi: we talked about this 1:1

crates/sui-types/src/gas_model/gas_common.rs
```diff
@@ -0,0 +1,226 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+pub use checked::*;
+
+mod checked {
```
lxfind: Did you forget about the sui_macros::with_checked_arithmetic derivative?
Same for gas_v3

crates/sui-types/src/gas_model/gas_common.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    pub fn half_digits_rounding(n: u64) -> u64 {
+        if n < 1000 {
+            return 1000;
+        }
+        let digits = n.ilog10();
+        let drop = digits / 2;
+        let base = 10u64.pow(drop);
+        n.div_ceil(base).saturating_mul(base)
```
lxfind: Why is this changed?

dariorussi: because we do not have the macro that crashes on overflow and such this is to guarantee that operations cannot fail.
Given the operation here it should be safe and a saturating multiplication when rounding gas seems safe as it would imply a cost of U64::max.
Or so I think....

crates/sui-types/src/gas_model/gas_common.rs
dariorussi: yeah I purposely removed it so we could check the operation and avoid crashing. That was one of the goals of this refactor, try to eliminate as many crashes as possible.
Thoughts?

lxfind: I wonder if a better fix is to still rely on the macros but instead of crashing turning them into invariant violations?

crates/sui-types/src/gas_model/gas_common.rs
```diff
@@ -0,0 +1,232 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#![deny(clippy::arithmetic_side_effects)]
+
```
tnowacki: Definitely also add `clippy::cast_possible_truncation` and maybe even `clippy::indexing_slicing`

crates/sui-types/src/gas_model/gas_common.rs
tnowacki: Sorry, why do we need the macros at all if we have the clippy error? 
I really despise macros like this and we have as such done away with them in the PTB layer and down

crates/sui-types/src/gas_model/gas_common.rs
```diff
@@ -0,0 +1,232 @@
...
+            })
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Portion of the storage rebate that gets passed on to the transaction sender. The remainder
+    /// will be burned, then re-minted + added to the storage fund at the next epoch change
+    // Cannot overflow: u64::MAX * u64::MAX leaves ~2^65 of u128 headroom for the `+ 5000`.
+    #[allow(clippy::arithmetic_side_effects)]
```
tnowacki: Why allow here? 
Just switch to using saturating if we need it?

crates/sui-types/src/gas_model/gas_v3.rs
```diff
@@ -0,0 +1,393 @@
+// Copyright (c) 2021, Facebook, Inc. and its affiliates
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#![deny(clippy::arithmetic_side_effects)]
+
+pub use checked::*;
```
tnowacki: Same comment as `gas_common`

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
```diff
@@ -98,13 +104,19 @@ impl InvariantChecker {
         transaction_kind: &TransactionKind,
         gas_data: &GasData,
         transaction_signer: SuiAddress,
+        is_gasless: bool,
```
tnowacki: This is fine, but just curious, do we already have this information in `gas_data` or `transaction_kind`?

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
```diff
@@ -53,6 +53,12 @@ struct InvariantInputs {
     /// address balance, and gas-data coin-reservation digests. Consumed by
     /// `check_address_balance_changes` and `check_ownership_invariants`.
     input_reservations: BTreeMap<(SuiAddress, TypeTag), u64>,
+    /// For gasless transactions only (`Some` even when empty): the per-`(owner, coin type)`
+    /// withdrawal reservations declared by the PTB's `FundsWithdrawal` inputs. Unlike
+    /// `input_reservations` this is keyed by the coin type `T` (not `Balance<T>`) and carries no
```
tnowacki: But why...? This information is already in the `input_reservation`. Why not just use that (and add the `Balance` when querying it)?

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
```diff
@@ -539,6 +556,43 @@ fn compute_input_reservations(
     reservations
 }
 
+/// Compute the per-`(owner, coin type)` withdrawal reservations of a gasless transaction from its
+/// PTB `FundsWithdrawal` inputs. Keyed by the coin type `T` to match the gasless allowed-token
+/// table, unlike `compute_input_reservations` above (keyed `Balance<T>`).
+fn compute_gasless_reservations(
```
tnowacki: If we do need this separately, why is this in `invariants`? Aren't these used to actually reject transactions?

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -576,6 +576,56 @@ impl<'backing> TemporaryStore<'backing> {
         self.invariants.clear();
     }
 
+    /// Consume this (post-execution) store and return the "recorded no-op" store used by a BumpOnly bail:
+    /// keep the input-derived state it already computed, discard everything execution produced, then bump
+    /// the mutable inputs. Its effects then record ONLY those input version bumps and the input
+    /// dependencies — nothing the (now-discarded) execution touched.
+    pub(crate) fn into_recorded_noop(self) -> Self {
```
tnowacki: Should we name this `into_bump_only` or something? 
Just spitballing if we need two terms for this or not

tnowacki: (its also not really a noop since the inputs are bumped, which confuses me mildly)

tnowacki: ((not a lot, just mildly))

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -827,8 +888,15 @@ impl<'backing> TemporaryStore<'backing> {
         gas_data: &GasData,
         transaction_signer: SuiAddress,
     ) {
-        self.invariants
-            .set_transaction_inputs(transaction_kind, gas_data, transaction_signer);
+        // Same gasless predicate as the execution engine's `payment_kind`.
+        let is_gasless = self.protocol_config.enable_gasless()
+            && is_gasless_transaction(gas_data, transaction_kind);
```
tnowacki: Yeah it looks like the gasless information is computable from transaction kind and gas data, so why make it an input below?

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -658,12 +708,23 @@ impl<'backing> TemporaryStore<'backing> {
             .fold(0, |sum, obj| sum + obj.object_size_for_gas_metering())
     }
 
+    /// Validates gasless post-execution invariants, using the withdrawal reservations cached by
+    /// [`Self::set_invariant_inputs`].
+    pub(crate) fn check_gasless_execution_requirements(&self) -> Result<(), String> {
+        self.check_gasless_execution_requirements_with_reservations(
+            self.invariants.gasless_reservations(),
```
tnowacki: This makes me think that the input reservations should be in the temporary store instead of in invariants? though it is a tricky one

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -48,6 +48,9 @@ pub mod checked {
     #[derive(Debug)]
     enum PaymentMetadata {
         Unmetered,
+        /// Metered-but-free; carries no state. The gasless withdrawal reservations the
+        /// storage-charging validation needs are cached on the `TemporaryStore` (see
+        /// `set_invariant_inputs`).
```
tnowacki: I'm really confused by this comment, since it seems to be describing a situation that only existed in previous versions of this PR. 
I'm not sure if this variant needs a comment at all, but if it does, something more succinct about what happens in gas charging for gasless as a whole would probably make more sense

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -289,6 +299,21 @@ pub mod checked {
             self.gas_status.charge_publish_package(size)
         }
 
+        /// Charge `storage_read` for each input object (system packages excepted). Charges go into the
+        /// Move VM meter, which `summary()` reads live — so the running summary reflects them on return.
```
tnowacki: > which `summary()` reads live — so the running summary reflects them on return.

I feel dumb and am not sure how to parse this

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -307,19 +332,186 @@ pub mod checked {
             self.gas_status.charge_storage_read(owner_cost)
         }
 
-        /// Resets any mutations, deletions, and events recorded in the store, as well as any storage costs and
-        /// rebates, then Re-runs gas smashing. Effects on store are now as if we were about to begin execution
+        /// Restore the store + gas state to the post-input shape: drop execution writes, clear storage
+        /// cost/rebate, re-smash gas (`drop_writes` cleared it), and re-touch mutable inputs so their
+        /// versions still bump on the err path. Used by the legacy storage-OOG fallback and the v15+
+        /// err-path handler.
```
tnowacki: If this is used by the legacy fallback, should we give it a different name?
Or is it used by non-legacy codepaths

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -307,19 +332,186 @@ pub mod checked {
...
             temporary_store.drop_writes();
             self.gas_status.reset_storage_cost_and_rebate();
             self.smash_gas(temporary_store);
+            temporary_store.ensure_active_inputs_mutated();
+        }
+
+        /// Tail of `execute_ptb`: bucketize computation via `bucketize_computation`, which fixes the
+        /// effective gas price (lowered to the abort-tx cap on a Move abort) and signals OOG if priced
+        /// computation alone exceeds the budget. Runs on Ok or Err (to capture the abort flag); an
+        /// execution error keeps precedence over a rounding error. Skipped for unmetered.
```
tnowacki: Am I just tired? Cannot grok this at all

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -307,19 +332,186 @@ pub mod checked {
...
+            match result {
+                Ok(v) => round_res.map(|_| v).map_err(Into::into),
+                Err(e) => Err(e),
+            }
+        }
+
+        /// Collect per-object storage cost/rebate and verify it fits the remaining budget. For gasless,
+        /// validate execution requirements first (before `ensure_active_inputs_mutated` adds objects to
+        /// written_objects). Does not retry/reset (the caller does) and does not apply payment (`charge` does).
+        pub fn charge_storage(
```
tnowacki: Maybe we should have a slightly different name for this function

> does not apply payment (`charge` does).

Feels a bit silly to be like `charge_storage` does not charge, that is the job of `charge`. 
Maybe `meter_storage`? Thoughts?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -307,19 +332,186 @@ pub mod checked {
...
+                }
+            }
+        }
+
+        /// Apply the final gas charge derived from the current `SuiGasStatus`, per payment kind
+        /// (`Unmetered` / `Gasless` / `Smash`).
+        ///
+        /// Reads `metadata.gas_charge_location` at apply time (no pre-rewind snapshot): the SUIPR-753
+        /// fix — `reset_writes` runs before `charge`, undoing any override (e.g. `coin::send_funds`)
+        /// on the err path before we read here.
```
tnowacki:
```
        /// Reads `metadata.gas_charge_location` at apply time (no pre-rewind snapshot): the SUIPR-753
        /// fix — `reset_writes` runs before `charge`, undoing any override (e.g. `coin::send_funds`)
        /// on the err path before we read here.
```
What is SUIPR-753? What are these fixes? What is the pre-rewind snapshot? Am lost

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -307,19 +332,186 @@ pub mod checked {
...
+                    });
+                    // IFFW against an address balance: nothing was withdrawn, so charge nothing.
+                    if iffw_failure
+                        && matches!(
+                            metadata.gas_charge_location,
+                            PaymentLocation::AddressBalance(_)
+                        )
+                    {
+                        return GasCostSummary::default();
+                    }
```
tnowacki: I didn't think this was reachable here anymore?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -307,19 +332,186 @@ pub mod checked {
...
+                    if let PaymentLocation::Coin(_) = metadata.gas_charge_location {
+                        #[skip_checked_arithmetic]
```
tnowacki, line 443: Is this file still in checked arithmetic? We should probably strip that out too though it doesn't have to be this PR

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -307,19 +332,186 @@ pub mod checked {
...
+                        trace!(target: "replay_gas_info", "Gas smashing has occurred for this transaction");
+                    }
+                    let cost_summary = self.gas_status.summary();
+                    self.apply_payment(temporary_store, cost_summary, metadata.gas_charge_location)
+                }
+            }
+        }
+
+        /// Deduct the charge: from the gas coin (coin payment) or via an accumulator event (address
+        /// balance).
```
tnowacki: But it might not actually be deducting? It could be positive

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -307,19 +332,186 @@ pub mod checked {
...
+                            sui_types::gas_coin::GAS::type_tag(),
+                        );
+                        let event = AccumulatorEvent::from_balance_change(
+                            payer_address,
+                            balance_type,
+                            net_change
+                                .checked_neg()
+                                .expect("net gas usage is never i64::MIN"),
+                        )
+                        .expect("Failed to create accumulator event for gas charging");
```
tnowacki, line 474: We are okay with these `expect`s still?

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -307,19 +332,186 @@ pub mod checked {
...
+                        temporary_store.add_accumulator_event(event);
+                    }
+                }
+                PaymentLocation::Coin(gas_object_id) => {
+                    let mut gas_object = temporary_store
+                        .read_object(&gas_object_id)
+                        .expect("gas coin is an input object and present after smashing")
```
tnowacki, line 481: ditto

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -722,6 +1292,40 @@ pub(crate) mod checked {
             (cost_summary, result, timings)
         }
 
+        fn gasless_withdrawal_reservations(
```
tnowacki, line 1345: Wasn't this code in invariants now? Or is this for the legacy path?

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -103,10 +103,67 @@ pub(crate) mod checked {
...
         transaction_kind: &TransactionKind,
         protocol_config: &ProtocolConfig,
+    ) -> PaymentKind {
+        if gas_data.is_unmetered() || transaction_kind.is_system_tx() {
+            PaymentKind::unmetered()
+        } else if protocol_config.enable_gasless()
+            && is_gasless_transaction(gas_data, transaction_kind)
+        {
+            // The gasless withdrawal reservations are cached on the store by
+            // `set_invariant_inputs`, under the same gasless predicate as here.
```
tnowacki: I do not understand this comment

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -103,10 +103,67 @@ pub(crate) mod checked {
...
+            PaymentKind::gasless()
+        } else if gas_data.payment.is_empty() {
+            PaymentKind::smash(vec![PaymentMethod::AddressBalance(
+                gas_data.owner,
+                gas_data.budget,
+            )])
+            .expect("unable to create a payment kind with a single address balance")
```
tnowacki: Are we okay with this expect still?

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -103,10 +103,67 @@ pub(crate) mod checked {
...
+                .iter()
+                .map(|entry| {
+                    if let Ok(parsed) = ParsedDigest::try_from(entry.2) {
+                        PaymentMethod::AddressBalance(gas_data.owner, parsed.reservation_amount())
+                    } else {
+                        PaymentMethod::Coin(*entry)
+                    }
+                })
+                .collect();
+            PaymentKind::smash(payment_methods).expect(
```
tnowacki: ditto

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -146,6 +203,13 @@ pub(crate) mod checked {
         Vec<ExecutionTiming>,
         Result<<Mode as ExecutionMode>::ExecutionResults, <Mode as ExecutionMode>::Error>,
     );
+
+    /// Gas summary, execution result, and timings produced by `execute_transaction`.
+    type ExecutionOutcome<Mode> = (
```
tnowacki: Should we have a struct? Though this is fine by me

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -306,6 +523,363 @@ pub(crate) mod checked {
...
+    /// the one expected reason; every other variant is an execution bug and is reported to
+    /// `execution_bump_only_exits`, whose `reason` label is `Self::label`.
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    enum BumpOnlyReason {
+        InsufficientFundsForWithdraw,
+        /// Resetting writes after a failed execution could not charge even input-only storage.
+        WriteReset,
+        Conservation,
+        Ownership,
+    }
```
tnowacki, line 573: Really love how this all turned out!

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -306,6 +523,363 @@ pub(crate) mod checked {
...
+        epoch_timestamp_ms: u64,
+        protocol_config: &ProtocolConfig,
+        metrics: Arc<ExecutionMetrics>,
+        enable_expensive_checks: bool,
+        execution_params: ExecutionOrEarlyError,
+        trace_builder_opt: &mut Option<MoveTraceBuilder>,
+    ) -> Outcome<Mode> {
+        // Short-circuit insufficient_funds. No execution, `Outcome::BumpOnly`
+        if should_short_circuit_insufficient_funds(&execution_params) {
+            assert_reachable!("IFFW short-circuit fired");
```
tnowacki, line 656: Was this here before? What does the assert mean?

crates/sui-types/src/gas_model/gas_common.rs
tzakian: +1 to not using the macro and force handling these explicitly with the clippy lints.

I think the macro is a good "high ROI for low effort" type of thing, but IMO there are places where you may want to handle arithmetic failures in different ways (e.g., one way may be an invariant violation, the other may now be fallible, but the callers can do something based on that, other places may well and truly need to panic still).

crates/sui-types/src/gas_model/gas_common.rs
```diff
@@ -0,0 +1,232 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#![deny(clippy::arithmetic_side_effects)]
+
+pub use checked::*;
+
```
tzakian: Don't need to do this pub-use of an inner module if we don't use the checked arith macro -- it was introduced for that reason so since that's not used here it can go away.

crates/sui-types/src/gas_model/gas_common.rs
```diff
@@ -0,0 +1,232 @@
...
+        / BASIS_POINTS) as u64
+    }
+
+    pub fn half_digits_rounding(n: u64) -> u64 {
+        if n < 1000 {
+            return 1000;
+        }
+        let digits = n.ilog10();
+        let drop = digits / 2;
+        let base = 10u64.pow(drop);
```
tzakian: It's fascinating (and pretty neat) that Clippy can establish this as ok

crates/sui-types/src/gas_model/gas_common.rs
```diff
@@ -0,0 +1,232 @@
...
+            unmetered: bool,
+        ) -> Option<u64> {
+            if unmetered {
+                return self
+                    .unmetered_storage_rebate
+                    .checked_add(storage_rebate)
+                    .map(|total| {
+                        self.unmetered_storage_rebate = total;
+                        0
+                    });
```
tzakian: Nit I think this would be a bit more readable:

```suggestion
let total = self
        .unmetered_storage_rebate
        .checked_add(storage_rebate)?;

    self.unmetered_storage_rebate = total;
    return Some(0);
             
```

sui-execution/latest/sui-adapter/src/execution_engine.rs
tzakian: +1 would prefer an invariant violation and/or for this to be a BumpOnly if it occurs

sui-execution/latest/sui-adapter/src/execution_engine.rs
tzakian: I like the idea of a struct here.

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -306,6 +523,363 @@ pub(crate) mod checked {
...
+        coin: Option<ObjectID>,
+        status: SuiGasStatus,
+    }
+
+    struct Finalized<Mode: ExecutionMode> {
+        gas: GasOutcome,
+        status: ExecutionStatus,
+        timings: Vec<ExecutionTiming>,
+        execution_result: Result<Mode::ExecutionResults, Mode::Error>,
+    }
```
tzakian: Placement of these seems a bit odd to me. Maybe up alongside BumpOnly and similar?

sui-execution/latest/sui-adapter/src/execution_engine.rs
tzakian: It's something to do with anithesis IIRC

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -152,12 +155,15 @@ pub mod checked {
...
 
-        pub fn new_unmetered(tx_digest: TransactionDigest) -> Self {
+        pub fn new_unmetered(
+            tx_digest: TransactionDigest,
+            protocol_config: &ProtocolConfig,
+        ) -> Self {
             Self {
                 tx_digest,
-                gas_model_version: 6, // pick any of the latest, it should not matter
+                gas_model_version: protocol_config.gas_model_version(),
```
tzakian, line 161: nice!

sui-execution/latest/sui-adapter/src/gas_charger.rs
tzakian: I'd prefer if we could error instead but I realize that errors in these places can be very hard to handle.

sui-execution/latest/sui-adapter/src/temporary_store.rs
tzakian: (((I like `into_bump_only` FWIW)))

tzakian: ((((We're using the bump only terminology pretty consistently throughout so I think it makes sense to be consistent with that here as well))))

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -306,6 +523,363 @@ pub(crate) mod checked {
...
+
+        let checks = check_effects::<Mode>(temporary_store, gas_charger, protocol_config, metrics);
+        // Execution error wins; otherwise a failed effects check fails the tx.
+        let result = result.and_then(|v| checks.map(|()| v));
+
+        if result.is_err() {
+            gas_charger
+                .reset_writes(temporary_store)
+                .map_err(|error| (error.into(), BumpOnlyReason::WriteReset))?;
+        }
```
alex-mysten, line 824: It's a bit surprising that `reset_writes` not only resets writes but also decides how much gas to charge according to an implicit decision tree.
Would it be clearer if we modeled the outcomes explicitly and, if practical, decide how much gas to charge as a pure function, and then reset writes as a side effect? Something like:
```
enum GasStrategy {
    Full, // we can charge the full amount
    InputsOnly, // oog, charge only input storage
    Budget, // oog, charge budget
    Zero, // iffw
}
fn price(meter, writes, params) -> (strategy, cost_summary)
```

sui-execution/latest/sui-adapter/src/execution_engine.rs
dariorussi: moved to Invariant/BumpOnly

crates/sui-types/src/gas_model/gas_common.rs
dariorussi: I want to also point out that a change in the macro to return an error (or None) on arithmetic operator feels incredibly obscure as it would change an infallible operation into a fallible one which is what we are doing here explicitly.

crates/sui-types/src/gas_model/gas_common.rs
dariorussi: removed and changed the code

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
dariorussi: we do and I changed it, please check! thank you

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
dariorussi: I changed it and I hope I got that right, please review and I see your point, sorry

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
dariorussi: changed for the invariants to hold the function that returns the gasless reservation via `input_reservations` , I think that is what you meant.
As for `input_reservations` being in `TemporaryStore` is a valid argument and I commented on that

sui-execution/latest/sui-adapter/src/temporary_store.rs
dariorussi: renamed

sui-execution/latest/sui-adapter/src/temporary_store.rs
dariorussi: I'd like to do that in a different PR. I am not sure it's too tricky as we have the `TemporaryStore` where we need that info but I want to think about it and don't change much this PR

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: no really apology. Fixed. My bad

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: gone! apology again

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: yes legacy too. Maybe a different name would help but as a follow up for me

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: did I apologize already? my bad

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: changed name

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: my bad again, somehow a lot of those slipped through.

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: removed, please review

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: yes it is and it will be easy to clean up once legacy go

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: changed

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: I'd like to do in a different PR and make `charge()` fallible and see how BumpOnly is going... it's getting a bit out of hand right now and I'd like to do that on a clean base

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: see above

sui-execution/latest/sui-adapter/src/execution_engine.rs
dariorussi: gone

sui-execution/latest/sui-adapter/src/execution_engine.rs
dariorussi: this is the legacy path

sui-execution/latest/sui-adapter/src/execution_engine.rs
dariorussi: I am not sure what you mean here and please expand a bit. Meanwhile few comments that may or may not be answering anything.
`reset_writes()` is not properly named and it really is the failure path. So it's "remove all side effects", "reload input objects", "compute storage cost for input objects". It's the execution path for a Move abort for instance. In a sense that function is doing what it should but a better name would help.
Also it was not charging and not sure if `charge_storage` (now renamed to `meter_storage`) was misleading too.

As for the general approach of having more "pure" functions and less mutability sprinkled around, we are on the same page and it is for a follow up.
This PR is intended to stay relatively close to main for now, cleaning up few critical things.
You have a solid point in terms of trying that approach which feels much more natural

alex-mysten: Happy to consider this as a follow up. I can see that this is an improvement. Perhaps a name change would make it more clear ~ `meter` is definitely better than `charge` or `compute`. `reset_writes` doesn't really capture that we're also deciding how much gas to charge. 

A small note, `reset_writes` is currently in `mod::legacy`, don't think this is intentional.

I'm not sure I have a specific recommendation to apply here, just that state and orchestration seems to have leaked across `gas_charger`, `gas_status`, and `temporary_store`, and I'm not totally sure of the mental model. Do we have a north star?

dariorussi: I have renamed `charge_storage` and `reset_writes` (`handle_error()`) is the new name. Hope it makes it cleaner.
As for the north star, I do not have one. Once the legacy code disappears the objective is to make the code cleaner and better organized. We'll see.
Certainly there is a lot of back and forth across the components you mentioned and we need some serious clean up there.

dariorussi: and thanks for pointing out `reset_write` was in the wrong spot. That was great!

sui-execution/latest/sui-adapter/src/gas_charger.rs
```diff
@@ -402,12 +393,24 @@ pub mod checked {
             }
         }
 
+        /// Single reset for execution errors.
+        /// Invoked on any legitimate error (Move Abort, OOG, etc.).
```
tzakian: Nit: what is an illegitimate error? not necessary to change unless you're going in for other stuff/updating the PR again before landing but clarifying this a smidge more would be nice probably.

crates/sui-types/src/gas_model/gas_common.rs
```diff
@@ -2,231 +2,224 @@
...
-                    return Err(UserInputError::GasObjectNotOwnedObject {
-                        owner: obj.owner.clone(),
-                    });
-                }
-            } else {
-                // This case should never happen (because gas can't be a shared object), but we
-                // handle this case for future-proofing
-                return Err(UserInputError::MissingGasPayment);
+#![deny(clippy::cast_possible_truncation)]
+#![deny(clippy::indexing_slicing)]
```
tzakian, line 6: Here and in the other places, can you add denials for:

* `clippy::cast_possible_wrap`
* `clippy::cast_sign_loss`

Just to be extra safe?

tzakian (approved):
LGTM, although a couple small nits if you're feeling like addressing them before landing.

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: I'll rephrase, and may have spent time on this comment already.
The point that this was trying to make - and that hit me few times - was that the code in question deals with the most legitimate error possible and not just unexpected errors.
But clearly is understood by myself only....

crates/sui-types/src/gas_model/gas_v3.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// Execution cost table to be used.
+    pub execution_cost_table: CostTable,
+    /// RGP-multiplier cap on the effective gas price for aborted transactions.
+    max_gas_price_rgp_factor_for_aborted_transactions: Option<u64>,
+}
+
+impl std::fmt::Debug for SuiCostTable {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // TODO: dump the fields.
+        write!(f, "SuiCostTable(...)")
```
tnowacki, line 49: should this todo be todone?

tnowacki: Maybe a follow up PR would be easy for your dog 🌭

crates/sui-types/src/gas_model/gas_v3.rs
```diff
@@ -0,0 +1,389 @@
...
+}
+
+impl SuiCostTable {
+    pub(crate) fn new(c: &ProtocolConfig, gas_price: u64) -> Self {
+        // gas_price here is the Reference Gas Price, however we may decide
+        // to change it to be the price passed in the transaction
+        let min_transaction_cost = c
+            .base_tx_cost_fixed()
+            .checked_mul(gas_price)
+            .expect("base tx cost cannot overflow: gas_price is bounded by max_gas_price");
```
tnowacki, line 60: I assume no reason/way to non `expect` here?

crates/sui-types/src/gas_model/gas_v3.rs
```diff
@@ -0,0 +1,389 @@
...
+        if self.force_computation_cost_to_budget {
+            return self.gas_budget;
+        }
+        let raw_units = self.gas_status.gas_used_pre_gas_price();
+        let bucketed_units = half_digits_rounding(raw_units);
+        bucketed_units.saturating_mul(self.effective_gas_price)
+    }
+
+    /// Computation cost reported in `summary()`, capped so
+    /// `computation + storage_cost - sender_rebate ≤ gas_budget`
```
tnowacki:
```suggestion
    /// `computation + storage_cost - sender_rebate <= gas_budget`
``` 
😛

tnowacki: I'm fine with it I just notice that llms love unicode in comments (or maybe you do too, idk)

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
```diff
@@ -640,8 +662,15 @@ impl InvariantChecker {
             })
             .filter(|id| {
                 // remove any non-mutable inputs. This will remove deleted or readonly shared
-                // objects
-                mutable_inputs.contains(id)
+                // objects.
+                // REVIEW AND DELETE COMMENT AFTER: before we took a caller-passed `mutable_inputs`
```
tnowacki: What does this mean?

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
```diff
@@ -503,6 +520,7 @@ fn compute_input_reservations(
     transaction_kind: &TransactionKind,
     gas_data: &GasData,
     transaction_signer: SuiAddress,
+    is_gasless: bool,
```
tnowacki: Isn't this computable from transaction_kind and gas_data?

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
```diff
@@ -50,8 +50,7 @@ use crate::type_layout_resolver::TypeLayoutResolver;
 struct InvariantInputs {
     /// Per-`(address, type)` funds-accumulator reservation budget authorized by this transaction.
     /// Sources: PTB `FundsWithdrawalArg`s (sender/sponsor as owner), gas paid entirely from an
-    /// address balance, and gas-data coin-reservation digests. Consumed by
-    /// `check_address_balance_changes` and `check_ownership_invariants`.
+    /// address balance, and gas-data coin-reservation digests.
```
tnowacki, line 53: It is a wee-bit strange to me still that this lives here instead of the temporary store since it is used for actual checks in gasless transactions... though its fine? I guess?

tnowacki: (The inputs map below the line, not the comment)

crates/sui-types/src/gas_model/gas_v3.rs
dariorussi: good catch!

sui-execution/latest/sui-adapter/src/gas_charger.rs
dariorussi: given the function name was changed to a more meaningful `handle_error` I removed the entire comment

crates/sui-types/src/gas_model/gas_v3.rs
dariorussi: yeah another day....

crates/sui-types/src/gas_model/gas_v3.rs
dariorussi: I'll leave it for a follow up and there are few others to check too.
This error is obviously a pretty dysfunctional system which we want to cover but this is more of a configuration problem I think

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
dariorussi: this is intended to be deleted after catching the attention of reviewers.
Those are things that were bothering me and I left a big comment for the reviewer

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
dariorussi: done, it was done one level up but both `TransactionKind` and `GasData` are here so no point in passing it in

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
dariorussi: I thought we talked about this :) 
I would like to look at this in a follow up. It was there already and I did not want to change much.
As I said in the lost comment I think it may not be difficult to pull it off but did not want to add much churn.
Having said so I am going to do the work and see how it looks

dariorussi: it looks a bit more of a change than I'd like now and I think we are best do that in a follow up.
It'll be more obvious what we want

PR #27046, opened 2026-06-24
[trace-view] Improved UX for starting inlined instructions

awelc:
## Description 

When inlined instruction (for a macro) was the first one in a function, what the debugger was displaying wasn't ideal. It did put both the function's frame and the virtual (inlined) frame on the stack, but because the function's frame didn't really hit its first instruction with a concrete line number, it was displayed with line 0 (which meant that clicking this frame on the stack did not point to ANY line in the debugger window, other than top of the stack which is always there):
image

This PR finesses it by making sure that even if frame's line number is freshly initialized (i.e., it's 0) the debugger will actually display something meaningful, that is the line where the function's signature is defined (it's intentional - it is meant to indicate that something is off - that we are in fact no in the first real instruction of this frame; the user can switch to disassembly mode if they need to see more details):

image

In the implementation we don't actually replace 0 with the line where the function signature is defined to avoid disturbing the stepping logic. If we did that, in the rare case when the first real instruction of the function is on the same line as the signature, we'd have debugger visually skip it due to how it handles multiple instructions that belong to the same source code line.

## Test plan 

Test output has been adjusted to reflect this change. Also tested manually.

### DIFF
--- a/external-crates/move/crates/move-analyzer/trace-adapter/src/adapter.ts
+++ b/external-crates/move/crates/move-analyzer/trace-adapter/src/adapter.ts
@@ -26,6 +26,7 @@ import {
     ExecutionResult,
     ExecutionResultKind,
     IMoveCallStack,
+    moveStackFrameDisplayLine,
 } from './runtime';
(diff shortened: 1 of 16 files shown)


tzakian (approved):
Seems sensible to me!

PR #27051, opened 2026-06-24
Extract execution-layer invariant checks into a InvariantChecker module

lxfind:
## Description

The execution layer's conservation-check logic (SUI conservation, balance-accumulator authorization and representability, plus the recovery orchestration) was interleaved through `temporary_store.rs` and `execution_engine.rs`, distracting from the core store/engine code. This moves it behind a dedicated `ConservationChecker` that owns the conservation-only bookkeeping and exposes the checks as its API; the main files just record data and call it. Pure refactor — no behavioral change.

## Test plan

Covered by existing conservation tests in `sui-e2e-tests`/`sui-core`; relying on CI. No logic changed (functions moved verbatim).


dariorussi (approved):
I like it! thanks

PR #27052, opened 2026-06-24
Reduce and pipeline consensus commit-handler hot-path work

andll:
## Description

The consensus commit handler is effectively single-threaded and is the bottleneck at high TPS. This reduces and pipelines its per-commit work:

- **Pipeline BCS deserialization** onto a dedicated single-threaded deserialize worker, so parsing a commit overlaps the handler processing the previous one. A small bounded channel preserves ordering (one worker, FIFO, one handler) and applies backpressure instead of buffering. Runs on the default runtime.
- **Replace the finalized-tx moka cache with a generational `RwLock<HashSet>`.** The cache has a single writer (the handler, ~15k inserts/s) and many concurrent readers (the voting path); moka's per-insert maintenance was ~15% of the handler's CPU under load, while its lock-free reads weren't the constraint. Membership is all the read path needs, so reads take a shared lock and inserts are a cheap hashset insert plus an occasional O(1) generation rotation — cheap enough to stay inline. The cache is advisory (`is_recently_finalized` falls back to the executed-tx check), so the bounded two-generation retention window is safe.

(The fine-grained `monitored_scope`s that were in this PR have landed on `main` separately and dropped out on rebase.)

## Test plan

New unit tests cover the cache's insert/contains and generation-rotation eviction. The deser pipeline preserves commit ordering by construction (single worker, FIFO bounded channel, single handler) and is covered by existing `consensus_handler` tests.

## Release notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/consensus_handler.rs
+++ b/crates/sui-core/src/consensus_handler.rs
@@ -3052,15 +3052,9 @@ impl MysticetiConsensusHandler {
         );
         tasks.spawn(monitored_future!(async move {
             while let Some(consensus_commit) = commit_receiver.recv().await {
-                // Prior commits are only replayed for `observe_commit` metadata and don't use
-                // the parsed transactions, so skip parsing them.
                 let transactions: ParsedConsensusTransactions = {
                     let _scope = monitored_scope("ConsensusCommitHandler::deserialize_worker");
-                    if consensus_commit.commit_ref.index > last_processed_commit_at_startup {
-                        consensus_commit.transactions()
-                    } else {
-                        Vec::new()
-                    }
+                    consensus_commit.transactions()
                 };
                 // The send is intentionally outside the scope above: on the bounded channel it
                 // blocks when the handler is the bottleneck, and that idle-wait would otherwise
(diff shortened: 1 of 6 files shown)


crates/sui-core/src/consensus_handler.rs
```diff
@@ -3026,15 +3040,51 @@ impl MysticetiConsensusHandler {
...
+        // bottleneck. Single-threaded; ordering is preserved (one worker, FIFO channel, one handler).
+        let (parsed_sender, mut parsed_receiver) = monitored_mpsc::channel(
+            "consensus_deserialized_commits",
+            CONSENSUS_HANDLER_DESERIALIZE_CHANNEL_CAPACITY,
+        );
         tasks.spawn(monitored_future!(async move {
-            // TODO: pause when execution is overloaded, so consensus can detect the backpressure.
             while let Some(consensus_commit) = commit_receiver.recv().await {
+                // Prior commits are only replayed for `observe_commit` metadata and don't use
+                // the parsed transactions, so skip parsing them.
```
mystenmark: is this optimization worth it? prior commit replay only happens once at startup

mystenmark (approved):

mystenmark:
can you please randomize the cache size for tests? (use `randomize_cache_capacity_in_tests()`)

andll:
Added randomization to authority per epoch tests. The cache unit tests themselves dont use randomization because they are sensitive to the cache size (they test specific cache behaviour)

PR #27054, opened 2026-06-24
Suppress resubmission of consensus-processed transactions

mystenmark:
## Description

Validators were resubmitting transactions that had already executed long ago. `handle_submit_transaction` only short-circuited an already-executed transaction when it could *also* rebuild the full `ExecutedData` response, but `complete_executed_data` fails once the transaction's input/output object versions have been pruned — so the dedup checks silently fell through and the transaction was sent to consensus again.

Gate on `is_consensus_message_processed` (current epoch), which is true for any transaction consensus has already processed regardless of whether its effects data is still retrievable, also covering transactions that are sequenced but still deferred. `Rejected { TransactionAlreadyExecuted }` is not a perfect fit for the deferred case, but is acceptable here since well-behaved clients receive `Executed` before pruning and this path primarily guards against faulty clients resubmitting indefinitely.

## Test plan

- `test_submit_transaction_consensus_message_processed`: a single processed-but-not-executed (deferred) transaction is suppressed.
- `test_submit_soft_bundle_transactions_with_consensus_message_processed`: a processed transaction is removed from a soft bundle before submission while the rest of the bundle is still submitted.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority_server.rs
+++ b/crates/sui-core/src/authority_server.rs
@@ -882,7 +882,11 @@ impl ValidatorService {
                     .with_label_values(&[req_type])
                     .inc();
                 results[idx] = Some(SubmitTxResult::Rejected {
-                    error: UserInputError::TransactionAlreadyExecuted { digest: tx_digest }.into(),
+                    error: SuiErrorKind::TransactionProcessing {
+                        digest: tx_digest,
+                        status: "sequenced by consensus".to_string(),
+                    }
+                    .into(),
                 });
                 debug!(
                     ?tx_digest,
--- a/crates/sui-core/src/authority_server.rs
+++ b/crates/sui-core/src/authority_server.rs
@@ -29,7 +29,9 @@ use sui_network::{
 };
 use sui_types::effects::TransactionEffectsAPI;
 use sui_types::message_envelope::Message;
-use sui_types::messages_consensus::{ConsensusPosition, ConsensusTransaction};
+use sui_types::messages_consensus::{
+    ConsensusPosition, ConsensusTransaction, ConsensusTransactionKey,
+};
 use sui_types::messages_grpc::{
     ObjectInfoRequest, ObjectInfoResponse, RawSubmitTxResponse, SystemStateRequest,
     TransactionInfoRequest, TransactionInfoResponse,
@@ -67,6 +69,7 @@ use crate::gasless_rate_limiter::GaslessRateLimiter;
 use crate::{
     authority::{AuthorityState, consensus_tx_status_cache::ConsensusTxStatus},
     consensus_adapter::{ConsensusAdapter, ConsensusAdapterMetrics, ConsensusOverloadChecker},
+    consensus_handler::SequencedConsensusTransactionKey,
     traffic_controller::{TrafficController, parse_ip, policies::TrafficTally},
 };
 use crate::{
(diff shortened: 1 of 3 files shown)


crates/sui-core/src/authority_server.rs
```diff
@@ -856,6 +867,26 @@ impl ValidatorService {
...
+            // Suppress resubmission of transactions consensus already processed this epoch:
+            // executed transactions whose effects details could not be reconstructed above (e.g.
+            // objects pruned), and sequenced-but-deferred transactions. Rejected is imperfect for
+            // the deferred case, but acceptable since well-behaved clients get Executed before
+            // pruning; this mainly sheds continuous resubmissions from faulty clients.
+            let consensus_key = SequencedConsensusTransactionKey::External(
+                ConsensusTransactionKey::Certificate(tx_digest),
+            );
+            if epoch_store.is_consensus_message_processed(&consensus_key)? {
+                metrics.submission_suppressed_already_processed.inc();
```
akichidis: Could you include the `req_type` as a label on this metric?

aschran (approved):

akichidis (approved):

crates/sui-core/src/authority_server.rs
```diff
@@ -856,6 +867,26 @@ impl ValidatorService {
...
+                results[idx] = Some(SubmitTxResult::Rejected {
+                    error: UserInputError::TransactionAlreadyExecuted { digest: tx_digest }.into(),
```
mwtian: It seems a transaction deferred for any reason can reach this path. It might become confusing to see TransactionAlreadyExecuted but later found out the transaction was cancelled. It is better to add a new error code or reuse an existing one in SuiError, and categorize the error as `Aborted` in `fn categorize(&self)`.

mystenmark: currently all user transactions that are marked processed in consensus are executed whether they are deferred or not, so at worst this error will appear to "see into the future" because the transaction may not be cancelled/executed yet

mwtian: This still means the error should be marked as `Aborted` to have the client retry it, instead of treating it as a permanent failure. I can push a commit.

crates/sui-core/src/authority_server.rs
mystenmark: Done — the metric is now an `IntCounterVec` labeled by `req_type`.

crates/sui-core/src/authority_server.rs
mwtian: ptal 516789366e34d4a2b1bc4f5bd78762c5c12c5c81 and feel free to edit / revert if it doesn't make sense.

mystenmark: I think this can result in infinite retries if they hit the previous case which checks `transaction_executed_in_last_epoch()`

mwtian: If the transaction input objects has been updated due to the transaction's execution, the transaction will get rejected much earlier. Or has this been broken now?

mwtian: 629a96341f2e5052d2a2af56b6a48861457d10bf should address your concern.

PR #27055, opened 2026-06-24
rpc-store: add object versions by checkpoint index

bmwill:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-rpc-store/src/indexer/object_version_by_checkpoint.rs
```diff
@@ -0,0 +1,607 @@
...
+        // stale prior version.
+        let mut removed: BTreeMap<ObjectID, SequenceNumber> = BTreeMap::new();
+        for tx in &checkpoint.transactions {
+            let lamport = tx.effects.lamport_version();
+            for oref in tx
+                .effects
+                .deleted()
+                .into_iter()
+                .chain(tx.effects.unwrapped_then_deleted())
+                .chain(tx.effects.wrapped())
```
amnn, line 144: It would be nice to use `object_changes` for this, filtered to all changes that don't have an output version/digest -- personally I find this easier to reason about than chaining together the various cases from effects v1 (and confirm mentally that a case hasn't been missed).

crates/sui-rpc-store/src/schema/objects.rs
```diff
@@ -77,6 +77,18 @@ pub fn options(resolver: &sui_consistent_store::CfOptionsResolver) -> rocksdb::O
     resolver.options(NAME)
 }
 
+/// Prefix encoder for "every version of `id`". Encodes as the 32 raw
```
amnn, line 80: Is the explicit prefix necessary here? Could `ObjectID` be used directly instead?

amnn (approved):
LGTM!

crates/sui-rpc-store/src/schema/objects.rs
bmwill: yeah good point

bmwill: would need to add this to `sui-consistent-store` instead of here though due to orphan rule

crates/sui-rpc-store/src/indexer/object_version_by_checkpoint.rs
bmwill: yeah i always am so confused by the various apis we have in the mono repo for this because they all do subtlety different things as well. let me look at this

PR #27057, opened 2026-06-24
[consensus] fix block publish/subscribe issues

mwtian:
## Description

Review of the consensus block publish/subscribe path with coding agent surfaced several issues which are fixed:

- **Snapshot/live race (both stream servers).** Each subscription snapshots cached blocks and only *then* subscribes to the live broadcast, so a block produced in the gap is sent in neither segment. Fixed by subscribing before snapshotting, which turns a possible gap into a harmless duplicate the receiver dedups.
- **Stale resume round.** The validator subscriber captured `last_received` once at subscribe time and reused it on every reconnect, re-streaming and re-verifying already-accepted blocks. It is now recomputed from `DagState` before each connection attempt, matching the observer path.
- **Observer round narrowing.** `highest_round_per_authority` was `uint64` on the wire and narrowed with `as u32 + 1`, so an out-of-range value could wrap to 0 and replay the entire block cache. The field is now `uint32` / `Vec<Round>` end-to-end.

Plus two minor cleanups: reset the subscriber's reconnect backoff when a block is received (so a recovered peer doesn't inherit an escalated delay), and make `SubscriptionCounter` increment/decrement infallible so the inc/dec pairing can't drift.

## Test plan

New `subscriber_recomputes_resume_round_on_reconnect` covers the resume-round fix: it advances the accepted round mid-subscription and asserts that reconnects resume from the advanced round. The rest is covered by existing `subscriber` / `authority_service` / `observer_*` tests.

### DIFF
--- a/consensus/core/src/observer_service.rs
+++ b/consensus/core/src/observer_service.rs
@@ -221,7 +221,7 @@ impl ObserverNetworkService for ObserverService {
     async fn handle_stream_blocks(
         &self,
         peer: NodeId,
-        highest_round_per_authority: Vec<u64>,
+        highest_round_per_authority: Vec<Round>,
     ) -> ConsensusResult<ObserverBlockStream> {
         if highest_round_per_authority.len() != self.context.committee.size() {
             return Err(ConsensusError::InvalidSizeOfHighestAcceptedRounds(
@@ -230,6 +230,11 @@ impl ObserverNetworkService for ObserverService {
             ));
         }
 
+        // Subscribe to the live broadcast BEFORE snapshotting past blocks below, so a block
+        // accepted concurrently with the snapshot lands in both segments (a harmless duplicate
+        // that the receiver dedups) rather than in neither (a gap).
+        let broadcast_rx = self.rx_accepted_block_broadcast.resubscribe();
+
         // Collect all accepted blocks from DagState that the observer hasn't yet seen,
         // sorted by round for consistent ordering.
         let past_blocks = {
@@ -237,7 +242,7 @@ impl ObserverNetworkService for ObserverService {
             let mut past_blocks = Vec::new();
 
             for (authority, _) in self.context.committee.authorities() {
-                let from_round = highest_round_per_authority[authority.value()] as u32 + 1;
+                let from_round = highest_round_per_authority[authority.value()] + 1;
                 past_blocks.extend(dag_state.get_cached_blocks(authority, from_round));
             }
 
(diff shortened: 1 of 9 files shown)


consensus/core/src/observer_service.rs
```diff
@@ -230,14 +230,19 @@ impl ObserverNetworkService for ObserverService {
...
+
         // Collect all accepted blocks from DagState that the observer hasn't yet seen,
         // sorted by round for consistent ordering.
         let past_blocks = {
             let dag_state = self.dag_state.read();
             let mut past_blocks = Vec::new();
 
             for (authority, _) in self.context.committee.authorities() {
-                let from_round = highest_round_per_authority[authority.value()] as u32 + 1;
+                let from_round = highest_round_per_authority[authority.value()] + 1;
```
arun-koshy: should we make it `let from_round = highest_round_per_authority[authority.value()].saturating_add(1);`?

to prevent the wrap to 0 that you fixed else where in this PR?

consensus/core/src/observer_subscriber.rs
```diff
@@ -131,10 +131,9 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
             // already-seen blocks.
             let highest_round_per_authority = {
                 let ds = dag_state.read();
-                let mut rounds = vec![0u64; context.committee.size()];
+                let mut rounds = vec![0u32; context.committee.size()];
                 for (authority, _) in context.committee.authorities() {
-                    rounds[authority.value()] =
-                        ds.get_last_block_for_authority(authority).round() as u64;
+                    rounds[authority.value()] = ds.get_last_block_for_authority(authority).round();
```
arun-koshy: should we clamp this by gc round?

`rounds[authority.value()] =  ds.get_last_block_for_authority(authority).round().max(ds.gc_round());`

consensus/core/src/subscriber.rs
```diff
@@ -369,4 +368,138 @@ mod test {
...
+    // attempt. Previously it was captured once at subscribe() time and reused for every reconnect,
+    // causing already-accepted blocks to be re-streamed and re-verified.
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn subscriber_recomputes_resume_round_on_reconnect() {
+        use crate::block::TestBlock;
+
+        // Records the `last_received` round passed to each subscribe_blocks() call, then returns a
+        // single-block stream that ends immediately to force frequent reconnections.
+        struct RecordingClient {
+            calls: Arc<Mutex<Vec<Round>>>,
```
arun-koshy: SubscriberTestClient exists above, could you add this field to it and reuse that for this test?

arun-koshy (approved):

consensus/core/src/subscriber.rs
```diff
@@ -369,4 +368,138 @@ mod test {
...
+            dag_state.clone(),
+        );
+
+        let peer = context.committee.to_authority_index(2).unwrap();
+        subscriber.subscribe(peer);
+
+        // Before any block from the peer is accepted, every reconnect resumes from genesis (0).
+        for _ in 0..3 {
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
```
akichidis: Why not simply `tokio::time::sleep(Duration::from_secs(3)).await;`

akichidis (approved):
All good fixes, thanks @mwtian !

PR #27062, opened 2026-06-25
Automatic fork recovery gated on certified outcome and corrected binary version

alex-mysten:
## Description

When a validator detects a fork it records a marker and dies; recovery (clearing the marker and re-deriving) is deliberate equivocation, since the node may have already signed the forked result. This is safe only when the network has already certified the canonical outcome — a quorum certificate is irrevocable, so re-signing cannot change what finalizes — and pointless under the binary that forked, which would deterministically fork again.

Automatic recovery at startup therefore clears a fork marker only when both hold:

- the marker records the certified checkpoint the node diverged from (checkpoint forks store the certified digest; transaction forks store the sequence of the certified checkpoint that supplied the expected effects). Detection persists the certificate before the marker, so this provenance proves the network sealed the outcome. Self-divergence — the builder re-deriving its own previous checkpoint differently, or execution diverging from the node's own previously signed effects — carries no certified reference and is never auto-recovered; it requires explicit `ForkRecoveryConfig` overrides.
- the marker was recorded by a different binary version than the one now running, so the node hangs after a fork until a corrected binary is deployed. This also bounds recurrence without a separate attempt counter: if the new binary forks too, its own marker blocks it in turn.

Markers move to new column families (`checkpoint_fork_detected`, `transaction_fork_detected_v2`) carrying the provenance and forking version; the old watermark key and tuple table are retired. `ExecutionEnv`'s expected effects digest becomes an enum (`Certified { digest, checkpoint_seq }` / `Uncertified`) so provenance is explicit at every call site.

## Test plan

- sui-node unit tests cover the recovery gates: same-version forks hang until a new binary, self-divergence and own-signed-effects forks never auto-recover, certified forks recover, and operator overrides clear markers the auto path refuses.
- Simtests `test_auto_fork_recovery_checkpoint_fork` and `test_auto_fork_recovery_transaction_fork` exercise both fork classes end-to-end (detection, crash, restart under a corrected binary via an `override_binary_version` failpoint, recovery, rejoin); `test_split_brain_recovery_via_checkpoint_overrides` covers the manual path when nothing certifies.

## Release notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Validators automatically recover from checkpoint and transaction forks on restart when the network has certified the canonical outcome and a new binary version is deployed; forks caused by the running binary version, or without a certified outcome, halt the node awaiting a corrected binary or operator-supplied fork recovery overrides.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/tests/simtest.rs
+++ b/crates/sui-benchmark/tests/simtest.rs
@@ -1446,11 +1446,13 @@ mod test {
         }
     }
 
-    // One validator forks (checkpoint fork) while the other three keep a quorum and certify past it.
-    // After the injection is cleared it restarts under RecoverOncePerVersion, clears its fork state,
-    // and rejoins.
+    // A checkpoint fork (vs the transaction fork below): one validator participates in consensus
+    // live with fork injection on all execution paths, so its builder constructs a divergent
+    // local checkpoint while the other three keep a quorum and certify the canonical one. After
+    // the injection is cleared it restarts under RecoverOncePerVersion as a corrected binary,
+    // clears its fork state, and rejoins.
     #[sim_test(config = "test_config()")]
-    async fn test_auto_fork_recovery_minority_validator() {
+    async fn test_auto_fork_recovery_checkpoint_fork() {
         sui_protocol_config::ProtocolConfig::poison_get_for_min_version();
 
         let test_cluster = build_test_cluster(4, 10_000, 4).await;
@@ -1522,8 +1524,14 @@ mod test {
             "expected the forked validator to detect a checkpoint fork"
         );
 
-        // Corrected binary: stop injecting forks.
+        // Corrected binary: stop injecting forks and report a new binary version, since recovery
+        // refuses to clear a fork under the version that produced it.
         clear_fail_point("simulate_fork_during_execution");
+        register_fail_point_arg("override_binary_version", || {
+            Some(std::sync::Arc::new(std::sync::Mutex::new(
+                "corrected-binary".to_string(),
+            )))
+        });
 
         // Restart the forked validator under RecoverOncePerVersion (no overrides).
         for validator in test_cluster.swarm.validator_nodes() {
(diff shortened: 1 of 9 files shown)


mystenmark (approved):
looks good!

ebmifa:
Heads up @alex-mysten — `test::test_auto_fork_recovery_transaction_fork` appears schedule-sensitive. On the v1.77.0 version-bump PR (#27321, cut from tonight's main tip which includes this PR) it fails **deterministically across two runs**: the `assert!(forked, ...)` at simtest.rs:1820 — the failpoint-injected fork never trips within its 60s window ([run](https://github.com/MystenLabs/sui/actions/runs/29551659120), 3687/3688 pass otherwise).

Since simtests replay a fixed seed per binary, re-runs reproduce the identical schedule — the version-string delta in the bump PR is enough to shift the msim schedule into the losing execution, while the framework-snapshot PR (#27320) and main's own runs pass with their binaries. Same class as the recent autodebug simtest relaxations (#27315).

This is currently blocking the v1.77.0 version bump landing for the branch-cut ceremony. Could the fork window be widened / the injection made schedule-robust? Happy to re-run the bump PR CI once a fix lands to verify.

ebmifa:
Follow-up with corrected analysis + a fix: #27322.

Two corrections to my earlier comment after deeper investigation: (1) the PR simtest seed derives from the merge-commit SHA — so the two failures were the same seed, but (2) the failure is **not** purely seed-determined: on an idle 16-core box the unfixed test passes 10/10 including the exact CI merge tree + seed (`11397606143398739615`) and under CPU saturation. The race is in the test's arming order — the fork injection is registered after `start()`, and whether the target's catch-up re-execution outruns that arming depends on real-thread scheduling msim doesn't fully serialize (CI runners lose it systematically; this box never does).

#27322 arms the injection and kill hooks before restart (exclusion-set matching since the new sim node id is unknowable pre-start), which removes the ordering dependence entirely and also closes a latent sim-abort if a fork tripped before the kill hooks were armed. Validated on seeds 1/2/3 + the CI seed + both sibling tests.

PR #27064, opened 2026-06-25
refactor spam weight calculation for handle_submit_transaction

aschran:
## Description 

weights now calculated based on tx submission outcome

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority_server.rs
+++ b/crates/sui-core/src/authority_server.rs
@@ -679,8 +679,12 @@ impl ValidatorService {
         let mut results: Vec<Option<SubmitTxResult>> = vec![None; request.transactions.len()];
         // Total size of all transactions in the request.
         let mut total_size_bytes = 0;
-        // Traffic control spam weight to use for the transaction.
-        let mut spam_weight = Weight::zero();
+        // Whether the request contains any gasless transaction.
+        let mut has_gasless = false;
+        // Set when a transaction duplicates an in-flight submission at admission. Tracked
+        // separately because it is detected after the per-tx results are finalized (those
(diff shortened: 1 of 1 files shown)


crates/sui-core/src/authority_server.rs
```diff
@@ -1150,7 +1161,48 @@ impl ValidatorService {
...
+            }
+        }
+        Weight::zero()
+    }
+
+    fn submission_spam_weight(result: &SubmitTxResult) -> Weight {
+        match result {
+            SubmitTxResult::Submitted { .. } => Weight::zero(),
+            // Non-submitted results can't be charged.
+            SubmitTxResult::Executed { .. } | SubmitTxResult::Rejected { .. } => Weight::one(),
```
alex-mysten, line 1204: is there a risk of flagging txs that were rejected for reasons that were not the user's fault as spam and blocking their IPs? for instance, the system overload gate?

aschran: Yes, this is more of a philosophical question. if you are sending a lot of requests that get rejected during system overload and those requests can't be charged, do they count as spam? I decided yes, but could see an argument for no. a few tallies don't matter, so this only becomes a big problem if you are hammering us with low-price requests during an overload scenario, which is sort of the thing we are trying to prevent.

mystenmark (approved):

PR #27069, opened 2026-06-25
Suppress transaction resubmission

alex-mysten:
## Description

Suppress duplicate in-flight transactions

## Test plan

- End-to-end endpoint test: a transaction resubmitted while its first copy is still in a block is rejected at `handle_submit_transaction`.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/authority_server.rs
+++ b/crates/sui-core/src/authority_server.rs
@@ -7,21 +7,22 @@ use async_trait::async_trait;
 use fastcrypto::traits::KeyPair;
 use futures::{TryFutureExt, future};
 use itertools::Itertools as _;
+use moka::sync::Cache;
 use mysten_common::ZipDebugEqIteratorExt;
 use mysten_common::{assert_reachable, debug_fatal};
 use mysten_metrics::spawn_monitored_task;
 use prometheus::{
-    Gauge, Histogram, HistogramVec, IntCounter, IntCounterVec, Registry,
+    Gauge, Histogram, HistogramVec, IntCounter, IntCounterVec, IntGauge, Registry,
     register_gauge_with_registry, register_histogram_vec_with_registry,
     register_histogram_with_registry, register_int_counter_vec_with_registry,
-    register_int_counter_with_registry,
+    register_int_counter_with_registry, register_int_gauge_with_registry,
 };
 use std::{
     collections::HashSet,
     io,
     net::{IpAddr, SocketAddr},
     sync::Arc,
-    time::{Duration, SystemTime},
+    time::{Duration, Instant, SystemTime},
 };
 use sui_network::{
     api::{Validator, ValidatorServer},
@@ -43,7 +44,7 @@ use sui_types::sui_system_state::SuiSystemState;
 use sui_types::traffic_control::{ClientIdSource, Weight};
 use sui_types::{
     base_types::ObjectID,
-    digests::TransactionEffectsDigest,
+    digests::{TransactionDigest, TransactionEffectsDigest},
     error::{SuiErrorKind, UserInputError},
 };
 use sui_types::{
@@ -202,7 +203,9 @@ pub struct ValidatorServiceMetrics {
     num_rejected_tx_during_overload: IntCounterVec,
     submission_rejected_transactions: IntCounterVec,
     submission_suppressed_already_processed: IntCounterVec,
-    submission_suppressed_recently_in_block: IntCounterVec,
+    submission_suppressed_recently_submitted: IntCounterVec,
+    recently_submitted_cache_size: IntGauge,
+    recently_submitted_resubmission_interval: Histogram,
     connection_ip_not_found: IntCounter,
     forwarded_header_parse_error: IntCounter,
     forwarded_header_invalid: IntCounter,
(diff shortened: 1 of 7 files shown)


crates/sui-core/src/consensus_adapter.rs
```diff
@@ -242,6 +256,10 @@ impl ConsensusAdapter {
             metrics,
             submit_semaphore: Arc::new(Semaphore::new(max_pending_local_submissions)),
             inflight_slot_freed_notify,
+            recently_in_block: Cache::builder()
+                .max_capacity(100_000)
```
akichidis: Let's use a constant for that as well. I guess if we reach the cache capacity it will start evicting the oldest entries?

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -250,6 +268,47 @@ impl ConsensusAdapter {
...
+                ConsensusTransactionKey::Certificate(digest),
+            ) = key
+            {
+                debug!(?digest, "recording transaction in recently-in-block cache");
+                self.recently_in_block.insert(*digest, ());
+            }
+        }
+        self.metrics
+            .recently_in_block_cache_size
+            .set(self.recently_in_block.entry_count() as i64);
```
akichidis: I can only assume this is a cheap/cached count - it won't do any live counting.

crates/sui-core/src/consensus_adapter.rs
akichidis: I believe 100K might be a bit high. We can leave as is for now, but let's gather metrics and consider

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -250,6 +268,45 @@ impl ConsensusAdapter {
         self.num_inflight_transactions.load(Ordering::Relaxed)
     }
 
+    /// Returns true if this validator recently included `digest` in a consensus block.
+    /// Used by the submission endpoint to suppress duplicate resubmissions
+    /// before the transaction is committed.
+    pub fn was_recently_in_block(&self, digest: &TransactionDigest) -> bool {
```
akichidis: Is it possible to return the timestamp or the remaining ttl? I wonder whether we could capture some additional metrics on the frequency of resubmission. Maybe we could add a timestamp as the value on the entry of the cache?

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -250,6 +268,45 @@ impl ConsensusAdapter {
...
+
+    /// Records that `digests` have been included in a consensus block, so subsequent
+    /// resubmissions can be suppressed until the entries expire.
+    fn record_in_block(&self, transaction_keys: &[SequencedConsensusTransactionKey]) {
+        for key in transaction_keys {
+            if let SequencedConsensusTransactionKey::External(
+                ConsensusTransactionKey::Certificate(digest),
+            ) = key
+            {
+                self.recently_in_block.insert(*digest, ());
```
akichidis: Will any overrides keep pushing the ttl?

akichidis (approved):

mwtian (commented):
One alternative, keep a cache in submit transaction handler and allow a unique transaction every 1 / 5 / 10 sec. I think it is unnecessary to only track transactions already in consensus output. Pre-submission and inflight transations should be deduped too.

crates/sui-core/src/consensus_adapter.rs
alex-mysten: Yep - 100k might be bigger than needed, can tune this later - and yes, the oldest would be evicted, which should be fine for reducing spam

crates/sui-core/src/consensus_adapter.rs
alex-mysten: it's an approximate counter, not a scan

crates/sui-core/src/consensus_adapter.rs
alex-mysten: yes - reubmission will push the ttl.

crates/sui-core/src/authority_server.rs
```diff
@@ -390,8 +415,17 @@ pub struct ValidatorService {
     client_id_source: Option<ClientIdSource>,
     gasless_limiter: GaslessRateLimiter,
     admission_queue: Option<AdmissionQueueContext>,
+    /// Digests submitted within the last `RECENT_SUBMISSION_TTL` (value: when recorded), to drop
+    /// duplicate resubmissions before they reach consensus.
+    recently_submitted: Cache<TransactionDigest, Instant>,
 }
 
+/// Window during which a given transaction is allowed into consensus at most once.
+const RECENT_SUBMISSION_TTL: Duration = Duration::from_secs(1);
```
mystenmark: can we put this in the node config? it might end up needing tuning

crates/sui-core/src/authority_server.rs
```diff
@@ -390,8 +415,17 @@ pub struct ValidatorService {
...
+
+/// Eviction past this only weakens best-effort suppression, not correctness.
+const RECENT_SUBMISSION_MAX_CAPACITY: u64 = 100_000;
```
mystenmark: don't these two values together suggest we are planning for 100k submissions per second? seems maybe unnecessary

crates/sui-core/src/authority_server.rs
```diff
@@ -915,6 +959,31 @@ impl ValidatorService {
                 continue;
             }
 
+            // Allow a given transaction into consensus at most once per `RECENT_SUBMISSION_TTL`,
+            // dropping duplicate resubmissions (including pre-submission and in-flight) early.
+            if let Some(recorded_at) = self.recently_submitted.get(&tx_digest) {
```
mwtian: Should we also check elapsed time from recorded_at?

mwtian (approved):

crates/sui-core/src/authority_server.rs
alex-mysten: Done!

crates/sui-core/src/authority_server.rs
alex-mysten: Updated

crates/sui-core/src/authority_server.rs
alex-mysten: Lowered threshold

mystenmark (approved):

PR #27071, opened 2026-06-25
[1/6] System object availability check and out-of-band retry

lxfind:
## Description

Stack of 6; the base. Adds the in-execution check that detects when a system object a transaction reads has not yet caught up on the local node, together with the out-of-band retry that follows.

`TemporaryStore::is_system_object_available` — an inherent method, called directly on the store during execution rather than through a resolver trait — consults a per-transaction `system_object_versions` map threaded in via the executor. The check loads the object at exactly the required version with a single keyed read; its absence means the local node has not yet committed that version. When that happens it records `ExecutionRetryError::SystemObjectUnavailable { object_id, version }` on `InnerTemporaryStore` instead of committing: execution still runs to completion and produces effects, but the authority discards them, waits for the object to reach the required version (new `ObjectCacheRead::notify_read_system_object_at_version` primitive), and re-sends the cert to the execution driver. To form the wait key the authority recovers the object's initial shared version from the epoch start config (which records every system object), so the signal only needs to carry the object id and target version. A `debug_assert` in the commit path guards that a retrying transaction is never committed.

Today the only such system object is the accumulator root, whose version is sourced from the transaction's assigned accumulator version; [3/6] generalizes the source to an arbitrary set of system objects. Inert until a consumer calls the check ([5/6]).

## Test plan

Infrastructure with no consumer yet; exercised end to end by `object_balance_tests::test_object_balance_withdraw_stress` once the stack is complete, which drives the not-caught-up retry-and-re-enqueue path under simulation across multiple seeds. Reading a system object with no assigned version is an invariant violation guarded by `debug_fatal`.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -96,6 +98,14 @@ pub struct TemporaryStore<'backing> {
...
     /// [`invariants::InvariantChecker`].
     invariants: InvariantChecker,
+
+    /// Recorded when execution determines the transaction must be retried later rather than
+    /// committed. Execution still runs to completion (the triggering native also raises a Move
+    /// error); this signal is carried out on `InnerTemporaryStore` so the authority can discard the
+    /// effects and re-enqueue. A `RefCell` rather than a lock: execution is single-threaded, but the
+    /// condition is detected behind `&self` (`ChildObjectResolver`), so the field must be
+    /// interior-mutable.
+    retry_request: RefCell<Option<ExecutionRetryError>>,
```
tnowacki: Can this be replaced by a `OnceCell`?

lxfind: Yes! Fixed.

crates/sui-core/src/authority.rs
```diff
@@ -1919,6 +1922,47 @@ impl AuthorityState {
...
+        let init_shared_version = epoch_store
+            .epoch_start_config()
+            .system_object_initial_shared_version(object_id)
+            .expect("system object must be registered in the epoch start config");
+        let full_object_id = FullObjectID::Consensus((object_id, init_shared_version));
+        let cache_reader = self.get_object_cache_reader().clone();
+        let scheduler = self.execution_scheduler.clone();
+        let cert = certificate.clone();
+        let execution_env = execution_env.clone();
+        let epoch_store = epoch_store.clone();
```
mystenmark, line 1976: should we have a counter metric tracking this?

mystenmark (approved):

PR #27072, opened 2026-06-25
[2/3] Add object-funds sufficiency check API to the temporary store

lxfind:
## Description

Stack of 3 (builds on [1/3] #27071).

Adds the temporary-store API for checking object-funds withdrawal sufficiency during execution, ahead of the Move native that calls it in [3/3]. Threads the transaction's assigned accumulator version and an `UnsettledObjectFundsRead` reader into execution, and implements `check_object_funds_sufficiency` on `TemporaryStore`: it verifies the local accumulator root has caught up to the required version (recording a retry and returning `Unknown` otherwise — see [1/3]), then checks the owner's settled balance net of unsettled in-commit withdrawals and earlier in-transaction withdrawals. The object-funds checker gains `record_object_funds_withdraws` and an `UnsettledObjectFundsRead` impl.

No caller yet; the Move native that invokes the check lands in [3/3].

## Test plan

The check has no caller until [3/3], so it is covered transitively by `object_balance_tests::test_object_balance_withdraw_stress` once the stack is complete — that test concurrently withdraws from owned, party-owned, and shared vault objects, exercising the settled-balance and net-withdrawal arithmetic this PR adds.

## Release notes

Check each box that your change affects. If none of the boxes relate to your changes, release notes aren't required.

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

PR #27073, opened 2026-06-25
[3/3] Check object-funds withdraw sufficiency inside the Move VM

lxfind:
## Description

Stack of 3 (builds on [2/3] #27072... ); final PR that activates in-VM object-funds withdrawal checking.

A `check_sufficient_object_funds` native, called from `withdraw_from_object`, consults the temporary store's in-execution funds check ([2/3]). On insufficient funds it aborts — a committed failure; when the accumulator root has not caught up it aborts after the store has recorded a retry, so the authority discards the effects and re-enqueues ([1/3]). Gated by the `check_object_funds_withdraw_in_execution` feature flag (protocol version 128, devnet and testnet); when off, the prior post-execution checker path runs unchanged. When on, successful withdrawals are recorded as unsettled so later transactions in the same consensus commit, reading the same not-yet-advanced settled balance, account for them.

Note: the in-execution unsettled-withdrawal read is not yet replay-deterministic (see the FIXME in `temporary_store.rs`); the flag therefore stays off on mainnet pending that fix.

## Test plan

`object_balance_tests::test_object_balance_withdraw_stress` runs concurrent withdrawals from owned, party-owned, and shared vault objects under simulation (protocol v128, so the in-VM path is active), driving both the in-VM insufficiency abort and the not-caught-up retry-and-re-enqueue paths; it passes across multiple seeds. The `object_funds_checker` integration tests are updated to reflect that over-budget withdrawals now fail in the VM with a `MoveAbort` rather than retrying. Protocol-config snapshots are regenerated for version 128.

## Release notes

- [x] Protocol: Adds protocol version 128, enabling in-execution object-funds withdrawal sufficiency checks on devnet and testnet (off on mainnet).
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


lxfind:
Superseded by the restored 5-PR stack (see #27071 / #27080).

PR #27078, opened 2026-06-26
[4/6] Add is_system_object_available to the runtime object resolver

lxfind:
## Description

Stack of 6 (builds on [3/6]). Adds `RuntimeObjectResolver::is_system_object_available` and threads the `system_object_versions` map from [3/6] through the executor into the `TemporaryStore`, which implements the check: when a required system object has not yet reached its assigned version locally it records an `ExecutionRetryError` (carrying the object's full id) so the transaction is retried via the post-exec mechanism from [1/6]; otherwise it records the read (id, version, digest) for later inclusion in effects ([5/6]). This is the interface half of the old "waitable system object framework" PR. Inert until a consumer calls the check ([6/6]).

## Test plan

Framework with no consumer yet; exercised end to end by the object-balance withdraw stress test once the stack is complete, which drives the not-caught-up retry path through this check. Invariant-violation paths (reading a system object with no assigned version, or missing locally) are guarded by `debug_fatal`.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


lxfind:
Folded into [1/5] (#27071): `is_system_object_available` and its fields now live on `TemporaryStore` as an inherent method (called directly during execution, not through the `RuntimeObjectResolver` trait). The version-assignment generalization remains in [3/5] (#27104). This PR is no longer needed.

PR #27080, opened 2026-06-26
[7/7] Check object funds withdraw sufficiency inside the Move VM

lxfind:
## Description

Second half of the restructured stack, on top of #27403: moves the object-funds withdraw sufficiency check into the Move VM, superseding the post-execution `ObjectFundsChecker` wait/re-execute flow. Gated by `check_object_funds_withdraw_in_execution` (protocol v132, devnet and testnet).

`funds_accumulator::withdraw_from_object` now calls a `track_object_funds_withdrawal` native that keeps a per-`(owner, type)` running balance in the object runtime. When the in-transaction balance falls short, it reads the settled balance at the transaction's accumulator version through the temporary store — via the blocking `load_implicitly_read_system_object` from #27403, so a root that has not caught up locally simply waits instead of unwinding — and discounts unsettled withdrawals from earlier transactions in the same consensus commit, read from the standalone `UnsettledObjectWithdrawals` store (extracted out of the old checker in the first commit; now populated on fullnodes too, since the check affects execution results on every executing node). Insufficiency is an ordinary deterministic Move abort surfaced as `ExecutionErrorKind::InsufficientObjectFundsForWithdraw`.

This reworks the old [6/7] #27335 + [7/7] content onto the block-wait base: the retry-based availability plumbing the old version carried (the `SYSTEM_OBJECT_NOT_AVAILABLE_LOCALLY` unwind, the retry request on the temporary store, and the non-`Available` availability variants) is gone entirely — unavailability is handled by the blocking read and never surfaces in execution results.

Known caveat carried over from the old PR: isolated replay does not reconstruct in-commit unsettled withdrawals (`unsettled = 0`; see the TODO in `temporary_store.rs`), which can diverge from the original execution. Mainnet enablement is blocked on resolving that.

## Test plan

- The `object_funds_withdraw.move` transactional test covers the in-VM check at the Move level: withdrawals within the settled balance succeed, exceeding it aborts with the insufficient-funds error, and in-transaction deposits cover later withdrawals without consulting the store.
- sui-core unit and integration tests cover the `UnsettledObjectWithdrawals` store and its interaction with execution: net unsettled amounts recorded from effects, GC at checkpoint commit, multiple withdrawals against one account, withdraw-and-deposit in the same transaction, and zero-amount withdrawals.
- E2e tests in `address_balance_tests.rs` cover the flag-on validator path (sufficient and insufficient withdrawals, the accumulator root recorded as a read-only consensus object in effects), effects equivalence with the flag off, and the dry-run/simulate counterparts. These are simtests and are blocked from running by the known msim `block_in_place` compile issue inherited from #27403; they pass once the simulator fix lands.

## Release notes

- [x] Protocol: Adds `check_object_funds_withdraw_in_execution` (protocol v132, enabled on devnet and testnet): object-funds withdraw sufficiency is checked inside the Move VM at the point of withdrawal, and insufficient withdrawals abort the transaction with `InsufficientObjectFundsForWithdraw` instead of being cancelled before execution.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


crates/sui-core/src/authority.rs
```diff
@@ -2096,6 +2101,13 @@ impl AuthorityState {
 
         let tracking_store = TrackingBackingStore::new(self.get_backing_store().as_ref());
 
+        // Held across execution so the in-execution object-funds check can read unsettled
+        // withdrawals from the current consensus commit, and reused by the post-execution check
+        // below.
+        let object_funds_checker = self.object_funds_checker.load();
```
lxfind: Why does object_funds_checker need to be held earlier?

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -210,6 +214,39 @@ mod checked {
...
+    /// single point every executed transaction passes through, after the final execution result is
+    /// settled (including error overrides such as the conservation checks) and before effects are
+    /// built.
+    ///
+    /// The two mismatch directions are not equally dangerous. The unwind error without a retry
+    /// request would let a node-local, transient condition surface as committed failed effects,
+    /// diverging from validators that have caught up. A retry request without the unwind error is
+    /// benign in production: the authority keys retries off the request alone, so effects are
+    /// discarded and the transaction re-executed either way. `debug_fatal` panics in tests and
+    /// alerts in release; behavior past it stays keyed off the retry request.
```
lxfind: Seems a bit too verbose and unnecessary to explain

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -134,3 +138,39 @@ pub fn withdraw_from_accumulator_address(
...
+        .protocol_config
+        .check_object_funds_withdraw_in_execution()
+    {
+        return Ok(NativeResult::ok(context.gas_used(), smallvec![]));
+    }
+
+    // The `?` also propagates the system-object-unavailable retry unwind (minted by the temporary
+    // store together with its retry request) through the outer `PartialVMResult` channel — a
+    // node-local condition, deliberately not a `NativeResult` abort, which would be a deterministic
+    // Move-level outcome. Insufficiency, by contrast, is such an outcome, so it aborts normally.
```
tnowacki: These feels brittle. Should we be handling this error locally here?

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -24,6 +25,9 @@ use crate::{
 
 const E_OVERFLOW: u64 = 0;
 const E_ADDRESS_BALANCE_NOT_ENABLED: u64 = 1;
+/// Public so the adapter's `convert_vm_error` can recognize this abort and convert it to
+/// `ExecutionErrorKind::InsufficientObjectFundsForWithdraw`.
+pub const E_OBJECT_FUNDS_INSUFFICIENT: u64 = 2;
```
tnowacki: I feel this should be named slightly differently since it is _not_ a normal abort code. It is a sort of time-travelly explosion. A comment feels insufficient for noticing it does something drastically different.

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -1105,6 +1131,61 @@ impl RuntimeObjectResolver for TemporaryStore<'_> {
...
+        type_: &TypeTag,
+    ) -> PartialVMResult<u128> {
+        // The accumulator root must be available at the version this transaction requires.
+        // `check_system_object_available` errors if it has no assigned version (an invariant
+        // violation for a withdrawal, which is always sequenced against the root) and, when the
+        // root has not caught up, records the retry request and returns the unwind error as one
+        // atomic act — propagate it unmodified (`?`) so it reaches the calling native intact, and
+        // let the waitable-system-object framework handle the wait and re-enqueue.
+        // Settlement advances the accumulator root only after the per-account fields are mutated, so
+        // the root reaching its required version implies every account is settled to at least it.
```
tnowacki: I find all of the error handling here really really confusing. I don't like passing through things like that are intended on branching behavior (if I am understanding correctly)

I feel like a custom error enum of some sort. Or just an option? feels safer?

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -1105,6 +1131,61 @@ impl RuntimeObjectResolver for TemporaryStore<'_> {
...
+        // Availability guarantees a required version is recorded for the accumulator root.
+        let required_version = self
+            .system_object_versions
+            .get(&SUI_ACCUMULATOR_ROOT_OBJECT_ID)
+            .copied();
+
+        // Settled balance at the required accumulator version.
+        let settled = AccumulatorRootValue::load(self, required_version, owner, type_)
+            .map_err(|e| {
+                PartialVMError::new(StatusCode::STORAGE_ERROR).with_message(e.to_string())
```
tnowacki: Storage Error is an invariant violation. Is that what we want here?

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -210,6 +214,30 @@ mod checked {
         )
     }
 
+    /// The temporary store's retry request and the `SYSTEM_OBJECT_NOT_AVAILABLE_LOCALLY` unwind
```
tnowacki: I'm confused by this function. It doesn't seem to be enforcing anything? it just has a debug fatal? Is there supposed to be an invariant violation or some other kind of mutation?

sui-execution/latest/sui-adapter/src/error.rs
```diff
@@ -28,6 +30,18 @@ pub(crate) fn convert_vm_error_impl(
             // this is a Move VM invariant violation, the code should always be there
             ExecutionErrorKind::VMInvariantViolation
         }
+        // The abort minted by the `check_sufficient_object_funds` native on an overdraw. The raw
+        // code is unambiguous within `0x2::funds_accumulator`: the module's own Move-level aborts
+        // use clever-encoded error constants (or `EOverflow = 0`), so a plain
+        // `E_OBJECT_FUNDS_INSUFFICIENT` there can only come from the native. Converted to a
+        // dedicated kind so clients see a legible status instead of an opaque abort code.
```
tnowacki: I think this comment could be a bit more succinct

crates/sui-framework/packages/sui-framework/sources/funds_accumulator.move
```diff
@@ -123,6 +124,11 @@ native fun withdraw_from_accumulator_address<T: store>(
     value: u256,
 ): T;
 
+// Checks during execution that `owner` holds at least `limit` of `T`. If the funds are not yet
+// sufficient, signals that the transaction should be retried later instead of committed. No-op
+// unless the `check_object_funds_withdraw_in_execution` feature is enabled.
+native fun check_sufficient_object_funds<T: store>(owner: address, limit: u256);
```
tnowacki: We probably want to rename this so it is more obvious that it is keeping track of all of the withdrawals asked for thus far in the transaction.

tnowacki: That is the name feels like a "check" when it is really more of a "track"

lxfind:
Reworked onto the block-wait base as #27459.

PR #27082, opened 2026-06-26
Stress-test large object-balance withdrawals

mystenmark:
## Description

The composite stress workload had no coverage for large or boundary-value object-balance withdrawals, which are the inputs most likely to expose overflow or minting bugs in the funds accumulator.

Adds an `object_balance_large_withdraw` operation that withdraws boundary values (1, `u32::MAX`±1, `i64::MAX`/sign-bit, `u64::MAX`±1, `TOTAL_SUPPLY_MIST`±1, and a uniform full-range draw) from a balance pool that is deliberately never seeded. Since the pool's `Balance<T>` is always zero, every such withdrawal must fail with insufficient funds — a success would mean funds were minted from nothing, which trips a `debug_fatal!`.

## Test plan

Extends the `test_composite_workload` simtest to exercise the operation and assert it is both attempted and never succeeds.

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/src/workloads/composite/mod.rs
+++ b/crates/sui-benchmark/src/workloads/composite/mod.rs
@@ -10,9 +10,9 @@ pub use operations::{
     ALIAS_ADD, ALIAS_REMOVE, ALIAS_TX, ALL_OPERATIONS, AccumulatorBalanceRead,
     AddressBalanceDeposit, AddressBalanceOverdraw, AddressBalanceWithdraw, AuthenticatedEventEmit,
     CoinReservationWithdraw, INVALID_ALIAS_TX, ImmutableObjectRead, ObjectBalanceDeposit,
-    ObjectBalanceOverdraw, ObjectBalanceWithdraw, OperationDescriptor, RandomnessRead,
-    SharedCounterIncrement, SharedCounterRead, TestCoinAddressDeposit, TestCoinAddressWithdraw,
-    TestCoinMint, TestCoinObjectWithdraw,
+    ObjectBalanceLargeWithdraw, ObjectBalanceOverdraw, ObjectBalanceWithdraw, OperationDescriptor,
+    RandomnessRead, SharedCounterIncrement, SharedCounterRead, TestCoinAddressDeposit,
+    TestCoinAddressWithdraw, TestCoinMint, TestCoinObjectWithdraw,
 };
 use rand::seq::SliceRandom;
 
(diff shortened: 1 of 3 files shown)


crates/sui-benchmark/src/workloads/composite/mod.rs
```diff
@@ -1589,6 +1609,33 @@ impl Workload<dyn Payload> for CompositeWorkload {
             info!("Created balance pool {:?}", self.balance_pool);
         }
 
+        if init_requirements.contains(&InitRequirement::CreateEmptyBalancePool) {
+            // This pool is deliberately never seeded, so its `Balance<T>` stays zero for
+            // every type. Withdrawals from it must always trigger IFFW.
```
alex-mysten, line 1614: Would it also make sense to apply the same tests to a pool where we have just enough balance?

crates/sui-benchmark/tests/simtest.rs
```diff
@@ -1687,6 +1688,29 @@ mod test {
                 "expected at least one accumulator balance read"
             );
 
+            // Large withdrawals target an always-empty pool, so they must be exercised
+            // but must never succeed (a success would be a minting bug).
```
alex-mysten: "minting bug" repeated in comments

crates/sui-benchmark/src/workloads/composite/operations.rs
```diff
@@ -1007,6 +1013,123 @@ impl Operation for ObjectBalanceOverdraw {
...
+
+    fn resource_requests(&self) -> Vec<ResourceRequest> {
+        vec![ResourceRequest::EmptyBalancePool]
+    }
+
+    fn init_requirements(&self) -> Vec<InitRequirement> {
+        vec![InitRequirement::CreateEmptyBalancePool]
+    }
+
+    fn apply(
```
alex-mysten, line 1075: Did we have something that checks that occasionally operations must produce valid transactions? possibly outside of composite, if we have a string of failed txs we cannot continue the workload?

I may be misremembering

mystenmark: yes there are minimum success counts in `test_composite_workload`

crates/sui-benchmark/src/workloads/composite/mod.rs
mystenmark: we could but it probably needs a separate PR/operation, since it will have to figure out whether to expect success or not

crates/sui-benchmark/tests/simtest.rs
mystenmark: man i told claude to take that out, not a minting bug!

PR #27083, opened 2026-06-26
Cancel pre-submission when a transaction is already processed by consensus

arun-koshy:
## Description

When a transaction is submitted for a consensus position (mfp) that consensus has already
processed, the adapter used to skip the already-processed check, resubmit, and return a position
for a redundant duplicate — one that doesn't track the placement that actually determines the
transaction's outcome. This removes `skip_processed_checks` so the check runs on the position path
too, and reports the transaction per-transaction as a retriable `TransactionProcessing` instead of
that meaningless position. Reporting it per-transaction (rather than a top-level error) keeps a
multi-transaction request from failing wholesale.

Follow-up (separate PR): on `TransactionProcessing`, have the transaction driver drop to
`wait_for_effects` by digest instead of resubmitting.

## Test plan

New `test_submit_batched_transactions_with_already_processing`: a batch containing an
already-processing transaction returns `[Rejected { TransactionProcessing }, Submitted]` with the
RPC still succeeding. The single-tx and ping paths are covered by the existing consensus-adapter
and submit-transaction tests.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/consensus_adapter.rs
+++ b/crates/sui-core/src/consensus_adapter.rs
@@ -512,10 +512,7 @@ impl ConsensusAdapter {
         };
 
         // Skip submission if the tx is already processed via consensus output or
-        // checkpoint state sync. Unlike before, this check now also runs when a caller
-        // is waiting for a consensus position (mfp): rather than redundantly submitting
-        // an already-processed transaction, we cancel the pre-submission request and
-        // tell the caller the transaction is already processing.
+        // checkpoint state sync.
         let already_processed =
             self.check_processed_via_consensus_or_checkpoint(&transaction_keys, epoch_store);
         if let Some(method) = already_processed {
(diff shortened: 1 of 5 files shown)


crates/sui-core/src/consensus_adapter.rs
```diff
@@ -621,27 +632,36 @@ impl ConsensusAdapter {
...
+            // the retry loop is cancelled cleanly.
+            let processed_waiter = self
+                .processed_notify(transaction_keys.clone(), epoch_store)
+                .boxed();
+            let processed_via_notify;
+            guard.processed_method = match select(processed_waiter, submit_fut.boxed()).await {
+                Either::Left((observed, _submit_fut)) => {
+                    processed_via_notify = true;
+                    observed
+                }
```
alex-mysten, line 648: What happens when one tx in a soft bundle is already processed?

arun-koshy: It's all or nothing for suppression, so if only one tx is already processed the bundle is still submitted in full to consensus

mystenmark: but note that we remove already executed/processed transactions from soft bundles in authority_server

mwtian: I think each component can take the simplest approach to handle soft bundle edge cases. Removing processed from authority server is simple. But in consensus adapter it is not easy to stop when one or more txns are processed, then drop processed from transactions already submitted to consensus.

mwtian (approved):
Please review the last commit.

PR #27087, opened 2026-06-26
Passing of accumulator_version should not be conditional on network

lxfind:
## Description 

It is a footgun that the value of accumulator_version passed to execution layer depends on the network.
We should always pass it, and in places where we check it we check the network.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


alex-mysten (approved):
This is definitely easier to reason about!

PR #27089, opened 2026-06-26
[move-ide] Fixes to auto-import and quick-fix

awelc:
## Description 

This PR provides two fixes to the auto-imports algorithm:
- avoid suggesting member auto-imports when the typed prefix already resolves to an in-scope module, since normal module::member completion is sufficient and actually inserting auto-import would result in an immediately "unused" `use` declaration
- for unresolved module-qualified chains like `foo::Bar`, allow auto-importing the module with `use pkg::foo` instead of incorrectly importing the member with `use pkg::foo::Bar`

It also fixes quick-fix algorithm to correctly handle module-prefixed (but without module import) accesses similarly to how it's done for auto-imports (e.g., for `foo::Bar` where module is not yet imported, include `use pkg::foo` in the auto-fix)

## Test plan 

New test has been added and test output has been regenerated

### DIFF
--- a/external-crates/move/crates/move-analyzer/src/code_action.rs
+++ b/external-crates/move/crates/move-analyzer/src/code_action.rs
@@ -257,6 +257,7 @@ pub fn access_chain_autofix_actions_for_error(
                                 two_element_access_chain_autofixes(
                                     code_actions,
                                     symbols,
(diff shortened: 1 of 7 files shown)


cgswords (approved):
LGTM

PR #27091, opened 2026-06-26
[move-ir] Add constant support

tnowacki:
## Description 

- Added syntax and support for constants
- The access syntax is `copy(<Name>)`, which is in line with the source language and avoids conflicts with structs or functions. 
- For the value syntax two notable callouts:
  - No integer types without a suffix
  - `vector` uses `()` to pun off of the normal expression syntax, but it isn't really true since `vector` can be applied to any expression (like a function)
- Borrow syntax can then be `&<Name>` whenever we add that

## Test plan 

- New tests
- Notably:
  - A few testing that the BCS bytes carry no type information e.g. `const C: vector<bool> = 0`
  - A few round trip tests
 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-ir-to-bytecode-syntax/src/syntax.rs
```diff
@@ -1995,6 +2056,107 @@ fn parse_variant_decl(
...
+    check_no_modifier(native, "native", "constant")?;
+    check_no_modifier(entry, "entry", "constant")?;
+
+    consume_token(tokens, Tok::Const)?;
+    let name_loc = current_token_loc(tokens);
+    let name = ConstantName(parse_name(tokens)?);
+    if !is_constant_name(name.0.as_str()) {
+        return Err(ParseError::InvalidToken {
+            location: name_loc,
+            message: "constant names must begin with an uppercase letter".to_string(),
```
cgswords, line 2090: This message is accurate but we could consider suggesting it is all-caps.

cgswords: Similarly, we could attempt to enforce it is all-caps, which would propagate into usage sites by virtue of name resolvers.

external-crates/move/crates/move-ir-to-bytecode-syntax/src/syntax.rs
```diff
@@ -1995,6 +2056,107 @@ fn parse_variant_decl(
...
+//     <i: U8 | U16 | U32 | U64 | U128 | U256>  // integer literals require a suffix
+//     "true" | "false"
+//     <a: AccountAddress>
+//     <b: ByteArray>                           // h"..", lowered to vector<u8>
+//     "vector" "(" <elems: Comma<ConstantValue>> ")"
+// }
+// Each value is self-describing, so the parsed `MoveValue` is independent of the constant's
+// declared type.
+fn parse_constant_value(tokens: &mut Lexer) -> Result<MoveValue, ParseError<Loc, anyhow::Error>> {
+    // `vector(elem, ...)` — element values are self-describing; there is no
```
cgswords, line 2116: `vector(elem, ...)` instead of `vector[elem, ...]`?

external-crates/move/crates/move-ir-to-bytecode-syntax/src/syntax.rs
```diff
@@ -1995,6 +2056,107 @@ fn parse_variant_decl(
...
+        tokens.advance()?; // consume `vector`
+        consume_token(tokens, Tok::LParen)?;
+        let elems = parse_comma_list(tokens, &[Tok::RParen], parse_constant_value, true)?;
+        consume_token(tokens, Tok::RParen)?;
+        return Ok(MoveValue::Vector(elems));
+    }
+    // `@addr` — named or literal address, mirroring address expressions.
+    if tokens.peek() == Tok::At {
+        tokens.advance()?;
+        return Ok(MoveValue::Address(parse_account_address(tokens)?));
```
cgswords, line 2128: Could map the result

external-crates/move/crates/move-ir-to-bytecode-syntax/src/syntax.rs
```diff
@@ -1995,6 +2056,107 @@ fn parse_variant_decl(
...
+    }
+    // A bare integer (no suffix) lexes as `U64Value`; constant values must be
+    // self-describing, so require an explicit suffix.
+    if tokens.peek() == Tok::U64Value && !tokens.content().ends_with("u64") {
```
cgswords, line 2132: Surely we want to parse something larger here?

cgswords (commented):
Some questions about syntax and behavior, but they should be quick fixes.

external-crates/move/crates/move-ir-to-bytecode-syntax/src/syntax.rs
tnowacki: This is to pun off the expression syntax, if you recall is `"vector" <expr>`, which typically comes to `vector(<expr>, <expr>, ...)`

external-crates/move/crates/move-ir-to-bytecode-syntax/src/syntax.rs
tnowacki: In the IR non-suffixed integers are currently limited to u64

external-crates/move/crates/move-ir-to-bytecode-syntax/src/syntax.rs
tnowacki: I don't like all-caps because we do upper camel case for error codes in the source language

cgswords (approved):

PR #27093, opened 2026-06-27
[chore] enhance consensus handler metrics

akichidis:
## Description 

Small changes on the consensus handler metrics

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -639,14 +639,14 @@ impl AuthorityMetrics {
             .unwrap(),
             consensus_handler_processed: register_int_counter_vec_with_registry!(
                 "consensus_handler_processed",
-                "Number of transactions processed by consensus handler",
-                &["class"],
+                "Number of transactions processed by consensus handler, sliced by class, commit outcome (accepted/rejected) and block author",
+                &["class", "outcome", "authority"],
                 registry
             ).unwrap(),
             consensus_handler_transaction_sizes: register_histogram_vec_with_registry!(
                 "consensus_handler_transaction_sizes",
-                "Sizes of each type of transactions processed by consensus handler",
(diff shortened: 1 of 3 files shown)


crates/sui-core/src/authority.rs
```diff
@@ -639,14 +639,14 @@ impl AuthorityMetrics {
             .unwrap(),
             consensus_handler_processed: register_int_counter_vec_with_registry!(
                 "consensus_handler_processed",
-                "Number of transactions processed by consensus handler",
-                &["class"],
+                "Number of transactions processed by consensus handler, sliced by class, commit outcome (accepted/rejected) and block author",
+                &["class", "outcome", "authority"],
```
mwtian: Cardinality will be non trivial. I wonder if we should have a separate metric with ["outcome", "authority"] labels.

crates/sui-core/src/authority.rs
```diff
@@ -639,14 +639,14 @@ impl AuthorityMetrics {
...
                 registry
             ).unwrap(),
             consensus_handler_transaction_sizes: register_histogram_vec_with_registry!(
                 "consensus_handler_transaction_sizes",
-                "Sizes of each type of transactions processed by consensus handler",
-                &["class"],
+                "Sizes of each type of transactions processed by consensus handler, sliced by class, commit outcome (accepted/rejected) and block author",
+                &["class", "outcome", "authority"],
```
mwtian: This cardinality will be even larger. Maybe just an aggregated size metric by authority?

mwtian (approved):

PR #27094, opened 2026-06-29
sui: retain fullnode address from network config

amnn:
## Description

Currently `sui` CLI always overrides the socket address configured for the fullnode to set it to `0.0.0.0`. It offers no way to set the socket address to bind to in the `sui start` invocation, and it ignores the socket address that's part of the network config.

The ideal solution would be to allow the socket address to be provided along with the port in the `sui start` invocation (similar to how it works for other RPCs exposed), but that would be an interface breaking change.

Instead, this change makes the `sui` CLI retain the socket address from the network config, if there is one, as a way to provide this information without breaking the current interface.

## Test plan

Manually tested.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [x] CLI: `sui start` respects the fullnode RPC socket address that's part of its network config, if one is provided.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/src/sui_commands.rs
+++ b/crates/sui/src/sui_commands.rs
@@ -1027,21 +1027,16 @@ async fn start(
         swarm_builder = swarm_builder.with_data_ingestion_dir(dir.clone());
     }
 
-    let mut fullnode_rpc_address = sui_config::node::default_json_rpc_address();
-    fullnode_rpc_address.set_port(fullnode_rpc_port);
-
-    if no_full_node {
+    let fullnode_rpc_address = if no_full_node {
         swarm_builder = swarm_builder.with_fullnode_count(0);
+        let mut fullnode_rpc_address = sui_config::node::default_json_rpc_address();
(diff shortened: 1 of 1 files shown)


stefan-mysten (approved):
I don't fully recall, but I don't think there are any tests that cover this; so wondering if we could add a few.

amnn:
I did some digging on how easy this would be to test -- unfortunately not that easy in a robust way: practically, the feature that is being added is that I can choose to bind only on loopback, rather than listen from any IP, but that is quite hard to test in a reproducible (automated) way.

Really what would give some confidence is if this logic was built in an obvious, easy to understand, uniform way. If we did that, then the desire to test the ability to customize the socket address would be low. Unfortunately, that's not the case, but I don't think any tests we could feasibly add would do anything meaningful to our confidence.

Obviously I am noting this though, for jellyfish.

PR #27096, opened 2026-06-29
[move] Tighten up some additional binary bounds

tzakian:
## Description 

These were already constrained by explicit bounds on other pools, but not explicitly. This makes those bounds explicit.

## Test plan 

CI + some local testing/backtesting

---

## Release notes

- [X] Protocol: Introduces a new protocol version that adds explicit bounds for binary pools that were otherwise implicitly constrained before. No user-visible change.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


dariorussi (approved):
I like it and we talked about it too, fine for me to go when you want

PR #27100, opened 2026-06-29
[consensus adapter] improve logging

mwtian:


crates/sui-core/src/consensus_adapter.rs
```diff
@@ -663,7 +663,11 @@ impl ConsensusAdapter {
                     tx_consensus_positions.send(Err(make_processing_error(guard.processed_method)));
             }
         }
-        debug!("{transaction_keys:?} processed by consensus");
+        let processed_via = match guard.processed_method {
+            ProcessedMethod::Consensus => "consensus output",
```
akichidis: Maybe instead we could derive the Debug marco for the `ProcessedMethod` enum?

akichidis (approved):

crates/sui-core/src/consensus_adapter.rs
mwtian: Adding another method to return string.

PR #27101, opened 2026-06-29
Batch per-transaction work in the consensus commit-handler filter stage

andll:
## Description

Follow-up to #27052. The consensus commit handler is the bottleneck at high TPS; `filter_consensus_txns` did two things once per transaction that are cheaper batched per commit:

- **Consensus-tx-status writes** — each `set_consensus_tx_status` took the status-cache write lock and notified waiters under it. Now collected and flushed once per commit (single lock acquisition, notifications outside the lock). Filter never reads status back, so end-of-filter visibility is equivalent.
- **Owned-object lock reads** — `try_acquire_owned_object_locks_post_consensus` read existing locks (quarantine + DB) per transaction. The cross-commit locks are constant for a commit's duration (new locks are written only at commit end), so they're prefetched once and passed into the per-tx conflict check. A loop-invariant hoist; intra-commit conflict logic is unchanged.

## Test plan

Both changes are behavior-preserving (equivalence argued in the commit messages). Covered by existing tests in `consensus_handler` and `consensus_tx_status_cache`.

## Release notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/sui-core/src/authority/authority_per_epoch_store.rs
@@ -1970,12 +1970,33 @@ impl AuthorityPerEpochStore {
             .get_owned_object_locks(&tables, obj_refs)
     }
 
+    /// Batched read of existing owned-object locks, returned as a map containing only
+    /// the refs that are currently locked. The consensus commit handler prefetches the
+    /// cross-commit lock state (constant for the duration of a commit) once with this,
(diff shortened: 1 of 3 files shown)


mystenmark (approved):
Looks good - lets make sure to run a split cluster test though to catch any divergences in behavior

PR #27102, opened 2026-06-29
`CursorToken` in proto

wlmyng:
## Description 

Use `proto` instead of `bcs` for `CursorToken` on the wire.

## Test plan 

Needed to migrate some cursor-related tests off transactional test runner and into Rust.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-rpc-cursor/proto/sui/rpc/cursor/v1/cursor.proto
```diff
@@ -0,0 +1,20 @@
...
+    // Ledger-history query this cursor was minted for. 1 = checkpoints, 2 = transactions, 3 =
+    // events.
+    uint32 query_type = 1;
+
+    // Whether `position` is a matched row returned to the client (1 = item) or a scan frontier the
+    // server reached (2 = boundary).
+    uint32 kind = 2;
+
+    uint64 checkpoint = 3;
+    uint64 position = 4;
```
tpham-mysten: nit: Maybe add some docs for this?

crates/sui-rpc-cursor/proto/sui/rpc/cursor/v1/cursor.proto
```diff
@@ -0,0 +1,20 @@
...
+
+// Pagination cursor for the bitmap-backed ledger-history endpoints.
+message CursorToken {
+    // Ledger-history query this cursor was minted for. 1 = checkpoints, 2 = transactions, 3 =
+    // events.
+    uint32 query_type = 1;
+
+    // Whether `position` is a matched row returned to the client (1 = item) or a scan frontier the
+    // server reached (2 = boundary).
+    uint32 kind = 2;
```
tpham-mysten: Could we make these proper proto enums rather than uint32 with the values documented in a comment? 

```
enum QueryType {
    QUERY_TYPE_UNSPECIFIED = 0;
    QUERY_TYPE_CHECKPOINTS = 1;
    ...
}
```

It keeps the value names in the schema instead of a comment that can drift.

WDYT?

crates/sui-rpc-cursor/src/lib.rs
```diff
@@ -73,3 +75,65 @@ impl CursorToken {
...
+    fn from_proto(value: u32) -> Option<Self> {
+        Some(match value {
+            1 => CursorKind::Item,
+            2 => CursorKind::Boundary,
+            _ => return None,
+        })
+    }
+}
+
+impl From<&CursorToken> for ProtoCursorToken {
```
tpham-mysten, line 117: Now that we have the proto type, do we still need the hand-written `CursorToken` + the `From`/`TryFrom`? Should we just use the ProtoCursorToken in all places?

crates/sui-rpc-cursor/proto/sui/rpc/cursor/v1/cursor.proto
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+package sui.rpc.cursor.v1;
+
+// Pagination cursor for the bitmap-backed ledger-history endpoints.
+message CursorToken {
```
tpham-mysten, line 26: qq on `AlphaLedgerGrpcReader`: should it expose and use the decoded `CursorToken` in StreamPage, FrameKind, etc, so we don't need to deal with `Bytes` object.

tpham-mysten (commented):
Looking great! Some questions on usage

crates/sui-rpc-cursor/proto/sui/rpc/cursor/v1/cursor.proto
wlmyng: I don't think it's needed on the read/ pagination side as we don't make additional modifications, just pass through to graphql client

Only time we need to touch cursortoken is when spoofing one on subscription side

crates/sui-rpc-cursor/src/lib.rs
amnn: I think it's still useful to have the manual type to deal with the fact that everything in the proto type is optional.

crates/sui-indexer-alt-e2e-tests/src/transaction.rs
amnn, line 1: I have a slight preference to not include helpers to build specific transaction shapes in central places -- we've done this a couple of times, and it's always devolved into a mess, because then people feel compelled to use or adapt the helper than just write the PTB that they need for their test, so the helpers end up growing lots of strange extra configs/optionals, etc.

crates/sui-indexer-alt-e2e-tests/tests/graphql_checkpoint_transactions_tests.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Query a checkpoint's `transactions` connection. Returns `None` if the checkpoint does not exist.
+async fn transactions(
+    cluster: &FullCluster,
+    seq: u64,
+    first: Option<u32>,
+    last: Option<u32>,
+    after: Option<String>,
+    before: Option<String>,
+) -> anyhow::Result<Option<graphql::Connection<TxNode>>> {
+    let query = format!(
```
amnn: Use a variable for `seq` as well, instead of interpolating it into the query.

crates/sui-indexer-alt-e2e-tests/tests/graphql_epoch_transactions_tests.rs
```diff
@@ -0,0 +1,193 @@
...
+    cluster: &FullCluster,
+    epoch_id: u64,
+    first: Option<u32>,
+    last: Option<u32>,
+    after: Option<String>,
+    before: Option<String>,
+) -> anyhow::Result<Option<graphql::Connection<TxNode>>> {
+    let query = format!(
+        r#"query($first: Int, $last: Int, $after: String, $before: String) {{
+            epoch(epochId: {epoch_id}) {{
```
amnn: ditto

amnn (commented):
Just passing through -- agreed with @tpham-mysten's comments that using enums instead of plain integers in the proto format seems wise! Will leave it to @tpham-mysten to give the final stamp.

crates/sui-rpc-cursor/proto/sui/rpc/cursor/v1/cursor.proto
evan-wall-mysten: Do we know if cursors will always be a `u64`? If not, `query_type` and `position` can be replaced with `oneof` to allow cursors with different types.

wlmyng: if that changes we can always add a new field to accommodate

crates/sui-indexer-alt-e2e-tests/src/transaction.rs
wlmyng: I get that, but for the new tests added these really are identical components ... added a comment discouraging subsequent builders from extending these helpers if they don't directly satisfy their requirements

crates/sui-rpc-cursor/tests/bootstrap.rs
```diff
@@ -0,0 +1,65 @@
...
+    // Sort files by name to have deterministic codegen output
+    fds.file.sort_by(|a, b| a.name.cmp(&b.name));
+
+    // No gRPC services in this crate's protos, so client/server stub generation is off and no
+    // FileDescriptorSet is emitted — these messages are only ever serialized as opaque cursor
+    // bytes.
+    if let Err(error) = tonic_prost_build::configure()
+        .build_client(false)
+        .build_server(false)
+        .bytes(".")
```
tpham-mysten: nit: is this necessary?

tpham-mysten (approved):
Looking great! Thanks @wlmyng

crates/sui-rpc-cursor/tests/bootstrap.rs
wlmyng: i don't think so, in fact gonna replace the tonic-prost-build with just prost-build

PR #27105, opened 2026-06-30
[sui-framework] Scratchpad

tnowacki:
## Description 

- Adds `sui::scratch`--an ephemeral, per-transaction key-value store. Entries are kept in a native context extension where all entires are dropped at the end of the transaction (nothing is persisted). As such, the `K` and `V` types need `drop` but do not need `store`. 
- Access is gated by a `sui::scratch::Permit<K>` which is issued from an `internal::Permit`. As such, access is controlled by the module that defines `K`.
- Natives are backed by a new ScratchRuntime extension (in-memory `BTreeMap<address, (Type, Value)>` built fresh per transaction). This seemed a bit simpler than stuffing it into the `ObjectRuntime`.
- Added a high limit of 16 entries per PTB command (based on the maximum number of PTB commands allowed)

## Test plan 

- New unit tests, including test_scenario support
- A simple adapter transactional test for additional testing of the ephemerality. 
- Verified the gas cost for `read` always exceeds `ReadRef`

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-framework/packages/sui-framework/sources/scratch.move
```diff
@@ -0,0 +1,131 @@
...
+/// `internal::Permit<K>`.
+module sui::scratch;
+
+/// A `Permit<K>` gates access to all entries keyed by values of type `K`.
+/// It is issued from an `internal::Permit<K>`, allowing the module that defines `K` to control
+/// all access to scratch entries.
+public struct Permit<phantom K: copy + drop>() has drop;
+
+/// Stand-in parent address used when hashing scratch keys.
+const DUMMY_ROOT: address = @0;
```
cgswords, line 23: Is there a reason to prefer this over the package address or similar?

crates/sui-framework/packages/sui-framework/sources/scratch.move
```diff
@@ -0,0 +1,131 @@
...
+/// Issues a `Permit<K>` from the privileged `internal::Permit<K>`, granting access to the
+/// scratch entries keyed by values of type `K`.
+public fun permit<K: copy + drop>(_: internal::Permit<K>): Permit<K> {
+    Permit()
+}
+
+/// Adds the `key`-`value` pair to the scratch store. Requires a `Permit<K>` for the key type.
+/// Aborts with `EEntryAlreadyExists` if there is already an entry for `key`, regardless of its
+/// value type.
+public fun add<K: copy + drop, V: drop>(_: Permit<K>, key: K, value: V, _: &mut TxContext) {
```
cgswords: Should there be an Option variant that returns V if it exists?

cgswords: I am surprised about this argument order. I would have expected:
```
ctxt.add(permit, k, v)
```

Also, I am surprised `K` has to match. What about this:
```
public fun add<P, K: copy + drop, V: drop>(_: Permit<P>, key: K, value: V, _: &mut TxContext) 
```

The permit gives ownership and the key is flexible, so you could use a string:
```
ctxt.add(permit, "call", 0);
```

crates/sui-framework/packages/sui-framework/sources/scratch.move
```diff
@@ -0,0 +1,131 @@
...
+public fun exists_with_type<K: copy + drop, V: drop>(_: Permit<K>, key: K, _: &TxContext): bool {
+    exists_with_type_impl<V>(hash_type_and_key(key))
+}
+
+// === Macro Functions ===
+
+/// A wrapper for `add` that constructs the `Permit<$K>` directly.
+/// Aborts with `EEntryAlreadyExists` if there is already an entry for `$key`, regardless of its
+/// value type.
+public macro fun internal_add<$K: copy + drop, $V: drop>(
```
cgswords, line 78: I am surprised about this. Usage as:
```
struct Called() has copy, drop;

fun record_called(ctxt: &mut TxContext, name: String) {
    let called = Called();
    let mut called_vec: Vec<String> =
        if (called.internal_exists(ctxt)) called.internal_read(ctxt)
        else vec![];
    called_vec.push(name);
    called.internal_add(ctxt, called_vec);
}
```

If you did it the other way, with `ctxt.interal_add(k, v)`, it may read better:

```
struct Called() has copy, drop;

fun record_called(ctxt: &mut TxContext, name: String) {
    let called = Called();
    let mut called_vec: Vec<String> =
        if (ctxt.internal_exists(called)) ctxt.internal_read(called)
        else vec![];
    called_vec.push(name);
    ctxt.internal_add(called, called_vec);
}
```

This reads more like "we're scratching stuff down for the transaction" to me.

crates/sui-framework/packages/sui-framework/sources/scratch.move
tzakian: I don't dislike this... although I get why Todd is going for this API -- to make it feel and behave as close to dynamic fields as is possible.

There are tricks we could play to allow for bifurcating based on `K + key` (instead of using dummy root for hash type and key get that off of the `K`)

tzakian: IMO if we went this route this should expose much more of a map-like API (i.e., implement `index`).

crates/sui-framework/packages/sui-framework/sources/scratch.move
tzakian: +1 and in general I think having the tx context as first argument (also for the public functions) would be good for this reason as well.

crates/sui-framework/packages/sui-framework/sources/scratch.move
tnowacki: What is `index`?

tnowacki: There will be yes, I just wanted to start with the core function set first (since these need natives its easier for the natives to abort rather than returning the option)

crates/sui-framework/packages/sui-framework/sources/scratch.move
tnowacki: > This reads more like "we're scratching stuff down for the transaction" to me.

I went back and forth on this, but am switching to this because I think it provides a good intuition for where this data is stored and how long it lives. So switching shortly!

crates/sui-framework/packages/sui-framework/sources/scratch.move
tnowacki: Which package? That of `K`? The downside for that would mean we need to hit a native function on each call (though we could stuff it in the `Permit` to "cache" the result). 
I don't think it really matters as far as collisions or safety is concerned. The type `K` is included in the hash and provides all the separation we need.

crates/sui-framework/packages/sui-framework/sources/tx_context.move
```diff
@@ -3,6 +3,23 @@
 
 module sui::tx_context;
 
+use sui::scratch;
+
+/// Allows calling `sui::scratch::add` as `.scratch_add(permit, key, value)`.
+public use fun scratch::add as TxContext.scratch_add;
```
tnowacki, line 9: @tzakian @cgswords this is based on the feedback for method syntax and making `TxContext` the first argument. I figured just having "add" and "read" on `TxContext` without the `scratch_` prefix would be a bit confusing

crates/sui-framework/packages/sui-framework/sources/scratch.move
```diff
@@ -0,0 +1,133 @@
...
+/// Removes the entry bound to `key` and returns its value. Requires a `Permit<K>` for the key type.
+/// Aborts with `EEntryDoesNotExist` if there is no entry for `key`.
+/// Aborts with `EEntryTypeMismatch` if the entry exists, but its value is not of type `V`.
+public fun remove<K: copy + drop, V: drop>(_: &mut TxContext, _: Permit<K>, key: K): V {
+    remove_impl(hash_type_and_key(key))
+}
+
+/// Returns true if and only if the scratch store has an entry for `key`, without regard to the
+/// value type. Requires a `Permit<K>` for the key type.
+public fun exists<K: copy + drop>(_: &TxContext, _: Permit<K>, key: K): bool {
```
tnowacki, line 63: Each function is gated by `Permit` out of caution. Especially since a lot of the intended use cases might be simple markers/counters, we probably want to make reads just as guarded and intentional as writes.

crates/sui-framework/packages/sui-framework/tests/scratch_tests.move
```diff
@@ -0,0 +1,167 @@
...
+// `internal::Permit`s (which is what the `internal_*` macros do under the hood).
+public struct Marker() has copy, drop;
+public struct WrappedU8(u8) has copy, drop;
+public struct WrappedBool(bool) has copy, drop;
+
+#[test]
+fun add_read_remove_exists() {
+    let mut ctx = tx_context::dummy();
+
+    assert!(!scratch::internal_exists!<Marker>(&ctx, Marker()));
```
tzakian: More just to double check -- can you write these as `ctx.internal_exists!<Marker>(Marker())`?

tzakian (approved):
Looks great!

cgswords (approved):
LGTM!

crates/sui-adapter-transactional-tests/tests/scratch/scratch.move
```diff
@@ -0,0 +1,47 @@
...
+// that the store is cleared between transactions (a fresh `ScratchRuntime` per transaction).
+
+//# init --addresses test=0x0 --accounts A
+
+//# publish
+module test::scratch_test {
+    public struct Marker() has copy, drop;
+
+    // Full add / read / remove lifecycle within a single transaction.
+    public entry fun add_read_remove(ctx: &mut TxContext) {
```
damirka: why is it entry?

crates/sui-framework/packages/sui-framework/tests/scratch_tests.move
```diff
@@ -0,0 +1,167 @@
...
+    // the slot is free again after removal, so the key can be re-added (with a new value/type)
+    scratch::internal_add!<Marker, bool>(&mut ctx, Marker(), true);
+    assert!(scratch::internal_exists!<Marker>(&ctx, Marker()));
+    assert!(!scratch::internal_exists_with_type!<Marker, u64>(&ctx, Marker()));
+    assert!(scratch::internal_exists_with_type!<Marker, bool>(&ctx, Marker()));
+    assert_eq!(scratch::internal_read!<Marker, bool>(&ctx, Marker()), true);
+}
+
+#[test]
+#[expected_failure(abort_code = sui::scratch::EEntryAlreadyExists)]
```
damirka: nit: we're slowly switching (and recommending) unified, inlined `#[test, expected_failure(...)]`

crates/sui-adapter-transactional-tests/tests/scratch/scratch.move
tnowacki: bad copy pasta

crates/sui-framework/packages/sui-framework/tests/scratch_tests.move
tnowacki: personally I don't like that because it muddies git  blame and lines changed but... sure

crates/sui-framework/packages/sui-framework/tests/scratch_tests.move
tnowacki: It is `ctx.scratch_internal_exists` but I'll switch to that to be sure

PR #27107, opened 2026-06-30
Deduplicate transaction submissions in authority service handler

mwtian:
## Summary

- add a ValidatorService in-flight digest guard so concurrent duplicate transaction submissions are rejected before consensus submission
- generalize repeated-transaction validation across batch and soft-bundle submit requests
- demote completed in-flight digests into the recent-submission cache under the same lock before removal
- add coverage for repeated request transactions, recent duplicate handling, guard lifecycle behavior, and handler-level concurrent duplicate submission

## Validation

- `cargo fmt --all -- --check`
- `SUI_SKIP_SIMTESTS=1 cargo test -p sui-core inflight_guard_tests --lib`
- `SUI_SKIP_SIMTESTS=1 cargo test -p sui-core test_submit_batched_transactions_with_repeated_transaction --lib`
- `SUI_SKIP_SIMTESTS=1 cargo test -p sui-core test_submit_soft_bundle_with_repeated_transaction --lib`
- `SUI_SKIP_SIMTESTS=1 cargo test -p sui-core test_concurrent_duplicate_submission_rejected_as_inflight --lib`
- `cargo xclippy` from `crates/sui-core`
- `git diff --check -- crates/sui-core/src/unit_tests/submit_transaction_tests.rs`

### DIFF
--- a/crates/sui-core/src/authority_server.rs
+++ b/crates/sui-core/src/authority_server.rs
@@ -341,7 +341,7 @@ impl ValidatorServiceMetrics {
             .unwrap(),
             inflight_transactions: register_int_gauge_with_registry!(
                 "validator_service_inflight_transactions",
-                "Number of transaction digests from inflight submit requests",
+                "Number of transactions from inflight submit requests",
                 registry,
             )
             .unwrap(),
@@ -805,10 +805,16 @@ impl ValidatorService {
 
             // A request must not repeat a transaction.
             if !request_digests.insert(tx_digest) {
-                return Err(SuiErrorKind::UserInputError {
+                let error: SuiError = SuiErrorKind::UserInputError {
                     error: UserInputError::RepeatedTransactions { digest: tx_digest },
                 }
-                .into());
+                .into();
+                // Reject individual repeated transactions in batch.
+                if is_soft_bundle_request {
+                    return Err(error);
+                }
+                results[idx] = Some(SubmitTxResult::Rejected { error });
+                continue;
             }
 
             // Soft bundles require all transactions to use the same gas price.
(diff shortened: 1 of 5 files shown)


crates/sui-core/src/authority_server.rs
```diff
@@ -1431,6 +1473,94 @@ impl ValidatorService {
...
+/// the in-flight set and demoted into `recently_submitted` cache, so resubmissions
+/// arriving shortly after the handler returns are still suppressed.
+struct InflightTransactionsGuard {
+    // Handle to inflight map and recently submitted cache.
+    inflight: Arc<Mutex<HashSet<TransactionDigest>>>,
+    recently_submitted: Cache<TransactionDigest, Instant>,
+    window: Duration,
+    metrics: Arc<ValidatorServiceMetrics>,
+    /// Digests this request successfully acquired (in acquisition order).
+    owned: Vec<TransactionDigest>,
```
akichidis: Do we consider it cheaper than holding a `HashSet` - mostly for the search part? Otherwise we could switch to a `HashSet`?

akichidis: nit: Maybe `acquired` would better capture the intention instead of `owned` ?

crates/sui-core/src/authority_server.rs
```diff
@@ -328,6 +331,20 @@ impl ValidatorServiceMetrics {
...
+                "validator_service_submission_suppressed_inflight",
+                "Number of submitted transactions suppressed because the same transaction was \
+                 already being handled by a concurrent in-flight submit request",
+                &["req_type"],
+                registry,
+            )
+            .unwrap(),
+            inflight_transactions: register_int_gauge_with_registry!(
+                "validator_service_inflight_transactions",
+                "Number of transaction digests from inflight submit requests",
```
akichidis: maybe: "Number of transactions from inflight submit requests"

akichidis: shall we also pass in the request type?

akichidis (approved):

crates/sui-core/src/authority_server.rs
mwtian: Let's see if we need to track inflight per request type later.

PR #27114, opened 2026-06-30
sui-indexer-alt-framework: surface underlying source in ingestion fetch-retry log

evan-wall-mysten:
## Description

The ingestion fetch-retry warning logs reqwest's terse `error sending request`, which hides why a fetch failed (request timeout vs connection refused vs reset). Today the cause is only inferable from the line's duration. Walk the error's `source()` chain at the log site so the underlying cause (`operation timed out`, `Connection refused`, …) is printed inline.

A small de-duping helper is used rather than `anyhow`/`eyre` `{:#}` because those join the chain without de-duping, and the object_store/reqwest wrappers already inline their child's message.

## Test plan

New unit test covers the helper: an already-inlined fragment is not repeated, and a genuinely-new leaf cause is appended.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/metrics.rs
+++ b/crates/sui-indexer-alt-framework/src/metrics.rs
@@ -332,7 +332,12 @@ impl IngestionMetrics {
         reason: &str,
         error: Error,
     ) -> backoff::Error<Error> {
-        warn!(checkpoint, reason, "Retrying due to error: {error}");
+        warn!(
+            checkpoint,
+            reason,
+            "Retrying due to error: {}",
+            error_with_sources(&error)
+        );
 
         self.total_ingested_transient_retries
             .with_label_values(&[reason])
@@ -803,6 +808,33 @@ impl CheckpointLagMetricReporter {
     }
 }
 
+/// Render an error together with its [`std::error::Error::source`] chain. Several wrappers
+/// (object_store, reqwest, anyhow) already inline their source's message in their own `Display`,
+/// so we skip any source whose message is already contained in its parent's message to avoid
+/// repetition, and append only genuinely-new sources. This surfaces the underlying `io::Error`
+/// ("Connection refused" / "operation timed out") that is otherwise hidden behind reqwest's terse
+/// "error sending request" leaf.
+fn error_with_sources(mut err: &dyn std::error::Error) -> String {
+    let mut out = String::new();
+    let mut next = err.to_string();
+
+    let mut prefix = "";
+    while let Some(src) = err.source() {
+        err = src;
+        let msg = err.to_string();
+        if !next.contains(&msg) {
+            out.push_str(prefix);
+            out.push_str(&next);
+            prefix = ": ";
+            next = msg;
+        }
+    }
+
+    out.push_str(prefix);
+    out.push_str(&next);
+    out
+}
+
 #[cfg(test)]
 pub(crate) mod tests {
     use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-framework/src/metrics.rs
```diff
@@ -820,4 +845,57 @@ pub(crate) mod tests {
...
+                Some(&self.0)
+            }
+        }
+
+        // The already-inlined "error sending request" fragment is not repeated, but the
+        // genuinely-new "operation timed out" cause (hidden behind the terse leaf) is appended.
+        let err = Inlining(Terse(Leaf));
+        assert_eq!(
+            error_with_causes(&err),
+            "Error performing GET: error sending request: operation timed out"
```
wlmyng, line 879: it seems like just the leaf error provides the most meaningful information, as we already know we're making GET requests, and encountered an issue that caused a retry (given the context of ingestion fetch-retry)

Among the possible leaf errors, they largely boil down to checking if the ingestion source is reachable and has the particular checkpoint, and if the ingestor is properly configured to access those checkpoints

evan-wall-mysten: I was trying to come up with something that was general purpose where all info from the causal chain would get logged. The goal is to add more detail to logs like this to help identify what is actually wrong.

Image

The test was inspired by this which is why claude made the test related to an HTTP response.

crates/sui-indexer-alt-framework/src/metrics.rs
```diff
@@ -820,4 +845,57 @@ pub(crate) mod tests {
...
+        impl fmt::Display for Inlining {
+            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                write!(f, "Error performing GET: {}", self.0)
+            }
+        }
+        impl StdError for Inlining {
+            fn source(&self) -> Option<&(dyn StdError + 'static)> {
+                Some(&self.0)
+            }
+        }
```
amnn: Use `thiserror` for this.

```rust
#[derive(thiserror::Error, Debug)]
#[error("operation timed out")]
struct Leaf;

#[derive(thiserror::Error, Debug)]
#[error("error sending request")]
struct Terse(#[source] Leaf);

#[derive(thiserror::Error, Debug)]
#[error("Error performing GET: {0}")]
struct Inlining(#[source] Terse);
```

crates/sui-indexer-alt-framework/src/metrics.rs
```diff
@@ -803,6 +808,26 @@ impl CheckpointLagMetricReporter {
...
+/// so we skip any fragment that is already a suffix of the accumulated string to avoid repetition,
+/// and append only genuinely-new causes. This surfaces the underlying `io::Error`
+/// ("Connection refused" / "operation timed out") that is otherwise hidden behind reqwest's terse
+/// "error sending request" leaf.
+fn error_with_causes(error: &dyn std::error::Error) -> String {
+    let mut out = error.to_string();
+    let mut source = error.source();
+    while let Some(cause) = source {
+        let msg = cause.to_string();
+        if !out.ends_with(&msg) {
```
amnn: Do you know that the embedded error message is always at the end? There's a risk that there's punctuation following it that will mess this strategy up. The following approach would address that:

```rust
fn error_with_causes(mut err: &dyn std::error::Error) -> String {
    let mut out = String::new();
    let mut next = err.to_string();

    let mut prefix = "";
    while let Some(src) = err.source() {
        err = src;
        let msg = err.to_string();
        if !next.contains(&msg) {
            out.push_str(prefix);
            out.push_str(&next);
            prefix = ": ";
            next = msg;
        }
    }

    out.push_str(prefix);
    out.push_str(&next);
    out
}
```

Keeps track of the next message that's going to be added to the output, and discards children that are contained in it. This prevents accidentally quadratic behaviour on the output size while being able to check for descendant content anywhere within the message.

Also, nit: shouldn't use the term `cause` to refer to the `source` because `cause()` is a distinct (deprecated) API on `std::error::Error`.

amnn (approved):
Looks reasonable!

crates/sui-indexer-alt-framework/src/metrics.rs
evan-wall-mysten: I'll change it since I'm not sure what every possible format could be. Also renamed `cause` to `source`.

PR #27115, opened 2026-06-30
[zklogin] enable v2 verify on devnet

joyqvq:
## Description 

depends on https://github.com/MystenLabs/fastcrypto/pull/937

also https://github.com/MystenLabs/fastcrypto/pull/987

add a protocol config enum:
0 -> v1 vk verify only (current behavior)
1 -> try verify with v2 vk first, then fallback to v1 vk (this enables)
2 -> v2 vk verify only (for the future when we fully deprecate v1)
## Test plan 

```
SUI_SKIP_SIMTESTS=1 cargo nextest run -p sui-e2e-tests zklogin_v1_v2_migration_scenario_test --run-ignored all
```

```
┌────────┬──────────┬───────────────────────────────┬─────────────────────────┐
│ prover │   flag   │             path              │         result          │
├────────┼──────────┼───────────────────────────────┼─────────────────────────┤
│ v2     │ enabled  │ v2 VK first → passes, no      │ is_ok +                 │
│        │          │ fallback                      │ assert_no_fallback()    │
├────────┼──────────┼───────────────────────────────┼─────────────────────────┤
│ v1     │ enabled  │ v2 VK first → fails →         │ is_ok +                 │
│        │          │ fallback to v1 VK → passes    │ assert_fallback()       │
├────────┼──────────┼───────────────────────────────┼─────────────────────────┤
│ v2     │ disabled │ only v1 VK tried → rejected,  │ is_err +                │
│        │          │ v2 never attempted            │ assert_no_fallback()    │
├────────┼──────────┼───────────────────────────────┼─────────────────────────┤
│ v1     │ disabled │ only v1 VK → passes           │ is_ok +                 │
│        │          │                               │ assert_no_fallback()    │
└────────┴──────────┴───────────────────────────────┴─────────────────────────┘
```
also tested against localnet for log capture and v1 and v2 verify both worked

```
fastcrypto_zkp::bn254::zk_login_api: [zkLogin] V2 verify failed
(env=Test, iss=https://oauth.sui.io): GeneralError("Groth16 proof verify failed"), falling back to V1

┌─────────────┬─────────────────────────┬───────────────────────────────────┬─────────┐
│    Proof    │         Prover          │              Digest               │ Status  │
├─────────────┼─────────────────────────┼───────────────────────────────────┼─────────┤
│             │                         │                                   │ success │
│ v1 (prover- │ prover-dev.mystenlabs.c │ ECUDYNfktX8Gtt76twAQ2eJTT5qavSYn7 │ , check │
│ dev)        │ om/v1                   │ qPbqwSa1TBZ                       │ point   │
│             │                         │                                   │ 307     │
├─────────────┼─────────────────────────┼───────────────────────────────────┼─────────┤
│             │                         │                                   │ success │
│ v2 (prover- │ prover-dev-v2.mystenlab │ F9igv99XYc9FMvdtEEZcRgHpZvZHakhiK │ , check │
│ dev-v2)     │ s.com/v1                │ aEtfjLVhA8W                       │ point   │
│             │                         │                                   │ 391     │
└─────────────┴─────────────────────────┴───────────────────────────────────┴─────────┘
```
---

## Release notes

- [] Protocol:  
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-types/src/zk_login_authenticator.rs
```diff
@@ -45,26 +45,28 @@ struct ZkLoginCachingParams {
...
+    zklogin_auth_v2: bool,
 }
 
 impl ZkLoginAuthenticator {
     /// The caching key for zklogin signature, it is the hash of bcs bytes of
-    /// ZkLoginInputs || max_epoch || flagged_pk_bytes. If any of these fields
+    /// ZkLoginInputs || max_epoch || flagged_pk_bytes || zklogin_auth_v2. If any of these fields
     /// change, zklogin signature is re-verified without using the caching result.
-    fn get_caching_params(&self) -> ZkLoginCachingParams {
+    fn get_caching_params(&self, zklogin_auth_v2: bool) -> ZkLoginCachingParams {
```
joyqvq: we technically do not need to care about the version in cache bc the cache is by per epoch, and a flag enablement is on epoch boundary only. adding it here as a layer of defence, so that a v2-proof success cached under the flag-on can never satisfy a lookup under flag-off.

crates/sui-core/src/signature_verifier.rs
```diff
@@ -63,6 +63,8 @@ struct ZkLoginParams {
     pub supported_providers: Vec<OIDCProvider>,
     /// The environment (prod/test) the code runs in. It decides which verifying key to use in fastcrypto.
     pub env: ZkLoginEnv,
+    /// Flag to determine whether zkLogin v2 circuit verification is accepted.
+    pub zklogin_auth_v2: bool,
```
benr-ml: we will also need to deprecate v1 later

crates/sui-types/src/zk_login_authenticator.rs
benr-ml: maybe use `version: u8` instead

crates/sui-indexer-alt-graphql/src/api/types/epoch.rs
```diff
@@ -558,6 +558,20 @@ impl Epoch {
         .map(Some)
     }
 
+    /// The protocol version this epoch runs at.
+    ///
+    /// Returns `None` when the epoch start information is not available in scope.
+    pub(crate) async fn protocol_version(
```
benr-ml, line 564: nit - surprised there isn't a similar function already

benr-ml (approved):

crates/sui-indexer-alt-graphql/src/api/types/epoch.rs
joyqvq: i dont seem to find one! cc @bmwill maybe know if there is a better way to read the protocol version for graphql?

PR #27079, opened 2026-06-26
[4/7] Record read system objects in effects, and reproduce them on replay

lxfind:
## Description

Stack of 7 (builds on [3/7], merged). Emits the system objects read during execution into the transaction effects as `ReadOnlyRoot` unchanged-consensus objects, and reconstructs their versions when executing from effects via `assign_versions_from_effects`. This makes checkpoint execution deterministic for transactions that gate on system-object versions, without re-deriving them from live consensus assignment.

## Test plan

Exercised by replay/checkpoint paths and, end to end, by the object-balance withdraw stress test once the stack is complete.

## Release notes

Check each box that your change affects. If none of the boxes relate to your changes, release notes are not required.

- [ ] Protocol:
- [ ] Nodes (Validators and Full nodes):
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

PR #27118, opened 2026-06-30
[combined view] Check object funds withdraw sufficiency inside the Move VM (stack of 7)

lxfind:
## Description

**Combined view only — do not merge.** This PR bundles the full 5-PR stack against `main` so the whole change ("check object-funds withdraw sufficiency inside the Move VM") can be read in one diff. Review and merge happen on the individual PRs below, in order:

1. #27071 — System object availability check and out-of-band retry
2. #27077 — Rename ChildObjectResolver to RuntimeObjectResolver
3. #27104 — Generalize accumulator version into a system object versions map
4. #27079 — Record read system objects in effects, and reproduce them on replay
5. #27080 — Check object funds withdraw sufficiency inside the Move VM

In short: a transaction's object-funds withdrawal is checked for sufficiency inside the VM against the settled balance (net of in-commit-unsettled and in-transaction withdrawals). If the local node has not caught up to the accumulator-root version the transaction was sequenced against, the VM records an out-of-band retry; the authority discards the effects, waits for the object, and re-enqueues. The system-object versions a transaction may read are pinned via consensus version assignment and recorded into effects for replay. Gated by `check_object_funds_withdraw_in_execution` (protocol v130, devnet + testnet).

## Test plan

See the individual PRs. End to end, `object_balance_tests::test_object_balance_withdraw_stress` exercises the in-VM insufficiency abort and the not-caught-up retry-and-re-enqueue paths under simulation.

## Release notes

Tracked on #27080 (adds protocol version 130). No separate release note for this combined view.

- [x] Protocol: Adds protocol version 130, enabling in-execution object-funds withdrawal sufficiency checks on devnet and testnet (off on mainnet).
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -134,3 +142,43 @@ pub fn withdraw_from_accumulator_address(
...
+            E_OBJECT_FUNDS_INSUFFICIENT,
+        )),
+        // Not ready yet: the temporary store has recorded the retry request on its inner store.
+        // Abort to unwind the VM; execution still completes and produces effects, but the authority
+        // discards them and re-enqueues the transaction once it sees `retry_request` set, so this
+        // code is only a placeholder to keep error conversion well-formed.
+        ObjectFundsSufficiency::Unknown => Ok(NativeResult::err(
+            context.gas_used(),
+            E_OBJECT_FUNDS_NOT_SETTLED,
+        )),
```
tnowacki: This scares me a bit. I wonder if we can instead replace it with some sort of error code that panics if it ever reaches the end of execution? Even if we need to change NativeResult a bit, the extra security of knowing we can't panic here on a bug seems worthwhile

crates/sui-framework/packages/sui-framework/sources/funds_accumulator.move
```diff
@@ -92,6 +92,7 @@ public(package) fun withdraw_from_object<T: store>(obj: &mut UID, limit: u256):
         EObjectFundsWithdrawNotEnabled,
     );
     let owner = obj.to_address();
+    check_sufficient_object_funds<T>(owner, limit);
```
tnowacki: I don't think this is sufficient. In that I think we need to check this in the withdrawal, no? Like this isn't sufficient in that it doesn't actually perform the withdrawal since you can call `withdraw_from_object` multiple times and then go to redeem.

PR #27121, opened 2026-07-01
[move][adapter] Add support for `init` on upgrade

tzakian:
## Description 

This adds support for `init` on upgrade according to the following rules:
1. `init` functions are only called in new modules introduced in that upgrade;
2. It is an error (now) to add an `init` function to an existing module during an upgrade;
3. `init` functions in new modules at upgrade are called in the order they appear in the `Upgrade` command.

Init functions in upgrades introduce restrictions on the global LUMPY linkage as follows:
1. If the package being upgraded (of any version) doesn't occur in the PTB then any `Upgrade` commands that have `init` functions introduce `exact` constraints (just like `Publish`).
2. If package being upgraded (of any version) occurs in the PTB then for any `(original_id, version_id)` as defined in the `Upgrade` command either: a. It is not in the existing LUMPY linkage, and a `original_id -> exact(version_id)` constraint is introduced; or b. It is in the existing LUMPY linkage, in which case `LUMPY[original_id].id` must equal `version_id`.

Ignore exact protocol versions -- once we're ready to land this I'll do that whole song and dance.

## Test plan 

Added a bunch of new tests, updated existing tests (and also made these protocol-versioned tests to memorialize them).

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-adapter-transactional-tests/tests/init/upgrade_existing_module_add_init_error.move
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//# init --addresses v0=0x0 v1=0x0 --accounts A
```
tnowacki, line 7: Maybe a small comment on the nature of the error

crates/sui-adapter-transactional-tests/tests/upgrade/upgrade_init_lumpy_examples.move
```diff
@@ -0,0 +1,159 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Additional end-to-end coverage for upgrade-init LUMPY examples
```
tnowacki: Lumpy or LUMPY?

tnowacki: I thought we were Lumpy now.

tnowacki: But maybe... LuMpY

crates/sui-adapter-transactional-tests/tests/init/upgrade_existing_module_add_init_error.move
```diff
@@ -0,0 +1,15 @@
...
+
+//# publish --upgradeable --sender A
+module v0::a {
+}
+
+//# upgrade --package v0 --upgrade-capability 1,1 --sender A
+module v1::a {
+    fun init(_: &mut TxContext) {
```
tnowacki, line 15: Do we have a test (maybe pre-existing) for removing an init? 

Maybe test add-remove-add too?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -93,32 +101,146 @@ fn analyze_command<E: ExecutionErrorTrait>(
...
             }
         }
+        Command::Upgrade(payload, _, current_package_id, _, resolved_linkage) => {
+            if !protocol_config.enable_init_on_upgrade() {
+                return Ok(());
+            }
+
+            let current_pkg = get_package(current_package_id, store)?;
+
+            debug_assert!(
```
tnowacki: invariant violation seems fine here too

crates/sui-adapter-transactional-tests/tests/init/upgrade_existing_module_add_init_error.move
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//# init --addresses v0=0x0 v1=0x0 --accounts A
+
+//# publish --upgradeable --sender A
+module v0::a {
+}
+
+//# upgrade --package v0 --upgrade-capability 1,1 --sender A
+module v1::a {
```
tnowacki, line 14: Looking at the implementation we should be sure to have a version of this test where there is also an upgrade with init. And we probably want to make sure we have various lexicographical orderings between the old and new module (both with init). In short what I'm after is making sure we don't short circuit on the new module having init without first checking that all the old modules did not add init

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -93,32 +101,146 @@ fn analyze_command<E: ExecutionErrorTrait>(
...
+                module_has_init(module.compiled_module()),
+            )
+        })
+        .collect::<BTreeMap<_, _>>();
+
+    let mut has_new_module_init = false;
+    for module in modules {
+        let module_name = module.identifier_at(module.self_handle().name).as_str();
+        let has_init = module_has_init(module);
+        match current_module_inits.get(module_name) {
```
tnowacki: I wonder if we want to split this init two separate passes for hardening. Though this is fine, just a thought.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -93,32 +101,146 @@ fn analyze_command<E: ExecutionErrorTrait>(
...
+}
+
+fn module_has_init(module: &CompiledModule) -> bool {
+    module.function_defs().iter().any(|func_def| {
+        let handle = module.function_handle_at(func_def.function);
+        module.identifier_at(handle.name) == INIT_FN_NAME
+    })
+}
+
+/// Add the linkage constraints needed to run `init` functions introduced by an upgrade.
```
tnowacki: I like succinct comments, but might be a tad too succinct. We should shortly describe what those rules are

crates/sui-protocol-config/src/lib.rs
```diff
@@ -357,6 +357,7 @@ const MAINNET_USDB: &str =
 // Version 127: Enable always_advance_dkg_to_resolution.
 //              Update gas prices for range proofs and ristretto group operations.
 //              Enable Ristretto255 group operations and bulletproofs verification on testnet.
+//              Enable init functions for newly introduced modules during package upgrade.
```
tnowacki, line 360: Why is this in 127?

tzakian: just was the current max version when I started working on this (and forgot to update before putting it up). Will push this into the correct version once it comes time to land.

crates/sui-adapter-transactional-tests/tests/upgrade/upgrade_init_lumpy_examples.move
tzakian: I think we're `Lumpy` now. Will fix..

tnowacki (approved):

PR #27123, opened 2026-07-01
Prioritize system messages over user transactions in consensus adapter

alex-mysten:
## Description

Under high load, user transactions saturate the consensus adapter's submit semaphore, so system messages (checkpoint signatures, `EndOfPublish`, capability notifications, randomness DKG) block behind them.

System messages now bypass the submit semaphore (both the blocking `submit_and_wait_inner` path and the best-effort path used for execution-time observations); user transactions still contend for a permit. `submit_best_effort` now rejects user transactions, which are never expected on that path.

Part of CORE-192.

## Test plan

New unit test `system_message_bypasses_exhausted_submit_semaphore`: with a 0-permit semaphore, a user transaction blocks while an `EndOfPublish` completes.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/consensus_adapter.rs
+++ b/crates/sui-core/src/consensus_adapter.rs
@@ -181,6 +181,14 @@ pub trait SubmitToConsensus: Sync + Send + 'static {
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) -> SuiResult;
 
+    /// Submits a system transaction to consensus once, without waiting for it to
+    /// be sequenced and without retrying if it is garbage collected, bounded by
+    /// `timeout`. Suits periodic, self-superseding messages (e.g. execution time
+    /// observations) where a missed submission is replaced by the next one.
+    ///
+    /// For system transactions only. User transactions are rejected:
+    /// this fire-and-forget, no-retry, backpressure-free path would
+    /// silently mishandle them.
     fn submit_best_effort(
         &self,
         transaction: &ConsensusTransaction,
@@ -469,6 +477,7 @@ impl ConsensusAdapter {
         // - If is_soft_bundle, then all transactions are of CertifiedTransaction or UserTransaction kind.
         // - If not is_soft_bundle, then transactions must contain exactly 1 tx, and transactions[0] can be of any kind.
         let is_soft_bundle = transactions.len() > 1;
+        let is_system_message = !transactions[0].is_user_transaction();
 
         let mut transaction_keys = Vec::new();
         let mut tx_consensus_positions = tx_consensus_positions;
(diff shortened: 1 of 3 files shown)


aschran (approved):

crates/sui-core/src/unit_tests/consensus_tests.rs
```diff
@@ -265,6 +267,64 @@ async fn submit_multiple_transactions_to_consensus_adapter() {
     waiter.await.unwrap();
 }
 
+#[tokio::test]
```
mystenmark: might be worth making this a simtest so that the timeout doesn't actually take 2 seconds of wall time

mystenmark (approved):

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -554,12 +558,20 @@ impl ConsensusAdapter {
...
-                .submit_semaphore
-                .acquire()
-                .count_in_flight(self.metrics.sequencing_in_flight_semaphore_wait.clone())
-                .await
-                .expect("Consensus adapter does not close semaphore");
+            // System messages (checkpoint signatures, EndOfPublish, capability
+            // notifications, randomness DKG, etc.) are not buffered behind user
+            // tx; they are excluded from the semaphore.
+            let _permit: Option<SemaphorePermit> = if is_system_message {
+                None
```
akichidis, line 570: Let's keep in mind that although we do bypass this here, the consensus maintains a separate queue as well so under stress this shortcut will get throttled downstream.
I think once we refine a bit further how we consume transactions between `consensus_adapter <-> consensus` we can even have some special reserved capacity in the blocks for system transactions only.

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -1074,13 +1086,15 @@ impl SubmitToConsensus for Arc<ConsensusAdapter> {
         // timeout is required, or the spawned task can run forever
         timeout: Duration,
     ) -> SuiResult {
-        let permit = match self.submit_semaphore.clone().try_acquire_owned() {
-            Ok(permit) => permit,
-            Err(_) => {
-                return Err(SuiErrorKind::TooManyTransactionsPendingConsensus.into());
+        if transaction.is_user_transaction() {
```
akichidis, line 1094: Can we add a method comment that is supposed to be used only for non-user transactions and why is that important?

akichidis (approved):

crates/sui-core/src/consensus_adapter.rs
alex-mysten: This makes a lot of sense. I did know that the system messages do share a queue downstream; just removing one bottleneck here. Your idea of reserving capacity for system messages makes sense.

PR #27124, opened 2026-07-01
Add discovery address prober for trusted-peer reachability metrics

aschran:
## Description

Validators have no signal today when a peer advertises an unreachable address (e.g. a wrong external address gossiped via discovery), which silently degrades connectivity. This adds a prober that runs on validators and periodically checks whether each trusted peer's advertised P2P and consensus addresses are actually reachable, exposing the results as Prometheus metrics plus an admin endpoint that dumps full per-address detail for operators to act on.

Bumps `anemo` to mystenlabs/anemo#75 for the non-disruptive `Network::probe_address` the P2P check uses.

## Test plan

New `discovery_tests` sim tests cover the core differential (a validator's real address probes reachable, its bad gossiped address does not), plus no-false-positives and reconfiguration cleanup cases.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-node/src/lib.rs
+++ b/crates/sui-node/src/lib.rs
@@ -160,6 +160,7 @@ use typed_store::rocks::default_db_options;
 
 use crate::metrics::{GrpcMetrics, SuiNodeMetrics};
 
+pub mod address_prober;
 pub mod admin;
 pub mod db_shell;
 mod handle;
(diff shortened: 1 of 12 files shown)


crates/sui-node/src/lib.rs
```diff
@@ -966,6 +971,34 @@ impl SuiNode {
             None
         };
 
+        if let Some(prober_config) = &config.address_prober {
+            prober_config.validate()?;
```
mystenmark: should validation failure exit this entire function?

mystenmark (approved):

crates/sui-node/src/lib.rs
aschran: yes it was intentional to crash the process if it's started with an invalid config, but will move this somewhere more sensible

PR #27125, opened 2026-07-01
feat(transaction-driver): wait for effects on durably-processing transactions

arun-koshy:
## Description

Follow-up to #27083. When a validator rejects a submission with `TransactionProcessing`, the transaction driver treated it as a retriable error and resubmitted — which the validator just suppresses again, causing retry churn. When the transaction is durably in flight (sequenced by consensus, or executed via a checkpoint) its outcome is already final, so the driver now waits for that outcome by digest instead of resubmitting. The non-durable "recently submitted" dedup status (#27069) is not evidence the transaction reached consensus, so it stays retriable.

The wait path is only taken when the rejection's digest matches the request, and a processing rejection for the driver's own transaction no longer counts against the validator's client-monitor score (the dedup hit is caused by the driver's own resubmission, not a validator fault).

## Test plan

New unit tests cover: a durable rejection is surfaced as a result after a single submission (no churn) and routes the certifier into the wait-for-effects path; "recently submitted" and wrong-digest rejections stay retriable and resubmit to all validators; validator score feedback is recorded for malformed (wrong-digest) responses but not for own-transaction dedup hits; and the durable/non-durable classification helper in `sui-types`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/transaction_driver/effects_certifier.rs
+++ b/crates/sui-core/src/transaction_driver/effects_certifier.rs
@@ -104,12 +104,22 @@ impl EffectsCertifier {
                 None => (None, None),
             },
             SubmitTxResult::Rejected { error } => {
-                return Err(TransactionDriverError::ClientInternal {
-                    error: format!(
-                        "Unexpected submission error in get_certified_finalized_effects(): {}",
-                        error
-                    ),
-                });
+                if matches!(
+                    error.as_inner(),
+                    SuiErrorKind::TransactionProcessing { digest, .. } if Some(*digest) == tx_digest
+                ) {
+                    // Already being processed by consensus: no consensus position to track, so
+                    // wait for the finalized outcome by digest — same path as an `Executed`
+                    // response with no inline effects.
+                    (None, None)
+                } else {
+                    return Err(TransactionDriverError::ClientInternal {
+                        error: format!(
+                            "Unexpected submission error in get_certified_finalized_effects(): {}",
+                            error
+                        ),
+                    });
+                }
             }
         };
 
--- a/crates/sui-core/src/transaction_driver/mod.rs
+++ b/crates/sui-core/src/transaction_driver/mod.rs
@@ -333,7 +333,16 @@ where
                 options,
             )
             .await?;
-        if let SubmitTxResult::Rejected { error } = &submit_txn_result {
+        // A `TransactionProcessing` rejection for this transaction is expected: the submitter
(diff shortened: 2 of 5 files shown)


crates/sui-core/src/transaction_driver/effects_certifier.rs
```diff
@@ -103,6 +103,16 @@ impl EffectsCertifier {
                 // But if it is not set, continuing to get full effects and certify the digest are still correct.
                 None => (None, None),
             },
+            // Durably processing (sequenced or checkpointed): no consensus position to track, so
+            // wait for the finalized outcome by digest — same path as an `Executed` response with
+            // no inline effects.
+            SubmitTxResult::Rejected { error }
```
mwtian: Merge with the case below?

crates/sui-types/src/error.rs
```diff
@@ -1001,6 +1007,26 @@ impl SuiError {
...
+    /// Returns the digest of a `TransactionProcessing` rejection whose outcome is already finalized
+    /// by consensus (sequenced, or executed via a checkpoint). Waiting for that outcome by digest is
+    /// guaranteed to resolve — to effects or a certified rejection — so the driver waits instead of
+    /// resubmitting.
+    ///
+    /// Returns `None` for the sole non-durable status ([`TX_PROCESSING_RECENTLY_SUBMITTED`], a
+    /// per-validator dedup hit) and for any other error, so the driver keeps resubmitting. Treating
+    /// every other processing status as durable is deliberate: new durable states are handled
+    /// automatically, with only the one non-durable status pinned by a constant.
+    pub fn durable_transaction_processing_digest(&self) -> Option<TransactionDigest> {
```
mwtian: Since this function is pretty specific to transaction driver, I think it should be moved there.

crates/sui-core/src/authority_server.rs
```diff
@@ -979,7 +979,7 @@ impl ValidatorService {
                 results[idx] = Some(SubmitTxResult::Rejected {
```
mwtian: Continuing the discussion, maybe we can add a SubmitTxResult::Processed variant, or turn SubmitTxResult::Submitted::consensus_position to be Option, ser/de to empty Bytes value?

PR #27127, opened 2026-07-02
[Stress] allow create duplicate and amplified traffic

akichidis:
## Description 

Stress nodes, given a desired configuration, can now send:
* same duplicate transaction(s) to the same node
* amplified transaction copies across multiple nodes
* the combination of above two

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/src/drivers/bench_driver.rs
+++ b/crates/sui-benchmark/src/drivers/bench_driver.rs
@@ -55,7 +55,7 @@ use tokio::{time, time::Instant};
 use tracing::{debug, error, info, warn};
 
 use super::Interval;
-use super::{BenchmarkStats, StressStats};
+use super::{BenchmarkStats, StressStats, SubmissionAmplification};
 
 /// Randomly partitions a list of transactions into groups for soft bundle submission.
 /// Each group will be submitted as a separate soft bundle.
@@ -257,6 +257,7 @@ pub struct BenchWorker {
     pub execution_proxy: Arc<dyn ValidatorProxy + Send + Sync>,
     pub group: u32,
     pub duration: Interval,
+    pub submission_amplification: SubmissionAmplification,
 }
 
 impl Debug for BenchWorker {
@@ -349,6 +350,7 @@ impl BenchDriver {
                     execution_proxy: execution_proxy.clone(),
                     group: workload_info.workload_params.group,
                     duration: workload_info.workload_params.duration,
+                    submission_amplification: workload_info.submission_amplification,
                 });
                 payloads = remaining;
                 qps -= target_qps;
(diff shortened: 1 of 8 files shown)


crates/sui-benchmark/src/drivers/mod.rs
```diff
@@ -5,11 +5,128 @@ use duration_str::parse;
...
+            amplification_validators_per_tx,
+            duplicate_probability,
+            duplicate_copies_per_validator,
+            validator_selection,
+        })
+    }
+
+    pub fn is_enabled(&self) -> bool {
+        (self.amplification_probability > 0.0 && self.amplification_validators_per_tx > 1)
+            || (self.duplicate_probability > 0.0 && self.duplicate_copies_per_validator > 1)
```
alex-mysten, line 118: There is some dependency that perhaps we should validate in `new` - if we have `amplification_probability > 0.0`, we must also have `amplification_validators > 1`

crates/sui-benchmark/src/drivers/mod.rs
```diff
@@ -5,11 +5,128 @@ use duration_str::parse;
...
+            validators_per_tx,
+            copies_per_validator,
+            validator_selection: self.validator_selection,
+        }
+    }
+
+    pub fn expected_submission_multiplier(&self) -> f64 {
+        (1.0 + self.amplification_probability * (self.amplification_validators_per_tx - 1) as f64)
+            * (1.0 + self.duplicate_probability * (self.duplicate_copies_per_validator - 1) as f64)
+    }
```
alex-mysten, line 148: Claude says this undercounts

crates/sui-benchmark/src/drivers/bench_driver.rs
```diff
@@ -1117,18 +1119,16 @@ async fn run_bench_worker(
                         // TODO: clone committee for each request is not ideal.
                         let committee = worker.execution_proxy.clone_committee();
 
-                        // Occasionally submit to multiple validators to test unpaid amplification deferral.
-                        // With 5% probability, submit to 3 to N validators to trigger deferral logic.
-                        let use_amplification = rand::thread_rng().gen_bool(0.05);
-                        let committee_size = committee.num_members();
+                        let submission_amplification = worker.submission_amplification;
```
alex-mysten, line 1122: Are we unintentionally removing coverage by not including some amplification to existing workloads out of the box?

alex-mysten (commented):
Were we able to repro the latency spikes in PT with these changes?

crates/sui-benchmark/src/drivers/bench_driver.rs
akichidis: Good catch - I've updated that now

mwtian (approved):

PR #27128, opened 2026-07-02
[5/7] Support implicitly-read system object versions in dry-run/simulate and replay

lxfind:
## Description

Stack of 7 (builds on [4/7] #27079). Completes the implicitly-read system object machinery for the execution paths that are not sequenced by consensus and so have no assigned system-object versions:

- **Dev-inspect / dry-run**: the executor pins every member of `IMPLICITLY_READ_SYSTEM_OBJECTS` (today just the accumulator root) at its latest committed version read directly from the object store and threads them in as system-object versions, so implicit reads during simulation resolve deterministically. A latest (non-fork-safe) read is acceptable because this is a simulation; the only edge case — that version being pruned mid-execution — surfaces as a recoverable error rather than a panic. The pinning read is captured by the simulate path's tracking store, so these objects are filtered out of the effects' `unchangedLoadedRuntimeObjects` (real execution records them as unchanged consensus objects instead, never there).
- **sui-replay-2**: derives the versions from the expected effects the same way checkpoint execution does — every accessed consensus object at its recorded version, skipping cancelled-transaction sentinels.

## Test plan

The maps built here have no consumer until the in-VM funds check lands in [7/7]; the e2e simulate tests and the transactional `.move` test there exercise both the dev-inspect pinning and the `unchangedLoadedRuntimeObjects` filtering. Existing dev-inspect and replay tests cover these paths passing the (unread) versions through unchanged.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -1864,7 +1864,7 @@ impl AuthorityState {
         self.metrics.total_effects.inc();
         self.metrics.total_certs.inc();
 
-        let consensus_object_count = effects.input_consensus_objects().len();
+        let consensus_object_count = effects.accessed_consensus_objects().len();
         if consensus_object_count > 0 {
             self.metrics.shared_obj_tx.inc();
         }
@@ -6719,7 +6719,7 @@ impl NodeStateDump {
 
         // Record all the shared objects
         let mut shared_objects = Vec::new();
-        for kind in effects.input_consensus_objects() {
+        for kind in effects.accessed_consensus_objects() {
             match kind {
                 InputConsensusObject::Mutate(obj_ref) | InputConsensusObject::ReadOnly(obj_ref) => {
                     if let Some(w) = object_store.get_object_by_key(&obj_ref.0, obj_ref.1) {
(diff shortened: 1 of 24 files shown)


crates/sui-core/src/authority.rs
```diff
@@ -2693,12 +2695,18 @@ impl AuthorityState {
...
+        let mut unchanged_loaded_runtime_objects =
             crate::transaction_outputs::unchanged_loaded_runtime_objects(
                 &transaction,
                 &effects,
                 &loaded_runtime_objects,
             );
+        // Dev-inspect reads implicitly-read system objects only to pin their versions as
+        // system-object versions; the tracking store captures those metadata reads, but they must
+        // not surface as loaded runtime objects. Real execution never records these objects here —
+        // when a transaction actually reads one, it is emitted as an unchanged consensus object
```
tzakian, line 2707: Just to double-check: there's nothing the transaction could do where these objects could show up in the `unchanged_loaded_runtime_objects` for any other reason right?

tzakian (approved):
Changes look good to me just one small double-check (pretty sure the answer is no though).

crates/sui-core/src/authority.rs
lxfind: Correct

lxfind:
Superseded: folded into #27403 (block-wait implicitly-read system objects).

PR #27129, opened 2026-07-02
indexer-alt: group pipelines into ingestion cohorts by distance from tip

evan-wall-mysten:
## Description

An indexer's pipelines all share one ingestion service that starts at the minimum watermark across them, so its slowest subscriber gates fetch concurrency for everyone — one far-behind pipeline (e.g. a backfill) holds back pipelines already at the tip. This groups pipelines into cohorts by how far behind the tip they resume and gives each cohort its own ingestion service, so a backfill only backpressures similarly-lagging peers. The earlier commits in the stack are supporting refactors: a shareable (`Arc`) streaming client, an `IngestionFactory` that mints services on demand, and per-cohort metric labels.

## Test plan

New unit tests for the cohort-grouping boundaries, plus end-to-end tests that a near-tip pipeline receives zero checkpoints while a far-behind cohort backfills to completion (through both the args- and client-driven constructors).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [x] Indexing Framework: Pipelines are now grouped into cohorts by distance from the tip, each with its own ingestion service, so a lagging pipeline no longer throttles caught-up ones. New optional `IngestionConfig.min_cohort_boundary` (default 25,000 checkpoints) tunes the grouping; no action required.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/cohort.rs
+++ b/crates/sui-indexer-alt-framework/src/cohort.rs
@@ -1,0 +1,187 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Grouping of pipelines into ingestion cohorts based on how far behind the network tip each
+//! pipeline's resume checkpoint is. Pipelines in the same cohort share one ingestion service, so
+//! a far-behind pipeline only applies backpressure to similarly-lagging peers, not to pipelines
+//! that are already at the tip.
+
+use crate::PendingPipeline;
+
+/// Default for [`IngestionConfig::min_cohort_boundary`]: a cohort's boundary (its maximum member
+/// distance from the tip) is at least this many checkpoints, so that pipelines that are roughly
+/// caught up are not fragmented into many tiny cohorts, each with its own ingestion service.
+///
+/// [`IngestionConfig::min_cohort_boundary`]: crate::ingestion::IngestionConfig::min_cohort_boundary
+pub(crate) const DEFAULT_MIN_COHORT_BOUNDARY: u64 = 25_000;
+
+/// Group pipelines into cohorts by their distance from the network tip.
+///
+/// Each pipeline's distance is how far behind `latest_checkpoint` it will resume from. `pipelines`
+/// may be in any order; they are sorted (nearest-tip first) internally. The closest unassigned
+/// pipeline (distance `d`) seeds a cohort that absorbs every pipeline whose distance is at most
+/// `max(2 * d, min_cohort_boundary)`; the next unassigned pipeline seeds the next cohort, and so
+/// on.
+///
+/// Returns the cohorts ordered nearest-tip first, each holding its members in ascending distance
+/// order (pipelines at equal distances keep registration order).
+pub(crate) fn cohorts(
+    mut pipelines: Vec<PendingPipeline>,
+    latest_checkpoint: u64,
+    min_cohort_boundary: u64,
+) -> Vec<Vec<PendingPipeline>> {
+    // Stable sort: pipelines at equal distances keep registration order.
+    pipelines.sort_by_key(|p| latest_checkpoint.saturating_sub(p.next_checkpoint));
+
+    // The closest unassigned pipeline seeds a cohort that absorbs every peer within
+    // `max(2 * its distance, min_cohort_boundary)`; the first pipeline beyond that boundary seeds
+    // the next cohort, and so on.
+    let mut cohorts: Vec<Vec<PendingPipeline>> = vec![];
+    let mut boundary = None;
+    for pipeline in pipelines {
+        let dist = latest_checkpoint.saturating_sub(pipeline.next_checkpoint);
+        if boundary.is_none_or(|b| dist > b) {
+            cohorts.push(vec![]);
+            boundary = Some(dist.saturating_mul(2).max(min_cohort_boundary));
+        }
+        cohorts.last_mut().unwrap().push(pipeline);
+    }
+    cohorts
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// A pipeline resuming exactly `distance` checkpoints behind the tip. Tests fix the tip at
+    /// `u64::MAX` so scenarios can be written directly in distances and the overflow case is
+    /// reachable.
+    fn at(distance: u64) -> PendingPipeline {
+        PendingPipeline {
+            name: "test",
+            next_checkpoint: u64::MAX - distance,
+            build: Box::new(|_| unreachable!("cohort grouping never builds pipelines")),
+        }
+    }
+
+    /// Group `pipelines` (tip at `u64::MAX`) and read each cohort back out as member distances.
+    fn grouped(pipelines: Vec<PendingPipeline>, min_cohort_boundary: u64) -> Vec<Vec<u64>> {
+        cohorts(pipelines, u64::MAX, min_cohort_boundary)
+            .iter()
+            .map(|cohort| {
+                cohort
+                    .iter()
+                    .map(|p| u64::MAX - p.next_checkpoint)
+                    .collect()
+            })
+            .collect()
+    }
+
+    #[test]
+    fn test_empty() {
+        assert_eq!(
+            grouped(vec![], DEFAULT_MIN_COHORT_BOUNDARY),
+            Vec::<Vec<u64>>::new()
+        );
+    }
+
+    #[test]
+    fn test_singleton() {
+        assert_eq!(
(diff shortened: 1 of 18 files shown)


crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
nickvikeras: Now we can have multiple ingestion services step on each other and set these metrics to their own value on the same `IngestionMetrics` struct so these metrics are broken when we have multiple cohorts. 

I think we either need to split these out into "per-cohort" metrics with their own labels, or, just delete them for now. They were more critical when I was debugging backfill throughput. They might not be pulling their weight anymore. I wouldn't block the rest of the PR on saving these metrics, but I also wouldn't leave in if they are broken.

crates/sui-indexer-alt-framework/src/ingestion/mod.rs
```diff
@@ -79,16 +82,41 @@ pub struct IngestionConfig {
...
 }
 
+/// Creates the [IngestionService]s that an indexer runs -- one per cohort of pipelines with
+/// similar distances from the network tip. The factory holds the clients services are made from;
+/// services are only constructed when requested through [Self::create], and every service is
+/// created the same way: it shares the factory's ingestion client (and so reports to the same
+/// metrics -- counters aggregate across services, and latest-checkpoint style gauges interleave
+/// while multiple services run), and it gets its own clone of the streaming client, if there is
+/// one.
+pub struct IngestionFactory {
```
nickvikeras: Can this be `pub (crate)`?

crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
evan-wall-mysten: I'll change them to a vec for now then.

nickvikeras (approved):

crates/sui-indexer-alt-framework/src/ingestion/mod.rs
evan-wall-mysten: This was originally for consistency with IngestionService, but it looks like that was a special case for usage outside of the Indexer (https://github.com/MystenLabs/sui/commit/be5267f315c6bcbf52d768a85839d505b7c8a5d7) so I'll change it.

crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
```diff
@@ -316,9 +332,7 @@ async fn stream_and_broadcast_range(
                 lo, "Skipping already processed checkpoint"
             );
             metrics.total_skipped_streamed_checkpoints.inc();
```
amnn: Why is this metric also no parameterized by the cohort?

crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
```diff
@@ -340,7 +354,7 @@ async fn stream_and_broadcast_range(
 
         debug!(checkpoint = lo, "Streamed checkpoint");
         metrics.total_streamed_checkpoints.inc();
```
amnn: ditto.

crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
amnn, line 1: Mixing ingestion-wide and cohort-wide metrics will lead to confusion -- in most cases let's lean towards making a metric cohort-wide. We can always sum them together when querying metrics if we want the true global metric.

amnn: A good way of structuring this is to create a `CohortMetrics` struct that is constructed from `IngestionMetrics` by parametrizing using label values. That way everywhere inside `IngestionService` we have the appropriately labeled metric series to work with.

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
```diff
@@ -317,6 +313,16 @@ impl IngestionClient {
...
 
+    /// A clone of this client that reports checkpoint-lag gauges under `cohort`, so services minted
+    /// for different cohorts don't overwrite each other. The checkpoint source and chain-id cache
+    /// are shared; only the metric reporter is rebound.
+    pub(crate) fn for_cohort(&self, cohort: usize) -> Self {
+        let mut client = self.clone();
+        client.checkpoint_lag_reporter =
+            CheckpointLagMetricReporter::new_for_cohort(&self.metrics, cohort);
+        client
+    }
```
amnn: Not a huge fan of setting this metrics lag reporter with a dummy `0` cohort with the expectation/hope that it gets overridden later. Will see if there's a way to improve that.

crates/sui-indexer-alt-framework/src/ingestion/mod.rs
```diff
@@ -573,29 +636,41 @@ mod tests {
...
-        let mut streaming = Some(
+        let streaming = Some(
             MockStreamingClient::new(std::iter::empty::<u64>(), None).fail_connection_times(1),
         );
-        let result = latest_checkpoint_number(streaming.as_mut(), &client).await;
+        let result = latest_checkpoint_number(streaming.as_ref(), &client).await;
         assert_eq!(result.unwrap(), FALLBACK);
     }
 
+    /// An `IngestionConfig` deserialized without `min_cohort_boundary` falls back to the default.
```
amnn: This is kind of a weird test -- it's mostly testing `serde`'s behaviour, and it's testing it with `serde_json` which is not even the format we use to serialize/deserialize configs.

Would suggest just dropping this test entirely.

crates/sui-indexer-alt-framework/src/cohort.rs
```diff
@@ -0,0 +1,193 @@
...
+        lo += count;
+    }
+
+    // Split the sorted vec into owned cohorts. `into_iter().take(count)` moves elements out
+    // front-to-back without shifting the buffer, so this is O(n) overall.
+    let mut pipelines = pipelines.into_iter();
+    sizes
+        .into_iter()
+        .map(|count| pipelines.by_ref().take(count).collect())
+        .collect()
```
amnn:
```rust
let mut cohorts = vec![];
let mut threshold = None;
for pipeline in pipelines {
    let dist = latest_checkpoint.saturating_sub(pipeline.next_checkpoint);
    if threshold.is_none_or(|t| dist > t) {
        cohorts.push(vec![]);
        threshold = Some(dist.saturating_mul(2).min(min_cohort_boundary));
    }

    cohorts.last().unwrap().push(pipeline)
}

cohorts
```

Single pass way to achieve the same, I think?

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
amnn: Okay, I think maybe an acceptable middle-ground would be to expose a `CheckpointLagMetricReporter::with_label(metrics: &IngestionMetrics, label: &str)`, which we can use here with an explicit "<default>" label or something when we construct the client for the first time, so that we don't conflate requests that happen before the indexer has even started with the ones that belong to the first cohort.

amnn (commented):
Looks good to me, thanks for the quick turnaround on this @evan-wall-mysten. Nothing blocking but the main themes in my comments were around:

- Metrics: Parametrization of all ingestion metrics into `CohortMetrics` (or perhaps `ClientMetrics`) that are labeled versions of the global `IngestionMetrics`), and a clear story of how a client or cohort's metrics are derived from the global metrics. (Current approach parametrizes some but not all metrics, and the hand-off from global metrics to per-cohort is ad-hoc and crosses abstraction boundaries).
- Simpler core algorithm for grouping into cohorts.

amnn (approved):
(Forgot to accept as well)

crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
evan-wall-mysten: Originally I only changed the metrics where a value was being set and not incremented because those metrics would have been broken by this change, but I will parameterize all by cohort for consistency.

crates/sui-indexer-alt-framework/src/ingestion/ingestion_client.rs
evan-wall-mysten: Done. I also migrated `new_for_pipeline` to use this since it was doing the same thing. I can revert this later if you think they should be separated and add a comment as to why.

amnn: All good -- that's what I had in mind as well!

PR #27131, opened 2026-07-02
[graphql] Migrate `Checkpoint` and `Event` cursor to grpc format

wlmyng:
## Description 

Since multiple endpoints/ types share the same `CursorToken`, each module of `Checkpoint`, `Event`, `Transaction` adds a newtype wrapper over `CursorToken`, then use `MultiCursor` for backwards compatibility.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
@@ -1,6 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::ops::Deref;
 use std::sync::Arc;
 
 use anyhow::Context as _;
@@ -8,6 +9,8 @@ use async_graphql::Context;
 use async_graphql::Object;
 use async_graphql::connection::Connection;
 use sui_indexer_alt_reader::kv_loader::KvLoader;
+use sui_rpc_cursor::CursorToken;
+use sui_rpc_cursor::Position;
 use sui_types::crypto::AuthorityStrongQuorumSignInfo;
 use sui_types::digests::CheckpointDigest;
 use sui_types::message_envelope::Message;
@@ -17,7 +20,10 @@ use sui_types::messages_checkpoint::CheckpointSummary;
 
 use crate::api::query::Query;
 use crate::api::scalars::base64::Base64;
+use crate::api::scalars::cursor::ByteCursor;
 use crate::api::scalars::cursor::JsonCursor;
+use crate::api::scalars::cursor::MultiCursor;
+use crate::api::scalars::cursor::OpaqueCursor;
 use crate::api::scalars::date_time::DateTime;
 use crate::api::scalars::id::Id;
 use crate::api::scalars::uint53::UInt53;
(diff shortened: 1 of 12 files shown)


crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
```diff
@@ -383,3 +393,95 @@ impl CheckpointContents {
...
+        ))))
+    }
+}
+
+impl CCheckpoint {
+    pub(crate) fn sequence_number(&self) -> u64 {
+        match self {
+            CCheckpoint::Primary(c) => match c.position {
+                Position::Checkpoints { checkpoint } => checkpoint,
+                _ => unreachable!("validated at decode"),
```
amnn: The fact that this is a thing does make me want to consider not re-using `CursorToken` everywhere but rather introducing types to go into `OpaqueToken` that are guaranteed to be checkpoint, transaction, event tokens and then marshal between those types and `CursorToken` in the `ByteCursor` impls.

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -188,12 +198,129 @@ impl Event {
...
+    /// Mint the edge cursor for the event at the given coordinates.
+    pub(crate) fn cursor(checkpoint: u64, tx_seq: u64, ev_sequence_number: u64) -> CEvent {
+        MultiCursor::new(OpaqueCursor::new(Self(CursorToken::item(
+            Position::Events {
+                checkpoint,
+                tx_seq,
+                // Event counts per transaction are protocol-bounded, far below u32::MAX.
+                event_index: ev_sequence_number
+                    .try_into()
+                    .expect("event index fits in u32"),
```
amnn: Does the user have any control over this value at any point? Could they create a situation where this is no longer true (e.g. when providing a cursor or some other query param?)

amnn: Should we/could we use `u32` everywhere (i.e. have `EventToken::cursor` accept a `u32`) to avoid this `expect`?

amnn (approved):
Looks reasonable overall, thanks  @wlmyng  -- now you have these wrapper types though, you might consider making them not wrapper types but true types in their own right that are convertible to/from `CursorToken`, to avoid some of the type assertions.

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
wlmyng: Looking at the life of a cursor on event, it's not used in the first phase when we look up the event indexes, the query is keyed by `tx_sequence_number`. Then 
when filtering the events of a transaction in memory, it is clamped by `.min(event_count)`. At least from `CursorToken`, it is a u32 on the proto, so only place to get u64 is the legacy JSON arm. Moved it out of `fn cursor` and removed the `.expect`, so callsites do `as u32` instead

amnn: Need to double check but `as u32` is not usually a safe thing to do to something that's coming from user input (e.g. a cursor) -- it would panic.

It would be okay to update the JSON Cursor to accept a `u32` directly, because then the cursor parsing logic would do the bounds check for you and return a non-panicking error.

wlmyng: It looks like Rust would truncate the upper bits of the number to fit into 32 bits, but I can also update the JsonCursor to accept u32 instead of u64

PR #27133, opened 2026-07-02
[refactor] update the processed message table for the dropped transactions

akichidis:
## Description 

Dropped transactions do not update  neither the processed messages table nor they do notify anyone who waiting to hear on updates via the `epoch_store.consensus_messages_processed_notify` . This can affect `consensus_adapter` with (dropped) transactions waiting there without having any way to get notified.

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/consensus_handler.rs
+++ b/crates/sui-core/src/consensus_handler.rs
@@ -98,6 +98,7 @@ use crate::{
 struct FilteredConsensusOutput {
     transactions: Vec<(SequencedConsensusTransactionKind, u32)>,
     owned_object_locks: HashMap<ObjectRef, TransactionDigest>,
+    dropped_transaction_keys: Vec<ConsensusTransactionKey>,
 }
 
 pub struct ConsensusHandlerInitializer {
@@ -1164,6 +1165,7 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
         let FilteredConsensusOutput {
             transactions,
             owned_object_locks,
+            dropped_transaction_keys,
         } = self.filter_consensus_txns(
             state.initial_reconfig_state.clone(),
             &commit_info,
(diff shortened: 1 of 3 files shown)


crates/sui-core/src/consensus_handler.rs
```diff
@@ -1173,6 +1175,13 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
         if !owned_object_locks.is_empty() {
             state.output.set_owned_object_locks(owned_object_locks);
         }
+
+        // Still record the dropped transactions as consensus message processed.
```
arun-koshy, line 1179: For a follow up PR or you can add here. This will fix the clogging issue but it will return back to the client the error TransactionProcessing in the positionless path which is a retriable error instead of ObjectLockConflict which is the right error and should be non retriable.  So I think there isnt much choice if we want to handle this right other than waiting on consensus position to get the proper error or record dropped digests digest-keyed (alongside the position-keyed reason)

crates/sui-core/src/consensus_handler.rs
```diff
@@ -3684,6 +3701,123 @@ mod tests {
...
+        assert_eq!(locks.get(&owned_object_ref), Some(&winner_digest));
+        assert!(
+            epoch_store
+                .is_consensus_message_processed(&winner_key)
+                .unwrap()
+        );
+        assert!(
+            epoch_store
+                .is_consensus_message_processed(&loser_key)
+                .unwrap()
```
arun-koshy, line 3817: nit: How about also awaiting epoch_store.consensus_messages_processed_notify(vec![loser_key]) here ?

arun-koshy (approved):

mwtian (approved):

PR #27134, opened 2026-07-02
[sui][sui-adapter] Strip TxContext borrows from PTB call sources in reference safety

cgswords:
## Description

Under `allow_references_in_ptbs`, every `TxContext` borrow shared a single root, so a result derived from a ctx-taking call conflicted with every later `TxContext` injection, making such results unusable across commands.

Since no function can return a reference into `TxContext`, both memory-safety checkers now strip the injected `TxContext` borrows from what call results extend, after the usual exclusivity checks run over the full argument set (so multiple `&mut TxContext`, or `&mut` + `&`, within one call are rejected exactly as without the flag). The regex checker filters them from the sources for `extend_by_dot_star_for_call`; the path-set invariant checker does the same strip and additionally enforces the post-condition that no result path is ever rooted at `TxContext`. Flag-off behavior is unchanged, including dev-inspect, which permits reference results without the flag.

## Test plan

New transactional tests cover coexistence across commands (mut/imm results held across later `TxContext` uses, two mutable results side by side), within-call exclusivity (`tx_context_multiple_mut_borrows_invalid` mirrors the flag-off `tx_context_multiple_borrows_invalid`), and flag-off dev-inspect twins pinning the pre-flag conflict behavior. A new e2e simtest covers the synthesized withdrawal-conversion command, the only synthesized-command path, which is unreachable from the adapter transactional tests. @tzakian will backtest before landing.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


tnowacki (commented):
I think we will need some more test coverage here to do some basic things to just demonstrate that the TxContext is not considered.

For example

```
public fun borrow_mut(my_obj: &mut X, ctx: &mut TxContext): &mut Y { ... }
public fun write(&mut Y, ctx: &mut TxContext) { ... }
```

A decent test here would be a PTB that calls `borrow_mut` does some other things that mutably uses the `TxContext` and then calls `write` using the `Result` from `borrow_mut`

cgswords:
Added some tests

tnowacki (changes requested):
I did not catch this until reading the description of the index-based approach in the invariant checks... sorry about that. 
But I don't think we can use the approach taken here `memory_safety`. Additionally as written we have a fork risk of existing behavior. 

That is that a function like 
```
public fun foo(_: &mut TxContext, _: &mut TxContext) { ... }
```
is not callable today, but with this change it would be.

I would propose the following, but this is just my first thought so definitely open to alternative ideas:
1. For the normal memory safety check, just remove the tx context from the sources after the call to `find_non_transferrable` 
2. For the invariant check, use the command index for the TxContext instead of a globally unique ID per usage

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/invariant_checks/memory_safety.rs
```diff
@@ -379,7 +419,12 @@ impl Context {
 
     fn location(&self, loc: T::Location) -> anyhow::Result<&Location> {
         Ok(match loc {
-            T::Location::TxContext => &self.tx_context,
+            T::Location::TxContext => {
+                let i = self.tx_context_index()?;
+                self.tx_contexts
+                    .get(i as usize)
+                    .ok_or_else(|| anyhow::anyhow!("TxContext index out of bounds {i}"))?
```
tnowacki: What about flipping this aroudn a bit, so we move the gating essentially into the `tx_contexts`: 
- If references are enabled, you supply the index `i` when creating the vector
- If references are not enabled, always supply 0. 

Then we can always use the current command index in this function

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/memory_safety.rs
```diff
@@ -22,11 +22,26 @@ use sui_types::{
...
+/// single shared root exactly as before the flag existed.
+///
+/// A `TxContext` root is minted only while its command is the current one, and positions
+/// strictly increase, so a live reference rooted in `TxContext` at some position can never have
+/// that root re-borrowed by a later command.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
-struct Location(T::Location);
+enum BorrowRoot {
+    TxContext(u16),
+    Loc(T::Location),
```
tnowacki: Not important but worth thinking about: any way to ensure this is never `TxContext`?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/memory_safety.rs
```diff
@@ -22,11 +22,26 @@ use sui_types::{
...
+/// exclusivity rules (at most one `&mut TxContext`, and no `&TxContext` alongside it). When
+/// references are not allowed, the position is always 0, giving every `TxContext` borrow a
+/// single shared root exactly as before the flag existed.
+///
+/// A `TxContext` root is minted only while its command is the current one, and positions
+/// strictly increase, so a live reference rooted in `TxContext` at some position can never have
+/// that root re-borrowed by a later command.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
-struct Location(T::Location);
+enum BorrowRoot {
```
tnowacki: I'm fine with keeping this as `Location`? 
Or maybe `RootLocation` like the invariant check?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/memory_safety.rs
```diff
@@ -22,11 +22,26 @@ use sui_types::{
...
+/// the command list) of the borrowing command when references are allowed in PTBs: borrows from
+/// different commands then root disjointly, allowing them to co-exist without interfering with
+/// each other. Borrows within a single command share a root and are subject to the usual
+/// exclusivity rules (at most one `&mut TxContext`, and no `&TxContext` alongside it). When
+/// references are not allowed, the position is always 0, giving every `TxContext` borrow a
+/// single shared root exactly as before the flag existed.
+///
+/// A `TxContext` root is minted only while its command is the current one, and positions
+/// strictly increase, so a live reference rooted in `TxContext` at some position can never have
+/// that root re-borrowed by a later command.
```
tnowacki: 🐶 
Can we make this comment a bit more succinct?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/memory_safety.rs
```diff
@@ -152,13 +169,38 @@ impl Context {
             .map_err(graph_meter_err::<E>)
     }
 
+    /// The root for a borrow of `l` by the command at `position` in the command list. Note the
+    /// position is different from `Command::idx`, because a single input command may be split
+    /// into multiple actual commands. See `BorrowRoot`.
```
tnowacki: I feel like it is not necessary to comment this everywhere but I'm fine having it if we want it

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/memory_safety.rs
```diff
@@ -152,13 +169,38 @@ impl Context {
...
+
     /// Used for checking if a location is borrowed
     /// Used for updating the borrowed marker in Copy, and for correctness of Move
     fn is_location_borrowed<E: ExecutionErrorTrait>(&self, l: T::Location) -> Result<bool, E> {
+        // `TxContext` cannot be moved or copied, so this check--used only by move and copy--never
+        // needs to consider the per-command `TxContext` labels
+        assert_invariant!(
+            !matches!(l, T::Location::TxContext),
+            "TxContext is never moved or copied"
+        );
```
tnowacki: Can this be hit in dev inspect?

tnowacki (commented):
Do we no longer want to do two separate approaches? One where we carve out from `sources` and the other where we use the command index? 
I'm fine if we don't, I just am wondering why the change of plan

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/invariant_checks/memory_safety.rs
```diff
@@ -359,11 +361,24 @@ impl Context {
...
+                    // produces ctx-rooted results.
+                    if allow_references_in_ptbs && let Some(Value::Ref { paths, .. }) = &v {
+                        anyhow::ensure!(
+                            paths
+                                .0
+                                .iter()
+                                .all(|p| p.root != RootLocation::Known(T::Location::TxContext)),
+                            "TxContext can never be returned"
+                        );
+                    }
```
tnowacki, line 381: I was fine with doing the index based shenanigans in the invariant checks, but this looks a bit less invasive, so seems good to me!

crates/sui-e2e-tests/tests/address_balance_compatibility_tests.rs
```diff
@@ -651,6 +651,48 @@ async fn test_split_from_fake_coin() {
     test_env.cluster.trigger_reconfiguration().await;
 }
 
+#[sim_test]
+async fn test_fake_coin_conversion_with_references_in_ptbs() {
```
tnowacki: Do we still need this now that we aren't doing index based things?

crates/sui-adapter-transactional-tests/tests/dev_inspect/tx_context_result_across_mut_tx_use.move
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Dev-inspect twin of programmable/tx_context_mut_result_across_mut_tx_use: under
+// --enable-feature-flags allow_references_in_ptbs, a `&mut Y` returned from `borrow_mut(&mut X, &mut TxContext)`
+// roots only in the object, so it stays valid across a later mutable TxContext use.
+// Compare with tx_context_result_across_mut_tx_use_invalid: dev-inspect permits reference
+// return values even without the flag, so the pair isolates exactly what stripping
+// TxContext borrows from call sources changes.
```
tnowacki: I feel like this comment could be significantly shorter. Makes my eyes spin a bit reading it towards the end

crates/sui-adapter-transactional-tests/tests/dev_inspect/tx_context_result_across_mut_tx_use_invalid.move
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Flag-off twin of tx_context_result_across_mut_tx_use. Dev-inspect permits reference
+// return values even without --enable-feature-flags allow_references_in_ptbs, but without the flag every
+// injected TxContext shares a single borrow root, so the `&mut Y` result (which extends
+// that root) conflicts with the later `&mut TxContext` injection in `mut_tx`.
```
tnowacki: Maybe like... all these test comments are a bit long lol 
😵‍💫

tnowacki (commented):
I think the tests look fine? Though a lot of them aren't really about the tx context, and are more about other reference safety checks. 
Like probably only one or two of these would be necessary to see that the TxContext isn't involved in the sources.

But I'm fine with more overall coverage!

tnowacki (approved):

crates/sui-e2e-tests/tests/address_balance_compatibility_tests.rs
cgswords: Nope, removed.

crates/sui-adapter-transactional-tests/tests/dev_inspect/tx_context_result_across_mut_tx_use.move
cgswords: Fixed 🐕

crates/sui-adapter-transactional-tests/tests/dev_inspect/tx_context_result_across_mut_tx_use_invalid.move
cgswords: Fixed 🐕

PR #27138, opened 2026-07-03
[consensus adapter] end submission when transaction is rejected or dropped

mwtian:
## Description

A consensus adapter submission only completes when its transaction key is recorded as consensus-processed, which never happens for user transactions that consensus votes to reject or that the commit handler drops. Each such submission parks until end of epoch holding an inflight slot and a submit-semaphore permit, so sustained rejections can clog the adapter. #27133 covers the lock-conflict dropped case by marking those digests processed, but that mechanism cannot cover rejections — rejection is per-position and a rejected digest must stay resubmittable, while a processed digest is suppressed at resubmission and skipped by post-consensus dedup.

Instead, settle user-transaction submissions on the per-position `ConsensusTxStatus` (Finalized / Rejected / Dropped) that the commit handler assigns once the block is sequenced. To make that signal total, the commit handler now assigns a terminal `Dropped` status on the user-transaction drop paths that previously emitted no signal: invalid input objects (also marked consensus-processed), and transactions ignored at epoch close (consensus certs no longer accepted, or the block author already sent EndOfPublish) — the epoch-close paths previously relied solely on epoch termination for cleanup, which can itself stall when reconfiguration is stuck. A position whose status expires from the cache before it is read ends the submission rather than resubmitting: the block was committed and its commit processed long ago, so a terminal outcome existed and was merely missed, and transaction-level retries belong to the client. System messages receive no per-position statuses and keep the processed-notification path, and the race against the processed/checkpoint notifications remains as the cross-position and state-sync backstop.

Also, per review feedback on #27133: a digest-keyed resubmission of a consensus-processed transaction is now checked against the epoch owned-object lock table (same conflict logic as post-consensus, covering the window before the conflict winner executes) and revalidated against live state, returning the concrete non-retriable error (`ObjectLockConflict`, or stale owned-object version once the winner executed) instead of the retriable `TransactionProcessing` suppression; the invalid-input drop path records a terminal reject reason; and `consensus_handler_dropped_transactions` is labeled by drop reason (`lock_conflict`, `invalid_input`, `end_of_epoch`, `end_of_publish`).

## Test plan

New tests: the adapter settles (releasing its inflight slot) when its position is rejected, when a soft bundle has mixed terminal statuses, and when a position status expires — without resubmitting; commit-handler tests that a voting-rejected transaction gets a Rejected status without being marked consensus-processed, and that transactions ignored at epoch close (certs closed, and post-EndOfPublish author) get a Dropped status and per-reason drop counts; submission tests that resubmitting a processed transaction returns the lock-conflict error while the winner holds the lock, and the stale-version validation error after the winner executed, rather than TransactionProcessing.

### DIFF
--- a/crates/sui-core/src/consensus_adapter.rs
+++ b/crates/sui-core/src/consensus_adapter.rs
@@ -11,7 +11,7 @@ use std::time::Instant;
 use consensus_core::BlockStatus;
 use futures::FutureExt;
 use futures::StreamExt;
-use futures::future::{self, Either, select};
+use futures::future::{self, Either, join_all, select};
 use futures::stream::FuturesUnordered;
 use mysten_common::debug_fatal;
 use mysten_metrics::{
@@ -44,6 +44,9 @@ use tokio::time::{self};
 use tracing::{Instrument, debug, debug_span, info, instrument, warn};
 
 use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
+use crate::authority::consensus_tx_status_cache::{
+    ConsensusTxStatus, NotifyReadConsensusTxStatusResult,
+};
 use crate::checkpoints::CheckpointStore;
 use crate::consensus_handler::{SequencedConsensusTransactionKey, classify};
 use crate::epoch::reconfiguration::{ReconfigState, ReconfigurationInitiator};
@@ -58,6 +61,7 @@ pub struct ConsensusAdapterMetrics {
     pub sequencing_certificate_success: IntCounterVec,
     pub sequencing_certificate_failures: IntCounterVec,
     pub sequencing_certificate_status: IntCounterVec,
+    pub sequencing_certificate_settled_status: IntCounterVec,
     pub sequencing_certificate_inflight: IntGaugeVec,
     pub sequencing_acknowledge_latency: HistogramVec,
     pub sequencing_certificate_latency: HistogramVec,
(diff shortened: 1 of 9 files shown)


crates/sui-core/src/consensus_adapter.rs
```diff
@@ -612,11 +623,48 @@ impl ConsensusAdapter {
...
+                                .wait_for_position_statuses(&consensus_positions, epoch_store)
+                                .await
+                            {
+                                Some(statuses) => break SequencingOutcome::Sequenced(statuses),
+                                None => {
+                                    // When positions expired before its status was read,
+                                    // resubmit like garbage collection.
+                                    // Duplicates are deduplicated post-consensus,
+                                    // and the processed waiter cancels this loop if the
+                                    // transaction has already finalized.
```
akichidis: If that's the case why we need to retry then? It seems like we try to tackle a potential race condition with a resubmission?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2672,6 +2672,9 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
                 {
                     let tx = tx_with_claims.tx();
                     let Some(owned_object_refs) = owned_object_refs_to_lock(tx_with_claims) else {
+                        // Invalid input object error is deterministic across all validators.
+                        status_updates.push((position, ConsensusTxStatus::Dropped));
+                        dropped_transaction_keys.push(parsed.transaction.key());
```
akichidis, line 2709: Could you also please use the metric `self.metrics.consensus_handler_dropped_transactions.inc();` here as well? Maybe we could even add a label to differentiate the reason/position where those are dropped.

akichidis (approved):

crates/sui-core/src/consensus_adapter.rs
mwtian: Good point, I will leave this case to be retried by transaction driver if needed.

PR #27140, opened 2026-07-03
[graphql-stream] Resumable transactions subscription [15/n]

tpham-mysten:
## Description

Part of the GraphQL streaming series (follows #26714, checkpoints [14/n]). Adds a resumable `transactions` subscription: it backfills the filter's matching transactions from a resume point via the scanning API, then hands off to the live checkpoint broadcast at a pinned seam, so delivery is contiguous with no gap or duplicate. The backfill scan retries transient errors (e.g. a rolling indexer deploy) with bounded exponential backoff before it gives up and disconnects. Staging-gated and not yet exposed.

Each payload is a batch of transaction edges (`[TransactionEdge!]!`) rather than a single edge: the backfill packs matches up to a fixed batch size so a deep scan coalesces its navigation reads, while live delivers each checkpoint's matches as one batch. Per-transaction cursors keep resume exact even when a batch splits a checkpoint.

The scan-then-live handoff, sparse-filter coverage, cursor unification, batching, and retry are documented at the top of `api/subscription/transactions.rs`.

## Test plan

- e2e parity test (`test_transaction_subscription_live_backfill_parity`): the same transactions resolve identically whether delivered live or through the backfill scan, plus the existing streaming/ordering/resume subscription tests.
- Batching e2e tests: `test_transaction_subscription_backfill_batches_matches` (a deep backfill packs matches into multi-edge payloads) and `test_transaction_subscription_live_batches_per_checkpoint` (each live payload is one checkpoint's matches).
- Unit tests for the scan retry: recovers within the budget, gives up once it is exhausted.

---

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425
   - #26495
   - #26731
   - #26714
   - #27140

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Support transaction subscription from historical cursor
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-graphql/src/api/types/transaction/filter.rs
tpham-mysten, line 1: I borrow this logic from @wlmyng's PR #26871

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,348 @@
...
+) -> impl Stream<Item = Result<Edge<String, Transaction, EmptyFields>, RpcError>> {
+    let handoff_threshold = config.broadcast_buffer as u64 / 2;
+    let proto_filter = filter.to_bitmap_filter();
+
+    stream! {
+        let mut pending_receiver = None;
+        let mut handoff: Option<u64> = None;
+        let mut last_checkpoint: Option<u64> = None;
+
+        // Phase 1: backfill toward the tip, pinning the live receiver near it.
```
tpham-mysten: For `transactions_stream`, the scan-then-live approach is quite similar to checkpoint resumption, but I intentionally kept it as its own path rather than a shared driver (I tried extracting one earlier and backed it out). The phase-1/phase-2 skeleton (backfill, pin a handoff near the tip, follow live) is common, but the plug-in points aren't: checkpoints enumerate sequence numbers and yield a neutral `ProcessedCheckpoint`, while transactions page a filter-driven scan with sparse matches + coverage markers + opaque cursors and build GraphQL edges directly (filter, hydrate, mint cursor). A generic driver ends up mostly callbacks, which buries the genuinely subtle part (the pin/seam race). 

With the events subscription coming, transactions and events should be able to share a common core (both are filter-driven scans that build edges), so I expect to extract it once events lands. Checkpoint might still stay on its own, given it enumerates sequence numbers and yields a neutral type, but we will see.

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -44,14 +44,30 @@ pub struct StreamPage<T> {
     pub items: Vec<PageItem<T>>,
     first_wm_cursor: Option<Bytes>,
     last_wm_cursor: Option<Bytes>,
+    /// Highest checkpoint the server fully scanned this page (max `checkpoint_hi` over its
+    /// watermarks). Reports how far the scan reached, not where the last match was, so it can run
+    /// ahead of the last matched item through non-matching checkpoints.
+    ///
+    /// E.g. matches in cp 101 and 102, then 103-104 scanned with no matches before the page ends:
+    /// `checkpoint_hi` = 104.
+    checkpoint_hi: Option<u64>,
```
wlmyng: `hi` implies exclusivity, but this represents the accumulated completion boundary as of that frame (computed server side correctly for either item or watermark.) Maybe a better name is `checkpoint_boundary` or something? Or the doc comment reflects this

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -144,16 +160,38 @@ impl<T> StreamPage<T> {
...
+    /// Highest checkpoint fully scanned this page (see the field doc). `None` means the page reported
+    /// no scan progress (genesis, or a caught-up short-circuit); callers treat it as "no coverage to
+    /// add this page".
+    pub fn checkpoint_hi(&self) -> Option<u64> {
+        self.checkpoint_hi
+    }
+
+    /// Remember the latest `checkpoint_hi` a frame reported. The server emits it non-decreasing, so
+    /// the latest non-`None` value is the highest; `None` frames (genesis) are ignored.
+    fn record_checkpoint_hi(&mut self, checkpoint_hi: Option<u64>) {
```
wlmyng: i think this method is self-documenting

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -144,16 +160,38 @@ impl<T> StreamPage<T> {
             .or_else(|| self.items.last().map(|item| &item.cursor))
     }
 
+    /// Highest checkpoint fully scanned this page (see the field doc). `None` means the page reported
+    /// no scan progress (genesis, or a caught-up short-circuit); callers treat it as "no coverage to
+    /// add this page".
+    pub fn checkpoint_hi(&self) -> Option<u64> {
```
wlmyng: I think we can decouple this accessor from the process that the comment is current narrating. This represents the completion boundary as of the page's last boundary-bearing frame (item or watermark.) In other words, this checkpoint is the one that the server could vouch has fully scanned up to. If it's `None` then the page makes no such coverage claim (doesn't mean we covered checkpoint 0, for example)

crates/sui-indexer-alt-graphql/src/api/types/transaction/filter.rs
```diff
@@ -236,6 +237,86 @@ impl TransactionFilter {
 
         true
     }
+
+    pub(crate) fn to_bitmap_filter(&self) -> Option<v2alpha::TransactionFilter> {
```
wlmyng: do we want to return a bad user input error if clients provide after/at/beforeCheckpoint on the filter?

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,465 @@
...
+                yield Ok(Scanned { checkpoint, edge: Some(edge) });
+            }
+            // Emit a coverage marker so the handoff can advance through pages with no match.
+            if let Some(checkpoint) = checkpoint_hi {
+                yield Ok(Scanned { checkpoint, edge: None });
+            }
+
+            // Advance to the next page; a caught-up short-circuit has no cursor, so re-request as-is.
+            if let Some(cursor) = next_cursor {
+                position = ResumeFrom::Cursor(cursor);
```
wlmyng: i think we also want to surface this so the client can resume. Right now it seems like clients can only advance on matches should they disconnect

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,465 @@
...
+    resume: ResumeFrom,
+) -> impl Stream<Item = Result<Scanned, RpcError>> {
+    stream! {
+        let mut position = resume;
+        loop {
+            let page = list_with_retry(scan_backoff(), || {
+                reader.list_transactions(build_list_request(&position, &proto_filter))
+            })
+            .await?;
+            let checkpoint_hi = page.checkpoint_hi();
```
wlmyng: So, every watermark's `checkpoint` should equal its own cursor's checkpoint - 1. By that, some item at `(C, t)` has a cursor cp at `C`, with boundary at `C-1`. Frontier watermark at F has cursor cp at F, boundary at `F-1`.

So we can represent coverage as 
```
page.last_cursor()
.map(CursorToken::decode)
.transpose()?
.and_then(|t| t.position.checkpoint().checked_sub(1))
```

I think this relation is acsending-only, which works for subscriptions as it only scans acsending

wlmyng (commented):
looks p good but we should also surface cursors from watermark (on no matches). also biased towards not having the new `checkpoint_hi` field on the page as the same information is present in the cursors, which graphql also understands

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
tpham-mysten: Agreed. Rather than rename, I removed the field entirely since it just duplicated the cursor as you suggested

crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
tpham-mysten: Ack and it's a great point. I have removed this checkpoint_hi()

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: SG. I have removed the old checkpoint_hi() and use the `last_cursor()` logic instead

crates/sui-indexer-alt-graphql/src/api/types/transaction/filter.rs
tpham-mysten: SG. Applied

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: Good call-out, but I'd hold off and keep the payload a plain edge list. Emitting a no-match payload carrying a watermark cursor breaks the current model and leaks scan internals into the public shape.

The cost it avoids is also small: the scan already fast-forwards through no-match ranges via the bitmap (round-trips scale with matches, not checkpoints), so a client resuming from its last match just replays that cheap fast-forward to reach live. It re-scans but never re-delivers.

WDYT?

tpham-mysten (commented):
Thanks @wlmyng! I have dropped checkpoint_hi as suggested.

For surface cursors from watermark (on no matches), I'm leaning towards not supporting that, and you can check more details in the comment

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
wlmyng: Oh I see, `Subscription::transactions` returns edges not a connection

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/snapshots/graphql_subscription__transaction_subscription__transaction_subscription.snap
```diff
@@ -4,41 +4,43 @@ expression: item
...
-        },
-        "effects": {
-          "status": "SUCCESS",
-          "balanceChanges": {
-            "nodes": [
-              {
-                "amount": "1000",
-                "coinType": {
-                  "repr": "[pkg]::sui::SUI"
+    "transactions": [
```
wlmyng: this is just a consequence of us returning batches (vec) instead of a single txn per payload

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/transaction_subscription.rs
```diff
@@ -94,11 +106,508 @@ async fn test_transaction_subscription_object_changes() {
...
+    let (digest, _) = create_item(&mut cluster.validator, package_id, 42).await;
     let item = wait_for_matching_item(&mut stream, &[digest], transaction_digest).await;
 
     graphql_redactions().bind(|| {
         insta::assert_json_snapshot!("transaction_subscription_object_changes", item);
     });
 }
+
+/// Field coverage: publishes a Move package, calls into it, and probes a broad set of transaction,
+/// effects, and Move-object fields at once. The snapshot documents which fields resolve in streaming
```
wlmyng: as in, do we expect all the selected fields below to resolve? and any fields not named should error?

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/transaction_subscription.rs
```diff
@@ -94,11 +106,508 @@ async fn test_transaction_subscription_object_changes() {
...
+    let backfill_nodes = collect_nodes(&mut backfill, &expected).await;
+
+    // 5. Identical across phases (parity), and matches the recorded shape (correctness).
+    assert_eq!(
+        live_nodes, backfill_nodes,
+        "live and backfill resolved the same transactions differently",
+    );
+    rich_tx_redactions().bind(|| {
+        insta::assert_json_snapshot!("transaction_subscription_parity", live_nodes);
+    });
```
wlmyng: does this snapshot provide additionally meaningful value? it seems the live vs backfill parity check is the main add of this test?

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/transaction_subscription.rs
```diff
@@ -94,11 +106,508 @@ async fn test_transaction_subscription_object_changes() {
...
+        }
+    }
+    assert!(
+        checkpoints_seen.len() >= 2,
+        "expected matches across >= 2 checkpoints, got {checkpoints_seen:?}",
+    );
+}
+
+/// The `{ "sender": ... }` variables shared by the filtered subscription queries.
+fn sender_var(sender: SuiAddress) -> Option<Value> {
```
wlmyng: nit fns before tests?

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/transaction_subscription.rs
```diff
@@ -94,11 +106,508 @@ async fn test_transaction_subscription_object_changes() {
...
+    settings.add_redaction(".**.lamportVersion", "[lamportVersion]");
+    settings.add_redaction(".**.gasSummary", "[gasSummary]");
+    settings.add_redaction(".**.json", "[json]");
+    settings.bind(|| {
+        insta::assert_json_snapshot!("transaction_subscription_field_coverage", item);
+    });
+}
+
+/// Decode a transaction edge's `cursor` field into its `(checkpoint, tx_sequence)` position.
+fn decode_tx_cursor(edge: &serde_json::Value) -> (u64, u64) {
```
wlmyng: nit organize helpers before tests

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,472 @@
...
+//!
+//! - Phase 1 scans the filter's matches forward: checkpoint 6, 7, 8, ... toward the tip.
+//! - As it nears the tip, it pins the handoff: it subscribes to the live broadcast (which will
+//!   deliver checkpoint 11 next) and records `handoff = 10`.
+//! - Phase 1 finishes delivering matches through checkpoint 10, then stops.
+//! - Phase 2 takes over from the live broadcast: checkpoint 11, 12, 13, ...
+//!
+//! The seam at 10 -> 11 has no gap and no duplicate. Subscribing *before* reading the tip is what
+//! guarantees it: the live feed's first checkpoint is always `handoff + 1` or earlier (any overlap
+//! is dropped), never past it. See [`live_transactions`] for the overlap skip and the gap check.
```
wlmyng: i think what's missing here, is that at some point the `handoff` between backfill pagination and live subscription is pinned, so backfill is not chasing a constantly moving "tip". When this handoff is pinned, the live subscriber starts working and filling up the buffer, thus ensuring that there's no gap

in practice the buffer threshold value is a small number of checkpoints, 256 checkpoints, which should be easily handled by pagination (and at 4 cp/s buys about 60 seconds of buffer time)

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,472 @@
...
+            if !has_more {
+                tokio::time::sleep(Duration::from_millis(100)).await;
+            }
+        }
+    }
+}
+
+/// Follow the live broadcast from `last_checkpoint + 1`, delivering matching transactions. Drops the
+/// one-checkpoint seam overlap (the resubscribe/tip race), gap-checks, and disconnects on anomalies.
+fn live_transactions(
```
wlmyng, line 249: seems we could add a test feeding checkpoints `[10, 10, 11, 13]` and expect on skip/deliver/disconnect

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,472 @@
...
+        let mut pending_receiver = None;
+        let mut handoff: Option<u64> = None;
+        let mut last_checkpoint: Option<u64> = None;
+
+        // Phase 1: backfill toward the tip, pinning the live receiver near it. Matches are batched
+        // by count (not by checkpoint) so a deep backfill coalesces its reads: a batch fills to
+        // `max_batch` edges then flushes, spanning or splitting checkpoints freely. Per-transaction
+        // cursors make mid-checkpoint resumption exact, so splitting is safe. The reactive throttle
+        // then meters one batch at a time.
+        if let Some(resume) = resume {
```
wlmyng, line 116: i think there are some components here worth testing explicitly
- does it matter if pinning is triggered by coverage marker vs a match?
- L155: match-at-handoff: assert on sending `handoff` checkpoint from scan but skipping matches past this `handoff` cutoff
- L169 test coverage to assert that we break when `checkpoint >= h`

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,472 @@
...
+            for await scanned in scan {
+                let Scanned { checkpoint, edge } = scanned?;
+
+                // Resubscribe-first and pin once within threshold of the tip (match or coverage).
+                if pending_receiver.is_none()
+                    && broadcast.network_tip().saturating_sub(checkpoint) <= handoff_threshold
+                {
+                    pending_receiver = Some(broadcast.broadcaster().resubscribe());
+                    handoff = Some(broadcast.network_tip());
+                }
```
wlmyng, line 141: and testing what happens when L145 `broadcast.network_tip()` yields a different checkpoint than L148 (not an issue today since the implementation and ordering ensures no gaps)

wlmyng (approved):
lgtm but some comments on things we could test assert on

crates/sui-indexer-alt-graphql/staging.graphql
```diff
@@ -4149,13 +4149,13 @@ type Subscription {
...
-	checkpoints. Transactions are ordered by checkpoint, then by position within
-	the checkpoint.
+	Pass `after` (opaque cursor) or `afterCheckpoint` (sequence number), but not both, to resume from a known point. The subscription first backfills the matching transactions after that point via the scanning API, then continues with the live stream.
+	
+	Each payload is a batch of matching transactions — edges ordered by checkpoint, then by position within the checkpoint — and a single batch may span multiple checkpoints. Matches are delivered in batches, rather than one per payload, for efficiency. Each edge carries a cursor for resumption.
 	
 	This subscription is not yet available for use.
 	"""
-	transactions(filter: TransactionFilter): TransactionEdge!
+	transactions(filter: TransactionFilter, after: String, afterCheckpoint: UInt53): [TransactionEdge!]!
```
amnn: This is unexpected/non-idiomatic, why are we returning a batch here, how does it impact efficiency -- is that measurable?

crates/sui-indexer-alt-graphql/src/scope.rs
```diff
@@ -61,6 +61,13 @@ pub(crate) enum DataSource {
     Streamed {
         checkpoint: Arc<ProcessedCheckpoint>,
     },
+    /// A scanned, already-finalized transaction's input/output objects, taken from the scanning
```
amnn: It does not make sense to treat this as a special case for scanning/resumption, which is in all ways similar to paginating over transactions -- we should just serve nested queries through the normal indexed pathways.

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,472 @@
...
+use crate::task::streaming::ProcessedCheckpoint;
+use crate::task::streaming::StreamingPackageStore;
+use crate::task::streaming::SubscriptionBroadcast;
+use crate::task::streaming::broadcast_error;
+use crate::task::streaming::hydrate_executed_transaction;
+use crate::task::streaming::reconnect_error;
+
+/// Backfill scan batch size: matches are packed into one payload up to this many, so a deep backfill
+/// coalesces its reads. Live ignores it (a whole checkpoint is the natural unit).
+const SCAN_MAX_TRANSACTIONS_PER_BATCH: usize = 100;
```
amnn: This should be consistent with the max page size for transaction pagination (we should use that value instead of defining a new one).

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,472 @@
...
+    package_store: Arc<StreamingPackageStore>,
+    resolver_limits: sui_package_resolver::Limits,
+    proto_filter: Option<proto::TransactionFilter>,
+    resume: ResumeFrom,
+) -> impl Stream<Item = Result<Scanned, RpcError>> {
+    stream! {
+        let mut position = resume;
+        loop {
+            let page = list_with_retry(scan_backoff(), || {
+                reader.list_transactions(build_list_request(&position, &proto_filter))
```
amnn: Rather than this code directly calling into `AlphaLedgerGrpcReader`, could we have it use the `Transaction::paginate` API?

amnn (commented):
I think there's more opportunities for re-use here between the paginated and scanning/backfill codepaths.

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/snapshots/graphql_subscription__transaction_subscription__transaction_subscription.snap
tpham-mysten: Right!

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/transaction_subscription.rs
tpham-mysten: Done

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/transaction_subscription.rs
tpham-mysten: Done

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: Good catch. I added that

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: This is quite similar to this comment (https://github.com/MystenLabs/sui/pull/27140/changes#r3596644193), so I reply there

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: For first bullet, I added test_transaction_subscription_empty_backfill_hands_off_to_live

For second and third, I added test_transaction_subscription_exactly_once_across_handoff to make sure there is no duplicate in the stream result

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: Ah good catch. I fixed that by sampling the tip.

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: skip/deliver are covered by the two new handoff tests (empty_backfill_hands_off_to_live, exactly_once_across_handoff). The gap-disconnect (seq > last + 1) is a defensive guard that never
fires normally (contiguous broadcast; a lagging receiver hits the separate Lagged branch), so I would prefer not having a test for that

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/transaction_subscription.rs
tpham-mysten: Yeah, all the fields we list here should resolve without an error. Fields we did not list are not unsupported, they just are not checked in this test as we just focus on critical field

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/transaction_subscription.rs
tpham-mysten: Good catch. I have removed it

crates/sui-indexer-alt-graphql/src/scope.rs
tpham-mysten: I think this is quite similar to your other comment about whether scanning can just use `Transaction::paginate`, so I'll answer both here.

The reason I kept this off the indexed pathway is that a subscription resolves fields differently from pagination.

1. `Transaction::paginate` is anchored at `checkpoint_viewed_at` for query consistency and to return everything up to that checkpoint. A subscription has no single "as of" checkpoint: it delivers transactions as they finalize and resolves each one from the data it streamed, so there's no result set to bound and nothing to anchor.

2. There's also an efficiency angle: paginate read-masks digest-only and resolves each field lazily through the index, a round trip per field, whereas the scanning reader already returns the full checkpoint contents via the read mask, so we hydrate straight from those and skip the round trip. It also keeps the backfill consistent with the live path, which delivers a whole checkpoint the same way.

For those two, we could adapt paginate to allow a None `checkpoint_viewed_at` or a read mask for the full checkpoint, but I'm not sure that's cleaner than calling AlphaLedgerGrpcReader directly, or that we earn much from it.

So I'd prefer keep scanning calling AlphaLedgerGrpcReader directly rather than fitting it into Transaction::paginate.

crates/sui-indexer-alt-graphql/staging.graphql
tpham-mysten: it's because async-graphql resolves a subscription one payload at a time: it fully resolves a yielded value and
sends it before pulling the next (https://github.com/async-graphql/async-graphql/blob/82cd1f15c2a66c6134be6b93e3ac6331847934c6/src/subscription.rs#L52)

So if we emit one transaction per payload, a checkpoint with say 100 matches resolves those 100 sequentially, and the DB reads each one does can't be batched across them. That only gets worse as the number of matches in a checkpoint grows. Delivering the checkpoint's matches as a single list keeps them in one payload, where `async-graphql` resolves the edges concurrently and their reads batch (via DataLoader), so it stays one round of work instead of N.

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: I think these are quite orthogonal, since they solve different problems.

IIUC, max_page_size exists to cap how much one request can demand. OTOH, this batch cap is internal, it's how many matched transactions we coalesce into a single subscription payload, which is the concurrency tradeoff for the speed of the scan.

So the value that's right for a client-facing page limit isn't necessarily right here. If we want this tunable I'd make it a SubscriptionConfig knob rather than borrow the Query page size.

crates/sui-indexer-alt-graphql/staging.graphql
amnn: Got it -- agreed we should fix that performance issue, but we should find a way to do that without leaking it into the schema.

crates/sui-indexer-alt-graphql/src/scope.rs
amnn: > resolves each field lazily through the index, a round trip per field, whereas the scanning reader already returns the full checkpoint contents via the read mask, so we hydrate straight from those and skip the round trip.

This should not be the case -- pagination implemented on top of gRPC can/should read transaction contents alongside digests (but probably not the object set etc).

To me, the differences you are describing sound incidental and non-desirable. Both scanning and pagination built on top of gRPC should behave the same way in terms of what is eagerly fetched, how limits are applied, what is considered "latest data" etc.

amnn: Maybe it's worth raising as a design topic on Thursday.

crates/sui-indexer-alt-graphql/staging.graphql
tpham-mysten: I have been thinking about this and have a few options.

**Option 1** — Keep the batched schema (current)
- Pros: works today, coalesces every DB read generically (nested included), no maintenance.
- Cons: non-idiomatic, and inconsistent with the checkpoint/event subscriptions, which are single-edge.

**Option 2** — Pipeline subscription resolution in the fork
`async-graphql` resolves subscription payloads serially (`StreamExt::then`). Make it configurable:
buffered(N) resolves N items concurrently with output order preserved, so one-per-payload coalesces
reads like the batch, generically, for every subscription. Opt-in via a `SchemaBuilder` knob, default
serial.
- Pros: idiomatic and fully general (nested + all subscriptions), no per-field prefetch to maintain.
- Cons: it's a fork change that upstream async-graphql doesn't support, so there may be something I
  haven't foreseen that blocks it; and the one real fix it needs is per-item error isolation, so
  concurrently-resolved payloads don't share an error sink.
- Draft PR: https://github.com/amnn/async-graphql/pull/2

**Option 3** — One-per-payload + prefetch
Pull the keys a checkpoint's matches need out of the payload and batch-load them before yielding, so
serial per-payload resolution hits a warm cache.
- Pros: idiomatic schema, no framework change.
- Cons: only covers the first hop. A nested DB field (previousTransaction inside previousTransaction)
  is keyed by data we haven't fetched yet, so it falls back to serial. Also needs a prefetch rule per
  DB-backed field.

I lean Option 2 as the cleanest, but if there is some blocker on Option 2,  I think keeping the batch (Option 1) is not the end of the world either. It's just batching, and it works fine.

WDYT?

amnn: I agree, Option 2 seems like the cleanest, and this also seems like a very reasonable candidate for upstreaming.

tpham-mysten: Sure. I have started mplementing Option 2 by supporting concurrent resolve in https://github.com/amnn/async-graphql/pull/2

When that PR is merged. I will send a separate PR to incorporate it

crates/sui-indexer-alt-graphql/src/scope.rs
tpham-mysten: As discussed in the weekly. We will have a scan function from Transaction to support use case with Non `checkpoint_view_at`. 

I have updated the scan logic to use `scan_grpc` from https://github.com/MystenLabs/sui/pull/27478/changes

tpham-mysten (commented):
Thanks all! 2 main blockers have been cleared:
- For concurrency resolve. I have sent PR https://github.com/amnn/async-graphql/pull/2 and will make a follow-up PR to incorporate that concurrency when the PR is merged
- For pagination, I have incorporated the `Transaction::scan_grpc` (from #27478) into the transaction backfill

crates/sui-indexer-alt-graphql/src/api/types/transaction_effects.rs
```diff
@@ -593,12 +589,15 @@ impl EffectsContents {
             return Ok(self.clone());
         };
 
-        let cp_num = transaction
-            .cp_sequence_number()
-            .context("Fetched transaction should have checkpoint sequence number")?;
-
-        if cp_num > checkpoint_viewed_at {
-            return Ok(self.clone());
+        // Enforce the consistency cutoff only when viewing as of a specific checkpoint. A
```
amnn, line 592: @tpham-mysten this does not seem like a safe/correct change in the context of a transaction that was just executed or simulated. In that case we also do not have `checkpoint_viewed_at` set in the scope, and we do not want to try and fetch the effects from the KV loader (that was the original reason this test existed).

I understand that you made this change so that the effects could be fetched from KV loader on a transaction that is being backfilled during streaming, but we need a way to distinguish between these scenarios, maybe taking into account the `DataSource` enum in `Scope`?

Until we have designed that properly we should roll back this part of the change.

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,429 @@
...
+        // No end cursor means nothing past the indexer tip is indexed yet; wait, then re-request. A
+        // range that was scanned but matched nothing still returns a boundary cursor, not `None`.
+        let Some(end_cursor) = result.last_cursor() else {
+            tokio::time::sleep(BACKFILL_POLL_INTERVAL).await;
+            continue;
+        };
+        let end_cursor = CursorToken::decode(end_cursor).context("Failed to decode scan cursor")?;
+
+        // Hold the page until both pipelines have indexed through its end.
+        wait_for_pipelines_catching_up_at(end_cursor.position.checkpoint(), watermarks_rx).await?;
```
amnn, line 362: Is this necessary? I thought the idea was that we only perform this check during start-up and from then on, any request that tries to fetch something "in the future" would fail. Otherwise, what happens to this code path if I provide a checkpoint that is way off in the future?

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,429 @@
...
+        let edges = build_grpc_connection(scope.clone(), &page, result)?.edges;
+        assert_eq!(edges.len(), checkpoints.len(), "one edge per scanned item");
+
+        let mut items = Vec::with_capacity(edges.len() + 1);
+        for (i, edge) in edges.into_iter().enumerate() {
+            items.push(Scanned {
+                checkpoint: checkpoints[i],
+                edge: Some(edge),
+            });
+        }
```
amnn, line 383: This sequence reads as though we're working too hard to re-use `build_grpc_connection`:

- `build_grpc_connection` is doing more work than you need it to -- you only care about the `edge`s.
- It looks like what you want to do is call `zip_eq` on `result.items` and `edges`, but you can't because `build_grpc_connection` consumes` result`.

Can we factor out the edge creation from `build_grpc_connection` and use that instead? Ideally in a way where you don't consume the `result` so you can perform that zipping operation.

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -0,0 +1,429 @@
...
+            limits,
+            Some(limits.default as u64),
+            after.cloned(),
+            None,
+            None,
+        )
+        .map_err(anyhow::Error::from)?;
+        match Transaction::scan_grpc(reader, cp_bounds.clone(), &page, filter).await {
+            Ok(result) => return Ok((page, result)),
+            Err(e) => match backoff.next_backoff() {
```
amnn, line 420: Is it not possible to use one of `backoff`'s helper functions to handle the loop? Also the loop only needs to encapsulate `scan_grpc` -- we don't need to recreate `page` for each iteration.

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
```diff
@@ -91,60 +117,75 @@ impl Subscription {
...
+        after: Option<String>,
+        after_checkpoint: Option<UInt53>,
     ) -> Result<
         impl futures::Stream<Item = Result<Edge<String, Transaction, EmptyFields>, RpcError>>,
-        RpcError,
+        RpcError<Error>,
     > {
+        // `after` (resume from a specific transaction) and `afterCheckpoint` (resume from a
+        // checkpoint) are distinct resume modes, not bounds to reconcile, so only one is allowed.
+        if after.is_some() && after_checkpoint.is_some() {
```
amnn, line 137: Do you need to introduce this mutual exclusion? The cursor includes the checkpoint sequence number, so it should be possible to intersect the two filters.

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
```diff
@@ -91,60 +117,75 @@ impl Subscription {
...
+
+        // Decode `after` here so a bad cursor surfaces as `BadUserInput`; the backfill resumes
+        // pagination from it.
+        let resume = if let Some(cursor) = after {
+            let ctransaction = CTransaction::decode_cursor(&cursor)
+                .map_err(|_| bad_user_input(Error::InvalidCursor(cursor)))?;
+            Some(ResumeFrom::Cursor(ctransaction))
+        } else {
+            after_checkpoint.map(|cp| ResumeFrom::Checkpoint(u64::from(cp)))
+        };
```
amnn, line 176: Cursor parsing should not be handled as part of the body here -- we should be accepting the cursor typed as part of the parameters, so that `async_graphql`'s input parsing handles it for us (take a look at other cases where we accept a cursor as a parameter -- we take it as its type, not as a `String`).

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
```diff
@@ -91,60 +117,75 @@ impl Subscription {
...
+        // How many matching transactions the backfill scans per page.
+        let scan_page_size = SCAN_PAGE_SIZE;
+
+        // Pin the handoff once the scan comes within half the live buffer of the tip, leaving room
+        // for checkpoints that arrive during the handoff so the receiver does not lag.
+        let handoff_threshold = config.broadcast_buffer as u64 / 2;
+
+        // A subscription streams forward from its resume point, so filter-level checkpoint bounds
+        // have no meaning; reject them rather than silently dropping them.
+        if filter.after_checkpoint.is_some()
```
amnn, line 161: Why shouldn't we support checkpoint bounds? In fact, it's a bit confusing that we now have two parameters called `afterCheckpoint` one of which creates an error and another of which is fine.

It would have been preferable to treat the `afterCheckpoint` inside `TransactionFilter` as the `afterCheckpoint` for resumption and not introduce another parameter for this.

amnn (commented):
Spotted a couple of things that we should address and/or discuss. The issue related to the `checkpoint_viewed_at` check inside effects content hydration is probably the most significant thing, but the API design around checkpoint filters is also important.

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
amnn: > IIUC, max_page_size exists to cap how much one request can demand. OTOH, this batch cap is internal, it's how many matched transactions we coalesce into a single subscription payload, which is the concurrency tradeoff for the speed of the scan.

You're right that one of these limits is externally facing and the other is used internally, but their intentions seem the same to me: To limit the amount of data we can extract from downstream data sources in one shot. Put another way: When would you set a different max scan page size to the max transaction page size?

crates/sui-indexer-alt-graphql/src/api/types/transaction_effects.rs
tpham-mysten: Thanks for the catch, this was a regression. Rather than reverting, I went with the `DataSource::Executed` check: execution and simulation scopes now never read the store, while indexed and streaming scopes hydrate as before. PR: https://github.com/MystenLabs/sui/pull/27589

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
tpham-mysten: Yes we can. I was thinking putting 2 options together in the argument is more explicit, but yeah, it's confusing given that filter already has `afterCheckpoint`.

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
tpham-mysten: I was thinking that cursor is opaque and if we intersect the two filters, we kinda expose the implementation details of gRPC cursor. 

But yeah, users have to know that opaque cursor will refer to a transaction, and with that mental model, we can do the intersect.

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
tpham-mysten: SG. I will fix it

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: Sounds good. I have made that split in https://github.com/MystenLabs/sui/pull/27596

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: Yes we can with backoff::future::retry. I initially had a loop and Claude put this on top of that loop 😅

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: You're right about the live path: the start-up check waits once for `kv_packages` to reach the tip, and past the tip live checkpoints carry their packages in the in-memory store, so live never re-waits on the indexer.

The wait is there for the scanning phase, where a backfilled item still resolves fields like objects by reading the indexer rather than from an in-memory checkpoint. We could avoid that by having `scan_grpc` return the full proto checkpoint the way live does, so those fields resolve from the payload and stay consistent with live. But that carries the waste we discussed before, and in practice this wait is a no-op for all but the last few checkpoints and only a brief pause near the tip, so I think it's fine to keep it as-is for now.

Also, when I first added this, I also thought it wasn't necessary, assuming that once a full node delivers checkpoint N all indexers are caught up at N. But a subsequent object resolve might land on a different replica, so we still need it here.

On the far-future checkpoint, I think that's an edge case and kinda orthogonal to this wait. There's nothing to backfill ahead of the tip, so it comes down to waiting until the stream reaches that checkpoint and then delivering from live. We could also just reject a far-future resume point up front.

tpham-mysten (commented):
Thanks @amnn! I have applied all the fix in follow-up PRs:
- #27589: Critical regression bug in scope (Thanks for the review)
- #27596: Extract the edge build logic 
- #27580: All remaining logic

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
amnn: Ah right, this is to handle a potential gap between the live stream and the indexed stream (which indeed could be ~5-10 checkpoints).

Usually, this should not be an issue because that should be covered by the broadcast buffer, but as the system is warming up we may need to wait -- so makes sense that we need some kind of check here.

I do foresee a complication related to keeping the set of pipelines/sources we need to maintain here in-sync with the rest of the schema though: We need to wait for all data sources that could be accessed via nested queries during scanning.

Initially, I was thinking the answer was to just wait for all watermarks to catch up, but that's not strictly necessary because, e.g. you can't perform a nested live object set query because we don't set `checkpoint_viewed_at`. However, maybe this is the simplest thing to do to avoid things going out-of-sync?

> On the far-future checkpoint, I think that's an edge case and kinda orthogonal to this wait. There's nothing to backfill ahead of the tip, so it comes down to waiting until the stream reaches that checkpoint and then delivering from live.

I agree that that's the right way to serve such a request, but if the request involves waiting for a very long time (days/months/years) then someone can quite cheaply tie up the service with requests that are just waiting, which is not good.

We should limit that -- e.g. it's fine to ask to wait for up to a minute ahead of tip-of-chain, but anymore and you get rejected.

tpham-mysten: >  However, maybe this is the simplest thing to do to avoid things going out-of-sync?

Sounds good, I'll wait for all watermarks as the consistent store lag is just ~1.23 second (p99), so the handoff pause is negligible. The tradeoff we're accepting is that a consistent-store incident would stall resumption even though scanning doesn't read that data, but it only hits catching-up clients (live subscriptions don't wait and keep streaming). We can revisit if it becomes annoying.

> We should limit that -- e.g. it's fine to ask to wait for up to a minute ahead of tip-of-chain, but anymore and you get rejected.

SG. I will implement in a follow-up PR

PR #27142, opened 2026-07-03
rpc-store: dedup per-object pruner retractions to one range tombstone

nickvikeras:
## Summary

The embedded rpc-store pruner retracted `object_version_by_checkpoint` rows with one `delete_range [id||0, id||cp)` per **(object × superseding transaction × pruned checkpoint)**. Hot objects (`0x5` SuiSystemState, `0x6` Clock — superseded in every checkpoint) therefore accumulated up to ~28,000 **nested** range tombstones (same start key `id||0`, growing end key) inside a single prune write batch / memtable.

RocksDB fragments a nested tombstone stack of depth K into **K²/2 (fragment, seqnum) pairs**: `FragmentedRangeTombstoneList::FragmentTombstones` records, for every fragment, the seqnum of *every* covering tombstone (`range_tombstone_fragmenter.cc`, inner loop over `cur_end_keys`). Flush-time re-fragmentation through `CompactionRangeDelAggregator::NewIterator` then enumerates every pair via `TruncatedRangeDelMergingIter` with **unpinned keys**, copying both keys of every pair into `pinned_slices_` (2 list nodes + 2 string buffers, ≈200B per pair, all live until the flush finishes).

This OOM-killed `sui-node-mainnet-rpc-alpha-0` (120G limit) and then wedged it permanently: WAL recovery runs the same flush path (`DBImpl::RecoverLogFiles → WriteLevel0TableForRecovery → BuildTable`), so every restart replayed the same poisoned WALs, OOMed at ~320MB/s of allocation before writing a single SST, and never truncated the WALs — a deterministic crash loop (122+ restarts, byte-identical RocksDB `LOG` every run, last line `Recovering log #128207 mode 2`).

## Evidence (mainnet-rpc-alpha, image `4ca99939`)

jemalloc heap profiles pulled from the crash-looping pod (`MALLOC_CONF=prof:true`, dumps at 64GiB allocation intervals), symbolized against the pod's binary — est. live heap 31.8GB at dump i0 → **75.9GB** at i1, 100% of it under `Db::open → DBImpl::RecoverLogFiles → WriteLevel0TableForRecovery`:

* 4 sites × ~16.8GB of **48-byte live objects** (~1.5×10⁹) — the per-pair `pinned_slices_` list nodes + string buffers from `range_tombstone_fragmenter.cc:223-235`, reached via `BuildTable → CompactionRangeDelAggregator::NewIterator → FragmentTombstones`;
* one single **8,589,934,592-byte (8GiB) allocation** — the `tombstone_seqs_` vector for 1.07×10⁹ pairs, from `MemTable::NewRangeTombstoneIteratorInternal → FragmentTombstones`.

Parsed the on-disk WALs (WriteBatch records) to quantify the write pattern:

| WAL | CF11 (`object_version_by_checkpoint`) DeleteRanges | distinct ids | max K (one object) | ΣK²/2 |
|---|---|---|---|---|
| `128209.log` | 162,885 | 11,883 | 10,641 | 1.4×10⁸ |
| `128213.log` | 457,494 | 22,587 | **28,365** | **1.07×10⁹** |
| `128217.log` | 575,083 | 29,568 | 23,587 | 1.17×10⁹ |

`128213.log`'s ΣK²/2 = 1.07×10⁹ matches the observed 8GiB `tombstone_seqs_` allocation (×8B) to three significant figures. Across the replayed WALs: ~2.4×10⁹ pairs ≈ 480GB demanded against a 120G limit.

(The earlier `cache_index_and_filter_blocks` commit targeted table-reader block memory; these allocations never touch the block cache, which is why it didn't help.)

## Fix

Coalesce retractions per object within a prune batch (`Retractions` collector) and stage exactly **one** `delete_range [id||0, id||max_cp)` per object, in both `prune_chunk` and `prune_history_cohort`. Semantics-preserving because the staged ranges for one object are nested: the widest subsumes every narrower range, including any intermediate point delete of a removed checkpoint row; on an equal checkpoint the `removed` flags are ORed (any same-checkpoint removal must still drop the row at that checkpoint). This collapses K from ~28k to 1 per object per batch, reducing fragmentation work from O(K²) to O(1) per object.

## Verification

* `cargo check -p sui-rpc-store`; pruner-scoped unit tests pass (20, including 2 new tests covering widest-range dedup and equal-checkpoint removed-OR through the public `get_object_version_at_checkpoint` read path).
* Deployed to `mainnet-rpc-alpha` (`SUI_SHA_OVERRIDE=c1fd4fff…`) and wiped the poisoned `rpc_store` dir (its existing WALs make recovery OOM regardless of code, and no open could ever complete to truncate them). Before: OOMKilled every ~8 min at the 120G limit, 132 restarts, ~320MB/s RSS climb inside `Db::open`. After: fresh store, `Bootstrap::Restore` running, RSS flat at ~11.5GB well past the old OOM horizon, 0 restarts.
* **Still pending (in progress):** the live-cohort restore + history backfill takes ~12h; once the first pruner cycle fires I'll parse a fresh WAL to confirm the deduped pattern on-disk (expect ≤1 `object_version_by_checkpoint` DeleteRange per object per prune batch, vs. K up to 28,365 before) and update this PR with the numbers.

## Follow-ups (not in this PR)

* Consider setting `memtable_max_range_deletions` on `object_version_by_checkpoint` as defense-in-depth (forces a flush before a memtable can accumulate a pathological tombstone count). Not exposed by the `rocksdb` 0.22 Rust crate yet.


nickvikeras:
Verification data is in — closing the loop on the "still pending" item.

**Status of this PR's commit:** the dedup fix was taken into the base branch via rebase as `36d9939b8c`, with a follow-up on top (`61c056d9` "point-delete object_version_by_checkpoint retractions") that eliminates CF11 range tombstones entirely — the per-batch dedup here bounded K within a batch but not across batches, so point deletes are the right endgame; the `Retractions` collector is retained to coalesce each object to one prefix walk per batch. `mainnet-rpc-alpha` was redeployed on `61c056d9` (which contains both).

**On-disk confirmation** (WAL `031312.log`, written 2026-07-05 19:11 UTC, 18,860 batches, parsed with the same WriteBatch parser used in the root-cause analysis):

| metric | before (`128213.log`, Jul 2) | after (`031312.log`, Jul 5) |
|---|---|---|
| `object_version_by_checkpoint` DeleteRanges | 457,494 (max K = 28,365 for one object) | **0** |
| ΣK²/2 (fragment,seq) pairs | 1.07×10⁹ | **0** |
| `object_version_by_checkpoint` point deletes | 0 | 101,842 across 8,534 ids |

**Node health** (vs. the incident baseline of OOMKilled every ~8 min at 120G, 132 restarts, dead inside `Db::open`):

- Ready and serving, executing at tip (checkpoint ~295.05M)
- History backfill + pruning active (pruner running the point-delete path)
- RSS 25.4GB, HWM 32GB against the 120G limit; 0 restarts over ~18h on the current pod
- rpc_store fully rebuilt (109G) from the fresh restore

The quadratic tombstone-fragmentation blowup is gone in both its manifestations (steady-state flush and WAL-recovery crash loop).

PR #27147, opened 2026-07-06
Reserve a priority lane in consensus input for system transactions

alex-mysten:
## Description

Follow-up to the consensus-adapter change that let system messages bypass the submit semaphore: consensus's single input channel is the remaining place a validator's own system messages (checkpoint signatures, `EndOfPublish`, randomness DKG, …) can queue behind a user backlog under load.

Adds a reserved, drained-first priority lane to the consensus input and routes system messages to it. Consensus stays agnostic to transaction type — the submitter picks a `Priority`; Sui maps a lone system message to `High`.

## Test plan

New consensus-core unit tests: priority-first ordering, reserved capacity (a full normal lane blocks a normal send but not a high one), and count/byte overflow held and drained first.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/transaction.rs
+++ b/consensus/core/src/transaction.rs
@@ -297,6 +297,7 @@ pub enum ClientError {
 }
 
 impl TransactionClient {
+    /// Returns the client and the receivers for normal and priority transactions, in that order.
     pub(crate) fn new(
         context: Arc<Context>,
     ) -> (
@@ -311,6 +312,7 @@ impl TransactionClient {
         )
     }
 
+    /// Returns the client and the receivers for normal and priority transactions, in that order.
     fn new_with_max_pending_transactions(
         context: Arc<Context>,
         max_pending_transactions: usize,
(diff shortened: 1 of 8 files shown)


consensus/core/src/transaction.rs
```diff
@@ -269,18 +297,38 @@ pub enum ClientError {
 }
 
 impl TransactionClient {
-    pub(crate) fn new(context: Arc<Context>) -> (Self, Receiver<TransactionsGuard>) {
-        Self::new_with_max_pending_transactions(context, MAX_PENDING_TRANSACTIONS)
+    pub(crate) fn new(
```
mwtian, line 301: Add a comment on the order of returned receivers (priority vs normal), for new() and new_with_max_pending_transactions().

mwtian (approved):

PR #27149, opened 2026-07-06
Remove admission queue bypass mode

alex-mysten:
## Description

Follow-up to the `sui_mainnet_high_traffic` incidents ([CORE-198](https://linear.app/mysten-labs/issue/CORE-198/refactor-admission-queue)). Bypass let a transaction skip the gas-price priority queue whenever consensus inflight was below a threshold, so admission prioritization and load-shedding were bypassed for effectively all steady-state traffic — and during the transient where inflight dipped, a low-gas transaction could jump straight to consensus ahead of higher-gas entries still waiting in the queue. Remove bypass entirely so all user transactions are admitted through the queue. The only remaining direct-to-consensus paths are: queue turned off by config, failover (queue actor stalled), and ping requests.

## Test plan

Covered by the existing admission queue e2e tests in `sui-e2e-tests/tests/admission_queue_tests.rs` (admission, eviction, rejection, reconfig), which previously forced queue mode via `bypass_fraction: 0.0` and now exercise the only behavior.

---

## Release notes

- [x] Nodes (Validators and Full nodes): Removes the `admission-queue-bypass-fraction` validator config field; all transactions are now admitted through the gas-price priority queue. The removed field is ignored if still present in a config file, but should be deleted.

### DIFF
--- a/crates/sui-config/src/node.rs
+++ b/crates/sui-config/src/node.rs
@@ -1446,12 +1446,6 @@ pub struct AuthorityOverloadConfig {
     #[serde(default = "default_admission_queue_capacity_fraction")]
     pub admission_queue_capacity_fraction: f64,
 
-    // Fraction of max_pending_transactions below which the admission queue is
-    // bypassed (transactions are submitted directly to consensus). Above this
-    // threshold, transactions go through the priority queue.
-    #[serde(default = "default_admission_queue_bypass_fraction")]
-    pub admission_queue_bypass_fraction: f64,
-
     // Enables use of a gas-price-based priority queue for load shedding of
     // transactions at admission time. If false, when consensus is saturated, transactions
     // are rejected with TooManyTransactionsPendingConsensus.
@@ -1510,10 +1504,6 @@ fn default_admission_queue_capacity_fraction() -> f64 {
     0.5
 }
 
-fn default_admission_queue_bypass_fraction() -> f64 {
-    0.9
-}
-
 fn default_admission_queue_enabled() -> bool {
     true
 }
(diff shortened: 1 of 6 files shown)


crates/sui-config/src/node.rs
```diff
@@ -1510,16 +1504,12 @@ fn default_admission_queue_capacity_fraction() -> f64 {
...
-    0.9
-}
-
 fn default_admission_queue_enabled() -> bool {
     true
 }
 
 fn default_admission_queue_failover_timeout() -> Duration {
-    Duration::from_secs(30)
+    Duration::from_secs(10)
```
akichidis: My understanding is that the queue will get tripped now quicker in case of a stall right? For example, if consensus deals with some temporary regression and queue doesn't move within 10 seconds, then it's possible to start seeing low value transactions coming through as they'll bypass it? Isn't this a point where we actually need to be more aggressive on prioritising high value transactions ?

akichidis (approved):
Changes make sense. Left a comment regarding the choice of 10 seconds timeout. It might not offer as is - to my understanding - the protection we need during a regression.

crates/sui-config/src/node.rs
alex-mysten: Correct. If we don't drain a tx in 10s, we would fail over to direct submission. I lowered this as a safeguard in case we don't entirely trust the queue implementation, given that our risk of a stall now impacts 100% of transactions. Now I'm thinking we leave failover at 30s for the reasons you've stated.

https://github.com/MystenLabs/sui/pull/27202 changes failover detection logic to "if consensus has capacity, AND we haven't drained in a time period" which perhaps is a little bit more indicative of a real halt with the queue implementation. It could be argued at that point that failover adds complexity during an incident without buying us meaningful safety.

alex-mysten: Failover timeout raised back to 30s (the current value)

PR #27151, opened 2026-07-06
[gql] Transaction connections understand the legacy `JsonCursor<u64>`

wlmyng:
## Description 

After the migration to the new cursor format, we can still support the previous `JsonCursor<u64>` format. The connection will always return a cursor in the new format.

## Test plan 

New integration and e2e tests
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


nickvikeras (approved):

amnn (changes requested):
Wait one on this -- there is a better way to tackle this that I will build on top of #27163

wlmyng:
> Wait one on this -- there is a better way to tackle this that I will build on top of #27163

Roger that

PR #27153, opened 2026-07-06
rust-sdk: bump pointer and migrate v2alpha ledger proto changes

bmwill:
Update the `sui-rust-sdk` git pointers and adapt to the four v2alpha ledger service proto changes that came with them:

- `EventTypeFilter.type` was renamed to `event_type`.
- `QueryOptions.limit_items` became `limit`, and `ordering` became optional. `QueryEnd.reason` became optional; consumers now read it through the generated `reason()` accessor, which defaults an absent or unknown value to `QUERY_END_REASON_UNSPECIFIED`.
- `Watermark.checkpoint_hi` and `checkpoint_lo` were merged into a single `checkpoint` field carrying the direction-correct completion boundary.
- `TransactionItem.transaction_offset` was removed in favor of `ExecutedTransaction.transaction_index`, and `EventItem` was slimmed to `{event, watermark}` with its ledger position (`checkpoint`, `transaction_digest`, `transaction_index`, `event_index`) moved onto the embedded `Event`.

Both list backends (`sui-rpc-api` and `sui-kv-rpc`) now populate the relocated position fields on the message unconditionally, since the list endpoints always report where an item sits regardless of the read mask, which the authenticated-events light client and the e2e tests rely on. The light client, the alpha ledger gRPC reader, and the affected tests were updated to read from the new field locations.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
+++ b/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
@@ -51,7 +51,7 @@ pub struct StreamPage<T> {
 enum FrameKind<T> {
     Item { payload: T, cursor: Bytes },
     Watermark { cursor: Option<Bytes> },
-    End { reason: i32 },
+    End { reason: proto::QueryEndReason },
     Unknown,
 }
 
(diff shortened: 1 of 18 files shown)


crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
```diff
@@ -211,7 +208,9 @@ impl TryFrom<proto::ListTransactionsResponse> for FrameKind<ExecutedTransaction>
             Response::Watermark(watermark) => FrameKind::Watermark {
                 cursor: watermark.cursor,
             },
-            Response::End(end) => FrameKind::End { reason: end.reason },
+            Response::End(end) => FrameKind::End {
+                reason: end.reason(),
+            },
```
wlmyng, line 213: I can't seem to find this on `sui.rpc.v2alpha.rs`, is this some typed accessor method that already handles "unexpected" numbers?

bmwill: It's generated and it should show up if you look at the docs. But yes it returns the default value if it's not set

amnn (approved):

PR #27154, opened 2026-07-06
[gql] reject invalid txn cursor upfront

wlmyng:
## Description 

```
impl TxBoundsCursor for CTransaction {
    fn tx_sequence_number(&self) -> u64 {
        CursorToken::decode(self)
            .expect("cursor already validated as ByteCursor")
            .position
    }
}
```

The expect from above pops up when we receive a legacy cursor format, which is valid base64 but doesn't decode into a `CursorToken`. Before this, we should validate in `fn paginate` whether the cursor is indeed an encoded `CursorToken`.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-bridge/src/e2e_tests/basic.rs
+++ b/crates/sui-bridge/src/e2e_tests/basic.rs
@@ -61,7 +61,7 @@ async fn test_bridge_from_eth_to_sui_to_eth() {
         .new_bridge_events(HashSet::from_iter([
             TokenTransferApproved.get().unwrap().clone(),
             TokenTransferClaimed.get().unwrap().clone(),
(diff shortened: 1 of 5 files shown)


evan-wall-mysten (approved):

crates/sui-bridge/src/e2e_tests/basic.rs
amnn, line 1: Hmmm... why?

PR #27155, opened 2026-07-06
[sui][sui-verifier] reject `&mut TxContext -> &mut T` on system packages

cgswords:
## Description 

Ensure `&mut TxContext -> &mut T` is not a valid function signature, and reject them at publication time if they exist.

## Test plan 

New and updated tests demonstrate the feature. @tzakian will backtest before landing, too

---

## Release notes

- [x] Protocol: Bumps the max protocol version to 131 and adds the `framework_tx_context_mut_restrictions` feature flag. Starting at version 131, system-package functions that take `&mut TxContext` and return a mutable reference must also take a non-`TxContext` `&mut` parameter; publication is rejected otherwise. User packages are unaffected.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


sui-execution/latest/sui-verifier/src/tx_context_restrictions_verifier.rs
```diff
@@ -0,0 +1,90 @@
...
+        })?;
+    }
+    Ok(())
+}
+
+fn verify_function(module: &CompiledModule, fdef: &FunctionDefinition) -> Result<(), String> {
+    let fhandle = module.function_handle_at(fdef.function);
+    let returns = &module.signature_at(fhandle.return_).0;
+    let returns_non_tx_context_mut = returns.iter().any(|t| match t {
+        SignatureToken::MutableReference(inner) => !is_tx_context(module, inner),
```
tnowacki: Why do we need this exception?

sui-execution/latest/sui-verifier/src/tx_context_restrictions_verifier.rs
```diff
@@ -0,0 +1,90 @@
...
+use sui_types::{
+    SUI_FRAMEWORK_ADDRESS,
+    base_types::{TX_CONTEXT_MODULE_NAME, TX_CONTEXT_STRUCT_NAME},
+    error::ExecutionError,
+};
+
+use crate::verification_failure;
+
+pub fn verify_module(module: &CompiledModule) -> Result<(), ExecutionError> {
+    for func_def in &module.function_defs {
```
tnowacki, line 52: I had maybe thought we should do this only for framework modules. Just as a checksum on our own implementations. I think adding this generally will be annoying... maybe?

sui-execution/latest/sui-verifier/src/tx_context_restrictions_verifier.rs
```diff
@@ -0,0 +1,90 @@
...
+        verify_function(module, func_def).map_err(|error| {
+            let name = module.identifier_at(module.function_handle_at(func_def.function).name);
+            verification_failure(format!("{}::{}. {}", module.self_id(), name, error))
+        })?;
+    }
+    Ok(())
+}
+
+fn verify_function(module: &CompiledModule, fdef: &FunctionDefinition) -> Result<(), String> {
```
tnowacki: Can probably do 
```suggestion
fn verify_function(module: &CompiledModule, fdef: &FunctionDefinition) -> Result<(), &'static> {
```
if we really want to

sui-execution/latest/sui-verifier/src/tx_context_restrictions_verifier.rs
tnowacki: And we will need to protocol gate this

cgswords: Protocol gated, let me know about the framework modules thing. It's easy to add.

sui-execution/latest/sui-verifier/src/tx_context_restrictions_verifier.rs
cgswords: It was an attempt to be more-relaxed, but is not necessary. Removed.

sui-execution/latest/sui-verifier/src/tx_context_restrictions_verifier.rs
tnowacki: I think it should be a framework only thing. Since we can't avoid this problem generally because of generics. 
As such for the framework, we might want to even extend it to doing `<T>(x: &mut T) -> &mut u` style patterns, if we are being paranoid

cgswords: It is framework-only, but I think in the framework case that reflexive one is likely uninhabitable unless T = U? Let me know if you want that more-general restriction.

tnowacki (approved):

crates/sui-protocol-config/src/lib.rs
```diff
@@ -1009,6 +997,15 @@ struct FeatureFlags {
...
     allow_references_in_ptbs: bool,
 
+    // If true, the tx_context_restrictions_verifier pass runs at Move module
+    // publish time and rejects system-package signatures with `&mut TxContext`
+    // + any `&mut _` return that have no non-`TxContext` `&mut U` parameter.
+    // User packages are exempt: they can express the same shape through
+    // generic instantiation, so PTB arity and auto-injection checks are the
+    // safety mechanism there.
+    #[serde(skip_serializing_if = "is_false")]
+    check_tx_context_restrictions: bool,
```
tnowacki: Nit, something more declarative than `check_` might be nice.
Maybe `framework_tx_context_mut_restrictions`? or something similar

cgswords: Renamed to `framework_tx_context_mut_restrictions`. Also rebased on main: the flag moved to protocol version 131 (130 was taken), and the getter/test-setter are now macro-generated, so the tests use the new generic `--enable-feature-flags` mechanism instead of a bespoke CLI flag.

PR #27156, opened 2026-07-07
[move-ide] Follow Move conventions for auto-imports and quick-fixes

awelc:
## Description 

This PR implements a refinement to auto-imports and quick-fix algorithms to follow Move coding conventions better. The description below is for auto-completion but quick-fix update follows the same strategy.

When suggesting auto-completion for single words, we used to import package/module and leave the single word in the code regardless of the type. For example, to auto-complete on letter `f` towards a function `foo` in module `mod` in package `pkg`, we would insert the following:
```
use pkg::mod::foo;

...
foo()
...
```

This is technically correct and follows Move convention for structs and enums but not for functions. For functions, it is customary to prefix their name with module name. As a result, a better auto-import (or quick-fix) experience is to insert the following:
```
use pkg::mod;

...
mod::foo()
...
```

Of course we should not insert module import more than once. Instead we should check if it's already in scope. This was already easily done for auto-imports but for quick-fixes we had to make some additional changes to make this information available

## Test plan 

New tests have been added. All tests must pass. Also tested manually

### DIFF
--- a/external-crates/move/crates/move-analyzer/editors/code/package.json
+++ b/external-crates/move/crates/move-analyzer/editors/code/package.json
@@ -5,8 +5,7 @@
   "publisher": "mysten",
   "icon": "images/move.png",
   "license": "Apache-2.0",
-  "version": "1.0.44",
-  "preview": true,
+  "version": "2.0.0",
   "repository": {
     "url": "https://github.com/MystenLabs/sui.git",
     "type": "git"
--- a/external-crates/move/crates/move-analyzer/src/code_action.rs
+++ b/external-crates/move/crates/move-analyzer/src/code_action.rs
@@ -5,6 +5,7 @@ use crate::{
     completions::utils::{
         addr_to_ide_string, all_mod_enums_to_import, all_mod_functions_to_import,
         all_mod_structs_to_import, auto_import_text_edit, compute_cursor, import_insertion_info,
+        module_import_info,
     },
     context::Context,
     symbols::{
@@ -36,7 +37,7 @@ use move_compiler::{
     expansion::ast::ModuleIdent,
     linters::LintLevel,
     parser::ast::{LeadingNameAccess_, NameAccessChain_},
-    shared::{Identifier, Name},
+    shared::{Identifier, Name, ide::AliasAutocompleteInfo},
 };
 use move_package_alt::MoveFlavor;
 
@@ -149,7 +150,7 @@ fn access_chain_autofix_actions<F: MoveFlavor>(
         LintLevel::None,
         move_flavor,
         flavor,
-        None,
+        Some(&file_path),
     ) else {
         return code_actions;
     };
@@ -192,6 +193,9 @@ pub fn access_chain_autofix_actions_for_error(
     {
         return;
     }
+    if let Some(autocomplete_info) = compiled_pkg_info.compiler_autocomplete_info.clone() {
+        symbols.compiler_autocomplete_info = autocomplete_info;
+    }
     // compute cursor and update symbols with it
     let Some(file_path) = canonical_path_from_uri(&file_url) else {
         return;
@@ -231,7 +235,7 @@ pub fn access_chain_autofix_actions_for_error(
                 }
                 SingleElementChainDiagPrefix::UnboundFunction => {
                     if err_msg.starts_with(prefix.as_str()) {
-                        single_element_access_chain_autofixes(
+                        single_element_function_autofixes(
                             code_actions,
                             symbols,
                             cursor,
@@ -311,7 +315,7 @@ fn single_element_access_chain_autofixes<I, K>(
             let title = format!("Qualify as `{}{}`", autofix_prefix, unbound_name);
             code_actions.push(access_chain_code_action(
                 title,
-                text_edit,
+                vec![text_edit],
                 diag.clone(),
                 file_url.clone(),
             ));
@@ -321,7 +325,7 @@ fn single_element_access_chain_autofixes<I, K>(
                 let text_edit = auto_import_text_edit(import_text, import_insertion_info);
                 code_actions.push(access_chain_code_action(
                     title,
-                    text_edit,
+                    vec![text_edit],
                     diag.clone(),
                     file_url.clone(),
                 ));
@@ -345,7 +349,7 @@ fn single_element_access_chain_autofixes<I, K>(
                 let title = format!("Qualify as `{}{}`", autofix_prefix, unbound_name);
                 code_actions.push(access_chain_code_action(
                     title,
-                    text_edit,
+                    vec![text_edit],
                     diag.clone(),
                     file_url.clone(),
                 ));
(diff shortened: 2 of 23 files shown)


external-crates/move/crates/move-analyzer/src/code_action.rs
```diff
@@ -149,7 +150,7 @@ fn access_chain_autofix_actions<F: MoveFlavor>(
         LintLevel::None,
         move_flavor,
         flavor,
-        None,
+        Some(&file_path),
```
awelc, line 153: This is one of the changes that looks a bit random but was needed to give quick-fix algorithm insight into already imported modules (similarly to how auto-import algorithm obtains them)

external-crates/move/crates/move-analyzer/src/code_action.rs
```diff
@@ -355,7 +359,111 @@ fn single_element_access_chain_autofixes<I, K>(
...
+    let Some(unbound_name_lsp_pos) =
+        loc_start_to_lsp_position_opt(&symbols.files, &unbound_name.loc)
+    else {
+        return;
+    };
+    let info = symbols
+        .compiler_autocomplete_info
+        .get_path_autocomplete_info(&unbound_name.loc)
+        .cloned()
+        .unwrap_or_else(AliasAutocompleteInfo::new);
```
cgswords: Is the path autocomplete info guaranteed to be recorded at exactly `unbound_name.loc` for a single-element chain? If not (or if we can't be sure), should we skip the module-import fix instead of defaulting to an empty `AliasAutocompleteInfo`?

external-crates/move/crates/move-analyzer/src/completions/utils.rs
```diff
@@ -94,6 +102,36 @@ pub fn auto_import_text_edit(
...
+        if in_scope_mod_ident.value == mod_ident.value {
+            return Some(ModuleImportInfo {
+                module_prefix: alias.to_string(),
+                import_text: None,
+            });
+        }
+    }
+
+    let module_name = mod_ident.value.module.value();
+    if in_scope_modules.contains_key(&module_name) {
```
cgswords: What happens if the target module's name is taken by a *member* alias rather than a module alias? Do we need to scan `info.members` aliases too?

external-crates/move/crates/move-analyzer/src/code_action.rs
cgswords, line 153: Have you measured latency on a large package? Since the symbolicator already compiled with a cursor for this file on the last edit, is there a way to reuse that instead of forcing incremental recompilation here?

external-crates/move/crates/move-analyzer/src/completions/name_chain.rs
```diff
@@ -673,8 +675,30 @@ fn all_single_name_member_completions(
...
+        for (mod_ident, import_functions) in all_mod_functions_to_import(symbols, cursor) {
+            let Some(mod_defs) = mod_defs(symbols, &mod_ident.value) else {
+                continue;
+            };
+            for function_name in import_functions {
+                // exclude functions that are already imported
+                if members_info.contains_key(&(mod_ident, function_name)) {
+                    continue;
+                }
+                let function_imports = function_auto_imports_with_module_import(
```
cgswords, line 710: We compute the same `module_import_info` for every function in a module, but it only depends on the module name. Could we hoist it out of the per-function loop and compute it once per module instead?

external-crates/move/crates/move-analyzer/src/code_action.rs
```diff
@@ -192,6 +193,9 @@ pub fn access_chain_autofix_actions_for_error(
     {
         return;
     }
+    if let Some(autocomplete_info) = compiled_pkg_info.compiler_autocomplete_info.clone() {
```
cgswords, line 196: We run `access_chain_autofix_actions_for_error` once per diagnostic, and it clones the whole `compiler_autocomplete_info` map each time. Could we clone it once or RC it instead?

external-crates/move/crates/move-analyzer/tests/ide_testsuite.rs
```diff
@@ -721,6 +730,32 @@ fn test_symbols_for_autocomplete<F: MoveFlavor + Default>(
     Ok(symbols)
 }
 
+/// Compute symbols for code-action tests with autocomplete information for the target file.
+/// The code-action test computes the exact cursor position separately for each diagnostic.
+fn test_symbols_for_code_action<F: MoveFlavor + Default>(
```
cgswords: Could we parameterize `test_symbols_for_autocomplete` to handle the code-action case too, instead of duplicating the helper?

external-crates/move/crates/move-analyzer/tests/ide_testsuite.rs
```diff
@@ -1010,6 +1037,16 @@ fn access_chain_quick_fix_test_suite<F: MoveFlavor + Default>(
         fpath.push(format!("sources/{file}"));
         let cpath = canonicalize_path(fpath.clone());
 
+        // Compile per file to get autocomplete/alias info for that file. The exact cursor position
+        // is computed later for each diagnostic triggerring a quick fix action, so it does not
```
cgswords: Nit: "triggerring" → "triggering".

external-crates/move/crates/move-analyzer/editors/code/package.json
```diff
@@ -5,8 +5,7 @@
   "publisher": "mysten",
   "icon": "images/move.png",
   "license": "Apache-2.0",
-  "version": "1.0.44",
-  "preview": true,
+  "version": "2.0.0",
```
cgswords, line 8: Intentional to cut 2.0.0 / drop `"preview": true` here?

external-crates/move/crates/move-analyzer/tests/autofix/sources/function_import_convention.move
cgswords, line 1: Maybe two more tests: (1) a missing `AliasAutocompleteInfo` to see if we emit a duplicate `use`, and (2) a member alias that collides with the module name, to see if we avoid the collision.

external-crates/move/crates/move-analyzer/editors/code/package.json
awelc: Intentional. This is the last bit on the list of tasks (as per our internal tracking) to be completed before taking the VSCode extension out of preview. From now on the intention is to be more reactive and predominantly work on bug fixes (if any problems pop up in the future) and/or requested features (when deemed to make sense, particularly in this agentic era)

external-crates/move/crates/move-analyzer/src/code_action.rs
awelc: For large packages, incremental re-compilation can be on the order of 0.6-0.7s

More importantly, though, we already do this on every auto-completion request and these are much more frequent than quick-fix requests, so this should not be a noticeable overhead for the user.

Still, we may want to revisit it at some point, but in a separate PR?

external-crates/move/crates/move-analyzer/src/code_action.rs
awelc: > Is the path autocomplete info guaranteed to be recorded at exactly unbound_name.loc for a single-element chain?

As far as I can tell this holds but it's a good idea to be defensive here no matter what. I changed it to only offer fully qualified import in case auto-complete info is unavailable to avoid potentially inserting an incorrect import

external-crates/move/crates/move-analyzer/src/completions/utils.rs
awelc: Good catch! I clearly did not think about it... Fixed now here and in other relevant places with tests for all these situations added.

external-crates/move/crates/move-analyzer/src/completions/name_chain.rs
awelc: Good point! Done!

external-crates/move/crates/move-analyzer/src/code_action.rs
awelc: Arc-ed the field which was probably the smallest and cleanest change

external-crates/move/crates/move-analyzer/tests/ide_testsuite.rs
awelc: Good catch. Dedupplicated!

external-crates/move/crates/move-analyzer/tests/ide_testsuite.rs
awelc: Fixed!

external-crates/move/crates/move-analyzer/tests/autofix/sources/function_import_convention.move
awelc: Added the second one but could not figure out how to force a situation that would trigger the first one without "faking" it in the test harness which kind of defeats the purpose I'd think

cgswords (approved):

PR #27158, opened 2026-07-07
Support setting boolean feature flags by name in transactional tests

lxfind:
## Description

Enabling a boolean feature flag in a transactional test currently requires adding a dedicated init arg to the test runner and a hand-written `set_*_for_testing` method on `ProtocolConfig`, per flag. Stacked on #26976, this generates a string-based setter from the feature-flags derive (`ProtocolConfig::set_feature_flag_for_testing(name, bool)`, mirroring the existing scalar `set_attr_for_testing`) and exposes it in the test runner as generic `--enable-feature-flags` / `--disable-feature-flags` init args, so new flags are settable from .move tests with no runner changes.

Existing bespoke init args are migrated:
- Dropped where the flags are already on by default at the protocol versions the tests run (verified against config snapshots): `--enable-accumulators`, `--enable-authenticated-event-streams`, `--enable-coin-reservations`, `--enable-address-balance-gas-payments`. Three tests pinning `--protocol-version 108` solely because it was current when they were written are unpinned to make the last of these droppable.
- Replaced with the generic args where not default: `--shared-object-deletion`, `--allow-references-in-ptbs`, `--enable-non-exclusive-write-objects`.
- Kept: `--enable-gasless`, which sets scalar limits (e.g. `gasless_max_tps`) not expressible as bool flags.

## Test plan

New unit tests cover setting a flag by name (round-trip true/false) and panicking on an unknown flag name. The migration is validated by the full `sui-adapter-transactional-tests` suite and the migrated `sui-indexer-alt-e2e-tests` graphql tests passing with zero expectation/snapshot changes, confirming the dropped args and unpinned protocol versions are behavior-preserving, including the shared-object-deletion tests still pinning version 90.

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes are not required.

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/transaction_effects/address_balance_changes.move
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/transaction_effects/address_balance_changes.move
@@ -3,7 +3,7 @@
 
 // Test send_funds and redeem_funds from sui::balance
 
-//# init --protocol-version 108 --addresses test=0x0 --accounts A B C --enable-accumulators --simulator --enable-address-balance-gas-payments
+//# init --protocol-version 128 --addresses test=0x0 --accounts A B C --simulator
 
 // Send 1000000000 from A to B and A to C
 //# programmable --sender A --inputs 1000000000 @B @C
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql/transaction_effects/address_balance_gas_effects.move
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql/transaction_effects/address_balance_gas_effects.move
@@ -1,7 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//# init --protocol-version 108 --accounts A --addresses test=0x0 --enable-accumulators --simulator --enable-address-balance-gas-payments
+//# init --protocol-version 128 --accounts A --addresses test=0x0 --simulator
 
 //# programmable --sender A --inputs 1000000000 @A
 //> 0: SplitCoins(Gas, [Input(0)]);
--- a/crates/sui-transactional-test-runner/src/args.rs
+++ b/crates/sui-transactional-test-runner/src/args.rs
@@ -6,7 +6,7 @@ use std::path::PathBuf;
 use crate::test_adapter::{FakeID, SuiTestAdapter};
 use anyhow::{bail, ensure};
 use clap;
-use clap::{Args, Parser};
+use clap::{ArgAction, Args, Parser};
 use move_compiler::editions::Flavor;
 use move_core_types::parsing::{
     parser::Parser as MoveCLParser,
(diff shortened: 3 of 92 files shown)


crates/sui-transactional-test-runner/src/args.rs
```diff
@@ -119,6 +99,14 @@ pub struct SuiInitArgs {
...
     #[clap(long = "gasless-max-unused-inputs")]
     pub gasless_max_unused_inputs: Option<u64>,
+    /// Enable boolean protocol feature flags by name (comma separated), e.g.
+    /// `--enable-feature-flags zklogin_auth,enable_party_transfer`. Flag names match the
+    /// field names of `FeatureFlags` in the sui-protocol-config crate.
+    #[clap(long = "enable-feature-flags", value_delimiter = ',')]
+    pub enable_feature_flags: Vec<String>,
+    /// Disable boolean protocol feature flags by name (comma separated).
+    #[clap(long = "disable-feature-flags", value_delimiter = ',')]
+    pub disable_feature_flags: Vec<String>,
```
amnn: nit: I tend to prefer the form where you provide the flag multiple times because it's a bit more forgiving for parsing the value, but that's personal preference.


```suggestion
    #[clap(long = "enable-feature-flags", action = ArgAction::Append)]
    pub enable_feature_flag: Vec<String>,
    /// Disable boolean protocol feature flags by name (comma separated).
    #[clap(long = "disable-feature-flags", action = ArgAction::Append)]
    pub disable_feature_flag: Vec<String>,
```

crates/sui-indexer-alt-e2e-tests/tests/graphql/transaction_effects/address_balance_gas_effects.move
amnn, line 1: Note that you will need to set a protocol version here, but it can be one where these features are enabled by default now. If we don't do this, the snapshots will need to be regenerated every time the framework is updated.

crates/sui-indexer-alt-e2e-tests/tests/graphql/transaction_effects/address_balance_changes.move
amnn, line 1: ditto -- need to pick a protocol version, but it can be a later one.

amnn (approved):
Nice, thank you @lxfind -- for the GraphQL E2E tests, we still need to pick a fixed protocol version, but otherwise this looks good to me. Accepting to unblock.

crates/sui-transactional-test-runner/src/args.rs
lxfind: Done — switched to `ArgAction::Append` (repeated flags), and updated the one test that used the comma form.

crates/sui-indexer-alt-e2e-tests/tests/graphql/transaction_effects/address_balance_gas_effects.move
lxfind: Done — pinned both tests to `--protocol-version 128`, where these features are all enabled by default (so the bespoke arg is still unnecessary). Snapshots unchanged.

crates/sui-indexer-alt-e2e-tests/tests/graphql/transaction_effects/address_balance_changes.move
lxfind: Done — pinned to `--protocol-version 128`.

PR #27163, opened 2026-07-07
graphql: OpaqueCursor<C>

amnn:
## Description

Introduce a cursor type that supports custom byte encodings, but performs encoding/decoding as part of parsing, so that the rest of the system can trust that the cursor is valid.

## Test plan

Existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/scalars/cursor.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/scalars/cursor.rs
@@ -9,10 +9,20 @@ use async_graphql::Scalar;
 use async_graphql::ScalarType;
 use async_graphql::Value;
 use async_graphql::connection::CursorType;
+use bytes::Bytes;
 use fastcrypto::encoding::Base64;
 use fastcrypto::encoding::Encoding;
 use serde::Serialize;
 use serde::de::DeserializeOwned;
+
+/// Custom byte encoding for cursors.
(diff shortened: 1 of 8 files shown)


crates/sui-indexer-alt-graphql/src/api/scalars/cursor.rs
```diff
@@ -45,6 +52,9 @@ pub enum Error {
 
     #[error("Invalid JSON")]
     BadJson,
+
+    #[error("Invalid encoding: {0:#}")]
+    BadEncoding(#[from] anyhow::Error),
```
wlmyng, line 57: it seems this error variant is not actually used anywhere?

Edit:

  - ByteCursor::decode_cursor returns anyhow::Result<Self> (cursor.rs:22-24).
  - OpaqueCursor::decode_cursor returns Result<Self, Error> and does Ok(OpaqueCursor(C::decode_cursor(&bytes)?)).
  - That ? converts the `anyhow::Error` into `Error` matching the `#[from]` on `BadEncoding`

wlmyng (approved):
lgtm

PR #27174, opened 2026-07-07
Add scheduled workflow to auto-fix cargo-deny advisories

ebmifa:
## Description

`cargo-deny (advisories)` failures on main are a recurring manual chore (#27167, #27137, #26968, #26948) and the fix is usually a mechanical, semver-compatible lockfile bump. Add a daily scheduled workflow that detects advisory errors in the CI-gated workspaces (root and `external-crates/move`), fixes them with `cargo update -p <pkg>@<version>`, propagates the bumps to the `examples/rust/*` lockfiles, and opens a PR. Advisories that can't be fixed by a compatible bump (need a major-version upgrade or a `deny.toml` ignore) are reported on a GitHub issue instead and the run is marked failed.

## Test plan

- `shellcheck` and `actionlint` pass on the new script and workflow.
- Ran `scripts/cargo-deny-autofix.sh` against a scratch crate pinned to `crossbeam-epoch 0.9.18` (RUSTSEC-2026-0204): it bumps to 0.9.20 — reproducing #27167 — verifies the check passes, and emits `has_fixes=true` with the fixed-advisory markdown used for the PR body.
- Ran it against a scratch crate pinned to `time 0.1` (RUSTSEC-2020-0071, no semver-compatible patch): the advisory lands in `remaining.md` with `has_remaining=true`, which drives the issue-reporting path.
- Ran it against the real root and `external-crates/move` workspaces (currently clean): no-op, no lockfile changes.


stefan-mysten (approved):
Let's give it a try and see how it works. Not sure I fully follow the logic for fixing the deny report.

Thanks @ebmifa

PR #27176, opened 2026-07-07
indexer-alt-graphql: configurable per-pipeline availability

evan-wall-mysten:
## Description

Stacked on #27341 and #27342.

Today `checkpoint_viewed_at` is the MIN high-watermark across every tracked pipeline, so a single lagging pipeline pins the whole GraphQL server to its stale checkpoint. This lets an operator gate pipelines by availability policy: `max-checkpoint-lag` serves a pipeline's data only while it is within that many checkpoints of the network tip, `enabled = false` never serves it (kill-switch), and `enabled = true` always serves it. A gated pipeline is dropped from the consistency boundary so the caught-up pipelines keep serving fresh reads. It's the read-side companion to ingestion cohorts.

Gating a pipeline excludes it from the `checkpoint_viewed_at` MIN and adds `FeatureUnavailable` guards on the read paths that bypass `reader_lo` (`consistent`, `obj_versions`, `tx_balance_changes`, and the `kv_*` content pipelines) so a gated pipeline can never silently serve stale or out-of-range data. Pipelines with no policy are always served, so behavior is unchanged unless configured. The network tip is approximated as the MAX high-watermark across tracked pipelines (the true tip when a Bigtable/Ledger-gRPC KV source is configured).

Availability policy lives in the same `[pipeline-defaults]`/`[pipeline.<name>]` config surface #27341/#27342 use for pipeline enablement, as a nested `availability` section; a policy sets exactly one of `enabled` or `max-checkpoint-lag`:

```toml
[pipeline-defaults.availability]
max-checkpoint-lag = 100          # default for every tracked pipeline

[pipeline.kv_objects.availability]
enabled = true                    # always serve, exempt from the default

[pipeline.tx_kinds.availability]
enabled = false                   # never serve

[pipeline.tx_calls.availability]
max-checkpoint-lag = 1000         # lag-gated override
```

## Test plan

New unit tests cover `is_available` (lag boundaries inclusive/exclusive, `enabled`/`disabled` ignoring lag), `apply_availability` (advancing the boundary past a lagging pipeline, the furthest-ahead pipeline never lag-gated, `disabled` gating even the tip, an `enabled` override exempting a pipeline from the default, an all-disabled snapshot staying initialized), TOML parse/round-trip of every policy form (including mutual-exclusion and empty-section rejection), the `reader_lo` gate, and `require_pipeline` (present-and-gated fires, untracked is a no-op). Covered by existing tests otherwise.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Operators can now gate pipelines in the RPC config with availability policies: `[pipeline-defaults.availability]` sets a default for every tracked pipeline and `[pipeline.<name>.availability]` overrides it per pipeline. A policy is either `max-checkpoint-lag = N` (serve only while within N checkpoints of the network tip), `enabled = false` (never serve), or `enabled = true` (always serve). A gated pipeline no longer holds back the server's consistency boundary, and queries requiring it return `FeatureUnavailable`. Pipelines with no policy are always served, so no action is needed unless you opt in.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/available_range.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/available_range.rs
@@ -103,6 +103,7 @@ impl AvailableRangeKey {
             let watermark = watermarks
                 .per_pipeline()
                 .get(pipeline)
+                .filter(|p| p.available())
                 .ok_or_else(|| pipeline_unavailable(pipeline))?;
             Ok(acc.max(watermark.lo().checkpoint()))
         })
@@ -143,6 +144,9 @@ pub(crate) fn pipeline_unavailable(pipeline: &str) -> RpcError {
         "cp_sequence_numbers" => feature_unavailable("querying checkpoints"),
         "ev_emit_mod" => feature_unavailable("querying events by emitting module"),
         "ev_struct_inst" => feature_unavailable("querying events by type"),
+        "kv_checkpoints" => feature_unavailable("fetching checkpoint contents"),
+        "kv_objects" => feature_unavailable("fetching object contents"),
+        "kv_transactions" => feature_unavailable("fetching transaction contents"),
         "obj_versions" => feature_unavailable("querying object versions"),
         "tx_affected_addresses" => {
             feature_unavailable("filtering transactions by affected address")
@@ -153,6 +157,26 @@ pub(crate) fn pipeline_unavailable(pipeline: &str) -> RpcError {
         "tx_digests" => feature_unavailable("querying transactions"),
         "tx_kinds" => feature_unavailable("filtering transactions by kind"),
         _ => anyhow!("unrecognized pipeline name: {}", pipeline).into(),
+    }
+}
+
+/// Guard a read path that consumes `pipeline` but does not go through [`AvailableRangeKey::reader_lo`].
+///
+/// Returns `FeatureUnavailable` when `pipeline` is tracked but gated off by the configured
+/// availability policy. A pipeline that is not tracked at all is a no-op: its data is either served
+/// from an external store (e.g. `kv_*` content under a Bigtable/Ledger backend) or genuinely
+/// absent, and those cases are handled by the read path itself.
+pub(crate) fn require_pipeline(ctx: &Context<'_>, pipeline: &str) -> Result<(), RpcError> {
+    let watermarks: &Arc<Watermarks> = ctx.data()?;
+    check_pipeline_available(watermarks, pipeline)
+}
+
+/// The availability decision behind [`require_pipeline`], split out so it can be unit-tested
+/// without a GraphQL context.
+fn check_pipeline_available(watermarks: &Watermarks, pipeline: &str) -> Result<(), RpcError> {
+    match watermarks.per_pipeline().get(pipeline) {
+        Some(p) if !p.available() => Err(pipeline_unavailable(pipeline)),
+        _ => Ok(()),
     }
 }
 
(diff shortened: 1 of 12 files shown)


crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -77,6 +87,102 @@ pub struct IndexerConfig {
...
+/// max-checkpoint-lag = 1000         # lag-gated override
+/// ```
+///
+/// The lag distance is measured against the network tip, approximated as the highest checkpoint
+/// high-watermark across all tracked pipelines (equal to the true network tip when a Bigtable or
+/// Ledger gRPC KV source is configured, and the fastest local pipeline otherwise). The default
+/// also covers the virtual `bigtable`/`ledger_grpc`/`consistent` pipelines tracked from those
+/// stores; the KV ones define the tip, so in practice a lag default never gates them. Gating a
+/// `kv_*` content pipeline only takes effect in a pure-Postgres deployment; when content is served
+/// from an external KV store those pipelines are not tracked, so the policy is inert.
```
wlmyng, line 121: it sounds like this availability feature is largely moot when we move off postgres to grpc though? We'd probably want grpc to expose a similar availability range?

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -537,6 +653,41 @@ impl ZkLoginLayer {
     }
 }
 
+impl TryFrom<AvailabilityLayer> for PipelineAvailability {
+    type Error = String;
+
+    fn try_from(layer: AvailabilityLayer) -> Result<Self, Self::Error> {
+        match (layer.enabled, layer.max_checkpoint_lag) {
+            (Some(_), Some(_)) => {
+                Err("'enabled' and 'max-checkpoint-lag' are mutually exclusive".to_string())
```
wlmyng, line 690: true + Some could be valid?

crates/sui-indexer-alt-graphql/src/api/types/available_range.rs
```diff
@@ -156,6 +160,26 @@ pub(crate) fn pipeline_unavailable(pipeline: &str) -> RpcError {
...
+    check_pipeline_available(watermarks, pipeline)
+}
+
+/// The availability decision behind [`require_pipeline`], split out so it can be unit-tested
+/// without a GraphQL context.
+fn check_pipeline_available(watermarks: &Watermarks, pipeline: &str) -> Result<(), RpcError> {
+    match watermarks.per_pipeline().get(pipeline) {
+        Some(p) if !p.available() => Err(pipeline_unavailable(pipeline)),
+        _ => Ok(()),
+    }
```
wlmyng, line 180: and this is set in watermark_task through `Watermarks::apply_availability` -> `policy.is_available()`

wlmyng (approved):

crates/sui-indexer-alt-graphql/src/api/types/object.rs
```diff
@@ -856,6 +861,8 @@ impl Object {
             return Ok(None);
         };
 
+        require_pipeline(ctx, "kv_objects")?;
```
wlmyng, line 864: i think this also requires `obj_versions` below at L871

wlmyng (commented):
should `epoch.rs:615` also require `cp_sequence_numbers`?

crates/sui-indexer-alt-graphql/src/config.rs
evan-wall-mysten: `enabled = true` means it's enabled unconditionally so it is not compatible with `max-checkpoint-lag` even though `max-checkpoint-lag` could result in it becoming enabled.

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -77,6 +87,102 @@ pub struct IndexerConfig {
...
+/// Per-pipeline configuration, under `[pipelines.<name>]`.
+#[derive(Clone, Default, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(default, rename_all = "kebab-case", deny_unknown_fields)]
+pub struct PipelineLayer {
+    /// Overrides the default availability policy for this pipeline. An override exists only to
+    /// set a policy, so its section must set `enabled` or `max-checkpoint-lag`.
+    pub availability: Option<PipelineAvailability>,
+}
+
+impl PipelineAvailability {
```
amnn: file order

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -77,6 +87,102 @@ pub struct IndexerConfig {
...
+
+    /// Per-pipeline overrides, keyed by pipeline name.
+    pub pipelines: BTreeMap<String, PipelineAvailability>,
+}
+
+/// TOML mirror of [`PipelineAvailability`]: a policy section sets exactly one of these keys,
+/// validated when converting to the enum.
+#[derive(Clone, Default, Debug, Deserialize, Serialize)]
+#[serde(default, rename_all = "kebab-case", deny_unknown_fields)]
+pub struct AvailabilityLayer {
```
amnn: The naming is off here -- it suggests that it's paired with `AvailabilityConfig` above, which is not the case.

crates/sui-indexer-alt-graphql/src/config.rs
amnn, line 1: Please try and keep to the pattern used by other configs in this file, otherwise it becomes a nightmare to maintain. E.g.

- Config vs layer separation with the same transitions between them (conversions to/from, `finish` API etc).
- File order.
- Conventions of where doc comments go (on the config types and their fields, not on the layers).

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -310,6 +327,56 @@ impl Watermarks {
...
+        // gRPC source is configured its virtual pipeline tracks the true network tip; computing the
+        // max over every pipeline (including any that end up gated) keeps the reference stable.
+        let network_tip = self
+            .per_pipeline
+            .values()
+            .map(|p| p.hi.checkpoint)
+            .max()
+            .unwrap_or(0) as u64;
+
+        for (name, pipeline) in self.per_pipeline.iter_mut() {
```
amnn, line 361:
```suggestion
        for (name, pipeline) in &mut self.per_pipeline {
```

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -310,6 +327,56 @@ impl Watermarks {
...
+            // A pipeline's own override wins, then the configured default; a pipeline with
+            // neither is always available.
+            pipeline.available = match config.policy_for(name) {
+                Some(policy) => policy.is_available(pipeline.hi.checkpoint as u64, network_tip),
+                None => true,
```
amnn, line 366:
```suggestion
            pipeline.available = config
                .policy_for(name)
                .is_none_or(|p| p.is_available(pipeline.hi.checkpoint as u64, network_tip));
```

amnn: Is it possible for a pipeline to show up in the list to be watermarked, and yet not have a policy? Previously a pipeline needed to be explicitly mentioned in GraphQL's config to be considered enabled.

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -310,6 +327,56 @@ impl Watermarks {
...
+        }
+
+        let mut global_hi = Watermark {
+            epoch: i64::MAX,
+            checkpoint: i64::MAX,
+            transaction: i64::MAX,
+        };
+        let mut timestamp_ms_hi_inclusive = i64::MAX;
+        let mut any_available = false;
+        for pipeline in self.per_pipeline.values().filter(|p| p.available) {
```
amnn, line 377: This is duplicating work that happens in the `merge` function -- can we get rid of it from there?

amnn (changes requested):
There are two major changes in this PR that I don't see addressed in the description or review -- requesting changes to discuss those and ideally have them split out and handled separately:

### `require_pipelines` pattern

This seems to be to introduce guarding for point look-ups, but we need to make some important decisions and/or changes here: because pipelines like `kv_checkpoints`, `kv_objects` etc are rarely enabled in production, because we fetch point lookup data from BigTable or Archival, and the watermark that we write for those is currently `bigtable` or `ledger_grpc` respectively.

- What should we do to normalise these cases? This is also important for the work of moving paginated APIs to be supported by gRPC.
- These pipelines are currently unpruned, what happens when that's no longer true?
  - What's the right way to think about their retention/available range? 
  - What kind of guarantees are we looking to offer if we are serving off of pruned point-lookup data?
- There's currently an existing abstraction that deals with pipeline availability and lowerbounds via `AvailableRangeKey::reader_lo` which abstracts the mapping from GraphQL schema types/field/parameters to pipeline names. This change introduces a new concept -- `require_pipeline` -- that seems to overlap, but needs to be supplied pipeline names directly which leaks that abstraction and weakens the guarantees that `AvailableRangeKey::reader_lo` and all its supporting machinery provides. How can we bring these two concepts closer together?

### Pipelines failing open or failing closed

Some parts of this change seem to imply that pipelines fail open (if they are not configured then they will be enabled), but the previous behaviour was that they failed closed (you had to explicitly name a pipeline in the config for GraphQL to use it and enable schema that depended on it).

The issue here is that now there are two ways to configure pipelines that are known to GraphQL -- by accepting `--indexer-config`s and extracting pipeline names from there, or via this new availability config.

We should not have two ways to configure pipelines for GraphQL, it's going to be a source of bugs. At the same time, we need to think carefully about the roll-out of this change to avoid breaking or churning existing operators.

evan-wall-mysten:
Will open a new PR with macro for handling pipeline availability

PR #27179, opened 2026-07-07
Encapsulate event bitpacking

nickvikeras:
## Description

Previously the LedgerService cursor was assumed to be opaque and hence could contain implementation details. It sort of graduated to a semi-public API that other internal services like GraphQL and gRPC's SubscriptionService need to understand.

wlmyng@ pointed out that this sort of couples all implementors of the gRPC interface to one event bitpacking scheme, which could cause problems if we ever need to split the rocksdb/bigtable impls, or add a third impl someday on a completely different backend database.

It also required gRPC SubscriptionService and GraphQL to understand the bitpacking scheme, which isn't ideal.

This PR refactors so the cursor is just the checkpoint/txn/event offsets, and the encoding into the bit-packed u64 happens inside gRPC, closer to the storage layer.

## Test plan

Unit tests all pass.


crates/sui-rpc-cursor/proto/sui/rpc/cursor/v1/cursor.proto
```diff
@@ -5,31 +5,56 @@ syntax = "proto3";
...
+    }
+}
 
-    // Whether `position` is a matched row or a scan frontier.
-    CursorKind kind = 2;
+message CheckpointsPosition {
+    // Checkpoint sequence coordinate. Item cursors resume strictly after it.
+    // Boundary cursors resume at it ascending / strictly below it descending.
+    optional uint64 checkpoint = 1;
+}
```
wlmyng, line 43: thanks for the great doc comments :pray:

crates/sui-rpc-cursor/proto/sui/rpc/cursor/v1/cursor.proto
```diff
@@ -5,31 +5,56 @@ syntax = "proto3";
...
 
-// Pagination cursor for the bitmap-backed ledger-history endpoints.
+// Pagination cursor for ledger-history endpoints. The cursor scope is encoded
+// by which `position` member is set.
 message CursorToken {
-    // Ledger-history query this cursor was minted for.
-    QueryType query_type = 1;
+    // Orthogonal to `position`: the same coordinate can be either a returned
+    // item or a scan boundary, with different resume inclusivity.
+    CursorKind kind = 1;
```
wlmyng: should we set the tags so that they don't collide with the previous version? given that this cursor format is in testnet graphql

crates/sui-kv-rpc/src/v2alpha/list_events.rs
```diff
@@ -573,25 +585,33 @@ fn unfiltered_event_refs(
...
     .boxed()
 }
 
-fn tx_range_for_event_range(event_range: std::ops::Range<u64>) -> Option<std::ops::Range<u64>> {
-    let last_event_seq = event_range.end.checked_sub(1)?;
-    let start_tx = event_bitmap_index::decode_event_seq(event_range.start).0;
-    let end_tx = event_bitmap_index::decode_event_seq(last_event_seq)
-        .0
-        .checked_add(1)?;
+fn tx_range_for_bounds(bounds: &EventScanBounds) -> Option<std::ops::Range<u64>> {
```
wlmyng: this might be better defined on `EventScanBounds`, since I see a verbatim in `sui-rpc-api/...event_scan.rs`

along with `bound_to_tuple` and `bounds_to_packed`


maybe a `fn tx_range -> Option<Range<u64>>` and `fn packed_range` which could subsume the `event_bound_to_tuple` as it's just an intermediate form on the way to packing

Since sui-rpc-api already has a dependency on sui-inverted-index

crates/sui-kv-rpc/src/v2alpha/list_events.rs
```diff
@@ -642,22 +661,38 @@ fn expand_event_refs(
...
 
+fn event_bound_to_tuple(bound: Bound<EventPosition>) -> Bound<(u64, u32)> {
+    match bound {
+        Bound::Included(position) => Bound::Included((position.tx_seq, position.event_index)),
+        Bound::Excluded(position) => Bound::Excluded((position.tx_seq, position.event_index)),
+        Bound::Unbounded => Bound::Unbounded,
+    }
+}
+
+fn decode_event_position(event_seq: u64) -> EventPosition {
```
wlmyng: I think this could be `EventPosition::decode` which should still work with `m.map_item(EventPosition::decode)...`

which should dedupe the fullnode's `event_scan.rs`

crates/sui-kv-rpc/src/v2alpha/list_events.rs
```diff
@@ -131,70 +136,73 @@ pub(crate) async fn list_events(
...
+                options.scan_direction(),
+                scan_budget,
+                ctx.bitmap_scan_observer(),
             )
-        };
+            .map_ok(|m| {
+                m.map_item(decode_event_position)
+                    .map_watermark(decode_event_position)
+                    .map_item(|position| EventRef {
+                        position,
```
wlmyng: with the methods on `EventScanBounds` and `EventPosition`, this could look like
```
.eval_bitmap_query_stream(
    query,
    event_bounds.packed_range(),
    BitmapIndexSpec::event(),
).map_ok(|m| {
    m.map_item(EventPosition::decode) // and so on
})
```

crates/sui-rpc-api/src/grpc/v2alpha/ledger_service/event_scan.rs
```diff
@@ -0,0 +1,310 @@
...
+        LedgerBitmapKind::Event,
+        EVENT_BITMAP_BUCKET_SIZE,
+        query,
+        pending_bucket,
+        packed_range,
+        direction,
+        hit_limit,
+        scan_budget,
+        cancel,
+    )?;
```
wlmyng: and here, with the methods on the two types, we could have

```
packed_range = bounds.map(|bounds| bounds.packed_range());
...
items: hits.items.into_iter().map(EventPosition::decode).collect(),
...
next_bounds: hits.next_range.map(EventScanBounds::from_packed)
...
```
and so on

crates/sui-kv-rpc/src/v2alpha/list_events.rs
wlmyng: oh, if each backend owns their own encoding, `packed_range` should accept a pack fn

```
pub fn packed_range(
          &self,
          pack: fn(Bound<(u64, u32)>, Bound<(u64, u32)>) -> Range<u64>,
      ) -> Range<u64> {
          pack(
              self.lo.map(EventPosition::to_tuple),
              self.hi.map(EventPosition::to_tuple),
          )
      }
```

wlmyng (commented):
Initial pass, thanks for taking this on! Had some thoughts on how we could consolidate the bounding and (un)packing into sui-rpc-api while allowing each backend to define their own encoding. Got a few more files to review

crates/sui-kv-rpc/src/v2alpha/list_events.rs
nickvikeras: Good call on the tx_range() helper, but I think a packed_range helper in sui-rpc-api would be a step backwards. IMO it's better to keep the packing logic in sui-inverted-index (since that is ultimately the "storage layer" that understands/defines ) it.

I made the existing packed_range() fn take Into<(u64, u32>) and made EventPosition implement that so now the rpc layer just passes its types directly into the index layer which I think has the same LoC reduction effect you were going after with this comment.

crates/sui-rpc-api/src/ledger_history/watermark.rs
```diff
@@ -137,18 +136,19 @@ pub fn boundary_cursor_cp(cp: u64, direction: ScanDirection) -> u64 {
...
-    options: &QueryOptions,
-    end_checkpoint: u64,
-    end_position: u64,
-) -> Watermark {
+pub fn terminal_boundary_watermark(options: &QueryOptions, end_position: Position) -> Watermark {
+    let end_checkpoint = match end_position {
+        Position::Checkpoints { checkpoint }
+        | Position::Transactions { checkpoint, .. }
+        | Position::Events { checkpoint, .. } => checkpoint,
+    };
```
wlmyng: we could lift this into `Position::checkpoint`

crates/sui-rpc-api/src/ledger_history/query_options.rs
```diff
@@ -186,12 +267,119 @@ impl QueryOptions {
...
+
+    pub fn cursor_for_item(&self, position: Position) -> Bytes {
+        CursorToken::item(position).encode()
     }
 
-    pub fn cursor_for_boundary(&self, checkpoint: u64, position: u64) -> Bytes {
-        CursorToken::boundary(self.query_type, checkpoint, position).encode()
+    pub fn cursor_for_boundary(&self, position: Position) -> Bytes {
+        CursorToken::boundary(position).encode()
+    }
```
wlmyng: I think we can delete these methods as they don't reference anything on `QueryOptions`

crates/sui-rpc-api/src/grpc/v2alpha/ledger_service/event_scan.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+fn bounds_from_packed(range: Range<u64>) -> EventScanBounds {
+    EventScanBounds {
+        lo: Bound::Included(event_seq::decode_event_seq(range.start).into()),
+        hi: Bound::Excluded(event_seq::decode_event_seq(range.end).into()),
+    }
+}
+
+pub(super) struct DrainedEventHits {
```
wlmyng, line 31: nit: structs before the fns

crates/sui-rpc-api/src/grpc/v2alpha/ledger_service/event_scan.rs
```diff
@@ -0,0 +1,267 @@
...
+        next_bounds: hits.next_range.map(bounds_from_packed),
+        buckets_scanned: hits.buckets_scanned,
+        frontier: hits
+            .coalesced_frontier
+            .map(|seq| EventPosition::from(event_seq::decode_event_seq(seq))),
+        scan_limit_hit: hits.scan_limit_hit,
+    })
+}
+
+pub(super) struct UnfilteredScan {
```
wlmyng: nit: structs before the fns

crates/sui-kv-rpc/src/v2alpha/list_events.rs
```diff
@@ -131,70 +135,70 @@ pub(crate) async fn list_events(
...
+                options.scan_direction(),
+                scan_budget,
+                ctx.bitmap_scan_observer(),
             )
-        };
+            .map_ok(|m| {
+                m.map_item(|seq| EventPosition::from(event_seq::decode_event_seq(seq)))
+                    .map_watermark(|seq| EventPosition::from(event_seq::decode_event_seq(seq)))
+                    .map_item(|position| EventRef {
+                        position,
```
wlmyng:
```

  m.map_item(|seq| EventRef {
      position: EventPosition::from(event_seq::decode_event_seq(seq)),
      tx_seq_digest: None,
  })
  .map_watermark(|seq| EventPosition::from(event_seq::decode_event_seq(seq)))
```

crates/sui-rpc-cursor/src/lib.rs
```diff
@@ -60,39 +63,6 @@ impl CursorToken {
     pub fn decode(bytes: &[u8]) -> anyhow::Result<Self> {
         Self::try_from(grpc::CursorToken::decode(bytes)?)
```
wlmyng: For compatibility what do you think of adding a `LegacyCursorToken` to the proto mirroring the previous version of `CursorToken` and then adding something like

```
impl CursorToken {
      pub fn decode(bytes: &[u8]) -> anyhow::Result<Self> {
          // Current format first. Legacy bytes decode "successfully" here too —
          // their tags 1-4 are unknown fields and get skipped — but arrive with
          // no position set, which try_from rejects, sending us to the fallback.
          let token = grpc::CursorToken::decode(bytes)?;
          match Self::try_from(token) {
              Ok(token) => Ok(token),
              Err(err) => Self::decode_legacy(bytes).map_err(|_| err),
          }
      }

      fn decode_legacy(bytes: &[u8]) -> anyhow::Result<Self> {
          let legacy = grpc::LegacyCursorToken::decode(bytes)?;
          let kind = match legacy.kind {
              1 => CursorKind::Item,
              2 => CursorKind::Boundary,
              k => anyhow::bail!("unknown legacy cursor kind: {k}"),
          };
          let checkpoint = legacy.checkpoint.context("legacy cursor missing checkpoint")?;
          let position = legacy.position.context("legacy cursor missing position")?;
          let position = match legacy.query_type {
              1 => Position::Checkpoints { checkpoint },
              2 => Position::Transactions { checkpoint, tx_seq: position },
              // Legacy event cursors carry a bit-packed position this crate
              // deliberately cannot interpret.
              3 => anyhow::bail!("legacy event cursors are not supported"),
              q => anyhow::bail!("unknown legacy cursor query_type: {q}"),
          };
          Ok(Self { kind, position })
      }
  }
```

wlmyng (approved):
lgtm and approving to unblock. Main changes of cursor tracking just the offsets, en/decoding in grpc, graphql/ subscriptions no longer understand bitpacking are welcome changes. Thank you!

As follow-ups, I'd like to put up a few PRs around the following, flagging them here for input on the direction:

- replacing the deconstructed `(end_checkpoint, end_position)` pair in `ResolvedRange/ResolvedEventRange` with a single `sui_rpc_cursor::Position` to elminate `cursor_checkpoint`, `ChunkTerminal` reassembly in handlers, checkpoint-extraction match in `terminal_boundary_watermark` call
- unify `Range<u64>` and `Bound<EventPosition>`, I think the two apply cursor bounds impls can be merged
- optional `checkpoint` handling (position-only cursors hydrate checkpoint at ingress)

crates/sui-rpc-cursor/src/lib.rs
nickvikeras: This doesn't make the cutover fully seamless because the old service instances still won't understand the new cursor format. 

I initially considered adding the new fields and setting the old and new fields for one release, then ripping out the old fields in the next release, but given the low stakes of a blip of cursor errors, that didn't seem worth the overhead. That's why I just bumped the tags. 


But I guess the main benefit of what you are proposing here that if someone gets a new cursor and sends it to an old server, that request will _eventually_ succeed on a retry (because the old servers will be drained). Whereas an old cursor against a new server will just never succeed (without this suggested change).

So with that in mind, it seems worth tacking this on.

PR #27180, opened 2026-07-07
indexer-framework: clamp lower bound of processor channel size to 1

bmwill:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-framework/src/pipeline/sequential/mod.rs
+++ b/crates/sui-indexer-alt-framework/src/pipeline/sequential/mod.rs
@@ -150,7 +150,7 @@ pub(crate) fn pipeline<H: Handler>(
         .unwrap_or(ConcurrencyConfig::Adaptive {
             initial: 1,
             min: 1,
-            max: num_cpus::get(),
+            max: num_cpus::get().max(1),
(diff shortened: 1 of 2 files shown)


crates/sui-indexer-alt-framework/src/pipeline/sequential/mod.rs
amnn: Could you also move the `max(4)` here outside the `unwrap_or_else`, so it matches the pattern in your change?

amnn (approved):
Thanks @bmwill !

We might as well make the same change here (for concurrent pipelines):

https://github.com/MystenLabs/sui/blob/c55bccc97b17fe80c322037f9d44436cc4ae2246/crates/sui-indexer-alt-framework/src/pipeline/concurrent/mod.rs#L277-L286

We won't hit this code path now, because we only use sequential pipelines in `sui-rpc-store`, but for consistency it would make sense.

PR #27181, opened 2026-07-08
[graphql] Fix backwards pagination on `paginate_preloaded_transactions`

wlmyng:
## Description 

- paginate_preloaded_transactions needs to be direction-aware to correctly return results on backwards pagination
- Cursors minted when paginating a streamed checkpoint now carry the real checkpoint sequence number

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
@@ -75,6 +75,7 @@ pub(crate) struct TransactionContents {
 
 pub type CTransaction = MultiCursor<OpaqueCursor<CursorToken>, JsonCursor<u64>>;
 
+/// Custom `Connection` for transactions to support partially-filled pages.
 pub(crate) struct TransactionConnection {
     pub edges: Vec<Edge<String, Transaction, EmptyFields>>,
     pub page_info: PageInfo,
@@ -281,6 +282,7 @@ impl Transaction {
     // TODO(DVX-2068): Add cursor consistency test between subscriptions and query API.
     pub(crate) fn paginate_preloaded_transactions(
         scope: Scope,
+        source_cp_sequence_number: u64,
         transactions: &[ProcessedTransaction],
         page: &Page<CTransaction>,
         filter: TransactionFilter,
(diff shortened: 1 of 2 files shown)


crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -281,27 +282,35 @@ impl Transaction {
...
             .filter(|tx| before.is_none_or(|b| tx.tx_sequence_number <= b))
-            .take(page.limit_with_overhead())
             .collect();
 
+        // `paginate_results` expects the window of results surrounding the requested page, so a
+        // `last` page keeps the tail of the filtered set, not the head.
+        if page.is_from_front() {
+            filtered.truncate(page.limit_with_overhead());
+        } else {
+            filtered.drain(..filtered.len().saturating_sub(page.limit_with_overhead()));
```
tpham-mysten, line 305: When filtered.len() <= limit_with_overhead(), I believe it is a no-op, right? Maybe add a test for it too?

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -281,27 +282,35 @@ impl Transaction {
     // TODO(DVX-2068): Add cursor consistency test between subscriptions and query API.
     pub(crate) fn paginate_preloaded_transactions(
         scope: Scope,
+        source_cp_sequence_number: u64,
```
tpham-mysten, line 285: nit: is this change related?

tpham-mysten (approved):
Looking great! Thanks @wlmyng

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
wlmyng: ... not related to backwards pagination, no, but this is so we can set the correct `checkpoint_sequence_number` rather than 0

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
wlmyng: added!

PR #27183, opened 2026-07-08
[cli] Make `sui client chain-identifier` return Base58 and Hex formats

stefan-mysten:
## Description 

Make `sui client chain-identifier` return both Base58 and Hex for the chain identifier. Either can be used for `Move.toml` environments section. 

## Test plan 

New and existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [x] CLI: `sui client chain-identifier` now displays both Base58 and Hex format for the chain digest.
The Move package system now accepts both formats interchangeably: chain IDs in `Move.toml` `[environments]` section.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-package-alt/src/chain_id.rs
+++ b/crates/sui-package-alt/src/chain_id.rs
@@ -4,14 +4,16 @@
 //! Utilities for working with chain identifier strings.
 //!
 //! A chain identifier can appear in two formats:
-//! - the canonical form: the full genesis checkpoint digest, Base58-encoded (the format returned
-//!   by the gRPC and GraphQL APIs), e.g. `4btiuiMPvEENsttpZC7CZ53DruC3MAgfznDbASZ7DR6S`;
-//! - the legacy short form: the first 4 bytes of the digest, hex-encoded (the format returned by
-//!   JSON-RPC and written by older tooling into `Move.toml` and `client.yaml`), e.g. `35834a8a`.
+//! - the legacy short form: the first 4 bytes of the digest, hex-encoded, e.g. `35834a8a`. This is
+//!   the format returned by JSON-RPC, the format rendered by `ChainIdentifier`'s `Display`, and the
+//!   only format this tooling ever *writes* — into `Move.toml` and `client.yaml`.
+//! - the full Base58 form: the full genesis checkpoint digest, Base58-encoded (the format returned
+//!   by the gRPC and GraphQL APIs), e.g. `4btiuiMPvEENsttpZC7CZ53DruC3MAgfznDbASZ7DR6S`.
 //!
-//! New identifiers are written in the canonical form, but manifests and caches produced by older
-//! tooling still contain short forms, so comparisons must treat the two encodings of the same
-//! digest as equal.
+//! We never write the Base58 form, but a user may paste it (e.g. copied from a gRPC/GraphQL
+//! response or an explorer) into a `Move.toml` `[environments]` entry, while the CLI environment
+//! and `client.yaml` cache hold the hex short form. Comparisons must therefore treat the two
+//! encodings of the same digest as equal.
 
 use std::str::FromStr;
 
--- a/crates/sui-package-alt/src/chain_id.rs
+++ b/crates/sui-package-alt/src/chain_id.rs
@@ -1,0 +1,128 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Utilities for working with chain identifier strings.
+//!
+//! A chain identifier can appear in two formats:
+//! - the canonical form: the full genesis checkpoint digest, Base58-encoded (the format returned
+//!   by the gRPC and GraphQL APIs), e.g. `4btiuiMPvEENsttpZC7CZ53DruC3MAgfznDbASZ7DR6S`;
+//! - the legacy short form: the first 4 bytes of the digest, hex-encoded (the format returned by
+//!   JSON-RPC and written by older tooling into `Move.toml` and `client.yaml`), e.g. `35834a8a`.
+//!
+//! New identifiers are written in the canonical form, but manifests and caches produced by older
+//! tooling still contain short forms, so comparisons must treat the two encodings of the same
+//! digest as equal.
+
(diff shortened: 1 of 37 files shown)


stefan-mysten:
The only issue I see with this change is that if someone goes back and forth with different versions, a new CLI will overwrite the Published.toml file and it will have the new digest which will not be parsed by the older versions.

crates/sui-package-alt/src/chain_id.rs
```diff
@@ -0,0 +1,128 @@
...
+//!
+//! A chain identifier can appear in two formats:
+//! - the canonical form: the full genesis checkpoint digest, Base58-encoded (the format returned
+//!   by the gRPC and GraphQL APIs), e.g. `4btiuiMPvEENsttpZC7CZ53DruC3MAgfznDbASZ7DR6S`;
+//! - the legacy short form: the first 4 bytes of the digest, hex-encoded (the format returned by
+//!   JSON-RPC and written by older tooling into `Move.toml` and `client.yaml`), e.g. `35834a8a`.
+//!
+//! New identifiers are written in the canonical form, but manifests and caches produced by older
+//! tooling still contain short forms, so comparisons must treat the two encodings of the same
+//! digest as equal.
```
stefan-mysten: This is wrong, and I have fixed it in a local commit that I will push shortly.

crates/sui-sdk/src/digests.rs
mdgeorge4153, line 1: I wonder if this should live in one of the dependency crates instead; it seems not completely package-system-specific. Not sure where exactly though.

mdgeorge4153: maybe sui_sdk::digests, for example

crates/sui-sdk/src/digests.rs
```diff
@@ -0,0 +1,128 @@
...
+/// identifier format. (`ChainIdentifier`'s `Display` renders the legacy 4-byte hex short form.)
+pub fn chain_id_base58(chain_id: &ChainIdentifier) -> String {
+    CheckpointDigest::new(*chain_id.as_bytes()).base58_encode()
+}
+
+/// Compare two chain identifier strings, each of which may independently be in the canonical
+/// Base58 form or the legacy hex short form. A short form matches a full form if it encodes the
+/// first 4 bytes of the digest. Strings in neither format (e.g. ad-hoc identifiers used by tests
+/// or other flavors) only match by exact string equality.
+pub fn chain_ids_match(a: &str, b: &str) -> bool {
```
mdgeorge4153, line 32: My instinct is to define a chain ID enum with variants for the two flavors, but I don't feel strongly about it. The advantage would be that other places that use them don't have to just pass raw strings around.

crates/sui-package-alt/src/find_env.rs
```diff
@@ -103,13 +103,16 @@ impl EnvFinder<'_> {
     }
 
     /// Check that the exact environment `active_env` is present in the manifest; returns an error
-    /// if the name is present but the chain ID differs; returns `None` if it's not present
+    /// if the name is present but the chain ID differs; returns `None` if it's not present.
+    ///
+    /// On success, the returned environment carries the manifest's chain ID, which may be a
```
mdgeorge4153, line 109: IIUC this means that if the manifest contains the short chain ID then the published.toml will also contain the short chain ID, is that right? We could alternatively prefer the new-style chain ID if we have it, but I think I agree with your approach.

crates/sui-sdk/src/digests.rs
mdgeorge4153: You could also implement `Eq` on that type, so that the other parts of the code need to change less.

crates/sui-source-validation/tests/shell_tests/local_publish/publish.sh
```diff
@@ -3,7 +3,7 @@
 
 sui client --client.config $CONFIG switch --env localnet
 
-chain_id=$(sui client --client.config $CONFIG chain-identifier)
+chain_id=$(sui client --client.config $CONFIG chain-identifier | awk '/^Hex:/ {print $2}')
```
mdgeorge4153, line 6: I'm surprised about this change - isn't part of the goal of this PR that the output of `sui client chain-identifier` can match the entry in `[environments]`? Why do we need to mangle it?

crates/sui/src/client_commands.rs
```diff
@@ -193,7 +192,10 @@ pub enum SuiClientCommands {
         processing: TxProcessingArgs,
     },
 
-    /// Query the chain identifier from the rpc endpoint.
+    /// Query the chain identifier from the rpc endpoint. Prints it in both encodings: the full
```
mdgeorge4153, line 196: just summarizing the conversation from slack: the goals for the output are that we want something clear for the human user and also easy for scripting. we considered
1. `sui client chain-identifier` outputs both by default, and we add `--short` and `--long` or `--base58` or whatever for scriptability
2. `sui client chain-identifier` outputs json containing both
I may be partial to (1) at this point just so that users don't have to parse json, but I don't feel strongly about it.

crates/sui-source-validation/tests/shell_tests/local_publish/publish.sh
mdgeorge4153: Oh I see now, you're just extracting the right one.

mdgeorge4153 (approved):
Left a few thoughts but this looks good

crates/sui/src/client_commands.rs
stefan-mysten: Added a `--format=base58|hex` flag for this.

crates/sui-source-validation/tests/shell_tests/local_publish/publish.sh
stefan-mysten: No longer needed due to the new `--format` flag.

crates/sui-sdk/src/digests.rs
stefan-mysten: Good point, moved.

PR #27184, opened 2026-07-08
Add consensus submission pipeline doc

mwtian:
## Description

Add a doc for the transaction submission pipeline: the dataflow and behaviors from the `SubmitTransaction` RPC through the admission queue and consensus adapter, the design principles behind it, and the per-transaction error/effects surfaces. A `sui-core` CLAUDE.md entry and module doc pointers direct future changes to these files to keep the doc in sync.

The doc is mostly AI generated (and reviewed against the code for factual accuracy). the design principles and major themes are human (my) input.

## Test plan

Docs and comment-only changes; no behavior change. `cargo check -p sui-core` passes.

### DIFF
--- a/crates/sui-core/src/consensus_submission_pipeline.md
+++ b/crates/sui-core/src/consensus_submission_pipeline.md
@@ -1,0 +1,422 @@
+# Transaction Submission Pipeline: Admission Queue and Consensus Adapter
+
+This documents the dataflow from a `SubmitTransaction` RPC arriving at a validator,
+through the gas-price admission queue (`admission_queue.rs`) and the consensus adapter
+(`consensus_adapter.rs`), to the point where the RPC request completes and the
+background submission task settles.
+
+**Keep this doc in sync**: when changing behavior in `admission_queue.rs`,
+`consensus_adapter.rs`, or the submit path in `authority_server.rs`, update this file.
+
+## Overview
+
+```
+gRPC SubmitTx
+     │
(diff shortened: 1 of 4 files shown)


crates/sui-core/src/consensus_submission_pipeline.md
```diff
@@ -0,0 +1,437 @@
...
+```
+gRPC SubmitTx
+     │
+ValidatorService::handle_submit_transaction        (authority_server.rs)
+     │  per-tx validation, dedup, voting
+     │
+classify_submit_mode()
+     │
+     ├─ Bypass ───────────────────────────────┐
+     ├─ Disabled (per-tx overload reject) ────┤
```
mystenmark: imo the only thing that should go before the queue is amplification related submission delays. We are removing bypass, and per-tx overload can also be handled during queue draining.

alex-mysten: Bypass has been removed. 
Just wondering if now is the right time to check in this doc, as we are actively iterating on these components, and this doc will need to evolve quickly.

mwtian: I have updated the doc on this point. This PR also asks AI to keep the doc in sync with code changes, and we can review it periodically for drift. Alternatively, I can remove some of the more detailed descriptions. I do think if we intend to make further changes to the submission path, the design principles of this code path should be documented for both AI and ourselves.

PR #27194, opened 2026-07-08
indexer-alt: merge ingestion cohorts as they converge

evan-wall-mysten:
## Description

Continues work started in #27209 (merged).

Cohorts are fixed at startup, so a backfilling cohort that catches up keeps running its own ingestion service — duplicate fetching and streaming subscriptions — forever. This PR merges cohorts at runtime: once a trailing cohort's frontier comes within `cohort-merge-threshold` checkpoints (default 1,000) of the cohort ahead, it hands its subscribers over exactly once and winds down, so a caught-up indexer converges back to a single ingestion service.

The handoff goes through the cohort table:

- Each service advertises its frontier and the exclusive end of its in-flight range — the exact point at which a subscriber can still join in full — and absorbs subscribers registered there when it commits its next range (ingestion legs are chunked when merging is enabled; streaming commits per checkpoint).
- A merging cohort registers its subscribers at that point in one atomic step under the table lock, delivers everything below it, and winds down.
- The target stays completely passive — a trailing cohort can merge into a service parked on backpressure — and adoptees become plain subscribers of the target: native backpressure and shutdown, no extra hops.

## Test plan

New tests cover:

- The table protocol: registration and absorption at range commits, refusal while no join point is advertised, pass-along of pending adoptees across cascaded merges, and the guard closing pending adoptees' channels on wind-down.
- Broadcaster-level merges from both the ingestion and streaming paths, and immediate wind-down when a subscriber hangs up mid-stream.
- End-to-end merges: a two-cohort merge where a backfill converges on a backpressured near-tip cohort, and a three-cohort cascade where a handed-off subscription is handed off again -- in both, every pipeline receives its exact range exactly once.
- Adoption from both broadcast paths: subscribers registered at ingestion-chunk and streamed join points, absorbed mid-run and carried across leg transitions.
- Handoffs compose with per-subscriber resume points: `next_checkpoint` rides through registration and absorption unchanged, including an adoptee whose resume point lies beyond the join point.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [x] Indexing Framework: Ingestion cohorts now merge back together at runtime once a trailing cohort catches up to within `cohort-merge-threshold` checkpoints (default 1,000) of the cohort ahead; set the new `ingestion.cohort-merge-threshold` config to tune how eagerly they merge. Pipelines sharing an ingestion service no longer receive checkpoints below their own resume watermark.

### DIFF
--- a/crates/sui-analytics-indexer/src/config.rs
+++ b/crates/sui-analytics-indexer/src/config.rs
@@ -134,6 +134,7 @@ pub struct IngestionLayer {
     pub streaming_connection_timeout_ms: Option<u64>,
     pub streaming_statement_timeout_ms: Option<u64>,
     pub min_cohort_boundary: Option<u64>,
+    pub cohort_merge_threshold: Option<u64>,
 
     /// Deprecated: accepted (and ignored) so old configs don't fail to parse. Replaced by
     /// per-pipeline `ingestion.subscriber_channel_size`.
@@ -167,6 +168,9 @@ impl IngestionLayer {
                 .streaming_statement_timeout_ms
                 .unwrap_or(base.streaming_statement_timeout_ms),
             min_cohort_boundary: self.min_cohort_boundary.unwrap_or(base.min_cohort_boundary),
+            cohort_merge_threshold: self
+                .cohort_merge_threshold
+                .unwrap_or(base.cohort_merge_threshold),
         }
     }
 }
--- a/crates/sui-checkpoint-blob-indexer/src/config.rs
+++ b/crates/sui-checkpoint-blob-indexer/src/config.rs
@@ -119,6 +119,7 @@ pub struct IngestionConfig {
     pub streaming_connection_timeout_ms: u64,
     pub streaming_statement_timeout_ms: u64,
     pub min_cohort_boundary: u64,
+    pub cohort_merge_threshold: u64,
 
     /// Deprecated: accepted (and ignored) so old configs don't fail to parse. Replaced by
     /// per-pipeline `ingestion.subscriber-channel-size`.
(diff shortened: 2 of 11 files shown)


evan-wall-mysten:
Had claude create a diagram to help show how the cohort merge happens
cohort-convergence

amnn (commented):
This is very complex (in the literal sense -- binds lots of different components together). There are significantly less complex ways to achieve the same goal (merging cohorts but never splitting them) which maintain cleaner boundaries.

If we wanted to ship this feature, that work would be a must, but it won't be no complexity, so the question is: Is this problem worth the complexity it takes to solve it?

An indexer that needs to regularly backfill pipelines is also being regularly redeployed (at which point cohorts are recalculated) and we redeploy way more often than we add new pipelines. So in practice we are talking about marginal savings for a significant increase in maintenance burden (even for a less complex solution).

My opinion is that we should not take this burden lightly, and in fact we should wait for a clear sign that what is being left on the table is non-trivial before we try and tackle it, but let us discuss as a team next week.

nickvikeras:
sorry did not mean to close this 😅

nickvikeras:
> This is very complex (in the literal sense -- binds lots of different components together). There are significantly less complex ways to achieve the same goal (merging cohorts but never splitting them) which maintain cleaner boundaries.
> 
> If we wanted to ship this feature, that work would be a must, but it won't be no complexity, so the question is: Is this problem worth the complexity it takes to solve it?
> 
> An indexer that needs to regularly backfill pipelines is also being regularly redeployed (at which point cohorts are recalculated) and we redeploy way more often than we add new pipelines. So in practice we are talking about marginal savings for a significant increase in maintenance burden (even for a less complex solution).
> 
> My opinion is that we should not take this burden lightly, and in fact we should wait for a clear sign that what is being left on the table is non-trivial before we try and tackle it, but let us discuss as a team next week.

I think a different approach that would result in less complexity would be to wait until the new cohorts are near, then just stop both broadcasters, wait for the channels to be drained, then drop them and start a new broadcaster with all of the subscribers at the lower height of the two cohorts that are being merged. 

Then you don't have to touch sui-futures at all and I think the life cycling of cohorts and their broadcasters will become its own state machine.

evan-wall-mysten:
Will leave this unchanged and depend on operators restarting the indexer

PR #27197, opened 2026-07-08
nickv/filtered subscriptions

nickvikeras:
## Description

Filtered subscription APIs

## Test plan

Added some basic unit tests, I'll also have LLMs hammer it with synthethic traffic generated from snowflake to hunt for bugs and test it under load like I did for the list APIs.

---

## Release notes

- [ ] Protocol:
- [x] gRPC: Adding filtered subscription APIs for checkpoints, transactions, and events
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-rpc-api/src/subscription/matcher.rs
```diff
@@ -0,0 +1,920 @@
...
+                    sub.checkpoints_since_frame += 1;
+                    if sub.checkpoints_since_frame < interval {
+                        continue;
+                    }
+                    SubscriptionUpdate::WatermarkTick {
+                        checkpoint: cp,
+                        tx_hi,
+                    }
+                }
+            };
```
bmwill, line 564: here we could do a quick check if the channel is closed

bmwill (approved):

PR #27198, opened 2026-07-08
[chore] Bump to Rust 1.96.1

stefan-mysten:
## Description 

Upgrade the Rust toolchain from 1.92 to 1.96.1.

- Bump rust-toolchain.toml channel 1.92 → 1.96.1.
- Bump the builder base image rust:1.92-bullseye → rust:1.96.1-bullseye across all 16 docker/*/Dockerfiles.
- Fix all new Clippy lints introduced/tightened by the newer toolchain so both cargo xclippy -Dwarnings (in crates/) and cargo move-clippy -D warnings (in external-crates/move/) pass clean. No #[allow(...)] suppressions were added — every finding is fixed at the source.

The Clippy fixes are mechanical and behavior-preserving. By lint:

- useless_conversion — drop redundant .into_iter() in IntoIterator-accepting calls (zip, zip_eq, zip_debug_eq, chain, extend, batch insert/delete). Largest group.
- unnecessary_sort_by → sort_by_key (using std::cmp::Reverse where the comparator was reversed).
- iter_kv_map → .values() / .into_values() instead of .iter()/.into_iter() + key-discarding map.
- unnecessary_unwrap — replace x.is_some() + x.unwrap() with if let Some(..) / let-chains.
- collapsible_if / collapsible_match — fold nested ifs into match guards / let-chains (excluded from external crates move code)
- manual_checked_ops / manual_saturating_arithmetic → checked_div / saturating_sub.
- manual_option_zip → Option::zip.
- result_large_err — shrink oversized error enums by boxing the largest variant (ReplayEngineError::UnexpectedEventFormat) and returning the boxed SuiError wrapper instead of the large SuiErrorKind.
- while_let_loop, unnecessary_cast, unused_variables — one-off fixes.

## Test plan 

Existing CI tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/config/src/committee.rs
+++ b/consensus/config/src/committee.rs
@@ -83,6 +83,7 @@ impl Committee {
     /// from the nominal total stake computed above. We will scale `f` and `c`
     /// from the nominal value to the largest possible values where intersection
     /// properties still hold. Then the thresholds are computed with scaled `f` and `c`.
+    #[allow(clippy::int_plus_one)] // Avoid clippy warning about `+ 1` in threshold asserts.
     pub fn new_v3(
         epoch: Epoch,
         authorities: Vec<Authority>,
@@ -123,7 +124,7 @@ impl Committee {
 
         // Ensure intersection between committed certification and quorum thresholds.
         assert!(
-            certification_threshold + quorum_threshold > actual_total_stake + f,
+            certification_threshold + quorum_threshold >= actual_total_stake + f + 1,
             "Stake-safety invariant violated: \
                 committed_cert ({certification_threshold}) + \
                 quorum ({quorum_threshold}) < \
--- a/consensus/config/src/committee.rs
+++ b/consensus/config/src/committee.rs
@@ -123,7 +123,7 @@ impl Committee {
 
         // Ensure intersection between committed certification and quorum thresholds.
         assert!(
-            certification_threshold + quorum_threshold >= actual_total_stake + f + 1,
+            certification_threshold + quorum_threshold > actual_total_stake + f,
             "Stake-safety invariant violated: \
                 committed_cert ({certification_threshold}) + \
                 quorum ({quorum_threshold}) < \
--- a/consensus/core/src/leader_schedule.rs
+++ b/consensus/core/src/leader_schedule.rs
@@ -270,7 +270,7 @@ impl LeaderSwapTable {
         assert_eq!(authorities_by_score.len(), context.committee.size());
         authorities_by_score.shuffle(&mut rng);
         // Stable sort the authorities by score descending. Order of authorities with the same score is preserved.
-        authorities_by_score.sort_by(|a1, a2| a2.1.cmp(&a1.1));
+        authorities_by_score.sort_by_key(|a2| std::cmp::Reverse(a2.1));
 
         // Calculating the good nodes
         let good_nodes = Self::retrieve_first_nodes(
--- a/consensus/core/src/leader_schedule.rs
+++ b/consensus/core/src/leader_schedule.rs
@@ -270,7 +270,7 @@ impl LeaderSwapTable {
         assert_eq!(authorities_by_score.len(), context.committee.size());
         authorities_by_score.shuffle(&mut rng);
         // Stable sort the authorities by score descending. Order of authorities with the same score is preserved.
-        authorities_by_score.sort_by_key(|a2| std::cmp::Reverse(a2.1));
+        authorities_by_score.sort_by_key(|a| std::cmp::Reverse(a.1));
 
         // Calculating the good nodes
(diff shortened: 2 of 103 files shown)


consensus/config/src/committee.rs
```diff
@@ -123,7 +123,7 @@ impl Committee {
 
         // Ensure intersection between committed certification and quorum thresholds.
         assert!(
-            certification_threshold + quorum_threshold >= actual_total_stake + f + 1,
+            certification_threshold + quorum_threshold > actual_total_stake + f,
```
mwtian: I wonder which lint is this. Maybe we can disable it?

consensus/core/src/leader_schedule.rs
```diff
@@ -270,7 +270,7 @@ impl LeaderSwapTable {
         assert_eq!(authorities_by_score.len(), context.committee.size());
         authorities_by_score.shuffle(&mut rng);
         // Stable sort the authorities by score descending. Order of authorities with the same score is preserved.
-        authorities_by_score.sort_by(|a1, a2| a2.1.cmp(&a1.1));
+        authorities_by_score.sort_by_key(|a2| std::cmp::Reverse(a2.1));
```
mwtian:
```suggestion
        authorities_by_score.sort_by_key(|a| std::cmp::Reverse(a.1));
```

consensus/core/src/leader_schedule_v3.rs
```diff
@@ -218,7 +218,7 @@ impl LeaderScheduleV3 {
         by_score.shuffle(&mut rng);
 
         // Use stable sort to produce deterministic order of authorities with the same score.
-        by_score.sort_by(|a, b| b.1.cmp(&a.1));
+        by_score.sort_by_key(|b| std::cmp::Reverse(b.1));
```
mwtian:
```suggestion
        by_score.sort_by_key(|a| std::cmp::Reverse(a.1));
```

consensus/core/src/proposer.rs
```diff
@@ -223,7 +223,7 @@ impl ValidatorProposer {
 
         // Sort scores descending so we can include the best of the pending excluded
         // ancestors first until we reach the threshold.
-        score_and_pending_excluded_ancestors.sort_by(|a, b| b.0.cmp(&a.0));
+        score_and_pending_excluded_ancestors.sort_by_key(|b| std::cmp::Reverse(b.0));
```
mwtian:
```suggestion
        score_and_pending_excluded_ancestors.sort_by_key(|a| std::cmp::Reverse(a.0));
```

mwtian (approved):
A few comments but lgtm for consensus/ code.

bmwill (approved):

external-crates/move/crates/move-ir-to-bytecode/src/compiler.rs
```diff
@@ -272,10 +272,10 @@ fn verify_move_function_body(code: &[Block]) -> Result<()> {
             match &statement.value {
                 Statement_::Jump(label)
                 | Statement_::JumpIf(_, label)
-                | Statement_::JumpIfFalse(_, label) => {
-                    if !labels.contains(&label.value) {
-                        undeclared.push(&label.value);
-                    }
+                | Statement_::JumpIfFalse(_, label)
+                    if !labels.contains(&label.value) =>
+                {
```
tnowacki: I do not like this set of changes in external-crates. 
Can we exclude the rule (whatever it is about `if` in a `match`) from move-clippy?

tzakian: +1 -- while "correct" I don't like the way the code reads with it/the code motion that it induces is a bit discombobulating.

tzakian: I think this is coming from the `collapsible_match` clippy?

consensus/config/src/committee.rs
stefan-mysten: Fixed it.

consensus/core/src/leader_schedule.rs
stefan-mysten: Fixed

consensus/core/src/leader_schedule_v3.rs
stefan-mysten: Fixed

consensus/core/src/proposer.rs
stefan-mysten: Fixed!

mwtian (approved):

ebmifa (approved):

PR #27201, opened 2026-07-08
[trace-view] Support for adding source-level metadata

awelc:
## Description 

This PR modifies the debugger extension to simplify enabling source-level debugging for replayed on-chain transactions. Prior to this PR, in order to enable source-level debugging, the user needed to understand the structure of the replay tool's directory, the structure of Move package's directory, and do a some manual copying operations. This created too much friction discouraging some users from even trying debugging replayed on-chain transactions as they did not want to do disassembly debugging only and also did not want to bother with complicated instructions.

With improvements implemented in this PR, they can just point the debugger to the local Move package directory with a click of a button:

image

## Test plan 

Tested manually, both the success path as well as various error scenarios (like artifacts already copied to a directory in the previous session, or debugger's inability to reconcile correct package ID)


external-crates/move/crates/move-analyzer/trace-debug/src/extension.ts
```diff
@@ -281,7 +304,7 @@ export function activate(context: vscode.ExtensionContext) {
     // When opening compressed trace file in the "default" editor,
     // close the editor and open another one showing decompressed
     // content.
-    vscode.workspace.onDidOpenTextDocument(async doc => {
+    context.subscriptions.push(vscode.workspace.onDidOpenTextDocument(async doc => {
```
awelc, line 307: Unrelated small fix

external-crates/move/crates/move-analyzer/trace-debug/src/extension.ts
```diff
@@ -317,23 +340,62 @@ class MoveTraceViewProvider implements vscode.CustomReadonlyEditorProvider {
             // Do not fire custom editor for trace files already displayed
             // correctly via mtrace scheme.
             vscode.commands.executeCommand('workbench.action.closeActiveEditor');
+            return;
```
awelc, line 343: Another unrelated small fix

external-crates/move/crates/move-analyzer/trace-debug/tsconfig.json
```diff
@@ -8,6 +8,7 @@
         ],
         "sourceMap": true,
         "rootDir": "src",
+        "esModuleInterop": true,
```
awelc, line 11: This was needed due to TOML parsing library addition

jessiemongeon1:
## Style Guide Audit

**6** violation(s) found. All must be fixed before merge.

### `docs/content/references/ide/debugger.mdx` (6 violation(s))

_6 violation(s) (0 regex, 6 claude)_

- **Line 178** — passive_voice
  - Current: `The source code version of each package should be the same version that was used to build the package`
  - Fix: `The source code version of each package should match the version that built the package`
- **Line 180** — passive_voice
  - Current: `The package's source code used to generate the debugging metadata`
  - Fix: `The package's source code that generates the debugging metadata`
- **Line 180** — passive_voice
  - Current: `if a call to `bar` was part of the execution trace`
  - Fix: `if the execution trace includes a call to `bar``
- **Line 180** — passive_voice
  - Current: `when the call is reached during the debugging session`
  - Fix: `when the debugger reaches the call during the session`
- **Line 193** — passive_voice
  - Current: `does not contain the source code for Move calls executed throughout the transaction`
  - Fix: `does not contain the source code for Move calls that execute throughout the transaction`
- **Line 193** — passive_voice
  - Current: `are the instructions on how to handle this simplified case, followed by instructions`
  - Fix: `the following instructions explain how to handle this simplified case, followed by instructions`

---

external-crates/move/crates/move-analyzer/trace-debug/src/add_source.ts
```diff
@@ -0,0 +1,393 @@
...
+ * id of the latest published version in the `latest-published-id` entry.
+ *
+ * @param pkgRoot root directory of the package (containing the lock file).
+ * @returns published ids found across all environments (empty if the lock file is
+ * missing or records no published ids).
+ */
+function readLockPublishedIds(pkgRoot: string): string[] {
+    const lockFilePath = path.join(pkgRoot, LOCK_FILE_NAME);
+    if (!fs.existsSync(lockFilePath)) {
+        return [];
```
tzakian, line 149: Why does this one return `[]` in this case and not `undefined`?

external-crates/move/crates/move-analyzer/trace-debug/src/add_source.ts
```diff
@@ -0,0 +1,393 @@
...
+/**
+ * Reads the on-chain published package ids recorded in a package's published file.
+ * The file records one publication per environment in its `published` table, with
+ * the id of the latest published version in the `published-at` entry.
+ *
+ * @param pkgRoot root directory of the package (containing the published file).
+ * @returns published ids found across all environments (empty if the file records
+ * none), or `undefined` if the file does not exist.
+ */
+function readPublishedFileIds(pkgRoot: string): string[] | undefined {
```
tzakian: I know why you didn't name it this way, but for consistency I kinda want this to be `readPublishedPublishedFileIds`...

tzakian: I'd maybe see if there's some consistency that can be had between these three functions (i.e., all returning `string[]` or `string[] | undefined`) but having them all return slightly different things makes the place that stitches these all together a little annoying.

e.g., if they all returned `string[] | undefined` in specific manner I _think_ you could do something like
```
readPublishedFileIds(..) || readLockPublishedIds(..) || readManifestPublishedIds(..)
```

which feels kinda nice

tzakian (approved):
Looks good to me, Jessie flagged some docs stuff, so please fix that before landing.

Otherwise some comments that could maybe clean the reading of published IDs a bit, but otherwise LGTM!

external-crates/move/crates/move-analyzer/trace-debug/src/add_source.ts
awelc: Renamed all three to "readIdsFromXXXFile" which should do the trick

awelc: It's actually intentional, though (added a comment to make it a bit more explicit) as each of these files is indeed treated differently:
- if Published.toml exists, it's the source of truth (don't look for IDs in any other file even if it's empty, corrupted, etc.) as we are dealing with the newest package system and if package is published the IDs should be in Published.toml
- if Move.lock file contains IDs, use them (it's a newer package system version), but if it does not have them (or it does not exist), still search for them in Move.toml file (as we may not be in the newer package system version, or we may be looking at system stuff which still encodes IDs in Move.toml)

external-crates/move/crates/move-analyzer/trace-debug/src/add_source.ts
awelc: Explained in more detail in another [comment](https://github.com/MystenLabs/sui/pull/27201#discussion_r3555755294), but the idea is that if IDs are not in Move.lock or if it does not exist, we still consult Move.toml to try and find them

PR #27202, opened 2026-07-08
Load-shed system execution overload via admission queue drain throttle

alex-mysten:
## Description

CORE-197. Stacked on #27149.

Under execution overload we previously shed load by probabilistically rejecting transactions at signing, regardless of gas price. This routes that shedding through the admission queue instead: the drain is throttled in proportion to the overload monitor's shed percentage, so the bounded gas-price priority queue fills and sheds the lowest-gas-price transactions by eviction. The signing-time probabilistic reject is skipped when the queue is enabled.

The drainer wakes on shed-level changes so the throttle lifts promptly, and failover distinguishes a deliberate throttle (queue holding work with no capacity under the reduced cap) from a genuinely stuck actor by computing capacity live.

## Test plan

New unit tests in `admission_queue.rs` / `execution_driver_tests.rs` cover: the shed-scaled drain cap and price-ordered eviction, failover suppressed under a throttle/saturation but tripped by a stuck actor mid-shed, the event-driven resume when shed clears, and the signing-time reject being skipped when the queue is enabled.

## Release notes

- [ ] Protocol: 
- [x] Nodes (Validators and Full nodes): Under execution overload, validators now shed load by gas price via the admission queue rather than rejecting transactions at random. During congestion, low-gas-price transactions may be rejected with an outbidding error instead of a generic retry-after.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -935,7 +935,7 @@ pub struct AuthorityState {
     pub config: NodeConfig,
 
     /// Current overload status in this authority. Updated periodically.
-    pub overload_info: AuthorityOverloadInfo,
+    pub overload_info: Arc<AuthorityOverloadInfo>,
 
     /// The chain identifier is derived from the digest of the genesis checkpoint.
     chain_identifier: ChainIdentifier,
@@ -1317,6 +1317,13 @@ impl AuthorityState {
     }
 
     fn check_authority_overload(&self, tx_data: &SenderSignedData) -> SuiResult {
+        // When the admission queue is enabled, execution-overload load shedding is
+        // done by throttling the queue drain (which sheds the lowest-gas-price
+        // transactions by priority), not probabilistically here.
+        if self.overload_config().admission_queue_enabled {
+            return Ok(());
+        }
+
         if !self.overload_info.is_overload.load(Ordering::Relaxed) {
             return Ok(());
         }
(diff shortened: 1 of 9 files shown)


crates/sui-core/src/authority.rs
```diff
@@ -1317,6 +1317,13 @@ impl AuthorityState {
     }
 
     fn check_authority_overload(&self, tx_data: &SenderSignedData) -> SuiResult {
```
akichidis, line 1319: just to confirm that this didn't get unnoticed: this does change the rejection mechanics both for the transactions submitted to the validator, and when voting for other transactions (arriving from blocks). 
So the node now will not reject a transaction if it's overloaded - as this goes purely via the admission queue. I think we should maintain some level of rejection in this case.

akichidis (commented):
I understand that transactions arriving to a validator now will not get probabilistically rejected when overload shedding gets calculated, but they'll rather get routed via the admission queue. However, IIUC the actual shedding will happen only when the queue has backed up, and that will also depend on the consensus round rate (how fast we are able to submit transcations to consensus).
With current setting of `max_pending_transactions = 20K` , even with load shedding of 50%, the effective max number of transactions will be 10K. Consensus capacity - checked via `consensus_adapter.num_inflight_transactions()`  - might still have lots of it , unless consensus round rate suffers as well (which we don't know) . Where I am getting to is that although execution might ask for backpressure/tx prioritisation, unless consensus slows down too and effectively see the admission queue back up and start prioritising more aggressively.

Let me do another pass, but that's my understanding for now unless I am missing some critical piece.

crates/sui-core/src/authority.rs
alex-mysten: This is a valid concern - overload now reduces the effective num_inflight, but if consensus is working normally, but execution is still overloaded, we don't actually load shed. I'll look into this.

mystenmark (approved):
looks good to me, lets make sure to test in PTN though

PR #27203, opened 2026-07-09
Improve address balance withdrawal error messages

lxfind:
## Description 

The insufficient-balance error at signing time leaked the accumulator object ID, which is an internal implementation detail (e.g. `Available amount in account for object id 0xdcf0f9... is less than requested: 0 < 5000000000`) and gave users no hint about what "account" refers to. The message now names the coin type and the owner address of the balance (sender or sponsor), states that the address balance is insufficient, and explains that funds held in Coin objects are not part of the address balance and should be passed as transaction inputs directly. The gasless remaining-balance error had the same object-ID leak and is reworded the same way.

To make the owner address available at the check site, the withdrawal maps returned by `process_funds_withdrawals_for_signing`/`_for_estimation` now carry the owner address alongside the amount and type tag.

## Test plan 

Covered by existing tests, updated to the new wording: the `insufficient_gas_payment` and `gasless_remaining_balance_reject` transactional-test snapshots capture the full new messages for both error paths (including that the reported owner address distinguishes different balance owners), and the e2e address-balance tests assert the message when withdrawing from a non-existent balance, withdrawing more than available, when an explicit withdrawal plus implicit gas withdrawal exceeds the balance, and when a coin reservation exceeds the reserved coin's balance.

---

## Release notes

- [ ] Protocol: 
- [x] Nodes (Validators and Full nodes): Improved the error message returned when a transaction's address balance withdrawal exceeds the available balance: it no longer exposes internal accumulator object IDs, and now names the coin type and owner address and clarifies that Coin objects owned by the address are not part of its address balance.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/accumulators/funds_read.rs
+++ b/crates/sui-core/src/accumulators/funds_read.rs
@@ -7,7 +7,7 @@ use move_core_types::language_storage::TypeTag;
 use sui_types::{
     accumulator_root::AccumulatorObjId,
     balance::Balance,
-    base_types::SequenceNumber,
+    base_types::{SequenceNumber, SuiAddress},
     error::{SuiErrorKind, SuiResult, UserInputError},
 };
 
@@ -40,9 +40,9 @@ pub trait AccountFundsRead: Send + Sync {
     /// where deterministic results are not required.
     fn check_amounts_available(
         &self,
-        requested_amounts: &BTreeMap<AccumulatorObjId, (u64, TypeTag)>,
+        requested_amounts: &BTreeMap<AccumulatorObjId, (u64, TypeTag, SuiAddress)>,
     ) -> SuiResult {
-        for (object_id, (requested_amount, type_tag)) in requested_amounts {
+        for (object_id, (requested_amount, type_tag, owner)) in requested_amounts {
             let actual_amount = self.get_latest_account_amount(object_id);
 
             if actual_amount < *requested_amount as u128 {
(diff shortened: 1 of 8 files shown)


crates/sui-core/src/accumulators/funds_read.rs
```diff
@@ -42,15 +42,21 @@ pub trait AccountFundsRead: Send + Sync {
...
 
             if actual_amount < *requested_amount as u128 {
+                let coin_type = Balance::maybe_get_balance_type_param(type_tag)
+                    .unwrap_or_else(|| type_tag.clone());
                 return Err(SuiErrorKind::UserInputError {
                     error: UserInputError::InvalidWithdrawReservation {
                         error: format!(
-                            "Available amount in account for object id {} is less than requested: {} < {}",
-                            object_id, actual_amount, requested_amount
+                            "Insufficient address balance of coin type {coin_type}: \
```
alex-mysten: This is definitely better. Should we include the owner address for situations like sponsored gas payments or object-owned balances

alex-mysten (approved):
Approved with one comment

crates/sui-core/src/accumulators/funds_read.rs
lxfind: Good idea — done. The withdrawal map now carries the owner address (computed in `accumulate_funds_withdrawals` via `owner_for_withdrawal`, so it's the sender or the sponsor as appropriate), and both error messages report it. The `insufficient_gas_payment` snapshot shows it distinguishing the two owners (`@A` vs `@B`).

PR #27204, opened 2026-07-09
[Block Streaming] Release blocks on quorum

akichidis:
## Description 

In the observer service, buffer live accepted blocks per round and release them to the
subscriber once the round has a quorum, instead of forwarding every block as soon as it
is accepted:

- Blocks are buffered per round, aggregating the author stake as they arrive. When the
  blocks buffered for round `R` reach a quorum, all blocks of rounds `<= R` are released -
  a quorum at round `R` implies the lower rounds are settled, so there is no reason to
  keep them buffered.
- A release timeout (reusing `leader_timeout`) guards liveness: if round `R` has not
  reached a quorum within the timeout of its first buffered block (e.g. blocks were
  dropped due to broadcast lag, or the subscription started mid-round), rounds `<= R`
  are released anyway.
- Since releases always cover a contiguous prefix of rounds, the highest released round
  is tracked as a watermark: blocks arriving at or below it are passed straight through
  to the subscriber without being buffered again.

Blocks are always released in ascending round order per batch. The quorum is counted only
from blocks observed on the subscription's own stream, so a round can miss quorum locally
even when the validator has it - such rounds simply fall back to the timeout release.

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/observer_service.rs
+++ b/consensus/core/src/observer_service.rs
@@ -1,16 +1,22 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{sync::Arc, time::Duration};
+use std::{
+    collections::BTreeMap,
+    sync::Arc,
+    time::{Duration, Instant},
+};
 
 use async_trait::async_trait;
 use bytes::Bytes;
+use consensus_config::Committee;
 use consensus_types::block::{BlockRef, Round};
 use futures::{StreamExt as _, stream};
 use parking_lot::RwLock;
 use sui_macros::fail_point_async;
 use tap::TapFallible;
-use tokio::sync::broadcast;
+use tokio::sync::{broadcast, mpsc};
+use tokio_stream::wrappers::ReceiverStream;
 
 use crate::{
     BlockVerifier, RandomnessSignatureHandler, TransactionVoteTracker,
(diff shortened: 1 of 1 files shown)


mwtian:
@copilot resolve the merge conflicts in this pull request

consensus/core/src/observer_service.rs
```diff
@@ -326,9 +337,194 @@ impl ObserverNetworkService for ObserverService {
...
+
+    fn next_timeout(&self) -> tokio::time::Instant {
+        // Effectively no timeout when nothing is buffered - the select loop re-evaluates on
+        // every incoming batch anyway.
+        const NO_BUFFERED_ROUNDS_TIMEOUT: Duration = Duration::from_secs(3600);
+
+        let next_timeout = self
+            .buffered_rounds
+            .values()
+            .map(|buffered| buffered.first_buffered_at + self.release_timeout)
```
mwtian, line 500: Why not just use the 1st buffered round's timestamp?

mwtian (approved):

consensus/core/src/observer_service.rs
akichidis: Yes we could do that as well, and in theory it should be the same as we do here - unless blocks arrive out of order, which shouldn't be the case

mwtian: Yeah just a question of simplicity.

PR #27213, opened 2026-07-09
[source-verification 2] Reimplement verify-source

mdgeorge4153:
## Description

`sui client verify-source` did not work: the toolchain-versioned rebuild only ran behind an env var and otherwise built with the *current* compiler, the wrong environment's toolchain was read from the metadata, and a missing toolchain version silently fell back to the running compiler.

Reimplemented around one idea: rebuild the source with the toolchain that published it, and compare. Given only a package path, the tool reads the on-chain address, original id, and toolchain version from the package's own publication metadata (via `read_publication` from #27241 — the same resolution the package system uses when linking against the package, so the verifier checks exactly the package a dependent would link), downloads and caches that release's binary, rebuilds the source with it in a subprocess, and compares the root package's bytecode and linkage against the chain. The verifier compiles nothing itself — editions, dependency resolution, and tree-shaking are the toolchain's job.

Scope is deliberately narrow: only the **root** package. To verify a dependency, run the command again on that dependency's source. The old `--verify-deps` / `--skip-source` / `--address-override` surface is gone, and there is no longer an on-chain-id argument. `--toolchain-version` remains, to rebuild with a different release when the recorded one is unusable (e.g. a garbage-collected framework revision); it warns when it differs from the recorded version.

**Linkage** is compared as an intersection, since either build may have tree-shaken unused dependencies away: every dependency shared by both linkages must resolve to the same storage id, and a one-sided entry — unreferenced by the compared-equal bytecode, hence unused — is ignored.

Renames `sui-source-validation` to `sui-source-verification` (hence the rewrite-shaped diff) and drops the old bytecode-comparison crate. See `crates/sui-source-verification/DESIGN.md`.

## Test plan

- Unit tests for the comparison (`0x0` rewrite, module diff, linkage intersection), toolchain/dependency-pinning parsing, and the archive-layout `find_binary`.
- Shell tests publish to a localnet with the current CLI, verify, and confirm a tampered source is rejected. No downloads.
- `tests/version_matrix.rs` (`#[ignore]`d) publishes a fixture with a historical binary and verifies it with the current CLI — modern packages end-to-end through the CLI, legacy packages through the library directly. `era_matrix` asserts over a curated set (v1.23.1 → v1.74.1, v1.23 being where `publish` first records a resolvable address) and runs nightly in a **non-required** job (`.github/workflows/source-verification-matrix.yml`) that also verifies the latest release. A one-off sweep of every mainnet release established the coverable range and the three gaps (no binary ≤ v1.8.1; protocol-config deserialization v1.25.1–v1.29.2; v1.64.1's missing framework revision).
- Spot-checked against mainnet DeepBook (v4–v8): each verifies at its published version; a tampered source and the wrong version fail.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: `sui client verify-source <path>` now rebuilds the package with the toolchain it was published with (downloading that release if needed) and reads the on-chain address to compare against from the package's own publication metadata — it no longer takes a package-id argument. `--verify-deps`, `--skip-source`, and `--address-override` are removed; to verify a dependency, run the command against that dependency's source. Pass `--toolchain-version` to rebuild with a different toolchain.
- [ ] Rust SDK:
- [ ] Indexing Framework:


docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,114 @@
...
+Only the **root** package is verified: the bytecode of its modules and its linkage table (the
+versions of its dependencies). The source of the dependencies themselves is not checked. To verify a
+dependency, run `verify-source` again against that dependency's own source and address.
+
+The command performs no compilation itself. It resolves the toolchain version the package was
+published with, downloads that release's `sui` binary if it is not already cached under
+`$MOVE_HOME/binaries`, and uses it to build the source. Because the on-chain bytecode was produced by
+that same toolchain, a matching build is strong evidence that the source is authentic.
+
+:::note
```
mdgeorge4153: should maybe be stronger and/or mentioned in the first paragraph.

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,114 @@
...
+The most common cause is source drift: the working tree is close to, but not exactly, what was
+published. The command reports which modules differ, and a small number of named modules is a strong
+hint that only those files have changed since publication.
+
+Check out the exact commit that was published. The commit that *records* a package's `published-at`
+in source control is often a later commit than the one that was actually built, so search for when
+the address first appeared:
+
+```bash
+git log -S <PACKAGE_ID> -- '**/Published.toml' '**/Move.lock'
```
mdgeorge4153: Is there a simple CLI command that can tell you the date that something was published, and then we could suggest using git log with that date? I wouldn't want a hugely complicated snippet here, but I don't think the way it's described here is correct.

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+`--toolchain-version` is also the workaround when a package's recorded toolchain cannot be used at
+all — for example, a release whose framework revision is no longer available. The command fails early
+in that case and suggests trying a different version.
+
+## When verification fails
+
+A failure does not necessarily mean the source is wrong. Reproducing a build is exacting, and there
+are a few common reasons a genuine source fails to verify.
+
```
mdgeorge4153, line 95: We might recommend publishing a tag with a fixed source package once you get verification to pass (including the toolchain-version and pinned deps).

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,114 @@
...
+:::note
+
+Verification compares **bytecode**, not source text. Changes that do not affect the compiled output —
+comments, formatting, and the bodies of `macro` functions that are never called — can differ between
+the source and what was published and still verify successfully. Conversely, a package that verifies
+has bytecode identical to what is on chain.
+
+:::
+
+## Choosing the toolchain version
```
mdgeorge4153: Maybe we should move this section into the "when verification fails" below?

.github/workflows/source-verification-matrix.yml
```diff
@@ -0,0 +1,61 @@
+name: Source Verification Version Matrix
+
+# Rebuilds a fixture with historical `sui` release binaries and verifies the result with the current
+# CLI. This downloads release archives from GitHub, so it is deliberately kept out of the required
+# pull request checks: it runs on a schedule and on demand only, and a failure here never blocks a
+# merge.
+on:
+  schedule:
+    - cron: '0 10 * * *' # UTC timing is every day at 2am PST
```
mdgeorge4153, line 9: should this be folded into the existing nightly checks? I'm not familiar with how our CI is currently set up.

crates/sui-source-verification/DESIGN.md
mdgeorge4153, line 1: It is possible that move-analyzer will want to use the comparison infrastructure. We should think about whether some parts of the infrastructure are not sui-specific and can be moved into external-crates.

crates/sui-source-verification/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+/// on-chain original id, and the modules and linkage are compared against the on-chain package
+/// fetched via `client`.
+pub async fn verify_source(
+    source_path: &Path,
+    on_chain_id: ObjectID,
+    env: &Environment,
+    client: &Client,
+    opts: VerifyOptions,
+) -> Result<(), AggregateError> {
+    if AccountAddress::from(on_chain_id) == AccountAddress::ZERO {
```
mdgeorge4153: I'm not sure we need this special case - we'd just try to fetch 0x0 and conclude that the address doesn't resolve to a package.

crates/sui/src/client_commands.rs
```diff
@@ -620,29 +614,27 @@ pub enum SuiClientCommands {
...
-        /// Path to directory containing a Move package
-        #[clap(name = "package_path", global = true, default_value = ".")]
+        /// Path to directory containing a Move package. Required: clap does not permit an optional
+        /// positional to precede a required one, and the package id is required.
+        #[clap(name = "package_path")]
         package_path: PathBuf,
 
+        /// The on-chain package id to verify against.
+        #[clap(name = "package_id")]
+        on_chain_id: ObjectID,
```
mdgeorge4153: In retrospect, I wonder if either this should be optional (and taken from the pubfile) or if we should check it against the pubfile. Otherwise it seems like a supply-chain attack vector - I publish the wrong address in my source package and then someone links against my source package and ends up linking against a different on-chain package. In fact, now that I think about it, maybe we should disallow overriding it and require it to come from the package metadata.

crates/sui/src/client_commands.rs
```diff
@@ -620,29 +614,27 @@ pub enum SuiClientCommands {
...
-
-        /// Don't verify source (only valid if --verify-deps is enabled).
-        #[clap(long)]
-        skip_source: bool,
-
-        /// If specified, override the addresses for the package's own modules with this address.
-        /// Only works for unpublished modules (whose addresses are currently 0x0).
-        #[clap(long)]
-        address_override: Option<ObjectID>,
+        toolchain_version: Option<String>,
```
mdgeorge4153, line 639: Similarly to my comment about the publication address, maybe we should actually disallow overriding the toolchain version and just force the user to add the toolchain version to the pubfile, or at least check if the pubfile contains a toolchain version that mismatches the provided toolchain version.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: In fact, I wonder how hard it would be to detect unused macros and produce a warning. Probably requires getting into bed with the compiler, in which case it's probably not worth the complexity.

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,233 @@
...
+        |e| Error::BinaryDownload {
+            version: version.to_string(),
+            message: e.to_string(),
+        },
+    )?;
+
+    Ok(canonical)
+}
+
+/// Download the `sui` release tarball for `version` and install it atomically into `version_dir`.
```
mdgeorge4153:
```suggestion
/// Download the `sui` release tarball for `version`; extract the `sui` binary from it, and install it atomically into `version_dir`.
```

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,233 @@
...
+    version: &str,
+    platform: &str,
+    binary_name: &str,
+    cache_root: &Path,
+    version_dir: &Path,
+) -> anyhow::Result<()> {
+    fs::create_dir_all(cache_root).context("creating binary cache directory")?;
+
+    // Temp dir on the same filesystem as the cache so the final rename is atomic.
+    let tmp = cache_root.join(format!(".tmp-{version}-{}", std::process::id()));
```
mdgeorge4153, line 73: not thread-safe but probably fine

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,233 @@
...
+    let result = (|| {
+        // Unpack into a scratch subdirectory: a release archive carries every shipped binary
+        // (`sui-debug`, `sui-node`, ...), which is well over a gigabyte, and only `sui` is wanted.
+        let extract = tmp.join("extract");
+        fs::create_dir_all(&extract).context("creating extraction directory")?;
+        download_and_extract(version, platform, &extract)?;
+
+        // Release archives have used two layouts: modern releases place `sui` at the archive root,
+        // while releases up to ~v1.8 nest it at `target/release/sui-<platform>`. Locate whichever is
+        // present and install it under the canonical cache path.
```
mdgeorge4153: If this ever changes again for a future release, would that be caught by the unit tests, or nightly tests on the next release, or not at all?

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,233 @@
...
+            .ok_or_else(|| anyhow!("no sui binary found in the {version} release archive"))?;
+
+        let release_dir = tmp.join("target").join("release");
+        fs::create_dir_all(&release_dir).context("creating release directory")?;
+        let canonical = release_dir.join(binary_name);
+        fs::rename(&found, &canonical).context("installing extracted binary")?;
+        set_executable_permission(canonical.as_os_str())?;
+
+        // Drop the archive and the binaries we do not use.
+        fs::remove_dir_all(&extract).context("cleaning up extracted archive")?;
```
mdgeorge4153, line 96: if something fails before this step, we'll have big orphaned temp dirs. Maybe we should always do the cleanup?

mdgeorge4153: oh I see, we are.

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,233 @@
...
+    fs::create_dir_all(&tmp).context("creating temporary install directory")?;
+
+    let result = (|| {
+        // Unpack into a scratch subdirectory: a release archive carries every shipped binary
+        // (`sui-debug`, `sui-node`, ...), which is well over a gigabyte, and only `sui` is wanted.
+        let extract = tmp.join("extract");
+        fs::create_dir_all(&extract).context("creating extraction directory")?;
+        download_and_extract(version, platform, &extract)?;
+
+        // Release archives have used two layouts: modern releases place `sui` at the archive root,
```
mdgeorge4153: I think this comment duplicates the `find_binary` comment and could probably be removed.

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,233 @@
...
+        "DOWNLOADING".bold().green(),
+        version.yellow()
+    );
+
+    let reader = match ureq::get(&mainnet_url).call() {
+        Ok(response) => response,
+        Err(ureq::Error::Status(404, _)) => {
+            debug!("no mainnet release for {version}, trying testnet");
+            let testnet_url = format!(
+                "https://github.com/MystenLabs/sui/releases/download/testnet-v{version}/sui-testnet-v{version}-{platform}.tgz",
```
mdgeorge4153, line 143: This is another place where I worry we are baking in conventions about the release process that will just silently break if the conventions ever change.

crates/sui-source-verification/src/build.rs
```diff
@@ -0,0 +1,147 @@
...
+struct DumpOutput {
+    modules: Vec<String>,
+    dependencies: Vec<ObjectID>,
+}
+
+/// Rebuild the package at `source_path` with `binary` and return its compiled modules and linkage
+/// dependencies via `--dump-bytecode-as-base64`.
+///
+/// The flag set is adapted to what `binary` supports (probed from its `--help`): older binaries lack
+/// `--no-tree-shaking` (their output is whatever they shake) and the `--environment` selector.
```
mdgeorge4153: we removed `--environment`, we should probably remove the note too.

crates/sui-source-verification/src/compare.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    if errs.is_empty() {
+        Ok(())
+    } else {
+        Err(AggregateError(errs))
+    }
+}
+
+/// Rewrite each generated module's `0x0` self-address to `original_id` and key the results by module
+/// name. Modules whose self-address is missing or already populated are reported as errors.
```
mdgeorge4153: I don't understand the last sentence here

crates/sui-source-verification/src/compare.rs
```diff
@@ -0,0 +1,255 @@
...
+
+/// Resolve the generated linkage's storage ids to `original id -> storage id` by fetching each
+/// dependency package. A storage id that cannot be fetched is reported as
+/// [`Error::SourceDependencyNotOnChain`].
+async fn fetch_generated_linkage(
+    client: &Client,
+    deps: &[ObjectID],
+    errs: &mut Vec<Error>,
+) -> BTreeMap<AccountAddress, AccountAddress> {
+    let results = future::join_all(deps.iter().map(|id| pkg_for_address(client, *id))).await;
```
mdgeorge4153: not a big deal, but can this be a single RPC call instead of one per address?

crates/sui-source-verification/src/compare.rs
```diff
@@ -0,0 +1,255 @@
...
+    let mut map = BTreeMap::new();
+    for (storage_id, result) in deps.iter().zip_debug_eq(results) {
```
mdgeorge4153: not sure what zip_debug_eq is, but could we just have the map closure above hold onto the id?

crates/sui-source-verification/src/pinning.rs
```diff
@@ -0,0 +1,159 @@
...
+        });
+    }
+
+    for (name, child) in table {
+        collect(child, Some(name), found);
+    }
+}
+
+fn is_commit_hash(rev: &str) -> bool {
+    rev.len() == 40 && rev.bytes().all(|b| b.is_ascii_hexdigit())
```
mdgeorge4153: I think the old package system would allow you to pin to truncated hashes; I think those are reliable enough that we should allow them. I believe git refs can't be named entirely in hex (though I could be wrong).

crates/sui-source-verification/src/pinning.rs
```diff
@@ -0,0 +1,159 @@
...
+    };
+
+    let mut found = vec![];
+    collect(&lock, None, &mut found);
+    found.sort_by(|a, b| (&a.dependency, &a.rev).cmp(&(&b.dependency, &b.rev)));
+    found.dedup();
+    found
+}
+
+/// Walk `value`, recording any table with a `source.rev` that is not a commit hash. Lockfiles have
```
mdgeorge4153, line 42: A little sad that we have to implement our own parser here instead of using something existing, but I guess  the package system legacy support doesn't go all the way back.

crates/sui-source-verification/src/pinning.rs
mdgeorge4153, line 1: it occurs to me that mvr dependencies can also change over time (I think?) in the legacy package system. I'm not sure I care enough to deal with it though.

mdgeorge4153: unless those get pinned as git deps? I'm not sure.

crates/sui-source-verification/src/toolchain_version.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// record this separately from the chain they were published to (a package can be published to
+    /// a localnet while building against mainnet dependencies), so it must be used to reproduce the
+    /// bytecode. `None` when the publication's environment is the chain's own environment.
+    pub build_env: Option<String>,
+}
+
+impl ToolchainVersion {
+    /// Resolve the toolchain version to rebuild the package at `root_path` with, for `env`.
+    ///
+    /// Precedence: `override_version` if given, then the ephemeral pubfile written by
```
mdgeorge4153: it feels weird to me that we're looking in ephemeral pubfiles - those are intended to be local and discarded, whereas source verification is specifically for publications on persistent networks. I guess this is mainly for our tests? We should consider changing the test harness so that this works better (e.g. by producing fake Published.toml files if needed).

crates/sui-source-verification/src/toolchain_version.rs
```diff
@@ -0,0 +1,204 @@
...
+    };
+
+    Ok(toolchain_from_metadata(
+        entry.metadata.toolchain_version,
+        Some(build_env),
+    ))
+}
+
+/// Read the toolchain version from `Published.toml`, selecting the publication whose recorded
+/// chain-id matches `env`. Returns `Ok(None)` if the file is absent or has no matching, complete
```
mdgeorge4153: Multiple environments can have the same chain-id. This should be matching on environment name.

crates/sui-source-verification/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+pub struct VerifyOptions {
+    /// Override the toolchain version used to rebuild the package. When `None`, the version is read
+    /// from the package's publish metadata.
+    pub toolchain_override: Option<String>,
+
+    /// The ephemeral pubfile written by `test-publish`. When `None`, the `Pub.<env>.toml` default
+    /// is consulted before falling back to `Published.toml` / `Move.lock`.
+    pub pubfile_path: Option<PathBuf>,
+
+    /// Override the environment whose dependencies the package is rebuilt against. When `None`, the
```
mdgeorge4153: I think this might be the wrong thing to do. I think ideally we'd match the environment selection process of the rest of the sui cli so that users don't get confused.

crates/sui-source-verification/tests/version_matrix.rs
```diff
@@ -0,0 +1,465 @@
...
+//! surface how well an old client interoperates with a current localnet). Run manually, and build
+//! the current CLI first so `get_cargo_bin("sui")` resolves:
+//!
+//! ```text
+//! cargo build -p sui
+//! cargo test -p sui-source-verification --test version_matrix -- --ignored --nocapture
+//! ```
+//!
+//! [`era_matrix`] asserts on a small curated set (Tier 3). [`historical_sweep`] reports a table over
+//! whatever versions are passed via `SUI_MATRIX_VERSIONS` (Tier 4 discovery); it does not assert.
```
mdgeorge4153: remove the "Tier 4 discovery" parenthetical.

crates/sui-source-verification/tests/version_matrix.rs
```diff
@@ -0,0 +1,465 @@
...
+        return false;
+    };
+    let text =
+        String::from_utf8_lossy(&out.stdout).into_owned() + &String::from_utf8_lossy(&out.stderr);
+    text.contains("test-publish")
+}
+
+/// Publish the fixture with the `version` binary in `mode` and verify it with `current_sui`. Returns
+/// `None` when the binary cannot support the mode.
+fn run_one(
```
mdgeorge4153: This function is very long; consider splitting it up if there are natural semantic helpers.

crates/sui-source-verification/DESIGN.md
mdgeorge4153: ah, I understand the requirements better now - I don't think splitting into external-crates is worth the effort. They actually want a `--no-build` option that skips the building and just does a binary comparison between an on-chain package and the already-generated build artifacts (without fetching a historical toolchain or rebuilding).

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Strengthened the caveat: it now spells out that the bodies of macros that *are* called are inlined and therefore covered, while only comments, formatting, and uncalled-macro bodies can differ. Kept it as a prominent `:::note` right after "What is verified" rather than in the first paragraph. See source-verification.mdx.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Reworked so it no longer needs a publish-date lookup: for the common case (a framework dependency) it points at `crates/sui-framework-snapshot/manifest.json` (revision per protocol version) or the `sui` release current at publish time. See the "Dependencies are not pinned to commit hashes" section.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Added a "Making a package verifiable" section recommending tagging the exact published commit, plus recording the toolchain-version and pinning deps to hashes. source-verification.mdx.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Done — it is now the "The toolchain version is not recorded" subsection under "When verification fails".

.github/workflows/source-verification-matrix.yml
mdgeorge4153: [CLAUDE] Kept it as its own scheduled workflow: it downloads release archives from GitHub and can take up to 2h, so it shouldn't share a job with the required nightly checks; it runs on schedule + `workflow_dispatch` and never blocks a merge. Happy to fold it under a single nightly umbrella if you'd prefer.

crates/sui-source-verification/DESIGN.md
mdgeorge4153: [CLAUDE] Noted as future work; nothing moved yet. The comparison currently leans on sui types (`ObjectID`, the gRPC `Client`, on-chain package fetch), so factoring the sui-agnostic bytecode-diff parts into external-crates is a separate follow-up rather than part of this PR.

crates/sui-source-verification/src/lib.rs
mdgeorge4153: [CLAUDE] Removed — the zero-address special case is gone. The address now comes from the package's publication rather than a CLI arg, so there is no 0x0 path to guard; a bad address simply fails to fetch. See `verify_source` in lib.rs.

crates/sui/src/client_commands.rs
mdgeorge4153: [CLAUDE] Addressed — this drove the rewire. The `on_chain_id` positional is removed; the address, original-id, and toolchain are read from the package's own publication via `read_publication`, the same resolution the package system uses to link this package. See the `VerifySource` handler in client_commands.rs.

crates/sui/src/client_commands.rs
mdgeorge4153: [CLAUDE] The address (the real supply-chain vector) is no longer overridable — it comes from the publication. `--toolchain-version` stays, because a recorded toolchain can be unusable (e.g. a gc'd framework rev), but `resolve_toolchain` now emits a WARNING when it differs from the recorded version instead of silently overriding. See `resolve_toolchain` in lib.rs.

crates/sui-source-verification/src/binary.rs
mdgeorge4153: [CLAUDE] Applied your suggestion to the `download_and_install` doc comment (binary.rs).

crates/sui-source-verification/src/binary.rs
mdgeorge4153: [CLAUDE] Agreed it's fine in practice: `ensure_binary` isn't called concurrently for the same version within a process, and the cross-process race is handled by the atomic rename plus the post-rename existence check. Left as-is.

crates/sui-source-verification/src/binary.rs
mdgeorge4153: [CLAUDE] Not by the unit tests (they cover only the two known layouts), but the nightly matrix now also verifies the *latest* release (via `SUI_MATRIX_EXTRA_VERSIONS`), so a layout change in a new release surfaces there as a download/`find_binary` failure. See the workflow and the `find_binary_handles_both_layouts` test.

crates/sui-source-verification/src/binary.rs
mdgeorge4153: [CLAUDE] The error path already cleans up: `download_and_install` runs the work in a closure and does `if let Err(e) = result { fs::remove_dir_all(&tmp); ... }`, so a failure before this step removes `tmp`. See the block right after the closure in binary.rs.

crates/sui-source-verification/src/binary.rs
mdgeorge4153: [CLAUDE] Removed — the call-site comment now just states what it does; the layout detail lives only in `find_binary`'s doc. binary.rs.

crates/sui-source-verification/src/binary.rs
mdgeorge4153: [CLAUDE] Acknowledged — the cache-path convention (`target/release/<name>`) is coupled to the release layout. Same mitigation as the archive-layout concern: the nightly now verifies the latest release, so a convention break surfaces there. Didn't add indirection for it.

crates/sui-source-verification/src/build.rs
mdgeorge4153: [CLAUDE] Fixed — the note referenced the removed `--environment`; it now reads `--build-env` (the flag actually probed from `--help`). See the `dump` doc in build.rs.

crates/sui-source-verification/src/compare.rs
mdgeorge4153: [CLAUDE] Reworded: a module already at `original_id` is kept as-is (some older toolchains emit the published address rather than 0x0); any other non-zero self-address is reported as an error. See the `rewrite_modules` doc in compare.rs.

crates/sui-source-verification/src/compare.rs
mdgeorge4153: [CLAUDE] There's no batch package-fetch on the gRPC `Client`, so it stays one call per dependency — but they run concurrently via `future::join_all`, so it's parallel rather than serial. Left as-is.

crates/sui-source-verification/src/compare.rs
mdgeorge4153: [CLAUDE] Done — the map closure now returns `(id, result)`, so the paired `zip` (and the `ZipDebugEqIteratorExt` import) are gone. See `fetch_generated_linkage` in compare.rs.

crates/sui-source-verification/src/pinning.rs
mdgeorge4153: [CLAUDE] Done — `is_commit_hash` now accepts any all-hex string of 7-40 chars (git's shortest abbreviation up to a full SHA). Added the `truncated_hash_is_not_reported` test. pinning.rs.

crates/sui-source-verification/src/pinning.rs
mdgeorge4153: [CLAUDE] Agreed it's unfortunate. The modern package system doesn't expose a reader for the legacy lock's moving-revision info, so this small walker stays for now.

crates/sui-source-verification/src/pinning.rs
mdgeorge4153: [CLAUDE] True — legacy MVR deps could also move over time. Left out of scope; `moving_revisions` only flags git branch/tag revs today. Noting it as a known gap.

crates/sui-source-verification/src/toolchain_version.rs
mdgeorge4153: [CLAUDE] Dropped ephemeral-pubfile support entirely. `toolchain_version.rs` now only reads the legacy `Move.lock`'s `[move].toolchain-version`; the modern toolchain comes from the persistent publication metadata.

crates/sui-source-verification/src/toolchain_version.rs
mdgeorge4153: [CLAUDE] Fixed by the environment rework: resolution keys on environment *name* throughout — `find_environment` selects the named env and `read_publication`/legacy lookups are name-keyed — so two environments sharing a chain-id no longer collide. That chain-id matcher is gone.

crates/sui-source-verification/src/lib.rs
mdgeorge4153: [CLAUDE] Done — the handler now resolves the environment with `find_environment`, the same selection the rest of the CLI uses (active env / `--build-env`), instead of a bespoke path. See the `VerifySource` handler in client_commands.rs.

crates/sui-source-verification/tests/version_matrix.rs
mdgeorge4153: [CLAUDE] Gone — the module doc was rewritten and no longer mentions tiers. See the version_matrix.rs module doc.

crates/sui-source-verification/tests/version_matrix.rs
mdgeorge4153: [CLAUDE] Split — extracted `stage_fixture` and `publish_fixture`; `run_one` went from ~97 to ~69 lines. version_matrix.rs.

crates/sui-source-verification/tests/version_matrix.rs
```diff
@@ -0,0 +1,417 @@
...
+            extra
+                .split(',')
+                .map(str::trim)
+                .filter(|s| !s.is_empty())
+                .map(String::from),
+        );
+    }
+
+    let outcomes = run_matrix(&versions).await;
+    print_table(&outcomes);
```
mdgeorge4153, line 126: Is there a way that these could show up as separate tests in nextest instead of one test that hits the whole matrix? I'd honestly even be ok with having a separate test for each of the curated versions that calls a common helper if there's not a better way.

crates/sui-source-verification/tests/version_matrix.rs
```diff
@@ -0,0 +1,417 @@
...
+    let failed: Vec<_> = outcomes.iter().filter(|o| !o.passed()).collect();
+    assert!(failed.is_empty(), "curated versions failed: {failed:#?}");
+}
+
+#[tokio::test]
+#[ignore = "downloads historical sui release binaries and requires network; run manually"]
+async fn historical_sweep() {
+    let versions: Vec<String> = match std::env::var("SUI_MATRIX_VERSIONS") {
+        Ok(list) => list.split(',').map(|s| s.trim().to_string()).collect(),
+        Err(_) => CURATED_VERSIONS.iter().map(|s| s.to_string()).collect(),
```
mdgeorge4153: This isn't the right fallback - we don't want to silently only test some of the versions if there's a config problem, we should just fail.

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,137 @@
...
+:::
+
+## Making a package verifiable
+
+If you publish a package and want others to be able to verify it out of the box:
+
+- **Pin every dependency to a commit hash**, not a branch or tag. The current package system does
+  this automatically; see [Dependencies are not pinned to commit
+  hashes](#dependencies-are-not-pinned-to-commit-hashes) below for older packages.
+- **Tag the exact commit that was published**, so verifiers can check out the precise source. The
```
mdgeorge4153: This is a little misleading - you want to tag the version that verifies. For example, if you added a Published.toml file you'd want to tag that commit.

mdgeorge4153: What we should say is something about finding the actual source code that you published

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,137 @@
...
+  commit that records the `published-at` address in source control is often a later commit than the
+  one that was actually built.
+- **Make sure the publication records the toolchain version.** Modern publishing writes it to
+  `Published.toml`. If yours does not (see below), you can add it.
+
+If a package's publication does not record its address or toolchain — for example, it was published
```
mdgeorge4153: We should either link to the appropriate section of the docs if there's already a page describing the format, or else give a little example here

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,137 @@
...
+```bash
+sui client verify-source <PACKAGE_PATH> --toolchain-version 1.71.1
+```
+
+If you do not know the exact version, the release that was current when the package was published is a
+good first guess. Compiler output is stable across neighbouring releases, so an adjacent version
+usually reproduces the same bytecode; a version from far in the past or future may not, because the
+bytecode file format changes occasionally.
+
+`--toolchain-version` is also the workaround when a package's recorded toolchain cannot be used at
```
mdgeorge4153: I think the list of supported versions should either be here or possibly in the error message for unsupported versions.

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+### Dependencies are not pinned to commit hashes
+
+The current package system pins every dependency to a specific commit hash, so its builds are
+reproducible. Packages published under the older system (before v1.63) may record a dependency's
+revision as a **branch** rather than a hash. Rebuilding then fetches whatever that branch points at
+now, which is almost never what was published — and often does not even compile against the older
+toolchain. When this happens, the command reports each dependency that is not pinned to a hash.
+
+The fix is to pin those dependencies yourself, in both `Move.toml` and `Move.lock`. Replace each
```
mdgeorge4153: The user should only change Move.toml; a rebuild should update Move.lock I think (we should do an experiment).

crates/sui-source-verification/tests/version_matrix.rs
mdgeorge4153: [CLAUDE] Left as a follow-up per discussion. The matrix shares one localnet across all curated versions, so splitting into per-version nextest tests would either spin up a cluster each (slow) or need a shared-fixture harness; not worth it right now. `era_matrix` still asserts over the whole curated set.

crates/sui-source-verification/tests/version_matrix.rs
mdgeorge4153: [CLAUDE] Fixed — `historical_sweep` now `expect()`s `SUI_MATRIX_VERSIONS` and fails if it's unset, rather than silently falling back to the curated set. See version_matrix.rs.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Reworded: it now says to tag a commit whose source *verifies* — not necessarily the original publish commit; if you added a `Published.toml` or pinned deps to make it verify, tag that commit. source-verification.mdx.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Added an inline `Published.toml` example (one `[published.<env>]` section with `chain-id`, `published-at`, `original-id`, `version`, `toolchain-version`). source-verification.mdx.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Added a supported-versions note to the docs: any release with a binary for your platform (roughly v1.8+) can be used as the rebuild toolchain, and packages published from v1.23 verify from metadata without extra arguments. The full sweep result (coverable range + the three gaps: no binary <= v1.8.1, protocol-config deserialization v1.25.1-v1.29.2, v1.64.1's missing framework rev) is now documented in DESIGN.md's testing section. I can also surface the specifics in the download error message if you'd prefer that over the docs.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Fixed — the docs now say to pin in `Move.toml` and rebuild, which updates `Move.lock`. Confirmed by experiment: editing only the manifest's `rev` (A->B) and running `sui move build` re-pinned the lock to B. source-verification.mdx and DESIGN.md.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Now surfaced in the error message too, hardcoded (no git/network query): a toolchain download or build failure appends a targeted suggestion. The versions known not to work for verification — v1.8.1 and earlier (no platform binary) and v1.64.1 (gc-ed framework) — get a hint naming a nearby working release (e.g. "1.63.4 or 1.65.2" for 1.64.1); anything else gets generic "try an adjacent release" advice. See `toolchain_suggestion` in error.rs (with a test).

mdgeorge4153: [CLAUDE] Restructured per your suggestion: rather than appending on failure, `check_toolchain_version` now runs up front and fails fast with a precise message for the known gaps — e.g. "The recorded toolchain sui 1.64.1 cannot verify this package: v1.64.x pins a framework revision that is no longer available. Passing --toolchain-version 1.65.2 may work." (<= v1.8.1 -> 1.9.0; v1.64.x -> 1.65.2). See `check_toolchain_version` in lib.rs, with a test.

jessiemongeon1:
## Style Guide Audit

crates/sui/src/client_commands.rs
```diff
@@ -1863,24 +1845,29 @@ impl SuiClientCommands {
...
-                    flavor: SuiFlavor::with_client(context),
-                };
-                let compiled_package = build_config
-                    .build_async_from_root_pkg(&mut root_pkg)
-                    .await?;
+                let flavor = SuiFlavor::with_client(context);
+                let publication =
+                    read_publication::<SuiFlavor>(&package_path, &environment, &flavor)
+                        .await?
+                        .ok_or_else(|| {
```
amnn: `with_context`

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,219 @@
...
+appeared and check out the earliest match:
+
+```bash
+git log --all -S <PACKAGE_ID> -- '**/Published.toml' '**/Move.lock'
+```
+
+Use `--all`: publishers frequently publish from a short-lived branch and merge it afterward, so the
+commit you want might not be on the main branch at all.
+
+That commit is a starting point, not an answer. A publish records the address *after* the build, so the
```
amnn: nit: reads very AI-esque (specifically the first sentence).

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,219 @@
...
+```bash
+sui client verify-source <PACKAGE_PATH> --toolchain-version 1.71.1
+```
+
+If you do not know the exact version, try the release that was current when the package was published.
+Compiler output is stable across neighboring releases, so an adjacent version usually reproduces the
+same bytecode; a version from far in the past or future might not, because the bytecode file format
+changes occasionally.
+
+That stability cuts both ways, and it is worth taking seriously before you start guessing. Because a
```
amnn: Also very AI-esque

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,219 @@
...
+Older toolchains require every dependency to declare its published address, and refuse to build without
+it:
+
+```
+Failed to publish the Move module(s), reason: Package dependency "Wormhole" does not specify a
+published address (the Move.toml manifest for "Wormhole" does not contain a published-at field).
+```
+
+This happens when the pinned revision of a dependency is a plain library checkout whose manifest leaves
+its address unset (`wormhole = "_"`), even though the package it links against is onchain. The publisher
```
amnn: switch to referencing a generic, or example package.

crates/sui-source-verification/DESIGN.md
```diff
@@ -0,0 +1,142 @@
...
+`sui client verify-source <package-path> [--toolchain-version <version>]` decides whether a Move
+source package compiles to the bytecode and linkage of the on-chain package it was published as.
+
+It rebuilds the source with the toolchain the package was published with, and compares the result to
+what is on chain:
+
+```
+verify_source(source_path, publication, toolchain_override, env, client, ...)
+  ├─ resolve the toolchain version   (publication metadata, the legacy Move.lock, or the override)
+  ├─ ensure that toolchain's binary  (cached under $MOVE_HOME/binaries, downloaded if absent)
```
amnn: It being under `$MOVE_HOME` but being the `sui` CLI is a bit interesting -- should it belong under `.sui` instead?

crates/sui-source-verification/DESIGN.md
```diff
@@ -0,0 +1,142 @@
...
+  ├─ rebuild the source with it      (`move build --dump-bytecode-as-base64`, in a subprocess)
+  ├─ fetch the on-chain package      (modules, linkage table, original id) at the publication's address
+  ├─ check the original id matches   (the package there is the one the source claims to be)
+  └─ compare                         (module bytecode, then linkage)
+```
+
+The `publication` — the on-chain address, the original id, and the recorded toolchain — is **not** a
```
amnn, line 20: I do think it's still worth allowing people to override whatever is in `Published.toml` with command-line flags. You have support for doing this to the toolchain version, but by the same token, it makes sense to allow that for the on-chain address and original ID.

This allows someone to work with packages where the `Published.toml` was overwritten or somehow messed with without bending over backwards.

crates/sui-source-verification/DESIGN.md
```diff
@@ -0,0 +1,142 @@
...
+
+The verifier performs no compilation itself. Everything about compiling Move — editions, flavors,
+dependency resolution, tree-shaking — is delegated to the downloaded toolchain, which is the same
+program that produced the on-chain bytecode. What the verifier knows is narrow:
+
+- how to read a toolchain version out of the publish metadata written by any historical release;
+- how to invoke a historical binary, whose command-line flags have changed over time;
+- how to fetch an on-chain package and compare bytecode and linkage.
+
+## Resolving the toolchain version
```
amnn, line 38: It may also be worth offering a `--toolchain` flag which would allow you to pass the path of a toolchain binary you already have on hand, skipping over the toolchain fetching logic.

crates/sui-source-verification/DESIGN.md
```diff
@@ -0,0 +1,142 @@
...
+
+A linkage table maps each dependency's **original id** to the **storage id** of the version being
+linked. Two linkages for the same package can legitimately differ in size, because a build may or may
+not have tree-shaken unused dependencies out. So the comparison is an intersection:
+
+1. Map every storage id in the rebuilt linkage to its original id (by fetching the packages).
+2. Intersect the two sets of original ids.
+3. For every original id in the intersection, require the storage ids to be equal.
+
+An original id present in only one linkage is ignored. That is sound because the module bytecode is
```
amnn: Linkage differences outside the intersection do have on-chain consequences (they apply versioning constraints when the package appears in another package's linkage).

Should we indicate when we detect IDs on only one side but default to treating that as a warning (and opt-in as an error?)

crates/sui-source-verification/DESIGN.md
```diff
@@ -0,0 +1,142 @@
...
+  the one the source claims to be. A `Published.toml` takes precedence over legacy addresses in both
+  the verifier and the package system, so adding one to make an old package verifiable opens no
+  discrepancy; and publications are keyed by environment **name**, not chain id, so two environments
+  that share a chain id are never confused.
+
+- **The wrong linkage.** Module bytecode does not by itself pin the *versions* of the dependencies a
+  package calls into — that is the linkage table's job — so a package whose modules match but whose
+  linkage does not is running against different dependency code. Sui's linkage is flat: the root
+  package's table alone chooses the version of every transitive dependency, and a dependency's own
+  table is ignored at runtime. The verifier therefore compares the whole table, not just the root's
```
amnn: > and a dependency's own table is ignored at runtime

This is not strictly true -- a dependency's linkage impacts the root's by imposing version constraints.

crates/sui-source-verification/DESIGN.md
```diff
@@ -0,0 +1,142 @@
...
+linkage is unreachable, and therefore unused.
+
+## Security
+
+Verifying source against the wrong on-chain data would let a package look authentic when it is not.
+Two failure modes drive the design:
+
+- **The wrong address.** If the verifier compared the source against an address chosen independently
+  of the package system, an attacker could arrange for a source to verify against one address while a
+  package *depending on it* links a different one — so the "verified" source is not the code that
```
amnn, line 87: I'm not really following the implications of this attack. It seems to be a reason to not allow an on-chain ID override to be supplied (which I advocated for above), but the way I see it, there could be two questions you're trying to answer using source verification:

- When I publish my package depending on a source package S, the package I link with on-chain will behave how S behaved during testing.
- When I want to understand how some on-chain package P works, the source I have in my hand is the source that compiled to that on-chain package.

In the first scenario, as long as the default behaviour of `verify-source` is to check the package's `Published.toml`, the ability to override the address you're checking against doesn't seem to pose a threat -- the attacker would have to make you test against a different ID to "trick" you. If you are trusting the attacker to tell you how to verify the package, they have simpler ways to trick you.

In the second scenario, you may be dealing with packages of unknown provenance, and having the flexibility to supply *a* copy of the source and test it against some custom on-chain ID can be very useful.

crates/sui-source-verification/src/pinning.rs
```diff
@@ -0,0 +1,176 @@
...
+
+/// The moving revisions recorded in the lockfile `contents`.
+fn moving_revisions_in(contents: &str) -> Vec<MovingRevision> {
+    let Ok(lock) = contents.parse::<toml::Value>() else {
+        return vec![];
+    };
+
+    let mut found = vec![];
+    collect(&lock, None, &mut found);
+    found.sort_by(|a, b| (&a.dependency, &a.rev).cmp(&(&b.dependency, &b.rev)));
```
amnn: nit: you would get this for free if you derived `PartialOrd` and `Ord` on `MovingRevision` (just call `sort`).

crates/sui-source-verification/src/onchain.rs
```diff
@@ -0,0 +1,68 @@
...
+use sui_types::move_package::MovePackage;
+use sui_types::object::Data;
+
+use crate::error::Error;
+
+/// The on-chain representation of a package: its modules (keyed by name), its original (runtime)
+/// package id, and its linkage table reduced to `original id -> storage id`.
+pub struct OnChainPackage {
+    pub original_id: AccountAddress,
+    pub modules: BTreeMap<Symbol, CompiledModule>,
```
amnn: Is there a reason to use `Symbol` rather than `String` here? `Symbol` is usually reserved for use in the compiler for strings that we need to compare against a lot.

amnn (approved):
This looks good to me! Overall very clean and easy to review, thanks @mdgeorge4153.

I do think long-term, having a separate binary that gets downloaded for the purposes of toolchain version checks would be beneficial:

- It would result in a much, much smaller download 1GB is *a lot*, even 150MB for the `sui` CLI is a lot.
- We avoid accidentally breaking the tooling by making a change to the command-line structure or the release archive directory structure that's intended primarily for humans.
- We can make it the responsibility of this new binary to handle compatibility issues, obviating the need for all the shenanigans where you need to check help output to see which features are available -- we can create a binary and pick it back across however many releases we want to support, without worrying about messing with or accommodating legacy CLIs.
- It adds some distance between this feature and the `sui` CLI so that we can port it into jellyfish more easily.

Having said that, this is strictly better than what we had before, so I don't think that work should be a blocker now, and I don't see any problem with, e.g. MVR building source verification on top of this machinery as-is.

These considerations will become more relevant if we start seeing individuals running this command over many packages s.t. the weight of each archive download becomes noticeable (doubly so if we start to recommend this or build this as an automated step into the `mvr` CLI etc), or when we look into bringing this feature into `jellyfish` (or an extension thereof).

Separately, I noticed in a couple of places that you're using `Symbol` (this may also be vestigial from the old implementation) -- we should probably stop doing that.

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,265 @@
...
+            message: format!("could not locate the running executable: {e}"),
+        });
+    }
+
+    let platform = detect_platform(version)?;
+    let binary_name = if platform == "windows-x86_64" {
+        "sui.exe"
+    } else {
+        "sui"
+    };
```
stefan-mysten: I would probably use an enum for this (I got bitten once or twice in suiup with string names for this stuff) or `whoami`. 

For example

```rust
enum Platform {
    MacosArm64,
    MacosX86_64,
    UbuntuX86_64,
    WindowsX86_64,
}

impl Platform {
    /// The release-artifact platform string used in download URLs.
    fn artifact_str(&self) -> &'static str {
        match self {
            Platform::MacosArm64 => "macos-arm64",
            Platform::MacosX86_64 => "macos-x86_64",
            Platform::UbuntuX86_64 => "ubuntu-x86_64",
            Platform::WindowsX86_64 => "windows-x86_64",
        }
    }

    fn binary_name(&self) -> &'static str {
        match self {
            Platform::WindowsX86_64 => "sui.exe",
            _ => "sui",
        }
    }
}

fn detect_platform(version: &str) -> Result<Platform, Error> {
    match (std::env::consts::OS, std::env::consts::ARCH) {
        ("macos", "aarch64") => Ok(Platform::MacosArm64),
        ("macos", "x86_64") => Ok(Platform::MacosX86_64),
        ("linux", "x86_64") => Ok(Platform::UbuntuX86_64),
        ("windows", "x86_64") => Ok(Platform::WindowsX86_64),
        (os, arch) => Err(Error::BinaryDownload {
            version: version.to_string(),
            message: format!(
                "no downloadable sui {version} release for your platform \
                 (OS: {os}, architecture: {arch}); place a matching binary in \
                 {}/binaries/{version}/target/release/",
                &*MOVE_HOME,
            ),
        }),
    }
}
```

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,265 @@
...
+}
+
+#[cfg(not(unix))]
+fn set_executable_permission(path: &OsStr) -> anyhow::Result<()> {
+    use std::process::Command;
+    Command::new("icacls")
+        .args([path, OsStr::new("/grant"), OsStr::new("Everyone:(RX)")])
+        .status()?;
+    Ok(())
+}
```
stefan-mysten: Not too familiar how things are in Windows, but setting the permission to everyone is a bit scary, right?

crates/sui-source-verification/src/binary.rs
stefan-mysten: I mean, not a lot you can do here (both from our side where we can change things, or from GitHub if endpoints change) unless you want to use GitHub API (which also evolves).

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,265 @@
...
+        // Locate the `sui` binary in the extracted archive (its layout varies by release) and
+        // install it under the canonical cache path.
+        let found = find_binary(&extract, platform)
+            .ok_or_else(|| anyhow!("no sui binary found in the {version} release archive"))?;
+
+        let release_dir = tmp.join("target").join("release");
+        fs::create_dir_all(&release_dir).context("creating release directory")?;
+        let canonical = release_dir.join(binary_name);
+        fs::rename(&found, &canonical).context("installing extracted binary")?;
+        set_executable_permission(canonical.as_os_str())?;
```
stefan-mysten, line 93: Left a note on this function for Windows.

crates/sui-source-verification/src/binary.rs
stefan-mysten: Also, does this really need to change? Another question is why spawning another process rather than using `window-sys` or some other crate that can modify metadata of a file?

crates/sui-source-verification/src/build.rs
```diff
@@ -0,0 +1,147 @@
...
+struct DumpOutput {
+    modules: Vec<String>,
+    dependencies: Vec<ObjectID>,
+}
+
+/// Rebuild the package at `source_path` with `binary` and return its compiled modules and linkage
+/// dependencies via `--dump-bytecode-as-base64`.
+///
+/// The flag set is adapted to what `binary` supports (probed from its `--help`): older binaries lack
+/// `--no-tree-shaking` (their output is whatever they shake) and the `--build-env` selector.
```
stefan-mysten:
```suggestion
/// `--no-tree-shaking` (their output includes all dependencies) and the `--build-env` selector.
```

crates/sui-source-verification/DESIGN.md
stefan-mysten: Probably under `.sui` is fine. An alternative would be `XDG_CACHE_HOME`?

crates/sui-source-verification/Cargo.toml
```diff
@@ -15,9 +15,11 @@ path = "src/lib.rs"
 [dependencies]
```
stefan-mysten, line 15: Some dependencies can be removed.

```
[deps]
 mysten-common, move-bytecode-source-map, move-compiler, sui-package-managementd
[dev-deps]
expect-test, rand, sui-test-transaction-builder
```

.github/workflows/source-verification-matrix.yml
```diff
@@ -0,0 +1,75 @@
+name: Source Verification Version Matrix
```
stefan-mysten, line 1: This seems like a good idea, though it does not do any alerting if these checks fail. I assume we'd want to know if this happens, as I am certain it will fall through the cracks at some point and nobody will look at it.

stefan-mysten (commented):
Great work!

As we briefly talked, it could be useful to check first if `suiup` has any of the release required. It uses `XDG_HOME_CACHE/suiup`, where you'll find a `releases` folder with all .tgz files (e.g., `sui-testnet-v1.49.2-macos-arm64.tgz`).

Full path: `XDG_HOME_CACHE/suiup/releases/sui-testnet-v1.49.2-macos-arm64.tgz`

crates/sui-source-verification/DESIGN.md
mdgeorge4153: For published-at it seems like a big potential footgun because you could end up saying something that verifies but then link against something else. With the toolchain version the worst that can happen is you say it verified with the wrong compiler version. The former could lead to supply-chain vulnerabilities while the latter doesn't really cause any problems.

That said, I could see a case for the tool allowing it but the service not allowing it, but given that the whole point of source verification is checking that you match the on-chain bytecode, I think I prefer just preventing it altogether.

crates/sui-source-verification/DESIGN.md
mdgeorge4153: seems reasonable.

crates/sui-source-verification/DESIGN.md
mdgeorge4153: In the latter case, I agree, but the right thing to do is to put the on-chain ID in the published.toml so that you record what was actually checked. In the former, providing a different address is very dangerous, because you would say "aha, this source behaves the same as it does on-chain", and then when you publish the package system would pull the published address from the published.toml, and you'd link against something you didn't check.

crates/sui-source-verification/DESIGN.md
amnn: I'm not advocating for the service offering this because that would indeed be confusing (the source that you point to in the attestation would not point back), but for me that's a different constraint than the one to impose on the tool, where I think it's generally useful to offer overrides for things that it otherwise does automatically.

crates/sui-source-verification/DESIGN.md
mdgeorge4153: changed it to XDG_CACHE_HOME/sui/binaries with a fallback to MOVE_HOME.

mdgeorge4153: looks like I still need to update the doc though

crates/sui-source-verification/DESIGN.md
mdgeorge4153: I'm adding this in a later PR.

crates/sui-source-verification/DESIGN.md
mdgeorge4153: warning is added; I didn't bother with opt-in error for now.

crates/sui-source-verification/DESIGN.md
mdgeorge4153: updated.

crates/sui-source-verification/src/pinning.rs
mdgeorge4153: done

crates/sui-source-verification/src/onchain.rs
mdgeorge4153: fixed.

crates/sui-source-verification/src/binary.rs
```diff
@@ -0,0 +1,303 @@
...
+fn detect_platform(version: &str) -> Result<Platform, Error> {
+    match (std::env::consts::OS, std::env::consts::ARCH) {
+        ("macos", "aarch64") => Ok(Platform::MacosArm64),
+        ("macos", "x86_64") => Ok(Platform::MacosX86_64),
+        ("linux", "x86_64") => Ok(Platform::UbuntuX86_64),
+        ("windows", "x86_64") => Ok(Platform::WindowsX86_64),
+        (os, arch) => Err(Error::BinaryDownload {
+            version: version.to_string(),
+            message: format!(
+                "no downloadable sui {version} release for your platform \
```
mdgeorge4153, line 234: the advice here is incorrect. (a) it looks like we still fail even if that binary exists, and (b) we don't want to allow this unless you pass --toolchain (so that the returned metadata is correct). I'd remove the suggestion here, and fix it in the `--toolchain` PR.

crates/sui-source-verification/src/binary.rs
mdgeorge4153: done.

crates/sui-source-verification/src/binary.rs
mdgeorge4153: took it out; it was unnecessary anyway

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,218 @@
...
+every difference it found and exits non-zero.
+
+## What it verifies
+
+Only the **root** package is verified: the bytecode of its modules and its linkage table (the
+versions of its dependencies). The command does not check the source of the dependencies themselves.
+To verify a dependency, run `verify-source` again against that dependency's own source.
+
+The command performs no compilation itself. It resolves the toolchain version that published the
+package, downloads that release's `sui` binary if it is not already cached under `$MOVE_HOME/binaries`,
```
mdgeorge4153: it's not $MOVE_HOME anymore

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,218 @@
...
+and uses it to build the source. Because that same toolchain produced the onchain bytecode, a matching
+build is strong evidence that the source is authentic.
+
+:::info
+
+Verification compares **bytecode**, not source text. Anything that does not affect the compiled output
+(comments, formatting, and the bodies of `macro` functions that no caller invokes) can differ between
```
mdgeorge4153:
```suggestion
(comments, formatting, and **the bodies of `macro` functions**) can differ between
```

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,218 @@
...
+macros into their callers, so those affect the bytecode and verification covers them. A package that
+verifies has bytecode identical to the onchain version.
+
+:::
+
+## Making a package verifiable
+
+If you publish a package and want others to verify it out of the box:
+
+- **Pin every dependency to a commit hash**, not a branch or tag. The current package system does this
```
mdgeorge4153:
```suggestion
- **Ensure Move.lock has commit hashes**, not branches or tags. The current package system does this
```

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,218 @@
...
+(comments, formatting, and the bodies of `macro` functions that no caller invokes) can differ between
+the source and the published package and still verify. The compiler inlines the bodies of called
+macros into their callers, so those affect the bytecode and verification covers them. A package that
+verifies has bytecode identical to the onchain version.
+
+:::
+
+## Making a package verifiable
+
+If you publish a package and want others to verify it out of the box:
```
mdgeorge4153:
```suggestion
If you want to fix a published source package so that it verifies out of the box:
```

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,218 @@
...
+  necessarily the original publish commit: if you added a `Published.toml` or pinned dependencies to
+  make it verify (see below), tag the commit that carries those changes.
+- **Record the toolchain version in the publication.** Modern publishing writes it to `Published.toml`.
+  If yours does not (see below), you can add it.
+
+If a package's publication does not record its address or toolchain (for example, the package predates
+`Published.toml`, or you published it through an offline or multisig flow), you can add a
+`Published.toml` next to the manifest that records the onchain address and toolchain version for each
+environment. Older `sui` versions ignore this file, so its only effect is to let `verify-source`, and
+anyone depending on the package, find the metadata. Both the verifier and the package system prefer a
```
mdgeorge4153:
```suggestion
environment. Older `sui` versions ignore this file, so its only effect is to let `verify-source` and
anyone depending on the package find the metadata. Both the verifier and the package system prefer a
```

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,218 @@
...
+with one section per environment:
+
+```toml
+[published.mainnet]
+chain-id = "35834a8a"                          # the network's chain id
+published-at = "0x0e735f8c93a95722efd735..."   # the package's current onchain address
+original-id = "0x2c8d603bc51326b8c13cef..."    # its original, first-published address
+version = 8
+toolchain-version = "1.71.1"                   # the sui release that built it
+```
```
mdgeorge4153, line 89:
````suggestion
toolchain-version = "1.71.1"                   # the sui release that built it

[published.testnet]
...
```
````

File: docs/content/develop/manage-packages/source-verification.mdx [hunk 57]
```diff
@@ -0,0 +1,218 @@
...
+### The toolchain version is not recorded
+
+By default the toolchain version comes from the package's publish metadata. Some publications omit it,
+notably packages from before v1.23 and some multisig or offline flows. The command then stops and asks
+you to supply the version:
+
+```bash
+sui client verify-source <PACKAGE_PATH> --toolchain-version 1.71.1
+```
+
```
mdgeorge4153, line 144: Once you find a version that verifies, you can write it into Published.toml to make future reverification easier.

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,218 @@
...
+Failed to publish the Move module(s), reason: Package dependency "MyDep" does not specify a
+published address (the Move.toml manifest for "MyDep" does not contain a published-at field).
+```
+
+This happens when the pinned revision of a dependency is a plain library checkout whose manifest leaves
+its address unset (`my_dep = "_"`), even though the package it links against is onchain. The publisher
+usually supplied the address out of band, by depending on a network-specific branch of the dependency
+or on a local checkout.
+
+To rebuild, vendor the dependency: copy it at its pinned revision into your tree, add its onchain
```
mdgeorge4153: check me on this, but I think you can override the address in the old Move.toml format or use `dep-replacements.<env>.<dep> = { published-at = "...", original-id = "..." }` in the new package system instead of vendoring.

docs/content/develop/manage-packages/source-verification.mdx
```diff
@@ -0,0 +1,218 @@
...
+my_dep = "0x5306f64e..."
+```
+
+Supplying the address yourself does not weaken the check. Verification compares the root package's
+linkage table as well as its bytecode, so an address other than the one the package actually links
+against causes verification to fail.
+
+### The package cannot be rebuilt from its repository
+
+Some packages do not build directly from their committed manifest. For example, a build script might
```
mdgeorge4153, line 230: I'd soften this - you'd just have to set up the Move.toml by hand.

mdgeorge4153 (commented):
Some comments on the docs.

.github/workflows/source-verification-matrix.yml
mdgeorge4153: I added a notify step that does the same thing as our nightly builds.

crates/sui-source-verification/DESIGN.md
mdgeorge4153: I'm going to leave as-is for now, and we can discuss adding the feature in the future.

crates/sui-source-verification/DESIGN.md
mdgeorge4153: I'm going to leave as-is and we'll discuss adding this in the future.

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Added after the version-guessing paragraph: "Once you find a version that verifies, record it in the package's `Published.toml` (the `toolchain-version` field) so future verifications need no guessing."

docs/content/develop/manage-packages/source-verification.mdx
mdgeorge4153: [CLAUDE] Reworked the section to lead with supplying the address via `[dep-replacements.<env>]` (`MyDep = { published-at = "0x...", original-id = "0x..." }`), or the dependency's `published-at` in older manifests, with vendoring kept as the fallback for when you also need source changes. Documented as intended behavior per your note; any bug in applying it is a separate workstream.

PR #27222, opened 2026-07-09
[move-compiler] split sui from lint categories

damirka:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-compiler/src/linters/mod.rs
```diff
@@ -45,141 +46,183 @@ pub enum LinterDiagnosticCategory {
     Suspicious,
     Deprecated,
     Style,
-    Sui = 99,
 }
 
+impl Flavor {
```
tnowacki: Having impl blocks like there where the type isn't defined makes me feel weird inside

tnowacki: Like personally I would have just done `pub(crate) const fn lint_category_marker(flavor: Flavor) -> u8 {`

I personally think for impl blocks we try to stick to one of the following:
- The `impl` is in the same module as the declaration
- The `impl` is in the same module (or submodule) as the corresponding `trait` declaration

If you don't do this and then  ever try to split up these modules into separate crates, you might be very sad.

tnowacki: (see also https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules)

external-crates/move/crates/move-compiler/src/linters/mod.rs
```diff
@@ -45,141 +46,183 @@ pub enum LinterDiagnosticCategory {
...
+    /// the lint's source (e.g. `Lint W94002` is a Sui-flavor Style lint) while the shared
+    /// categories stay comparable across sources. `Core` marks the flavor-independent lints,
+    /// which run for every flavor and sit at the base of the space. Disjoint blocks under the
+    /// common `Lint` prefix let filters (and, eventually, configuration such as
+    /// `lints.sui.suspicious` vs `lints.suspicious`) target one source's category without the
+    /// other's.
+    pub(crate) const fn lint_category_marker(self) -> u8 {
+        match self {
+            Flavor::Core => 0,
+            Flavor::Sui => 90,
```
tnowacki: Can we just add a letter for `Sui`? if I understand correctly this limits the number of categories to 9 or 10?

external-crates/move/crates/move-compiler/src/linters/mod.rs
damirka: Thanks for pointing that out! I wasn't sure what's the right approach here and took a wrong path.

external-crates/move/crates/move-compiler/src/linters/mod.rs
damirka: That would mean codes being:

```
[WS0090] 
[ES0090]
```
...where S stands for Sui. While core warnings and errors would feature a `0` ? 

```
[W00090]
[E00090]
```

PR #27227, opened 2026-07-09
[jsonrpc-alt] Add `getProtocolConfig` and `devInspectExecuteTransactionBlock` apis

wlmyng:
## Description 

Add `getProtocolConfig` and `devInspectExecuteTransactionBlock` reading from `kv_epoch_starts` for relevant info (protocol config, ref gas price, and so on)

## Test plan 

new e2e tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_write_api_tests.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_write_api_tests.rs
@@ -30,9 +30,6 @@ use sui_indexer_alt_reader::kv_loader::KvArgs;
 use sui_pg_db::DbArgs;
 use sui_pg_db::temp::TempDb;
 use sui_pg_db::temp::get_available_port;
-use sui_rpc::field::FieldMask;
-use sui_rpc::field::FieldMaskUtil;
-use sui_rpc::proto::sui::rpc::v2 as proto;
 use sui_swarm_config::genesis_config::AccountConfig;
 use sui_types::MOVE_STDLIB_PACKAGE_ID;
 use sui_types::TypeTag;
@@ -135,7 +132,9 @@ impl WriteTestCluster {
         // checkpoint, so wait for it before handing the cluster to a test.
         tokio::time::timeout(Duration::from_secs(60), async {
             loop {
-                let response = cluster.execute_jsonrpc("sui_getProtocolConfig", json!([])).await;
+                let response = cluster
+                    .execute_jsonrpc("sui_getProtocolConfig", json!([]))
+                    .await;
                 if matches!(&response, Ok(r) if r["error"].is_null()) {
                     break;
                 }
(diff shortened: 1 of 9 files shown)


crates/sui-indexer-alt-jsonrpc/src/api/write/mod.rs
```diff
@@ -141,6 +186,86 @@ impl Write {
...
+            gas_objects.unwrap_or_default(),
+            gas_budget.map(|budget| *budget).unwrap_or(max_tx_gas),
+            gas_price.map(|price| *price).unwrap_or(reference_gas_price),
+            gas_sponsor.unwrap_or(sender_address),
+        );
+
+        // The raw transaction data reflects what the caller specified: the gas payment stays empty
+        // here, and the mock gas coin the fullnode injects during simulation only shows up in the
+        // effects (matching the legacy implementation, which captures these bytes before simulation
+        // for the same reason).
```
amnn, line 228: Thank you for calling this out, because I had the same question!

crates/sui-indexer-alt-jsonrpc/src/api/write/mod.rs
```diff
@@ -185,6 +310,52 @@ impl Write {
...
+    let chain = ctx
+        .chain_identifier()
+        .context("Chain identifier not available")?
+        .chain();
+
+    let config =
+        ProtocolConfig::get_for_version_if_supported((protocol_version as u64).into(), chain)
+            .with_context(|| {
+                format!("Protocol version {protocol_version} of the latest epoch is not supported")
+            })?;
```
amnn: Prefer to fetch `max_tx_gas` by joining against `kv_protocol_configs` so that we don't need to keep the binary strictly up-to-date with the validator software to keep running properly.

I did also consider whether it's worth storing this information in a cache like you did for validator APYs, but because we don't have watermarking here, you would need to fetch the latest epoch anyway, at which point you may just fetch the other stuff.

crates/sui-indexer-alt-jsonrpc/src/api/protocol.rs
```diff
@@ -0,0 +1,110 @@
...
+                ProtocolVersion::MIN.as_u64(),
+                ProtocolVersion::MAX.as_u64(),
+            ))
+        })?;
+
+    Ok(ProtocolConfigResponse::from(config))
+}
+
+/// Fetch the protocol version of the latest epoch from the database.
+async fn latest_protocol_version(ctx: &Context) -> Result<u64, RpcError<Error>> {
```
amnn, line 147: Same deal here, use `kv_protocol_configs` and `kv_feature_flags`.

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_write_api_tests.rs
```diff
@@ -637,3 +694,241 @@ async fn test_execute_and_dry_run_gas_costs_agree() {
...
+    );
+
+    let result = &response["result"];
+    assert_eq!(result["effects"]["status"]["status"], "success");
+    assert_eq!(result["results"].as_array().unwrap().len(), 1);
+}
+
+#[tokio::test]
+async fn test_dev_inspect_execution_failure() {
+    telemetry_subscribers::init_for_testing();
```
amnn: nit: skip telemetry subscriber initialisation by default -- we can add it when we need to get telemetry from the test, otherwise get spotty/inconsistent coverage on this and it gets confusing.

crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_write_api_tests.rs
```diff
@@ -637,3 +694,241 @@ async fn test_execute_and_dry_run_gas_costs_agree() {
...
+        .unwrap();
+
+    let result = &response["result"];
+    assert_eq!(result["effects"]["status"]["status"], "failure");
+    assert!(!result["error"].is_null(), "Expected an execution error");
+    assert!(result["results"].is_null());
+}
+
+#[tokio::test]
+async fn test_dev_inspect_matches_grpc_simulation() {
```
amnn: Shouldn't we be testing against fullnode JSONRPC, rather than gRPC which is what it is implemented using?

amnn (approved):
Thanks @wlmyng -- accepting to unblock, but would still recommend switching to fetching protocol configs from `kv_protocol_configs` and `kv_feature_flags` so that (a) we fetch this information from the same place in all RPCs, and (b) we maintain the loose coupling between RPCs and node software, so that they remain forward compatible for longer.

PR #27229, opened 2026-07-10
Fix bridge e2e flakiness reading ETH coin from owned-object index

lxfind:
## Description

The bridge e2e tests read the recipient's ETH coin with a single `get_owned_objects` call right after the transfer's onchain status becomes `Claimed`. The status is read from live object state, but the owned-object index only updates once the claiming transaction's checkpoint is indexed, so the coin may not be listed yet. This raced in CI: `test_v2_sui_with_v1_evm` failed all 4 retries and both `test_bridge_from_eth_to_sui_to_eth` variants went flaky 2/4 in [the same run](https://github.com/MystenLabs/sui/actions/runs/29061189538/job/86263842593).

Replaces the five single-shot reads with a helper that polls the index until the coin appears. The re-read in `test_v2_sui_with_v1_evm` excludes the coin from the earlier transfer, which was already bridged away but could still be listed by a lagging index.

## Test plan

Test-only change; the five affected e2e tests exercise the helper directly, covering both the coin-not-yet-indexed case (poll until visible) and the stale-previous-coin case (exclusion in `test_v2_sui_with_v1_evm`, where the Test 1 coin was bridged away before the Test 3 re-read).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


Bridgerz (approved):
Lgtm

PR #27232, opened 2026-07-10
[wip] allowances

manolisliolios:
## Description 

## Test plan 

- Tests in this PR for the move code
- Adapter tests here: https://github.com/MystenLabs/sui/pull/27254
- todo: needs workload tests, more edge case tests, more adapter tests etc

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,335 @@
...
+    /// Always `Some` today: the sign-time DoS gate needs a concrete signer.
+    /// `Option` so app-bound allowances can later go keyless.
+    spender: Option<address>,
+    /// When set, only the app's module can spend and rotate; the signer path
+    /// is disabled and `spender` is just the sign-time gate.
+    app: Option<TypeName>,
+    /// `None` = no lifetime total; at least one of cap / rate limit must be
+    /// set. Amounts are `u256` (matching `Withdrawal.limit`); times are ms.
+    lifetime_cap: Option<u256>,
+    current_spend: u256,
```
mystenmark, line 82: can you document this field?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,335 @@
...
+
+/// Revocation for an allowance, sent to the funder at issuance (key-only, non-transferrable).
+/// Also used for discoverability (funder -> allowances)
+public struct AllowanceCap<phantom T> has key {
+    id: UID,
+    allowance: ID,
+}
+
+/// A tumbling per-window cap: at most `limit` per `period_ms`, resetting each
+/// window.
```
mystenmark: we should have some plans for how to evolve this. For instance, it can't express monthly payments since months are not constant times.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,335 @@
...
+    self.lifetime_cap.do_ref!(|lifetime_cap| {
+        assert!(self.current_spend + amount <= *lifetime_cap, EExceedsLifetimeCap);
+    });
+
+    self.current_spend = self.current_spend + amount;
+
+    self.rate_limit.do_mut!(|rl| {
+        // Tumbling window: reset once the period has elapsed.
+        if (now >= rl.window_start_ms + rl.period_ms) {
+            rl.window_start_ms = now;
```
mystenmark: should window start become window_start_ms + period_ms? otherwise if the spender is late to withdraw, the window gets shifted

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,335 @@
...
+    name: String,
+    spender: address,
+    app: Option<TypeName>,
+    lifetime_cap: Option<u256>,
+    start_timestamp_ms: Option<u64>,
+    expiration_timestamp_ms: Option<u64>,
+    rate_limit: Option<RateLimit>,
+    ctx: &mut TxContext,
+) {
+    // we do not allow unlimited allowances (TODO: Do we?)
```
mystenmark, line 373: this is a funny comment

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub fn is_allowance(s: &StructTag) -> bool {
+        s.address == SUI_FRAMEWORK_ADDRESS
+            && s.module.as_ident_str() == ALLOWANCE_MODULE_NAME
+            && s.name.as_ident_str() == ALLOWANCE_STRUCT_NAME
+            && s.type_params.len() == 1
+    }
+}
+
+/// Sign-time view of an `Allowance<T>`. The spender can rotate, so never
+/// reuse a resolution across transactions.
```
mystenmark, line 89: why do we need a separate view for sign-time? we can't just use the BCS mirror?

crates/sui-types/src/transaction.rs
```diff
@@ -3114,6 +3146,82 @@ impl TransactionDataAPI for TransactionDataV1 {
...
+        }
+        // One pass over the inputs; each allowance below is a map lookup. The
+        // allowance must be among the tx's inputs (spending needs it as an
+        // input anyway), or it fails to resolve here.
+        let objects_by_id: BTreeMap<ObjectID, &Object> = input_objects
+            .iter()
+            .filter_map(|input| Some((input.id(), input.as_object()?)))
+            .collect();
+        // Parses are cached per object, but every withdrawal is checked: two
+        // may source the same allowance with different declared funders.
```
mystenmark, line 3165: what? how would that work?

manolisliolios: You could declare (when building the tx) 
`AllowanceWithdrawal { funder: @0xa, allowance: @0x1 }` 
and 
`AllowanceWithdrawal { funder: @0xb, allowance: @0x1 }`

Which would be caught here & fail

crates/sui-types/src/allowance.rs
manolisliolios: It adds the `funds_type`, as well as the parsed (best-effort) `spend_limit` to check against during signing. The latter could be provided from a helper fn, but I think the type arg is not part of the bcs data?

crates/sui-types/src/transaction.rs
```diff
@@ -3114,6 +3146,82 @@ impl TransactionDataAPI for TransactionDataV1 {
...
+            *total = total.checked_add(U256::from(amount)).unwrap_or(U256::max_value());
+            // Over the spend limit can never clear; reject
+            if *total > resolved.spend_limit {
+                return Err(UserInputError::InvalidWithdrawReservation {
+                    error: format!(
+                        "Reservations against allowance {allowance} exceed its spend limit {}",
+                        resolved.spend_limit
+                    ),
+                });
+            }
```
mystenmark: i'm not sure whether it makes sense to do any checks here other than `spender == sender`. All of these checks get repeated in Move (right??), and relying on signing time checks may mask issues. For instance, signing can clear two concurrent transactions that will cumulatively exceed the spend limit, but normally in testing we might not send concurrent transactions. So, if there's any problem in the move checks, we might not reach it during testing.

crates/sui-types/src/transaction.rs
```diff
@@ -3114,6 +3146,82 @@ impl TransactionDataAPI for TransactionDataV1 {
...
+            };
+            if !resolved_allowances.contains_key(allowance) {
+                let object = objects_by_id.get(allowance).ok_or_else(|| {
+                    UserInputError::InvalidWithdrawReservation {
+                        error: format!("allowance object {allowance} not found"),
+                    }
+                })?;
+                resolved_allowances.insert(*allowance, parse_allowance_object(object)?);
+            }
+            let resolved = &resolved_allowances[allowance];
```
mystenmark: you can avoid multiple lookups:

```
let resolved = match resolved_allowances.entry(*allowance) {
    Entry::Occupied(entry) => entry.into_mut(),
    Entry::Vacant(entry) => {
        let object = objects_by_id.get(allowance).ok_or_else(|| {
            UserInputError::InvalidWithdrawReservation {
                error: format!("allowance object {allowance} not found"),
            }
        })?;
        entry.insert(parse_allowance_object(object)?)
    }
};
```

crates/sui-types/src/transaction.rs
manolisliolios: Yes that's right (the checks and enforcement lives in move). 

The only reason I added this check is to minimize the reservation amount per tx for the **spender**, based on the best-effort calculated maximum. It's a guard that protects against reservations against the entire balance of the funder. 

Given the check is almost free, don't you think it's good to have?

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: if we did `window_start_ms + period_ms`, that'd enable the spender to always be able to withdraw. That feels more like a scheduler design instead of a throttler.

We could introduce scheduling logic, but maybe we could scope this for a future version (this also touches on the comment around monthly payments above). If we were to do that, we could introduce it through DFs or type evolution over a new type version (so we should be OK to introduce without core changes).

Update on this: Ended up doing correct fixed windows & monthly/yearly rate limit options for better built-in support.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: This is an enum now. Discussed with core, we're OK adding enum variants in the framework if we need to!

manolisliolios:
Closing for #27528 stack

PR #27238, opened 2026-07-10
add opt-in gRPC request capture on the grpc_request tracing target

evan-wall-mysten:
## Description

gRPC counterpart to the `graphql_request` capture that motivated #27026: there is currently no way to collect per-request logs from the gRPC servers — they record only method/latency/status for metrics, never the request message.

Adds a `GrpcRequestLogLayer` tower middleware to `mysten-network` that tees each gRPC request body, decodes messages with `prost-reflect` using the same `FileDescriptorSet`s the server registers for reflection, and emits one `trace`-level event per message on the `grpc_request` target with the payload rendered as canonical protobuf JSON — collectable via `RUST_LOG_TAILS` as an NDJSON replay corpus. Wired into `sui-rpc-api` (under grpc-web translation, so browser-client traffic is captured too), `sui-kv-rpc`, and the consistent store, one commit each.

Off by default; opt in with e.g.:

```
RUST_LOG="info,grpc_request=trace" RUST_LOG_TAILS="grpc_request=/var/log/sui/grpc_request.jsonl"
```

`service`/`method` are recorded on an enclosing span, so `RUST_LOG='info,grpc_request[{service=sui.rpc.v2.LedgerService}]=trace'` subsets a single service (under `telemetry-subscribers` this additionally requires `TOKIO_SPAN_LEVEL=trace`). Captured payloads are capped at 512 KiB per message and 1 message per request; oversized, compressed, or undecodable frames emit payload-less events instead.

## Test plan

Unit tests cover the frame parser (chunk boundaries, oversized/compressed/undecodable frames, per-request truncation), `EnvFilter` interaction (plain and span-field directives, including under a `telemetry-subscribers`-style global span-level cap), grpc-web layer ordering in `sui-rpc-api`, and that each server's registered descriptor sets build one pool.

Verified on a localnet: `RUST_LOG="info,grpc_request=trace" RUST_LOG_TAILS="grpc_request=/tmp/grpc.jsonl" sui start --force-regenesis`, then `grpcurl -plaintext -d '{"object_id":"0x5"}' localhost:9000 sui.rpc.v2.LedgerService.GetObject` and a grpc-web POST both land in `/tmp/grpc.jsonl` as NDJSON; plain `RUST_LOG=info` captures nothing; `grpc_request[{service=grpc.health.v1.Health}]=trace` with `TOKIO_SPAN_LEVEL=trace` captures only that service.

---

## Release notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Full nodes (and `sui-kv-rpc`) can capture decoded gRPC request payloads on the new opt-in `grpc_request` tracing target: set `RUST_LOG="info,grpc_request=trace"`, optionally with `RUST_LOG_TAILS="grpc_request=/path/file.jsonl"`, to log each request's method and canonical-JSON payload. Unset by default, so log output is unchanged and no action is required.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [x] Indexing Framework: The consistent store's gRPC service supports the same opt-in `grpc_request` capture target.


crates/mysten-network/src/request_log/body.rs
```diff
@@ -0,0 +1,375 @@
...
+            // event.
+            let _span = self.span.enter();
+            match event {
+                FrameEvent::Message(message) => {
+                    match DynamicMessage::decode(self.input.clone(), message.as_slice()) {
+                        Ok(message) => trace!(
+                            target: TARGET,
+                            method = %self.path,
+                            message_index = self.message_index,
+                            payload = %serde_json::to_string(&message).unwrap_or_default(),
```
nickvikeras: IIUC this is transcoding the binary serialized protobuf to ProtoJSON. Did you consider just storing the binary protobuf request message directly as a base64 encoded string instead? I think that would be substantially cheaper if we don't need to be human-readable in the logs.

crates/mysten-network/src/request_log/body.rs
```diff
@@ -0,0 +1,375 @@
...
+use super::TARGET;
+
+/// Length of the gRPC frame prefix: a 1-byte compressed flag and a 4-byte big-endian message
+/// length.
+const FRAME_PREFIX_LEN: usize = 5;
+
+/// Per-request cap on captured messages. Requests are unary on the servers this wraps (only
+/// reflection streams legitimately carry more), so this exists to stop a hostile body packed with
+/// tiny frames from amplifying into unbounded decode and log volume.
+const MAX_CAPTURED_MESSAGES: u32 = 64;
```
nickvikeras: Does this ever need to be more than 1 if we don't have client-streaming APIs? I think for our unary and server-streaming APIs there is always exactly one request message. Just calling it out cuz I think combined with the json-transcoding above and our large message size limits, this could use a substantial amount memory if an adversarial client tried to DoS a gRPC server.

evan-wall-mysten: No, changed to 1 and updated the comment.

crates/mysten-network/src/request_log/body.rs
evan-wall-mysten: ATM there isn't any use case to be able to read the JSON so I'll switch it. The benchmark code will need a protobuf library anyways.

nickvikeras (approved):

crates/mysten-network/src/request_log/body.rs
```diff
@@ -0,0 +1,363 @@
...
+pub(crate) struct CaptureState {
+    span: Span,
+    path: String,
+    max_captured_message_size: usize,
+    parser: FrameParser,
+    message_index: u32,
+}
+
+/// Incremental parser for the gRPC message framing (a sequence of `FRAME_PREFIX_LEN`-byte
+/// prefixes, each followed by that many message bytes). Chunk-boundary-agnostic: one chunk may
```
wlmyng: "each a `FRAME_PREFIX_LEN-byte prefix followed by that message's bytes"

crates/mysten-network/src/request_log/body.rs
```diff
@@ -0,0 +1,363 @@
...
+    /// Parsing is over for this request — the framing was structurally invalid, or the
+    /// per-request capture limit was reached. The rest of the stream passes through unparsed.
+    Stopped,
+}
+
+/// A frame boundary [`FrameParser::feed`] reached.
+enum FrameEvent {
+    /// A complete, uncompressed message within the size cap.
+    Message(Vec<u8>),
+    /// A frame whose bytes were counted but not buffered.
```
wlmyng: "message" instead of "frame"

crates/mysten-network/src/request_log/body.rs
```diff
@@ -0,0 +1,363 @@
...
+                let n = (FRAME_PREFIX_LEN - *filled).min(chunk.len());
+                buf[*filled..*filled + n].copy_from_slice(&chunk[..n]);
+                *filled += n;
+                if *filled < FRAME_PREFIX_LEN {
+                    return (n, None);
+                }
+
+                let flag = buf[0];
+                let len = u32::from_be_bytes(buf[1..].try_into().unwrap()) as usize;
+                if flag > 1 {
```
wlmyng, line 177: do we want to emit a marker here anyways? Or is the idea to treat as if the request never existed

crates/mysten-network/src/request_log/body.rs
```diff
@@ -0,0 +1,363 @@
...
+                        n,
+                        Some(FrameEvent::Skipped {
+                            len,
+                            reason: "message too large",
+                        }),
+                    );
+                }
+                if len == 0 {
+                    *self = Self::new();
+                    return (n, Some(FrameEvent::Message(Vec::new())));
```
wlmyng, line 209: is it possible for me to send `N` requests each claiming a 512 KiB message but otherwise nothing further/ rest of the messages really slowly, causing the server to hold equivalent number of empty buffers?

crates/mysten-network/src/request_log/body.rs
```diff
@@ -0,0 +1,363 @@
...
+            };
+
+            // Enter the span so `EnvFilter` span-field directives (service/method) apply to the
+            // event.
+            let _span = self.span.enter();
+
+            // Checked before handling `event`, not after, so this only fires when a message
+            // beyond the cap actually arrives — a body with exactly `MAX_CAPTURED_MESSAGES`
+            // messages and nothing more must not end in a spurious "too many messages" event.
+            if self.message_index >= MAX_CAPTURED_MESSAGES {
```
wlmyng: it seems if I send a body containing two frames, the following occurs:

1. `self.parser.feed` matches on the `Prefix` arm, transitions to `Message`, returns `None` event, so we just loop
2. `feed` hits the `Message` arm, returns an event payload, `message_index += 1`. We know we don't need to capture any more for this request, but we can't prove that excess frames exist
3. `feed` matches `Prefix` arm, inside the arm, we extraneously allocate `Vec::with_capacity`, as part of transitioning to `Message`. 
4. At some point we make it past L100, and finally L111 fires

This means that a second frame of 5 bytes could buy up additional held memory before we reject it. I think the `MAX_CAPTURED_MESSAGES` check in `observe` essentially runs when a frame completes. When we have processed the excess frame's prefix, we already know that we have too many messages in the body, sot he cap could be enforced there, before the `Prefix` arm allocates for a message.

wlmyng (commented):
Only managed to review `body.rs` as I have a huge void where grpc knowlege is supposed to be in my head ... but had some questions around how we're allocating for a `Message` payload

crates/mysten-network/src/request_log/body.rs
evan-wall-mysten: It means that the flag byte value is invalid. I'll change it to emit a `FrameEvent::Skipped` with a reason to differentiate it.

crates/mysten-network/src/request_log/body.rs
evan-wall-mysten: I changed it to lazily grow the buffer here, but nothing forces the client to eventually send the bytes specified by the message size.

Tonic will reserve 4MB by default so this is not much worse: https://github.com/hyperium/tonic/blob/6cb6056b5a748bc5a29bd48f4602dbc4e552bb7d/tonic/src/codec/decode.rs#L189-L199

crates/mysten-network/src/request_log/body.rs
evan-wall-mysten: Good catch, moved the check above the allocation.

PR #27241, opened 2026-07-10
[source-verification 1] Read a package's publication without loading its graph

mdgeorge4153:
## Description

`Package::load` reads the manifest and resolves the publication for an environment — modern pubfile first, then a legacy lockfile — before it resolves dependencies. Factor that into a helper and expose a public `read_publication(dir, env, flavor)` that stops there, so a caller can learn a package's recorded addresses without loading the dependency graph and without reimplementing the modern/legacy precedence.

Base of a stack: source verification (#27213) uses this to resolve the address it verifies against the same way the package system resolves it when linking against the package, so the two cannot disagree.

## Test plan

Unit tests over `read_publication`: it returns the recorded addresses, `None` when unpublished, and — when a package records its publication in more than one place — prefers the `Published.toml` address over both a legacy manifest `published-at` and a legacy `Move.lock` `[env]` address (so the verifier and the package system, which share this resolution, cannot disagree). The refactor leaves `Package::load` calling the extracted helper, so existing tests cover it.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-package-alt/src/package/package_impl.rs
+++ b/external-crates/move/crates/move-package-alt/src/package/package_impl.rs
@@ -2,7 +2,7 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::BTreeMap, sync::Arc};
+use std::{collections::BTreeMap, path::Path, sync::Arc};
 
 use derive_where::derive_where;
 use sha2::{Digest as _, Sha256};
@@ -66,6 +66,25 @@ pub struct Package<F: MoveFlavor> {
     pub dummy_addr: OriginalID,
 }
 
+/// Read the publication recorded for `env` at the package rooted in `dir`, consulting the modern
+/// pubfile and then a legacy lockfile, without loading the dependency graph. Returns `None` if the
+/// package records no publication for `env`.
+///
+/// This resolves the publication the same way [`Package::load`] does, so a caller learns a package's
+/// published address exactly as the package system would resolve it when linking against that
+/// package, without paying for a full graph load.
+pub async fn read_publication<F: MoveFlavor>(
+    dir: &Path,
+    env: &Environment,
+    flavor: &F,
+) -> PackageResult<Option<Publication<F>>> {
+    let path = PackagePath::new(dir.to_path_buf())?;
+    let mtx = path.lock()?;
+    let (_, _, publication) =
+        Package::<F>::read_manifest_and_publication(&path, env, true, &mtx, flavor).await?;
+    Ok(publication)
+}
+
 impl<F: MoveFlavor> Package<F> {
     /// Fetch `dep` (relative to `self`) and load a package from the fetched source.
     /// Makes a best effort to translate old-style packages into the current format.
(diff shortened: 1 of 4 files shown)


external-crates/move/crates/move-package-alt/src/package/package_impl.rs
```diff
@@ -589,6 +624,126 @@ mod tests {
 
     /// Create a basic package and then call cache_package on a local dependency to it; check that
     /// the returned fields are correct
```
stefan-mysten: These two comment lines need to be moved to the now below test that used to be here before the change.

external-crates/move/crates/move-package-alt/src/package/package_impl.rs
```diff
@@ -66,6 +66,25 @@ pub struct Package<F: MoveFlavor> {
...
+/// package, without paying for a full graph load.
+pub async fn read_publication<F: MoveFlavor>(
+    dir: &Path,
+    env: &Environment,
+    flavor: &F,
+) -> PackageResult<Option<Publication<F>>> {
+    let path = PackagePath::new(dir.to_path_buf())?;
+    let mtx = path.lock()?;
+    let (_, _, publication) =
+        Package::<F>::read_manifest_and_publication(&path, env, true, &mtx, flavor).await?;
```
stefan-mysten: Why are we passing `true` to this function? Claude tells me this:
```
 So reading the publication of a legacy package that explicitly depends on Sui/MoveStdlib will print unsolicited advice about a package the caller doesn't own.
```

external-crates/move/crates/move-package-alt/src/package/package_impl.rs
```diff
@@ -66,6 +66,25 @@ pub struct Package<F: MoveFlavor> {
...
+/// This resolves the publication the same way [`Package::load`] does, so a caller learns a package's
+/// published address exactly as the package system would resolve it when linking against that
+/// package, without paying for a full graph load.
+pub async fn read_publication<F: MoveFlavor>(
+    dir: &Path,
+    env: &Environment,
+    flavor: &F,
+) -> PackageResult<Option<Publication<F>>> {
+    let path = PackagePath::new(dir.to_path_buf())?;
+    let mtx = path.lock()?;
```
stefan-mysten, line 91: This does manual lock, instead of taking a PackageSystemLock like in other parts of the code. Could this lead to a deadlock? Do we even need to lock given that we're only reading?

stefan-mysten: Oh I see this is actual a PackageSystemLock...

mdgeorge4153: you still need to hold locks while reading since you can have read-write conflicts, although I guess in this case they would be different files. Nevertheless, the package system requires you to hold a lock. I don't think there's risk of deadlock, since there's no hold-and-wait.

external-crates/move/crates/move-package-alt/src/package/package_impl.rs
mdgeorge4153: I think this analysis comes from claude's habit of treating the implementation as the interface. Arguably this /is/ the root package, since we're reading the manifest. I'll change the name of that argument to `display_warnings` and then make it `false`.

external-crates/move/crates/move-package-alt/src/package/package_impl.rs
stefan-mysten: So this is safe to call only when the caller does not hold already a package's lock. I assume within the package system itself does not happen, but it could happen in places where this is externally used (i.e. source verification).

mdgeorge4153: This isn't a new concern - you would see the same issue if you acquired two RootPackages for the same directory. Fixing it would require some engineering to expose the lock to the outside world (so that callers could share the lock between the RootPackage and this function) or make the locks reentrant. I'm just going to document the restriction I think.

stefan-mysten: Documenting is good enough!

stefan-mysten (approved):

PR #27246, opened 2026-07-10
[cli] Add faucet retry for finding gas coins

### LINKED ISSUE #27240: sui start --with-faucet --force-regenesis: local faucet intermittently fails to start ("No address found with sufficient coins") and takes down the whole node
## Summary

`sui start --with-faucet --force-regenesis` intermittently fails to start the faucet, logging:

```
Wallet Error: `No address found with sufficient coins`
```

The faucet never binds `:9123`. Because the faucet is built with `?` *before* the node health loop, the error propagates and terminates the **entire `sui start` process** (fullnode included) within ~2s of "Cluster started" — so depending on timing it presents as either "faucet never came up" or "RPC never came up." It is a startup race, so it is intermittent, and it reproduces reliably on hosts with slow disk `fsync` (e.g. Docker Desktop's Linux VM, or CI on slow/contended storage).

## Affected code (current `main`)

- `crates/sui-faucet/src/local_faucet.rs` — `find_gas_coins_and_address()` scans the wallet's addresses **once** for a gas coin `>= config.amount`; if none is found at that instant it returns `FaucetError::Wallet("No address found with sufficient coins")`. No retry, no wait.
- `crates/sui/src/sui_commands.rs` — the faucet is built during startup via `LocalFaucet::new(...).await?`, **before** the node health loop, so that error tears down the whole process.

The faucet's source address is the genesis funding account, which **is** funded — so this is purely a timing race against genesis state becoming readable, not a missing or underfunded account.

## Root cause (verified)

The faucet's coin check fires ~7 ms after "Cluster started" and reads gas objects through the wallet's node client. On a fresh `--force-regenesis` network the genesis coin objects are created but must be **persisted and made readable** before that query returns them. When persistence/readability lags the ~7 ms check, the scan returns empty and the faucet fails hard.

The race window is set by how quickly genesis state becomes durable and readable, so **storage speed is the dominant factor**. Changing only the genesis store's backing storage (same binary, same host) shifts the failure rate substantially:

- On a slow / virtualized disk (e.g. Docker Desktop's Linux VM), failures are common (tens of percent of boots).
- Moving the store to RAM (tmpfs) sharply reduces them (roughly 5× fewer in testing, though not to zero under heavy concurrency).
- On a fast local disk the check almost always wins, even under many concurrent boots.

So this reproduces on slow or contended storage and is rare on fast disks — but it is a latent one-shot race regardless, and a slow-`fsync` or busy host can hit it.

### Timeline of a failing boot (`RUST_LOG=off,sui_faucet=debug,sui::sui_commands=info`)

```
T+0.000  INFO sui::sui_commands: Cluster started
T+0.007  INFO sui_faucet::server: Initialize wallet from config path: "/tmp/.tmpXXXX/client.yaml"
T+0.007  Wallet Error: `No address found with sufficient coins`   # one-shot check; process exits ~1-2s later
```

Passing boots follow the identical sequence but the check happens to see the coins and logs `Starting local faucet service on 0.0.0.0:9123` / `Faucet started`.

## Reproduction

Boots the localnet N times and counts faucet-bind failures. It reproduces where disk `fsync` is slow or contended — a Docker Desktop Linux VM, CI runners, or many concurrent boots; a fast idle local disk may show 0:

```bash
fails=0
for i in $(seq 1 20); do
  RUST_LOG=off sui start --with-faucet --force-regenesis >"/tmp/sui.$i.log" 2>&1 &
  pid=$!
  ok=0
  for _ in $(seq 1 30); do
    kill -0 "$pid" 2>/dev/null || break          # process self-terminated
    case "$(curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:9123)" in
      200|404|405) ok=1; break ;;
    esac
    sleep 0.5
  done
  kill "$pid" 2>/dev/null; wait "$pid" 2>/dev/null
  if [ "$ok" = 1 ]; then echo "boot $i: OK"; else
    fails=$((fails+1)); echo "boot $i: FAIL — $(grep -m1 'No address found' "/tmp/sui.$i.log")"
  fi
done
echo "faucet failed to start on $fails/20 boots"
```

To confirm the cause directly, re-run with the genesis store on RAM and watch the failure rate drop sharply — e.g. in a container, `docker run --tmpfs /tmp ...` (the store is created under the temp dir).

## Suggested fix

Two independent improvements (either helps; both is ideal):

1. **Retry the lookup (targeted).** Poll `find_gas_coins_and_address()` for a bounded time before erroring — genesis coins are guaranteed to appear shortly. A ~15-line change; patch below.
2. **Don't fail-hard the whole node (defense-in-depth).** A local faucet that can't initialize shouldn't tear down `sui start`. Constructing/spawning the faucet after the node health loop, or logging-and-continuing, would keep the fullnode up even if the faucet can't start.

### Patch for (1) — `crates/sui-faucet/src/local_faucet.rs`

Wraps the existing scan in a bounded poll loop (default 20 s / 250 ms) and tolerates transient query errors. Preserves the original error message on timeout. (Attached as a separate diff.)

## Workaround (for anyone hitting this today)

**Supervise `sui start` and relaunch on the fast self-termination** — reliable, since the failure is transient:

```bash
for attempt in $(seq 1 20); do
  RUST_LOG=off,sui_node=info sui start --with-faucet --force-regenesis &
  pid=$!
  while kill -0 "$pid" 2>/dev/null; do
    case "$(curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:9123 2>/dev/null)" in
      200|404|405) echo "ready on attempt $attempt"; break 2 ;;
    esac
    sleep 1
  done
  wait "$pid" 2>/dev/null   # it self-terminated; try again
done
```

Optionally, **RAM-back the genesis store** (`--tmpfs /tmp` in a container) to cut the failure rate as well — but it does not fully eliminate the race under load, so keep the retry as the reliable mechanism.

stefan-mysten:
## Description 

On slower hardware (e.g., in a Docker container), the faucet fails to find the gas coins at startup, when used with `sui start`. This should fix it and closes #27240.

## Test plan 

Existing tests. Spinned up a local container and with some hdd contention to be able to reproduce the issue.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-faucet/src/local_faucet.rs
+++ b/crates/sui-faucet/src/local_faucet.rs
@@ -28,6 +28,12 @@ use sui_sdk::wallet_context::WalletContext;
 const GAS_BUDGET: u64 = 10_000_000;
 const NUM_RETRIES: u8 = 2;
 
+/// Number of times to retry the gas-coin scan before giving up. On a freshly created
+/// `--force-regenesis` network the genesis coin objects may not yet be persisted and
+/// readable the instant the cluster reports started — especially on slow or contended
+/// storage — so we retry the scan a bounded number of times before failing.
+const GAS_COIN_LOOKUP_RETRIES: u8 = 3;
+
 pub struct LocalFaucet {
     wallet: WalletContext,
     active_address: SuiAddress,
@@ -176,10 +182,42 @@ impl LocalFaucet {
 /// Finds gas coins with sufficient balance and returns the address to use as the active address
 /// for the faucet. If the initial active address in the wallet does not have enough gas coins,
 /// it will iterate through the addresses to find one with sufficient gas coins.
+///
+/// Retries the scan a bounded number of times (see [`GAS_COIN_LOOKUP_RETRIES`]) so that a
+/// transient startup race — where genesis coins are not yet readable — does not fail the faucet.
 async fn find_gas_coins_and_address(
     wallet: &mut WalletContext,
     config: &FaucetConfig,
 ) -> Result<(Vec<GasCoin>, SuiAddress), FaucetError> {
+    let mut retry_delay = Duration::from_millis(500);
+    // Preserved and returned if every attempt comes up empty, so the caller sees the original
+    // "no coins" message (or the last transient query error) rather than a generic timeout.
+    let mut last_err =
+        FaucetError::Wallet("No address found with sufficient coins".to_string());
+
+    for attempt in 0..=GAS_COIN_LOOKUP_RETRIES {
+        match scan_for_gas_coins(wallet, config).await {
+            Ok(Some(found)) => return Ok(found),
+            Ok(None) => {}
+            Err(e) => last_err = e,
+        }
+
+        if attempt < GAS_COIN_LOOKUP_RETRIES {
+            tokio::time::sleep(retry_delay).await;
+            retry_delay *= 2;
+        }
+    }
+
+    Err(last_err)
+}
+
+/// Scans the wallet's addresses once for a gas coin with a balance of at least `config.amount`,
+/// returning the matching coins and the address that holds them, or `Ok(None)` if no such coin is
+/// currently visible.
+async fn scan_for_gas_coins(
+    wallet: &mut WalletContext,
+    config: &FaucetConfig,
+) -> Result<Option<(Vec<GasCoin>, SuiAddress)>, FaucetError> {
     let active_address = wallet
         .active_address()
         .map_err(|e| FaucetError::Wallet(e.to_string()))?;
@@ -200,13 +238,11 @@ async fn find_gas_coins_and_address(
             .collect();
 
         if !coins.is_empty() {
-            return Ok((coins, address));
+            return Ok(Some((coins, address)));
         }
     }
 
-    Err(FaucetError::Wallet(
-        "No address found with sufficient coins".to_string(),
-    ))
+    Ok(None)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


crates/sui-faucet/src/local_faucet.rs
```diff
@@ -176,10 +182,41 @@ impl LocalFaucet {
...
+    let mut last_err = FaucetError::Wallet("No address found with sufficient coins".to_string());
+
+    for attempt in 0..=GAS_COIN_LOOKUP_RETRIES {
+        match scan_for_gas_coins(wallet, config).await {
+            Ok(Some(found)) => return Ok(found),
+            Ok(None) => {}
+            Err(e) => last_err = e,
+        }
+
+        if attempt < GAS_COIN_LOOKUP_RETRIES {
```
amnn: nit (optional): In other places where we need this kind of functionality, we use the `backoff` crate. Granted, that crate is unmaintained (the recommendation is to use `backon` instead), but we use it everywhere anyway, and so it may just be easier to use it for consistency so that when we do migrate off it across the mono-repo, we spot that this is a place where we are performing an exponential backoff too. Those crates do some nice things like correctly account for the time it takes to perform the action, and adding jitter when adding delays between retries.

amnn (approved):

jordanjennings-mysten (approved):

crates/sui-faucet/src/local_faucet.rs
stefan-mysten: Good point, will move to using that.

PR #27248, opened 2026-07-11
Do not floor transaction lookups at the object-pruning watermark

lxfind:
## Description

Root cause of the `fuzz_dynamic_committee` simtest flake ("Transaction executed but checkpoint wait timed out", [CI failure](https://github.com/MystenLabs/sui/actions/runs/29061189525/job/86263843868)); replaces #27230, which retried the wait and could not have fixed this.

`GetTransaction` refuses to serve a transaction whose checkpoint is below the lowest-available watermark. Since #25397 that floor includes the object-pruning watermark (and on the embedded rpc-store path, the index floor the object pruner drags along with it — the object pruner prunes the rpc-store history cohort in lockstep). With default object pruning (`num_epochs_to_retain: 0`; in simtest the pruner runs every 2s), the floor trails the executed tip by only a couple of checkpoints — so a transaction checkpointed moments ago is reported `NotFound` while all of its data is still present.

This breaks the already-checkpointed fast path in the SDK's `execute_transaction_and_wait_for_checkpoint`: the fast path exists precisely for the case where the transaction's checkpoint streamed past before the client subscribed, and its spurious `NotFound` leaves the client scanning future checkpoints for a digest that will never appear — a guaranteed 30s timeout on a healthy, fully-caught-up node. Reproduced deterministically with `MSIM_TEST_SEED=1783726865041` / `...047`: the fatal lookup was rejected with `checkpoint 89 outside available range 91..=91`.

The fix keeps the two existing watermarks in their documented roles instead of maxing them together for every read: the transaction lookup is bounded by transaction/checkpoint retention (`ReadStore::get_lowest_available_checkpoint`, restored to that meaning on `RpcStoreReadStore`), and the object bound is checked where object contents are actually read (rendering `balance_changes`/`effects`, which also converts an internal fetch error into `NotFound` for `ListTransactions` on a pruned range). The rpc-store index floor already rides in `get_lowest_available_checkpoint_objects`, so the combined watermark served in response headers and used by checkpoint reads is unchanged. The `StateReader` wrapper method that computes that combined value is renamed to `get_lowest_available_checkpoint_combined` so it can no longer be mistaken for the store-level transaction floor of the same name.

## Test plan

Reproduced the flake deterministically with two seeds found via `seed-search.py` (`MSIM_TEST_SEED=1783726865041` and `1783726865047`, `cargo simtest fuzz_dynamic_committee -p sui-e2e-tests`); both fail on main at the exact CI panic site and pass with this change. Instrumented runs confirm the mechanism: on main the fatal `get_transaction` is rejected by the object-pruning floor (`89 outside 91..=91`), with the fix the transaction floor is checkpoint retention (0) and only benign upper-bound races remain (tx-to-checkpoint mapping momentarily ahead of latest-executed, which the checkpoint stream covers).

## Release notes

- [ ] Protocol:
- [x] gRPC: `GetTransaction`/`BatchGetTransactions` no longer report a transaction as not found when only historic object versions below it have been pruned; the object-pruning floor now applies only to requests that render object contents (`balance_changes`/`effects`).
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/storage.rs
+++ b/crates/sui-core/src/storage.rs
@@ -689,12 +689,13 @@ impl ReadStore for RpcStoreReadStore {
     }
 
     fn get_lowest_available_checkpoint(&self) -> Result<CheckpointSequenceNumber> {
-        // A consistent read needs both the raw chain data (perpetual
-        // store) and the index rows (rpc-store), so the available range
-        // starts at the higher of the two lower bounds.
-        let perpetual = self.rocks.get_lowest_available_checkpoint()?;
-        let rpc_store = self.reader.get_lowest_available_checkpoint()?;
-        Ok(perpetual.max(rpc_store))
+        // Every transaction and checkpoint read on this store is served from the raw chain
+        // stores, so only their retention bounds it. The rpc-store's index floor is pruned in
+        // lockstep with objects and is carried by `get_lowest_available_checkpoint_objects`;
+        // folding it in here would floor transaction lookups at the object-pruning watermark,
+        // which under aggressive object pruning (`num_epochs_to_retain: 0`) rides within a
+        // couple of checkpoints of the executed tip.
+        self.rocks.get_lowest_available_checkpoint()
     }
 
     fn get_checkpoint_by_digest(&self, digest: &CheckpointDigest) -> Option<VerifiedCheckpoint> {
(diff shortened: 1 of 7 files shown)


nickvikeras:
I am not sure if we want this fix as-is, because we could have a request where we can load the requested transaction but not its objects. I suppose maybe the compromise would be to only clamp to the object pruner's watermark if objects are requested in the read mask. 

Edit - I submitted a new PR with a 2nd commit - https://github.com/MystenLabs/sui/pull/27267

We can just close this one.

PR #27250, opened 2026-07-11
Refactor gRPC APIs onto stable proto service definition

nickvikeras:
## Description 

Refactored onto the stable proto service definition. Emitting the end reason, final watermark, and (maybe) last item in a single frame required a surprising amount of refactoring to make everything consistent across both implementations. 

Now (in the happy path) the last frame will always have the end reason and a terminal watermark. If the requested ended because `limit_items` was hit, it will also have the last item. If the scan limit is hit, it's possible that the final watermark is emitted twice - once a standalone frame, and again in the final frame. 

The evaluator now produces a "scan frontier" when the scan limit budget is hit, and that propagates all through the pipelines to the grpc response. I also consolidated the logic to hydrate the watermarks with checkpoint sequence numbers so it's all done in the same place. 

I updated the documentation in the proto to explain more precisely how the APIs work. https://github.com/MystenLabs/sui-rust-sdk/pull/279

## Test plan 

All the unit tests, plus a bunch of new tests. I'm also gonna re-run my test suite I generated from snowflake that I ran on the last iteration of these APIs.


bmwill:
Two issues to look at:

> 1. Watermark hydration is no longer degradable — this is the cause of the failing CI (major)
> 
>   CI's test (ubuntu-ghcloud) job fails with two tests panicking on Internal: missing checkpoint mapping for transaction sequence
>   number N (kv_rpc_tests.rs:397 and :447). Root cause, verified in the code:
> 
>   - bigtable_client.rs:350-362 — tx_seq_checkpoint now returns Err(Internal) when the tx_seq_digest row is missing; pre-PR it
>   returned Ok(None) and the resolver simply suppressed the un-resolvable boundary watermark. The two resolver closures
>   (tx_wm_resolver, event_wm_resolver) map this to ScanStop::Fault, faulting the whole stream.
>   - The miss is deterministic, not a race: boundary watermarks can legitimately sit at the exclusive upper edge of the scanned tx
>   range (checkpoint_to_tx_range maps end_checkpoint to the first tx of the next checkpoint), and that edge tx has no row when the
>   range ends at/past the ledger tip or an un-created checkpoint. Descending lookups hit this most directly since they look up
>   position itself.
>   - Same non-degradability appears in pipeline.rs:1113 and :1134: a failed checkpoint lookup on the terminal frontier converts a
>   successful budget-limited response into Internal after items were already streamed, where the old code kept the ScanLimit
>   outcome with a last-good cursor.
> 
>   Suggested fix: restore tx_seq_checkpoint to Result<Option<u64>> and let the resolvers propagate Ok(None). The handlers already
>   tolerate checkpoint = None at numeric edges via scan_frontier_cursor_cp, so the terminal-ScanLimit path won't regress.
> 
>   2. list_checkpoints can re-deliver a checkpoint on ScanLimit resume (major)
> 
>   In sui-rpc-api/src/grpc/v2alpha/ledger_service/list_checkpoints.rs:531, the terminal frontier watermark is now emitted whenever
>   request_scan_limit_reached — including when the chunk emitted checkpoints (the old code emitted it only on empty chunks). The
>   frontier is a tx sequence number that checkpoint_frontier_watermark coarsens to Boundary(C) in checkpoint space, and ascending
>   boundary cursors resume inclusively at C (boundary_cursor_cp, watermark.rs:130). So if the scan budget dies mid-checkpoint after
>   checkpoint C was already emitted (e.g., matches at tx 10 and tx 50, both in cp 3; budget exhausts at frontier tx 40), the
>   client resumes from Boundary(cp 3) and receives cp 3 a second time. The frame also reports a covered bound of C-1 after an item
>   frame claimed C. This is unique to list_checkpoints because its output granularity (checkpoint) is coarser than its frontier
>   granularity (tx); transactions and events carry exact positions. The added tests exercise checkpoint_frontier_watermark only on
>   the empty-result path, so this case isn't covered.

Also the `Cargo.lock` update seems to pick up a bunch of downgrades in unrelated deps.

bmwill (approved):

nickvikeras:
> Two issues to look at:
> 
> > 1. Watermark hydration is no longer degradable — this is the cause of the failing CI (major)

This was intentional. In production I'd never expect the checkpoint to be missing from the tx_digest index table.

> > 2. list_checkpoints can re-deliver a checkpoint on ScanLimit resume (major)

This should be fixed by `e33e9eb918abc8f15f1ef665c21b403bb8ecd621`. I am guessing you had reviewed an earlier version of the code, cuz I fixed that later.

PR #27251, opened 2026-07-12
Stop reading live_owned_object_markers table

andll:
## Description

Step 1 of a two-step removal of the `live_owned_object_markers` perpetual table (a legacy liveness index; see #27243 for the full context). This PR stops all reads — `check_owned_objects_are_live` and the test-only `get_lock` are reimplemented on the latest-object read path (`get_object_impl`), which answers exact-ref liveness from the objects table — while **keeping the writes**, so a rollback to a previous binary still finds a correctly maintained table. The field is renamed with a `deprecated_` prefix (the on-disk CF name is unchanged via `#[rename]`).

Step 2 (#27243) removes the writes and the table itself once this version is deployed everywhere.

## Test plan

Covered by existing `sui-core` tests; the writeback-cache and object-lock suites exercise both rewritten readers.

## Release notes

Check each box that your changes affect. If none apply, ignore this section. When releasing, delete unchecked boxes.

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/execution_cache/writeback_cache.rs
+++ b/crates/sui-core/src/execution_cache/writeback_cache.rs
@@ -46,7 +46,7 @@ use crate::authority::authority_store::{
 use crate::authority::authority_store_tables::LiveObject;
 use crate::authority::backpressure::BackpressureManager;
 use crate::authority::epoch_start_configuration::{EpochFlag, EpochStartConfiguration};
-use crate::fallback_fetch::{do_fallback_lookup, do_fallback_lookup_fallible};
+use crate::fallback_fetch::do_fallback_lookup;
 use crate::global_state_hasher::GlobalStateHashStore;
 use crate::transaction_outputs::TransactionOutputs;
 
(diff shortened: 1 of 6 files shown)


crates/sui-core/src/execution_cache/writeback_cache.rs
```diff
@@ -1949,33 +1929,22 @@ impl ObjectCacheRead for WritebackCache {
     }
 
     fn check_owned_objects_are_live(&self, owned_object_refs: &[ObjectRef]) -> SuiResult {
```
mystenmark: we could also delete this entire function - its really just a paranoid check at this point that can never (and has never) failed as far as i know

andll: Done

mystenmark (approved):

PR #27253, opened 2026-07-12
sui-rpc-api: add a request handler latency metric

bmwill:
## Description 

See commit messages for a description of this change

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


amnn (approved):
Change makes sense, thanks!

Previously we had issues related to the feature-based CryptoProvider selection for some of the RPC and indexing binaries.

The issues did not show up at diff-time, only as a panic during deploy because release binaries are built together with a merged feature set.

Just flagging this in case you think the changes related to `rustls` configuration might affect that, but I don't see changes to the lines we added to configure that explicitly (#25301)

bmwill:
Yeah if you noticed there is a section in sui-node that explicitly picks the provider so that this isn't an issue

amnn:
Got it -- so this would only have been an issue for codebases that previously were not explicitly setting this, and everything has now standardised on that approach (tangentially, it makes me quite sad that rustls operates in this way...)

PR #27261, opened 2026-07-13
sui-core: reuse a per-epoch VM instance for transaction simulation

bmwill:
Previously, simulate_transaction created a brand new Move VM on every call. This was a workaround for issues with the old VM implementation that no longer apply.

With this commit, the per-epoch store now holds a dedicated simulate executor in ExecutionComponents, created alongside the normal execution VM and therefore refreshed at the same cadence, on epoch boundaries. All simulate calls (including the insufficient-object-funds retry) now reuse this instance. The simulate VM is kept separate from the normal execution VM so that simulating arbitrary, unvalidated transactions cannot pollute the internal VM state (e.g. package caches) of the executor used for committed transaction execution.

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -2472,12 +2472,7 @@ impl AuthorityState {
             )?
         };
 
-        // TODO see if we can spin up a VM once and reuse it
-        let executor = sui_execution::executor(
-            protocol_config,
-            true, // silent
-        )
-        .expect("Creating an executor should not fail here");
+        let executor = epoch_store.simulate_executor();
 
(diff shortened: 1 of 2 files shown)


amnn (commented):
This looks good to me, but I'd like to get @tzakian's take on whether we need to do anything else at all to prevent publishes/upgrades from tainting the shared cache with the new VM.

tzakian (approved):
👍🏻

PR #27266, opened 2026-07-13
[source-verification] add --verify-only bytecode comparison mode

mdgeorge4153:
## Description

Adds a hidden `--verify-only` mode to `sui client verify-source`, for tooling (the debugger) that
wants to check an existing build against an on-chain package without rebuilding.

`sui client verify-source --verify-only <package-path> <package-id>` compares the modules already
compiled under `<package-path>/build` against the on-chain package, skipping the compiler, toolchain
resolution, and publication-metadata lookup. Module bytecode only, not linkage. The caller passes the
id explicitly — this is a local-build-versus-chain diff, not an authenticity check on recorded source,
so there is no metadata to disagree with. The flag is hidden; its consumer is other tooling.

Stacked on #27213.

## Test plan

Unit test for reading the compiled modules out of the build directory; the shell test gains a
`--verify-only` case that compares the already-built package against its published id.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/docs/content/develop/manage-packages/source-verification.mdx
+++ b/docs/content/develop/manage-packages/source-verification.mdx
@@ -1,0 +1,154 @@
+---
+title: Source Verification
+sidebar_label: Source Verification
+description: "Verify that a Move source package compiles to the bytecode of an on-chain package using sui client verify-source."
+keywords: [source verification, verify source, verify-source, verify package, reproducible build, bytecode verification, toolchain version]
+---
+
+Source verification checks that a Move source package compiles to the bytecode and linkage of a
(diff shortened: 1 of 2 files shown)


mdgeorge4153:
[CLAUDE] This PR was auto-closed by GitHub during a force-push that briefly left the branch with no commits ahead of its base; it can't be reopened. Continued as #27349 on the same branch.

PR #27267, opened 2026-07-13
Do not floor transaction lookups at the object-pruning watermark

nickvikeras:
## Description

Slight tweak to <https://github.com/MystenLabs/sui/pull/27248> to tighten up the logic on when to require clamping to object pruning floor.

### DIFF
--- a/crates/sui-core/src/storage.rs
+++ b/crates/sui-core/src/storage.rs
@@ -689,12 +689,13 @@ impl ReadStore for RpcStoreReadStore {
     }
 
     fn get_lowest_available_checkpoint(&self) -> Result<CheckpointSequenceNumber> {
-        // A consistent read needs both the raw chain data (perpetual
-        // store) and the index rows (rpc-store), so the available range
-        // starts at the higher of the two lower bounds.
-        let perpetual = self.rocks.get_lowest_available_checkpoint()?;
-        let rpc_store = self.reader.get_lowest_available_checkpoint()?;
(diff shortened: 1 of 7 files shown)


crates/sui-core/src/storage.rs
```diff
@@ -689,12 +689,13 @@ impl ReadStore for RpcStoreReadStore {
...
-        let perpetual = self.rocks.get_lowest_available_checkpoint()?;
-        let rpc_store = self.reader.get_lowest_available_checkpoint()?;
-        Ok(perpetual.max(rpc_store))
+        // Every transaction and checkpoint read on this store is served from the raw chain
+        // stores, so only their retention bounds it. The rpc-store's index floor is pruned in
+        // lockstep with objects and is carried by `get_lowest_available_checkpoint_objects`;
+        // folding it in here would floor transaction lookups at the object-pruning watermark,
+        // which under aggressive object pruning (`num_epochs_to_retain: 0`) rides within a
+        // couple of checkpoints of the executed tip.
+        self.rocks.get_lowest_available_checkpoint()
```
bmwill, line 698: If we make this change we also probably want to better advertise the available range in our headers as well as the GetServiceInfo, I had change things such that we only advertised a single range taking into account both lower bounds (so take the max) to hopefully avoid confusion.

nickvikeras: Let's discuss this more in our 1:1 tomorrow

PR #27272, opened 2026-07-13
[Stress] disable duplicate/amplified submissions by default in bench CLI

arun-koshy:
## Description 

#27127 enabled submission amplification by default in the stress `bench` CLI (amplification probability 0.05, duplicate probability 0.02). Amplification requires local validator execution, but Antithesis runs stress with `--use-fullnode-for-execution true` (`docker/stress/entrypoint.sh`), so workload configuration bails at startup and the stress container crash-loops without generating any traffic. This is why nearly all reachability assertions have gone unseen in Antithesis reports since the merge (confirmed by a green run with #27127 reverted: https://mysten-labs.slack.com/archives/C07KB2X10TU/p1783966658688209).

Default both probabilities to 0.0 so the feature is opt-in. Simtests are unaffected — they enable it explicitly via `SubmissionAmplification::default_enabled()` and use the local validator proxy, so coverage of duplicate/amplified traffic is retained there.

## Test plan 

Covered by existing tests; `test_simulated_load_basic` continues to exercise amplification via `default_enabled()`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

---
_Generated by [Claude Code](https://claude.ai/code/session_01TRaYtB9qt2LqEe38eu6UNK)_

### DIFF
--- a/crates/sui-benchmark/src/drivers/mod.rs
+++ b/crates/sui-benchmark/src/drivers/mod.rs
@@ -52,8 +52,10 @@ impl Default for SubmissionAmplification {
 }
 
 impl SubmissionAmplification {
-    // Default values shared between the stress CLI options and simtests, so both
-    // exercise the same duplicate/amplified submission behavior out of the box.
+    // Default values used when duplicate/amplified submissions are explicitly enabled
+    // (e.g. via `default_enabled()` in simtests). The stress CLI defaults to disabled
+    // (zero probabilities), since amplification requires local validator execution and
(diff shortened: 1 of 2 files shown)


arun-koshy:
Closing in favor of #27271, which makes the same change.

PR #27274, opened 2026-07-13
[move-stackless-bytecode-2] fix dead snapshot-test harness

awelc:
## Description

`tests/tests.rs` drives the `modules.txt`-based snapshot directories via `datatest_stable::harness!`, but has no `[[test]] … harness = false` entry in `Cargo.toml`, so it compiles under the default libtest harness, discovers zero tests, and passes vacuously. As a result the `binary_ops/` and `structs/` snapshots were never asserted. Fold those two directories into the live `from_source` harness, and delete `tests.rs` (an older near-duplicate of `from_source.rs`) along with the now-redundant `modules.txt` manifests and the orphaned `*@sources.*.sbir.snap` files (no current suffix can produce them).

Refreshing the two revived directories' snapshots surfaced drift that accumulated while they were dead: display improvements (register types, struct names in `Unpack`), and one behavioral fix that had landed unobserved — `unpack_bar`'s multi-value `Return` order was corrected (the stale snapshot recorded `Return(reg_2, reg_1)` for `(val, other)`). The refreshed snapshots match current, correct translator output; binop operand order (`Divide(a, b)` etc.) is unchanged.

## Test plan

The revived directories now run and fail on drift: `cargo nextest run -p move-stackless-bytecode-2` runs 10 snapshot tests (previously 8 live).

## Release notes

Check each box that your changes affect. If none apply, ignore this section. For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates.

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


awelc:
Withdrawn.

PR #23777, opened 2025-09-29
[move][move-decompiler][5/n] Fix argument ordering

cgswords:
## Description 

The stackless bytecode was largely mis-generating code by putting arguments in the wrong order. This PR fixes that by modifying the stackless translation to do the right thing.

It also fixes some some type information that was being mis-handled.

## Test plan 

Several new tests, plus revisions to tests across the board with hand verification.

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:

PR #27277, opened 2026-07-14
[move-model] Stackless bytecode test fixes

awelc:
## Description 

Testing infrastructure for stackless bytecode does not seem quite up to date with some changes to the code that have been made in the past and this PR attempts to clean it up.

In particular, one testing mode seems to be defunct (the one implemented in `test.rs`) and its functions subsumed by `from_source.rs` and `from_bytecode.rs`. Most tests previously ran in `test` are now `from_source` so their old snapshots and `module.txt` files that were used by `test` are deleted.

The exception are tests for `structs` and `binary_ops` which have been "promoted" from `test` to `from_source`, which triggered update of their (stale) snapshots to the new output format and to reflect argument ordering [fix](https://github.com/MystenLabs/sui/pull/23777).

Another subtlety is related to what happened to `counter_constant` tests whose snapshots were also deleted. They don't have the exact equivalence in `from_source` due to renaming that happened at some point (and their functionality is now subsumed by `constant_constant` tests)

## Test plan 

Snapshots have been re-generated. All tests must pass


cgswords (approved):
LGTM, thanks.

PR #27281, opened 2026-07-14
Leap Frog Bitmap Joins

nickvikeras:
## Description

Implementing leap-frogging for the bitmap merge joins.

This was "one-shot" by planning with Fable and implementing with 5.6 Sol subagents. I just wrote some of the comments.

## Test plan

I'm gonna delay merging this until I can find a good test candidate query and verify the improvement empirically.

Edit - It works

```
 ┌─────────┬────────────┬───────────┬───────────┬─────────┐
 │ Latency │ Before     │ After     │ Reduction │ Speedup │
 ├─────────┼────────────┼───────────┼───────────┼─────────┤
 │ Mean    │ 68.558 ms  │ 30.655 ms │ 55.3%     │ 2.24×   │
 ├─────────┼────────────┼───────────┼───────────┼─────────┤
 │ p50     │ 64.826 ms  │ 28.495 ms │ 56.0%     │ 2.27×   │
 ├─────────┼────────────┼───────────┼───────────┼─────────┤
 │ p90     │ 87.561 ms  │ 36.480 ms │ 58.3%     │ 2.40×   │
 ├─────────┼────────────┼───────────┼───────────┼─────────┤
 │ p95     │ 100.753 ms │ 44.059 ms │ 56.3%     │ 2.29×   │
 ├─────────┼────────────┼───────────┼───────────┼─────────┤
 │ p99     │ 117.230 ms │ 69.726 ms │ 40.5%     │ 1.68×   │
 ├─────────┼────────────┼───────────┼───────────┼─────────┤
 │ Maximum │ 312.033 ms │ 96.962 ms │ 68.9%     │ 3.22×   │
 └─────────┴────────────┴───────────┴───────────┴─────────┘
 ```

### CODE REVIEW

File: crates/sui-inverted-index/src/bitmap_query/iter.rs [hunk 1]
```diff
@@ -104,32 +139,25 @@ where
...
                     }
-                    // Budget exhaustion: leave `front[i]` at its last scanned
-                    // position so the resume cursor cannot claim past it.
                     Some(Err(_)) => class[i] = Some(LeafHead::Error),
                 }
             }
 
-            // An include at EOF makes its term unsatisfiable (the intersection
-            // is permanently empty). With dedup, an EOF'd leaf may be an
-            // include for several terms; all of them become unsatisfiable.
```
bmwill, line 132: Is there a reason these comments got nuked?

bmwill (approved):

crates/sui-inverted-index/src/bitmap_query/iter.rs
nickvikeras: No good reason. I was just mostly reading the code in normal view (not diff view) and didn't notice.

crates/sui-kv-rpc/src/config.rs
```diff
@@ -217,6 +220,14 @@ pub struct LedgerHistoryConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub bitmap_bucket_budget_event: Option<u64>,
 
+    /// Dead bucket rows drained from an open leaf scan per lag episode before
+    /// abandoning the stream and reopening it past the gap. A fresh reopen is
+    /// disabled by `0`; provably dead gaps still advance the resume cursor.
+    ///
+    /// Defaults to `50` if not specified.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub bitmap_drain_probe_rows: Option<u32>,
```
amnn, line 229: My understanding is that this controls how many successively calls to `next` the execution strategy will tolerate before calling `seek`, and the rationale is that there can be a cost to calling `seek`.

But this extra cost to calling `seek` is highly dependent on the particular seekable stream implementation you're dealing with: For a synthetic stream it'll be completely free, for RocksDB iterators it's also basically free. The only implementation where this isn't the case is the implementation based on BigTable streams, right?

So in that case, does it make sense to make this an implementation detail of those implementations, rather than a configuration parameter? I.e. you could make it so that the seek implementation for the BigTable-backed stream tries to satisfy the seek by some successive calls to `next()` initially.

amnn (commented):
Haven't had a chance to look at the full implementation yet, but just sharing an initial thought.

crates/sui-inverted-index/src/bitmap_query/iter.rs
```diff
@@ -18,30 +18,74 @@
...
+        match self.peeked.take() {
+            Some(item) => item,
+            None => self.iter.next(),
+        }
+    }
+}
+
+impl<I: super::SeekableBucketIterator> Leaf<I> {
+    fn seek_bucket(&mut self, bucket: u64) {
+        if matches!(self.peeked.as_ref(), Some(Some(Ok(_)))) {
```
amnn, line 81: What happens if the bucket being seeked is the one that was already peeked?

crates/sui-inverted-index/src/bitmap_query/iter.rs
amnn, line 10: This comment looks like it needs to be updated to account for leap-frogging -- we can now end up peeking more than one bucket ahead. Similar comment applies to the streaming implementation also.

crates/sui-inverted-index/src/bitmap_query/iter.rs
```diff
@@ -104,32 +147,29 @@ where
...
             for i in 0..leaf_count {
                 if unreferenced[i] {
                     continue;
                 }
                 match leaves[i].peek() {
                     Some(Ok((bucket, _))) => {
-                        let (pre, _post) = bucket_edges(*bucket, bucket_size, &range, direction);
-                        front[i] = pre;
+                        let (pre, _) = bucket_edges(*bucket, bucket_size, &range, direction);
+                        front[i] = advance_in_direction(front[i], pre, direction);
```
amnn, line 160: What was the motivation for this change? I couldn't figure out why we needed to guard the write in this way.

amnn: Same question for the leap-frogging operation later on (why do we need `advance_in_direction`).

crates/sui-inverted-index/src/bitmap_query/iter.rs
```diff
@@ -18,30 +18,74 @@
...
 use super::eval_term_at_bucket;
 use super::frontier_advanced;
+use super::leaf_skip_targets;
 use super::recompute_unreferenced;
+use super::strictly_before;
 use super::take_snapshot_bitmap;
 
+struct Leaf<I> {
+    iter: I,
+    peeked: Option<Option<BucketItem>>,
```
amnn, line 51: nit: I got a little stuck wondering why we needed this over `Peekable`, I can see below that it's because peek needs to interact with seek, but maybe we should document that (need access to the underlying iterator to be able to `seek` it).

crates/sui-inverted-index/src/bitmap_query/iter.rs
amnn: I think the answer is that invariants that are handled elsewhere prevent this situation from happening, but that sounds a bit hairy. Seems worth either being defensive for that case in case that invariant is not maintained in future, asserting it or at least documenting it.

PR #27291, opened 2026-07-14
[cli] Rework gas selection in the CLI

stefan-mysten:
## Description 

This PR reworks the gas selection in the CLI to allow for address balances to be used.

The CLI now lets the fullnode select gas, matching what the TS SDK does. When a gas coin is not provided with `--gas`, all of the Coin<SUI> objects are used as the gas payment, and the network smashes them into a single coin. There are two consequences here:
- Available SUI coins for that address get consolidated. After any command where the CLI picks gas for you, expect to hold one SUI coin rather than several. If you rely on holding multiple separate coins (for example, to submit transactions in parallel), name the coin to use with --gas, which opts out of gas selection entirely and leaves your other coins untouched.
- If you hold a SUI address balance and the transaction uses the gas coin, your coins are absorbed into that address balance. They stop existing as coin objects, and gas is charged against the balance.

In exchange, gas selection now works in cases where it previously failed: an address whose SUI is spread across many small coins no longer hits "Cannot find gas coin ... sufficient for the required gas budget", and an address holding only an address balance — with no coins at all — can now run any command.

This PR fixes also the `sui client send-funds --all-coins`. It previously always failed with `InsufficientCoinBalance` when trying to smash your coins and send them to the AB. It now sends every Coin<SUI> object you own to the recipient's address balance, less the gas it costs to do so. Your own address balance is not touched — only your coin objects are used. Use `--from-address-balance` to send funds from your address balance instead.

## Test plan 

Existing tests and new tests to cover for sending funds from all coins to AB, paying for gas with AB, paying for gas with coins, etc.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [x] CLI: new logic for gas selection to allow usage of address balances. Note that if no gas coin is passed, it will smash all available gas coins into one.

Breaking change: `--stateless` is no longer available for `send-funds` command. Use `--from-address-balance` instead to send funds from AB to another address.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/src/client_commands.rs
+++ b/crates/sui/src/client_commands.rs
@@ -128,6 +128,18 @@ pub(crate) static USER_AGENT: &str =
 /// Only to be used within CLI
 pub const GAS_SAFE_OVERHEAD: u64 = 1000;
 
+/// The protocol limits that bound how many coins a single transaction can drain.
+struct CoinLimits {
+    /// Maximum number of objects a gas payment may contain.
+    max_gas_payment_objects: usize,
+    /// Maximum number of arguments a single command may take. A command's target does not count
+    /// towards this, and the limit is exclusive.
+    max_arguments: usize,
+    /// Maximum number of object inputs a transaction may have. The gas payment does not count
+    /// towards this.
+    max_input_objects: usize,
+}
+
 #[derive(Parser)]
 #[clap(rename_all = "kebab-case")]
 pub enum SuiClientCommands {
@@ -3420,18 +3432,6 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
             ),
         },
     )
-}
-
-/// The protocol limits that bound how many coins a single transaction can drain.
-struct CoinLimits {
-    /// Maximum number of objects a gas payment may contain.
-    max_gas_payment_objects: usize,
-    /// Maximum number of arguments a single command may take. A command's target does not count
-    /// towards this, and the limit is exclusive.
-    max_arguments: usize,
-    /// Maximum number of object inputs a transaction may have. The gas payment does not count
-    /// towards this.
-    max_input_objects: usize,
 }
 
 /// Queries the protocol config for the limits that bound how many coins fit in one transaction.
--- a/crates/sui/src/client_commands.rs
+++ b/crates/sui/src/client_commands.rs
@@ -1369,13 +1369,6 @@ impl SuiClientCommands {
 
                 let is_sui = coin_type_tag == sui_type_tag;
 
-                let TypeTag::Struct(coin_struct_tag) = &coin_type_tag else {
-                    bail!("coin type must be a struct type, got {coin_type_tag}");
-                };
-                let balance_info = client.get_balance(signer, coin_struct_tag).await?;
-                let coin_balance = balance_info.coin_balance();
-                let address_balance = balance_info.address_balance();
-
                 if all_coins {
                     return send_all_coins(
                         context,
@@ -1387,6 +1380,13 @@ impl SuiClientCommands {
                     )
                     .await;
                 }
+
+                let TypeTag::Struct(coin_struct_tag) = &coin_type_tag else {
+                    bail!("coin type must be a struct type, got {coin_type_tag}");
+                };
+                let balance_info = client.get_balance(signer, coin_struct_tag).await?;
+                let coin_balance = balance_info.coin_balance();
+                let address_balance = balance_info.address_balance();
 
                 let (amount, use_address_balance) = if let Some(amount) = amount {
                     let use_address_balance = if stateless {
@@ -3437,28 +3437,6 @@ async fn max_gas_payment_objects(client: &Client) -> Result<usize, anyhow::Error
         })
 }
 
-/// Every `Coin<T>` object owned by `owner`, paired with its value.
-async fn owned_coins(
-    client: &Client,
-    owner: SuiAddress,
-    coin_type_tag: &TypeTag,
-) -> Result<Vec<(ObjectRef, u64)>, anyhow::Error> {
-    let objects: Vec<Object> = client
-        .list_owned_objects(owner, Some(Coin::type_(coin_type_tag.clone())))
-        .try_collect()
-        .await?;
-
-    objects
-        .iter()
-        .map(|object| {
-            let coin = object
-                .as_coin_maybe()
-                .ok_or_else(|| anyhow!("Object {} is not a Coin<{coin_type_tag}>", object.id()))?;
-            Ok((object.compute_object_reference(), coin.value()))
-        })
-        .collect()
-}
-
 /// Send every `Coin<T>` object owned by `signer` to `recipient`'s address balance. The signer's
 /// own address balance is left untouched: only their coin objects are drained.
 async fn send_all_coins(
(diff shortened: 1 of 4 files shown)


crates/sui/src/client_commands.rs
```diff
@@ -3416,6 +3422,126 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
...
+        .ok_or_else(|| {
+            anyhow!(
+                "Could not find the maximum number of gas payment objects in the protocol config."
+            )
+        })
+}
+
+/// Send every `Coin<T>` object owned by `signer` to `recipient`'s address balance. The signer's
+/// own address balance is left untouched: only their coin objects are drained.
+async fn send_all_coins(
```
manolisliolios, line 3474: why do we need that one? was this pre-supported? And even so, can't grpc simulation fill this in with the same code path?

manolisliolios: Ah I see why we need (it's in the PR desc 😆). But the grpc simulation question stands!

stefan-mysten: Before, it would take the coins amount and try to split the gas coin to pay for gas, and then it would leave some dust. Essentially, you could not transfer all your coins into the address balance, and more importantly, it would simply fail because it wouldn't have gas to pay (because the gas coin was used as well). 

This function makes sure all the coins will be used for gas payment (I think the current limit is 256 coins for gas), and the grpc server will smash that into one coin. `send_funds` will use that coin to move the balance into the address balance and no dust will be left.

crates/sui/src/client_commands.rs
```diff
@@ -3416,6 +3422,126 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
...
+    let cfg = client.get_protocol_config(None).await?;
+    cfg.attributes()
+        .get("max_gas_payment_objects")
+        .and_then(|s| s.parse::<usize>().ok())
+        .ok_or_else(|| {
+            anyhow!(
+                "Could not find the maximum number of gas payment objects in the protocol config."
+            )
+        })
+}
```
amnn:
```suggestion
        .and_then(|s| s.parse().ok())
        .context(
            "Could not find the maximum number of gas payment objects in the protocol config."
        )
}
```

crates/sui/src/client_commands.rs
```diff
@@ -3416,6 +3422,126 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
...
+    signer: SuiAddress,
+    recipient: SuiAddress,
+    coin_type_tag: TypeTag,
+    gas_data: GasDataArgs,
+    processing: TxProcessingArgs,
+) -> Result<SuiClientCommandResult, anyhow::Error> {
+    let client = context.grpc_client()?;
+    ensure!(
+        coin_type_tag == GAS::type_tag(),
+        "--all-coins is only supported for SUI. Use --amount with --stateless to send another \
```
amnn: It would be surprising to me as a user that `--all-coins` only works for `SUI`. Why can't somebody ask to send all of their USDC and have the CLI figure out based on what kind of coin it is whether they need to pay for gas with the coin being sent or not?

crates/sui/src/client_commands.rs
```diff
@@ -3416,6 +3422,126 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
...
+
+    // Every coin has to fit in a single gas payment, so when the signer holds more than that, drain
+    // as many as will fit and let them run the command again for the rest.
+    let max_payment = max_gas_payment_objects(&client).await?;
+    if coin_refs.len() > max_payment {
+        let remaining = coin_refs.len() - max_payment;
+        coin_refs.truncate(max_payment);
+        eprintln!(
+            "Warning: a gas payment holds at most {max_payment} coins, so {remaining} of your \
+             coins will not be sent. Run the command again to send the rest."
```
amnn: nit: technically they may need to run the command multiple times to send the rest, how about:


```suggestion
             coins will not be sent. Run the command repeatedly to send the rest."
```

crates/sui/src/client_commands.rs
```diff
@@ -3416,6 +3422,126 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
...
+        .try_collect()
+        .await?;
+    ensure!(!coins.is_empty(), "No coins available to send");
+
+    let mut coin_refs: Vec<ObjectRef> =
+        coins.iter().map(|c| c.compute_object_reference()).collect();
+
+    // Every coin has to fit in a single gas payment, so when the signer holds more than that, drain
+    // as many as will fit and let them run the command again for the rest.
+    let max_payment = max_gas_payment_objects(&client).await?;
```
amnn: You can send more than coins by accepting them as inputs and merging them into the gas coin.

stefan-mysten: I see, ok. I can rework this part.

crates/sui/src/client_commands.rs
stefan-mysten: I think I misunderstood the original code here. Yes, we should be able to use `--all-coins` together with `--coin-type` and transfer whatever `Coin<T>` the address has into its address balance.

stefan-mysten: However, in the original implementation that @mystenmark did, it seems that this was not possible. Not sure if there's a specific reason for this.

stefan-mysten: Reworked this completely. You can now send all coins of various types, and it will fold them into the AB. I also renamed the `stateless` flag because it was strange, and reworked the logic.

jessiemongeon1:
## Style Guide Audit

Audited **2** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

**3** violation(s) found. All must be fixed before merge.

### `docs/content/onchain-finance/asset-custody/address-balances/using-address-balances.mdx` (3 violation(s))

_3 violation(s) (0 regex, 3 claude)_

- **Line 65** — passive voice
  - Current: `Gas is chosen independently of this flag`
  - Fix: `The system chooses gas independently of this flag`
- **Line 67** — passive voice
  - Current: `gas is paid from the sender's address balance`
  - Fix: `the system pays gas from the sender's address balance`
- **Line 67** — passive voice
  - Current: `gas is paid from the sender's coins`
  - Fix: `the system pays gas from the sender's coins`

---

crates/sui/src/client_commands.rs
```diff
@@ -128,6 +128,18 @@ pub(crate) static USER_AGENT: &str =
 /// Only to be used within CLI
 pub const GAS_SAFE_OVERHEAD: u64 = 1000;
 
+/// The protocol limits that bound how many coins a single transaction can drain.
```
amnn: nit: file order -- private type in front of the main public type in this module.

crates/sui/src/client_commands.rs
```diff
@@ -1376,12 +1399,16 @@ impl SuiClientCommands {
...
-                    let use_address_balance = if stateless {
+                let (amount, use_address_balance) = if let Some(amount) = amount {
+                    let use_address_balance = if from_address_balance {
+                        ensure!(
+                            address_balance >= amount,
+                            "Insufficient address balance to send {} MIST. \
+                            Address balance: {}, Coin balance: {}",
+                            amount,
+                            address_balance,
+                            coin_balance
```
amnn: nit: inline into the `{}`s?

crates/sui/src/client_commands.rs
```diff
@@ -3416,62 +3439,181 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
...
+    Ok(CoinLimits {
+        max_gas_payment_objects: limit("max_gas_payment_objects")?,
+        max_arguments: limit("max_arguments")?,
+        max_input_objects: limit("max_input_objects")?,
+    })
+}
+
+/// Warn about, and drop, any coins past what one transaction can hold.
+fn truncate_to_max_coins(coin_refs: &mut Vec<ObjectRef>, max_coins: usize) {
+    if coin_refs.len() > max_coins {
```
amnn: nit: early return style?

crates/sui/src/client_commands.rs
```diff
@@ -3416,62 +3439,181 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
...
+        .await?;
+    ensure!(
+        !coins.is_empty(),
+        "No {coin_type_tag} coins available to send"
+    );
+
+    let mut coin_refs: Vec<ObjectRef> =
+        coins.iter().map(|c| c.compute_object_reference()).collect();
+
+    // Coins that do not fit in the gas payment are merged in by a single `MergeCoins`, so they are
```
amnn: Why a single `MergeCoins` (as opposed to multiple, up to the transaction input limit?)

crates/sui/src/client_commands.rs
```diff
@@ -3416,62 +3439,181 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
...
+    let limits = coin_limits(&client).await?;
+    let smashed = if is_sui {
+        limits.max_gas_payment_objects
+    } else {
+        0
+    };
+    let max_merged = std::cmp::min(
+        limits.max_arguments - if is_sui { 1 } else { 0 },
+        limits.max_input_objects,
+    );
```
amnn:
```suggestion
    let max_merged = limits
        .max_arguments
        .saturating_sub(is_sui as u64)
        .min(limits.max_input_objects)
```

amnn: (although based on my earlier comment about supporting multiple merge coins, this may need to change further)

crates/sui/src/client_commands.rs
```diff
@@ -3416,62 +3439,181 @@ pub async fn max_gas_budget(client: &Client) -> Result<u64, anyhow::Error> {
...
+        .map(|r| builder.obj(ObjectArg::ImmOrOwnedObject(*r)))
+        .collect::<Result<Vec<_>, _>>()?;
+
+    // Gas smashing has already merged the gas payment into the gas coin, so that is what the rest
+    // merge into; without one, the first coin takes the role. `MergeCoins` only borrows its target,
+    // so either way it is still ours to send afterwards.
+    let target = if is_sui {
+        Argument::GasCoin
+    } else {
+        merged_args.remove(0)
```
amnn: nit: pop rather than remove from the front? I don't think it matters which one, so might as well make the algorithmically efficient choice?

amnn (approved):
Looks good -- there's an opportunity to support even more merging, but that's not a requirement by any means.

crates/sui/src/client_commands.rs
stefan-mysten: Mainly kept it simple. I reworked it to use multiple `MergeCoins`.

crates/sui/src/client_commands.rs
stefan-mysten: Good point, fixed.

PR #27299, opened 2026-07-15
Added replay-based support for tracing dry-run

awelc:
## Description 

What the subject says. Enabled via `sui client ptb --dry-run --trace`. This implementation has two parts:
- on the sui client side, transaction data is constructed the same way it would be for a "regular" dry-run, but instead of being sent to the full node it is passed on to the replay tool for local execution
- on the replay tool side;
   - owned inputs and gas coins are fetched at the versions recorded in the transaction data, while shared objects, packages, and dynamic loads are fetched at their newest versions
   - an artificial gas coin is injected if the transaction has no gas payment (in a manner similar to how it's done at a full node for a "regular" dry-run)
   - transaction is executed locally producing the same artifacts as a "regular" replayed transaction (but no effects comparison as there are no effects)


awelc:
Subsumed by https://github.com/MystenLabs/sui/pull/27592

PR #27300, opened 2026-07-15
Add epoch close deadline failsafe for stuck deferred transactions

aschran:
## Description

If a bug leaves deferred transactions that never get scheduled, the epoch cannot close: the network gets stuck partially closed — rejecting user transactions but unable to advance the epoch — until manual intervention. This has caused outages.

Add a deadline after which the epoch closes anyway, abandoning whatever is still deferred. Two deliberate design choices: the deadline is a pure function of consensus-agreed inputs, so all validators force-close at the same commit without any new persisted state; and stragglers are dropped rather than cancel-executed, because this path only runs when deferral machinery is already broken, so per-validator deferred state cannot be assumed consistent — executing it would feed divergent state into end-of-epoch checkpoints where it could stall certification.

## Test plan

New unit tests cover the deadline trigger and abandonment accounting. A new e2e simtest injects a different unschedulable deferred transaction on every validator and asserts the cluster still reconfigures — this hangs without the fix, and successful checkpoint certification doubles as the check that divergent deferred state never reaches checkpoint contents.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/consensus_handler.rs
+++ b/crates/sui-core/src/consensus_handler.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
-    collections::{BTreeMap, HashMap, HashSet, VecDeque},
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque},
     hash::Hash,
     num::NonZeroUsize,
     sync::{Arc, Mutex},
@@ -927,6 +927,17 @@ struct CommitHandlerState {
 }
 
 impl CommitHandlerState {
+    fn new(epoch_store: &AuthorityPerEpochStore, consensus_round: u64) -> Self {
+        Self {
+            output: ConsensusCommitOutput::new(consensus_round),
+            dkg_failed: false,
+            randomness_round: None,
+            indirect_state_observer: Some(IndirectStateObserver::new()),
+            initial_reconfig_state: epoch_store.get_reconfig_state_read_lock_guard().clone(),
+            occurrence_counts: HashMap::new(),
+        }
+    }
+
     fn get_notifications(&self) -> Vec<SequencedConsensusTransactionKey> {
         self.output
             .get_consensus_messages_processed()
@@ -981,6 +992,14 @@ impl CommitHandlerState {
 
         randomness_manager
     }
+}
+
+/// Deferred transactions abandoned because the epoch close deadline forced the epoch
+/// closed while they were still unscheduled.
+struct AbandonedDeferredTxns {
+    count: usize,
+    // At most 10 (key, digest) pairs for logging.
+    sample: Vec<(DeferralKey, TransactionDigest)>,
 }
 
 impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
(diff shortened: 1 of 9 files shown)


crates/sui-core/src/consensus_handler.rs
```diff
@@ -2243,13 +2292,23 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+        // past the deadline it closes regardless, abandoning whatever remains.
+        let should_close = !has_deferred_txns || deadline_reached;
+        let final_round = should_close && !start_state_is_reject_all_tx;
 
-        if !commit_has_deferred_txns && !previous_commits_have_deferred_txns {
-            if !start_state_is_reject_all_tx {
-                info!("Transitioning to RejectAllTx");
+        let mut abandoned_deferred_txns = None;
+        if final_round {
+            info!("Transitioning to RejectAllTx");
```
mystenmark, line 2304: why did we lose the start state check? before this log?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2258,11 +2317,52 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
-        } else {
-            (reconfig_state, false)
+        (reconfig_state, final_round, abandoned_deferred_txns)
+    }
+
+    /// Abandons every deferred transaction still unscheduled when the epoch close deadline
+    /// forces the epoch closed: drops this commit's newly staged deferrals and stages all
+    /// remaining deferral keys for deletion.
+    ///
+    /// (Without this, post-close commits would reload and re-defer them forever, growing the
```
alex-mysten, line 2327: Should we instead have a limit on the number of times a transaction can be deferred?

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2243,13 +2292,23 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
-        if !commit_has_deferred_txns && !previous_commits_have_deferred_txns {
-            if !start_state_is_reject_all_tx {
-                info!("Transitioning to RejectAllTx");
+        let mut abandoned_deferred_txns = None;
+        if final_round {
+            info!("Transitioning to RejectAllTx");
+            if has_deferred_txns {
+                // Closing with deferred transactions remaining implies the deadline forced it.
+                debug_assert!(deadline_reached);
+                abandoned_deferred_txns = self.abandon_deferred_transactions(state);
```
alex-mysten, line 2308: Does this presume that deferred transactions are the only thing that can prevent epoch closure, or will the epoch close regardless?

mystenmark (approved):

crates/sui-core/src/consensus_handler.rs
mystenmark: my read is no - but if there are deferred transactions we have to clear them

crates/sui-core/src/consensus_handler.rs
aschran: we already do for congestion deferrals, i guess we could also add that for randomness deferrals. we'd have to set it high enough that we don't see any cancellations at start of epoch when dkg is pending.

my intention was for this mechanism to also apply to any epoch-close-blockers that may be added in the future, but i guess that might be unlikely ever to happen that we add another one.

as a backstop, this seemed less likely to me to have bugs vs an approach where we only add deferral limit to randomness tx, bc the logic lives exactly in the place where reconfig state is advanced.

crates/sui-core/src/consensus_handler.rs
aschran: i think epoch definitely will close once we advance reconfig state to close_all_tx and return final_round==true here

(or at least we will definitely advance to final checkpoint construction)

crates/sui-core/src/consensus_handler.rs
aschran: i think we did not lose it? it's part of
```
let final_round = should_close && !start_state_is_reject_all_tx;
```

the only behavior change here is that now we won't repeatedly call `reconfig_state.close_all_tx()` when before we were doing that, but it seems like we didn't actually mean to be doing that?

PR #27301, opened 2026-07-15
Remove unused epoch close logic

mwtian:
## Description 

In every network, `EndOfPublish` message is no longer emitted during normal epoch closing. The consensus handler closes the epoch automatically when the consensus commit timestamp reaches the reconfiguration timestamp. The logic to transition into `ReconfigCertStatus::RejectUserCerts` and send `EndOfPublish` message based on timestamp are removed.

`EndOfPublish` message is still sent by manual epoch closing on individual validators. The logic to count `EndOfPublish` quorum and recover manual epoch close state are kept.

`ReconfigCertStatus::RejectUserCerts` is now reserved for manual epoch close.

## Test plan 

CI

### DIFF
--- a/crates/sui-core/src/consensus_handler.rs
+++ b/crates/sui-core/src/consensus_handler.rs
@@ -77,7 +77,6 @@ use crate::{
         CheckpointHeight, CheckpointRoots, CheckpointService, CheckpointServiceNotify,
         PendingCheckpoint, PendingCheckpointInfo,
     },
-    consensus_adapter::ConsensusAdapter,
     consensus_throughput_calculator::ConsensusThroughputCalculator,
     consensus_types::consensus_output_api::{ConsensusCommitAPI, ParsedTransaction},
     epoch::{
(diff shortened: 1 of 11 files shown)


mystenmark:
this will probably have substantial conflicts with #27300

crates/sui-core/src/consensus_handler.rs
```diff
@@ -2274,8 +2246,8 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
         false
     }
 
-    /// After we have collected 2f+1 EndOfPublish messages, we call this function every round until the epoch
-    /// ends.
+    /// Once the timestamp deadline is reached or 2f+1 EndOfPublish messages are collected, we call
+    /// this function every round until the epoch ends.
```
mystenmark, line 2250: We should remove any mention of EoP (also from this function name) since it isn't used anymore.

mystenmark (approved):

crates/sui-core/src/consensus_handler.rs
mwtian: EndOfPublish is still used by the manual epoch close feature. Good point on the function name.

PR #27304, opened 2026-07-15
[sui-framework] Add `get_*` macros to `sui::scratch`

tnowacki:
## Description 

- Adds `get_do`, `get_do_mut`, `get_fold`, and `get_fold_mut` macros for "borrow"-like patterns
- As an implementation detail, added borrow "guards". It might be a bit overkill? But it seems potentially helpful to prevent footguns

## Test plan 

- New tests 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-framework/packages/sui-framework/tests/scratch_tests.move
```diff
@@ -217,6 +220,239 @@ fun replace_wrong_old_type() {
...
+// is held by a `BorrowMarker`, so an `add` to that key aborts.
+#[test, expected_failure(abort_code = sui::scratch::EEntryAlreadyExists)]
+fun get_do_borrow_protection() {
+    let mut ctx = tx_context::dummy();
+    ctx.scratch_internal_add!<Marker, u64>(Marker(), 42);
+    ctx.scratch_internal_get_do!(Marker(), |_v: &u64| {
+        // add fails due to borrow protection, even though the value was temporarily removed
+        // TODO(regex) Remove once we have the regex borrow checker
+        let mut ctx = tx_context::dummy();
+        ctx.scratch_internal_add!<Marker, bool>(Marker(), true);
```
tnowacki, line 265: This is a limitation due to `copy` of `ctx: &mut TxContext` and epsilon/alias borrows. This will go away in the new borrow checker

crates/sui-framework/packages/sui-framework/sources/scratch.move
```diff
@@ -19,6 +19,14 @@ module sui::scratch;
 /// all access to scratch entries.
 public struct Permit<phantom K: copy + drop>() has copy, drop;
 
+/// Occupies a key's slot while `get_do`, `get_fold`, or their mutable variants have the value of
+/// type `V` removed, so nothing can add to or read the slot while the value is being "borrowed".
+/// Each one carries a transaction-unique id, so a marker cannot be forged to match one in use.
+public struct BorrowMarker<phantom V: drop>(u64) has copy, drop;
```
tzakian: Should this not be `copy`?

crates/sui-framework/packages/sui-framework/sources/scratch.move
```diff
@@ -105,8 +116,140 @@ public fun replace<K: copy + drop, VNew: drop, VOld: drop>(
...
+/// by the borrow-style macros (`get_do`, `get_fold`, and their mutable variants).
+/// Aborts with `EEntryDoesNotExist` if there is no entry for `key`.
+/// Aborts with `EEntryTypeMismatch` if the entry exists, but its value is not of type `V`.
+public fun begin_borrow<K: copy + drop, V: drop>(
+    ctx: &mut TxContext,
+    permit: Permit<K>,
+    key: K,
+): (V, BorrowMarker<V>) {
+    let value = remove<K, V>(ctx, permit, key);
+    let marker = borrow_marker<V>(ctx);
```
tzakian, line 134: why not make `BorrowMarker` non-copy and just make a manual copy to hand back here?

tzakian (approved):
LGTM. One small thing about making `BorrowMarker` non-copy since that seems a bit too loose, but not a big deal either way if you decide to keep it as-is.

crates/sui-framework/packages/sui-framework/sources/scratch.move
tnowacki: Good call. 
It used to need to be before I encapsulated things a bit more in begin/end.

PR #27311, opened 2026-07-16
[consensus] add crash-and-restart simtests

mwtian:
## Description 

Stacked on #27319 (consensus-core shutdown cleanup); this PR contains only the simtest changes.

Adds consensus-only simtests that exercise crash recovery under load, modeled on `test_committee_start_simple`:
- `test_consensus_crash_and_restart`: starts a 10-authority committee under continuous transaction load, and registers fail points at key consensus points (store before/after write, after propose, after leader-schedule change) where each node has a small probability of killing itself. The test harness recreates the killed node after a short delay, forcing recovery from persisted on-disk state, and asserts every authority recovers and keeps making commit progress.
- `test_consensus_rolling_restarts_all_validators`: repeatedly restarts every validator in random order under load, with varying downtime, and asserts the fully restarted committee reconnects and makes new commit progress.
- `test_committee_fast_path` is renamed to `test_consensus_transaction_votes` and now runs under the same random crash injection (with a cool-down between crashes so the test fits its time budget). Transactions are submitted from separate tasks so a down or recovering authority cannot stall the load, and the cross-authority commit comparison tolerates restarts by re-acquiring commit consumer channels and skipping replayed commits.

Harness changes:
- The simtest `AuthorityNode` deletes and recreates its simulator node across crashes, and `stop()` awaits `ConsensusAuthority::stop()` before deleting the node, mirroring the full-node restart fixture.
- Committee startup boilerplate and crash injection are shared across the tests.

## Test plan 

- `cargo simtest -p consensus-simtests`: all 10 tests pass, including the new crash-and-restart and rolling-restart tests and the crash-enabled transaction votes test.

### DIFF
--- a/consensus/simtests/tests/consensus_tests.rs
+++ b/consensus/simtests/tests/consensus_tests.rs
@@ -3,24 +3,35 @@
 
 #[cfg(msim)]
 mod consensus_tests {
-    use std::sync::atomic::{AtomicU64, Ordering};
-    use std::{sync::Arc, time::Duration};
+    use std::{
+        collections::{BTreeMap, HashMap},
+        sync::{
+            Arc,
+            atomic::{AtomicBool, AtomicU64, Ordering},
+        },
+        time::Duration,
+    };
 
     use consensus_config::{
         Authority, AuthorityIndex, AuthorityName, Committee, ConsensusProtocolConfig, Epoch,
         NetworkKeyPair, Parameters, ProtocolKeyPair, Stake,
     };
     use consensus_core::NoopTransactionVerifier;
-    use consensus_core::{BlockAPI, BlockStatus, Priority, TransactionVerifier, ValidationError};
+    use consensus_core::{
+        BlockAPI, BlockStatus, CommitIndex, CommitRef, CommittedSubDag, Priority,
+        TransactionVerifier, ValidationError,
+    };
     use consensus_simtests::node::{AuthorityNode, Config};
-    use consensus_types::block::{BlockRef, TransactionIndex};
+    use consensus_types::block::{BlockRef, BlockTimestampMs, TransactionIndex};
     use fastcrypto::traits::{KeyPair as _, ToFromBytes as _};
     use mysten_metrics::RegistryService;
+    use mysten_metrics::monitored_mpsc::UnboundedReceiver;
     use mysten_network::{Multiaddr, multiaddr::Protocol};
+    use parking_lot::Mutex;
     use prometheus::Registry;
-    use rand::{Rng, SeedableRng as _, rngs::StdRng};
+    use rand::{Rng, SeedableRng as _, rngs::StdRng, seq::SliceRandom as _};
     use sui_config::local_ip_utils;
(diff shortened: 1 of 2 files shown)


consensus/simtests/tests/consensus_tests.rs
```diff
@@ -131,71 +127,212 @@ mod consensus_tests {
...
+
+        crashes.stop().await;
+
+        // Every authority should have recovered from its crashes and made commit progress.
+        for i in 0..NUM_OF_AUTHORITIES {
+            let highest_committed_index = authorities[i]
+                .commit_consumer_monitor()
+                .highest_handled_commit();
+            tracing::info!("Authority {i} highest handled commit: {highest_committed_index}");
+            assert!(
```
arun-koshy, line 209: A node can make progress here by replaying commits from disk without actually ever rejoining the network right? Should we track the highest commit and ensure it can recover to the point where it fully rejoins and gets newer commits than what was available locally?

consensus/simtests/tests/consensus_tests.rs
```diff
@@ -577,6 +726,201 @@ mod consensus_tests {
...
+        authority: &AuthorityNode,
+        receiver: &mut UnboundedReceiver<CommittedSubDag>,
+        last_seen_commit_index: CommitIndex,
+    ) -> CommittedSubDag {
+        loop {
+            let Some(sub_dag) = receiver.recv().await else {
+                *receiver = authority.commit_consumer_receiver();
+                continue;
+            };
+            if sub_dag.commit_ref.index <= last_seen_commit_index {
```
arun-koshy, line 960: Seems  like a good place to add tracking of the rejected tx set and that replay does not diverge that rejected set.

consensus/simtests/tests/consensus_tests.rs
```diff
@@ -131,71 +127,212 @@ mod consensus_tests {
...
+
+                authority.start().await.unwrap();
+                authority.spawn_committed_subdag_consumer().unwrap();
+            }
+            restart_orders.push(restart_order);
+        }
+
+        // Measure after every validator has been replaced for the final time. Each final instance
+        // must advance from here, proving the fully restarted committee reconnected.
+        let post_restart_commit_indexes = authorities
```
arun-koshy: The last restarted validator will not have been awaited since spawn at this point so its snapshot will be 0, right?

arun-koshy (approved):

akichidis (approved):

consensus/simtests/tests/consensus_tests.rs
mwtian: Gave 30s which should be more than enough for catching up.

consensus/simtests/tests/consensus_tests.rs
mwtian: Added a wait before taking snapshot.

consensus/simtests/tests/consensus_tests.rs
mwtian: Added.

PR #27314, opened 2026-07-16
[move][move-decompiler] Proper loop exit fusion

cgswords:
## Description 

Replaces the decompiler's bespoke dispatch-table machinery for multi-exit loops with No More Gotos §V-B: each loop reduces to an abstract node whose outgoing edges carry per-exit reach formulas, letting the existing acyclic structuring produce post-loop if/else cascades naturally — a net −600 lines. Combined with an NMG Algorithm 2 dominance fix in loop-successor refinement, it eliminates the last residual gotos we were producing.

## Test plan 

Tests updated.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


tzakian (approved):
seems reasonable to me 👍🏻

PR #27317, opened 2026-07-16
Resolve post-consensus owned-object conflicts from in-memory state and the objects table

andll:
## Description

PR 1 of 2 toward removing the `owned_object_locked_transactions` epoch table. Post-consensus owned-object conflict detection stops reading the lock table; the table is still fully written, and PR 2 (#27302) deletes it.

**Release / rollback plan.** This split exists for rollback safety:

- **Step 1 (this PR)**: every write to the lock table is preserved byte-for-byte — only reads move to in-memory state + the objects table. Sequencing decisions (finalize/drop/defer/cancel) are identical to main in all reachable states, so the fleet can run mixed versions, and rolling back to any earlier binary is safe at any point, mid-epoch included: the table an old binary would read never stopped being maintained.
- **Step 2 (#27302)**: deletes the table, the backstop read, and the differential check. Deploy only once step 1 is running everywhere; a step-2 rollback targets the step-1 binary (not older), since rows stop being written under step 2.

**What changed.** The commit handler resolves the claim state of each owned input ref through layers:

1. **Quarantine locks** (in-memory, existing): locks of finalized commits not yet flushed.
2. **Deferred-locks map** (new, in-memory): locks of deferred transactions, which are finalized but unexecuted. Entries live from deferral until the flush of the commit that reloads the transaction, and are re-seeded from the durable deferred-transactions table at restart. Deferral-key collisions (a re-deferral keeps its original `deferred_from_round`) retain main's exact scheduling semantics — the displaced entries' lock coverage alone is preserved under a sentinel deferral key that no reload can reach, read back only by restart seeding.
3. **Live-object cache** (new, in-memory): a monotone lower bound on each object's latest version, warmed by vote-time validation (both accept and reject paths) and bumped at quarantine flush. Owned per epoch store, so bounds derived from state the end-of-epoch revert discards die with the epoch.
4. **Objects table** (authoritative fallback): verdict rule `conflict ⟺ latest_version(id) > claimed_version`.
5. **Lock-table backstop**: a table read only for refs whose object is immutable at exactly the claimed version (see safety argument).

Also included: strict vote-time immutable-claims verification (the check no longer short-circuits on an empty claims list), and in debug builds (including simtests) every verdict is differentially validated against the lock-table-based resolution, dumping all resolution layers on divergence.

**Safety argument.** Locks are never released within an epoch, and execution bumps every exclusive-owned input ref on all paths (success, abort, cancel). A commit leaves the quarantine only after its checkpoints are executed, so any lock absent from the memory layers belongs to a durably executed holder — meaning the objects table already shows `latest > claimed` for that ref. Conversely, `latest > claimed` implies a finalized consumer this epoch (stale claims are vote-rejected, so they cannot reach quorum). Hence the objects-table rule is exact for everything outside the memory layers:

- *Pipelined inputs* (claimed version not yet visible locally): the producer finalizes before the claimant can be voted on, so any competing lock is still in the memory layers; resolving "unclaimed" from `latest ≤ claimed` is exact.
- *Cache verdicts are decisive both ways* because bounds are lower bounds (versions never regress) and every path that removes a lock from the memory layers bumps the cached bound past the consumed version first, under the quarantine write lock; the resolver holds the read guard across each commit's batch, so it always sees either the lock or the bump.
- *The one undecidable case*: an object immutable at exactly its claimed version. A lock on such a ref never bumps (immutable inputs aren't mutated), so the objects table cannot see it. These refs — reachable only via byzantine under-claiming of immutable inputs, never by honest traffic — are routed to the still-written lock table, keeping verdicts exact. Strict claims verification makes such locks quorum-unreachable once deployed everywhere, which is what allows PR 2 to delete the table and backstop. The cache tracks an `Option<bool>` immutable bit per version for this routing; observations that didn't inspect the object (consumption bumps, ref-only reads) record `unknown`, and an equality hit with an unknown bit re-reads the object once to learn it.
- *Duplicate sequencing across restart* (the same transaction finalized again after its first commit was flushed): the old path accepted it by table digest-equality; the new path accepts it via "this digest already executed in this epoch", durably true wherever the memory layers no longer hold the first sequencing's locks.
- *Lock sets are derived per occurrence* of a transaction in a commit (immutable claims are not covered by the digest, so occurrences of the same digest can carry different claim sets — a vote-rejected duplicate must not shrink the accepted occurrence's lock set).
- *Restart / reconfig*: the cache is per-epoch and rebuilt from authoritative reads; the deferred-locks map is re-seeded from the durable deferred-transactions rows, including the displaced-deferral sentinel rows. A store opened mid-epoch may be rebuilding from rows an older binary wrote (upgrade/rollback windows), which cannot convey displaced-deferral coverage - such a node resolves every potential-clear at exactly the claimed version from the lock table for the remainder of that epoch, making its verdicts equal the table's (and every other node's) in any mixed binary history. Fresh epochs and never-restarted nodes run purely in memory.

Net: verdicts are identical to the lock-table path in all reachable states — asserted per finalized transaction in debug builds — so no protocol gating is needed. Reachability assertions (antithesis/simtest) cover every new resolution branch.

## Test plan

- New unit tests: live-object cache bound/merge semantics, deferred-locks map coverage across quarantine flush, per-occurrence lock-set derivation under a vote-rejected divergent-claims duplicate, and sentinel-row seeding (locks rebuilt, no reload eligibility).
- The debug-build differential check asserts objects-table-based verdicts match lock-table-based verdicts for every finalized transaction across CI and simtests; it caught (and its state dump root-caused) a pre-existing deferral-key-collision bug during development, reproduced deterministically with `MSIM_TEST_SEED=7427020612626106405` on `test_simulated_load_with_reconfig_and_correlated_crashes`.
- Antithesis: single-version runs plus a split-cluster (this PR vs main) upgrade run exercising mixed-fleet verdict parity.
- The superset of this change (including the table deletion, #27302) ran on private-testnet at 16K TPS with lower commit-handler utilization than main and zero steady-state objects-DB lookups (`consensus_owned_object_lock_resolutions{source="objects_db"}` ≈ 0).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/consensus_handler.rs
+++ b/crates/sui-core/src/consensus_handler.rs
@@ -159,6 +159,7 @@ pub(crate) fn resolve_owned_object_lock_states(
     live_object_cache: &LiveObjectCache,
     obj_refs: &[ObjectRef],
 ) -> (HashMap<ObjectRef, LockResolution>, LockResolutionStats) {
+    let _scope = monitored_scope("ConsensusCommitHandler::resolve_owned_object_locks");
     let mut stats = LockResolutionStats::default();
     let mut resolutions = HashMap::new();
     let mut backstop_refs = Vec::new();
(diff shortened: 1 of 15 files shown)


crates/sui-core/src/live_object_cache.rs
```diff
@@ -0,0 +1,351 @@
...
+            return;
+        }
+        if shard.len() >= self.capacity_per_shard {
+            // Evict one arbitrary entry to make room; evicting is always sound
+            // (a missing entry just means an authoritative fallback read).
+            if let Some(&victim) = shard.keys().next() {
+                shard.remove(&victim);
+            }
+        }
+        shard.insert(id, bound);
```
lxfind, line 176: **Cache eviction defeats the monotonic lower bound; a stale re-warm can flip a conflict verdict** -- Confidence: 100/100 | Importance: 100/100

(found by: Codex)
- **File**: `crates/sui-core/src/live_object_cache.rs` lines 169-176
- **Problem**: The resolver relies on cache entries never moving backward. Max-merge only protects a bound while the entry is resident. Sequence: a quarantine flush bumps object A from v5 to v6 after its consumer executed. A later insert evicts A from a full shard. A delayed vote-time read then inserts A at v5. Nothing remains to reject the stale value. The resolver now sees a known-mutable v5 entry matching the claim, treats the ref as unclaimed, and skips the authoritative objects-table read. The comment at lines 29-36 describes this exact window and accepts it as unlikely.
- **Impact**: A competing transaction can be finalized on one validator while another validator, whose cache still holds v6 or whose eviction order differed, drops it. Eviction order and thread interleaving are local to each node, so this is a checkpoint-fork risk in release builds. The debug-only differential check against the old table will not catch it in production. Small randomized shard capacities in tests and the production `LIVE_OBJECT_CACHE_SIZE` both make the ordering reachable.
- **Suggestion**: Preserve a monotonic fence across eviction rather than accepting the window. Options: generation or read tickets in the style of `MonotonicCache`, non-evictable consumption markers, or using cache bounds only to prove conflicts and never to prove absence of conflict. A simple safe fallback is to stop evicting resident entries and drop inserts for unseen IDs when a shard is full. Both the consumption bump and any later stale warm would then miss, forcing an authoritative read. Add a deterministic regression test for bump, evict, stale reinsert.

---

crates/sui-core/src/consensus_handler.rs
```diff
@@ -113,6 +116,255 @@ struct FilteredConsensusOutput {
...
+                live_object_cache.record_absent(obj_ref.0);
+            }
+        }
+    }
+
+    if !backstop_refs.is_empty() {
+        stats.backstop += backstop_refs.len() as u64;
+        let locks = epoch_store
+            .multi_get_owned_object_locks_from_db(&backstop_refs)
+            .expect("cannot read owned object lock table");
```
lxfind, line 357: **Resolver panics on `EpochEnded` when called from the RPC submit path** -- Confidence: 95/100 | Importance: 85/100

(found by: Codex)
- **File**: `crates/sui-core/src/consensus_handler.rs` lines 353-357
- **Problem**: `resolve_owned_object_lock_states` calls `expect` on the result of its lock-table backstop read. The same function is invoked from `authority_server.rs` lines 1008-1012 for resubmitted transactions. There, an old epoch-store guard can outlive reconfiguration and `release_db_handles()`. In recovery mode, ordinary equality claims take the backstop path, so a release between the earlier processed-message check and this read yields `EpochEnded`, which hits the `expect`.
- **Impact**: An epoch-boundary RPC race aborts a release validator process, even though the submit path explicitly intends to surface failed reads as retriable RPC errors. Client retries around reconfiguration turn this into an availability problem.
- **Suggestion**: Return `SuiResult<(HashMap<...>, LockResolutionStats)>` from the resolver and propagate `EpochEnded` through the RPC path as a retriable error. Let the consensus handler decide separately that an unexpected failure is fatal while it owns an active epoch. Add a test that releases epoch DB handles while the RPC resolver enters the backstop.

---

crates/sui-core/src/lib.rs
```diff
@@ -28,6 +28,7 @@ mod fallback_fetch;
 pub mod gasless_rate_limiter;
 pub mod global_state_hasher;
 pub mod jsonrpc_index;
+pub mod live_object_cache;
```
lxfind, line 31: **`LiveObjectCache` and its mutators are public across the crate boundary** -- Confidence: 100/100 | Importance: 74/100

(found by: Codex)
- **File**: `crates/sui-core/src/lib.rs` line 31; `crates/sui-core/src/authority/authority_per_epoch_store.rs` lines 1952-1954
- **Problem**: The cache is an implementation detail of consensus-handler conflict resolution, yet it is declared as a public module, and the public epoch-store accessor exposes every mutation method. No crate outside `sui-core` uses it.
- **Impact**: External crates can couple to or directly mutate consensus-critical resolution state. This violates the repository's explicit no-abstraction-leaks requirement and makes future changes to conflict verdicts harder to reason about.
- **Suggestion**: Make the module, cache types, mutators, and the epoch-store accessor `pub(crate)`. Route both the consensus handler and the RPC server through a small private owned-object conflict-resolution API, ideally in its own module, instead of exposing consensus-handler internals to the RPC layer.

---

lxfind (commented):

---

PR #27318, opened 2026-07-16
[consensus] add metrics on block bytes size and proposed block latency

mwtian:


consensus/core/src/commit_finalizer.rs
```diff
@@ -254,6 +255,22 @@ impl CommitFinalizer {
             }
         }
 
+        let utc_now = self.context.clock.timestamp_utc_ms();
+        for commit in &finalized_commits {
```
arun-koshy, line 259: do recovered/synced commits throw the data off here?

arun-koshy (approved):

consensus/core/src/commit_finalizer.rs
mwtian: Synced commits go through this path but recovered commits don't. This only measures own proposed blocks in finalized commits, so these edge cases should be rare.

PR #27319, opened 2026-07-17
[consensus] finish tasks and propagate panics during authority shutdown

mwtian:
## Description 

Authority shutdown previously aborted component tasks and dropped their JoinHandles, which could leak resources across an in-process restart and silently swallow task panics. This PR makes shutdown deterministic, in preparation for crash-and-restart simtests (stacked on top as #27311):

- Components stop by finishing their tasks instead of dropping them. Observer subscriptions shut down cooperatively so in-flight block handlers are awaited; replaced subscriptions are retired and joined at stop, with finished tasks reaped on resubscription to keep the retired list bounded. The observer's current and retired subscriptions are tracked under a single lock so replace-and-retire is atomic w.r.t. concurrent subscribe() and stop().
- Subscription and network proxy tasks hold weak references to services, and `AuthorityNode::stop()` verifies DagState ownership was released.
- Task panics propagate consistently through every stop path (core thread, synchronizer, commit syncer, commit finalizer, subscribers) via shared helpers in `consensus/core/src/task.rs`, replacing six hand-rolled abort/join/resume_unwind copies.
- Documents shutdown ownership of consensus components, and avoids deep-cloning reject transaction vote indices in commit syncer fetches.

## Test plan 

- `cargo nextest run -p consensus-core`: all 286 unit tests pass, including new shutdown tests (e.g. `test_stop_waits_for_block_handlers`).
- `cargo simtest -p consensus-simtests`: the existing simtests pass unchanged against the new shutdown behavior.
- clippy (xclippy lint set) and `cargo fmt` clean.

### DIFF
--- a/consensus/core/src/authority_node.rs
+++ b/consensus/core/src/authority_node.rs
@@ -1,7 +1,10 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{sync::Arc, time::Instant};
+use std::{
+    sync::{Arc, Weak},
+    time::Instant,
+};
 
 use consensus_config::{
     Committee, ConsensusProtocolConfig, NetworkKeyPair, NetworkPublicKey, Parameters,
@@ -9,6 +12,7 @@ use consensus_config::{
 };
 use consensus_types::block::Round;
 use itertools::Itertools;
+use mysten_common::debug_fatal;
 use mysten_network::Multiaddr;
 use parking_lot::RwLock;
 use prometheus::Registry;
@@ -142,10 +146,10 @@ enum SubscriberType<N: NetworkManager> {
 }
 
 impl<N: NetworkManager> SubscriberType<N> {
-    fn stop(&self) {
+    async fn stop(&self) {
         match self {
-            SubscriberType::Validator(subscriber) => subscriber.stop(),
-            SubscriberType::Observer(subscriber) => subscriber.stop(),
+            SubscriberType::Validator(subscriber) => subscriber.stop().await,
+            SubscriberType::Observer(subscriber) => subscriber.stop().await,
         }
     }
 }
@@ -159,12 +163,18 @@ where
     transaction_client: Arc<TransactionClient>,
     synchronizer: Arc<SynchronizerHandle>,
     store: Arc<RocksDBStore>,
+    // Only use for verification and logging during shutdown.
+    // To avoid keeping the DagState alive at the end of shutdown, this is only a weak reference.
+    dag_state: Weak<RwLock<DagState>>,
 
     commit_syncer_handle: CommitSyncerHandle,
     round_prober_handle: Option<RoundProberHandle>,
     leader_timeout_handle: LeaderTimeoutTaskHandle,
     core_thread_handle: CoreThreadHandle,
     subscriber: SubscriberType<N>,
+    // Network proxies hold ObserverService weakly, so AuthorityNode keeps the service alive until
+    // the network servers have stopped.
+    observer_service: Option<Arc<ObserverService>>,
     network_manager: N,
 }
 
@@ -406,7 +416,7 @@ where
             store.clone(),
         ));
 
-        let (subscriber, round_prober_handle) = if context.is_validator() {
+        let (subscriber, round_prober_handle, observer_service) = if context.is_validator() {
             let authority_service = Arc::new(AuthorityService::new(
                 context.clone(),
                 block_verifier.clone(),
@@ -451,7 +461,7 @@ where
             );
 
             // Start the observer server if the observer server is enabled in the parameters.
-            if context.parameters.observer.is_server_enabled() {
+            let observer_service = if context.parameters.observer.is_server_enabled() {
                 let observer_service = Arc::new(ObserverService::new(
                     context.clone(),
                     core_dispatcher.clone(),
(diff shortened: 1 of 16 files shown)


consensus/core/src/observer_subscriber.rs
```diff
@@ -37,10 +36,42 @@ pub(crate) struct ObserverSubscriber<C: ObserverNetworkClient, S: ObserverNetwor
...
+    retired: Vec<ObserverSubscription>,
+}
+
+struct ObserverSubscription {
+    // Signals the subscription task to stop cooperatively, instead of aborting it. Aborting
+    // would drop the block handler JoinSet owned by the task, which aborts the handlers without
+    // awaiting them: they could briefly keep running (holding strong references to the observer
+    // service) and any panic recorded in the set would be discarded. On this signal the task
+    // stops streaming and drains the JoinSet instead, so join() returns only after every block
+    // handler has finished and handler panics have propagated.
```
arun-koshy: because you abort first the block handlers wont actually "finish" and complete there work right? So its really just canceling the handler and then awaiting the cancellation

consensus/core/src/observer_subscriber.rs
```diff
@@ -259,36 +345,35 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
...
-            let remaining = active_tasks.load(Ordering::Acquire);
-            if remaining > 0 {
-                warn!(
-                    "Warning: {} tasks still marked as active after cleanup",
-                    remaining
-                );
+    fn handle_task_result(result: Option<Result<(), JoinError>>) {
+        if let Some(Err(error)) = result {
+            if error.is_panic() {
+                std::panic::resume_unwind(error.into_panic());
```
arun-koshy, line 358: will this silently kill the live observers block stream? It previously warned on these errors and streaming continued

consensus/core/src/task.rs
```diff
@@ -0,0 +1,42 @@
...
+        std::panic::resume_unwind(error.into_panic());
+    }
+}
+
+/// Aborts all tasks in the set and awaits them, propagating any panic.
+/// Task cancellations are ignored.
+pub(crate) async fn shutdown_join_set<T: 'static>(tasks: &mut JoinSet<T>) {
+    tasks.abort_all();
+    while let Some(result) = tasks.join_next().await {
+        if let Err(error) = result
```
arun-koshy, line 24: what do you think about draining the whole set instead of stopping at the first panic?

consensus/core/src/authority_node.rs
```diff
@@ -531,47 +550,75 @@ where
...
-        };
-        self.commit_syncer_handle.stop().await;
-        if let Some(round_prober_handle) = self.round_prober_handle {
+        synchronizer.stop().await;
+        commit_syncer_handle.stop().await;
+        if let Some(round_prober_handle) = round_prober_handle {
             round_prober_handle.stop().await;
         }
-        self.leader_timeout_handle.stop().await;
+        leader_timeout_handle.stop().await;
```
arun-koshy, line 594: LeaderTimeoutTaskHandle::stop() and inflight_fetches.shutdown() still swallow panics, might be worth migrating them to task.rs helpers

consensus/core/src/synchronizer.rs
arun-koshy, line 491: AI review flagged this early return as potentially breaking the patterns you are trying to fix in this PR because the "JoinSet drop aborts but doesnt await"

consensus/core/src/observer_subscriber.rs
mwtian: Good catch — the handlers are cancelled and awaited, not run to completion. Reworded the comment: on the shutdown signal the task stops streaming, then cancels the remaining block handlers and awaits their termination.

consensus/core/src/task.rs
mwtian: Done — `shutdown_join_set()` now drains the entire set first, remembering the first panic, and resumes unwinding only after the drain, so no task is dropped without being awaited.

consensus/core/src/authority_node.rs
mwtian: Done — migrated `LeaderTimeoutTaskHandle::stop()` and both `inflight_fetches.shutdown()` call sites to the task.rs helpers, plus `RoundProberHandle::stop()` which hand-rolled the same shape. A sweep of consensus/core shows no remaining `.await.ok()` / `JoinSet::shutdown()` on stop paths; `CommitFinalizer`'s chunk-collection loop is left as-is since it intentionally logs and returns on cancellation.

consensus/core/src/synchronizer.rs
mwtian: Fixed — extracted `Synchronizer::shutdown_tasks()` and call it from both loop exits (the `Shutdown` command and the early return when Core is shutting down), so the JoinSets are always drained before `run()` exits. The commands arm cannot exit the loop on its own (a closed channel just disables that select arm), so those are the only two exits.

consensus/core/src/observer_subscriber.rs
mwtian: This does not silently kill. The panic is resumed loudly in tests. This is desirable because panics in `handle_block()` would surface bugs in tests. In production, panics abort the whole process so this path is irrelevant.

arun-koshy (approved):

PR #27326, opened 2026-07-17
[tracing] support per-instruction/event filtering

tzakian:
## Description 

Replace `wants_effects` with a per-instruction `EventFilter` which can provide more precise upstream filtering/determination if an event should be created.

The `per_instruction_filter` is called for each instruction with the bytecode and current pc. It can then return:
- `None` -- Do not keep any of this instructions effects (instruction event will still be created though).
- Some(EventFilter)` -- Keep this instructions effects, based on the `EventFilter` function returned.

See the comments on the `EventFilter` type but the returned function is called with the relative index to the instruction where the instruction is at index 0 (so negative for pre-effects, and positive for post-effects).

E.g., if you wanted to only keep the result of `Add` instructions, and the arguments to the `Sub` instructions, and not keep any any others you could do so with something like the following:

```rust
fn instruction_filter_example(instruction: &Opcodes, _pc: u16) -> Option<EventFilter> {
     match instruction {
         // Only interested in the post effects of an `Add` instruction
         Opcodes::ADD => Some(|relative_event_index: i16| relative_event_index > 0),
         // Only interested in the pre effects of a `Sub` instruction
         Opcodes::SUB => Some(|relative_event_index: i16| relative_event_index < 0),
         // Otherwise, we don't care about the effects of this instruction
         _ => None,
     }
}
```

## Test plan 

CI/existing tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
@@ -48,7 +48,6 @@ pub(crate) struct VMTracer<'a> {
     type_stack: Vec<StackType>,
     loaded_data: BTreeMap<TraceIndex, GlobalValue>,
     effects: Vec<EF>,
-    wants_effects: bool,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
```diff
@@ -48,7 +48,6 @@ pub(crate) struct VMTracer<'a> {
     type_stack: Vec<StackType>,
     loaded_data: BTreeMap<TraceIndex, GlobalValue>,
     effects: Vec<EF>,
-    wants_effects: bool,
```
dariorussi, line 51: 👍

dariorussi (approved):
makes sense as we talked about

PR #27328, opened 2026-07-17
[random beacon] allow DKG to complete after timeout, re-enabling randomness mid-epoch

aschran:
## Description

Today, if random-beacon DKG fails to complete by `random_beacon_dkg_timeout_round`, randomness is disabled for the entire epoch and all randomness-using transactions are canceled until the next epoch. This lets DKG keep running past the timeout (gated on the new `allow_dkg_completion_after_timeout` protocol flag): transactions are still canceled while DKG is incomplete, but if DKG completes later in the epoch, randomness re-enables at the following commit instead of forcing users to wait for a new epoch.

Rather than persisting a terminal failure, "timed out" is now derived per commit from `(dkg_output, commit round)`, so the DKG state machine is never stopped and can complete late. Epoch-close liveness is preserved: randomness-using transactions are canceled (not deferred) during the timed-out window, so nothing new blocks reconfiguration, and a late completion follows the existing successful-DKG close path. `DkgStatus::Failed` is renamed to `DkgStatus::TimedOut` to reflect that the state is no longer terminal.

## Test plan

New unit tests in `sui-core` cover late completion, restart-then-complete, confirmation re-send on restart, and the round-derived status. New `sui-e2e-tests/tests/randomness_tests.rs` simtests cover (a) timeout → cancellation → mid-epoch recovery → epoch change, and (b) timeout with no recovery, asserting epoch change still succeeds. `sui-benchmark` DKG simtests additionally assert reconfiguration completes.

## Release notes

- [x] Protocol: Under protocol version 131, random-beacon DKG that misses its timeout round now keeps running and can re-enable randomness mid-epoch if it completes, instead of disabling randomness for the whole epoch.


crates/sui-core/src/epoch/randomness.rs
```diff
@@ -960,11 +997,73 @@ impl RandomnessManager {
...
+
+        if !self.dkg_timeout_reported {
+            error!(
+                "random beacon: DKG timed out; randomness-using transactions will be canceled unless/until DKG completes"
+            );
+            epoch_store.metrics.epoch_random_beacon_dkg_failed.set(1);
+            self.dkg_timeout_reported = true;
+        }
+        DkgStatus::TimedOut
+    }
```
akichidis, line 1044: I believe avoiding the double negate checks `!epoch_store
            .protocol_config()
            .allow_dkg_completion_after_timeout()
            || !Self::is_past_dkg_timeout_round(round, epoch_store.protocol_config())` will help on the code readability. I suggest the following:

```suggestion
pub fn dkg_status_for_commit_round(&mut self, round: Round) -> DkgStatus {
    let status = self.dkg_status();
    if status != DkgStatus::Pending {
        return status;
    }

    let epoch_store = self
        .epoch_store()
        .expect("epoch store must be alive while computing DKG status for a commit");
    let protocol_config = epoch_store.protocol_config();

    if protocol_config.allow_dkg_completion_after_timeout()
        && Self::is_past_dkg_timeout_round(round, protocol_config)
    {
        self.report_dkg_timeout_once(&epoch_store);
        return DkgStatus::TimedOut;
    }

    DkgStatus::Pending
}

fn report_dkg_timeout_once(&mut self, epoch_store: &Arc<AuthorityPerEpochStore>) {
    if self.dkg_timeout_reported {
        return;
    }
    self.dkg_timeout_reported = true;
    error!(
        "random beacon: DKG timed out; randomness-using transactions will be canceled unless/until DKG completes"
    );
    epoch_store.metrics.epoch_random_beacon_dkg_failed.set(1);
}
```

akichidis (approved):

crates/sui-core/src/epoch/randomness.rs
aschran: thanks, simplified as suggested. I kept the error report inline because it's only called from the one spot

mystenmark (approved):

PR #27330, opened 2026-07-17
[adapter] Add original package id function on upgrade cap

tzakian:
## Description 

This adds a new function `original_package_id` on an `UpgradeCap` that returns the original package ID for the package that the upgrade cap controls.

In order to do this, we need to be able to load a package from the object runtime as the upgrade cap doesn't hold this data. However, to be extra careful to make sure that there is no possibility of running into issues with unsequenced reads the API for reading a package from the object runtime is very narrow -- you can only read a package by providing a specific `(version, version_id)` pair. This will return `Some(package)` only if the package with the specified `version` exists at `version_id`. In any other case `None` is returned. 

(Note: this ability to read the package at a specific version in a a safe-yet-unsequenced manner will be important for min-versioning in order to validate the package being set as the min version).

Bottom commit is the addition of `BackingPackageStore` as a super trait of the `RuntimeObjectResolver`, middle commit is the actual implementation, and top commit is snapshots. 

## Test plan 

Added additional tests for this. 

---

## Release notes

- [X] Protocol: Adds a new function to the `sui-framework` to allow getting the original package ID of an `UpgradeCap`. 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-framework/packages/sui-framework/sources/package.move
```diff
@@ -308,6 +314,18 @@ public fun commit_upgrade(cap: &mut UpgradeCap, receipt: UpgradeReceipt) {
...
+/// The original (first-version) ID of the package that this cap authorizes upgrades for.
+public fun original_package_id(cap: &UpgradeCap): ID {
+    let latest = cap.upgrade_package();
+    assert!(latest.to_address() != @0x0, EUpgradeInProgress);
+    assert!(cap.version > 0, EInvalidPackageVersion);
+    object::id_from_address(original_package_id_impl(latest.to_address(), cap.version))
+}
+
+/// Returns the original package id for the given package storage id and version.
+/// The package_id supplied must have the exact `version` otherwise
```
tnowacki: otherwise... what?

tnowacki: I'm not understanding the comment in general

crates/sui-framework/packages/sui-framework/sources/package.move
```diff
@@ -308,6 +314,18 @@ public fun commit_upgrade(cap: &mut UpgradeCap, receipt: UpgradeReceipt) {
     cap.version = cap.version + 1;
 }
 
+/// The original (first-version) ID of the package that this cap authorizes upgrades for.
```
tnowacki, line 317: Let's make sure this comment is consistent with what is in `type_name`, even if it means changing type_name.

crates/sui-types/src/storage/mod.rs
```diff
@@ -208,6 +208,31 @@ pub trait RuntimeObjectResolver {
...
+    /// To be extra careful, we simply return `None` in all cases unless the object is a package
+    /// with the exact version supplied, and let the caller decide how to handle it.
+    fn get_package_at_version(
+        &self,
+        package_id: &ObjectID,
+        package_version: SequenceNumber,
+    ) -> Option<MovePackage> {
+        let package_obj = self.get_package_object(package_id).ok().flatten()?;
+        if package_obj.move_package().version() == package_version {
+            Some(package_obj.move_package().clone())
```
tnowacki: Is there an arc under this? Is this clone efficient?

crates/sui-transactional-test-runner/src/simulator_persisted_store.rs
```diff
@@ -673,6 +673,15 @@ impl ReadStore for PersistedStoreInnerReadOnlyWrapper {
...
 }
 
+impl BackingPackageStore for PersistedStoreInnerReadOnlyWrapper {
+    fn get_package_object(
+        &self,
+        package_id: &ObjectID,
+    ) -> sui_types::error::SuiResult<Option<PackageObject>> {
+        load_package_object_from_object_store(self, package_id)
+    }
+}
```
tnowacki, line 683: This just has me thinking, should we have some additional non-transactional test coverage?

crates/sui-framework/packages/sui-framework/sources/package.move
tzakian: double-checked and matches up 👍🏻

crates/sui-types/src/storage/mod.rs
tzakian: no and no...probably something worth fixing in a different PR that I'll put up :)

crates/sui-transactional-test-runner/src/simulator_persisted_store.rs
tzakian: I think that would be great. I'll definitely work on adding some unit tests .

crates/sui-types/src/storage/mod.rs
tzakian: Clone here is not efficient, but there's a way of making this more efficient. Will update.

crates/sui-framework/packages/sui-framework/sources/package.move
```diff
@@ -66,6 +66,12 @@ const EAlreadyAuthorized: u64 = 2;
 const ENotAuthorized: u64 = 3;
 /// Trying to commit an upgrade to the wrong `UpgradeCap`.
 const EWrongUpgradeCap: u64 = 4;
+/// Trying to read the current package ID from the `UpgradeCap`
+/// while an upgrade is in progress.
+const EUpgradeInProgress: u64 = 5;
+/// Invalid package version in the `UpgradeCap`, or there was a mismatch
+/// between the package ID and version supplied (in the native).
+const EInvalidPackageVersion: u64 = 6;
```
tnowacki, line 74: Is there anyway to exercise this in tests?

tnowacki (approved):
Move side of things LGTM, not sure if more review needs to be done on the resolver side

crates/sui-framework/packages/sui-framework/sources/package.move
tzakian: There is! (gotta use dev-inspect though). Will be adding that here shortly.

tzakian:
I think the resolver side is fine -- no major changes there, so will land :)

PR #27331, opened 2026-07-17
sui-rpc-api: list and subscription streaming metrics

nickvikeras:
## Description

Added a bunch of metrics that AI recommended for GA. Most of these are symmetric with what I added to help debug sui-kv-rpc bottlenecks.

List APIs (ListCheckpoints / ListTransactions / ListEvents)

All labeled {method, resolution} unless noted. method is the handler name; resolution is the coarse response detail derived only from the
validated read mask (checkpoints: summary|transactions|objects, transactions: digest|full|full_objects, events: no_json|json).

- list_first_frame_seconds — handler entry to the first response frame of any kind (data, watermark-only, or terminal). Time to the client's
  first actionable signal; streams that match nothing still count, since a watermark carries the resume cursor.
- list_response_page_bytes — encoded_len() of data-bearing frames only, so the percentiles read as real page sizes. Watermark/terminal frames
  are excluded (they're all tiny and would just pile into the bottom bucket) and counted by:
- list_watermark_frames_total {method} — watermark-only + terminal-only frames emitted.
- list_stream_yield_wait_seconds — from yielding any frame until the handler stream is polled again. Pure downstream-consumption/backpressure
  signal; a stream dropped while suspended records nothing.
- list_render_seconds — building one data item into a response frame. Chunk-level batch reads are excluded; note the checkpoint render includes
  its per-item RocksDB read (hydration and response construction are one function on the fullnode path).
- list_chunk_seconds {method, phase=queue|work} — one blocking scan chunk: queue = time from just before spawn_blocking until the closure runs
  (blocking-pool saturation signal), work = the blocking execution itself, errors included. Exactly one sample of each per chunk.
- list_query_ends_total {method, reason=item_limit|scan_limit|ledger_tip|checkpoint_bound|cursor_bound} — successful streams by effective end
  reason. Errors, cancellation, and client drops never count.
- list_bitmap_buckets_evaluated {method} — total bitmap buckets consumed across all chunks of one successfully completed filtered request.
  Per-request filter cost; unfiltered requests aren't observed.

Subscriptions (SubscribeCheckpoints / SubscribeTransactions / SubscribeEvents)

All labeled {type} (checkpoint|transaction|event) unless noted.

- subscription_payload_messages — data frames delivered, by type.
- subscription_watermark_messages_total — progress-only frames, including a filtered subscription's initial start frame.
- subscription_payload_bytes — encoded_len() of payload frames. Progress-only frames are excluded (counted by the watermark counter above);
  byte percentiles read as payload sizes.
- subscription_stream_yield_wait_seconds — yield of any frame (payload or watermark) until the stream is polled again; slow-consumer signal.
- subscription_terminations_total {type, reason=client_closed|slow_consumer|source_lag|service_shutdown} — every admitted subscription ends
  exactly once here, accounted by an RAII guard owned by the matcher (the tonic stream itself never touches lifecycle metrics). Admission
  rejections are excluded.
- subscription_index_wait_seconds (unlabeled) — time checkpoint dispatch spent waiting for the subscription index to catch up; checkpoints that
  don't wait aren't observed.
- subscription_index_wait_timeouts_total (unlabeled) — waits that hit the 10s bound and dispatched the checkpoint anyway.

Changed, not added: subscription_inflight_subscribers kept its name but went from one unlabeled gauge to {type, filtered} — same total, now
sliceable by stream type and whether a filter is attached.

### DIFF
--- a/crates/sui-rpc-api/src/metrics.rs
+++ b/crates/sui-rpc-api/src/metrics.rs
@@ -2,11 +2,17 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use axum::http;
-use std::{borrow::Cow, collections::HashSet, sync::Arc, time::Instant};
+use std::{
+    borrow::Cow,
+    collections::HashSet,
+    sync::Arc,
+    time::{Duration, Instant},
+};
 
 use prometheus::{
-    HistogramVec, IntCounterVec, IntGauge, IntGaugeVec, Registry,
-    register_histogram_vec_with_registry, register_int_counter_vec_with_registry,
+    Histogram, HistogramTimer, HistogramVec, IntCounter, IntCounterVec, IntGauge, IntGaugeVec,
+    Registry, register_histogram_vec_with_registry, register_histogram_with_registry,
+    register_int_counter_vec_with_registry, register_int_counter_with_registry,
     register_int_gauge_vec_with_registry, register_int_gauge_with_registry,
 };
 use prost::Message;
@@ -18,6 +24,14 @@ pub struct RpcMetrics {
     num_requests: IntCounterVec,
     request_latency: HistogramVec,
     request_handler_latency: HistogramVec,
+    list_first_frame_seconds: HistogramVec,
+    list_response_page_bytes: HistogramVec,
+    list_watermark_frames_total: IntCounterVec,
+    list_stream_yield_wait_seconds: HistogramVec,
+    list_render_seconds: HistogramVec,
+    list_chunk_seconds: HistogramVec,
+    list_query_ends_total: IntCounterVec,
+    list_bitmap_buckets_evaluated: HistogramVec,
 }
 
 const LATENCY_SEC_BUCKETS: &[f64] = &[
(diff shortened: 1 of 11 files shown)


crates/sui-rpc-api/src/metrics.rs
```diff
@@ -18,6 +24,14 @@ pub struct RpcMetrics {
     num_requests: IntCounterVec,
     request_latency: HistogramVec,
     request_handler_latency: HistogramVec,
+    list_first_frame_seconds: HistogramVec,
```
bmwill: Should we just add time to first byte (TTFB) metric to the metrics handler?

crates/sui-rpc-api/src/metrics.rs
```diff
@@ -310,19 +420,191 @@ fn code_as_str(code: tonic::Code) -> &'static str {
     }
 }
 
+#[derive(Clone)]
+pub(crate) struct ListApiMetrics {
```
bmwill: I didnt understood the need for this, then I realized you were able to populate the labels. TIL

bmwill (approved):

PR #27335, opened 2026-07-19
[6/7] Extract unsettled object-withdrawal tracking into a standalone store

lxfind:
## Description

Stack of 7 (builds on [5/7] #27128). The in-execution object-funds sufficiency check ([7/7]) needs every executing node — fullnodes included — to track unsettled object withdrawals, but that bookkeeping lives inside `ObjectFundsChecker`, which is only initialized on validators when `enable_object_funds_withdraw` is on. This PR extracts the bookkeeping (data plus record/GC logic) into a standalone `UnsettledObjectWithdrawals` store, held unconditionally by `AuthorityState` and shared with the checker, so fullnode correctness under the upcoming in-execution check cannot depend on role-aware init gating. Checkpoint-commit GC goes through the store directly instead of the checker.

No functional change: the store is only written by the flag-gated checker path, exactly as before; on nodes where the checker is not initialized it stays empty and GC is a no-op.

## Test plan

Pure refactoring with no intended behavior change, covered by the existing `object_funds_checker` unit and integration tests (unsettled-withdraw accounting, settlement watch, GC of settled versions).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/accumulators/design_docs/object_funds_checking.md
+++ b/crates/sui-core/src/accumulators/design_docs/object_funds_checking.md
@@ -309,7 +309,7 @@ would produce an `InsufficientFundsForWithdraw` early error without executing th
 ## 4. Garbage collection
 
 The unsettled withdrawals tracking must be cleaned up to prevent unbounded memory growth.
-This happens via `ObjectFundsChecker::commit_effects`, called once a batch of effects has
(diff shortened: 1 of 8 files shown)


lxfind:
Folded into #27080 (reworked onto #27403's block-wait base).

PR #27341, opened 2026-07-20
indexer-alt-graphql: fold indexer pipeline config into a single RPC config file

evan-wall-mysten:
## Description

`sui-indexer-alt-graphql` previously learned which indexer pipelines were enabled by merging an arbitrary number of `--indexer-config` files (simple string-list concatenation, not a real merge), separately from its own `--config` file. This replaces that with a single `[pipeline-defaults]`/`[pipeline.<name>]` section on the RPC's existing config file, so the service takes exactly one config file. Per-pipeline `enabled` overrides fall back to a shared default; the default lives in its own section (rather than alongside the per-pipeline map) so that a pipeline can be named anything, including `enabled` -- TOML forbids redefining the same key as both a scalar and a table, so a combined section would make a pipeline named `enabled` inexpressible.

## Test plan

Added unit tests in `sui-indexer-alt-graphql/src/config.rs` covering default-enablement resolution (including the fallback when `[pipeline-defaults]` is omitted) and that unlisted pipelines are never enabled. Manually verified `generate-config` output and end-to-end parsing against the updated example config.

---

## Release notes

- [x] GraphQL: `sui-indexer-alt-graphql`'s `--indexer-config` CLI flag has been removed. Pipeline enablement is now configured via `[pipeline-defaults]`/`[pipeline.<name>]` sections in the existing `--config` file -- see `examples/prod-config/graphql.toml`. A pipeline listed under `[pipeline.<name>]` is enabled by default; set `enabled = false` to opt out.

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/config.rs
+++ b/crates/sui-indexer-alt-graphql/src/config.rs
@@ -43,6 +43,9 @@ pub struct RpcConfig {
 
     /// Configuration for the request-logging extension.
     pub logging: LoggingConfig,
+
+    /// Configuration for which pipelines this service can expect to find populated.
+    pub pipeline: PipelineConfig,
 }
 
 #[derive(Clone, Default, Debug, Deserialize, Serialize)]
@@ -55,6 +58,8 @@ pub struct RpcLayer {
     pub zklogin: ZkLoginLayer,
     pub subscription: SubscriptionLayer,
     pub logging: LoggingLayer,
+    pub pipeline_defaults: PipelineLayer,
+    pub pipeline: BTreeMap<String, PipelineLayer>,
 }
 
 #[derive(Clone)]
@@ -69,12 +74,35 @@ pub struct HealthLayer {
     pub max_checkpoint_lag_ms: Option<u64>,
 }
 
-/// Config for an indexer writing to a database used by this RPC service. It is simplified w.r.t.
-/// to the actual indexer config to focus on extracting the names of pipelines enabled on that
-/// indexer.
-#[derive(Serialize, Deserialize, Default, Clone, Debug)]
-pub struct IndexerConfig {
-    pub pipeline: toml::Table,
+#[derive(Clone, Default, Debug)]
+pub struct PipelineConfig {
+    /// Resolved availability for every pipeline explicitly mentioned in config.
+    pub availabilities: BTreeMap<String, PipelineAvailability>,
+
+    /// Availability for any pipeline not present in `availabilities` above -- e.g. one that
+    /// starts producing watermarks after startup, without ever being explicitly listed.
+    pub default_availability: PipelineAvailability,
+}
+
+/// Whether a pipeline is enabled: tracked by GraphQL and expected to be populated in the database
+/// it reads from.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum PipelineAvailability {
+    #[default]
+    Enabled,
+    Disabled,
+}
+
+/// Configuration for a single pipeline's `enabled` setting -- used both for the shared default
+/// (`[pipeline-defaults]`) and for per-pipeline overrides (`[pipeline.<name>]`). Kept as its own
+/// type/section rather than flattened together with the per-pipeline map: TOML forbids redefining
+/// the same key as both a scalar and a table, so if the default lived directly in the `[pipeline]`
+/// table, a pipeline named `enabled` could never be expressed, regardless of how the Rust side
+/// deserialized it.
+#[derive(Clone, Default, Debug, Deserialize, Serialize)]
+#[serde(default, rename_all = "kebab-case", deny_unknown_fields)]
+pub struct PipelineLayer {
+    pub enabled: Option<bool>,
 }
 
 pub struct Limits {
@@ -346,6 +374,10 @@ impl RpcLayer {
             zklogin: ZkLoginConfig::default().into(),
             subscription: SubscriptionConfig::default().into(),
             logging: LoggingConfig::default().into(),
+            pipeline_defaults: PipelineLayer {
+                enabled: Some(true),
+            },
+            pipeline: BTreeMap::new(),
         }
     }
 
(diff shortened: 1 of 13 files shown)


jessiemongeon1:
## Style Guide Audit

amnn (commented):
IIUC, after this change, we still require operators to opt-in to pipelines, but now that opt-in is via the GraphQL config rather than the indexer config.

This would not be ideal because it's now increasing the burden on operators to keep configs up-to-date. I thought the plan was different: GraphQL would automatically discover pipelines that were available and its config would be used to parametrize them further (e.g. to introduce a max lag, or explicitly disable them)

evan-wall-mysten:
> IIUC, after this change, we still require operators to opt-in to pipelines, but now that opt-in is via the GraphQL config rather than the indexer config.
> 
> This would not be ideal because it's now increasing the burden on operators to keep configs up-to-date. I thought the plan was different: GraphQL would automatically discover pipelines that were available and its config would be used to parametrize them further (e.g. to introduce a max lag, or explicitly disable them)

Yeah that is a follow up PR that I was finishing up https://github.com/MystenLabs/sui/pull/27342. I'll number the related PRs to make this more clear.

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -55,6 +58,8 @@ pub struct RpcLayer {
     pub zklogin: ZkLoginLayer,
     pub subscription: SubscriptionLayer,
     pub logging: LoggingLayer,
+    pub pipeline_defaults: PipelineLayer,
+    pub pipeline: BTreeMap<String, PipelineLayer>,
```
wlmyng, line 62: pipelines?

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -69,12 +74,35 @@ pub struct HealthLayer {
...
-/// Config for an indexer writing to a database used by this RPC service. It is simplified w.r.t.
-/// to the actual indexer config to focus on extracting the names of pipelines enabled on that
-/// indexer.
-#[derive(Serialize, Deserialize, Default, Clone, Debug)]
-pub struct IndexerConfig {
-    pub pipeline: toml::Table,
+#[derive(Clone, Default, Debug)]
+pub struct PipelineConfig {
+    /// Resolved availability for every pipeline explicitly mentioned in config.
+    pub availability: BTreeMap<String, PipelineAvailability>,
```
wlmyng: availabilities?

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -69,12 +74,35 @@ pub struct HealthLayer {
...
+}
+
+/// Whether a pipeline is enabled: tracked by GraphQL and expected to be populated in the database
+/// it reads from.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum PipelineAvailability {
+    #[default]
+    Enabled,
+    Disabled,
+}
```
wlmyng, line 94: do we anticipate further evolution? Or is it just, "is a pipeline available or not"?

crates/sui-indexer-alt-graphql/src/config.rs
```diff
@@ -713,3 +753,127 @@ fn max_across_protocol<T: Ord>(f: impl Fn(&ProtocolConfig) -> Option<T>) -> Opti
...
+    fn pipeline_default_enabled_applies_to_unset_entries() {
+        let layer: RpcLayer = toml::from_str(
+            r#"
+            [pipeline-defaults]
+            enabled = true
+
+            [pipeline.tx_calls]
+
+            [pipeline.kv_objects]
+            enabled = false
```
wlmyng, line 805: what's the difference between naming a pipeline but setting `enabled = false`, vs not naming a pipeline in the config altogether?

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -135,7 +137,7 @@ pub(crate) type WatermarksLock = Arc<RwLock<Arc<Watermarks>>>;
 impl WatermarkTask {
     pub(crate) fn new(
         config: WatermarkConfig,
-        pg_pipelines: Vec<String>,
+        pipeline: PipelineConfig,
```
wlmyng, line 140: pipelines?

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -146,6 +148,13 @@ impl WatermarkTask {
             watermark_polling_interval,
         } = config;
 
+        let pg_pipelines = pipeline
+            .availability
+            .into_iter()
+            .filter(|(_, status)| matches!(status, PipelineAvailability::Enabled))
+            .map(|(name, _)| name)
+            .collect();
```
wlmyng, line 156: could we not do this filtering, or add a method on the config, to give us the enabled pipelines?

crates/sui-indexer-alt-graphql/src/config.rs
evan-wall-mysten: In a follow up there will be something like `MaxCheckpointLimit(u64)` that causes the pipeline to be enabled when within that distance from the tip or furthest ahead pipeline.

crates/sui-indexer-alt-graphql/src/config.rs
evan-wall-mysten: They are the same in this PR, but they will be enabled by default without needing to be specified in the config in https://github.com/MystenLabs/sui/pull/27342.

crates/sui-indexer-alt-graphql/src/task/watermark.rs
evan-wall-mysten: This is changed in https://github.com/MystenLabs/sui/pull/27342 so I am leaving it inlined here.

crates/sui-indexer-alt-graphql/src/task/watermark.rs
evan-wall-mysten: Except for `Limits`, the other `*Config` structs are singular https://github.com/MystenLabs/sui/blob/26c78168d2be95e0686b8a604b3ad0ec763829c2/crates/sui-indexer-alt-graphql/src/config.rs#L25-L46

crates/sui-indexer-alt-graphql/src/config.rs
evan-wall-mysten: This would cause the config to be different from `sui-indexer-alt` where it is singular.

PR #27342, opened 2026-07-20
indexer-alt-graphql: auto-discover pipelines from the watermarks table

evan-wall-mysten:
## Description

Stacked on #27341. `sui-indexer-alt-graphql` currently only enables pipelines explicitly listed in its config file, so a pipeline that's running and writing watermark rows still needs a matching `[pipeline.<name>]` entry to be picked up. This adds a second source: at startup, the service queries the `watermarks` table directly (retrying until the database is reachable) to discover any pipeline that has ever run. Discovered pipelines are enabled by default, same as config-listed ones, but explicit config always wins -- a `[pipeline.<name>] enabled = false` (or a global `[pipeline-defaults] enabled = false`) still disables a pipeline even if it was auto-discovered.

## Test plan

Added 5 unit tests in `sui-indexer-alt-graphql/src/config.rs` covering all 9 meaningful (discovery x config-state) combinations for pipeline resolution: discovered-and-unconfigured enabled by default (both directions of the global default), config override winning over discovery in both directions, and discovered/configured pipelines combining correctly.

Manually verified end-to-end against a local Postgres: created a `watermarks` table with rows for two pipelines and no config declaring them, confirmed both were picked up (visible in the resulting watermark-polling query's bind parameters); then added an explicit `enabled = false` override for one and confirmed it was excluded while the other remained enabled.

---

## Release notes

- [x] GraphQL: `sui-indexer-alt-graphql` now also auto-discovers pipelines from the database's `watermarks` table at startup, in addition to its config file. Auto-discovered pipelines are enabled by default; explicit `[pipeline.<name>]`/`[pipeline-defaults]` config entries still override this.

### DIFF
--- a/crates/sui-indexer-alt-graphql/README.md
+++ b/crates/sui-indexer-alt-graphql/README.md
@@ -82,10 +82,16 @@ the defaults need to be changed:
 cargo run --bin sui-indexer-alt-graphql -- generate-config > graphql_config.toml
 ```
 
-GraphQL also needs access to the configuration of the indexer(s) that are
-writing to the database it is reading from. It uses these to understand which
-pipelines are being populated and therefore what features should be available
-in the database.
+The config file's `[pipeline-defaults]` and `[pipeline.<name>]` sections tell
+GraphQL which pipelines it can expect to find populated in the database it is
+reading from, so it knows which watermarks to poll for and what features to
+enable. List each pipeline you expect to be populated as `[pipeline.<name>]`
+(e.g. `[pipeline.tx_calls]`); pipelines listed this way are enabled by default.
+Set `enabled = false` under `[pipeline-defaults]` to disable every pipeline
+listed below by default instead, or on a specific `[pipeline.<name>]` entry to
+opt just that one out (or back in, if the default has been flipped to
+disabled). Pipelines that aren't listed at all are never considered enabled.
+See `examples/prod-config/graphql.toml` for an example.
 
 ## Running
 
@@ -93,7 +99,7 @@ The service can be run with the following minimal command:
 
 ```sh
 cargo run --bin sui-indexer-alt-graphql -- rpc \
-  --indexer-config indexer_alt_config.toml
+  --config graphql_config.toml
 ```
 
 In this configuration, the RPC will respond at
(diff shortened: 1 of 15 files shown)


crates/sui-indexer-alt-graphql/README.md
```diff
@@ -82,16 +82,14 @@ the defaults need to be changed:
...
-disabled). Pipelines that aren't listed at all are never considered enabled.
-See `examples/prod-config/graphql.toml` for an example.
+GraphQL auto-discovers pipelines directly from the database's `watermarks`
+table: any pipeline with a watermark row is picked up on the next poll and
+enabled by default, whether or not it's explicitly listed in the config file.
+The `[pipeline-defaults]` and `[pipeline.<name>]` sections control that
+default and let you override it. Set `enabled = false` under
+`[pipeline-defaults]` to disable every pipeline by default instead, and add a
+`[pipeline.<name>]` entry (e.g. `[pipeline.tx_calls]`) to opt a specific
+pipeline out (or back in, if the default has been flipped to disabled).
```
wlmyng, line 92: Hmm... one thing operators don't have to worry about today are outdated entries on the watermarks table. For example, if I created one or more tasked tx_calls for various patches, I don't need to do any clean up because I've explicitly configured graphql to pay attention to `tx_calls`. If graphql reads directly from the watermarks table though, now I have (a) update the config to exclude all these historical markers or (b) remove them from the watermarks table.

I think there's a benefit to keeping those runs for record-keeping, which makes (a) the only realistic option

crates/sui-indexer-alt-graphql/src/task/watermark.rs
```diff
@@ -69,8 +67,9 @@ pub(crate) struct WatermarkTask {
     /// How long to wait between updating the watermark.
     interval: Duration,
 
-    /// Pipelines that we want to check the watermark for.
-    pg_pipelines: Vec<String>,
+    /// Configuration for which pipelines are enabled, used to determine which pipelines to check
```
wlmyng, line 70: i think this comment is not entirely correct, in that the watermarks table is the full set of pipelines, and the config masks over that to indicate which ones to exclude or track

PR #27345, opened 2026-07-20
[autodebug] test-cluster: resubmit on retryable consensus rejection in submit_and_execute

alex-mysten:
## Description

Fixes the nightly simtest flake in `receive_of_object_with_reconfiguration_receive_after_reconfig` ([failing run](https://github.com/MystenLabs/sui/actions/runs/29733469834), seed 1784541660 at c7aa4874f8).

A transaction can be indirectly rejected at commit time when its block fails to finalize within `INDIRECT_REJECT_DEPTH` (3) leader rounds — in this trace, the test submitted the post-reconfig `receive` 112 ms after reconfiguration completed, while consensus dissemination was still warming up: the proposer's block gathered 3 accept votes and 0 reject votes but was linked into peers' DAGs ~5 rounds late and missed the finalization window. This outcome is retryable by design — the production `TransactionDriver` maps it (`Rejected` with no reject-vote reason → `RejectedByConsensus` → `ErrorCategory::Aborted`) to a resubmission. `TestCluster::submit_and_execute` already resubmits on the sibling transient outcomes (`Expired` positions, transient wait errors; #26965) but hard-failed on `Rejected`, surfacing the generic "Transaction was rejected" error.

## Test Plan

- CI seed 1784541660 with CI-parity build: fails deterministically unfixed, passes fixed.
- Seed search: 800/800 pass on simtest-01.

## Release notes

Check each box that your changes affect. If none apply, release notes aren't required.

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/test-cluster/src/lib.rs
+++ b/crates/test-cluster/src/lib.rs
@@ -1098,6 +1098,7 @@ impl TestCluster {
             };
 
             let mut consensus_position = None;
+            let mut rejected_transient = false;
             for result in submit_response.results {
                 match result {
                     SubmitTxResult::Executed { details, .. } => {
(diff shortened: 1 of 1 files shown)


mwtian (approved):
This doesn't look incorrect but as you can see this will not have the same coverage as Transaction Driver. In future, I think we should look into using transaction driver here directly.

PR #27346, opened 2026-07-20
[move][move-vm-runtime] Type size formulae instead of type construction and traversal.

cgswords:
## Description

There are many places in the VM today where we build a type just to check safety properties. This PR attempts to avoid those creations whenever possible, opting instead for statically-known type size formulae that can be used to check them directly. This avoids doing so much work at runtime. 

**NOTE:** This will need to be feature-gated before rollout, but is not feature gated in the draft to make it easier to review the changes.

## Test plan

- `subst_formula_tests` covers the load-bearing algebra directly.
- `loader_tests` exercises end-to-end formula resolution through a real linkage.
- CI and everything else is unchanged.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
```diff
@@ -736,60 +747,63 @@ fn op_step_impl(
...
             let value = Vector::pack(specialization, elements)?;
             state.push_operand(value)?;
         }
         Bytecode::VecLen(ty_ptr) => {
             let vec_ref = state.pop_operand_as::<VectorRef>()?;
-            let ty = instantiate_single_type(ty_ptr, state.call_stack.current_frame.ty_args())?;
+            let specialization = vector_spec(ty_ptr, state.call_stack.current_frame.ty_args())?;
             gas_meter.charge_vec_len()?;
-            let value = vec_ref.len(&ty)?;
+            let value = vec_ref.len_internal(specialization)?;
```
tzakian: nit on naming (possibly?): thoughts on calling these functions `len_specialized` `borrow_elem_specialized` (or similar)? Might be a smidge cleaner as to what their purpose is over `_internal` possibly?

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
```diff
@@ -1150,76 +1164,28 @@ fn partial_error_to_error<T>(
...
+fn vector_spec(elem: &ArenaType, ty_args: &TypeArguments) -> PartialVMResult<VectorSpecialization> {
+    VectorSpecialization::from_element(elem, ty_args.types())
+}
+
+/// Check that a concrete value type's nesting depth is within the configured limit before a
+/// value of it is created (`Pack`/`PackVariant`). Solved from the type's resolved formula — the
+/// type is never walked field by field. `ty` is depth-bounded by construction, so `type_size_of`'s
+/// recursion over it is safe.
+fn check_value_depth_of_type(run_context: &mut RunContext, ty: &Type) -> PartialVMResult<()> {
     let Some(max_depth) = run_context
```
tzakian: JFYI/for future: I think we can change this to a `safe_unwrap!` as this value should never be unset in the new VM.

This gets back to the config flattening, and probably best to leave this as-is for now and then we can simplify as part of the config flattening later.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -37,49 +41,86 @@ use move_core_types::{
...
+/// from the (cached) `Sync` tables and dropped at the end. Because it is never shared across
+/// threads, its size cache can be unsynchronized. It derefs to [`DispatchTables`], so all the
+/// resolution methods read the shared tables directly.
+#[derive(Debug)]
+pub(crate) struct VMDispatchTables {
+    pub(crate) tables: DispatchTables,
+    pub(crate) size_formulas: TypeCache,
+}
+
+impl Deref for VMDispatchTables {
```
tzakian, line 136: Interesting.... 👀

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -512,114 +550,191 @@ impl VMDispatchTables {
...
+        self.arena_type_size_formula_impl(ty, &mut Visiting::new())
+    }
 
-                    datatype_formula.subst(ty_arg_map)?
-                }
-            })
+    pub(crate) fn arena_type_size_formula_impl(
+        &self,
+        ty: &ArenaType,
+        visiting: &mut Visiting,
```
tzakian: same here -- will need to pass budget since we are doing a deep traversal of the type.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -512,114 +550,191 @@ impl VMDispatchTables {
...
     }
 
+    /// The four sizes of a concrete runtime type, assuming no free parameters. Datatype nodes
+    /// resolve through the cache.
+    ///
+    /// Note: `ty` is depth-bound by construction here.
+    pub(crate) fn type_size_of(&self, ty: &Type) -> PartialVMResult<TypeSize> {
+        Ok(match ty {
+            Type::Vector(inner) | Type::Reference(inner) | Type::MutableReference(inner) => {
+                TypeSize::wrap(self.type_size_of(inner)?)
```
tzakian: nit: 
```suggestion
               self.type_size_of(inner)?.wrap()
```

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -512,114 +550,191 @@ impl VMDispatchTables {
...
-    ) -> PartialVMResult<DepthFormula> {
-        type_size.check()?;
-        // If we've already computed this datatypes depth, no more work remains to be done.
-        if let Some(form) = self.cached_type_depth(datatype_name) {
-            return Ok(form.clone());
+    /// `visiting` turns a cyclic datatype graph into an invariant violation.
+    fn virtual_key_size_formula_impl(
+        &self,
+        key: &VirtualTableKey,
+        visiting: &mut Visiting,
```
tzakian: I think this (and `arena_type_size_formula` and related threading) should take a `TypeTraversalBudget` or something similar?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
```diff
@@ -255,8 +254,6 @@ where
                     self.convert_linked_vm_error(e, &linkage)
                 }
             })?;
-        let runtime_signature = subst_signature(runtime_signature, &loaded_type_arguments)
```
tzakian, line 258: Given this, can we remove `TyParam` completely from the `Type` enum in the execution AST so that `Type` is always fully instantiated?

tzakian: If not, could you add a debug_assert! traversal (and/or invariant check) that `runtime_signature` doesn't contain any ty params?

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,623 @@
...
+//! - `value_depth`: the depth of a *value* of the type, through datatype fields;
+//! - `layout_size`: the node count of the type's generated layout, through datatype fields.
+//!
+//! Rather than realize a type and measure it, we predict each quantity with a closed-form
+//! formula and check the prediction, so rejection is arithmetic and no oversized type, value,
+//! or layout is ever built.
+//!
+//! Each quantity lives in one of two algebras — additive ([`LinearForm`], for `type_size` and
+//! `layout_size`) and max-plus ([`MaxPlusForm`], for `type_depth` and `value_depth`) — both flat
+//! and closed under substitution. The pipeline is a two-stage partial evaluator over three
```
tzakian: 😍

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,623 @@
...
+};
+use move_binary_format::{
+    checked_as,
+    errors::{PartialVMError, PartialVMResult},
+    partial_vm_error,
+};
+use std::collections::HashSet;
+
+/// Index of a type parameter.
+pub(crate) type TyParamIndex = u16;
```
tzakian: thoughts on reusing the type alias from the Move binary format?

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,623 @@
...
+            type_depth: inner.type_depth.saturating_add(1),
+            value_depth: inner.value_depth.saturating_add(1),
+            layout_size: inner.layout_size.saturating_add(1),
+        }
+    }
+}
+
+/// Check a solved `(type_size, type_depth)` against the type-traversal limits: depth first,
+/// then size.
+pub(crate) fn check_syntactic_limits(type_size: u64, type_depth: u64) -> PartialVMResult<()> {
```
tzakian, line 95: Fine to keep where it is, but also maybe consider moving this to be along the other limits checks on the dispatch tables -- eventually once we move these constants to be config values we'll need to do that.

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,623 @@
...
+pub(crate) struct LinearTerm {
+    pub(crate) param: TyParamIndex,
+    pub(crate) coefficient: u64,
+}
+
+/// One term of a [`MaxPlusForm`]: `offset + x_param`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct MaxPlusTerm {
+    pub(crate) param: TyParamIndex,
+    pub(crate) offset: u64,
```
tzakian: Nit: consistent order for the fields plz 🙏🏻

tzakian, line 162: offset feels like a weird name to me? But I can't think of a better one either...

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// Example: `self = max(1, 2+x0, x1)`, `args = [3, 5]` gives `max(1, 2+3, 5) = 5`.
+    pub(crate) fn solve(&self, args: &[u64]) -> PartialVMResult<u64> {
+        let mut acc = self.constant;
+        for term in &self.terms {
+            let value = args
+                .get(term.param as usize)
+                .ok_or_else(|| out_of_bounds_parameter(term.param, args.len()))?;
+            acc = acc.max(term.offset.saturating_add(*value));
+        }
+        Ok(acc)
```
tzakian: thoughts on either adding a debug assert or invariant violation if the solved value == 0?

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,623 @@
...
+                visit_field(
+                    field,
+                    1,
+                    &mut value_depth_local,
+                    &mut layout_size_local,
+                    &mut apps,
+                );
+            }
+            value_depth_local.canonicalize();
+            layout_size_local.canonicalize();
```
tzakian, line 923: ostensibly if the fields are private there's not need to canonicalize at this point I believe? But fine to keep just for sanity sake.

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,623 @@
...
+// -------------------------------------------------------------------------------------------------
+
+/// A type's four size formulae over some parameters, fully resolved and flat (no pending
+/// datatype applications). This is the value cached per datatype key on the vtable, and the
+/// intermediate the interpreter threads while resolving a term.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct PartialTypeSizeFormula {
+    pub(crate) type_size: LinearForm,
+    pub(crate) type_depth: MaxPlusForm,
+    pub(crate) value_depth: MaxPlusForm,
```
tzakian, line 402: uber nit: consistant ordering of these fields across here and `ArenaTypeSizeFormula` 🙏🏻

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,623 @@
...
+                        );
+                    }
+                    ArenaType::Datatype(_) | ArenaType::DatatypeInstantiation(_) => {
+                        apps.push(ArenaApply {
+                            field_type: VMPointer::from_ref(ty),
+                            value_depth_offset: prefix_depth,
+                            layout_size_coeff: 1,
+                        });
+                    }
+                    _ => {
```
tzakian: Nit: exhaustive match 🙏🏻

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,623 @@
...
+/// `type_size`/`type_depth` are closed. `value_depth`/`layout_size` are the `*_local` part (the
+/// contribution of the datatype's primitive/parameter/vector field structure) plus `apps` (the
+/// datatype-application fields), resolved by [`substitute`](Self::substitute).
+#[derive(Debug)]
+pub(crate) struct ArenaTypeSizeFormula {
+    pub(crate) type_size: LinearForm,
+    pub(crate) type_depth: MaxPlusForm,
+    pub(crate) value_depth_local: MaxPlusForm,
+    pub(crate) layout_size_local: LinearForm,
+    pub(crate) apps: ArenaVec<ArenaApply>,
```
tzakian: Can (should?) these fields be private? Feels like it would be nice if they could be as they are all w.r.t. the resolution context of `apps`.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -512,114 +550,191 @@ impl VMDispatchTables {
...
+            | Type::U128
+            | Type::U256
+            | Type::Address
+            | Type::Signer => 1,
+        })
+    }
+
+    /// Check a computed value's `value_depth` against the configured limit (if any).
+    fn check_value_depth(&self, value_depth: u64) -> PartialVMResult<()> {
+        if let Some(max) = self.vm_config.runtime_limits_config.max_value_nest_depth
```
tzakian: IMO we can `safe_unwrap!` this -- this should always be set, and set long before this execution version.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -512,114 +550,191 @@ impl VMDispatchTables {
...
+            | Type::U128
+            | Type::U256
+            | Type::Address
+            | Type::Signer => TypeSize::PRIMITIVE,
+        })
+    }
+
+    /// The `value_depth` of a concrete type — the single measure `Pack`/`PackVariant` need,
+    /// computed without the other three. `ty` is depth-bound by construction.
+    pub(crate) fn value_depth_of(&self, ty: &Type) -> PartialVMResult<u64> {
```
tzakian, line 703: Similarly here, I think we may need to pass a budget too(??) although am less certain.

external-crates/move/crates/move-vm-runtime/src/execution/interpreter/eval.rs
cgswords: Went ahead and `safe_unwrap!`'d this everywhere `max_value_nest_depth` is read, for uniformity — it's always set at any protocol version this VM runs under. Any broader config flattening can still simplify the Option away later.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
cgswords: This code was restructured: formula resolution is now an iterative worklist over vtable keys with cycle detection, and never traverses a type at all (dependencies come from the JIT-compiled `vtable_keys` list). The type walks that remain (`load_type`, `abilities`, `type_to_type_tag`, `type_size_of`, `value_depth_of`) all thread a `TypeTraversalBudget`; layout generation and substitution are instead bounded up front by their formula checks (`check_layout_limits` / `realize_type`) before anything is built.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
cgswords: Same as the other budget thread — the recursive `_impl` functions (and their `Visiting` threading) no longer exist; resolution is the linearized worklist and does no type traversal.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/env.rs
cgswords: Done, on both sides: the adapter's execution AST `Type` has no `TyParam` variant, and the VM's runtime `Type` enum no longer has one either: nothing constructed it, so runtime types are now fully instantiated by construction, and this arm is gone entirely. `ArenaType::to_type` raises an invariant violation if asked to convert an uninstantiated ty param (`ArenaType` keeps its `TyParam` — uninstantiated signatures and size formulas genuinely range over parameters). Split into its own commit (d90a0818a8) for easier review.

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: Keeping it here for now; agreed it should migrate to the dispatch-tables limit checks when these constants become config values.

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: Kept `offset`, since it's the name used in a max-plus algebraic semiring. 🤠

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: Added — `solve` debug-asserts a nonzero result (both form flavors), and the algebra-test generators now only produce forms/arguments with the real-formula invariant (everything >= 1).

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: Kept for sanity: canonicalized formulae substitute a little more nicely / predictably into each other, and canonical order keeps `Debug` output and any future structural comparison deterministic.

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: Done every place `ArenaType` is matched.

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: All fields are private now. The four forms also became arena-backed flavors in the process, to avoid some deallocation messiness.

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
cgswords: Added a `TypeTraversalBudget` here and on `type_size_of`, so every `Type` walk is metered the same way.

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -104,28 +102,63 @@ pub(crate) fn check_syntactic_limits(type_size: u64, type_depth: u64) -> Partial
...
+    let mut acc = constant;
+    for term in terms {
+        let value = args
+            .get(term.param as usize)
+            .ok_or_else(|| out_of_bounds_parameter(term.param, args.len()))?;
+        acc = acc.saturating_add(term.coefficient.saturating_mul(*value));
+    }
+    // Every measure of a real type is at least one node/level, and argument measures are
+    // themselves solved measures, so a zero here means a broken formula.
+    debug_assert_ne!(acc, 0, "linear form solved to zero");
```
tzakian, line 125: invariant violation for here and below in this case?

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
```diff
@@ -512,114 +562,212 @@ impl VMDispatchTables {
...
+    // with the cache consulted in the loop -- no recursion through type structure. Type terms
+    // carry their own JIT-compiled formulae (`SizedArenaType`) and close the same way
+    // (`term_size_formula`); `type_size_of` then solves a concrete runtime type to a `TypeSize`.
+
+    /// The resolved formula of a datatype under this linkage: an iterative DFS over the datatype
+    /// dependency graph -- keys staged on an explicit work stack, resolved bottom-up through the
+    /// cache, each formula closed by one `evaluate` pass on the way back up.
+    ///
+    /// Example: resolving `W` from `struct W<A> { x: u64, y: T<u64>, z: vector<R<S<A>>> }`
+    /// (`T`, `R`, `S` datatypes, none cached) stages `[W, T, S, R]`; `T`, `S`, and `R` have no
```
tzakian, line 580: Is this W,T,S,R or W,T,R,S? The latter seems right to me but maybe I'm missing something in the ordering

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,1162 @@
...
+/// contribution of the datatype's primitive/parameter/vector field structure) plus the
+/// datatype-application fields, carried symbolically until a linkage resolves them: given the
+/// resolved formula of every datatype in `keys`, [`evaluate`](Self::evaluate) computes this
+/// formula's own flat form directly.
+///
+/// Example: `struct W<A> { x: u64, y: T<u64>, z: vector<R<S<A>>> }` (`T`, `R`, `S` datatypes)
+/// builds
+///
+/// ```text
+/// keys         = [T, S, R]                       // first-mention order
```
tzakian, line 589: I'm confused by the ordering here, I would have expected T,R,S?

tzakian: Is this a stack not in the Rust sense? i.e., I view the "top" of the stack as the last element.

But if that's the case, it seems (even more) at odds with the ordering in the comment in `dispatch_tables` so some clarification/a bit more specificity on the ordering may be nice.

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,1162 @@
...
+    fn new() -> Self {
+        Self {
+            keys: vec![],
+            applications: vec![],
+        }
+    }
+
+    /// The index of `key` in the deduplicated key list, appending it on first mention.
+    fn intern_key(&mut self, key: &VirtualTableKey) -> PartialVMResult<KeyIndex> {
+        let ndx = match self.keys.iter().position(|k| k == key) {
```
tzakian, line 641: Nit: an `IndexMap` here _may_ be more performant?

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,1162 @@
...
+    /// then `R<r0>` and returns `R`'s index; the argument `vector<vector<A>>` compiles to
+    /// `{ vector_layers: 2, base: TypeParameter(A) }`.
+    fn emit_application(&mut self, ty: &ArenaType) -> PartialVMResult<ApplicationIndex> {
+        // One work item: compile a type into an `Argument`, or assemble an application from
+        // the last `argc` compiled arguments (its own, in order).
+        enum Item<'a> {
+            Compile(&'a ArenaType),
+            Assemble {
+                key: &'a VirtualTableKey,
+                argc: usize,
```
tzakian, line 665: Contrary to popular belief characters are free ;)

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,1162 @@
...
+        while let Some(item) = work.pop() {
+            match item {
+                Item::Compile(mut ty) => {
+                    // Strip the vector/reference layers, then dispatch on the base.
+                    let mut vector_layers: usize = 0;
+                    while let ArenaType::Vector(inner)
+                    | ArenaType::Reference(inner)
+                    | ArenaType::MutableReference(inner) = ty
+                    {
+                        vector_layers = vector_layers.saturating_add(1);
```
tzakian, line 681: mm I don't like this name anymore...

`primitive_layers` or `primitive_wraps` or something like that?

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,1162 @@
...
+                    argc,
+                    vector_layers,
+                } => {
+                    let split = compiled.len().checked_sub(argc).ok_or_else(|| {
+                        partial_vm_error!(
+                            UNKNOWN_INVARIANT_VIOLATION_ERROR,
+                            "argument underflow while linearizing a size formula"
+                        )
+                    })?;
+                    let arguments = compiled.split_off(split);
```
tzakian, line 741: Nit: debug assert (or invariant violation) that `split <= compiled.len()` above this 🙏🏻

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,1162 @@
...
+    }
+
+    /// Emit the applications of a datatype-application type in post-order (arguments before the
+    /// application that consumes them -- the linearization invariant) and return the root's
+    /// index. Runs on an explicit work stack: nothing recurs, however deeply the type nests.
+    ///
+    /// Example: for datatypes `R` and `S` and enclosing parameter `A`, `R<S<A>>` emits `S<x0>`
+    /// then `R<r0>` and returns `R`'s index; the argument `vector<vector<A>>` compiles to
+    /// `{ vector_layers: 2, base: TypeParameter(A) }`.
+    fn emit_application(&mut self, ty: &ArenaType) -> PartialVMResult<ApplicationIndex> {
```
tzakian, line 658: nit on name: since this is only meant to be called on datatypes, maybe `emit_datatype_application`?

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,1162 @@
...
+/// and its `R<S<A>>` is emitted at depth 2. Had `W` had a field `A` instead, it would add the
+/// terms `1 + x0` (depth) and `x0` (layout).
+fn visit_field(
+    mut ty: &ArenaType,
+    mut prefix_depth: u64,
+    value_depth_local: &mut MaxPlusForm,
+    layout_size_local: &mut LinearForm,
+    linearizer: &mut Linearizer,
+) -> PartialVMResult<()> {
+    loop {
```
tzakian, line 811: This should 1000% always terminate. But my paranoia says this should just be a massive while loop (or have some bound on it) and invariant violate if we breach that bound.

But I realize that this is probably overly paranoid, but just something worth considering.

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
```diff
@@ -0,0 +1,1162 @@
...
+                .map(|argument| -> PartialVMResult<PartialTypeSizeFormula> {
+                    let base = match argument.base {
+                        ArgumentBase::Primitive => PartialTypeSizeFormula::primitive(),
+                        ArgumentBase::TypeParameter(idx) => PartialTypeSizeFormula::parameter(idx),
+                        ArgumentBase::Application(ndx) => results
+                            .get(ndx as usize)
+                            .cloned()
+                            .ok_or_else(|| broken("out-of-order application reference"))?,
+                    };
+                    Ok((0..argument.vector_layers).fold(base, |formula, _| formula.wrap()))
```
tzakian, line 1136: This feels mildly obscene vs having a `wrap_n` that doesn't just call wrap n times...

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords:
```
vec/2 time:   [21.102 ns 21.200 ns 21.296 ns]
 indexmap/2 time:   [90.522 ns 90.742 ns 90.956 ns]
 vec/4 time:   [31.669 ns 31.788 ns 31.908 ns]
 indexmap/4 time:   [176.40 ns 177.23 ns 178.00 ns]
 vec/8 time:   [39.754 ns 39.882 ns 40.007 ns]
 indexmap/8 time:   [359.87 ns 360.70 ns 361.57 ns]
 vec/16 time:   [102.49 ns 102.78 ns 103.07 ns]
 indexmap/16 time:   [728.48 ns 730.31 ns 732.13 ns]
 vec/32 time:   [252.42 ns 253.07 ns 253.75 ns]
 indexmap/32 time:   [1.4413 µs 1.4491 µs 1.4561 µs]
 vec/64 time:   [957.55 ns 959.93 ns 962.26 ns]
 indexmap/64 time:   [2.8419 µs 2.8489 µs 2.8557 µs]
 ```

external-crates/move/crates/move-vm-runtime/src/execution/dispatch_tables.rs
cgswords: Revised comment. Does that clarify?

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: First-mention order. I updated the comment. Does that clarify?

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: renamed

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: invariant

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: `wrap_n` added

external-crates/move/crates/move-vm-runtime/src/shared/type_size_formulae.rs
cgswords: I added a bounds

PR #27349, opened 2026-07-20
[source-verification 4] Add --verify-only bytecode comparison mode

mdgeorge4153:
## Description

Adds a hidden `--verify-only` mode to `sui client verify-source`, for tooling (the debugger) that wants to check an existing build against an on-chain package without rebuilding.

The on-chain id is the flag's value:

```
sui client verify-source --verify-only <package-id> <package-path>
```

It compares the modules already compiled under `<package-path>/build` against the on-chain package, skipping the compiler, toolchain resolution, and publication-metadata lookup. Module bytecode only, not linkage. The caller supplies the id, so this is a local-build-versus-chain diff rather than an authenticity check on recorded source. The flag is hidden; its consumer is other tooling.

Stacked on #27213.

## Test plan

Unit test for reading the compiled modules out of the build directory; the shell test gains a `--verify-only` case that compares the already-built package against its published id.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-source-verification/src/lib.rs
+++ b/crates/sui-source-verification/src/lib.rs
@@ -17,6 +17,7 @@ mod build;
 mod compare;
 mod onchain;
 mod pinning;
+mod prebuilt;
 mod toolchain_version;
 
 pub use binary::ensure_binary;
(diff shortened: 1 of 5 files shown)


awelc (approved):
LGTM

PR #27356, opened 2026-07-20
Use up to 20 retries in benchmark to reduce antithesis flakiness

mystenmark:
We are getting pretty frequent failures here when the cluster is just on the cusp of recovering

### DIFF
--- a/crates/sui-benchmark/src/lib.rs
+++ b/crates/sui-benchmark/src/lib.rs
@@ -11,7 +11,7 @@ use anyhow::bail;
 use async_trait::async_trait;
 use fullnode_reconfig_observer::FullNodeReconfigObserver;
 use futures::TryStreamExt;
-use mysten_common::{fatal, random::get_rng};
+use mysten_common::{fatal, in_antithesis, random::get_rng};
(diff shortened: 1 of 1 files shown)


crates/sui-benchmark/src/lib.rs
```diff
@@ -11,7 +11,7 @@ use anyhow::bail;
 use async_trait::async_trait;
 use fullnode_reconfig_observer::FullNodeReconfigObserver;
 use futures::TryStreamExt;
-use mysten_common::{fatal, random::get_rng};
+use mysten_common::{fatal, in_antithesis, random::get_rng};
```
alex-mysten, line 14: we don't want to limit this to just antithesis?

alex-mysten (approved):

crates/sui-benchmark/src/lib.rs
mystenmark: we're not getting this flakiness anywhere else so it seems better to leave it as is in simtest?

mystenmark: oh, wait, i see what you mean, i lost some edits

PR #27358, opened 2026-07-21
[7/8] Streamline the object-funds checker flow around a shared update computation

lxfind:
## Description

Stack of 8 (builds on [6/8] #27335). Streamlines `should_commit_object_funds_withdraws` ahead of the in-execution check ([8/8]):

- The accumulator-version lookup moves to the top, right after the effects-error check. A transaction with no assigned accumulator version (from before accumulators were enabled, or an end-of-epoch transaction) cannot have object withdraws, so the check returns early — replacing a `debug_fatal` that could only fire after the withdraw scan for the same impossible combination.
- The address-reservation split and the running-max/net update computation fold into `compute_unsettled_withdraw_updates` in the unsettled-withdrawals module, returning both the object-account peaks (what a sufficiency check must cover) and the amounts to record as unsettled. It lives there because [8/8]'s in-execution recording path shares it.

No functional change. Note the no-object-withdraws early return still keys on the running-max map being empty (not the recordable updates): under `record_net_unsettled_object_withdraws`, a transaction that withdraws and deposits back nets to zero but still has a peak the sufficiency check must cover.

## Test plan

Pure refactoring covered by the existing `object_funds_checker` unit and integration tests, which exercise sufficient/insufficient/pending outcomes and both the running-max and net recording modes.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/accumulators/object_funds_checker/mod.rs
+++ b/crates/sui-core/src/accumulators/object_funds_checker/mod.rs
@@ -1,19 +1,15 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{
-    collections::{BTreeMap, BTreeSet},
-    sync::Arc,
-};
+use std::{collections::BTreeMap, sync::Arc};
 
-use mysten_common::{assert_reachable, debug_fatal};
+use mysten_common::assert_reachable;
 use sui_types::{
     accumulator_root::AccumulatorObjId,
     base_types::SequenceNumber,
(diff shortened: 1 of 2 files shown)


lxfind:
Dropping this per discussion: not worth restructuring the deprecated checker's flow, and the shared helper it created coupled the surviving in-execution code to the deprecated path. The in-VM PR keeps its own self-contained net computation instead; the checker stays frozen as-is until it is deleted after rollout.

PR #27360, opened 2026-07-21
Fix broadcaster hang on fatal ingestion error while streaming is active

evan-wall-mysten:
## Description

`broadcaster`'s main loop used `futures::future::join` to wait for both the streaming and ingestion tasks before checking ingestion's result. Once real streaming connects it runs until the broadcaster's overall end checkpoint (not just the current loop iteration), so in the common unbounded-range case it can run for a long time. If ingestion concurrently hit a fatal condition (closed subscriber or permanent error), the code was supposed to stop the broadcaster immediately, but `join` meant it wouldn't actually stop until the long-running streaming task also finished on its own.

Awaits the ingestion result first; on either fatal outcome, `stream_guard` is left undropped in that branch so the normal scope-exit `Drop` (which aborts the underlying task) cancels streaming immediately instead of waiting for it.

## Test plan

Added `ingestion_break_stops_broadcaster_without_waiting_for_streaming`: streaming sends one checkpoint then stalls for 3s, while ingestion takes ~300ms to notice a closed subscriber. Asserts the broadcaster stops within a 2s timeout — this fails against the old `join`-based code and passes in under 0.5s with the fix.

`cargo nextest run -p sui-indexer-alt-framework --lib ingestion::broadcaster` — 20/20 pass.

---

## Release notes

- [x] Indexing Framework: Fixes a bug where the checkpoint broadcaster could keep running instead of shutting down promptly after a fatal ingestion error, if real-time checkpoint streaming was active and healthy at the same time.

### DIFF
--- a/crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
+++ b/crates/sui-indexer-alt-framework/src/ingestion/broadcaster.rs
@@ -120,12 +120,16 @@ where
                 cohort_metrics.clone(),
             );
 
-            let (streaming_result, ingestion_result) =
-                futures::future::join(stream_guard, ingest_guard).await;
-
-            // Check ingestion result, exit on any error.
-            match ingestion_result.context("Ingestion task panicked, stopping broadcaster")? {
-                Ok(()) => {}
+            // Wait for ingestion first, rather than joining both tasks together, so that a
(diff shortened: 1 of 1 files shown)


evan-wall-mysten:
Was testing `/goal` in claud code to see if it would find any bugs and it discovered this.

amnn (approved):
LGTM!

PR #27363, opened 2026-07-21
fix(sui-tool): retry transient checkpoint fetches during formal-snapshot backfill

phips28:
## Problem

`sui-tool download-formal-snapshot`'s transaction-digest backfill (`backfill_epoch_transaction_digests`) fetches **every** checkpoint of the retained epoch from the checkpoint store with **no retry**, running concurrently with the CPU-bound state accumulation. On a busy/contended host, the async runtime gets starved enough that a single fetch times out:

```
Generic HTTP error: ... error sending request ... operation timed out
```

`.try_for_each(...).await?` then propagates that error and **aborts the entire restore** — discarding hours of already-downloaded and accumulated live-object state. Re-running restarts the backfill from scratch (its progress isn't persisted), so on a contended box the restore can fail repeatedly and never complete.

Observed on a real mainnet restore: repeated aborts at *different* checkpoints, always during the ~18-minute accumulation phase (all cores pinned). `--max-retries` had no effect because it was only wired into the object-snapshot reader, never into the backfill.

## Fix

- **Retry each backfill checkpoint fetch** with linear backoff (capped at 30s) so a transient failure / CPU-starvation window is waited out instead of aborting the whole multi-hour restore.
- **Thread the existing `--max-retries`** into `backfill_epoch_transaction_digests` (it previously ignored the flag) to bound the retries — no new flag introduced.
- **Raise the default `--max-retries` from 3 → 50**, so the out-of-the-box default survives a long accumulation phase; operators can still override.
- Warn-logs each retry so contention is visible in the logs.

## Testing

Ran the patched `sui-tool` against a mainnet formal-snapshot restore on a CPU-constrained host where the stock binary aborted 5×: the backfill now stalls (silently retrying) during accumulation and completes during the object-download phase, exiting 0. The node starts and syncs normally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17350,6 +17350,7 @@ dependencies = [
  "anemo",
  "anemo-cli",
  "anyhow",
+ "backoff",
  "base64 0.21.7",
  "bcs",
  "bin-version",
--- a/crates/sui-tool/Cargo.toml
+++ b/crates/sui-tool/Cargo.toml
@@ -11,6 +11,7 @@ workspace = true
 
 [dependencies]
 anyhow.workspace = true
+backoff.workspace = true
 num_cpus.workspace = true
 bcs.workspace = true
 clap = { version = "4.1.4", features = ["derive"] }
--- a/crates/sui-tool/src/commands.rs
+++ b/crates/sui-tool/src/commands.rs
@@ -273,8 +273,8 @@ pub enum ToolCommand {
         verbose: bool,
 
         /// Number of retries for failed HTTP requests when downloading snapshot files.
-        /// Defaults to 3 retries. Set to 0 to disable retries.
-        #[clap(long = "max-retries", default_value = "3")]
+        /// Set to 0 to disable retries.
+        #[clap(long = "max-retries", default_value = "10")]
         max_retries: usize,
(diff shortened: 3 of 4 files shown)


amnn (commented):
Thanks for the contribution @phips28, PTAL at the comments on #27362.

phips28:
@amnn thanks for the review on #27362 — addressed both points here (this PR supersedes that one):

- **Exponential backoff via the `backoff` crate**: the hand-rolled linear sleep loop is gone. The backfill fetch now uses `backoff::future::retry_notify` with `ExponentialBackoff { max_interval: 30s, max_elapsed_time: None, ..Default::default() }`, so we get the crate's default jitter (randomization factor) and multiplier — no more synchronized retry storms. Attempts are still bounded by `--max-retries` (the error is marked `permanent` once the cap is hit), and the `notify` hook warn-logs each retry with the backoff delay.
- **Default `--max-retries`**: dropped from 50 to 10, and removed the mention of the default from the doc comment since clap already renders it in `--help`.

`cargo check`/`cargo xclippy -p sui-tool` are clean.

amnn (approved):
Looks great, thank you!

PR #27366, opened 2026-07-21
[gql] Migrate `Events` to grpc ledger service `list_events` 1/3

wlmyng:
## Description 

Part 1 of 3 extends graphql so that, if the "alpha" ledger grpc reader is configured, graphql will instead read from ledger service instead of Postgres.

There are additional changes we can make. Given the events connection eagerly loads all event contents, we can widen the read mask on the grpc request, `read_mask = contents, package_id, module, sender, transaction_digest, event_index[, checkpoint]` to cover almost what `Event` wants.

The challenge is that `timestamp_ms` is not available on the grpc response today. A few approaches:
1. Extend grpc `list_events` to also return `timestamp_ms` somewhere
2. eagerly look up each event's checkpoint's timestamp
3. Change the `Event` node shape to hold `checkpoint_sequence_number`, lazily load `timestamp_ms`. This is a bigger lift as the subscription/PG/effects paths all construct Event too.

The second PR would involve changes inspired from the list above.

Finally, the third PR would remove Postgres support.

## Test plan 

e2e tests in graphql_bitmap_pagination_tests.rs, tests in `event/mod.rs`

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
@@ -7,6 +7,9 @@ use anyhow::Context as _;
 use async_graphql::Context;
 use async_graphql::Object;
 use async_graphql::connection::Connection;
+use async_graphql::connection::Edge;
+use async_graphql::connection::EmptyFields;
+use async_graphql::connection::PageInfo;
 use diesel::prelude::QueryableByName;
 use diesel::sql_types::BigInt;
 use itertools::Itertools;
@@ -86,6 +89,12 @@ pub(crate) struct Event {
     pub(crate) timestamp_ms: Option<u64>,
 }
 
+/// Custom `Connection` for events to support partially-filled pages.
+pub(crate) struct EventConnection {
+    pub edges: Vec<Edge<String, Event, EmptyFields>>,
+    pub page_info: PageInfo,
+}
+
 #[Object]
 impl Event {
     /// The Move value emitted for this event.
@@ -148,6 +157,24 @@ impl Event {
     }
 }
 
+#[Object]
+impl EventConnection {
+    /// Information to aid in pagination.
+    async fn page_info(&self) -> &PageInfo {
+        &self.page_info
+    }
+
+    /// A list of edges.
+    async fn edges(&self) -> &[Edge<String, Event, EmptyFields>] {
+        &self.edges
+    }
+
+    /// A list of nodes.
+    async fn nodes(&self) -> Vec<&Event> {
+        self.edges.iter().map(|e| &e.node).collect()
+    }
+}
+
 impl Event {
     /// Paginates events based on the provided filters and page parameters.
     ///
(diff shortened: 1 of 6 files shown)


crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -213,6 +261,82 @@ impl Event {
...
+        options.after = after;
+        options.before = before;
+        options.ordering = Some(if page.is_from_front() {
+            v2::Ordering::Ascending as i32
+        } else {
+            v2::Ordering::Descending as i32
+        });
+
+        let mut request = v2::ListEventsRequest::default();
+        // Position only — event contents and timestamps hydrate in a batch via `KvLoader`.
```
amnn, line 322: I'm not understanding the rationale for why we don't fetch the event content at this point? It would make sense if we did it lazily based on (fields resolved) later on, but we're not doing that -- `load_event_contents` is fetching that stuff anyway.

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -235,6 +359,34 @@ impl CEvent {
...
+                checkpoint: 0,
+                tx_seq: c.tx_sequence_number,
+                event_index: c.ev_sequence_number,
+            },
+        }
+    }
+}
+
+impl EventConnection {
+    fn empty() -> Self {
```
amnn, line 379: This looks like a derivable `Default`.

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -301,12 +453,161 @@ impl TxBoundsCursor for CEvent {
...
+
+/// The position a `ListEvents` stream item points at: the emitting transaction's digest and the
+/// event's index in that transaction's events list.
+fn item_position(payload: &v2::Event) -> Result<(TransactionDigest, u32), RpcError> {
+    let digest = payload
+        .transaction_digest
+        .as_deref()
+        .context("ListEvents item missing transaction digest")?
+        .parse::<TransactionDigest>()
+        .context("Failed to parse transaction digest from ListEvents")?;
```
amnn, line 483:
```suggestion
    let digest: TransactionDigest = payload
        .transaction_digest
        .as_deref()
        .context("ListEvents item missing transaction digest")?
        .parse()
        .context("Failed to parse transaction digest from ListEvents")?;
```

amnn (commented):
This looks mostly good, but I'm not following the decision to only fetch the keys/positions from the stream but then immediately hydrate them with a follow-up batch load.

amnn (approved):
I see that my questions/concerns are addressed in the next PR in the stack, so all good! Just nits left.

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
wlmyng: I could impl `Default`, but `PageInfo` itself doesn't derive Default, only SimpleObject

amnn: Ah okay, then less interesting/appealing I suppose

PR #27367, opened 2026-07-21
feat(sui-cluster-test): migrate to gRPC-native APIs (LedgerService/StateService/TransactionExecutionService)

ebmifa:
## Problem: gateway migration broke the client contract `sui-cluster-test` assumed

`sui-cluster-test` built a **full JSON-RPC** `SuiClient` against the public fullnode root (`wallet_client.rs` → `SuiClientBuilder::default().build(url).await.unwrap()`) and its test cases asserted the **legacy full JSON-RPC contract** (`suix_getAllCoins`, `unsafe_*` builders, `quorum_driver_api`, `governance_api`, `read_api`).

After the **v1.76.0 gateway migration**, the public fullnode root **serves gRPC only** — a JSON-RPC POST to root returns **HTTP 404**, so the client panicked on setup before any test ran. JSON-RPC moved to the reduced indexer-alt `/json-rpc` subset, which does **not** include every method the suite needs. The suite therefore failed during **client/setup** rather than detecting a real deployment regression.

The durable fix is to make `sui-cluster-test` **gRPC-native**. This PR does that. It does **not** restore the retired JSON-RPC endpoint and does **not** add a JSON-RPC fallback.

## Services used (authoritative API surface)

- **LedgerService** — object/ledger reads (`get_object`), transaction retrieval + checkpoint wait (`get_transaction`, `wait_for_transaction`), epoch/system-state/protocol/RGP reads, chain identifier.
- **TransactionExecutionService** — execution (`execute_transaction`, `execute_transaction_and_wait_for_checkpoint`).
- **StateService** — owned-object enumeration (`list_owned_objects` / `get_owned_objects`), `get_balance`, `list_balances`. **StateService is served on public fullnodes** (devnet/testnet return populated `ListOwnedObjects`/`GetBalance`/`ListBalances`; staging/CI return valid-empty for a zero-address query) and on the embedded new-local node.

Transactions are built with the **high-level gRPC transaction builder using an explicit gas coin** sourced from the faucet response (`CoinInfo { id, amount, transfer_tx_digest }`). This is a deliberate determinism choice — the faucet already returns exact object IDs, so construction never has to guess a gas coin — **not** a workaround for any missing service.

> Note on `LiveDataService`: it does not exist in current source; the deployed coin/balance service is **StateService** (`state_client()`). This PR introduces no dependency on `LiveDataService`.

## Type-key gotcha (documented in-code)

Coin APIs key on two different types, which caused two bugs during development (both fixed):
- **`GetBalance` / `ListBalances`** key on the coin's **inner** type `T` (e.g. `0x2::sui::SUI`, `pkg::managed::MANAGED`).
- **`ListOwnedObjects`** filters on the **object** type `Coin<T>` (e.g. `Coin<0x2::sui::SUI>`, `Coin<MANAGED>`), and its **parameterless** `0x2::coin::Coin` filter matches every `Coin<T>`.

## Test-by-test coverage mapping (all 8 preserved)

| Test | Construction (new) | Verification | Notes |
| --- | --- | --- | --- |
| **NativeTransfer** | high-level `transfer_object` (explicit gas) | native effects + `get_object` owner; native `balance_changes` exact `TypeTag` compare | **Bug fixed:** the second transfer now asserts the *second* object id (was re-checking the first). Refs refreshed between txs. |
| **CoinMergeSplit** | high-level `split_coin` / `merge_coins` (explicit gas, refs refreshed per tx) | created coins from effects → `get_object` coin value; merged coins verified **deleted via effects**; final primary value | |
| **SharedCounter** | existing gRPC-native builders (`publish_basics_package_and_make_counter`, `increment_counter`) | native effects (`input_consensus_objects`, `mutated`) + `get_object` shared owner/version | setup-panic fix only |
| **FullNodeExecuteTransaction** | direct `execute_transaction` (immediate) then `wait_for_transaction` + `get_transaction`; separately `execute_transaction_and_wait_for_checkpoint` | immediate success effects; waited result asserts `checkpoint.is_some()` | Renamed away from Quorum-Driver terminology. **Intentionally bypasses** the shared execute helper to exercise both execution paths. |
| **GrpcPublishTransaction** (was FullNodeBuildPublishTransaction) | high-level gRPC `publish(modules, deps, explicit gas)` | effects `created` immutable package | Renamed/rewritten — the old name claimed JSON-RPC server-side builder behavior. |
| **CoinIndex** | high-level gRPC `move_call` (SuiJsonValue args, explicit gas) for mint/wrap/unwrap/burn/mint_multi | **StateService `GetBalance` + `ListBalances` (required, at every stable MANAGED point: 10000 / 10010 / 0)** for SUI and the MANAGED custom coin; `ListOwnedObjects` filtered by `Coin<T>` and parameterless `0x2::coin::Coin`; coin values decoded from BCS and cross-checked against the aggregates; opaque page tokens | **Kept, not deleted.** See "behavior changed/retired". |
| **RandomBeacon** | existing gRPC-native builders | `get_protocol_config` feature flag; event count + `RandomU128Event` type | setup-panic fix only |
| **Staking** | low-level `TestTransactionBuilder::call_staking` (separate stake + gas coins) | validator from `get_system_state_summary`; StakedSui from effects, decoded natively (owner, principal, staking pool, **activation epoch = execution-checkpoint epoch + 1**); **plus balance/coin-count accounting** — snapshots `GetBalance(SUI)` and owned `Coin<SUI>` count before/after and asserts the stake consumes exactly one Coin<SUI> and the indexed balance changes by the effects-reported amount. The owned-object count uses `try_collect`, so a pagination/decoding error fails the test instead of being counted as a coin | Replaces `governance_api().get_stakes` + `quorum_driver_api`. |

## Shared infrastructure changes (`lib.rs`, `helper.rs`, `wallet_client.rs`, deps)

Beyond the per-test ports, the crate's shared scaffolding changed:

**`wallet_client.rs`** — dropped the separately-built JSON-RPC `SuiClient` field entirely. `get_fullnode_client() -> &SuiClient` is replaced by `grpc_client() -> GrpcClient`, returning a cheap clone of the connection `WalletContext` already caches. This is the change that removes the setup panic: nothing builds a JSON-RPC client against the fullnode root anymore.

**`lib.rs` (`TestContext`)** —
- `get_fullnode_client` / `clone_fullnode_client` / `get_fullnode_rpc_url` → a single `get_grpc_client()`.
- `sign_and_execute` simplified to `(txn_data, desc) -> ExecutedTransaction` (native effects/balance-changes), replacing the JSON-RPC response-options plumbing.
- **`build_transaction_remotely` removed** — this wrapped the `unsafe_*` server-side JSON-RPC transaction builders, which have no gRPC equivalent by design. All construction is now client-side (see above).
- `let_fullnode_sync(digests, timeout_sec)` → `wait_for_txns(&digests)`, backed by `LedgerService::wait_for_transaction` instead of a retry/sleep loop; the ad-hoc `get_tx_with_retry_times` helper is gone.
- **New `current_object_ref(id)`** — refetches an object's current `ObjectRef` so multi-transaction tests refresh refs between txs rather than reusing stale versions.

**`helper.rs`** — `ObjectChecker` was ported from JSON-RPC `SuiObjectData` to native `Object` (owner via native `Owner`, coins via `GasCoin::try_from(&Object)`). Its **`deleted()` mode was removed**: gRPC `GetObject` on a deleted object returns a not-found error rather than a tombstone with `SuiObjectDataOptions`, so deletion is now asserted from transaction **effects** (`deleted()`), which is the more direct signal anyway — `CoinMergeSplit` uses it. `BalanceChangeChecker` was ported to `sui_sdk_types::BalanceChange` with exact `TypeTag` comparison (no substring matching), and its `owner(Owner)` builder became `address(SuiAddress)` to match the native type.

**Dependencies** — removed `jsonrpsee` and `sui-json-rpc-types`; added `sui-rpc-api` (gRPC client) and `sui-sdk-types` (native effects/balance-change types). `Cargo.lock` updated accordingly.

**File rename** — `fullnode_build_publish_transaction_test.rs` → `grpc_publish_transaction_test.rs` (module list in `test_case.rs` updated).

## Behavior intentionally changed or retired (with justification)

In `coin_index_test`, which this PR **keeps and ports** to StateService:

1. ~~Recipient of a SUI transfer owns 0 coin objects~~ **Corrected after review (stefan-mysten) — no behavior change here after all.** Classic `TransferSui` creates a distinct recipient `Coin<SUI>` object; address-balance accumulator crediting applies only to `send_funds`-style transfers. Verified on live devnet (the round-2 recipient owns exactly one `Coin<SUI>`; `fundsInAddressBalance: 0`). The legacy assertion `recipient coin_object_count == 1` is **restored** alongside the `StateService::GetBalance` check (`a066c31b3b`). The earlier 0-objects observation was a wrong-filter artifact from an intermediate draft.
2. **Pagination assertions adapted to opaque page tokens.** The JSON-RPC object-ID-cursor assertions were rewritten against StateService's opaque `next_page_token` while preserving completeness (all-coins == SUI ∪ MANAGED) and exclusion (wrapped/removed coins absent) checks.

No behavioral area was silently dropped, and MANAGED balance-index coverage is **required, not optional** (see below).

## Custom-coin balance index: investigated, fixed, and required (not gated)

An earlier revision made `ListBalances` inclusion of the freshly-published MANAGED type *optional*, on the suspicion that the embedded node might not index custom-coin balances. **That suspicion was wrong and is now fixed.** A live diagnostic (poll loop) showed:
- `GetBalance` / `ListBalances` **correctly index the MANAGED coin at 10000, at t=0 (no lag)** — once queried with the **inner** type `pkg::managed::MANAGED`.
- The earlier "0 / not reported" reading was a **test bug**: the MANAGED `Coin<T>` object type was passed where the balance APIs expect the inner `T`.

CoinIndex now **requires** `GetBalance(MANAGED)` and `ListBalances(MANAGED)` to agree at **every** stable point: after the first mint (10000), after the second mint (10010), and after the final burn (0 / absent from the listing). There is no expected-failure gate, because there is no underlying defect.

## Test plan

**Performed (locally, this environment)** — commands, copy-pasteable as-is:

```bash
cargo fmt --all -- --check
cargo build -p sui-cluster-test
cargo clippy -p sui-cluster-test --all-targets -- -D warnings
cargo run -p sui-cluster-test -- new-local
```

Results, in the same order:

| Command | Result |
| --- | --- |
| `cargo fmt --all -- --check` | pass |
| `cargo build -p sui-cluster-test` | pass, no warnings |
| `cargo clippy ... -D warnings` | pass |
| `cargo run -p sui-cluster-test -- new-local` | **8 of 8 tests passed**, re-run at current head (also 8/8 at `c8ddf25`, `a066c31b3b`) |

The 8 tests are NativeTransfer, CoinMergeSplit, SharedCounter, FullNodeExecuteTransaction, GrpcPublishTransaction, CoinIndex, RandomBeacon, and Staking. The embedded fullnode serves LedgerService, TransactionExecutionService, **and StateService** on one port, so the full CoinIndex StateService path (owned-object enumeration + GetBalance + ListBalances) is exercised locally.

Additional checks:
- **Repo CI runs this suite too**, via the `cluster_test` integration test in `crates/sui-cluster-test/tests/local_cluster_test.rs`, which calls the same `ClusterTest::run(ClusterTestOpt::new_local())` path. The `test` job's `cargo nextest run --profile ci -E '!package(sui-bridge) and !package(sui-bridge-indexer)'` includes `sui-cluster-test`, so a regression in any of the 8 ported tests fails CI, not just build/clippy.
- Legacy-path audit: no remaining `SuiClient`/`SuiClientBuilder`/`get_fullnode_client`/`jsonrpsee`/`unsafe_*`/`quorum_driver_api`/`coin_read_api`/`governance_api`/`read_api`/`LiveDataService` in the crate. `SuiJsonValue` remains for **local** Move-call argument conversion (allowed); `SuiClientConfig`/`SuiEnv` remain only as `WalletContext` config types (not a network client).

**Live devnet smoke — PERFORMED (two rounds, from mp3-built `sui-tools` images, same invocation as the sui-operations CI gate: `docker run -v <sui>:/sui <image> sui-cluster-test devnet`):**

- **Round 1** (image at `c8ddf2560a`): NativeTransfer, CoinMergeSplit, SharedCounter passed against live devnet; FullNodeExecuteTransaction panicked at `lib.rs:82` — a **faucet-contract mismatch**, not a transport defect: `get_sui_from_faucet` assumed one request yields >= `minimum_coins`, but devnet's gas-station-backed faucet (sui-operations#8434) dispenses **one coin per request**. Pre-existing assumption, masked before this PR because the legacy binary always panicked at setup (the 404) before reaching the faucet.
- **Fix**: `a87f2309a7` — accumulate faucet requests (bounded by `minimum_coins`, early-break once met); no behavior change for multi-coin faucets.
- **Round 2** (image at `a87f2309a7`): **8 of 8 tests passed** against live devnet (exit 0, ~48s), including the full **CoinIndex StateService path on the public gateway** (GetBalance / ListBalances / ListOwnedObjects with `Coin<T>` + parameterless filters, managed-coin publish→mint→wrap→unwrap→burn lifecycle) and **Staking** (validator from system-state summary; StakedSui principal + pool verified; activation epoch = current+1 as asserted by that binary; the head derives the epoch from the execution checkpoint instead).
- The devnet faucet currently accepts unauthenticated requests, so no `FAUCET_AUTH_HEADER` was needed.

**Attempted, blocked by faucet auth:** live **testnet** smoke (`sui-cluster-test testnet` at `47b9c26c`) reached the faucet and exercised the bounded 429 retry exactly as designed (five honored waits), but the testnet faucet hard-throttles unauthenticated requests — an authenticated run (`FAUCET_AUTH_HEADER`, as CI's cluster-tests action passes) is required for a full 8/8. Not a migration failure; no code change indicated. **CI-Nomad** may only be resolvable from its internal environment.

## Rollout dependency

The remote devnet smoke has now been **performed and passed (8/8)** from a purpose-built image of this branch. The remaining gate before restoring the sui-operations devnet cluster-test gate is simply: **this PR lands on main and a standard `mysten/sui-tools:devnet` image containing the migrated binary ships** via the normal pipeline. This PR does **not** re-enable the gate itself. (StateService is served publicly; not a dependency.)

## References

- Design proposal: MystenLabs/sui-operations#8458 (documentation-only; **OPEN**). #8458 explicitly says **"Do not delete `coin_index_test`"**, which this PR honors — CoinIndex is kept and ported to StateService.
- Immediate pipeline repair that removed the broken devnet gate: MystenLabs/sui-operations#8446 — **CLOSED unmerged in favor of this PR**; the legacy devnet gate therefore remains in sui-operations `main` and keeps failing until this PR ships in a devnet image.

**Recommendation on #8458:** **amend and retain** as the durable design doc (do not close as superseded). The devnet-smoke condition is now satisfied (8/8 twice, at `a87f2309` and `47b9c26c`); the remaining condition is simply this PR landing. Worth folding in: (a) **StateService is served both publicly and on the embedded new-local node**, so CoinIndex is fully portable and locally validatable; (b) the `sui-test-transaction-builder` helpers are **already gRPC**; and (c) the inner-`T` vs `Coin<T>` type-key distinction for balance vs owned-object APIs.

> A "delete `coin_index_test`" idea appeared only in the original **local draft plan** during scoping; it was **not** part of #8458 and was not adopted — the test is preserved.

---

## Release notes

None required — this PR touches only `sui-cluster-test`, an internal test binary/suite. No protocol, node, gRPC/JSON-RPC/GraphQL server, CLI, SDK, or indexing behavior changes; the gRPC services consumed here are pre-existing and unmodified.

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-cluster-test/src/helper.rs
```diff
@@ -61,112 +58,54 @@ impl ObjectChecker {
...
-                SuiObjectDataOptions::new()
-                    .with_type()
-                    .with_owner()
-                    .with_bcs(),
-            )
-            .await
-            .or_else(|err| bail!("Failed to get object info (id: {}), err: {err}", object_id))?;
-
-        trace!("getting object {object_id}, info :: {object_info:?}");
+        let mut client = client.clone();
```
stefan-mysten, line 69: I wonder if this really needs to be mut.

ebmifa:
**Live devnet smoke — round 1 results + why commit a87f2309a7 was added**

Ran the suite from a sui-tools image built at `c8ddf2560a` against **live devnet** (`sui-cluster-test devnet`, same invocation as the sui-operations CI gate):

| Test | Result |
| --- | --- |
| NativeTransfer | ✅ (native gRPC execution + LedgerService reads) |
| CoinMergeSplit | ✅ |
| SharedCounter (publish + shared object) | ✅ |
| FullNodeExecuteTransaction | ❌ `lib.rs:82: Expect to get at least 2 Sui Coins … but only got 1` |
| GrpcPublish / CoinIndex / RandomBeacon / Staking | not reached |

The transport migration held: no `wallet_client.rs` 404 panic, gRPC client + faucet + real transactions all worked against the public gateway.

The failure is a **faucet-contract mismatch, not a transport defect**: `get_sui_from_faucet` assumed one request yields ≥ `minimum_coins`. The local test faucet sends several coins per request (hence 8/8 on new-local), but devnet's gas-station-backed faucet (sui-operations#8434) dispenses **one coin per request**, so any 2-coin minimum panics. This predates the migration and was masked because the legacy binary always panicked at setup before reaching the faucet.

**Commit a87f2309a7** fixes it harness-side: loop faucet requests (bounded by `minimum_coins`, early-break once met). No behavior change for multi-coin faucets.

Round-2 smoke from an `a87f2309a7` image is queued to exercise the remaining 5 tests (incl. CoinIndex/StateService and Staking on public devnet); results + PR body update to follow.

*Edited: corrected the NativeTransfer row — it verifies execution-provided balance changes and LedgerService object reads, not StateService balance APIs. The first full StateService balance/enumeration exercise is CoinIndex (round 2).*

bmwill (approved):

ebmifa:
**Live devnet smoke — round 2: 8/8 PASSED** ✅

Image built from `a87f2309a7` (mp3 build 43697), run with the CI-gate invocation (`docker run -v <sui>:/sui <image> sui-cluster-test devnet`) against live devnet. Exit 0, ~48s:

| Test | Result |
| --- | --- |
| NativeTransfer | ✅ |
| CoinMergeSplit | ✅ |
| SharedCounter | ✅ |
| FullNodeExecuteTransaction | ✅ (both execution paths — round-1 blocker cleared by the faucet fix) |
| GrpcPublishTransaction | ✅ |
| **CoinIndex** | ✅ — **first full StateService exercise on the public gateway**: GetBalance / ListBalances / ListOwnedObjects (`Coin<T>` + parameterless) + managed-coin publish→mint→…→burn lifecycle |
| RandomBeacon | ✅ |
| Staking | ✅ — StakedSui principal/pool verified, activation epoch = current+1 |

PR body updated accordingly (validation section now records round-1 partials, the faucet-contract fix, and round-2 results; #8446 reference corrected to CLOSED-unmerged). Remaining rollout gate: land this PR + ship a standard `sui-tools:devnet` image, then re-enable the sui-operations gate.


*Edited: removed the GetCoinInfo mention — the ported CoinIndex exercises GetBalance / ListBalances / ListOwnedObjects; the crate makes no GetCoinInfo calls.*

crates/sui-cluster-test/src/test_case/coin_index_test.rs
```diff
@@ -3,677 +3,678 @@
...
+use sui_types::gas_coin::{GAS, GasCoin};
 use sui_types::object::Owner;
+use sui_types::{Identifier, SUI_FRAMEWORK_ADDRESS};
 use tracing::info;
 
 pub struct CoinIndexTest;
 
+/// A decoded coin object owned by an address: its ID and current value.
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct OwnedCoin {
```
stefan-mysten, line 24: I don't understand what value means here. Is it version, or balance?

crates/sui-cluster-test/src/test_case/coin_index_test.rs
```diff
@@ -3,677 +3,678 @@
...
-        old_total_balance = total_balance;
+        // `old_total_balance` is refreshed after the publish below (its value
+        // here would be overwritten before use).
+
+        // The recipient balance change comes from the (already-checkpointed)
+        // effects and is authoritative.
+        assert!(recipient_change.1 > 0);
+        // NOTE (behavior change): under the current protocol, transferring SUI to
+        // an address credits the recipient's *address balance* accumulator rather
+        // than creating a distinct `Coin<SUI>` object, so the old JSON-RPC
```
stefan-mysten: I think this is incorrect... This is only if using `send_funds` function. The classic `transfer` function is not doing this.

crates/sui-cluster-test/src/helper.rs
ebmifa: It does — `GrpcClient::get_object` takes `&mut self` (`crates/sui-rpc-api/src/client/mod.rs:120`, tonic-style client), and we only hold `&GrpcClient` here, so it's clone-then-call on the owned clone (cheap channel clone).

crates/sui-cluster-test/src/test_case/coin_index_test.rs
ebmifa: Good call — it's the coin's balance (the u64 from `Coin::value`), not the version. Renamed the field to `balance` and documented it as MIST in `a066c31b3b`.

crates/sui-cluster-test/src/test_case/coin_index_test.rs
ebmifa: You're right, and I verified it both ways before fixing:

- **On-chain (live devnet)**: the round-2 smoke's recipient (`0x3294c2c0…`) queried post-hoc — `suix_getOwnedObjects` returns exactly **1 × `Coin<0x2::sui::SUI>`**, and `suix_getBalance` shows `coinObjectCount: 1`, `fundsInAddressBalance: '0'`. Classic `transfer_sui` created a distinct coin object; no accumulator crediting.
- **new-local**: full suite passes **8/8** with the legacy assertion restored.

Fixed in `a066c31b3b`: the incorrect note is replaced (now narrowly distinguishing `transfer_sui` from `send_funds`-style transfers), and the legacy `recipient owns exactly one Coin<SUI>` assertion is restored alongside the balance check. The earlier 'owns 0 objects' observation was a wrong-filter artifact (inner `0x2::sui::SUI` vs object type `Coin<SUI>`) from an intermediate draft.

ebmifa:
**Round-3 smoke + multi-environment matrix**

Round 3 ran the `47b9c26c` image (mp3 build 43699) against testnet and devnet:

- **devnet: 8/8 passed** (48s) — re-validates the already-green env after both faucet commits; no regression.
- **testnet: blocked by faucet policy, not by the migration.** The new 429 handling worked exactly as designed — five bounded retries honoring every advised wait (`3s, 0s, 5s, 0s, 5s`) — but the testnet faucet hard-throttles unauthenticated requests from this runner's IP, so attempts exhausted cleanly (`faucet.rs` exhaustion path; the observed final-attempt sleep is the quirk already fixed in `6917531664`). A testnet run needs `FAUCET_AUTH_HEADER` (the same secret CI's cluster-tests action passes); no code change indicated.

**Per-commit evidence matrix:**

| Environment | Result | Binary / commit |
| --- | --- | --- |
| new-local | 8/8 | working tree ≡ `a066c31b3b` (also 8/8 at `17840f4a`) |
| devnet | 8/8 (round 2) + 8/8 (round 3) | `a87f2309`, `47b9c26c` |
| ci | 8/8 (54s) | `a87f2309` |
| staging | 8/8 (56s) | `a87f2309` |
| testnet | faucet auth-gated; 429 retry verified | `47b9c26c` |
| GitHub head CI | pending at time of writing | `a066c31b3b` |

ci/staging excerpt (tri-env run, 2026-07-21 20:07–20:09Z, identical invocation):
```
[ci]      20:08:30Z Test Staking succeeded.  → 8 of 8 tests passed.  exit 0
[staging] 20:09:27Z Test Staking succeeded.  → 8 of 8 tests passed.  exit 0
```
Notable: ci and staging still serve the legacy JSON-RPC contract at root — these passes demonstrate the gRPC-native suite is safe on pre-gateway-migration environments, not only migrated ones.

Remote smoke commits (`a87f2309`/`47b9c26c`) vs head (`a066c31b3b`): the head-only delta is the final-429-sleep guard, Retry-After preference, the restored recipient `Coin<SUI>` assertion (validated on-chain — the round-2 recipient owns exactly one `Coin<SUI>`, `fundsInAddressBalance: 0` — and via new-local 8/8), and the `OwnedCoin.balance` rename. That delta is covered by head CI (the `cluster_test` integration test runs the new-local suite).

ebmifa:
**Final review round addressed — `262081b51b`**

1. **[P1] Swallowed `ObjectChecker` result (NativeTransfer)** — fixed: the `check()` result is now `.expect(...)`-ed, so a failed LedgerService `get_object` fails the test instead of vacuously passing the ownership assertion.
2. **[P2] Epoch-boundary race (Staking)** — fixed: the activation-epoch gate now derives the execution epoch from the transaction's own checkpoint (`response.checkpoint` → `get_checkpoint_summary(seq).data().epoch`) and asserts `activation == execution_epoch + 1`; the racy pre-execution system-state read is gone.
3. **[P2] PR-body contradictions** — corrected: the behavior-change section now records the *corrected* transfer semantics (classic `TransferSui` → one recipient coin object; assertion restored, credit to @stefan-mysten); new-local validation references the current head; testnet reads **attempted, blocked by faucet auth** (429 retry verified); the #8458 recommendation notes its devnet-smoke condition is satisfied, leaving only the landing of this PR.

**new-local: 8/8 with both code fixes.** Remaining gate: head CI at `262081b51b`.

ebmifa:
**Round-4 smoke — head-exact validation on ci, staging, devnet**

Image built from the current head `262081b51b` (mp3 build 43714), CI-gate invocation:

| Environment | Result | Binary |
| --- | --- | --- |
| ci | **8/8** (21:17Z) | `262081b51b` |
| staging | **8/8** (21:18Z) | `262081b51b` |
| devnet | **8/8** (21:19Z) | `262081b51b` |
| new-local | **8/8** | `262081b51b` |

Every reachable environment now has evidence at the **exact head**, eliminating the earlier smoke-commit/head distinction. Testnet remains faucet-auth-gated (429 retry verified at `47b9c26c`); mainnet is not a supported cluster-test environment. Remaining gate: head CI.

PR #27368, opened 2026-07-21
Decouple old-execution-version tests from the built-in framework

lxfind:
## Description

A handful of tests run the validators' built-in (current) framework bytecode under an outdated executor, so any native the current framework declares must stay registered in old natives tables just to keep them linking. This change decouples them so new framework natives can be registered in the latest execution version only.

- `test_create_coin_registry_object` and `test_create_display_registry_object` are deleted. They pin pre-119 protocol versions (95 and 115) to exercise the one-time creation of the coin and display registries at their upgrade epoch; the mismatch is inherent because the epoch-change transaction publishes the current framework under the outgoing epoch's executor. Those migrations completed on every network long ago, so the tests no longer guard anything that can regress.
- `test_upgrade_package_add_new_module_in_dep_only_mode_pre_v68` and `test_address_balance_gas_v3_accumulator_sign` still guard frozen v3 engine behavior, so instead of pinning the execution version (which keeps the protocol version at MAX and therefore boots genesis with the built-in framework), they now start at protocol version 118 — the last execution-v3 version. Genesis then loads the matching framework from the bytecode snapshot (`load_bytecode_snapshot(118)`), which pairs correctly with the v3 executor no matter what natives are later added to the latest one. This drops the previously-needed gas-model pin too, since the real 118 config is self-consistent; the only overrides left are the two consensus-handler flags introduced after 118 (`merge_randomness_into_checkpoint`, `timestamp_based_epoch_close`) and the pre-v68 dep-only flag under test.
- With no test pinning old executors anymore, the invariant is now enforced: `execution_version` is marked with a new `#[custom_setter]` derive attribute that suppresses the generated `set_execution_version_for_testing`, and a hand-written one asserts the new version is >= the config's current one, with an error message pointing at the protocol-version-pinning pattern. The `from_str`/`set_attr_for_testing` string paths route through the hand-written setter by name, so all external mutation routes are covered. Replay's executor override — the only remaining caller — only upgrades and is unaffected.

## Test plan

The two reworked tests were validated both here and on a branch where the current framework declares a new native registered only in the latest execution version: there, the old execution-version pin fails genesis with `MISSING_DEPENDENCY` on `0x2`, while the protocol-118 form passes — confirming genesis links the snapshot framework against the v3 executor. Registry-object existence and behavior, and dep-only upgrade rules under the current protocol, remain covered by the current-version test suites.

New unit tests cover the setter guard (upgrade and same-version allowed, downgrade panics). Protocol config snapshots are unchanged — the serialized config is identical for every version and chain.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


tzakian (approved):
Makes sense/seems like a nice cleanup to me.

PR #27372, opened 2026-07-21
[source-verification 3] Output the verification metadata, and expose it via --json

mdgeorge4153:
## Description

`verify-source` now reports the metadata a run relied on — the package's original id and published-at address, the toolchain version it rebuilt with, and the path to the `sui` binary it used — so callers (in particular enclave tooling) can capture what was verified against. It prints under the success message and, through the existing global `--json` flag, is emitted as parseable JSON. No new flag.

Stacked on #27365.

## Test plan

A unit test asserts the JSON uses stable camelCase keys and renders addresses as `0x` strings. The local-publish shell test gains a deterministic check that the `--json` output carries `toolchainVersion` (the metadata's addresses and binary path are machine-specific, so they're not snapshotted).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: `sui client verify-source` now prints the metadata it verified against (original id, published-at, toolchain version, binary path) on success, and emits it as JSON under the global `--json` flag.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-source-verification/src/lib.rs
+++ b/crates/sui-source-verification/src/lib.rs
@@ -1,10 +1,11 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use colored::Colorize;
 use move_package_alt::schema::{Environment, Publication};
+use serde::Serialize;
 use sui_package_alt::SuiFlavor;
 use sui_rpc_api::Client;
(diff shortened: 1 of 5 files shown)


mdgeorge4153:
[CLAUDE] Auto-marked merged by GitHub during the stack reorder (metadata's commit ended up in this PR's old base branch). The change was not actually merged to main; it continues as #27436, rebased into the linear stack.

PR #27373, opened 2026-07-21
[hardening][sui-framework] Harden owner tracking for derived objects and dynamic fields

tnowacki:
## Description 

- Track the ownership relationship of newly created IDs for derived objects and dynamic fields in the object runtime.
- Most importantly this is for tracking the object's root owner version, as is necessary for MVCC reads of dynamic fields. 
- This is not a functional/observable change today (outside of a gas cost reduction):
  - For derived objects they cannot be recreated (yet)
  - For dynamic fields, the UID is not accessible and it does not have its own dynamic fields in this way

## Test plan 

- Ran tests; not yet an observable change (outside of gas)

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-protocol-config/src/lib.rs
```diff
@@ -4560,6 +4568,7 @@ impl ProtocolConfig {
                     if chain != Chain::Mainnet && chain != Chain::Testnet {
                         cfg.feature_flags.defer_owned_object_double_spend = true;
                     }
+                    cfg.object_record_new_uid_from_hash_cost_base = Some(1);
```
tzakian, line 4573: why not the same as `object_record_new_uid_cost_base` for this?

sui-execution/latest/sui-move-natives/src/object_runtime/object_store.rs
```diff
@@ -435,6 +435,15 @@ impl<'a> ChildObjectStore<'a> {
         }
     }
 
+    /// When `parent` has a tracked root version, record the same root version for `id`.
+    /// Note that this is not observable at this time, but will be if we either allow for the
+    /// re-creation of derived objects, or if we grant access to the `id: UID` of a dynamic field.
+    pub(super) fn inherit_root_version_from_parent(&mut self, parent: ObjectID, id: ObjectID) {
+        if let Some(v) = self.inner.root_version.get(&parent).copied() {
+            self.inner.root_version.insert(id, v);
```
tzakian: should we have a check on the insertion?

sui-execution/latest/sui-move-natives/src/object.rs
```diff
@@ -115,3 +117,35 @@ pub fn record_new_uid(
...
+ * Marks `bytes` as a newly created id (as `record_new_uid` does) and, when `parent` has a tracked
+ * root version, records the same root version for `bytes`.
+ *   gas cost: object_record_new_uid_from_hash_cost_base       | this is a simple ID addition
+ **************************************************************************************************/
+pub fn record_new_uid_from_hash(
+    context: &mut NativeContext,
+    ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(ty_args.is_empty());
```
tzakian: Nit: can we switch these over to `safe_assert!`s? (like `save_unwrap!` but assert)

tzakian (approved):
Overall LGTM, couple small comments/nits.

crates/sui-protocol-config/src/lib.rs
tnowacki: Do you mean why is it 1 instead of 52? or why have a separate config?

PR #27377, opened 2026-07-21
Fix Move VM tracing bookkeeping for loaded global references.

### LINKED ISSUE #26389: sui replay --trace panics on tracer type_stack/operand_stack assert for various mainnet transactions
## Steps to Reproduce Issue                  
                           
1. Build or install `sui` with the `tracing` feature enabled (Homebrew's                                                               
   `sui` 1.70.2 has it; `cargo build --release -p sui --features tracing` on
   current `main` also reproduces).                                                                                                    
2. `sui client switch --env mainnet`          
3. `sui replay -d 6WNDjCX3W852hipq6yrHhpUaSFHSPWfTxuLKaQkgNfVL --trace -y --overwrite`
                                                                       
The same digest replays fine without `--trace`; only the trace path panics.

## Expected Result                                                                                                                     
                                                                                                                                       
`sui replay --trace` succeeds and writes `trace.json.zst` alongside the other                                                      replay artifacts.                                                                                                        
                                                                                                                                         
## Actual Result                                

```
sui replay -d 6WNDjCX3W852hipq6yrHhpUaSFHSPWfTxuLKaQkgNfVL --trace -y --overwrite
⠂: Executing transaction 6WNDjCX3W852hipq6yrHhpUaSFHSPWfTxuLKaQkgNfVL                                                                    
thread 'main' (16904895) panicked at external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs:915:9:
assertion `left == right` failed: Type stack and operand stack must be the same length sender 0
  left: 1
 right: 0
2026-04-27T21:45:35.705871Z ERROR telemetry_subscribers: panicked at external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs:915:9:
assertion `left == right` failed: Type stack and operand stack must be the same length sender 0
  left: 1
 right: 0 panic.file="external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs" panic.line=915 panic.column=9
zsh: abort      sui replay -d 6WNDjCX3W852hipq6yrHhpUaSFHSPWfTxuLKaQkgNfVL --trace -y 
```    

## Root Cause
I assume in `enter_frame_impl` the loop that resolves call arguments runs *before* the                                                         `split_off` that pops the matching arg types from `type_stack`,  however, I tried a local patch that snapshots type_stack.len() on entry to `enter_frame_impl `and truncates back to `initial_len - arg_count` on the                                                            way out regardless of error. It does fix the sender 0 panic, but the same reproducer then panics on a different signature:                                                                                       
```
  assertion `left == right` failed: Type stack and operand stack must be the same length mint 89                                         
    left: 1                                 
   right: 2  
```

I suspect this is more of a design question than a localized fix, so happy to help further — feel free to let me know if there's anything I can do (additional reproducers, testing a patch, etc.).

## System Information                                                                                                                     
                                                                                                                                     
- OS: macOS 15 (Darwin 25.4.0, arm64); also reproduced on Linux x86_64                                                                 
- Compiler: rustc 1.92 (workspace toolchain)                                                                                           
- sui versions tested:                        
    - sui 1.70.2-homebrew                                                                                                                
    - sui 1.72.0 built from main at c1d32ae3f1 with --features tracing
- Network: mainnet                                                                                                                     
- Tx digest: 6WNDjCX3W852hipq6yrHhpUaSFHSPWfTxuLKaQkgNfVL

tzakian:
## Description 

The tracer previously tracked each loaded global with a single mutable location, which could be overwritten by callee locals and left pointing at dead frames after frame exit. This could corrupt trace state and panic while resolving later call arguments during tracing.

This changes loaded global tracking to maintain a stack of GlobalLocations. The newest location is active; frame/local cleanup removes stale locations, and moving a local reifies its global reference to a snapshot before the local becomes unreadable. Same-frame aliases are tracked independently so moving or overwriting one local does not hide another live alias.

This also makes call-argument type stack handling robust by splitting argument types before fallible snapshotting, preventing trace errors from leaving type_stack inconsistent.

Fixes #26389 

## Test plan 
 
Validated with the failing Sui replay digest and existing Move CLI tracing tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
+++ b/external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
@@ -51,16 +51,66 @@ pub(crate) struct VMTracer<'a> {
 }
 
 #[derive(Debug, Clone)]
-pub(crate) enum GlobalValue {
-    // Currently loaded into a local
-    InLocal(TraceIndex, usize),
-    // Value loaded from a native function, or a value that was passed in externally (and may be
-    // passed back out).
-    Value(TraceValue),
-    // (ephemeral) Currently on the stack, but we don't have a snapshot of the value but the value is at offset
-    // `usize`. This is used when moving from a local to a stack value. We should always reify back
-    // to a value or or in local state.
-    AtStackOffset(usize),
+pub(crate) struct GlobalValue {
+    snapshot: TraceValue,
+    locations: Vec<GlobalLocation>,
+}
+
+/// A live place where a loaded global value can currently be read by the tracer.
+///
+/// `GlobalValue` keeps one stable fallback snapshot plus a stack of live locations. The newest
+/// location is active; if the stack is empty, the snapshot is used. Locals are pushed as references
+/// to globals flow through frames and are popped/removed before those locals become unreadable. When
+/// a local is moved to the operand stack, it is temporarily represented by `StackOffset` only long
+/// enough to snapshot the stack value, then the fallback snapshot is updated and the temporary
+/// location is removed.
+#[derive(Debug, Clone)]
+pub(crate) enum GlobalLocation {
+    // Currently loaded into a local.
+    Local(TraceIndex, usize),
+    // Ephemeral location used while reifying a moved local into the fallback snapshot.
+    StackOffset(usize),
+}
+
+impl GlobalValue {
+    // Start with a stable snapshot of the loaded global value.
+    fn new(snapshot: TraceValue) -> Self {
+        Self {
+            snapshot,
+            locations: vec![],
+        }
+    }
+
+    // Record that the current active copy of this global is held by a local.
+    // Track each same-frame local separately; moving one local must not hide another live alias.
+    fn push_local(&mut self, frame_identifier: TraceIndex, local_index: usize) {
+        if matches!(self.locations.last(), Some(GlobalLocation::Local(fidx, lidx)) if *fidx == frame_identifier && *lidx == local_index)
+        {
+            return;
+        }
+        self.locations
+            .push(GlobalLocation::Local(frame_identifier, local_index));
+    }
+
+    // Drop locations owned by an exiting frame so older caller locations or the snapshot become active.
+    fn remove_frame_locations(&mut self, frame_identifier: TraceIndex) {
+        while matches!(self.locations.last(), Some(GlobalLocation::Local(fidx, _)) if *fidx == frame_identifier)
+        {
+            self.locations.pop();
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
```diff
@@ -51,16 +51,66 @@ pub(crate) struct VMTracer<'a> {
...
+        }
+    }
+
+    // Record that the current active copy of this global is held by a local.
+    // Track each same-frame local separately; moving one local must not hide another live alias.
+    fn push_local(&mut self, frame_identifier: TraceIndex, local_index: usize) {
+        if matches!(self.locations.last(), Some(GlobalLocation::Local(fidx, lidx)) if *fidx == frame_identifier && *lidx == local_index)
+        {
+            return;
+        }
```
tnowacki: This might be less confusing with a `let-else` followed by an `if`

tnowacki: Or maybe we want some helpers for this? Lots of really long matches against the last items below

tnowacki: Or I guess one is a retain... 🤷

external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
```diff
@@ -785,11 +865,10 @@ impl VMTracer<'_> {
...
                 Some(val)
             })
-            .collect::<Option<Vec<_>>>()?;
+            .collect::<Option<Vec<_>>>();
 
-        let call_args_types = self
-            .type_stack
-            .split_off(self.type_stack.len() - function.arg_count());
+        let call_args_types = self.type_stack.split_off(arg_start);
+        let call_args = call_args?;
```
tnowacki, line 887: A comment about this ordering might be helpful so it doesn't get moved again later

tnowacki (approved):
I understand nothing, but that didn't stop me from leaving some comments.

external-crates/move/crates/move-vm-runtime/src/execution/tracing/tracer.rs
tzakian: some helpers probably make sense -- will do.

PR #27381, opened 2026-07-21
docs: rewrite testing page with comprehensive Move testing guide

jessiemongeon1:
## Summary
- Rewrites the testing page from a stub (only an AgentPrompt, no content) into comprehensive documentation covering unit tests, `test_scenario`, test utilities, running tests, coverage, and debugging
- Updates the Testing and Debugging index page to surface the Move Trace Debugger, Scenario Testing example, and `sui move` CLI reference as related resources
- Removes the `draft: true` flag from the testing page

## Test plan
- [ ] Verify the page renders correctly in the docs site (`pnpm start`)
- [ ] Verify all internal links resolve (`/references/cli/move`, `/references/ide/debugger`, `/getting-started/examples/scenario-testing`, `/getting-started/onboarding/local-network`, `/getting-started/onboarding/configure-sui-client`)
- [ ] Verify external links are reachable (GitHub source links for `test_utils`, `unit_test`, `debug` modules; TypeScript SDK link)
- [ ] Verify concept-map terms are present: `test_scenario`, `#[test]`, `sui move test`, `next_tx`, `take_shared`

### DIFF
--- a/docs/content/develop/testing-debugging/index.mdx
+++ b/docs/content/develop/testing-debugging/index.mdx
@@ -11,6 +11,7 @@ keywords:
   - smart contract testing
   - test framework
   - troubleshooting
+  - Move Trace Debugger
 pagination_prev: null
 goal:
   description: >-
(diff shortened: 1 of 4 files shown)


jessiemongeon1:
## Style Guide Audit

Audited **4** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

### `docs/content/develop/testing-debugging/index.mdx` (1 violation(s))

- **Line 49** — No em dashes in prose
  - Current: `—`
  - Fix: `Rewrite with comma, parentheses, or split sentence`

---

awelc (approved):
LGTM. Stamping but perhaps as future work we should consider some sort of consolidation between `develop/testing-debugging/testing` and `getting-started/examples/scenario-testing` as there is at least some overlap with respect to scenario testing between the two. In fact, the one in `develop/testing-debugging/testing` seems more introductory and perhaps better suited for "getting started" section. The other one seems longer, too, so at least perhaps we should link it as "extended example" from `develop/testing-debugging/testing`. Not sure what the best strategy here would be but it may be worth thinking about.

Also, when looking at `getting-started/examples/scenario-testing` I was thinking that the `Prerequisites` section there is perhaps a bit out of place - it looks quite general and might be useful to expose somewhere (else) on its own as in this place it's either buried a bit (if it's the only place where these things are mentioned) or repeats installation instructions that are already somewhere else

PR #27382, opened 2026-07-21
[sui-fork] refactor the data storage code and use `sui-rpc-store` instead of the hand rolled filesystem

stefan-mysten:
## Description 

This PR is a larger refactoring intended to move away from the filesystem (which had a few issues, but mainly was reinventing the wheel for data and indexes storage) and instead use the `sui-rpc-store` crate.

The main reasoning for moving to using `sui-rpc-store` is that we needed to support various indexes along with object storage, checkpoints, etc. After the initial POC implementation, when I started looking into adding more support for balance indexes, owned objects, etc., it became clear that we're reinventing the wheel. Around that time, @bmwill landed the `sui-rpc-store` crate which solves a bunch of problems (and adds a few others that are more manageable). This crate is perfectly suited for the fork tool, as it will allow us to have an embedded indexer that takes care of indexing balances, and objects, and other various stuff that are needed by the RPC layer.

While the PR is relatively big, it splits the logic as much as possible:
- remote rpc data fetching is now in its own module
- local store that is able to work with the DB to store data (objects, checkpoints, tx, etc.)
- ingestion module for enabling checkpoint indexing
- service manager that starts and manages the needed services (indexer, storage, etc), and makes them available for the tool to run
- various small fixes.

## Test plan 

Existing tests and added new tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-fork/src/runtime.rs
```diff
@@ -0,0 +1,430 @@
...
+const INDEXED_CHECKPOINT_POLL_INTERVAL: Duration = Duration::from_millis(20);
+const INDEXED_CHECKPOINT_TIMEOUT: Duration = Duration::from_secs(30);
+
+type ForkedSimulacrum = Simulacrum<OsRng, ForkStore>;
+
+/// Opened fork runtime state backed by `sui-rpc-store`.
+///
+/// The runtime owns the local RPC store and, once started, keeps the embedded
+/// `sui-rpc-store` indexer alive for local Simulacrum checkpoints.
+pub(crate) struct ForkRuntime {
```
stefan-mysten: Maybe `runtime` is not the right name for this.

crates/sui-fork/src/store.rs
```diff
@@ -67,88 +58,70 @@ use sui_types::sui_system_state::SuiSystemState;
...
 /// Cloned stores share the same inner state and local snapshot guard, so RPC readers and the local
-/// executor coordinate multi-file filesystem snapshots.
+/// executor coordinate index initialization.
 ///
 /// Implements [`SimulatorStore`] so it can be passed directly into
 /// [`simulacrum::Simulacrum::new_from_custom_state`].
 #[derive(Clone)]
-pub struct DataStore {
-    inner: Arc<DataStoreInner>,
+pub struct ForkStore {
```
stefan-mysten, line 82: This name `ForkStore` might also not be the best name.

stefan-mysten:
Marking this as draft as there are a few things to (re)consider and improve.

crates/sui-fork/src/node.rs
```diff
@@ -17,13 +17,6 @@ pub(crate) const MAINNET_GQL_URL: &str = "https://graphql.mainnet.sui.io/graphql
 pub(crate) const TESTNET_GQL_URL: &str = "https://graphql.testnet.sui.io/graphql";
 /// GraphQL endpoint for Sui devnet.
 pub(crate) const DEVNET_GQL_URL: &str = "https://graphql.devnet.sui.io/graphql";
-/// gRPC endpoint for Sui mainnet.
```
stefan-mysten, line 20: Some cleanup, we don't need this.

crates/sui-fork/src/node.rs
```diff
@@ -69,17 +62,6 @@ impl Node {
             Node::Custom(url) => url.as_str(),
         }
     }
-
-    /// Returns the gRPC endpoint URL for this node.
-    pub(crate) fn node_url(&self) -> &str {
```
stefan-mysten, line 74: Some cleanup, we don't need this.

crates/sui-fork/src/filesystem.rs
```diff
@@ -1,1102 +0,0 @@
-// Copyright (c) Mysten Labs, Inc.
```
stefan-mysten, line 1: bye bye filesystem!

crates/sui-fork/src/store.rs
stefan-mysten: Keeping the name, as it's the store backing up this forked network.

crates/sui-fork/src/local_store.rs
```diff
@@ -0,0 +1,1086 @@
...
+    ) -> anyhow::Result<()> {
+        let terminal_deleted: std::collections::BTreeSet<_> = removed_objects
+            .iter()
+            .filter_map(|removed| {
+                (removed.kind == TombstoneKind::Deleted).then_some(removed.object_id)
+            })
+            .collect();
+
+        let checkpoint = self.executing_checkpoint()?;
+        let mut batch = self.db.batch();
```
jordanjennings-mysten: because your batch is created here its possible to write objects -> fail to write a checkpoint -> and be left with objects on a checkpoint that doesn't exist. in lock-step mode I'd expect the batch variable to exist in some higher scope and passed down to the individual functions. Also curious if I try and use the same object twice with a failure on the checkpoint how versions get resolved and is there any potential double spend of an object in a single checkpoint.

crates/sui-fork/src/context.rs
```diff
@@ -34,53 +31,67 @@ pub(crate) struct CreatedCheckpointMetadata {
...
+/// manager running the embedded indexer.
 pub struct Context {
     simulacrum: Arc<RwLock<ForkedSimulacrum>>,
-    chain_identifier: Chain,
-    checkpoint_sender: broadcast::Sender<Arc<Checkpoint>>,
+    services: ServiceManager,
     checkpoint_publication_lock: Mutex<()>,
 }
 
 impl Context {
```
jordanjennings-mysten, line 42: I think you are probably generally failing fast when errors occur so this probably doesn't help much but I was sorta thinking through the usefulness of a state machine on context to prevent unexpected state changes due to either weird combinations of API calls or errors.

jordanjennings-mysten: If you only made sure you panic! when things are going wrong and lean on rocksdb to batch entire checkpoints this also becomes much less useful and you just ensure the db is always consistent and when things fail you force the user to restart the program.

crates/sui-fork/src/local_store.rs
jordanjennings-mysten: I guess you could also do a 
local_store.execute(|batch| // do things with batch ) 
something of that pattern to keep the caller code clean.

crates/sui-fork/design/storage.md
```diff
@@ -0,0 +1,389 @@
...
+Ultimately, as currently there does not seem to be a need to add additional data or column families
+to the current schemas in `sui-rpc-store`, it seems reasonable to use the same column families for both historical and locally executed data.
+
+## Known gaps
+
+The pending checkpoint buffer is memory only, so a crash mid-publication loses the
+unsealed checkpoint and its transactions, while the object rows and version-index entries
+that checkpoint had already written persist. On restart the fork resumes producing the
+same checkpoint number, so those orphaned rows still fall within a live-state read's bound
+and are served as though the checkpoint had sealed. This is the main known gap.
```
jordanjennings-mysten: I'd have to look closer but I have a feeling you can close this gap pretty easily now that you have rocksdb underlying things.

crates/sui-fork/src/tests/store_execution.rs
```diff
@@ -100,10 +113,39 @@ fn find_gas_coin(config: &NetworkConfig, owner: SuiAddress) -> Object {
         .clone()
 }
 
-fn test_data_store() -> (tempfile::TempDir, DataStore) {
+fn test_data_store() -> (tempfile::TempDir, ForkStore) {
```
jordanjennings-mysten: test_* sounds like a test

stefan-mysten: Will fix, thanks.

crates/sui-fork/src/context.rs
stefan-mysten: Yes, this is something I should look more carefully at and possibly rework it a bit. I'll do another pass on a later PR on this.

crates/sui-fork/design/storage.md
stefan-mysten: Yes, this is something that probably can be improved. I'll take a look.

crates/sui-fork/src/local_store.rs
stefan-mysten: You're right here, the objects are written ahead of the checkpoint being committed to the DB, which can lead to weird state. I'll fix it.

crates/sui-fork/src/remote.rs
```diff
@@ -0,0 +1,211 @@
...
+        Self {
+            gql,
+            forked_at_checkpoint,
+        }
+    }
+
+    /// The underlying GraphQL client, for callers with their own query needs
+    /// (seed resolution runs its own checkpoint-scoped queries).
+    pub(crate) fn gql(&self) -> &GraphQLClient {
+        &self.gql
```
jordanjennings-mysten: do the functions you need on `GraphQLClient` not make sense to give shims here and enforce your max checkpoint?

crates/sui-fork/src/store.rs
```diff
@@ -177,550 +132,364 @@ impl DataStore {
...
+        }
     }
 
     /// Query the remote GraphQL endpoint to determine the lowest checkpoint for
     /// which both checkpoint and transaction data are available.
     pub(crate) fn get_lowest_available_checkpoint(
         &self,
     ) -> anyhow::Result<CheckpointSequenceNumber> {
-        self.inner.gql.get_lowest_available_checkpoint()
+        self.inner.remote.lowest_available_checkpoint()
```
jordanjennings-mysten: if you have metadata around for a long time then this checkpoint number might be higher than your local state no? so you would want to `min(remote.get_lowest_available_checkpoint(), fork_checkpoint)` maybe?

crates/sui-fork/src/store.rs
```diff
@@ -177,550 +132,364 @@ impl DataStore {
...
             .map(|(_, contents)| contents))
     }
 
-    /// Look up a checkpoint summary by its digest. Local only: the GraphQL
-    /// checkpoint query is keyed by sequence number, so there is no remote
-    /// fallback for digest lookups.
+    /// Look up a checkpoint summary by its digest. RPC-store only: the
+    /// GraphQL checkpoint query is keyed by sequence number, so there is no
+    /// remote fallback for digest lookups.
     pub(crate) fn get_checkpoint_by_digest(
```
jordanjennings-mysten: Id be tempted to say `get_local...` in any cases where a remote version of the query is not possible.

crates/sui-fork/src/store.rs
stefan-mysten: I think that can only happen in a scenario where:
- you started a local forked network some time ago (e.g., 1 year ago)
- keep using that local forked network.
- recently, GraphQL RPC decided to only keep X recent checkpoints for objects, checkpoints, transactions, so this will return a very high value. If X is > than Y checkpoints that were done on the forked network, this would be problematic, yes.

stefan-mysten: It's possible that some RPC providers might want to do aggressive pruning, so this would be an issue.

crates/sui-fork/src/local_store.rs
stefan-mysten: Fixed. Objects, transactions, effects, events, checkpoint are all committed together, just once.

crates/sui-fork/src/remote.rs
stefan-mysten: Good catch. Removed this and reworked it.

crates/sui-fork/src/store.rs
stefan-mysten: To be on the safe side, added a min comparison here.

crates/sui-fork/src/store.rs
stefan-mysten: Fixed.

crates/sui-fork/design/storage.md
stefan-mysten: With the fixes on committing all data together, I think this is no longer an issue.

stefan-mysten: Well, maybe not. Derived indexes might still be an issue?

jordanjennings-mysten: I think you can detect and fix the derived index which makes batching unnecessary. In theory you could know which checkpoint the derived indexes is at and rebuild each checkpoint until you return the db to a consistent state.

jordanjennings-mysten (approved):
LGTM

PR #27385, opened 2026-07-22
[exploration] Block execution on unavailable system object versions instead of retrying

lxfind:
## Description

Exploration (do not merge as-is): an alternative to the retry flow in [7/7] (#27080). When execution reads a system object that has not yet caught up locally to its required version, block inside execution until the version is committed, instead of unwinding the VM, discarding effects, and re-enqueueing.

The wait lives in a `SystemObjectWaitingStore` wrapper around the backing store (stacked under the existing `TrackingBackingStore`): `get_object_by_key` on an implicitly-read system object that misses blocks on `notify_read_system_object_at_version` (via `block_in_place`, so the tokio worker is compensated). The execution layer can then assume system objects are always available at their required version — a missing one becomes an invariant violation — which deletes the whole retry propagation: `ExecutionRetryError`, the `retry_request` plumbing on both temporary stores, the `SYSTEM_OBJECT_NOT_AVAILABLE_LOCALLY` unwind and its `error.rs` conversion, the retry/unwind pairing invariant check, and the authority's discard-and-re-enqueue path. Net −139 lines, and only validators/fullnodes see the mechanism (dry-run, dev-inspect, replay, and the test runner pin versions that are already committed and never wait).

Known gaps, deliberately not solved here:
- Concurrent-execution limit: blocked transactions hold execution-driver semaphore permits. The limit is crudely bumped to `max(num_cpus, 100)` to avoid immediate starvation; a real solution (e.g. releasing the permit while blocked) is future work. Note the settlement transaction that writes the awaited root version needs a permit too, so exhausting all permits with blocked withdrawals would deadlock.
- The wait is not bounded by `within_alive_epoch`; a transaction blocked across reconfiguration would park its thread indefinitely.
- Simtests: execution blocking on a single-threaded simulator starves it (`block_in_place` is also unsupported there), so simtest CI is expected red.

Comments made stale by this change (left untouched for the author to rewrite): the `ExecutionRetryError` reference above the post-execution checker bypass in `authority.rs`; "Rate limit concurrent executions to # of cpus" in `execution_driver.rs`; the "nor triggers a system-object-availability retry" clause on `check_object_funds_sufficiency` in `object_runtime/mod.rs`; and the `SystemObjectNotAvailableLocally` variant docs in `execution_status.rs` (the variant stays for BCS compatibility but is no longer minted).

## Test plan

Exploration only; existing tests exercise the available path, and the blocking path is expected to break simtests (see above).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -74,7 +74,6 @@ use sui_types::accumulator_root::AccumulatorObjId;
 use sui_types::accumulator_root::UnsettledObjectFundsRead;
 use sui_types::dynamic_field::visitor as DFV;
 use sui_types::execution::ExecutionOutput;
-use sui_types::execution::ExecutionRetryError;
 use sui_types::execution::ExecutionTimeObservationKey;
 use sui_types::execution::ExecutionTiming;
 use sui_types::execution_params::ExecutionOrEarlyError;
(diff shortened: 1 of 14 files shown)


lxfind:
Exploration superseded by #27403.

PR #27387, opened 2026-07-22
consensus: reconnect idle block subscription streams

mwtian:
## Description

A validator can remain subscribed to a peer whose runtime has stalled while the connection stays open, preventing missed blocks from being replayed after the peer recovers. Bound stream silence to 10 seconds and resubscribe from the latest accepted round.

## Test plan

Unit tests cover reconnecting a silent stream and retaining a slow-but-live stream.

### DIFF
--- a/consensus/core/src/subscriber.rs
+++ b/consensus/core/src/subscriber.rs
@@ -11,7 +11,10 @@ use consensus_types::block::Round;
 use futures::StreamExt;
 use mysten_metrics::spawn_monitored_task;
 use parking_lot::{Mutex, RwLock};
-use tokio::{task::JoinHandle, time::sleep};
+use tokio::{
+    task::JoinHandle,
+    time::{sleep, timeout},
+};
 use tracing::{debug, error, info};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


consensus/core/src/subscriber.rs
arun-koshy, line 186: Looks like we need to apply the same type of fix around subscribe_blocks await so that it does not get stuck waiting for a response on resubscription

consensus/core/src/subscriber.rs
```diff
@@ -249,14 +259,22 @@ impl<C: ValidatorNetworkClient, S: ValidatorNetworkService> Subscriber<C, S> {
                         retries = 0;
```
arun-koshy, line 279: it looks like your changes now can cause a loop for a reachable but not proposing peer (catchup) where the retry count is never reset and replays the last proposed block which happens every 10s until the peer proposes a block?

arun-koshy (approved):

akichidis (approved):

consensus/core/src/subscriber.rs
mwtian: I think this churn is acceptable. I increased timeout to 30s.

PR #27395, opened 2026-07-22
[package-alt] Pin git deps to the peeled commit for annotated tags

mdgeorge4153:
## Description

When a git dependency's `rev` names an annotated tag, we pinned the sha of the tag object rather than of the commit the tag points at. The tag object is not reachable through branch history, and moving the tag orphans it — after which GitHub will not serve it by sha to a fresh fetch, so a checked-in `Move.lock` fails to build on a clean clone or in CI. This surfaced via MVR, where the `@mysten/attestations` dep pinned a tag object for tag `testnet/v1`.

Lightweight tags point straight at the commit and were never affected, which is likely why this went unnoticed — the existing test fixtures only created lightweight tags.

## Test plan

New tests cover both tag kinds resolving to the commit sha. The annotated-tag test first asserts that `rev-parse <tag>` differs from the commit, so it cannot pass trivially if the fixture ever stops producing a real tag object; I confirmed it fails against the old code (resolved the tag object instead of the commit).

## Release notes

For each box you select, include information after the relevant heading that describes the impact of your changes for a user of that component.

- [ ] Protocol: 
- [ ] gRPC: 
- [ ] GraphQL: 
- [x] CLI: A git dependency whose `rev` is an annotated tag now pins to the commit the tag points at instead of the tag object. Existing `Move.lock` files that pinned a tag object keep their old sha until repinned; you can use `sui move update-deps` to fix old lockfiles.
- [ ] Rust SDK:

### DIFF
--- a/external-crates/move/crates/move-package-alt/src/git/cache.rs
+++ b/external-crates/move/crates/move-package-alt/src/git/cache.rs
@@ -419,17 +419,32 @@ async fn find_branch_or_tag_sha(repo: &str, rev: &str) -> GitResult<GitSha> {
 
     // TODO(manos): Figure out if doing both lookups at once works fine (and add appropriate tests)
 
-    // Try to find a tag matching the `rev`:
-    // git ls-remote https://github.com/user/repo.git refs/heads/<tag_name>
-    let tag = run_git_cmd_with_args(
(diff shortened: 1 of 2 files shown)


stefan-mysten (approved):
Thanks for the fix.

PR #27400, opened 2026-07-22
[pkg-alt]: fix how system deps are downloaded when build environment is passed

stefan-mysten:
## Description 

This PR fixes an issue where system deps were resolved for the active environment, rather than the requested one.

`SuiFlavor::system_deps` ignored the `EnvironmentID` it was given and determined the protocol version from a gRPC client pinned to the CLI's active environment. As a result, `sui move build -e <env>` resolved manifest dependencies for `<env>` but pinned system packages (sui, std, sui_system, bridge) to the *active* environment's protocol version, diverging from `sui client switch --env <env>; sui move build` whenever the two networks run different protocol versions.

`SuiFlavor` now holds the CLI environment table instead of a single client and resolves the protocol version per environment ID, caching the result. Candidate endpoints for a chain are: configured environments whose cached chain ID matches, then environments whose chain has never been cached, then the default public endpoints of the well-known networks; the active environment is preferred within each group and environments cached as a different chain are excluded. Every candidate is verified live (its reported chain identifier must match the requested environment) before its protocol config is trusted, and each probe is bounded by a timeout. When no endpoint serves the chain, resolution falls back to the latest known system packages with a warning, as before. Offline mode (`SuiFlavor::new`) remains fully network-free.

`SuiFlavor::with_client` is renamed to `with_wallet` to reflect the new semantics.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-package-alt/src/sui_flavor.rs
+++ b/crates/sui-package-alt/src/sui_flavor.rs
@@ -1,7 +1,12 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::BTreeMap, path::PathBuf};
+use std::{
+    collections::BTreeMap,
+    path::PathBuf,
+    sync::{Arc, Mutex},
+    time::Duration,
+};
 
 use anyhow::Context;
 use async_trait::async_trait;
@@ -20,10 +25,9 @@ use sui_package_management::system_package_versions::{
     SYSTEM_GIT_REPO, latest_system_packages, system_packages_for_protocol,
 };
 use sui_protocol_config::ProtocolVersion;
-use sui_rpc_api::Client as RpcClient;
+use sui_sdk::sui_client_config::SuiEnv;
 use sui_sdk::types::{base_types::ObjectID, is_system_package};
 use sui_sdk::wallet_context::WalletContext;
-use tokio::sync::OnceCell;
 use tracing::warn;
 
 use sui_sdk::digests::chain_ids_match;
@@ -33,24 +37,35 @@ use crate::{mainnet_environment, testnet_environment};
 const EDITION: &str = "2024";
 const FLAVOR: &str = "sui";
 
+/// Give up on a candidate RPC endpoint after this long so that unreachable endpoints don't stall
+/// builds; resolution moves on to the next candidate endpoint (or falls back to the latest known
+/// system packages).
+const ENDPOINT_TIMEOUT: Duration = Duration::from_secs(10);
+
 /// The Sui-specific implementation of the [MoveFlavor] trait.
 ///
-/// Can be constructed in offline mode ([`SuiFlavor::new`]) or connected mode
-/// ([`SuiFlavor::with_client`]). In connected mode, queries the network via gRPC to determine the
-/// protocol version for correct system dependency resolution.
+/// Can be constructed in offline mode ([`SuiFlavor::new`]), where system dependencies resolve to
+/// the latest known system packages, or in connected mode ([`SuiFlavor::with_wallet`]), where the
+/// system dependencies for each environment are pinned to the protocol version reported by a gRPC
+/// endpoint serving that environment's chain.
 #[derive(Clone, Default)]
 pub struct SuiFlavor {
-    /// The gRPC client for the target network. `None` for offline/test usage.
-    client: Option<RpcClient>,
-    /// Lazily populated from gRPC when needed.
-    protocol_version: OnceCell<ProtocolVersion>,
+    /// The CLI environments used to locate a gRPC endpoint for the environment being built.
+    /// `None` for offline/test usage.
+    envs: Option<Vec<SuiEnv>>,
+    /// Alias of the CLI's active environment; preferred when several configured environments
+    /// serve the same chain.
+    active_env: Option<EnvironmentName>,
+    /// Protocol versions already resolved, keyed by environment ID. Shared across clones.
+    protocol_versions: Arc<Mutex<BTreeMap<EnvironmentID, ProtocolVersion>>>,
 }
 
 impl std::fmt::Debug for SuiFlavor {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("SuiFlavor")
-            .field("connected", &self.client.is_some())
-            .field("protocol_version", &self.protocol_version)
+            .field("connected", &self.envs.is_some())
+            .field("active_env", &self.active_env)
+            .field("protocol_versions", &self.protocol_versions.lock().unwrap())
             .finish()
     }
 }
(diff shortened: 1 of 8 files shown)


crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -61,48 +76,123 @@ impl SuiFlavor {
...
+                ),
+            }
+        }
+
+        warn!(
+            "Could not determine the protocol version of environment `{env_id}` from any \
+             configured RPC endpoint. Falling back to protocol version {}.",
+            ProtocolVersion::MAX.as_u64()
+        );
+        ProtocolVersion::MAX
```
mdgeorge4153, line 132: [CLAUDE] This is the main thing we'd change. Every failure path in `resolve_protocol_version` converges here: no candidates for the chain, every candidate erroring, every candidate timing out, and every candidate reporting a different chain. `ProtocolVersion::MAX` is the largest pin this binary can express, so it's the maximally-too-new answer — and it gets written into `[pinned.<env>]` in a committed `Move.lock`, where nothing will ever revisit it.

The premise it encodes is that the chain is at or beyond this binary's ceiling. That's systematically false for mainnet: a `main`-built CLI (any of us) or a testnet-channel release is ahead of it. So "couldn't reach mainnet, pinned your mainnet build to the newest framework I know of" is backwards precisely where it's most expensive.

Note also that the `warn!` just above is `tracing::warn!`, which isn't user-facing output in the CLI — `user_warning!` / `user_error!` in `external-crates/move/crates/move-package-alt/src/logging.rs` are. That's what makes this silent rather than merely wrong.

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -61,48 +76,123 @@ impl SuiFlavor {
...
+    }
+
+    /// The endpoints that may serve the chain identified by `env_id`, in the order they should be
+    /// tried: configured environments whose cached chain ID matches, then configured environments
+    /// whose chain has never been cached (their identity is verified when they are queried), and
+    /// finally the default public endpoints of the well-known networks. Within each group the
+    /// active environment comes first. Environments cached as a *different* chain are never
+    /// candidates.
+    fn candidate_envs(
```
mdgeorge4153, line 141: [CLAUDE] Not an objection — recording where we landed, since it affects how much this machinery is worth.

While an unreachable endpoint produces a silently-too-new pin, trying hard to find one is a *safety* argument, and safety arguments justify arbitrary complexity. Once the unreachable case resolves to keep-the-prior-pin-and-warn, this becomes a *usefulness* argument instead: someone building for mainnet who has never configured a mainnet environment should still get a working `update-deps`. That's real, but it trades against complexity rather than overriding it.

The public-endpoint fallback for the well-known networks looks like it earns its place on those grounds. The live verification in `query_protocol_version` stays regardless — that one is safety, not convenience.

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -61,48 +76,123 @@ impl SuiFlavor {
...
+        envs: &[SuiEnv],
+        active_env: Option<&str>,
+        env_id: &EnvironmentID,
+    ) -> Vec<SuiEnv> {
+        let serves_chain = |env: &SuiEnv| {
```
mdgeorge4153, line 146: [CLAUDE] `serves_chain` gets used with two different levels of trust, and I think the asymmetry is unintended.

When the cache says "match", the code correctly re-verifies live before trusting the answer — exactly the right call, and the doc comment on `query_protocol_version` gives the right reason. But when the cache says "no match", it's treated as authoritative and the environment is dropped from the candidate list entirely.

So an environment repointed at a different network, or one whose chain ID was cached before a localnet wipe, is excluded from candidates for the chain it now actually serves. Combined with the fallback below, an empty candidate list means a silent `MAX` pin.

Partly mitigated for the *active* environment, since `find_environment` refreshes its cached chain ID via `cache_chain_id()` (`crates/sui-package-alt/src/find_env.rs:85`) before we get here. It's the non-active environments where it bites.

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -178,12 +268,12 @@ impl MoveFlavor for SuiFlavor {
...
-        _: &EnvironmentID,
+        environment: &EnvironmentID,
     ) -> BTreeMap<SystemDepName, LockfileDependencyInfo> {
         let mut deps = BTreeMap::new();
         let deps_to_skip = ["DeepBook".into()];
 
-        let version = self.protocol_version().await;
+        let version = self.protocol_version(environment).await;
         let packages = match system_packages_for_protocol(version) {
             Ok((pkgs, _)) => pkgs,
```
mdgeorge4153, line 278: [CLAUDE] `system_packages_for_protocol` returns the protocol version that actually produced the pin, and it's dropped here. That's the one piece of information that distinguishes two cases which look identical from the outside:

1. the table has no entry at `version` because the framework didn't change then — the returned revision is *exactly* right, nothing to say;
2. this binary's ceiling is below `version` — the pin is correct but capped, and the user's own CLI is the binding constraint.

The doc comment on that function calls this out and tells the caller to compare against `ProtocolVersion::MAX` (`crates/sui-package-management/src/system_package_versions.rs:51-60`). Without it, a developer whose CLI is behind the chain runs `update-deps`, gets a successful no-op, still can't call the API they wanted, and has nothing pointing at their binary as the cause. Worth noting the naive test — "we returned something lower than we asked for" — is wrong and would fire constantly, since the table is sparse.

Separately, the `Err` arm just below responds to a version that was too *low* by falling back to `latest_system_packages()`, i.e. jumping to the top of the table. Unreachable in practice, since the lookup clamps downward — but it's the same reflex as the fallback above.

crates/sui-package-alt/src/sui_flavor.rs
```diff
@@ -285,3 +375,90 @@ impl Default for BuildParams {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
mdgeorge4153, line 380: [CLAUDE] All four tests exercise `candidate_envs`, which is the pure, network-free part. Nothing covers what happens when resolution *fails*, which is where the consequences are.

Mostly moot given the fallback is what's changing, but flagging it so the replacement aims its tests at the failure paths — no candidates, candidate reports the wrong chain, prior pin present vs. absent — rather than at the ordering.

mdgeorge4153:
[CLAUDE] Drafted by Claude at Michael's direction. This is the summary for the inline review comments above — it was dropped when the review was submitted, so posting it separately.

Thanks for this — the core fix is right and it's the part that matters. `system_deps` now resolves per environment ID instead of ignoring it, so `sui move build -e mainnet` stops pinning `[pinned.mainnet]` using whatever the CLI's active environment happens to be. Making the pin a function of its own lockfile key is the correct shape. The live chain-identifier check before trusting a protocol config is also exactly right: a cached chain ID can be stale, and querying the wrong chain yields a version with no relationship to the bound we care about.

We talked this through and concluded the remaining gap needs an architectural change rather than line edits, so I'm taking it over on a follow-up PR. Leaving these comments for the record so the reasoning is captured.

### The model

Writing `need` for the oldest framework that compiles the developer's code, `network` for the version running on the target chain, and `CLI` for the newest framework revision in this binary's table, we want `need <= pinned <= network` for correctness and `pinned <= CLI` for feasibility. Two facts shape the rest:

- **Getting it wrong is asymmetric.** Too old is a compile error on the developer's own machine, at the moment they hit it, with a fix they can perform themselves (`update-deps`). Too new compiles clean and fails at publish with a linker error against `0x2` — late, cryptic, and because the pin is committed to `Move.lock`, inflicted on everyone who builds the package rather than just the person who created it.
- **`network` only ever rises.** Protocol versions are monotonic, so a pin that satisfied the bound when it was made still does. Staleness can only push a pin toward "too old", never toward "too new".

### The main issue

Every failure path in the new resolution logic lands on `ProtocolVersion::MAX` — no candidates for the chain, all candidates erroring, all timing out, all reporting a different chain. That's the maximally-too-new pin, chosen silently, written durably into a shared lockfile. The premise it encodes, that the chain is at or beyond this binary's ceiling, is systematically false for mainnet: anyone on a `main`-built or testnet-channel CLI is ahead of it. So the guess is worst exactly where the cost is highest.

When no endpoint answers, the one thing resolution must not do is invent a version. That means `system_deps` needs to be able to fail, and the caller has to decide what a failure means. We considered carrying the previous pin forward — by monotonicity it is provably still within bounds — but `Move.lock` doesn't record which pins came from system-dep resolution (`Pinned` has no system variant; they serialize as ordinary git dependencies), so there's no clean way to locate it without a lockfile format change. The follow-up will most likely just error, with a message naming the environment and pointing at configuring an endpoint for it. Either way the property that matters holds: no too-new pin is ever written.

### Two smaller things

The resolved protocol version is still discarded, so the case where the *CLI* rather than the chain is the binding ceiling goes undiagnosed — see the comment on `system_deps`. And the ephemeral publish path still constructs `SuiFlavor::new()` (`crates/sui/src/client_commands.rs:4019`), so `test-publish --build-env E` pins to latest regardless of `E`: same bug, different call site, and untouched here despite being in scope of the PR title.

Full write-up of the intended behavior will be on the follow-up PR.

PR #27403, opened 2026-07-22
Replace system-object unavailability retry with a blocking wait in execution

lxfind:
## Description

Reworks the system-object availability mechanism from #27071. There, a transaction that read a system object not yet caught up locally unwound the VM with a dedicated status code, recorded an out-of-band retry request, and had its effects discarded and the certificate re-enqueued once the object caught up — propagation machinery spanning the temporary stores, `InnerTemporaryStore`, the error converter, and the authority.

Instead, execution now just waits: the temporary store fetches the object through a single `ObjectStore::load_implicitly_read_system_object` call, which returns the object at the required version, blocking first if it has not been committed yet. The default body is a plain keyed read — correct for every store that pins already-committed versions (dry-run, replay, tests) — and the writeback cache overrides it with read-or-block-then-read via `NotifyRead::read_one_blocking` from #27509's blocking sync primitives, on which this PR is stacked (the retry path's separate `notify_read_system_object_at_version` cache primitive is folded into it). A miss that survives the wait is an invariant violation. This deletes `ExecutionRetryError`, the `retry_request` plumbing, the discard-and-re-enqueue path, and the `SystemObjectNotAvailableLocally` execution-failure kind and `SYSTEM_OBJECT_NOT_AVAILABLE_LOCALLY` status code outright (the kind was the last BCS variant and is never committed to effects, so removing it is compatibility-safe).

A parked execution must not be able to starve the very transaction that writes the version it waits for: if every execution-driver permit were held by blocked withdrawers, the settlement transaction could never execute and the driver would deadlock. Two mechanisms prevent this, both from #27509's primitives plus driver routing. The driver installs each execution's permit on its blocking thread via `set_execution_permit`, and the blocking wait releases it before parking without reacquiring it (transient over-subscription resolves as executions complete). Additionally, transactions that take an implicitly-read system object as a *mutable* input — the sound classification, since exactly those transactions advance the versions that blocked executions wait on — draw from a dedicated permit pool, and the permit is acquired inside the spawned task rather than the dispatch loop so a writer can never queue head-of-line behind a blocked user transaction. Under `msim` the wait runs on the simulator's deterministic blocking pool and yields the thread quantum between checks (handled inside the primitive). One known open issue, deliberately not addressed here: a transaction blocked across reconfiguration holds its execution-lock read guard while parked (the old path bounded the wait with `within_alive_epoch`).

This PR also folds in the two companion pieces so the whole implicitly-read system-object story is reviewable in one place: recording the read into effects (previously #27079 — the read is emitted as a `ReadOnlyRoot` unchanged consensus object, reproduced by checkpoint execution via `assign_versions_from_effects`, with `input_consensus_objects` renamed to `accessed_consensus_objects` to reflect that undeclared system reads now appear there), and the simulation-side wiring (previously #27128).

The sequenced executor entry points accept exact per-object versions via a single `SystemObjectVersions` struct (today just the accumulator version; the struct's constructor and id-keyed getter concentrate future additions in one impl), so simulation semantics cannot leak onto the sequenced path: consensus-sequenced execution passes its assigned versions and blocks until they are committed, simulacrum assigns the current versions at execution time (`from_latest_in_store` — equivalent to real assignment under its sequential in-memory model), and replay/backtest reconstruct the versions the original execution recorded in its effects (`from_effects`). Dev-inspect/dry-run/simulate instead run in an execution-wide `Latest` mode, internal to the executor: each implicit read resolves to the object's latest version at read time, so a version being pruned mid-simulation can never strand or crash the node. This folds in what was previously #27128.

Like #27071, the mechanism is inert until a consumer calls the availability check; the follow-up PR moves the object-funds withdraw sufficiency check into the VM on top of it.

## Test plan

The transactional test runner's validator backend now runs consensus version assignment for every committed transaction, as production consensus does, so committed transactions in transactional tests carry an assigned accumulator version. This lets the follow-up PR's Move-level tests exercise the exact-version path end to end; on this PR it is behaviorally invisible (the full adapter transactional suite is unchanged).

No consumer exists on main, so behavior is unchanged and covered by existing suites; the notify-based wait keeps unit-test coverage through the new blocking getter (blocks while the version is uncommitted, resolves on the write at the target version, immediate when already committed, and releases the execution permit while blocked). The wait path itself gets exercised end to end by the follow-up in-VM funds-check PR, whose stress test drives the not-caught-up case under simulation.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


lxfind:
Superseded by #27523 (identical content, clean history).

PR #27406, opened 2026-07-22
[move-ide] Finessed testing for quick fixes

awelc:
## Description 

This PR simplifies how the tests for quick fixes are written. Previously the text of the diagnostic had to be provided by-hand in the .ide spec file and now it's obtained directly from compiler's diagnostics which makes the whole process more convenient and less error-prone

## Test plan 

All tests must pass under the new change


tzakian (approved):
Nice!

PR #27410, opened 2026-07-22
[move-compiler] Preserve macro expansion context through CFGIR

awelc:
## Description 

This is the first of two PRs whose goal is to encode information about macro frame transition needed for proper debugging in presence of macros (e.g., so that the debugger know precisely where the code execution is at any given moment, including code inside inlined macros).

This PR propagates this information from where it is originally introduced to the level of CFGIR. The subsequent [PR](https://github.com/MystenLabs/sui/pull/27411) threads this information further, up to debug info (aka. source map) generation.

This PR consists of two commits:
- the first commit represents the original design from the time when it was deemed sufficient to encode macro-frame information at the level of commands
-  the second commits is a result of realizing that command-level encoding does not provide enough fidelity to record all the information precisely (in other words, under command-level-only encoding some frame transitions, particularly those related to macro arguments, would be lost, and others could be incorrect)

## Test plan 

End-to-end macro-frame tests are included in the follow-up PR

### DIFF
--- a/external-crates/move/crates/move-compiler/src/cfgir/ast.rs
+++ b/external-crates/move/crates/move-compiler/src/cfgir/ast.rs
@@ -10,10 +10,7 @@ use crate::{
         Var, Visibility,
     },
     parser::ast::{ConstantName, DatatypeName, ENTRY_MODIFIER, FunctionName, TargetKind},
-    shared::{
-        ast_debug::*, macro_expansion::MacroExpansionInfo, program_info::TypingProgramInfo,
-        unique_map::UniqueMap,
-    },
+    shared::{ast_debug::*, program_info::TypingProgramInfo, unique_map::UniqueMap},
 };
 use move_core_types::runtime_value::MoveValue;
 use move_ir_types::location::*;
@@ -111,53 +108,10 @@ pub type BasicBlocks = BTreeMap<Label, BasicBlock>;
 
 pub type BasicBlock = VecDeque<SyntaxCommand>;
 
-/// Syntactic context for code that was produced by an expansion (e.g., a macro), tracking the
-/// chain of expansions that produced it, innermost first.
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct SyntaxInfo {
-    pub info: SyntaxInfoEntry,
-    pub prev: Option<Arc<SyntaxInfo>>,
-}
-
-/// A single kind of syntactic context. An enum so that other kinds of
-/// compiler-tracked context (beyond macro expansion) can be added over time.
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub enum SyntaxInfoEntry {
-    MacroExpansion(MacroExpansionInfo),
-}
-
-/// A location plus the expansion context (if any) the code at that location came from.
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct SyntaxLoc {
-    pub loc: Loc,
-    pub syntax_info: Option<Arc<SyntaxInfo>>,
-}
-
-/// Like `Spanned<T>`, but carrying a `SyntaxLoc` instead of a bare `Loc`.
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct SyntaxSpanned<T> {
-    pub sloc: SyntaxLoc,
-    pub value: T,
-}
+pub(crate) use crate::shared::syntax_info::ssp;
+pub use crate::shared::syntax_info::{SyntaxInfo, SyntaxInfoEntry, SyntaxLoc, SyntaxSpanned};
 
 pub type SyntaxCommand = SyntaxSpanned<Command_>;
-
-/// Macro used to create a tuple-like pattern match for SyntaxSpanned, mirroring `sp!`
-macro_rules! ssp {
-    (_, $value:pat) => {
-        crate::cfgir::ast::SyntaxSpanned { value: $value, .. }
-    };
-    ($sloc:pat, _) => {
-        crate::cfgir::ast::SyntaxSpanned { sloc: $sloc, .. }
-    };
-    ($sloc:pat, $value:pat) => {
-        crate::cfgir::ast::SyntaxSpanned {
-            sloc: $sloc,
-            value: $value,
-        }
-    };
-}
-pub(crate) use ssp;
 
 #[derive(PartialEq, Eq, Debug, Clone, Copy)]
 pub enum LoopEnd {
@@ -189,28 +143,6 @@ impl LoopEnd {
             LoopEnd::Unused => false,
             LoopEnd::Target(t) => *t == lbl,
         }
-    }
-}
-
-impl SyntaxInfo {
-    pub fn new(info: SyntaxInfoEntry, prev: Option<Arc<SyntaxInfo>>) -> Self {
-        Self { info, prev }
-    }
-}
-
-impl SyntaxLoc {
-    pub fn new(loc: Loc, syntax_info: Option<Arc<SyntaxInfo>>) -> Self {
-        Self { loc, syntax_info }
-    }
-}
-
-impl<T> SyntaxSpanned<T> {
-    pub fn new(sloc: SyntaxLoc, value: T) -> Self {
-        Self { sloc, value }
-    }
-
-    pub fn loc(&self) -> Loc {
-        self.sloc.loc
     }
 }
 
@@ -447,12 +379,6 @@ impl AstDebug for (&Label, &BasicBlock) {
     fn ast_debug(&self, w: &mut AstWriter) {
         w.write(format!("label {}:", (self.0).0));
         w.indent(4, |w| w.semicolon(self.1, |w, cmd| cmd.ast_debug(w)))
-    }
-}
-
-impl<T: AstDebug> AstDebug for SyntaxSpanned<T> {
-    fn ast_debug(&self, w: &mut AstWriter) {
-        self.value.ast_debug(w)
     }
 }
 
--- a/external-crates/move/crates/move-compiler/src/cfgir/ast.rs
+++ b/external-crates/move/crates/move-compiler/src/cfgir/ast.rs
@@ -6,11 +6,14 @@ use crate::{
     diagnostics::filter::FilterScope,
     expansion::ast::{Attributes, Friend, ModuleIdent, Mutability},
     hlir::ast::{
-        BaseType, Command, Command_, EnumDefinition, FunctionSignature, Label, SingleType,
-        StructDefinition, Var, Visibility,
+        BaseType, Command_, EnumDefinition, FunctionSignature, Label, SingleType, StructDefinition,
+        Var, Visibility,
     },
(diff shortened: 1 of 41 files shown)


cgswords (commented):
A few comments that don't anchor to changed lines, plus inline comments below.

**`src/typing/visitor.rs` (`same_value_exp_`, ~line 1327):** Does this need a `MacroExpansion` pass-through arm alongside `Dereference`/`TempBorrow`/`Cast`/`Annotate`? Every `$x` use inside a macro body is now `Annotate(MacroExpansion(Argument, …))`. `Annotate` unwraps, `MacroExpansion` falls into `_ => false`, so `combinable_comparisons` regresses to false negatives inside macros (e.g. `$a < $b || $a == $b`). Please add the arm plus a lint test with macro-argument operands pinning the behavior.

**`src/linters/unnecessary_conditional.rs` (`extract_value`, line 84):** Same class of issue: `UnannotatedExp_::Annotate(e, _) => extract_value(e)` should also look through `MacroExpansion(_, e)`, or `if (c) $a else $b` with literal arguments stops firing. Worth a quick audit of all typing-AST matches that unwrap `Annotate` with a catch-all, making them explicit instead.

**Testing:** Deferring end-to-end macro-frame tests to #27411 is reasonable, but this PR carries observable behavior of its own (the `ast_debug` rendering of expansion regions, the `assignment_lvalue_syntax_info` heuristic, the linter interactions above), and we should test it if possible.

external-crates/move/crates/move-compiler/src/shared/syntax_info.rs
```diff
@@ -0,0 +1,134 @@
...
+/// A location plus the expansion context (if any) the code at that location came from.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SyntaxLoc {
+    pub loc: Loc,
+    pub syntax_info: SyntaxContext,
+}
+
+/// Like `Spanned<T>`, but carrying a `SyntaxLoc` instead of a bare `Loc`.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SyntaxSpanned<T> {
```
cgswords, line 40: `Spanned` implements value-only `PartialEq`/`Hash`/`Ord` on purpose (locations never participate in equality anywhere in the compiler); deriving here makes `SyntaxSpanned` equality incorrectly include the loc and `syntax_info`.

external-crates/move/crates/move-compiler/src/shared/syntax_info.rs
```diff
@@ -0,0 +1,134 @@
...
+//! carried on HLIR expressions and CFGIR commands via [`SyntaxSpanned`].
+
+use crate::shared::{ast_debug::*, macro_expansion::MacroExpansionInfo};
+use move_ir_types::location::Loc;
+use std::sync::Arc;
+
+/// Syntactic context for code that was produced by an expansion (e.g., a macro), tracking the
+/// chain of expansions that produced it, innermost first.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SyntaxInfo {
```
cgswords, line 15: The naming trio is hard to keep straight: `SyntaxContext` is `Option<Arc<SyntaxInfo>>`, yet fields named `syntax_info` hold a `SyntaxContext`, and `SyntaxInfoEntry` is the payload of a `SyntaxInfo`. Each level's name should say what it means.

external-crates/move/crates/move-compiler/src/shared/syntax_info.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub info: SyntaxInfoEntry,
+    pub prev: Option<Arc<SyntaxInfo>>,
+}
+
+/// A single kind of syntactic context. An enum so that other kinds of
+/// compiler-tracked context (beyond macro expansion) can be added over time.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum SyntaxInfoEntry {
```
cgswords, line 23: Consider using `MacroExpansionInfo` directly until a second context kind actually exists.

external-crates/move/crates/move-compiler/src/shared/syntax_info.rs
```diff
@@ -0,0 +1,134 @@
...
+pub(crate) use ssp_macro::ssp;
+
+impl SyntaxInfo {
+    pub fn new(info: SyntaxInfoEntry, prev: Option<Arc<SyntaxInfo>>) -> Self {
+        Self { info, prev }
+    }
+
+    /// Compact one-line rendering of the expansion chain (outermost first)
+    /// for debug output, e.g. `[0x1::m::foo, lambda]`.
+    pub fn debug_chain(&self) -> String {
```
cgswords, line 101: I wonder if we could manifest this somehow as a test case. A nested macro + lambda + argument case would both pin the rendering and serve as living documentation of chain order (outermost first).

external-crates/move/crates/move-compiler/src/cfgir/ast.rs
```diff
@@ -106,7 +106,12 @@ pub struct Function {
 
 pub type BasicBlocks = BTreeMap<Label, BasicBlock>;
 
-pub type BasicBlock = VecDeque<Command>;
+pub type BasicBlock = VecDeque<SyntaxCommand>;
+
+pub(crate) use crate::shared::syntax_info::ssp;
```
cgswords, line 111: nit: The re-export chain through `cfgir::ast` adds an indirection layer; do we do this for `sp`? If not, just import directly from `crate::shared::syntax_info`.

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -147,6 +155,109 @@ impl<'env> Context<'env> {
         self.label_count = 0;
         self.loop_bounds = BTreeMap::new();
     }
+
+    /// Runs `body` with `info` installed as the current syntactic context,
+    /// restoring the previous one afterwards. The chain itself is built
+    /// during HLIR lowering (so that statement- and expression-level uses
+    /// share one identity per expansion); since expansions nest, its `prev`
+    /// must equal the context in place here (checked in debug builds).
+    fn with_syntax_info<T, F>(&mut self, info: Arc<SyntaxInfo>, body: F) -> T
```
cgswords, line 164: The chain-consistency check is only `debug_assert!`, so a release-build compiler silently proceeds with wrong attribution if HLIR and CFGIR ever disagree. An `ice!` or `ice_assert!` would be more appropriate.

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -147,6 +155,109 @@ impl<'env> Context<'env> {
...
+    ///   context.
+    /// * In (3), the result temporary receives the macro call's location,
+    ///   outside `add1`'s body. Searching the current context chain therefore
+    ///   reaches the caller's context, or none for ordinary user code.
+    ///
+    /// This choice is recorded while translating HLIR, where the assignment's
+    /// original RHS is still present. CFG optimizations may later replace that
+    /// RHS with an expression from another macro invocation; the assignment
+    /// must still remain associated with the invocation that created it.
+    fn assignment_lvalue_syntax_info(
```
cgswords, line 217: This heuristic encodes HLIR's location conventions (`bind_exp` gives temps the RHS loc; result temps get the call loc). Those conventions are load-bearing now, and nothing pins them. A couple of targeted tests in this PR would help with future drift issues.

external-crates/move/crates/move-compiler/src/typing/macro_expand.rs
```diff
@@ -20,7 +25,9 @@ use move_ir_types::location::*;
 use move_proc_macros::growing_stack;
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 
-type LambdaMap = BTreeMap<Var_, (N::Lambda, Loc, Vec<Type>, Type)>;
+/// Maps a lambda parameter to the lambda, its expression location, its
+/// function-type location, its parameter types, and its result type.
+type LambdaMap = BTreeMap<Var_, (N::Lambda, Loc, Loc, Vec<Type>, Type)>;
```
cgswords, line 30: nit: Would naming these fields to avoid confusion be better? 'No' is a fine answer.

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -2469,9 +2597,22 @@ fn value_evaluation_order(
...
+    //
+    //     macro fun add1($x: u64): u64 { $x + 1 }
+    //     macro fun double($x: u64): u64 { $x + $x }
+    //     fun test(v: u64): u64 { double!(add1!(v)) }
+    //
+    // lowering `double`'s body binds the value of its argument `add1!(v)`
+    // to a temp. The temp's read is located at `add1!(v)` (in `test`) and
+    // must stay attributed to the argument, not to the body of `double`
+    // where the binding takes place.
+    var_exp.exp.sloc.syntax_info = e.exp.sloc.syntax_info.clone();
```
cgswords, line 2615: This is asymmetrical: we only re-stamp the top-level `var_exp` with the bound value's context, while `reattribute_expansion_result` recurs into `Multiple`/`Freeze`. Why?

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -1054,6 +1067,37 @@ fn tail_block(
...
+///     mut_refs($x, $y)
+/// }
+/// fun test(x: &mut u64, y: &mut u64): (&u64, &u64) { through!(x, y) }
+/// ```
+///
+/// `through` is evaluated in its macro context, but its implicit conversions
+/// to immutable references happen while returning the result to `test`.
+/// Lowering represents this result with `Multiple` and `Freeze` wrappers, so
+/// those wrappers and their result reads all receive the caller's context.
+fn reattribute_expansion_result(e: &mut H::Exp, syntax_context: &SyntaxContext) {
```
cgswords, line 1086: nit: This recursion lacks `#[growing_stack]`, unlike the neighboring traversals.

external-crates/move/crates/move-compiler/src/naming/ast.rs
```diff
@@ -410,6 +410,9 @@ pub struct Lambda {
     pub return_label: BlockLabel,
     pub use_fun_color: Color,
     pub body: Box<Exp>,
+    /// The original argument span for a lambda forwarded between macro calls.
+    /// Populated on its first binding and used only for expansion provenance.
+    pub macro_expansion_loc: Option<Loc>,
```
cgswords, line 415: nit: Maybe call this `original_lambda_loc` to clarify, with a one-line doc comment?

external-crates/move/crates/move-compiler/src/shared/syntax_info.rs
awelc: Fixed!

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
awelc: Changed to `ice_assert!`

PR #21454, opened 2025-03-10
[move] Source maps become debug info

awelc:
## Description 

This PR replaces a concept of a source map with a concept of debug info. In this PR:
- we add versioning to (what is now) debug info
- rename directory where (what is now) debug info is deposited from `source_maps` to `debug_info`
- rename the binary file extension for (what is now) debug info from `mvsm` to `mvd`

Renaming of internal data structure(s), related functions, comments, etc. will happen in a separate PR.

## Test plan 

All tests must pass (modified disassembly test to reflect debug info format differing from source map format by the inclusion of a version number). 

Manually tested that the correct renamed subdirectory is crated in `build` and that extensions for debug info also reflect renaming.

PR #27411, opened 2026-07-22
[move-compiler] Record macro frames in source maps

awelc:
## Description 

This is the first of two PRs whose goal is to encode information about macro frame transition needed for proper debugging in presence of macros (e.g., so that the debugger know precisely where the code execution is at any given moment, including code inside inlined macros).

In this PR we record macro, lambda, and by-name argument frames in source maps while preserving per-instruction context through bytecode optimization.

## Test plan 

Test harness was extended to verify correctness of macro frame transitions. All tests must pass

### DIFF
--- a/external-crates/move/crates/move-bytecode-source-map/src/source_map.rs
+++ b/external-crates/move/crates/move-bytecode-source-map/src/source_map.rs
@@ -27,8 +27,51 @@ use std::{collections::BTreeMap, ops::Bound, path::PathBuf};
 
 pub type SourceName = (String, Loc);
 
-/// The current version of the trace format.
-const CURRENT_VERSION: u64 = 2;
+/// The current version of the source map format.
+const CURRENT_VERSION: u64 = 3;
+
+/// The type of macro expansion scope a frame represents.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum MacroFrameKind {
+    /// Expansion of a macro function body.
+    MacroBody {
+        module_addr: AccountAddress,
+        module_name: Identifier,
+        function_name: Identifier,
+    },
+    /// Expansion of a lambda invocation inside a macro.
+    Lambda,
+    /// Evaluation of a by-name argument substituted into a macro body.
+    Argument,
+}
+
+/// A single macro expansion frame stored in the source map for debugger
+/// support. Produced during code generation by converting the compiler's
+/// internal expansion records into bytecode-level entries with parent
+/// references resolved to indices.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MacroFrameInfoEntry {
+    /// The type of expansion this frame represents.
+    pub kind: MacroFrameKind,
+    /// Location of the expanded construct — used by the debugger to determine
+    /// which source file to display when stepping into this frame. In
+    /// principle redundant with bytecode instruction locations (which point
+    /// to the same file), but stored explicitly for convenience:
+    /// - `MacroBody`: the macro function definition.
+    /// - `Lambda`: the lambda body definition.
+    /// - `Argument`: the argument expression at the call site.
+    pub source_loc: Loc,
+    /// Location of the site that triggered this expansion — shown in the
+    /// debugger's call stack as the "called from" position in the parent
+    /// frame:
+    /// - `MacroBody`: the `macro_name!(args)` invocation.
+    /// - `Lambda`: the `$f(args)` call inside the macro body.
+    /// - `Argument`: the `$x` reference inside the macro body.
+    pub call_loc: Loc,
+    /// Index of the parent frame in the function's `macro_frame_info` vec,
+    /// or `None` for top-level macro calls (no enclosing expansion).
+    pub parent_index: Option<u32>,
+}
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct StructSourceMap {
@@ -93,6 +136,17 @@ pub struct FunctionSourceMap {
 
     /// Whether this function is a native function or not.
     pub is_native: bool,
+
+    /// Metadata for each macro/lambda/argument expansion frame.
+    #[serde(default)]
+    pub macro_frame_info: Vec<MacroFrameInfoEntry>,
+
+    /// Maps bytecode offsets to macro frame indices. Stores only the offsets
+    /// where the active frame changes — consecutive instructions with the
+    /// same frame are not repeated. `None` means user code (no macro frame),
+    /// `Some(i)` is an index into `macro_frame_info`.
+    #[serde(default)]
+    pub macro_frame_map: Vec<(CodeOffset, Option<u32>)>,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
@@ -232,6 +286,8 @@ impl FunctionSourceMap {
             is_native,
             nops: BTreeMap::new(),
             labels: BTreeMap::new(),
+            macro_frame_info: Vec::new(),
+            macro_frame_map: Vec::new(),
         }
     }
 
@@ -258,6 +314,15 @@ impl FunctionSourceMap {
             }
             _ => (),
         };
+    }
+
+    /// Record the macro frame index for a given bytecode offset.
+    /// Only appends when the index differs from the most recent entry,
+    /// keeping the map compact.
+    pub fn add_macro_frame_mapping(&mut self, start_offset: CodeOffset, frame_index: Option<u32>) {
+        if self.macro_frame_map.last().map(|(_, c)| *c) != Some(frame_index) {
+            self.macro_frame_map.push((start_offset, frame_index));
+        }
     }
 
     /// Record the code offset for an Nop label
@@ -653,6 +718,18 @@ impl SourceMap {
         Ok(())
     }
 
+    /// Iterates over all function source maps keyed by table index.
+    pub fn function_source_maps(&self) -> impl Iterator<Item = (TableIndex, &FunctionSourceMap)> {
+        self.function_map.iter().map(|(k, v)| (*k, v))
+    }
+
+    /// Mutable iterator over all function source maps keyed by table index.
+    pub fn function_source_maps_mut(
+        &mut self,
+    ) -> impl Iterator<Item = (TableIndex, &mut FunctionSourceMap)> {
+        self.function_map.iter_mut().map(|(k, v)| (*k, v))
+    }
+
     pub fn get_function_source_map(
         &self,
         fdef_idx: FunctionDefinitionIndex,
@@ -807,6 +884,13 @@ impl SourceMap {
             }
             for (_, loc) in function_map.code_map.iter_mut() {
                 *loc = Loc::new(file_hash, loc.start(), loc.end());
+            }
+            // Defensive for generated source maps. Disassembly bytecode maps currently
+            // leave macro metadata empty because bytecode does not retain macro frames.
+            for frame in function_map.macro_frame_info.iter_mut() {
+                frame.source_loc =
+                    Loc::new(file_hash, frame.source_loc.start(), frame.source_loc.end());
+                frame.call_loc = Loc::new(file_hash, frame.call_loc.start(), frame.call_loc.end());
             }
         }
     }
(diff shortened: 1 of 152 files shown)


cgswords (commented):
Comments inline — mostly questions about the source-map format/version story, plus a few structural suggestions. The test coverage in this one is great.

external-crates/move/crates/move-bytecode-source-map/src/source_map.rs
```diff
@@ -27,8 +27,51 @@ use std::{collections::BTreeMap, ops::Bound, path::PathBuf};
 
 pub type SourceName = (String, Loc);
 
-/// The current version of the trace format.
-const CURRENT_VERSION: u64 = 2;
+/// The current version of the source map format.
+const CURRENT_VERSION: u64 = 3;
```
cgswords, line 31: Is it okay that new tooling cannot read pre-existing v2 `.mvsm` files and old tooling cannot read v3, while the `version` field is never consulted on load? (On-disk maps go through `bcs::from_bytes`, which is positional — `#[serde(default)]` only helps the JSON path.) If the policy is "source maps are transient build artifacts, regenerate them" (as #21454 implies), confirm that and consider version checking?

external-crates/move/crates/move-bytecode-source-map/src/source_map.rs
```diff
@@ -260,6 +316,15 @@ impl FunctionSourceMap {
         };
     }
 
+    /// Record the macro frame index for a given bytecode offset.
+    /// Only appends when the index differs from the most recent entry,
+    /// keeping the map compact.
+    pub fn add_macro_frame_mapping(&mut self, start_offset: CodeOffset, frame_index: Option<u32>) {
```
cgswords, line 322: nit: When the map is empty, the first entry is pushed even when it's `(0, None)` (user code at offset 0). Should we avoid this?

external-crates/move/crates/move-bytecode-source-map/src/source_map.rs
```diff
@@ -27,8 +27,51 @@ use std::{collections::BTreeMap, ops::Bound, path::PathBuf};
...
 pub type SourceName = (String, Loc);
 
-/// The current version of the trace format.
-const CURRENT_VERSION: u64 = 2;
+/// The current version of the source map format.
+const CURRENT_VERSION: u64 = 3;
+
+/// The type of macro expansion scope a frame represents.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum MacroFrameKind {
```
cgswords, line 35: These mirror `MacroExpansionKind`/`MacroExpansionInfo` in the compiler almost field-for-field; maybe that's just life?

external-crates/move/crates/move-bytecode-source-map/src/source_map.rs
```diff
@@ -93,6 +136,17 @@ pub struct FunctionSourceMap {
...
+    /// Metadata for each macro/lambda/argument expansion frame.
+    #[serde(default)]
+    pub macro_frame_info: Vec<MacroFrameInfoEntry>,
+
+    /// Maps bytecode offsets to macro frame indices. Stores only the offsets
+    /// where the active frame changes — consecutive instructions with the
+    /// same frame are not repeated. `None` means user code (no macro frame),
+    /// `Some(i)` is an index into `macro_frame_info`.
+    #[serde(default)]
+    pub macro_frame_map: Vec<(CodeOffset, Option<u32>)>,
```
cgswords, line 149: I am not sure about the consumer story here: how does the trace adapter / debugger interpret the change-points encoding?

external-crates/move/crates/move-bytecode-source-map/src/source_map.rs
```diff
@@ -27,8 +27,51 @@ use std::{collections::BTreeMap, ops::Bound, path::PathBuf};
...
+    /// The type of expansion this frame represents.
+    pub kind: MacroFrameKind,
+    /// Location of the expanded construct — used by the debugger to determine
+    /// which source file to display when stepping into this frame. In
+    /// principle redundant with bytecode instruction locations (which point
+    /// to the same file), but stored explicitly for convenience:
+    /// - `MacroBody`: the macro function definition.
+    /// - `Lambda`: the lambda body definition.
+    /// - `Argument`: the argument expression at the call site.
+    pub source_loc: Loc,
```
cgswords, line 63: The comment says this is "in principle redundant" with instruction locations. Can you explain when the debugger actually needs it stored explicitly rather than deriving it?

external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -400,6 +415,160 @@ fn module(
...
+        let func_idx = fdef_idx as usize;
+        if func_idx >= compiled_module.function_defs.len() {
+            continue;
+        }
+        let fdef = &compiled_module.function_defs[func_idx];
+        let fhandle = &compiled_module.function_handles[fdef.function.0 as usize];
+        let fname_str = compiled_module.identifiers[fhandle.name.0 as usize].as_str();
+        let fname_sym = Symbol::from(fname_str);
+
+        let Some(info_data) = function_syntax_info.get(&fname_sym) else {
```
cgswords, line 442: The join between compiled functions and `function_syntax_info` is by function name string -> `Symbol`, and a lookup miss silently `continue`s. A miss is only legitimate for natives; for a function with code, that's dropped debug info with no signal. Consider an ice_assert or some other way to raise this issue (we can even filter out the natives via the native list).

external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -1049,14 +1240,20 @@ fn label(lbl: H::Label) -> IR::BlockLabel_ {
     IR::BlockLabel_(format!("{}", lbl).into())
 }
 
+/// Pushes an instruction, recording the current syntactic context for it in
+/// parallel (the source of the source map's macro frame info).
+fn push_instr(context: &mut Context, code: &mut IR::BytecodeBlock, instr: IR::Bytecode) {
```
cgswords, line 1245: The pairing invariant is enforced only by convention plus `debug_assert_alignment`. A future raw `code.push` compiles fine and silently skews the parallel vec. A small builder owning both vecs (one `push` method, destructured at block end) would make the illegal state unrepresentable rather than merely detected.

external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -400,6 +415,160 @@ fn module(
...
+            continue;
+        };
+
+        // The per-instruction data is indexed by bytecode offset, which
+        // relies on the IR compiler emitting exactly one instruction per IR
+        // bytecode. If the counts ever diverge, drop this function's macro
+        // frame info -- missing debug info is better than misattributed
+        // frames.
+        let code_len = fdef.code.as_ref().map_or(0, |code| code.code.len());
+        if info_data.len() != code_len {
```
cgswords, line 452: Could we get a test for the drop path? When the guard fires and a function's frame info is discarded, what does the debugger actually see for that function?

external-crates/move/crates/move-compiler/src/to_bytecode/context.rs
```diff
@@ -32,6 +34,16 @@ pub struct Context<'a> {
     current_module: Option<&'a ModuleIdent>,
     seen_datatypes: BTreeSet<(ModuleIdent, DatatypeName)>,
     seen_functions: BTreeSet<(ModuleIdent, FunctionName)>,
+    /// Syntactic context (macro expansion chain) of the command currently
+    /// being compiled, taken from its `SyntaxLoc`. `None` = user code.
+    pub syntax_info: Option<Arc<SyntaxInfo>>,
```
cgswords, line 39: nit: `syntax_info`, `current_block_info`, and `function_syntax_info` are the only `pub` fields in the struct, and they're only touched from `to_bytecode`. `pub(super)`, or private with `push_instr` as a method, seems fine.

external-crates/move/crates/move-compiler/src/to_bytecode/macro_frame_diagnostics.rs
```diff
@@ -0,0 +1,290 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Diagnostics describing macro frame info recorded in the source map,
+//! emitted only under the "macro-frames" compiler mode (see
+//! `shared::macro_frames::MACRO_FRAMES_MODE`) and consumed by the
```
cgswords, line 6: nit: Doc says `shared::macro_frames::MACRO_FRAMES_MODE`; the const lives in `shared::macro_expansion`.

external-crates/move/crates/move-compiler/src/to_bytecode/macro_frame_diagnostics.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    /// Formats the frame index plus kind without macro/function names for mismatch messages.
+    fn format_frame_idx_kind(frames: &[MacroFrameInfoEntry], frame_idx: u32) -> String {
+        let entry = &frames[frame_idx as usize];
+        format!("{}:{}", frame_idx, format_kind_short(&entry.kind))
+    }
+
+    /// Build the frame stack for a given expansion index. It's intended
+    /// to mirror frame transitions that would be visible in the debugger.
+    fn build_frame_stack(frames: &[MacroFrameInfoEntry], idx: Option<u32>) -> Vec<u32> {
```
cgswords, line 90: nit: The parent-chain walk will loop forever on a cyclic `parent_index`. Cycles are impossible for freshly built maps, but this function's input is a malformed `SourceMap`. Maybe it's fine to assume the input is well-formed?

external-crates/move/crates/move-compiler/src/to_bytecode/optimize/mod.rs
```diff
@@ -6,15 +6,24 @@ mod remove_nop_store;
...
+    collections::{BTreeSet, HashMap},
+    sync::Arc,
+};
+
+/// Per-block syntactic context maintained in parallel with BytecodeBlocks.
+/// `block_info[i][j]` is the syntactic context (macro expansion chain) for
+/// instruction `j` in block `i`.
+pub(crate) type BlockSyntaxInfo = Vec<Vec<Option<Arc<SyntaxInfo>>>>;
 
 pub type Optimization = fn(
```
cgswords, line 21: nit: The signature is up to five parameters and will presumably grow again the next time optimizations need shared state. Worth considering `OptimizationContext` or something since we are touching them all anyway.

external-crates/move/crates/move-compiler/tests/move_check_testsuite.rs
```diff
@@ -203,6 +228,576 @@ fn out_path(path: &Path, test_name: &str, test_kind: &Option<String>) -> PathBuf
     path.with_file_name(file_name).with_extension(OUT_EXT)
 }
 
+fn render_source_map_snapshot(files: &MappedFiles, units: &[AnnotatedCompiledUnit]) -> String {
```
cgswords, line 231: ~600 lines of bespoke JSON pretty-printing/hash-abbreviation in the testsuite entry file is sort of rough. Could we get `insta` to do this _somewhere_?

external-crates/move/crates/move-compiler/tests/move_check_testsuite.rs
```diff
@@ -203,6 +228,576 @@ fn out_path(path: &Path, test_name: &str, test_kind: &Option<String>) -> PathBuf
...
+        snippet.push_str(" ...");
+    }
+    if snippet.len() > 80 {
+        snippet.truncate(77);
+        snippet.push_str("...");
+    }
+    Some(snippet.replace('`', "\\`"))
+}
+
+fn render_raw_source_map_snapshot(
```
cgswords, line 505: Separate from whether insta can do the rendering: this machinery shouldn't live in the testsuite entry file at all.

external-crates/move/crates/move-compiler/tests/move_check_testsuite.rs
```diff
@@ -275,6 +874,17 @@ pub fn run_test(path: &Path) -> datatest_stable::Result<()> {
         fs::write(out_path, &diag_buffer)?;
     }
 
+    // Macro frame diagnostics render frame/location inconsistencies as `!!`
+    // markers. Fail outright rather than relying on snapshot comparison, so
+    // that a newly added test cannot commit a snapshot containing a
+    // violation unnoticed.
+    if matches!(test_kind, TestKind::MacroFrames) && rendered_diags.contains("!!") {
```
cgswords, line 881: Cool

PR #27417, opened 2026-07-23
Add TransactionPool trait to consensus-core

mwtian:
## Description

Add a TransactionPool interface to consensus-core and route proposer pulls and commit notifications through it. Preserve the current TransactionClient behavior with a TransactionConsumerPool adapter when ConsensusAuthority::start receives None; Sui can pass an external pool in a follow-up change.

## Test plan

- SUI_SKIP_SIMTESTS=1 cargo nextest run --no-capture -p consensus-core -p consensus-simtests -p sui-core (1065 passed, 1 skipped)
- cargo xclippy -p consensus-core -p consensus-simtests -p sui-core -- -D warnings
- cargo fmt --all

### DIFF
--- a/consensus/core/src/transaction.rs
+++ b/consensus/core/src/transaction.rs
@@ -123,7 +123,13 @@ impl TransactionConsumer {
     // and `max_num_transactions_in_block` parameters specified via protocol config.
     // This returns one or more transactions to be included in the block and a callback to acknowledge the inclusion of those transactions.
     // Also returns a `LimitReached` enum to indicate which limit type has been reached.
-    pub(crate) fn next(&mut self) -> (Vec<Transaction>, Box<dyn FnOnce(BlockRef)>, LimitReached) {
+    pub(crate) fn next(
+        &mut self,
+    ) -> (
+        Vec<Transaction>,
+        Box<dyn FnOnce(BlockRef) + Send>,
+        LimitReached,
+    ) {
         let mut transactions = Vec::new();
         let mut acks = Vec::new();
         let mut total_bytes = 0;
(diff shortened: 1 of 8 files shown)


consensus/core/src/transaction.rs
```diff
@@ -451,6 +457,77 @@ impl TransactionClient {
...
+    /// block's reference after the block is durably created, and which limit stopped the
+    /// take. Dropping the callback without invoking it means the transactions have not been
+    /// included in a block, and the implementation may make them available to take again.
+    fn take(
+        &self,
+        max_count: usize,
+        max_bytes: usize,
+    ) -> (
+        Vec<Transaction>,
+        Box<dyn FnOnce(BlockRef) + Send>,
```
aschran, line 476: looks good overall! I have a vague notion, but curious if you can say more about how you imagine the durable creation callback and notify callback would be used on the other side?

guesses:

durable creation callback = safe to remove tx from the pool at this point (whereas before it's called, multiple invocations of `take` would return the same results, assuming the pool itself hadn't changed)

notify committed = I'm not sure exactly, maybe this would replace some stuff currently tracked by the consensus adapter?

mwtian: durable creation callback: this is meant to allow returning `ConsensusPosition` per transaction, to the external transaction submitter. Because BlockRef is not known until the block is created, it has to be passed via a callback after take(). Consensus Core does guarantee that the creation callback runs before next block is created. I don't think we need to guarantee multiple take() calls should return the same results before creation callback. In practice each take() call will be for another block proposal. After take(), I think it is ok to move the taken transactions' states to a "staging" data structure. Taken transactions still need to be tracked until they are finalized or rejected, for retries (if system transaction) and latency measurements.

notify committed: this is mostly relevant to GC'ed transactions. Gc'ed system transactions need to be retried, and this is the only mechanism to support that. GC'ed user transactions should no longer wait for finality.

aschran (approved):

PR #27426, opened 2026-07-23
[gql] Lazy load `Event` node's `timestamp_ms` 2/3

wlmyng:
## Description 

Pending support on grpc to return a read-masked `timestamp_ms`, graphql for now fetches what it can from grpc, and lazily loads `timestamp_ms`

PR stack:
1. https://github.com/MystenLabs/sui/pull/27366
2. this one

## Test plan 

Existing tests pass

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
@@ -503,8 +503,8 @@ impl From<Connection<String, Event>> for EventConnection {
 }
 
 /// Hydrate an `Event` node from a `ListEvents` stream item. The read mask requests everything the
-/// node needs — the event envelope and its position — so no KV lookup is required; a missing
-/// field is an internal inconsistency.
+/// node needs — the event envelope and its position — so no KV lookup is required; a missing field
+/// is an internal inconsistency.
 fn event_from_stream_item(scope: Scope, payload: &v2::Event) -> Result<Event, RpcError> {
     // TODO: can we consolidate to using sui_sdk type? To explore, captured in DVX-2189
     let transaction_digest = payload
--- a/crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
@@ -106,9 +106,9 @@ pub(crate) enum EventTimestamp {
     /// Timestamp known at construction time. `None` when the transaction is not part of a
     /// checkpoint (simulated or just-executed transactions).
     Known(Option<u64>),
-    /// Only the containing checkpoint is known (events hydrated from the gRPC scan stream); its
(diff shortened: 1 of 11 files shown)


crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -151,8 +157,28 @@ impl Event {
     /// All events from the same transaction share the same timestamp.
     ///
     /// `null` for simulated/executed transactions as they are not included in a checkpoint.
-    async fn timestamp(&self) -> Option<Result<DateTime, RpcError>> {
-        Some(DateTime::from_ms(self.timestamp_ms? as i64))
+    async fn timestamp(&self, ctx: &Context<'_>) -> Option<Result<DateTime, RpcError>> {
```
amnn, line 150: Could you take a look at the pattern that we use in e.g. `Epoch`, `Transaction`, `TransactionEffects` etc. for using `OnceCell` to lazy load content, and create a separate path in the KV loader for fetching just the timestamp for a transaction? Fetching the whole transaction is quite a bit more expensive and all the sources we fetch it from also offer a way to fetch just the timestamp (I believe).

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -486,51 +518,34 @@ fn item_position(payload: &v2::Event) -> Result<(TransactionDigest, u32), RpcErr
...
-            Ok((
-                digest,
-                TxEventContents {
-                    events: contents.events().context("Failed to deserialize events")?,
-                    timestamp_ms: contents.timestamp_ms(),
-                },
-            ))
-        })
-        .collect()
+    let native = NativeEvent {
```
amnn: Isn't there an `impl From<sui_sdk_types::Event> for NativeEvent`?

amnn (approved):
The core change looks reasonable, but please do take a look at aligning this with the patterns we use elsewhere for lazy loading for consistency and efficiency reasons (details in inline comments!)

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
wlmyng: there doesn't seem to be one, but i could add an 
```
  bcs_convert_impl!(crate::event::Event, Event)
```

amnn: That seems like a reasonable addition to me?

wlmyng: done!

PR #27433, opened 2026-07-23
sui-execution: collect unified linkage for signing

tzakian:
## Description 

Add unified PTB linkage collection for signing-time analysis.

The execution layer now parses packages directly from `BackingPackageStore` (with intra-transaction caching of packages) and derives the same command-level linkage inputs used during execution. It returns:
 - package original IDs referenced through non-type linkage (i.e., "execution linkage"); and
 - the transaction-wide unified package-version resolution.

The endpoint is gated by unified-linkage and execution version, returning empty results where unsupported.

Signing checks receive `ProtocolConfig` since we need that in a number of areas (or it makes it simpler). If we would prefer not to pass this into the signing checks I can remove that.

Linkage-analysis failures remain execution-time errors and do not reject signing.

## Test plan 

CI + new unit tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


sui-execution/latest/sui-adapter/src/data_store/backing_package_metadata_store.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl<'a> BackingPackageMetadataStore<'a> {
+    pub(crate) fn new(
+        protocol_config: &ProtocolConfig,
+        backing: &'a dyn BackingPackageStore,
+    ) -> Self {
+        Self {
+            backing,
+            binary_config: protocol_config.binary_config(None).clone(),
```
tnowacki, line 39: The `None` is for metering yeah? Do we need that here?

sui-execution/latest/sui-adapter/src/data_store/backing_package_metadata_store.rs
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    pub(crate) fn get_package(
+        &self,
+        package_id: &ObjectID,
+    ) -> SuiResult<Option<Rc<BackingPackageMetadata>>> {
+        if let Some(package) = self.packages.borrow().get(package_id) {
+            return Ok(Some(package.clone()));
+        }
```
tnowacki: Not sure if this whole function can be wrapped up somehow with entry + insert_or_else.
It doesn't look like it given the Rcs and other custom logic for Resutls, but worth considering

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/mod.rs
```diff
@@ -4,18 +4,68 @@
...
+use move_vm_runtime::shared::types::{OriginalId, VersionId};
+use std::collections::{BTreeMap, BTreeSet};
 use sui_protocol_config::ProtocolConfig;
+use sui_types::{
+    error::{ExecutionError, SuiResult},
+    storage::BackingPackageStore,
+    transaction::ProgrammableTransaction,
+};
+
+#[allow(clippy::type_complexity)]
```
tnowacki: What is this warning? Return types?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/mod.rs
```diff
@@ -4,18 +4,68 @@
...
+pub fn collect_unification_information_for_signing(
+    protocol_config: &ProtocolConfig,
+    pt: &ProgrammableTransaction,
+    backing_package_store: &dyn BackingPackageStore,
+) -> SuiResult<(BTreeSet<OriginalId>, BTreeMap<OriginalId, (VersionId, u64)>)> {
```
tnowacki: Might be worth putting this in a struct, because from reading, I have no idea why there isn't just a single map here, lol

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/mod.rs
```diff
@@ -4,18 +4,68 @@
...
+    let linkage_analyzer = LinkageAnalyzer::new::<Normal<ExecutionError>>(protocol_config)
+        .map_err(|error| sui_types::error::SuiError::from(error.to_string()))?;
+
+    let mut execution_original_ids = BTreeSet::new();
+    let linkage = single_linkage::compute_unified_linkage::<ExecutionError, _>(
+        facts,
+        &linkage_analyzer,
+        &backing_package_metadata_store,
+        protocol_config,
+        Some(&mut execution_original_ids),
```
tnowacki, line 49: This isn't just the keys in the other map?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/raw_facts.rs
```diff
@@ -0,0 +1,176 @@
...
+};
+
+use crate::data_store::backing_package_metadata_store::{
+    BackingPackageMetadata, BackingPackageMetadataStore,
+};
+
+use super::facts::{LinkageCommandFacts, LinkageFacts, ModuleInitFacts};
+use sui_verifier::INIT_FN_NAME;
+
+pub(crate) fn linkage_facts_from_programmable_transaction(
```
tnowacki, line 26: I'm a bit confused what this does. A brief comment about why it is here and what is "raw" about it might help

sui-execution/src/lib.rs
```diff
@@ -3,17 +3,31 @@
...
 
 pub mod executor;
 pub mod verifier;
 
+pub type UnificationInformation = (
+    // OriginalId
+    BTreeSet<AccountAddress>,
+    // OriginalId => (VersionId, version)
+    BTreeMap<AccountAddress, (AccountAddress, u64)>,
+);
```
tnowacki: We could make this a struct and then use it in the other file?

sui-execution/src/lib.rs
```diff
@@ -3,17 +3,31 @@
...
 pub use executor::Executor;
 pub use verifier::Verifier;
 
 pub mod executor;
 pub mod verifier;
 
+pub type UnificationInformation = (
+    // OriginalId
+    BTreeSet<AccountAddress>,
+    // OriginalId => (VersionId, version)
```
tnowacki: Why do we need both?

sui-execution/latest/sui-adapter/src/data_store/backing_package_metadata_store.rs
tzakian: It's for overriding deprecation of global storage ops (I think we can actually remove this now actually). This was an interim thing for the rollout of disallowing them at signing IIRC.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/mod.rs
tzakian: yea, return types. I think if I follow your suggestion on making this a struct it should eliminate this.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/mod.rs
tzakian: They are not ;) 

The keys in the other map are all original IDs of all packages that occur in the Lumpy linkage.

This set is the set of original IDs of all packages that could have code executed in them (i.e., don't occur purely in a type argument context).

So this is a subset of the Lumpy keys, but the Lumpy key set may contain original IDs outside of this set.

sui-execution/src/lib.rs
tzakian: Yep. I think this is the way to go

sui-execution/src/lib.rs
tzakian: See comment above, but I will add comments in the struct definition explaining (along with some simple examples)

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/raw_facts.rs
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) Mysten Labs, Inc.
```
cgswords, line 1: This filename though :joy: I think `raw` has the wrong connotation in Rust. Can we pick something better here? `command_facts.rs`?

PR #27435, opened 2026-07-23
Run each simulation in its own process; make simulator test state process-global

mystenmark:
## Description

The deterministic simulator is gaining a blocking-task pool, so a simulation will run code across multiple threads. Per-simulation test state kept in `thread_local!`s (fail-point and debug-fatal registries, framework / advance-epoch / protocol-config injection overrides, the JWK injector, checkpoint-timeout override, fork-stake and synthetic-execution-time accounting, node-leak counter, sim IP/port allocator, RNG seed) is invisible across those threads — a value a test sets on the main thread is not seen on a pool thread. This moves all of it to process globals, which requires at most one simulation per process.

That single-sim-per-process model retires the in-process determinism machinery: `MSIM_TEST_NUM` (many seeds in one process), `MSIM_TEST_CHECK_DETERMINISM` / `#[sim_test(check_determinism)]` (run twice in one process, compare RNG logs), and `init_static_initializers` (which existed only to make that in-process double-run deterministic, and whose own comment notes it has no effect on process-level determinism). Determinism is now checked externally, by running a test twice in separate processes and diffing the output (`scripts/simtest/check-determinism.sh`, run in the `simtest` CI job); multi-seed searches use `seed-search.py` / a seed loop.

Independent of, and landing ahead of, the blocking pool itself: process globals behave identically single-threaded, so this is a no-op for current simulator behavior.

## Test plan

- `scripts/simtest/check-determinism.sh` runs `test_net_determinism` twice in two processes with the same seed and diffs the normalized logs; verified locally it passes (byte-identical sim output). It runs in the `simtest` job on every PR.
- Existing simulator tests continue to exercise the converted state; the `simulator_tests` determinism tests still run (now once each, as ordinary tests).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


lxfind (approved):
Probably also need to update the README file under sui-simulator

mystenmark:
Updated `crates/sui-simulator/README.md` (91cd3ba): documented that `MSIM_TEST_NUM` / `MSIM_TEST_CHECK_DETERMINISM` and the `check_determinism` attribute are retired (per-test state is now process-global, so a process runs one simulation), and pointed the seed-search and determinism-check workflows at `seed-search.py` (one process per seed) and `check-determinism.sh` (run twice in separate processes). Thanks @lxfind.

PR #27437, opened 2026-07-23
[gql] Migrate `checkpoints` node to read from grpc

wlmyng:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
@@ -1,6 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::ops::RangeInclusive;
 use std::sync::Arc;
 
 use anyhow::Context as _;
@@ -11,7 +12,12 @@ use async_graphql::connection::CursorType;
 use async_graphql::connection::Edge;
 use async_graphql::connection::EmptyFields;
 use async_graphql::connection::PageInfo;
+use prost_types::FieldMask;
+use sui_indexer_alt_reader::alpha_ledger_grpc_reader::AlphaLedgerGrpcReader;
+use sui_indexer_alt_reader::alpha_ledger_grpc_reader::StreamPage;
 use sui_indexer_alt_reader::kv_loader::KvLoader;
+use sui_rpc::field::FieldMaskUtil;
+use sui_rpc::proto::sui::rpc::v2;
 use sui_rpc_cursor::CursorKind;
 use sui_rpc_cursor::CursorToken;
 use sui_rpc_cursor::Position;
@@ -354,6 +360,10 @@ impl Checkpoint {
         page: Page<CCheckpoint>,
         filter: CheckpointFilter,
     ) -> Result<CheckpointConnection, RpcError> {
+        if let Some(reader) = ctx.data_opt::<AlphaLedgerGrpcReader>() {
+            return Self::paginate_grpc(ctx, reader, scope, page, filter).await;
+        }
+
         let watermarks: &Arc<Watermarks> = ctx.data()?;
         let available_range_key = AvailableRangeKey {
             type_: "Query".to_string(),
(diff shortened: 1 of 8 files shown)


crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
```diff
@@ -464,18 +613,188 @@ impl PartialEq for CCheckpoint {
...
+                has_next_page: conn.has_next_page,
+                start_cursor,
+                end_cursor,
+            },
+        }
+    }
+}
+
+/// Helper to extract the first and last checkpoint of an epoch.
+async fn epoch_cps(
```
amnn: In the postgres codepath, we could fuse this query into the other filters with a join to avoid a roundtrip. Now that we can't do that, it probably makes sense to use a data loader instead?

amnn (commented):
Needs more data loader (for the `atEpoch` codepath)! Otherwise looking good.

crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
wlmyng: yeah that would be helpful for deduping

wlmyng:
@amnn , given we plan to remove postgres and bigtable soon, that leaves just `LedgerGrpc` on the `KvLoader`. I'm not sure if I'm jumping the gun by adding a `Dataloader`ed grpc reader directly to the graphql schema

crates/sui-indexer-alt-reader/src/epochs.rs
```diff
@@ -146,3 +164,54 @@ impl Loader<EpochEndKey> for PgReader {
             .collect())
     }
 }
+
+#[async_trait::async_trait]
+impl Loader<EpochCheckpointsKey> for LedgerGrpcReader {
```
amnn, line 169: I had a question as I was reading this: If there is no multi-get path for fetching Epochs, is there value in packaging this up in a data loader? After all, we're just delaying issuing the request.

I concluded in this case that it was still worthwhile because there was a good chance Epoch requests would be deduplicated and that was still advantageous, but we could maybe still consider requesting a `BatchGetEpochs` be added.

crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
```diff
@@ -358,6 +401,95 @@ impl Checkpoint {
...
+        let cp_bounds = if let Some(epoch) = filter.at_epoch {
+            let loader: &Arc<DataLoader<LedgerGrpcReader>> = ctx.data()?;
+            match loader
+                .load_one(EpochCheckpointsKey(epoch.into()))
+                .await
+                .context("Failed to get epoch")?
+            {
+                Some(e) => {
+                    intersect_epoch_bounds(e.first_checkpoint, e.last_checkpoint, &cp_bounds)
+                        .context("Epoch's checkpoint range is disjoint from the requested bounds")?
```
amnn, line 452: This seems to be inconsistent with the code path above where we treat a disjoint bound as grounds to return an `Ok(empty())`, rather than an error.

crates/sui-indexer-alt-graphql/src/api/types/checkpoint/mod.rs
```diff
@@ -358,6 +401,95 @@ impl Checkpoint {
...
+        options.before = before;
+        options.ordering = Some(if page.is_from_front() {
+            v2::Ordering::Ascending as i32
+        } else {
+            v2::Ordering::Descending as i32
+        });
+
+        let mut request = v2::ListCheckpointsRequest::default();
+        // Sequence number only — checkpoint contents hydrate lazily via `KvLoader` on field
+        // access.
```
amnn, line 480: If we are only fetching sequence numbers, aren't we doing a network round-trip to calculate, effectively `(cp_bounds.start()..=cp_bounds.end())` -- is there a benefit in doing that?

Similar to the discussion around transactions, this step becomes more useful if we partially hydrate the checkpoint while paginating, but unlike the transactions case, the pagination call is not necessary at all here, so I would lean towards not doing it at all.

PR #27443, opened 2026-07-24
Run transaction execution on blocking threads

mystenmark:
## Description

Transaction execution is synchronous and CPU-bound but ran inline on the async runtime's worker threads, where a long-running execution stalls those threads. Run it via `spawn_blocking` instead.

Execution previously inherited the `epoch_terminated()` barrier for free from `within_alive_epoch` (execution has no `.await` points, so the epoch-alive read guard was held throughout). Moving execution behind a `spawn_blocking().await` introduces a yield point, so the new `AuthorityPerEpochStore::enter_alive_epoch` hands back that guard explicitly; the driver holds it across the unconditionally-awaited blocking task so execution is never detached at epoch end.

## Test plan

Covered by existing `execution_driver` unit tests and the `reconfiguration_tests` simtests (which exercise execution across epoch boundaries); both pass, including simtest determinism.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/execution_driver.rs
+++ b/crates/sui-core/src/execution_driver.rs
@@ -106,42 +106,57 @@ pub async fn execution_process(
 
         authority.metrics.execution_rate_tracker.lock().record();
 
-        // Certificate execution can take significant time, so run it in a separate task.
+        // Certificate execution is CPU-bound and can take significant time, so run it on a
+        // blocking thread to avoid stalling the async runtime's worker threads.
         let epoch_store_clone = epoch_store.clone();
-        spawn_monitored_task!(epoch_store.within_alive_epoch(async move {
+        spawn_monitored_task!(async move {
             let _scope = monitored_scope("ExecutionDriver::task");
-            let _guard = permit;
+            let _permit = permit;
             if authority.is_tx_already_executed(&digest) {
                 return;
             }
 
             fail_point_async!("transaction_execution_delay");
 
-            match authority.try_execute_immediately(
-                &certificate,
-                execution_env,
-                &epoch_store_clone,
-            ) {
-                ExecutionOutput::Success(_) => {
-                    authority
-                        .metrics
-                        .execution_driver_executed_transactions
-                        .inc();
+            // Hold the epoch-alive guard across execution so that `epoch_terminated()` waits
+            // for in-flight execution to finish (previously guaranteed by `within_alive_epoch`,
+            // since execution has no yield points). Skip if the epoch has already ended.
+            let Some(_alive_guard) = epoch_store.enter_alive_epoch().await else {
+                return;
+            };
+
+            // Await unconditionally: once dispatched, execution always runs to completion
+            // within the alive-epoch guard and is never detached at epoch end.
+            tokio::task::spawn_blocking(move || {
+                let _scope = monitored_scope("ExecutionDriver::blocking_task");
+                match authority.try_execute_immediately(
+                    &certificate,
+                    execution_env,
+                    &epoch_store_clone,
+                ) {
+                    ExecutionOutput::Success(_) => {
+                        authority
+                            .metrics
+                            .execution_driver_executed_transactions
+                            .inc();
+                    }
+                    ExecutionOutput::EpochEnded => {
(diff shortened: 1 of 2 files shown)


crates/sui-core/src/execution_driver.rs
```diff
@@ -106,42 +111,57 @@ pub async fn execution_process(
...
+                            .execution_driver_paused_transactions
+                            .inc();
+                    }
                 }
-            }
-        }.instrument(error_span!("execution_driver", tx_digest = ?digest))));
+            })
+            .await
+            .expect("transaction execution task panicked");
+        }.instrument(error_span!("execution_driver", tx_digest = ?digest)));
```
lxfind: `.instrument` only works for async, won't work for `spawn_blocking`

crates/sui-core/src/execution_driver.rs
```diff
@@ -29,8 +29,9 @@ pub async fn execution_process(
 ) {
     info!("Starting pending certificates execution process.");
 
-    // Rate limit concurrent executions to # of cpus.
-    let limit = Arc::new(Semaphore::new(num_cpus::get()));
+    // Rate limit concurrent executions to half of the available CPUs.
+    let execution_concurrency = std::cmp::max(1, num_cpus::get() / 2);
```
lxfind, line 33: I would assume execution still consume most cpu cores?

crates/sui-core/src/execution_driver.rs
```diff
@@ -106,42 +111,57 @@ pub async fn execution_process(
...
-                    // Transaction will be retried later and auto-rescheduled, so we ignore it here
-                    authority
-                        .metrics
-                        .execution_driver_paused_transactions
-                        .inc();
+            // Hold the epoch-alive guard across execution so that `epoch_terminated()` waits
+            // for in-flight execution to finish (previously guaranteed by `within_alive_epoch`,
+            // since execution has no yield points). Skip if the epoch has already ended.
+            let Some(_alive_guard) = epoch_store.enter_alive_epoch().await else {
+                return;
```
aschran, line 135: worth some kind of log here?

crates/sui-core/src/execution_driver.rs
aschran: I also wonder why this particular limit, particularly if we now expect some number execution threads to block and those will presumably count towards the limit? (worst case could this deadlock?)

crates/sui-core/src/execution_driver.rs
mystenmark: yes, this is not quite what i wanted. we need to re-enter the span inside the blocking thread.

crates/sui-core/src/execution_driver.rs
mystenmark: i don't think we had a log for the analogous case under `within_alive_epoch` but happy to add one

crates/sui-core/src/execution_driver.rs
mystenmark: i picked this by looking at the last week's worth of load tests, and couldn't find a time when we had more concurrent execution tasks than cpus/2. Using the prior limit of num_cpus causes CPU contention that did not exist previously (since the size of the async thread pool is capped) which leads to increased execution latency.

crates/sui-core/src/execution_driver.rs
mystenmark: Claude: this should be an info log

lxfind (approved):

crates/sui-core/src/execution_driver.rs
aschran: What about the possibility of deadlock if all threads are blocked waiting on a dependency that's queued to be executed?

mystenmark: that will be an issue when we add blocking (which is not yet present in this PR). Threads will temporarily release their permit when they block and then re-acquire it.

PR #27446, opened 2026-07-24
[source-verification 8] Reuse suiup's cached release tarball when present

mdgeorge4153:
## Description

Before downloading a `sui` release, check whether `suiup` has already cached it (under `<cache-dir>/suiup/releases/sui-<net>-v<version>-<platform>.tgz`) and extract the `sui` binary from that tarball instead. Avoids re-downloading a release the user already has.

Stacked on #27438.

## Test plan

Unit test covers the tarball lookup (mainnet/testnet naming, hit and miss).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: `sui client verify-source` reuses a `suiup`-cached release archive instead of re-downloading, when present.
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-source-verification/src/binary.rs
+++ b/crates/sui-source-verification/src/binary.rs
@@ -184,8 +184,30 @@ fn download_and_install(
 /// Download the `sui` release tarball for `version` and stream it, extracting only the `sui` binary
 /// to `dest`.
 fn stream_sui_binary(version: &str, platform: &str, dest: &Path) -> anyhow::Result<()> {
+    if let Some(tarball) = suiup_cached_tarball(version, platform) {
+        debug!("reusing suiup's cached release at {}", tarball.display());
+        let reader = fs::File::open(&tarball).context("opening suiup's cached release")?;
+        return extract_sui_from_stream(reader, version, platform, dest);
+    }
     let reader = download_reader(version, platform)?;
     extract_sui_from_stream(reader, version, platform, dest)
+}
+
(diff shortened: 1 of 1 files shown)


crates/sui-source-verification/src/binary.rs
```diff
@@ -184,10 +184,32 @@ fn download_and_install(
...
 
+/// The path to a `sui` release tarball already cached by `suiup`, if one exists. `suiup` keeps
+/// release archives under `<cache-dir>/suiup/releases/sui-<net>-v<version>-<platform>.tgz`; reusing
+/// one avoids re-downloading a release the user already has.
+fn suiup_cached_tarball(version: &str, platform: &str) -> Option<PathBuf> {
+    let releases = dirs::cache_dir()?.join("suiup").join("releases");
+    suiup_tarball_in(&releases, version, platform)
+}
+
+/// The path to a matching `suiup` release tarball under `releases` (checking the mainnet and testnet
```
stefan-mysten, line 204: If I recall correctly, the original mechanism was always downloading the mainnet release. Do you want to keep them in-sync?

stefan-mysten (approved):

crates/sui-source-verification/src/binary.rs
mdgeorge4153: I think I deliberately matched the suiup mechanism; I'm pretty sure this matches what exists.

stefan-mysten: I misread, all good.

PR #27451, opened 2026-07-24
[sui][sui-adapter] Restrict TxContext usage in PTB Move call signatures

cgswords:
## Description

The static PTB verifier had no signature-level rules for `TxContext`: duplicate mutable usages were only caught downstream by memory safety, dev-inspect could forge an owned `TxContext` from arbitrary pure bytes, and (under the not-yet-enabled `allow_references_in_ptbs`) a call returning `&mut TxContext` became a usable PTB result.

This adds two rules, checked up-front on the instantiated signature of each Move call and gated by the new `ptb_tx_context_restrictions` feature flag (protocol version 135):

1. In parameters, `TxContext` may appear at most once as `&mut TxContext`, or any number of times as `&TxContext`, and never by value.
2. `TxContext` may never appear in return position, by value or by reference, so it can never become a PTB result. This holds in all execution modes, including dev-inspect.

All violations report a new `CommandArgumentError::InvalidTxContext` (added in last position, emitted only when the feature flag is enabled), with a source message distinguishing the by-value, duplicate-`&mut`, mixed-reference, and return-position cases. The variant landed upstream in MystenLabs/sui-apis#32 and MystenLabs/sui-rust-sdk#297; this PR bumps the pinned `sui-rust-sdk` rev so the proto and sdk-types conversions map it 1:1.

## Test plan

New transactional tests `tx_context_by_value_invalid` and `tx_context_return_invalid` cover by-value parameters (reachable only via dev-inspect), by-value and by-reference returns, and generic instantiations of both; the existing `tx_context_*` tests gain generic-injection cases for the parameter rules, and `tx_context_multiple_borrows_invalid` distinguishes the duplicate-`&mut` error (reported at the second occurrence) from the mixed `&mut`/`&` error (reported at the `&mut`), including when both apply. Note that chained shapes like `mut_id(); use(Result(0))` still fail with `ArityMismatch` because arity is checked during translation, before the verify passes; the snapshots document this ordering.

---

## Release notes

- [x] Protocol: New feature flag `ptb_tx_context_restrictions` (protocol version 135). `TxContext` in a PTB Move call signature may appear at most once as `&mut TxContext` or any number of times as `&TxContext`, never by value, and never in return position. Transactions calling functions that violate these rules now fail up front with the new `InvalidTxContext` command argument error (previously they failed with borrow-checking, arity, or serialization errors, or in dev-inspect could partially succeed).
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
```diff
@@ -240,6 +240,14 @@ impl Type {
             TxContextKind::None
         }
     }
+
+    /// Is this the `TxContext` datatype itself, not behind a reference?
+    pub fn is_tx_context_by_value(&self) -> bool {
+        match self {
+            Type::Datatype(dt) => dt.qualified_ident() == RESOLVED_TX_CONTEXT,
+            _ => false,
+        }
```
tnowacki: `matches!` could be nice, but is fine as is

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/move_functions.rs
```diff
@@ -95,6 +100,73 @@ fn check_signature<Mode: ExecutionMode>(
...
+                        ExecutionErrorKind::command_argument_error(
+                            CommandArgumentError::TypeMismatch,
+                            checked_as!(idx, u16)?,
+                        ),
+                        "TxContext cannot be taken by value",
+                    ));
+                }
+            }
+        }
+    }
```
tnowacki: I don't mind this, but I would consider just split this up into 3 distinct passes for clarity. 
1. Check no by value usage (via `find`)
2. Then for reference usage:
    - First Determine mutable usage (via `find`)
    - Then count all usages

tnowacki (commented):
Overall looks fine, but if we can, we might want to just replace these errors with InvalidTxContext usage. In the summary it is stated that these changes preserve previous errors, but that isn't entirely true since the ordering of things has changed. 
As such, we might want to consider providing a better error message if it is convenient. 

It might also be good to add compiler warnings for these (if we don't already)

sui-execution/latest/sui-adapter/src/static_programmable_transactions/typing/verify/move_functions.rs
cgswords: Done as three different predicates

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
cgswords: Done

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
```diff
@@ -240,6 +240,11 @@ impl Type {
             TxContextKind::None
         }
     }
+
+    /// Is this the `TxContext` datatype itself, not behind a reference?
+    pub fn is_tx_context_by_value(&self) -> bool {
+        matches!(self, Type::Datatype(dt) if dt.qualified_ident() == RESOLVED_TX_CONTEXT)
+    }
```
tnowacki, line 247:
```suggestion
    }
    
```

tnowacki: nit

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4587,6 +4597,8 @@ impl ProtocolConfig {
 
                     cfg.consensus_max_transactions_in_block_bytes = Some(288 * 1024);
                     cfg.consensus_max_num_transactions_in_block = Some(128);
+
+                    cfg.feature_flags.ptb_tx_context_restrictions = true;
```
tnowacki: I believe 1.78 is on 134, so you will need to bump this

tnowacki (approved):

PR #27453, opened 2026-07-24
[move-model] Fix stackless bytecode from-bytecode tests

awelc:
## Description 

Prior to this PR `from_bytecode` tests for stackless bytecode were effectively dead. They were looking for `.mv` files of which there were none, accepting the empty file set, and marking the whole test run as successful.

Apparently, the  snapshots for these tests were previously committed to the repository (but NOT the`.mv` files), and were deleted during one of the cleanups.  As a result, this PR was created under the assumption that these tests were useful but perhaps we did not want to keep `.mv` files in the repository.

To address this, test harness for `from_bytecode` now builds source packages to generate bytecode and runs analysis on said bytecode rather than looking for pre-built `.mv` files.

Incidentally, this also allow testing the same cases we have in `from_source` at the bytecode level simply by putting the `from_bytecode.txt` files in the relevant package directories. We can decide what makes sense to be tested this way in a follow-up PR

## Test plan 

Proper bytecode testing was enabled and new snapshots for the existing fixture were generated. All tests must pass


external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
```diff
@@ -1,85 +1,136 @@
...
+    let mut writer = Vec::new();
+    let compiled_package = BuildPlan::create(&root_pkg, &config)?
+        .compile_no_exit(&mut writer, |compiler| compiler)
+        .with_context(|| {
+            format!(
+                "failed to compile {}:\n{}",
+                pkg_dir.display(),
+                String::from_utf8_lossy(&writer)
+            )
+        })?;
```
cgswords: This package-loading + compile is now a near-copy of `from_source.rs`, with only small differences. Can we pull the shared setup into a helper?

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
```diff
@@ -1,85 +1,136 @@
...
-
-    let (_mdl, bytecode) = from_compiled_modules(modules.clone(), /* optimize */ true)?;
-
     let test_module_names = test_module_names
         .into_iter()
         .map(|name| name.into())
         .collect::<BTreeSet<Symbol>>();
+    if test_module_names.is_empty() {
+        return Err(anyhow::anyhow!("no modules requested by {}", file_path.display()).into());
+    }
```
cgswords: Blank lines in `from_bytecode.txt` become empty `Symbol`s in `test_module_names`. Won't the `ensure!` at the bottom then fail, and the new `is_empty()` guard? A `.filter(|l| !l.trim().is_empty())` maybe?

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
```diff
@@ -1,85 +1,136 @@
...
+    root_modules: &BTreeSet<ModuleKey>,
+    test_module_names: &BTreeSet<Symbol>,
+    suffix: &str,
+) -> anyhow::Result<()> {
+    let mut observed_modules = BTreeSet::new();
 
     for pkg in &bytecode.packages {
         let pkg_name = pkg
             .name
             .unwrap_or(Symbol::from(pkg.address.to_hex_literal()));
```
cgswords: `pkg.name` is always `None` for packages coming out of `from_compiled_modules`; since you have `compiled_package` in hand, consider using the root package name for consistent snapshot naming.

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
```diff
@@ -1,85 +1,136 @@
...
         for (module_name, module) in &pkg.modules {
-            if test_module_names.contains(module_name) {
+            let module_key = (pkg.address, *module_name);
+            if root_modules.contains(&module_key) && test_module_names.contains(module_name) {
```
cgswords: nit: `// Filter to avoid emitting stdlib, etc`

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
```diff
@@ -1,85 +1,136 @@
...
 
+    ensure!(
+        &observed_modules == test_module_names,
+        "requested modules were not translated for {suffix}: {}",
+        test_module_names
+            .difference(&observed_modules)
+            .map(Symbol::as_str)
+            .collect::<Vec<_>>()
+            .join(", ")
+    );
```
cgswords: Nice!

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
```diff
@@ -1,85 +1,136 @@
...
+use move_package_alt_compilation::{build_config::BuildConfig, build_plan::BuildPlan};
 use move_symbol_pool::Symbol;
 use std::{collections::BTreeSet, io::BufRead, path::Path};
+use tempfile::TempDir;
 
-use move_binary_format::CompiledModule;
-use move_command_line_common::files::{MOVE_COMPILED_EXTENSION, extension_equals, find_filenames};
+type ModuleKey = (AccountAddress, Symbol);
 
 fn lib_test(file_path: &Path) -> datatest_stable::Result<()> {
```
cgswords: nit: `from_bytecode` now compiles from source, so it should probably get renamed

cgswords (approved):
LGTM for dead code elimination, but `from_bytecode` should really be used to exercise on-disk bytecode at some point as a follow-up. Please fix the other stuff before landing, and tag me in a re-review if necessary.

One more that I can't inline since the file isn't in the diff: should `from_source` get the same requested-module-must-exist validation added here? Maybe as part of unification?

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
awelc: Renamed to `from_compiled` but happy to accept other suggestions

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
awelc: Good idea! Done!

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
awelc: Good catch! Fixed to make sure that blank lines are ignored as susggested

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
awelc: Good idea! Done!

external-crates/move/crates/move-stackless-bytecode-2/tests/from_bytecode.rs
awelc: Added a comment about excluding dependencies but this code has been refactored a bit and looks a bit different now

awelc:
> LGTM for dead code elimination, but `from_bytecode` should really be used to exercise on-disk bytecode at some point as a follow-up. Please fix the other stuff before landing, and tag me in a re-review if necessary.

I can add this functionality in a separate PR, particularly if you have some suggestions on what would be worth testing in this mode (I can add some initial tests myself as part of this other PR of course and we can take it from there)

> One more that I can't inline since the file isn't in the diff: should `from_source` get the same requested-module-must-exist validation added here? Maybe as part of unification?

Done!

cgswords (approved):

PR #27459, opened 2026-07-25
Check object funds withdraw sufficiency during execution

lxfind:
## Description

Second half of the restructured stack, on top of #27403: moves the object-funds withdraw sufficiency check into the Move VM, superseding the post-execution `ObjectFundsChecker` wait/re-execute flow. Gated by `check_object_funds_withdraw_in_execution` (protocol v132, devnet only).

`funds_accumulator::withdraw_from_object` now calls a `reserve_object_funds_for_withdrawal` native that keeps a per-`(owner, type)` running balance in the object runtime. When the in-transaction balance falls short, it reads the settled balance at the transaction's accumulator version through the temporary store — via the blocking `load_implicitly_read_system_object` from #27403, so a root that has not caught up locally simply waits instead of unwinding — and discounts unsettled withdrawals from earlier transactions in the same consensus commit, read from the standalone `UnsettledObjectWithdrawals` store (extracted out of the old checker in the first commit; now populated on fullnodes too, since the check affects execution results on every executing node). Insufficiency is an ordinary deterministic Move abort, surfaced in effects as `MoveAbort(funds_accumulator, E_OBJECT_FUNDS_INSUFFICIENT)` like any other framework abort — deliberately not remapped to a dedicated `ExecutionErrorKind`. Rust consumers (the authority metric, benchmark, tests) match it via `sui_types::funds_accumulator::is_object_funds_insufficient_abort`.

This reworks the old [6/7] #27335 + [7/7] content onto the block-wait base: the retry-based availability plumbing the old version carried (the `SYSTEM_OBJECT_NOT_AVAILABLE_LOCALLY` unwind, the retry request on the temporary store, and the non-`Available` availability variants) is gone entirely — unavailability is handled by the blocking read and never surfaces in execution results.

Known caveat carried over from the old PR: isolated replay does not reconstruct in-commit unsettled withdrawals (`unsettled = 0`; see the TODO in `temporary_store.rs`), which can diverge from the original execution. Mainnet enablement is blocked on resolving that.

## Test plan

- The `object_funds_withdraw.move` transactional test covers the in-VM check at the Move level, in both modes. Dry-run tasks exercise the read-time-`Latest` path (withdrawals within the settled balance succeed, exceeding it aborts, in-transaction deposits cover later withdrawals without consulting the store). Committed tasks exercise the real certificate path end to end, including unsettled discounting: a second withdrawal in the same settlement window aborts because the first one's unsettled amount is deducted from the settled balance, and succeeds again once a checkpoint settles it. (The committed path relies on the base PR's runner change that assigns consensus versions to every committed transaction.)
- sui-core unit and integration tests cover the `UnsettledObjectWithdrawals` store and its interaction with execution: net unsettled amounts recorded from effects, GC at checkpoint commit, multiple withdrawals against one account, withdraw-and-deposit in the same transaction, and zero-amount withdrawals.
- E2e tests in `address_balance_tests.rs` cover the flag-on validator path (sufficient and insufficient withdrawals, the accumulator root recorded as a read-only consensus object in effects), effects equivalence with the flag off, and the dry-run/simulate counterparts. These are simtests and are blocked from running by the known msim `block_in_place` compile issue inherited from #27403; they pass once the simulator fix lands.

## Release notes

- [x] Protocol: Adds `check_object_funds_withdraw_in_execution` (protocol v132, enabled on devnet only): object-funds withdraw sufficiency is checked inside the Move VM at the point of withdrawal, and insufficient withdrawals abort the transaction with a `sui::funds_accumulator` Move abort instead of being cancelled before execution.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


lxfind:
Superseded by #27524 (identical content, clean history).

PR #27478, opened 2026-07-27
[gql] Split out and share scanning grpc logic at graphql boundary 1/n

wlmyng:
## Description 

Split out graphql-side scanning grpc usage from scope check and Connection building, so that `fn scan_grpc` can be reused by subscriptions on backfilling (catching up to streaming.)

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::ops::RangeInclusive;
+use std::ops::Range;
 use std::sync::Arc;
 
 use anyhow::Context as _;
@@ -439,17 +439,19 @@ impl Transaction {
             return Ok(Connection::new(false, false).into());
         };
 
+        // `cp_bounds` end is inclusive; `scan_grpc` takes an exclusive end.
+        let cp_bounds = *cp_bounds.start()..cp_bounds.end().saturating_add(1);
         let result = Self::scan_grpc(reader, cp_bounds, &page, &filter).await?;
 
         build_grpc_connection(scope, &page, result)
     }
 
-    /// Scan a page of transactions over `cp_bounds` via the streaming gRPC List API. Computing
-    /// checkpoint bounds is the caller's responsibility. Items are returned in scan order
-    /// (descending when paginating from the back).
+    /// Scan a page of transactions over the half-open checkpoint range `cp_bounds` via the
+    /// streaming gRPC List API. Computing checkpoint bounds is the caller's responsibility.
+    /// Items are returned in scan order (descending when paginating from the back).
     pub(crate) async fn scan_grpc(
         reader: &AlphaLedgerGrpcReader,
-        cp_bounds: RangeInclusive<u64>,
+        cp_bounds: Range<u64>,
         page: &Page<CTransaction>,
         filter: &TransactionFilter,
     ) -> Result<StreamPage<v2::ExecutedTransaction>, RpcError> {
@@ -478,9 +480,8 @@ impl Transaction {
         let mut request = v2::ListTransactionsRequest::default();
         // Digest only — contents hydrate lazily via `KvLoader` on field access.
         request.read_mask = Some(FieldMask::from_paths(["digest"]));
-        request.start_checkpoint = Some(*cp_bounds.start());
-        // `cp_bounds` end is inclusive; the request bound is exclusive.
-        request.end_checkpoint = Some(cp_bounds.end().saturating_add(1));
+        request.start_checkpoint = Some(cp_bounds.start);
+        request.end_checkpoint = Some(cp_bounds.end);
         request.filter = filter.to_grpc_filter();
         request.options = Some(options);
 
--- a/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
@@ -1,7 +1,8 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::ops::Range;
+use std::ops::Bound;
+use std::ops::RangeBounds;
 use std::sync::Arc;
 
 use anyhow::Context as _;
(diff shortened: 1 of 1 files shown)


crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -463,18 +480,15 @@ impl Transaction {
         let mut request = v2::ListTransactionsRequest::default();
         // Digest only — contents hydrate lazily via `KvLoader` on field access.
         request.read_mask = Some(FieldMask::from_paths(["digest"]));
-        request.start_checkpoint = Some(*cp_bounds.start());
-        // `cp_bounds` end is inclusive; the request bound is exclusive.
-        request.end_checkpoint = Some(cp_bounds.end().saturating_add(1));
+        request.start_checkpoint = Some(cp_bounds.start);
+        request.end_checkpoint = Some(cp_bounds.end);
```
tpham-mysten: Could `scan_grpc` accept an optional upper bound, taking `impl RangeBounds<u64>` and mapping an unbounded end to `None`?

```
    request.end_checkpoint = match cp_bounds.end_bound() {
        Bound::Included(&e) => Some(e + 1),
        Bound::Excluded(&e) => Some(e),
        Bound::Unbounded => None,
    };
```

The reason I ask is that we don't know up front how far a scan will reach, so the subscription can't pick an upper bound ahead of time: it scans forward and only pins the handoff once it sees it has caught up close to the tip. That holds whether the client is far behind or nearly current. An unset end fits that, scan forward to whatever's indexed, which is also how the direct `list_transactions` path worked. The proto already makes `end_checkpoint` optional, so it's just surfacing that.

wlmyng: i see, i was anchored on the handoff checkpoint, but that only comes in play when we're within the hand off threshold. Yeah that's reasonable

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -463,18 +480,15 @@ impl Transaction {
         let mut request = v2::ListTransactionsRequest::default();
         // Digest only — contents hydrate lazily via `KvLoader` on field access.
         request.read_mask = Some(FieldMask::from_paths(["digest"]));
```
tpham-mysten, line 482: I think this is the "Prefetching content during pagination" topic @amnn has on the weekly agenda, so not sure if we want to support prefetching here? The read mask is hardcoded to
digest-only today. Options for how far to go:

* Digest only (current)
  - Pros: smallest scan, no over-fetch
  - Cons: every content/object field is a lazy KvLoader round trip (but it's batched)

*  Contents, not objects (objects lazy)
    - Pros: contents come with the scan (fewer round trips); objects stay lazy, matching "read contents alongside digests, but not the object set"
    - Cons: over-fetches contents for a digest-only query

* Full proto (contents + objects)
  - Pros: trivial hydration (build the Transaction straight from the proto, no lazy reads); same shape live already returns
  - Cons: biggest over-fetch, and since this path is shared, Query pagination would over-fetch too

* Selection-aware read mask
  - fetch exactly what the query selects, no over-fetch and no lazy reads
  - the most work


I personally lean toward full proto: hydration stays trivial (build straight from the proto, no lazy reads) and it cuts round trips, which is particularly useful for deep backfill. It does over-fetch, and Query pays it too on this shared path, but I think that's a fair trade for the throughput gained.

WDYAT?

wlmyng: I think the challenge is that the subscription side wants objects, not for any graphql field, but because of `Scope::for_scanned_transactions`, otherwise it could be the same fields masked, "transaction, effects, events.bcs, signatures.bcs, balance_changes, checkpoint, timestamp"

wlmyng: can subscriptions side support full contents but not objects?

tpham-mysten: Yeah, contents-not-objects works. But I'd frame objects as an object cache rather than just scope plumbing: it lets objectChanges / asMoveObject resolve from memory instead of a kv read. Dropping it means backfill hits the kv store for object fields while live doesn't (live already serves them from the streamed checkpoint's execution_objects map). Prefetching keeps those consistent and saves the kv reads, at the cost of over-fetching when the query doesn't read objectChanges.

So I lean toward full proto to keep backfill and live consistent on object reads. The round trips saved when objects are read are worth the over-fetch when they aren't, imo.

wdyt?

wlmyng: > it lets objectChanges / asMoveObject resolve from memory instead of a kv read. Dropping it means backfill hits the kv store for object fields while live doesn't (live already serves them from the streamed checkpoint's execution_objects map).

But this is because when streaming live, the object set hasn't been indexed, so if we don't keep the data in memory we would not be able to yield object data at all. This is done on subscription and execution in order to enable read functionality that is otherwise impossible 

Otherwise, it's not an issue to read from the kv store for indexed checkpoints

tpham-mysten: To avoid conflating two paths: live doesn't call scan_grpc, it serves objects from the streamed checkpoint's in-memory map out of necessity (live data isn't indexed yet). Scanning is the only caller of scan_grpc, and it's over indexed data. So this is purely a scanning question, live isn't relevant here.

For scanning, agreed kv reads are fine correctness-wise. My angle is throughput: since scanning goes through LedgerService, which can return objects in the same response, prefetching them saves the separate kv round trips per page vs reading lazily. The catch is over-fetch when the query doesn't read objectChanges (mostly Query pagination on this shared path, subscriptions usually do read them), so selection-aware might be the clean answer. But for scanning, folding objects into the scan saves round trips.

WDYT?

tpham-mysten (approved):
Thanks @wlmyng, this is a clean split and the subscription scanning path can adopt it directly. We can continue discussing prefetching here, but the code can land on a follow-up PR.

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
amnn: I would opt for contents without object set here. Even for a query that actually needs nested objects on a transaction, because all those payloads are paginated, it's mostly impossible to fully take advantage of a transaction's object set. The same is not true for contents -- most queries that paginate transactions will definitely need to load the contents.

wlmyng: > Dropping [objects] means backfill hits the kv store for object fields while live doesn't (live already serves them from the streamed checkpoint's execution_objects map). Prefetching keeps those consistent and saves the kv reads, at the cost of over-fetching when the query doesn't read objectChanges.

The point I was trying to make is that we don't need to do this when backfilling a subscription, because we are reading indexed data. Aligned on lazy-loading is fine correctness-wise. Then that leaves throughput, but this is no different from prefetching/ lazy loading on the Query side

tpham-mysten: > Then that leaves throughput, but this is no different from prefetching/lazy loading on the Query side

Yeah the throughput will matter more in Subscription backfilling. Lazy loading costs about one extra DB round trip per batch (~10ms for the hop plus a few ms of CPU, call it ~15ms). At 100 per batch and a dense filter (roughly a match per checkpoint), backfilling all ~300M checkpoints is ~3M batches, so ~45,000s (~12 hours) of added latency. For a normal near-tip backfill it's negligible; only the deep-from-genesis case makes it add up.

So I think contents without the object set, as @amnn suggested, is the right balance: it folds the contents read into the scan where it matters most, without over-fetching an object set that does not benefit pagination well.

amnn: > At 100 per batch and a dense filter (roughly a match per checkpoint), backfilling all ~300M checkpoints is ~3M batches, so ~45,000s (~12 hours) of added latency.

This calculation assumes the latency is applied serially which should not be the  case -- while we are doing that second round trip to inflate objects, we can also be fetching the next batch of matching transactions, which means the latency is only ~15ms even across a full backfill from genesis.

tpham-mysten: Thanks for catching this, agreed. With concurrency buffer = 100 and p99 ~49 transactions per checkpoint (by Capy), a checkpoint's transactions don't fill the buffer, so the next checkpoint is already being scanned while the current DB read is in flight. The inflate round trip overlaps the next scan rather than adding to it, so the extra latency across a full backfill is ~15ms, not the per-batch figure I quoted.

PR #27479, opened 2026-07-27
Move self-signed effects equivocation check from execution path to RPC boundaries

alex-mysten:
## Description

A validator must never assert two different effects for the same transaction within an epoch, since clients aggregate effects signatures toward finality. This was enforced during execution: if re-executing an uncommitted transaction (e.g. after a restart with a new binary) produced effects differing from previously signed effects, the validator would panic. 

This PR limits the blast radius of this equivocation check in two ways:
1. a validator will refuse to broadcast effects different from those for which it has already returned a signature
2. we will not panic if we detect that effects are different, we will simply refuse to return different effects from rpc

We will also not return unsigned effects via RPC if those effects are already in the signed_effects table. Note that once the checkpoint is finalized, the record is cleared from the signed_effects table, so the check only applies between transaction execution and checkpoint finalization. 

## Test plan

New unit tests cover: divergent re-execution no longer fails execution; signing, `SubmitTransaction`, and `WaitForEffects` each refuse to contradict a previously signed digest; a matching signed digest does not block responses; conflicting digest records are rejected atomically. 

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/authority.rs
+++ b/crates/sui-core/src/authority.rs
@@ -297,6 +297,9 @@ pub struct AuthorityMetrics {
     authority_state_handle_vote_transaction_latency: Histogram,
 
     internal_execution_latency: Histogram,
+    /// Number of times the validator refused to report effects (signed or unsigned, labeled by
+    /// RPC surface) because it had previously signed different effects for the same transaction.
+    signed_effects_equivocation_prevented: IntCounterVec,
     execution_load_input_objects_latency: Histogram,
     prepare_certificate_latency: Histogram,
     commit_certificate_latency: Histogram,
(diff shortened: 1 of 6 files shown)


crates/sui-core/src/authority.rs
```diff
@@ -5469,13 +5467,61 @@ impl AuthorityState {
...
         let tx_digest = *effects.transaction_digest();
+
+        self.check_effects_against_previously_signed(
+            epoch_store,
+            &tx_digest,
+            &effects.digest(),
+            "sign_effects",
+        )?;
+
         let signed_effects = match epoch_store.get_effects_signature(&tx_digest)? {
```
mystenmark, line 5525: we could further simplify this to stop creating effects signatures (since we don't return those any more) and just verify that the digest matches the previously attested digest if there is one

alex-mysten: `transaction_info` rpc is the only upstream of `sign_effects`, and it still returns signed effects. if we didn't return signed effects here this whole problem goes away, no?

mystenmark: well, no, the problem will exist as long as we attest to effects in any way, whether by signatures or not. If we are still returning signed effects then we should probably just leave things as they are (although we could sign lazily, only when needed by transaction_info)

mystenmark (approved):

crates/sui-core/src/authority.rs
alex-mysten: we only sign effects when `transaction_info` is called and the tx is locally executed, and we always sign them on that path. we could detect any effects divergence, signed or unsigned, in a followup

PR #27399, opened 2026-07-22
Add GCP Confidential Spaces attestation native function

chargarlic:
## Description

Rebased rewrite of #25671 onto current `main` (protocol MAX 132 → **133**).

Adds on-chain verification of GCP Confidential Spaces attestation JWTs so Move contracts can verify code is running in a GCP Confidential Space (AMD SEV-SNP) and inspect workload claims.

### Design
- **Move API (Nitro-like):** `sui::gcp_attestation::verify_gcp_attestation(token, kid, clock)` — **no** `AuthenticatorState` arg
- **Trust (zkLogin-like):** GCP JWKS → consensus `NewJWKFetched` → `AuthenticatorState` → per-tx `JwkMap` native extension
- **Crypto:** `ring` RS256; numeric `e ≥ 65537`; `kid` bound to JWT header; issuer hardcoded to `https://confidentialcomputing.googleapis.com`
- **Node:** validator background GCP JWKS poller (+ msim injector)
- **Protocol:** `enable_gcp_attestation` at **v133**, **Unknown/devnet only** (Testnet/Mainnet off); gas on all chains

Closes the earlier AuthState-as-user-input approach (tzakian) — AuthState is not taken as a shared object in user txs.

Supersedes #25671 and #27398.

### Test plan
- [x] `SUI_SKIP_SIMTESTS=1 cargo test -p sui-types --lib -- gcp_attestation`
- [x] Move unit tests for `gcp_attestation`
- [x] `cargo simtest -p sui-e2e-tests --test gcp_attestation_tests` (3/3)
- [x] protocol + swarm + CLI summary snaps for v133
- [x] Full CI on this PR

```
SUI_SKIP_SIMTESTS=1 cargo test -p sui-types --lib -- gcp_attestation
cargo simtest -p sui-e2e-tests --test gcp_attestation_tests
cargo test -p sui-protocol-config snapshot_tests
```

---

## Release notes

- [x] Protocol: new native `verify_gcp_attestation_internal` in `sui::gcp_attestation`; `enable_gcp_attestation` at protocol version 133 (non-Mainnet/Testnet only)
- [x] Nodes (Validators): background task polls GCP JWKS and submits keys to consensus (same pipeline as zkLogin JWKs)

PR #27481, opened 2026-07-27
Add and harden GCP Confidential Spaces attestation

chargarlic:
## Description

Adds on-chain verification of GCP Confidential Spaces attestation JWTs, including the complete API, consensus hardening, and node updater implementation previously split across #27482, #27483, and #27484.

### Design

- **Kidless Move API:** `sui::gcp_attestation::verify_gcp_attestation(token, clock)` derives the required `kid` from the signed JWT header.
- **Consensus-backed trust:** validators fetch GCP JWKS, submit them through consensus, and execution reads a consensus-versioned `AuthenticatorState` snapshot.
- **Fixed cryptography:** exact GCP issuer, RS256 only, bounded RSA key parameters, exact signed-byte verification, and no remote-key URL or algorithm fallback.
- **Fail-closed key lifecycle:** malformed, stale, over-cap, or conflicting same-`kid` keys cannot be used; live consensus and replay apply identical protocol-gated validation.
- **Shared updater:** zkLogin and GCP reuse the polling/retry/dedup/submission loop while retaining source-specific validation and cap semantics.
- **Protocol rollout:** base attestation support at v133; consensus validation, activation limits, and one-epoch key age at v134. Testnet/Mainnet activation remains off.

The Move verifier authenticates the token but does not authorize a workload. Callers must enforce audience and workload claims. Nonce consumption remains application policy.

Supersedes #25671, #27398, #27399, #27482, #27483, and #27484.

## Test plan

- [x] GCP JWT parser, signature, issuer, time, `kid`, and RSA-bound tests
- [x] framework Move/build-system package tests
- [x] consensus live/replay validation and activation conflict/cap tests
- [x] stale/conflicting `JwkMap` tests
- [x] GCP and zkLogin updater regression tests
- [x] GCP msim E2E tests
- [x] full workspace and cfg(msim) checks
- [x] real Confidential Space JWT localnet deployment and continuous on-chain updates
- [ ] Full CI on the combined PR head

---

## Release notes

- [x] Protocol: add GCP Confidential Spaces JWT verification at v133 and protocol-gated JWK validation, conflict handling, activation limits, and bounded key age at v134
- [x] Nodes (Validators and Full nodes): validators fetch and consensus-validate GCP JWKs; execution consumes consensus-versioned keys through shared zkLogin/GCP updater infrastructure


crates/sui-protocol-config/src/lib.rs
```diff
@@ -29,7 +29,7 @@ use tracing::{info, warn};
 
 /// The minimum and maximum protocol versions supported by this build.
 const MIN_PROTOCOL_VERSION: u64 = 1;
-const MAX_PROTOCOL_VERSION: u64 = 132;
+const MAX_PROTOCOL_VERSION: u64 = 134;
```
lxfind, line 32: **Redefines frozen protocol versions 133 and 134** -- Confidence: 100/100 | Importance: 95/100

(found by: Claude, Codex)
- **File**: `crates/sui-protocol-config/src/lib.rs` lines 32, 4597-4619; new `crates/sui-protocol-config/src/snapshots/*_version_133.snap` and `*_version_134.snap`; `crates/sui-core` `execution_time_observer_v133/v134.snap`; `crates/sui-benchmark/tests/simtest.rs` lines 842-847; `crates/sui-e2e-tests/tests/address_balance_tests.rs` lines 107-110 and 194-197
- **Problem**: The branch is based on a July 29 merge-base where `MAX_PROTOCOL_VERSION` was 132, and it defines version 133 (gas costs plus `enable_gcp_attestation`) and 134 (`max_gcp_active_jwks`, `max_age_of_gcp_jwk_in_epochs`, `enable_gcp_consensus_validation`). On current main, 133 is "include function signatures in type-node limits" and 134 is "add package::original_package_id". The latest release branch is at 136 and main is at 136 or 137 (the two reviewers differed by one). Main already ships snapshot files for 133 through 136 with different contents and framework bytecode snapshots for 132, 135 and 136. The `simtest.rs` and `address_balance_tests.rs` changes that disable `create_forwarding_address_registry` exist only because the branch created intermediate versions without framework snapshots; main does not need them.
- **Impact**: The same protocol version would carry different semantics in different binaries. That breaks the released-versions-are-frozen invariant and risks consensus divergence, incompatible replay, and snapshot corruption. The two-version split also forces the `max_age_of_gcp_jwk_in_epochs_as_option().unwrap_or(u64::MAX)` fallback in `context.rs`, which only exists to survive v133.
- **Suggestion**: Rebase onto main first. Introduce the GCP changes at a fresh version above the current maximum, ideally a single version holding the gas costs, cap, age, and both feature flags gated to `Chain::Unknown`. Regenerate protocol-config, framework, execution-estimator, OpenRPC, and swarm snapshots. Drop the v133/v134 snapshot files, the forwarding-address test workarounds, and the `u64::MAX` fallback. Run `/protocol-config` as CLAUDE.md requires.

---

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs
```diff
@@ -415,13 +416,86 @@ where
...
+            }
+
+            let store = StateAsObjectStore {
+                state: &*env.state_view,
+                authenticator_state,
+                assigned_version,
+            };
+            let inner = get_authenticator_state(&store)
+                .map_err(|error| {
+                    Mode::Error::from(make_invariant_violation!(
```
lxfind, line 476: **AuthenticatorState inner child is read at the outer object's exact version** -- Confidence: 85/100 | Importance: 85/100

(found by: Claude)
- **File**: `sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/context.rs` lines 442-476 (line 456 in particular); `sui-execution/latest/sui-adapter/src/temporary_store.rs` lines 1142-1150
- **Problem**: `StateAsObjectStore::get_object` resolves the `AuthenticatorStateInner` dynamic-field child at the exact version assigned to the outer `0x7` object. The two do not always share a version. Every end-of-epoch transaction includes `AuthenticatorStateExpire`, and since the outer object is a mutable input, `ensure_active_inputs_mutated` bumps it unconditionally. The child is only written when its bytes change (`object_store.rs:787`, `object_runtime/mod.rs:840`). In steady state `expire_jwks` removes nothing, so the rebuilt `active_jwks` is byte-identical and the child keeps its older version. The same lag occurs after an `AuthenticatorStateUpdate` whose only new key is a same-kid rotation that the Move code ignores.
- **Impact**: From the start of every quiet epoch until the first update that really changes the inner list, every transaction calling `verify_gcp_attestation` is assigned the post-epoch version, finds the outer object but not the child, and `get_authenticator_state` fails. That failure is routed through `make_invariant_violation!`, which panics under `cfg!(debug_assertions)` and commits a failed `InvariantViolation` effect in release builds. No e2e test crosses an epoch boundary, so this is uncovered.
- **Suggestion**: Read the child the way the Move runtime does, with an upper bound rather than an exact version. Use the backing store's `RuntimeObjectResolver::read_child_object(&SUI_AUTHENTICATOR_STATE_OBJECT_ID, &child_id, assigned_version)` to get the latest child version at or below the outer's assigned version. That also removes the need for `Storage::get_object_by_key_including_store` and the inline `StateAsObjectStore` adapter. Add a simtest that activates a GCP key, crosses an epoch boundary with no JWK changes, and verifies a token before any new update lands.

---

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -151,9 +165,43 @@ impl<'backing> TemporaryStore<'backing> {
...
+    /// Ensure an implicitly read system object is available at its consensus-assigned version.
+    pub fn check_system_object_available(&self, object_id: &ObjectID) -> PartialVMResult<()> {
+        let Some(required_version) = self.system_object_versions.get(object_id).copied() else {
+            debug_fatal!("system object {object_id} read without an assigned version");
+            return Err(
+                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR).with_message(
+                    format!("system object {object_id} read without an assigned version"),
+                ),
+            );
+        };
```
lxfind, line 183: **Dry-run and dev-inspect supply no system-object versions** -- Confidence: 85/100 | Importance: 80/100

(found by: Claude)
- **File**: `sui-execution/latest/sui-adapter/src/temporary_store.rs` lines 175-183 (`check_system_object_available`); the empty map originates in `sui-execution/src/latest.rs` lines 196 and 217 (unchanged, now load-bearing)
- **Problem**: `dev_inspect_transaction` passes an empty `system_object_versions` map, and `simulate_transaction` (used by dry-run at `authority.rs:2406-2426` and by dev-inspect) goes through it. This PR is the first code to read that map. For any PTB calling `verify_gcp_attestation`, `Context::new` calls `read_system_object`, finds no assigned version, hits `debug_fatal!`, and returns `UNKNOWN_INVARIANT_VIOLATION_ERROR`.
- **Impact**: Dry-run and dev-inspect of GCP attestation transactions always fail with an invariant violation in release builds, so gas estimation in the CLI and SDKs is broken for this feature. In debug, msim, and any build with `SUI_ENABLE_DEBUG_ASSERTIONS`, `debug_fatal!` aborts the node on a plain RPC request, which is a remote crash vector for test clusters and local networks.
- **Suggestion**: For non-committing modes, populate `system_object_versions` with the latest `AuthenticatorState` version in `simulate_transaction`, or have `read_system_object` fall back to the latest object when the execution mode is `DevInspect`. Never `debug_fatal!` on user-controlled input; return a normal execution error. Add a test that dry-runs a GCP transaction.

---

crates/sui-types/src/execution_status.rs
```diff
@@ -298,6 +298,10 @@ pub enum ExecutionErrorKind {
 
     #[error("Non-exclusive write input object {id} has been modified")]
     NonExclusiveWriteInputObjectModified { id: ObjectID },
+    /// Node-local, transient failure used to unwind execution when an implicitly read system
+    /// object has not reached the consensus-assigned version. This is never committed.
+    #[error("System object not available locally; transaction must be retried")]
+    SystemObjectNotAvailableLocally,
```
lxfind, line 304: **Node-local retry condition leaks into the serialized ExecutionFailureStatus; the scheduler already provides the wait** -- Confidence: 95/100 | Importance: 72/100

(found by: Claude, Codex)
- **File**: `crates/sui-types/src/execution_status.rs` lines 301-304; `crates/sui-core/src/authority.rs` lines 2019-2060 and 2176-2197; `external-crates/move/crates/move-core-types/src/vm_status.rs` (`SYSTEM_OBJECT_NOT_AVAILABLE_LOCALLY`); `crates/sui-core/src/execution_cache.rs` line 435
- **Problem**: When the assigned `AuthenticatorState` version is not yet in the local cache, execution unwinds with a new Move VM status code, which is converted into a new `ExecutionFailureStatus::SystemObjectNotAvailableLocally` variant in the BCS-serialized on-chain enum. The variant appears in `exec_failure_status.yaml`, `format__sui.yaml.snap`, and both RPC conversion tables as `unreachable!()` arms, even though the comment says it must never be committed. The PR also adds an `InnerTemporaryStore::retry_request` field touched by all five executor versions, a `notify_read_system_object_at_version` that duplicates `notify_read_input_objects`, and a `tokio::task::spawn` plus re-enqueue that wastes one full execution per affected transaction. Meanwhile `ExecutionScheduler::schedule_transaction` (`execution_scheduler_impl.rs:219-233`) already pushes an `InputKey::VersionedObject` for the accumulator root at its assigned version so the transaction is not dispatched until the object exists. The same treatment for `SUI_AUTHENTICATOR_STATE_OBJECT_ID` would make the unavailable case impossible.
- **Impact**: Local scheduling state is exposed through a consensus-visible, versioned public type that SDKs must now handle. Any missed interception would produce a nondeterministic execution status or panic in RPC conversion, and release nodes abort on panic. This is an abstraction leak between the execution scheduler and the effects format, and a wasted execution on the hot path.
- **Suggestion**: Preferred: add the authenticator-state key to `input_object_keys` in `schedule_transaction` when `system_object_versions` contains it, turn the missing-object branch of `check_system_object_available` into an invariant violation, and delete the status code, the enum variant, `retry_request`, `wait_for_system_object_and_reenqueue`, and `notify_read_system_object_at_version`. Minimal alternative: intercept the VM status before it becomes an `ExecutionErrorKind` and surface it through the existing `ExecutionRetryError` channel, removing the variant from the enum, the staged schema, and the conversion code.

---

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2665,9 +2687,98 @@ impl AuthorityPerEpochStore {
...
+                outcome.cap_exceeded += 1;
+                warn!(
+                    ?id,
+                    cap, "GCP JWK reached quorum but max_gcp_active_jwks cap was reached"
+                );
+                continue;
+            }
+            info!(epoch = ?self.epoch(), ?round, ?id, ?jwk, "GCP jwk became active");
+            output.insert_active_jwk(round, (id, jwk));
+            active_count += 1;
```
lxfind, line 2778: **Same-kid conflicts across commits never reach JwkMap; the on-chain merge keeps the old key** -- Confidence: 90/100 | Importance: 72/100

(found by: Codex; Claude described the same behavior in its risk analysis)
- **File**: `crates/sui-core/src/authority/authority_per_epoch_store.rs` lines 2756-2778; `crates/sui-framework/packages/sui-framework/sources/authenticator_state.move` lines 232-238
- **Problem**: `activate_gcp_jwks` rejects two distinct quorate keys for one `kid` only when both land in the same consensus commit. The added test explicitly permits a different key for the same `kid` in a later commit. But `update_authenticator_state` in Move retains the existing key when a new key arrives under an existing `JwkId`, and only bumps its epoch. The execution-time `JwkMap` therefore never sees two keys for one `kid`, and its fail-closed conflict handling is unreachable through the on-chain path.
- **Impact**: After a same-`kid` rotation, or conflicting endpoint responses across commits, the old key remains trusted rather than the `kid` becoming unusable. Because each re-vote refreshes the old entry's epoch, a compromised or revoked old key can stay accepted indefinitely while tokens signed by the replacement key fail. This is pre-existing Move merge behavior inherited from zkLogin, but the PR's stated fail-closed guarantee is not achieved.
- **Suggestion**: Define conflict semantics across all active state, not only candidates in one commit. Either tombstone conflicted IDs on-chain or carry enough information for execution to reject the `kid`, and treat a same-`kid` re-vote with a different key as a conflict at activation time. Add an integration test that applies two real `AuthenticatorStateUpdate`s across separate commits and then builds the `JwkMap` from the resulting on-chain state.

---

lxfind (commented):
## Automated Review: 5 Actionable Finding(s)

5 issue(s) posted as inline comments on the relevant code.

---

PR #27485, opened 2026-07-27
[move-compiler] Compile constant references as LdConst

cgswords:
## Description

Constant references in function bodies now load their value directly with `LdConst` instead of `LdNamedConst`; the values deduplicate onto the pool entries seeded by the constant definitions, so the emitted bytecode is unchanged. Removes the `LdNamedConst` IR instruction. The IR compiler's named constant table remains: IR-source constant expressions and error constant bitsets still resolve constants by name.

## Test plan

No behavior change: `move-compiler`, `move-compiler-transactional-tests` (including the `print-bytecode` snapshots), and `move-cli` suites pass with zero snapshot changes.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-ir-types/src/ast.rs
```diff
@@ -691,7 +691,6 @@ pub enum Bytecode_ {
     LdTrue,
     LdFalse,
     LdConst(Type, MoveValue),
-    LdNamedConst(ConstantName),
```
tnowacki, line 694: I'm thinking about the IR as a target for the disassembler. I wonder if we want to still have a named variant here... or something.
It is strange since from source -> ir -> bytecode, we really don't care about names in the IR, and actively kind of don't want them.
BUT! As a flip, going from bytecode -> ir, it would be nice to have some sort of pool (whether it be by name or by index) to avoid blowing the thing up. 

And we want this ir <-> bytecode roundtrip tool. And I'm not sure right now as the best way to tie these competing things. Thoughts? Feelings?

PR #27486, opened 2026-07-27
[move][move-compiler] Support cross-module constant usage within a package

cgswords:
Replaces #27350 (and formerly stacked #27485, now dropped: the IR is unchanged).

## Description

Adds cross-module constant usage to Move, gated on the `2024.alpha` edition and scoped to the defining package. A constant must be declared `public(package)` to be usable outside its module; no other visibility is valid on constants, and `#[error]` constants cannot be `public(package)` (their name/value are encoded against the aborting module's tables), which is rejected at the definition.

Every cross-module use is resolved at compile time; there is no runtime call and no friend recording:

- All constants are compiled up front by a standalone global pass at the start of CFGIR: one dependency graph over `(module, constant)` with cross-module edges, folded via a work queue. Genuine constant cycles (including mutual recursion across modules) are reported directly at the constants involved.
- Because constants are fully erased at compile time, constant references add no module dependencies: constant usage does not feed module dependency ordering, and module "cycles" that exist only through constants are legal.
- In constant definitions, a reference to another module's constant is folded at compile time.
- In function bodies, a reference compiles to a module-local copy of the constant, synthesized by a selection pass after CFGIR translation from the already-folded value with mangled names. Values are baked in at compile time, so upgrading the defining package does not change values already compiled into its users. Note this duplicates large `vector` constants per using module, bounded by the constant pool limits.
- Constants in pre-compiled modules (the analyzer's interaction pattern) are seeded from the pre-compiled program info — the conversion into HLIR form lives with the rest of type lowering in `hlir::translate` — and behave like any other constant.
- A constant that fails to fold reports an error at each cross-module use, pointing at the definition.
- Cross-package access is an error even for `public(package)` constants.

Function and constant visibility checking now share one checker in typing (`check_member_visibility`); friend recording and entry/testing visibility apply only on the function side.

A follow-up PR (see the TODO in `to_bytecode`) will build the constant pool after CFGIR lowering — pool name synthesis and common-pool elimination/CSE — also giving the IR a pooled representation for bytecode <-> IR round-tripping.

As part of the CFGIR work, the optimization interface's constant-values parameter changed shape (a `(ModuleIdent, ConstantName)`-keyed map replaces the per-module map). No IR, protocol, verifier, or VM changes are needed.

## Test plan

- `move-compiler` tests (`tests/move_2024/constants/`): folding chains, function-body uses, match patterns, macro-expanded uses, unfoldable constants (definitions error, every use reports it, no ICE), copy-name collisions, visibility (private constants rejected cross-module, invalid visibilities, friend-declaration mixing, edition gating), direct cross-module constant-cycle errors, constant-only module cycles compiling (`cross_module_constant_module_cycle`), and `#[error]` + `public(package)` rejected at the definition (with and without uses; the macro-expansion path errors through visibility at the expansion site).
- `move-compiler-transactional-tests` (`tests/constants/`): fold and copy semantics at runtime, match patterns, macro expansion, `print-bytecode` showing copies in the constant pool, previously published modules, and package upgrade via versioned publish (`--location`/`--linkage`) — both the linked run and the original version read their own compiled-in values.
- `tests/precompiled_constants.rs`: pre-compiled (analyzer-pattern) constants, and the copy names recorded in emitted source maps.
- `move-cli` build tests: intra-package usage verified at runtime via `move test`; cross-package usage rejected across real package manifests.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Move packages on the `2024.alpha` edition can now declare constants `public(package)`, making them usable from other modules of the same package, both in code and in other constant definitions.
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-compiler/src/shared/program_info.rs
```diff
@@ -42,6 +42,8 @@ pub struct ConstantInfo {
     pub index: usize,
     pub attributes: Attributes,
     pub defined_loc: Loc,
+    /// `public(package)` makes the constant usable in other modules of the package
```
tnowacki:
```suggestion
```
Either no comment here, or a comment that we have pub package or internal

external-crates/move/crates/move-compiler/src/parser/ast.rs
```diff
@@ -436,6 +436,9 @@ pub struct Constant {
     pub doc: DocComment,
     pub attributes: Vec<Attributes>,
     pub loc: Loc,
+    /// `public(package)` makes the constant usable in other modules of the package; no other
+    /// visibility is valid on constants
```
tnowacki:
```suggestion
```
Either no comment here, or a comment that we have pub package or internal

external-crates/move/crates/move-compiler/src/naming/ast.rs
```diff
@@ -268,6 +268,8 @@ pub struct Constant {
     pub index: usize,
     pub attributes: Attributes,
     pub loc: Loc,
+    /// `public(package)` makes the constant usable in other modules of the package
```
tnowacki:
```suggestion
```
Either no comment here, or a comment that we have pub package or internal

external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -1226,7 +1226,13 @@ fn exp(context: &mut Context, code: &mut IR::BytecodeBlock, e: H::Exp) {
...
         E::Copy { var: v, .. } => code.push(sp(loc, B::CopyLoc(var(v)))),
 
-        E::Constant(c) => {
+        E::Constant(m, c) => {
+            // cross-module references are replaced with module-local copies during CFGIR
+            assert!(
+                context.current_module() == Some(&m),
+                "ICE cross-module constant should have been replaced with a module-local constant \
+                 copy"
+            );
```
tnowacki, line 1288: Can we have an ICE here? or do we not have diagnostics to add to?

external-crates/move/crates/move-compiler/src/typing/ast.rs
```diff
@@ -112,6 +112,8 @@ pub struct Constant {
     pub index: usize,
     pub attributes: Attributes,
     pub loc: Loc,
+    /// `public(package)` makes the constant usable in other modules of the package
```
tnowacki:
```suggestion
```
Either no comment here, or a comment that we have pub package or internal

external-crates/move/crates/move-compiler/src/expansion/ast.rs
```diff
@@ -257,6 +257,8 @@ pub struct Constant {
     pub index: usize,
     pub attributes: Attributes,
     pub loc: Loc,
+    /// `public(package)` makes the constant usable in other modules of the package
```
tnowacki:
```suggestion
```
Either no comment here, or a comment that we have pub package or internal

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -676,7 +677,7 @@ fn base_types<'a, R: std::iter::FromIterator<H::BaseType>>(
     tys.into_iter().map(|t| base_type(reporter, t)).collect()
 }
 
-fn base_type(reporter: &DiagnosticReporter, sp!(loc, nb_): &N::Type) -> H::BaseType {
+pub(crate) fn base_type(reporter: &DiagnosticReporter, sp!(loc, nb_): &N::Type) -> H::BaseType {
```
tnowacki: I'm lost, where is this pub crate? This seems like a strange thing to be public in such a way

external-crates/move/crates/move-compiler/src/typing/core.rs
```diff
@@ -1953,31 +1953,121 @@ pub fn make_constant_type(
...
+                m,
+                c,
+                Visibility::PACKAGE,
+            );
+            context.add_diag(diag!(
+                TypeSafety::Visibility,
+                (usage_loc, msg),
+                (defined_loc, internal_msg)
+            ));
+        }
```
tnowacki: Can this logic not be shared with functions?

external-crates/move/crates/move-compiler/src/typing/dependency_ordering.rs
```diff
@@ -281,6 +284,15 @@ fn function_signature(context: &mut Context, sig: &N::FunctionSignature) {
     type_(context, &sig.return_type)
 }
 
+//**************************************************************************************************
+// Constants
+//**************************************************************************************************
+
+fn constant(context: &mut Context, cdef: &T::Constant) {
+    type_(context, &cdef.signature);
+    exp(context, &cdef.value)
```
tnowacki: Since we are erasing these, do we need to add usage to the dependency ordering? 
Like it feels fine if we have cycles here... though a bit confusing for folks that don't realize they are inlined/erased?

external-crates/move/crates/move-compiler/src/typing/translate.rs
```diff
@@ -4324,6 +4327,19 @@ fn annotated_error_const(context: &mut Context, e: &mut T::Exp, abort_or_assert_
...
+                    module_ident, constant_name, abort_or_assert_str
+                );
+                let defined_msg =
+                    "'#[error]' constants can only be used in the module that defines them";
+                context.add_diag(diag!(
+                    TypeSafety::InvalidErrorUsage,
+                    (*const_loc, msg),
+                    (defined_loc, defined_msg)
+                ));
+            }
```
tnowacki, line 4340: Should we just disallow public(package) and error? 
Erroring at the usage feels strange compared to the definition

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -50,11 +61,35 @@ pub(super) struct CFGIRDebugFlags {
...
+    /// has already reported
+    Pending,
+    /// Failed to fold to a value; an error was already reported at the definition
+    Failed,
+    /// Folded; the value is in the shared constant value map
+    Defined {
+        loc: Loc,
+        signature: Box<H::BaseType>,
+    },
+}
```
tnowacki: I'm a bit confused at this sort of complexity and would suggest something hopefully simpler. It is more passes, but hopefully a bit easier to navigate:
1. Before we compile ALL constants globally before the rest of the AST.
2. We pass this generated map into the Context or pass it in separately when compiling the rest of the AST. So the map can be used to set the constant values
3. While translating, we just collect/record the non-local constnats used (the ones from another module)
4. We write a "constant selection" pass after CFGIR translation (over the CFGIR AST) that more or less builds a constant pool to be used for the IR. So when we pass the AST to the IR, we have either a pool where ALL names are generated (not just the public(packaged) inlined ones), or it is just indexed 

My hope with (4) is to also address the issue I raised on the other PR with the ir <-> bytecode round-tripping

external-crates/move/crates/move-compiler/src/to_bytecode/translate.rs
cgswords: The context is not threaded through, and we historically bail in this way. This convention is used in a few other places in `to_bytecode` (e.g. `function_body`/`command`/`exp`), as a backstop for never generating malformed bytecode. Following up in #27643 and will update this if that lands first.

external-crates/move/crates/move-compiler/src/hlir/translate.rs
cgswords: We needed this to compute constant types and convert precompiled constants' `N::Type` signatures into `H::BaseType` when seeding their already-folded values (the vector case needs the element type, so the signature conversion can't be skipped). That conversion now lives behind a single dedicated entry point in `hlir::translate` (`precompiled_constants`), which the CFGIR constant pass calls at startup — `base_type` is private again.

external-crates/move/crates/move-compiler/src/typing/core.rs
cgswords: Merged into a single `check_member_visibility` over a `Function`/`Constant` member enum. The package arms and the internal arm (including the messages) are fully shared now. It is a little messier for book-keeping, but hopefully the logic all living together is now preferable.

external-crates/move/crates/move-compiler/src/typing/dependency_ordering.rs
cgswords: Agreed, and removed. Constants are now compiled by a standalone pass at the start of CFGIR with their own global dependency graph over `(module, constant)`, so module dependency ordering no longer needs (or has) constant edges. Constant-only module 'cycles' compile fine, and a genuine cross-module constant cycle now reports the error directly.

external-crates/move/crates/move-compiler/src/typing/translate.rs
cgswords: Done and done: `#[error]` + `public(package)` errors in expansion, next to the other constant visibility checks. The check here is downgraded to an ICE-if-no-errors guard for regression and assurance.

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
cgswords: Per our discussion, this is partway there:

1. All constants are compiled up front by a standalone global pass at the start of CFGIR, with one dependency graph over `(module, constant)` including cross-module edges, cycles reported there and participants marked failed immediately (which also let me delete the `Pending` state), folding via a work queue.
2. The resulting value map is persisted on the context, to gather in cross-module uses.
3. Function translation just records non-local uses, and a selection step afterward synthesizes the module-local copies and rewrites the uses.

Step 4 will be a follow-on, with a TODO in `to_bytecode`.

Hopefully (3) also paves the way for your later 'fold constants in functions' work, which can now just reuse the map built in (2).

external-crates/move/crates/move-cli/tests/build_tests/cross_module_constants_different_packages/sources/usage.move
```diff
@@ -0,0 +1,12 @@
+module Usage::usage {
+    use Defn::defn;
+
+    // constants cannot be used from another package, in constant definitions...
+    const DOUBLE: u64 = defn::MAX * 2;
```
tnowacki, line 6: nit, maybe use or allow unused to limit the snap

external-crates/move/crates/move-cli/tests/build_tests/cross_module_constants/args.exp
```diff
@@ -0,0 +1,18 @@
+Command `build -v`:
+INCLUDING DEPENDENCY MoveStdlib
+BUILDING A
+warning[W09011]: unused constant
+  ┌─ ./sources/defn.move:5:27
+  │
+5 │     public(package) const TEST_USED: u64 = 9;
+  │                           ^^^^^^^^^ The constant 'TEST_USED' is never used. Consider removing it.
```
tnowacki: nit, maybe use or allow unused to limit the snap

external-crates/move/crates/move-cli/tests/build_tests/cross_module_constants/args.exp
```diff
@@ -0,0 +1,18 @@
+Command `build -v`:
```
tnowacki, line 1: Did I never migrate these to snap... I should have really done that. I wonder why I didn't

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
```diff
@@ -0,0 +1,600 @@
...
+    copies: BTreeMap<ConstantId, ConstantName>,
+    copy_defs: Vec<(ConstantName, G::Constant)>,
+    copy_names: BTreeSet<Symbol>,
+    /// continues after the module's own constants so copies are emitted after them
+    next_index: usize,
+}
+
+//**************************************************************************************************
+// Impls
+//**************************************************************************************************
```
tnowacki: ❤️

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
```diff
@@ -0,0 +1,600 @@
...
+    fn new(constants: &UniqueMap<ConstantName, G::Constant>) -> Self {
+        Self {
+            copies: BTreeMap::new(),
+            copy_defs: vec![],
+            copy_names: BTreeSet::new(),
+            next_index: constants
+                .key_cloned_iter()
+                .map(|(_, cdef)| cdef.index + 1)
+                .max()
+                .unwrap_or(0),
```
tnowacki: Can we not simplify this to `len()`? 
I suppose I am not sure at the time I am leaving this comment how the `index` is set

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -581,32 +464,139 @@ fn constant_(
...
+        "Invalid use of constant '{}::{}'. Its value could not be computed",
+        m, c
+    );
+    let mut diag = diag!(CodeGeneration::UnfoldableConstant, (use_loc, msg));
+    diag.add_secondary_label((defined_loc, format!("'{}' is defined here", c)));
+    diag
+}
+
+/// The error reported at each use of a constant defined outside the current compilation
+pub(super) fn outside_compilation_use(
```
tnowacki: What does outside of compilation mean? Is this just normal visibility? If so, why isn't this error reported in `typing`?

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -581,32 +464,139 @@ fn constant_(
...
+
+/// The error reported at each use of a constant whose definition could not be evaluated to a value
+pub(super) fn unfoldable_constant_use_error(
+    m: &ModuleIdent,
+    c: &ConstantName,
+    use_loc: Loc,
+    defined_loc: Loc,
+) -> Diagnostic {
+    let msg = format!(
+        "Invalid use of constant '{}::{}'. Its value could not be computed",
```
tnowacki: Is this in the case of cycles?I'm a bit lost at the code here of how this differs from the old `CANNOT_FOLD` error

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
```diff
@@ -0,0 +1,600 @@
...
+            if consts.contains_key(dep) {
+                *blocked_on.get_mut(user).unwrap() += 1;
+                dependents.entry(*dep).or_default().push(*user);
+            }
+        }
+    }
+
+    let mut ready: VecDeque<ConstantId> = blocked_on
+        .iter()
+        .filter_map(|(node, blockers)| (*blockers == 0).then_some(*node))
```
tnowacki: Interesting. My first thought here would have been to toposort them, but I imagine this is faster?

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -246,214 +280,48 @@ fn module(
...
+// Constants
 //**************************************************************************************************
 
-fn constants(
+/// Constants of modules outside this compilation (e.g. the analyzer's pre-compiled dependencies)
+/// were folded when they were compiled; seed their values so they can be folded into constant
+/// definitions and copied into function bodies like any other constant. The conversion into HLIR
+/// form lives with the rest of type lowering in `hlir::translate`; the map is transient to this
+/// pass.
+fn seed_precompiled_constants(
```
tnowacki: Should this be in `constants`?

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -581,32 +464,139 @@ fn constant_(
...
+        C::Mutate(lhs, rhs) => vec![lhs, rhs],
+        C::Break(_) | C::Continue(_) | C::Jump { .. } => vec![],
+    };
+    exps.into_iter()
+}
+
+/// Reports a constant definition that could not be folded to a value: cross-module references to
+/// constants that themselves failed to fold (or are outside the current compilation) are reported
+/// as the root cause, otherwise the generic unfoldable-constant error is reported at `loc`.
+fn report_cannot_fold<'a>(
```
tnowacki: I wonder if these items should be in `constants`

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -50,11 +50,16 @@ pub(super) struct CFGIRDebugFlags {
...
-    env: &'env CompilationEnv,
-    info: &'env TypingProgramInfo,
+pub(super) struct Context<'env> {
+    pub(super) env: &'env CompilationEnv,
+    pub(super) info: &'env TypingProgramInfo,
     reporter: DiagnosticReporter<'env>,
-    current_package: Option<Symbol>,
+    pub(super) current_package: Option<Symbol>,
+    /// the fold state of every constant in the program, populated by the global constant pass
+    pub(super) constant_defs: BTreeMap<(ModuleIdent, ConstantName), ConstantEntry>,
```
tnowacki: Thoughts on moving this into a separate context for `constants`?
I don't really mind, but the first thing that stood out to me were the crosslinks between `constants` and `translate`. Not a big deal, but it is nice to prevent the cycles when we can

external-crates/move/crates/move-compiler/src/hlir/translate.rs
```diff
@@ -3060,3 +3062,91 @@ fn gen_unused_warnings(
...
+pub(crate) fn precompiled_constants(
+    reporter: &DiagnosticReporter,
+    info: &TypingProgramInfo,
+    compiled: &BTreeSet<ModuleIdent>,
+    packages: &BTreeSet<Option<Symbol>>,
+) -> BTreeMap<(ModuleIdent, ConstantName), (Loc, H::BaseType, Option<H::Value>)> {
+    // Precompiled constants arrive as MoveValues, but we need them as H::Value to include them.
+    // This should only matter for partial compilation, such as for the IDE.
+    #[growing_stack]
+    fn value_from_move_value(mv: &MoveValue, ty: &H::BaseType) -> Option<H::Value> {
```
tnowacki, line 3086: We have another version of this floating around somewhere, yeah? Or is it the other direction?

external-crates/move/crates/move-compiler/tests/linter/move_2024/cross_module_constants.move
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Cross-module constant uses produce no lint output: the synthesized constant copies reuse
+// the constant's location, so a lint firing on them would point at the declaration
```
tnowacki: What is the lint here that should fire? Not sure what this is testing

external-crates/move/crates/move-compiler/tests/linter/move_2024/cross_module_constants.move
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
```
tnowacki:
```suggestion
```

external-crates/move/crates/move-compiler/tests/move_2024/constants/cross_module_alias.move
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
```
tnowacki:
```suggestion
```

external-crates/move/crates/move-compiler/tests/move_2024/constants/invalid_cross_module_unfoldable.snap
```diff
@@ -0,0 +1,44 @@
...
+   │       ^ Constant uses constant 0x42::c::X, which has a circular dependency
+
+error[EC08001]: cannot compute constant value
+   ┌─ tests/move_2024/constants/invalid_cross_module_unfoldable.move:27:25
+   │
+ 9 │ public(package) const BAD: u64 = 1 / 0;
+   │                       --- 'BAD' is defined here
+   ·
+27 │ public fun bad(): u64 { a::BAD }
+   │                         ^^^^^^ Invalid use of constant '0x42::a::BAD'. Its value could not be computed
```
tnowacki: I see now.
I wonder if we just don't give an error here. I feel like the cycle error is enough? Thoughts?

external-crates/move/crates/move-compiler/tests/precompiled_constants.rs
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! `public(package)` constants in pre-compiled modules can be used from source modules of the
```
tnowacki: It might be easier to test this by using transactional tests, where you first publish a module in the package that is written in `mvir`, and then publish a new module in the package using source. 
A a whole, this file scares me a bit, and will likely prove to be a bit annoying. Just in that having source code written in strings in a file has always proven to be a bit annoying in my experience.

external-crates/move/crates/move-cli/tests/build_tests/cross_module_constants_different_packages/sources/usage.move
cgswords: Added `#[allow(unused_const)]` and regenerated the exp.

external-crates/move/crates/move-cli/tests/build_tests/cross_module_constants/args.exp
cgswords: Added `#[allow(unused_const)]` and regenerated the exp.

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
cgswords: `len()` breaks when a constant failed to fold. Added a comment explaining.

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
cgswords: Typing already rejects cross-package constant access (E04001), but visibility errors are non-blocking, so CFGIR still runs afterward and this arm fired. I removed it, though, in favor of the `ice_assert`, and added a test case to demonstrate.

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
cgswords: Cycles don't get here any more, after the SCC revision. This happens now if we find a usage that has no value (like a pre-compiled constant that already failed in its own compilation).

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
cgswords: It _was_ Kahn's toposort interleaved with the folding, but it was easy enough to swap it for a direct SCC via petgraph. This also gets rid of the work-queue bookkeeping, and cycle reporting falls out for free.

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
cgswords: Sorry, yes. It got un-moved in some rebase shuffle. Moved.

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
cgswords: Moved.

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
cgswords: Done and done. There is still some cross-calling, especially after moving the code into `constants.rs`, but the constant state is now fully isolated from the rest of CFGIR.

external-crates/move/crates/move-compiler/src/hlir/translate.rs
cgswords: Other direction: `move_value_from_value` (cfgir) goes `H::Value` → `MoveValue`; this one seeds pre-compiled `MoveValue`s back into `H::Value`. We should really just unify them at some point.

external-crates/move/crates/move-compiler/tests/linter/move_2024/cross_module_constants.move
cgswords: The test is ensuring none fire: the synthesized copies must stay invisible to lints, else e.g. a naming lint would fire pointing at the other module's declaration (the copies reuse the original's loc). Reworded the comment to say so.

external-crates/move/crates/move-compiler/tests/move_2024/constants/invalid_cross_module_unfoldable.snap
cgswords: I thought it would be nicer in the IDE situation to see the use-site errors, but consistency with intra-module behavior is more important. Uses of a failed constant are now marked failed / rewritten to `UnresolvedError` silently.

external-crates/move/crates/move-compiler/tests/precompiled_constants.rs
cgswords: That makes sense, I agree about the string-source file. The mvir shape gets converted to an interface file with no constants, so that test doesn't work. This test was really for the analyzer with pre-compiled deps, though, so I moved it there.

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
```diff
@@ -0,0 +1,985 @@
...
+        context: &mut Context,
+        constant_values: &ConstantValues,
+        m: ModuleIdent,
+        c: ConstantName,
+        signature: H::BaseType,
+        defined_loc: Loc,
+        use_loc: Loc,
+    ) -> ConstantName {
+        let symbol: Symbol = format!(
+            "const{NEW_NAME_DELIM}{}{NEW_NAME_DELIM}{}",
```
tnowacki: I think we need the module address here as well, or something. 
Visibility errors are non-blocking, so it is possible to reach this without being in the same package, so just `module`+`c` is going to collide

```move
module 0x42::m { public(package) const C: u64 = 1; }
module 0x43::m { public(package) const C: u64 = 2; }
module 0x42::ex { public fun collision(): u64 { 0x42::m::C + 0x43::m::C } 
```

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
```diff
@@ -0,0 +1,985 @@
...
+                    loc,
+                    signature: Box::new(signature.clone()),
+                },
+                Some(move_value_from_value(value)),
+            )
+        }
+        // an error was reported during folding, at this definition or an earlier one
+        _ => (ConstantEntry::Failed, None),
+    };
+    let (entry, value) = entry;
```
tnowacki: nit, can this be move to line 384?

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
```diff
@@ -0,0 +1,985 @@
...
+    }
+
+    let mut output = BTreeSet::new();
+    let (_, block) = &constant.value;
+    dep_block(&mut output, block);
+    output
+}
+
+//**************************************************************************************************
+// Constant dependencies
```
tnowacki: Is this supposed to be a different header? 
```suggestion
// Constant cycles
```
?

tnowacki: Or `errors`?

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
```diff
@@ -222,10 +248,21 @@ fn module(
         functions: hfunctions,
         constants: hconstants,
     } = mdef;
+    assert!(
```
tnowacki: Can't be an ice_assert? 
I feel like we could at least point to the module location or something?

external-crates/move/crates/move-compiler/src/expansion/translate.rs
```diff
@@ -1411,6 +1412,47 @@ fn check_visibility_modifiers(
...
+            && !matches!(constant.visibility, E::Visibility::Internal)
+        {
+            let vis_loc = constant
+                .visibility
+                .loc()
+                .expect("ICE non-internal visibility must have a loc");
+            let msg = format!(
+                "Invalid constant declaration. '#[{}]' constants cannot be declared '{}'",
+                known_attributes::ErrorAttribute::ERROR,
+                E::Visibility::PACKAGE,
```
tnowacki: The visibility here could be `public`. Maybe best to flip this and say Something like "'#[{}]' constants must be internal to their module"? That or swap this for `constant.visiblity`

external-crates/move/crates/move-compiler/tests/move_2024/constants/invalid_cross_module_unfoldable_use_only.move
```diff
@@ -0,0 +1,18 @@
+// An unfoldable constant used only from another module's function body: one error at the
+// definition and one at the use, with no ICE
```
tnowacki:
```suggestion
// An unfoldable constant used only from another module's function body with one 
// error at the definition. 
```
I find it a bit strange that a lot of these test comments are like "with no ICE"... I fuckin hope so lol 
I hope the compiler doesn't crash too 😜

tnowacki (commented):
We should have tests for the two issues I cited, the collision and the visibility error message on `#[error]`. 

Additionally from a robo test coverage review 🤖 
- `cfgir::constants::exp`'s `E::Pack` / `E::PackVariant` and `command`'s `C::Mutate` arms have no
  test exercising a cross-module constant in a struct field or on a mutation RHS.
- A cross-module constant passed directly as a call argument (`E::ModuleCall`). Currently only
  reached via `assert!` and binops.

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
cgswords: Tested, whoops. Good catch, thanks.

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
cgswords: Moved

external-crates/move/crates/move-compiler/src/cfgir/constants.rs
cgswords: Changed to errors, plus restructred

external-crates/move/crates/move-compiler/src/cfgir/translate.rs
cgswords: It can and should be, plus one other one I found along the way was changed

external-crates/move/crates/move-compiler/src/expansion/translate.rs
cgswords: `constant.visibility` plus some rewording done

external-crates/move/crates/move-compiler/tests/move_2024/constants/invalid_cross_module_unfoldable_use_only.move
cgswords: 👀 Yeah....

external-crates/move/crates/move-compiler/tests/inq-review-20260828-152827.md
```diff
@@ -0,0 +1,25 @@
+# Review feedback
```
tnowacki: Got some 🤖 🐕 files in here by accident it seems

external-crates/move/crates/move-compiler/src/expansion/translate.rs
```diff
@@ -1420,23 +1422,32 @@ fn check_visibility_modifiers(
             .get_(&known_attributes::AttributeKind_::Error)
             && !matches!(constant.visibility, E::Visibility::Internal)
         {
-            let vis_loc = constant
-                .visibility
-                .loc()
-                .expect("ICE non-internal visibility must have a loc");
+            use known_attributes::ErrorAttribute as Err;
+
```
tnowacki: Nit shadowing `Err` seems a bit strange but.. sure? Maybe `ErrAttr`?

tnowacki (approved):

Issue #27464, opened 2026-07-27
Protocol-version mismatch crashes with a raw Rust panic instead of a clean CLI error


dwill6413:
## Summary

When a Sui CLI binary's maximum supported protocol version is older than the network it's talking to (a common situation, since Sui ships new protocol versions roughly biweekly and nothing prompts developers to update proactively), the CLI crashes with a raw `panic!` and a Rust backtrace hint, rather than a normal CLI error.

## Steps to Reproduce Issue

Run any `sui client` subcommand that touches the network (reproduced on `sui client upgrade`) with a CLI binary whose max protocol version is behind the target network's current epoch protocol version.

## Expected Result

A clean, normal CLI error explaining the CLI is out of date and needs to be upgraded, in the same style as the CLI's other argument/usage errors.

## Actual Result

```
thread 'main' (12536) panicked at crates\sui-protocol-config\src\lib.rs:2911:9:
Network protocol version is ProtocolVersion(129), but the maximum supported version by the binary is 128. Please upgrade the binary.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2026-07-27T02:20:54.583195Z ERROR telemetry_subscribers: panicked at crates\sui-protocol-config\src\lib.rs:2911:9: ...
```

The message itself is fine — "Please upgrade the binary" is correct, actionable advice. The problem is purely presentational: a panic plus backtrace prompt plus an ERROR-level telemetry log dump reads exactly like an internal crash/bug, not an expected, recoverable, user-facing condition. Anyone hitting this for the first time reasonably assumes something is broken rather than realizing they just need to run one `curl`/`suiup` command.

## Suggested Fix

Detect the protocol-version mismatch during the initial RPC handshake or epoch fetch, before entering any code path that would otherwise panic, and surface it through the CLI's normal error-printing path:

```
error: your Sui CLI (protocol 128) is older than the network you're connecting to (protocol 129).
Install a newer release: https://github.com/MystenLabs/sui/releases
(or, if using suiup: `suiup update sui@mainnet`)
```

All the information needed for this message already exists at the point of the panic — it just needs to be routed through normal error handling instead of `panic!`.

## System Information

* OS: Windows (also expected to reproduce on Linux/macOS, since this is a client/network handshake issue, not a platform-specific one)
* Sui CLI: mainnet-v1.72.5 talking to a network on protocol 129

dwill6413:
[...]

PR #27488, opened 2026-07-27
[cli] Replace panic error on protocol version incompatibilty with resaonable error message

stefan-mysten:
## Description 

This PR replaces the panic when a protocol version incompatiblity happens with a reasonable error. See https://github.com/MystenLabs/sui/issues/27464 for more details on the issue.
## Test plan 

New tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui/src/client_commands.rs
+++ b/crates/sui/src/client_commands.rs
@@ -4498,30 +4498,3 @@ fn find_faucet_url(address: SuiAddress, rpc: &str) -> anyhow::Result<String> {
         bail!("Cannot recognize the active network. Please provide the gas faucet full URL.")
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn protocol_config_out_of_range_errors_instead_of_panicking() {
-        let too_new = ProtocolVersion::MAX_ALLOWED + 1;
-        let err = protocol_config_for_version(too_new, Chain::Unknown)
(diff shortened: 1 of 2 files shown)


jordanjennings-mysten (approved):

crates/sui/src/client_commands.rs
```diff
@@ -3829,6 +3836,32 @@ async fn check_protocol_version_and_warn(client: &Client) -> Result<(), anyhow::
...
+
+    if version < ProtocolVersion::MIN {
+        bail!(
+            "protocol version {} is older than the minimum version {} supported by this CLI",
+            version.as_u64(),
+            ProtocolVersion::MIN.as_u64(),
+        );
+    }
+
+    Ok(ProtocolConfig::get_for_version(version, chain))
```
amnn, line 3862: Would it be easier to use `get_for_version_if_supported`? It does mean that you don't get to return a different error for versions that are too old vs too new, but practically speaking "too old" is not going to happen.

amnn (approved):

crates/sui/src/client_commands.rs
stefan-mysten: I see, I assume that'll work fine too. Not sure what's the benefit of using it though.

PR #27492, opened 2026-07-28
sui-indexer-alt-graphql: Fix affectedObject filter typo in TransactionFilter::active_filters()

evan-wall-mysten:
## Description

`TransactionFilter::active_filters()` pushed `"affectedObjects"` (plural) instead of `"affectedObject"`, so the `tx_affected_objects` pipeline was never selected when computing the available range for `affectedObject`-filtered `transactions` queries. Instead of returning a clean `FEATURE_UNAVAILABLE` error when that pipeline isn't configured, such queries silently proceeded.

First commit adds a regression test that captures the buggy silent-success behavior, second commit fixes the typo and updates the test's assertion to the correct error, third commit adds the same style of "requires pipeline" coverage for every other filter-driven pipeline branch in `Query.[transactions]`/`Query.[events]` (`affectedAddress`, `sentAddress`, `function`, `kind`, `module`, and the default type-filtered events path) that had the same latent risk, though none were actually buggy.

## Test plan

New tests in `crates/sui-indexer-alt-e2e-tests/tests/graphql_available_range_tests.rs` exercise the real `transactions`/`events` resolvers end-to-end against a cluster missing the relevant pipeline, asserting the `FEATURE_UNAVAILABLE` error via a shared `assert_filter_requires_pipeline` harness. Verified the `affectedObject` test actually catches the bug by temporarily reintroducing the typo and confirming the test fails (the query returns no error instead of the expected one).

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: `transactions(filter: {affectedObject: ...})` now correctly returns a `FEATURE_UNAVAILABLE` error when the `tx_affected_objects` pipeline isn't configured, instead of silently succeeding.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql_available_range_tests.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql_available_range_tests.rs
@@ -203,7 +203,9 @@ async fn test_transaction_pagination_pruning() {
         cluster.create_checkpoint().await;
     }
 
-    let transactions_in_range = query_transactions(&cluster, b).await;
+    let affected_address_filter = json!({ "affectedAddress": b.to_string() });
+
+    let transactions_in_range = query_transactions(&cluster, affected_address_filter.clone()).await;
     let actual = collect_digests(&transactions_in_range);
     assert_eq!(&a_txs, &actual);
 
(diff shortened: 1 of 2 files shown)


amnn (approved):
Nice fix/find!

PR #27494, opened 2026-07-28
[discovery] reduce state write-lock hold time in update_known_peers_versioned

akichidis:
## Description 
`update_known_peers_versioned` held the discovery state write lock for its entire body due to lifetime extension of the guard temporary. This kept per-peer Ed25519 signature verification (up to MAX_PEERS_TO_SEND peers) and endpoint_manager.update_endpoint calls — which acquire locks in other subsystems — under the lock, stalling the discovery RPC handlers and event loop that contend on it. The PR is reducing the lock held time so it gives the opportunity for other parts of the component to process commands.

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-network/src/discovery/mod.rs
+++ b/crates/sui-network/src/discovery/mod.rs
@@ -1372,8 +1372,11 @@ fn update_known_peers_versioned(
         .our_info_v2
         .as_ref()
         .and_then(|info| info.peer_id());
-    let known_peers_v2 = &mut state.write().unwrap().known_peers_v2;
-    let mut trusted_peer_changed = false;
+
+    // Filter and verify before taking the state write lock: signature
+    // verification is CPU-heavy (up to MAX_PEERS_TO_SEND peers per call).
+    let mut verified_peers = Vec::new();
+    let mut endpoint_updates = Vec::new();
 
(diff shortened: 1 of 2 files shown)


crates/sui-network/src/discovery/mod.rs
```diff
@@ -1409,34 +1412,53 @@ fn update_known_peers_versioned(
...
+
+    // Forward after the merge: handle_peer_address_change's Chain-source
+    // fallback reads known_peers_v2, so this batch must be visible there
+    // before these updates are processed. update_endpoint also acquires locks
+    // in other subsystems, so it must not run under the state lock.
+    for (endpoint_id, addrs) in endpoint_updates {
+        if let Err(e) =
+            endpoint_manager.update_endpoint(endpoint_id.clone(), AddressSource::Discovery, addrs)
+        {
+            warn!(?endpoint_id, "Failed to update endpoint: {e:?}");
```
aschran: This warn is redundant, there's already a warn inside the called fn

aschran (approved):

PR #27497, opened 2026-07-28
[sui-execution-backtest] Fix system package versions

tzakian:
## Description 

My shenanigans around system packages in the execution backtesting were incorrect -- would lead to wrong versions and transaction dep digests. 

Instead require a GQL endpoint to load the system packages correctly.

## Test plan 

Replayed a bunch of failing txs that ran afoul of these shenanigans.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


dariorussi (approved):
LGTM, just a question.
Are we loading a system package all the time?
do we have a cache for them?
we could also load all of them at startup and call it a day, maybe?

tzakian:
> Are we loading a system package all the time?
Nope!

> do we have a cache for them?

Yes! We cache them per-epoch, so we read them once per-epoch, and the cache and share those system packages across the whole epoch.

> we could also load all of them at startup and call it a day, maybe?

This is similar(ish) to what we do now. We don't a priori load _everything_ but pretty close/the overall overhead for this should be pretty much imperceptible.

amnn:
Note that if you know which epoch you are fetching the system packages for, this is a simpler query that means you don't even need to take the dependency on knowing the set of system packages at that epoch:

```graphql
query SystemPackages($epoch: UInt53) {
  epoch(epochId: $epoch) {
    systemPackages(first: 50) {
      pageInfo {
        hasNextPage
      }
      nodes {
        address
        objectBcs
      }
    }
  }
}
```

PR #27502, opened 2026-07-29
sui-adapter: share unified linkage command facts

tzakian:
## Description 

Refactor unified linkage analysis around shared command "facts".

Execution-time PTB linkage now extracts command-level facts before performing the existing unified constraint folding over this.

This will be used in the PR on top of this to allow for sharing the core linkage analysis logic across both signing and execution, without signing needing to construct a fully loaded PTB.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
+++ b/sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
@@ -2,21 +2,19 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    data_store::VerifiedPackageStore,
+    data_store::{PackageMetadata, PackageStore, VerifiedPackageStore},
     static_programmable_transactions::{
         linkage::{
             analysis::LinkageAnalyzer,
+            facts::{LinkageCommandFacts, LinkageFacts, ModuleInitFacts},
             resolution::{ResolutionTable, VersionConstraint, add_and_unify, get_package},
             resolved_linkage::{ExecutableLinkage, ResolvedLinkage},
         },
-        loading::ast::{
-            Command, DeserializedPackage, LoadedFunction, PackagePayload, Transaction, Type,
-        },
+        loading::ast::{Command, DeserializedPackage, PackagePayload, Transaction},
     },
 };
 use move_binary_format::{CompiledModule, file_format::Visibility};
-use move_vm_runtime::validation::verification::ast::Package as VerifiedPackage;
-use std::{collections::BTreeMap, sync::Arc};
+use std::collections::{BTreeMap, BTreeSet};
 use sui_protocol_config::ProtocolConfig;
 use sui_types::{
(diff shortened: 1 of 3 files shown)


sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -87,88 +95,203 @@ fn analyze_command<E: ExecutionErrorTrait>(
...
+            type_defining_ids: ty.all_addresses().into_iter().map(ObjectID::from).collect(),
+        }),
+        Command::MakeMoveVec(None, _)
+        | Command::TransferObjects(_, _)
+        | Command::SplitCoins(_, _)
+        | Command::MergeCoins(_, _) => Ok(LinkageCommandFacts::Noop),
+    }
+}
+
+pub(crate) fn compute_unified_linkage<E: ExecutionErrorTrait, S: PackageStore + ?Sized>(
```
tnowacki, line 145: This will become fully `pub`?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
```diff
@@ -87,88 +95,203 @@ fn analyze_command<E: ExecutionErrorTrait>(
...
+    facts: impl IntoIterator<Item = LinkageCommandFacts>,
+    linkage_analysis: &LinkageAnalyzer,
+    package_store: &S,
+    protocol_config: &ProtocolConfig,
+    mut execution_original_ids: Option<&mut BTreeSet<ObjectID>>,
```
tnowacki, line 150: What is this arg?

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
tzakian: will still be `pub(crate)` and be abstracted by a an execution entrypoint that takes a non-`loaded` PTB, constructs the linkage facts from that, and then calls into this function (but that happens on this side, and not on the authority side if that makes sense)

sui-execution/latest/sui-adapter/src/static_programmable_transactions/linkage/single_linkage.rs
tzakian: It's used to distinguish packages that occur as "executable" vs "for type only". This will be clearer in further PRs on this stack, which I should hopefully have up later today, but basically we want to allow you to use a type from forbidden version of a package, but not allow you to link against it for execution. 

Essentially, this list gives us the set of forbid lists we should be looking at after we gather the unified/lumpy linkage.

PR #27505, opened 2026-07-29
[consensus] wire FlexCommitter into Core

mwtian:
## Description

Follow-up to #26876. When `enable_v3` is on, Core uses FlexCommitter for the local commit decision rule (`try_commit_v3`) and for building sub-dags from certified commits synced by commit syncer (`process_certified_commits`), instead of the UniversalCommitter path. No behavior change while `enable_v3` is off (the default).

Additional changes on the v3 path:
- Report per-commit metrics and logs from `Core::post_commit`, matching the legacy `CommitObserver::handle_committed_leaders` behavior.
- Bound `DagState` recovery without relying on CommitInfo, which is not persisted with `enable_v3`: skip the commit scan that loads full subdags for legacy leader scoring, and recover `last_committed_rounds` from the existing gc-bounded backward scan that restores block committed statuses.
- Assert that certified commits chain onto the last local commit digest in `process_certified_commits`, restoring the fork detection the legacy path performs by re-linearizing certified commits locally.
- Refresh the proposer's propagation scores on each commit, so score-based ancestor exclusion tracks current v3 leader scoring instead of staying frozen at the scores from Core construction.

## Test plan

New Core-level tests for the v3 path:
- `try_commit_v3_local_commits`: local commits through the FlexCommitter decision rule.
- `add_certified_commits_v3`: direct commits of synced certified commits, continuation with the local rule, and skipping already-committed commits.
- `test_core_recover_from_store_v3`: restart recovery from a store without CommitInfo.

Also covered by existing tests in `consensus-core`.

### DIFF
--- a/consensus/core/src/commit_observer.rs
+++ b/consensus/core/src/commit_observer.rs
@@ -270,7 +270,9 @@ impl CommitObserver {
         );
     }
 
-    fn report_commit_metrics(&self, commit: &CommittedSubDag) {
+    /// Reports per-commit metrics and logs the commit. Called for every commit on the
+    /// legacy path via `report_metrics`, and directly by `Core::post_commit` on the v3 path.
+    pub(crate) fn report_commit_metrics(&self, commit: &CommittedSubDag) {
         let metrics = &self.context.metrics.node_metrics;
         let utc_now = self.context.clock.timestamp_utc_ms();
 
--- a/consensus/core/src/commit_observer.rs
+++ b/consensus/core/src/commit_observer.rs
@@ -312,6 +312,9 @@ impl CommitObserver {
         for commit in committed {
             self.report_commit_metrics(commit);
         }
+        // Only the legacy path batches multiple subdags per call. The v3 path
+        // handles one commit at a time, so this metric would always observe 1
+        // and is intentionally not reported there.
         self.context
             .metrics
             .node_metrics
--- a/consensus/core/src/commit_observer.rs
+++ b/consensus/core/src/commit_observer.rs
@@ -310,6 +310,9 @@ impl CommitObserver {
         for commit in committed {
             self.report_commit_metrics(commit);
         }
+        // Only the legacy path batches multiple subdags per call. The v3 path
+        // handles one commit at a time, so this metric would always observe 1
+        // and is intentionally not reported there.
         self.context
             .metrics
             .node_metrics
--- a/consensus/core/src/commit_observer.rs
+++ b/consensus/core/src/commit_observer.rs
@@ -95,7 +95,7 @@ impl CommitObserver {
     /// of selected leader blocks, and whether they come from local committer or commit sync remotely.
     ///
     /// Also, buffers the commits to DagState and forwards committed subdags to commit finalizer.
-    pub(crate) fn handle_commit(
+    pub(crate) fn handle_committed_leaders(
         &mut self,
         committed_leaders: Vec<VerifiedBlock>,
         local: bool,
@@ -105,7 +105,7 @@ impl CommitObserver {
             .metrics
             .node_metrics
             .scope_processing_time
-            .with_label_values(&["CommitObserver::handle_commit"])
+            .with_label_values(&["CommitObserver::handle_committed_leaders"])
             .start_timer();
 
         let mut committed_sub_dags = self.commit_interpreter.handle_commit(committed_leaders);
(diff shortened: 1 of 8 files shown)


consensus/core/src/commit_observer.rs
```diff
@@ -134,6 +134,11 @@ impl CommitObserver {
         Ok(committed_sub_dags)
     }
 
+    /// Forwards a committed subdag to the commit finalizer. Used by `Core::post_commit`.
+    pub(crate) fn send_to_finalizer(&self, subdag: CommittedSubDag) -> ConsensusResult<()> {
+        self.commit_finalizer_handle.send(subdag)
+    }
+
```
akichidis, line 143: Looking on this do we see the `CommitObserver` eventually getting removed once v3 gets permanently enabled? Probably the recovery/metrics processes could be folded to other components? Just curious whether you had any thoughts

consensus/core/src/dag_state.rs
```diff
@@ -260,6 +268,19 @@ impl DagState {
                     break;
                 }
 
+                // On the v3 path, last_committed_rounds are recovered from this scan
+                // instead of CommitInfo. Authorities without committed blocks above
+                // gc_round keep under-estimated values, which is fine because nothing
+                // on the v3 path reads last_committed_rounds at or below gc_round.
+                if context.protocol_config.enable_v3() {
+                    for block_ref in commit.blocks() {
+                        state.last_committed_rounds[block_ref.author] = max(
```
akichidis, line 278: Let's also document the semantics of the `last_committed_rounds`  - as now someone shouldn't expect accuracy here for authorities of committed rounds < gc_round

akichidis (approved):

consensus/core/src/core.rs
```diff
@@ -751,6 +793,200 @@ impl Core {
...
+            let Some((commit, subdag)) = self
+                .flex_committer
+                .as_mut()
+                .expect("FlexCommitter must be initialized on the v3 path")
+                .try_commit(next_commit_leaders)
+            else {
+                break;
+            };
+
+            self.post_commit(commit, subdag.clone())?;
```
arun-koshy, line 913: if post commit advances next_commit_index on every commit then maybe_refresh_pending_commit_state would reset FlexCommitters cached round state. So try_direct_commit would have to redecide every slot in every pending round from scratch. This would be a problem during catchup right?

consensus/core/src/flex_committer.rs
```diff
@@ -395,6 +392,22 @@ impl FlexCommitter {
             .collect();
         drop(dag_state);
 
+        // Every block in the commit's leader round is a committed leader (same-round
+        // blocks cannot be ancestors of each other, so they can only appear in the
+        // commit as leaders). Report them like `build_commit` does for local
+        // decisions, so `committed_leaders_total` covers the commit sync path too.
+        for block in &blocks {
```
arun-koshy, line 460: A round that ends fully skipped never reaches build_commit. Its slots direct-skip/indirect-skip series are never incremented, and the round state is dropped on the next schedule refresh. Should we ensure we dont omit these skip metrics

consensus/core/src/core.rs
```diff
@@ -2877,6 +3133,342 @@ mod test {
...
+        let store = core.store.clone();
+        let mut core = core.core;
+
+        let mut dag_builder = DagBuilder::new(core.context.clone());
+        dag_builder.layers(1..=12).build();
+
+        // Certified commits for leader rounds 1..=5. Their blocks have not been
+        // accepted locally and are provided by the certified commits themselves.
+        let certified_commits = dag_builder
+            .get_sub_dag_and_certified_commits(1..=5)
```
arun-koshy, line 3200: does this helper need to be updated to produce multi-leader certified commits?

consensus/core/src/core.rs
akichidis: That's good point and it was flagged by AI too actually, but for this to happen it will mean that we have dealed with some long period of asynchrony where we accumulated lots of rounds and for some reason didn't decide yet on the leaders? Otherwise during recovery we should expect to have a few rounds to run the decision rules so it should be ok?

consensus/core/src/commit_observer.rs
mwtian: Yes, where to keep the post commit logic is still to be decided. I want `Core` and related components to have a "hub and spoke" model, instead of having the data flow distributed among multiple components. We will remove Linearizer after v2 is disabled. Potentially CommitObserver will get folded into Core as well.

consensus/core/src/core.rs
mwtian: This logic should be on the same order of magnitude as existing v2 committers, which recomputes from scratch leader certificate of certificate with every incoming block. In v3, computations for next commit is cached, which is already an improvement. I switched v3 to cache leader computations for the same allowed leaders sequence.

consensus/core/src/core.rs
mwtian: Reworked the test.

PR #27507, opened 2026-07-30
Add pull-based consensus transaction pool for block proposals

aschran:
## Description

#27417 added a `TransactionPool` trait to consensus-core so the proposer can pull transactions at block-proposal time. This adds the Sui-side implementation as an opt-in mode (`consensus-transaction-pool` in NodeConfig): the validator RPC inserts user transactions into a gas-price-prioritized pool that consensus drains directly, replacing the admission-queue drain thread that pushes through `ConsensusAdapter`. Backpressure reduces to a single pool-capacity limit (no inflight counting, submit semaphore, or drain failover).

`ConsensusAdapter` keeps managing system-transaction and ping submission (GC retries, processed-elsewhere dedup); in pull mode its consensus client submits into the pool's dedicated system/ping lanes, so that logic stays mode-independent. Push mode remains the default; the admission queue actor is not spawned when pull mode is configured.

## Test plan

- Unit tests over the pool state machine: take/ack budgets and bundle atomicity, dropped-ack requeue, close and user-lane-close semantics at epoch end, epoch-mismatch handling, displaced-entry notification (delivered outside the pool mutex, with a drop guard enforcing that `try_insert` outcomes are notified), and the watch-based epoch handoff (including lost-wakeup races).
- New e2e simtests: basic flow + ping + epoch progress, gas-price eviction/rejection under a full pool, a pending submission resolving across an epoch boundary, and a submission arriving in the window where the new epoch store is published but the old epoch's consensus is still running (held open deterministically via a reconfiguration fail point).
- Existing `admission_queue` e2e tests cover the unchanged push mode.

## Release notes

- [ ] Protocol: 
- [x] Nodes (Validators and Full nodes): Adds an opt-in `consensus-transaction-pool` config for validators: when set, consensus pulls transactions from a validator-side, gas-price-prioritized pool at block-proposal time instead of the admission queue pushing them to consensus.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


mwtian (commented):
Have you double checked against the design principles in https://github.com/MystenLabs/sui/blob/e5d18feb21574d4b757bcc250edc38c4696ab87d/crates/sui-core/src/consensus_submission_pipeline.md?

crates/sui-config/src/node.rs
```diff
@@ -323,6 +329,22 @@ impl Default for TransactionDriverConfig {
     }
 }
 
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct ConsensusTransactionPoolConfig {
+    /// Maximum queued user-lane entries. A soft bundle counts as one entry,
+    /// matching the existing admission queue. Defaults to the consensus
```
akichidis, line 336: I haven't checked the rest of the PR yet, but will that effectively "undercount" the actual number of pending transactions to get admitted to? (ex 20K records of soft bundle transactions, with 5 txes each, will account for 100K pending transactions - although we do "count" here 20K)

aschran:
> Have you double checked against the design principles in https://github.com/MystenLabs/sui/blob/e5d18feb21574d4b757bcc250edc38c4696ab87d/crates/sui-core/src/consensus_submission_pipeline.md?

thanks for the suggestion. i gave it a pass. most of the principles still apply i think, and are done same or better. one question is re: principle #2, whether we want to add anything to track per-entry early completion of tx pending in the pool (i.e. filter out stuff that is pending in our proposal pool which was already proposed+sequenced elsewhere). debatable whether this is worth it (principle #4 says some dupes are fine if the alternative is too complicated), I think a case could be made. if so it shouldn't block this pr, even if it does block broader release to all validators.

crates/sui-core/src/consensus_transaction_pool.rs
```diff
@@ -0,0 +1,1631 @@
...
+
+        let mut total_bytes = 0usize;
+        let mut serialized = Vec::with_capacity(transactions.len());
+        let mut keys = Vec::with_capacity(transactions.len());
+        for transaction in transactions {
+            let bytes =
+                bcs::to_bytes(transaction).expect("Serializing consensus transaction cannot fail");
+            let transaction_bytes =
+                u64::try_from(bytes.len()).expect("transaction size fits into u64");
+            if transaction_bytes > protocol_config.max_transaction_size_bytes() {
```
mystenmark, line 619: shouldn't the transaction have been rejected already if it was too large?

aschran: I think this would be the only place that size of system tx is checked, but I can improve the error handling here

crates/sui-core/src/authority_server.rs
```diff
@@ -403,16 +404,26 @@ impl ValidatorServiceMetrics {
...
+    Pool,
     /// Submit directly to consensus, bypassing the queue — used when the queue
     /// is turned off by config, temporarily disabled by failover, or for a ping
     /// request. Individual txs are rejected when consensus is saturated
     /// (pre-queue behavior).
     Direct,
 }
 
+#[derive(Clone)]
+pub enum UserSubmissionPath {
```
alex-mysten, line 425: I was initially confused about the difference between this enum and the one above. Might be worth a comment or adjusting naming here that this describes authority config and the other pertains to request routing.

crates/sui-core/src/authority_server.rs
```diff
@@ -863,7 +875,8 @@ impl ValidatorService {
 
             // Use the pre-queue per-tx consensus overload reject on the direct
             // submission path (queue off, failover, or ping).
-            if matches!(submit_mode, AdmissionQueueSubmitMode::Direct)
+            if matches!(submit_mode, UserSubmissionMode::Direct)
+                && !matches!(&self.user_submission_path, UserSubmissionPath::Pool(_))
```
alex-mysten, line 889: Maybe a comment here about why overload does not apply when using Pool?

crates/sui-core/src/authority_server.rs
```diff
@@ -1514,24 +1587,26 @@ impl ValidatorService {
...
-        let Some(aq) = &self.admission_queue else {
-            return AdmissionQueueSubmitMode::Direct;
-        };
-
+    fn classify_submit_mode(&self, is_ping_request: bool) -> UserSubmissionMode {
         // Ping requests carry no transactions and must not wait behind queued
         // work; submit them directly to consensus.
         if is_ping_request {
-            return AdmissionQueueSubmitMode::Direct;
+            return UserSubmissionMode::Direct;
```
alex-mysten, line 1606: why use the adapter for pings?

crates/sui-config/src/node.rs
```diff
@@ -282,6 +282,12 @@ pub struct NodeConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub transaction_driver_config: Option<TransactionDriverConfig>,
 
+    /// When set, consensus pulls transactions directly from a validator-side pool
+    /// instead of the admission-queue drain thread pushing them. This takes
+    /// precedence over `authority_overload_config.admission_queue_enabled`.
```
alex-mysten, line 287: What's the desired end state? It seems like we want the pool to replace the queue, and then absorb the functionality of the adapter - is that right?

crates/sui-core/src/authority_server.rs
```diff
@@ -1373,6 +1389,63 @@ impl ValidatorService {
...
+                                (transaction.data().transaction_data().gas_price()
+                                    / epoch_store.reference_gas_price().max(1))
+                                .max(1);
+                            epoch_store.submitted_transaction_cache.record_submitted_tx(
+                                transaction.digest(),
+                                amplification_factor as u32,
+                                submitter_client_addr,
+                            );
+                        }
+                    }
```
alex-mysten: Should this and the other duplicated policy logic instead be factored out of consensus_adapter into a shared helper?

crates/sui-core/src/consensus_transaction_pool.rs
```diff
@@ -0,0 +1,2013 @@
...
+
+        let mut total_bytes = 0usize;
+        let mut serialized = Vec::with_capacity(transactions.len());
+        for transaction in transactions {
+            let bytes =
+                bcs::to_bytes(transaction).expect("Serializing consensus transaction cannot fail");
+            let transaction_bytes =
+                u64::try_from(bytes.len()).expect("transaction size fits into u64");
+            if transaction_bytes > protocol_config.max_transaction_size_bytes() {
+                return Err(consensus_client_error(ClientError::OversizedTransaction(
```
alex-mysten, line 620: Are we worried about the risk of divergence here with TransactionClient or is this just a backstop?

crates/sui-config/src/node.rs
aschran: yes ultimately push mode should be removed from the binary, pool and queue and adapter all merged.

crates/sui-core/src/authority_server.rs
aschran: added comments

crates/sui-core/src/authority_server.rs
aschran: added

crates/sui-core/src/authority_server.rs
aschran: the logic just already exists there so it was a diff-minimizing move. ultimate goal is to consolidate adapter, admission queue, and pool

crates/sui-core/src/consensus_transaction_pool.rs
aschran: it's not just a backstop, TransactionClient is bypassed when consensus proposer pulls directly from the pool. I'll add a comment. re: divergence risk, hopefully we can not wait too long to delete the push code

crates/sui-core/src/authority_server.rs
aschran: factored

crates/sui-core/src/authority_server.rs
alex-mysten: do we intend to consolidate before rolling this out, or is the change to factor the adapter logic something that would need to be independently gated & verified?

aschran: My plan was to remove push mode and consolidate only after pull mode is out on mainnet and we are very confident in its stability. Until then we'd keep it in case we need to flip back

alex-mysten: Will `consensus_adapter` continue to exist when push mode is removed, or will its current functionality be rolled into `pool` or something else?

crates/sui-core/src/consensus_transaction_pool.rs
```diff
@@ -0,0 +1,2016 @@
...
+        let entry = PoolEntry {
+            transactions,
+            total_bytes,
+            gas_price,
+            ack: PendingAck::new(EntryAck::Positions(sender), keys),
+            metrics: self.metrics.clone(),
+            processed,
+        };
+
+        let mut inner = self.inner.lock();
```
alex-mysten, line 418: It seems like lock contention could be a bottleneck, perhaps we should add instrumentation

crates/sui-core/src/consensus_transaction_pool.rs
```diff
@@ -0,0 +1,2016 @@
...
+        let user = match self.check_user_lane_open_locked(caller_epoch, &mut inner) {
+            Ok(user) => user,
+            Err(error) => {
+                drop(inner);
+                entry.ack.resolve_error(error.clone());
+                return Err(error);
+            }
+        };
+        let newly_inserted = user.insert(entry)?;
```
alex-mysten: If you follow the destructor chain for PoolEntry, each evicted or rejected entry takes locks within `notify_read::cleanup`, sharded on digest. If we see duplicate spam to the same validator, partitions could be hot. Wondering if we should cleanup evicted or rejected entries outside of the `inner` lock.

crates/sui-config/src/node.rs
```diff
@@ -323,6 +329,22 @@ impl Default for TransactionDriverConfig {
     }
 }
 
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct ConsensusTransactionPoolConfig {
+    /// Maximum queued user-lane entries. A soft bundle counts as one entry,
+    /// matching the existing admission queue. Defaults to the consensus
+    /// `max_pending_transactions` setting.
+    pub max_pending_transactions: Option<usize>,
```
alex-mysten, line 338: Does using the same value make sense here now that submit semaphore is out of the picture?

crates/sui-core/src/consensus_transaction_pool.rs
alex-mysten: should we add some defense on the consensus provider side?

crates/sui-core/src/authority_server.rs
alex-mysten: My thinking is that if we ultimately plan to remove `consensus_adapter`, maybe it's easier to verify if we do that behind the same gate. Also, the contract for `transaction_pool` and neighboring components is easier to reason about, and we avoid an extra variant mode of operation

crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1426,6 +1427,31 @@ impl AuthorityPerEpochStore {
...
+    pub(crate) fn record_submitted_user_transactions(
+        &self,
+        transactions: &[ConsensusTransaction],
+        submitter_client_addr: Option<IpAddr>,
+    ) {
+        for transaction in transactions {
+            if let Some(tx) = transaction.kind.as_user_transaction() {
+                let amplification_factor = (tx.data().transaction_data().gas_price()
+                    / self.reference_gas_price().max(1))
+                .max(1);
```
mystenmark, line 1488: with the allowed proposers work, we are allowing up to 3x amplification for free - not sure if we will need to account for that here

crates/sui-core/src/consensus_manager/mod.rs
```diff
@@ -283,7 +291,33 @@ impl ConsensusManager {
             protocol_config.version
         );
 
-        self.consensus_client.set(self.client.clone());
+        let pool_context = self
+            .transaction_pool_context
+            .as_ref()
+            .filter(|_| self.protocol_keypair.is_some() && epoch_store.is_validator());
```
akichidis: checking only `epoch_store.is_validator()` should be enough. The protocol keypair should be guaranteed to be present in that case.

crates/sui-core/src/consensus_transaction_pool.rs
```diff
@@ -0,0 +1,2016 @@
...
+
+    /// Inserts user transactions to the pool. Returns the receiver for their
+    /// eventual positions, and bool indicating whether all transactions in the
+    /// group were newly inserted.
+    /// Fails fast on epoch mismatch, a closed pool/user lane, block-limit
+    /// violations, or when outbid by the eviction policy.
+    pub fn try_insert(
+        &self,
+        caller_epoch: EpochId,
+        gas_price: u64,
```
akichidis, line 381: Could you please add a comment on the role of `gas_price` ?

crates/sui-core/src/consensus_transaction_pool.rs
```diff
@@ -0,0 +1,2016 @@
...
+            .iter()
+            .any(ConsensusTransaction::is_user_transaction)
+        {
+            debug_fatal!(
+                "user transaction submitted through the consensus transaction pool client"
+            );
+            return Err(SuiErrorKind::GenericAuthorityError {
+                error: "user transactions cannot be submitted through the transaction pool client"
+                    .to_string(),
+            }
```
akichidis, line 471: I guess this is a path that our code should never take? As in, having a component trying to use `submit` for user transactions? Just wondering whether it deservers an `assert` check instead as this would break the system assumptions?

crates/sui-core/src/authority_server.rs
aschran: the goal of the current design is to avoid adding copy-pasted code to TransactionPool so that we don't run into problems with edits that miss one spot or the other. after we get rid of push mode, we could either keep what remains of consensus adapter as a thin layer that just implements the special logic for handling system tx, or fold it into the pool. the latter is probably a bit cleaner but not a huge difference either way.

I'm not sure I understand what you mean about the "same gate"? can you elaborate?

crates/sui-core/src/consensus_transaction_pool.rs
aschran: we could, but that's probably a separate/orthogonal PR since this one just implements the interface provided by the consensus code.

crates/sui-config/src/node.rs
aschran: I agree this is definitely set way too high, but that's regardless of the semaphore and push vs pull doesn't change the appropriate setting afaict. we should prob lower it in another much simpler pr

crates/sui-core/src/consensus_transaction_pool.rs
aschran: thanks, improvement incoming

crates/sui-core/src/consensus_transaction_pool.rs
aschran: per your other comment I moved all the drops/entry teardowns outside of the locks. are there any other potential bottlenecks that concern you? the critical section here seems very small, and afaict there are no other locks in the codebase that we have instrumented with contention metrics.

i'm a bit wary to add a one-off if this lock isn't especially concerning compared to other locks we have, though maybe we should be instrumenting all of them better and this could be the first? (probably implementing something like `monitored_mutex` in `mysten_metrics` that is analogous to the other monitored types.) wdyt?

crates/sui-config/src/node.rs
aschran: yes, but this is no change in behavior from admission queue. worst case we store some extra tx in the queue, either way gas price remains the deciding factor for what gets in.

crates/sui-core/src/consensus_transaction_pool.rs
aschran: debug_fatal is always the preferred choice when a recoverable invariant violation occurs, no? what's the problem here?

crates/sui-core/src/authority/authority_per_epoch_store.rs
aschran: there are a lot of places where we tally uncharged actions towards spam, even if it's allowed (eg reading objects, wait for effects, etc.). seems like the same logic applies here, do you agree or am I missing any nuance?

crates/sui-core/src/consensus_transaction_pool.rs
aschran: done

crates/sui-core/src/consensus_manager/mod.rs
aschran: done

crates/sui-core/src/authority_server.rs
alex-mysten: I was thinking that that second change to fold that logic into the pool might involve some amount of risk. Since it seems like we're planning a data collection phase, maybe it makes sense to accept the risk of drift, in exchange for being able to test & activate the final version all behind the same config.

If that second part isn't a big risk or change, then it's not worth worrying about

crates/sui-core/src/consensus_transaction_pool.rs
alex-mysten: This seems like the most likely place to me where we could potentially hit a perf bottleneck, and would otherwise would be difficult to observe. I think that monitored_mutex makes a lot of sense. I'm also pretty convinced this strategy in general will perform a lot better than what we have. Won't block on it

crates/sui-config/src/node.rs
alex-mysten: Right now, this setting sizes both the queue and the semaphore, and also now the size of the pool. I'm not sure we should also tie the size of the pool to the same config. The number has a different meaning and a likely different ideal value in the different modes of operation. Having a different config would make it harder to configure incorrectly

alex-mysten: Sorry, reread the code. Looks fine to me

alex-mysten (approved):
LGTM. Left a few suggestions in comment replies, but none are blocking

crates/sui-core/src/consensus_transaction_pool.rs
akichidis: Is it recoverable though? For example if our codepath at some point tries to send checkpoint signatures via this method, how would we manage to certify any checkpoint?

crates/sui-core/src/consensus_transaction_pool.rs
```diff
@@ -0,0 +1,2026 @@
...
+    oneshot::Receiver<BlockStatus>,
+);
+
+/// How a queued submission is resolved once its fate is known.
+enum EntryAck {
+    /// RPC user path: positions on inclusion, error on eviction / epoch end.
+    Positions(oneshot::Sender<SuiResult<Vec<ConsensusPosition>>>),
+    /// `ConsensusAdapter` path (system transactions): block inclusion ack plus a
+    /// `BlockStatus` subscription.
+    Inclusion(oneshot::Sender<BlockInclusion>),
```
akichidis: Looking on the comments maybe those could be renamed to something like:

```
EntryAck::User
EntryAck::System
```

and probably match closer the domain?

crates/sui-core/src/consensus_transaction_pool.rs
```diff
@@ -0,0 +1,2026 @@
...
+            next_index += entry.transactions.len();
+            let indices = (start..next_index)
+                .map(|index| {
+                    TransactionIndex::try_from(index)
+                        .expect("validated transaction index must fit TransactionIndex")
+                })
+                .collect::<Vec<_>>();
+            self.metrics
+                .queue_wait_latency
+                .observe(entry.ack.created.elapsed().as_secs_f64());
```
akichidis, line 777: I would suggest we label by tx type, as user Vs system transactions will probably have different latency profiles. Similarly below we might want to track the ping latencies as well.

akichidis (approved):
Changes look good. I would just like to highlight your comment @aschran 

> principle https://github.com/MystenLabs/sui/pull/2, whether we want to add anything to track per-entry early completion of tx pending in the pool (i.e. filter out stuff that is pending in our proposal pool which was already proposed+sequenced elsewhere). debatable whether this is worth it

so there are a few observed scenarios where this was deemed useful. One noteworthy is when a validator halt their block proposals and it keeps submitting to consensus checkpoint signatures. We've seen occurrences in the past where this could happen (ex a validator catching up in consensus mid epoch) which was leading to a pile of checkpoint signature transactions which would later flush once block proposals were restored for them. 

Now, I don't suggest you address it in this PR, as it already bears enough scope, but I believe we should further consider the possible ramifications if we don't add any such mechanism. At minimum I believe for the system transactions during `take()` we should also check if those are already processed (as you do with user), so at least we'll avoid flushing those to blocks. But the pile up will probably need to get addressed differently.

crates/sui-core/src/consensus_transaction_pool.rs
aschran: I don't know what you mean? it wouldn't send the signatures, because it would return an error. the question is just do we want to panic the binary, or only raise an invariant violation. both of them trigger an alert, but only one immediately takes down the entire network.

PR #27509, opened 2026-07-30
Add blocking synchronization primitives

mystenmark:
## Description

Foundational sync primitives for the upcoming "block execution on unavailable object versions" work, where a transaction executing on a blocking thread must wait for a value another execution will produce.

Each primitive begins with a non-blocking `try_` check and only blocks if nothing is ready. When it does block it first releases a per-thread *execution permit* (`sync::execution_permit`), so that limited execution concurrency cannot deadlock on self-dependent work; the permit is not re-acquired afterward (any transient over-subscription self-resolves). Under msim, blocking is a `yield_blocking` spin loop rather than OS parking.

The primitives are standalone in this PR — wiring the permit into the execution driver lands with the consumer that actually blocks.

Depends on the msim blocking-pool change (mysten-sim#76) for `msim::task::yield_blocking`: the workspace `msim` pin must be bumped before this can merge, so the `simtest` job will not compile until then. Verified under `--cfg msim` locally with `LOCAL_MSIM_PATH` against that branch.

## Test plan

New unit tests in `sync::oneshot` and `sync::execution_permit` cover the try-first fast path, blocking until send / sender-drop, and that the execution permit is released when the receiver blocks but kept when a value is already available.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9649,8 +9649,10 @@ dependencies = [
  "antithesis_sdk",
  "either",
  "futures",
+ "msim",
  "mysten-metrics",
  "once_cell",
+ "oneshot",
  "parking_lot 0.12.3",
  "prost-types 0.14.1",
  "rand 0.8.5",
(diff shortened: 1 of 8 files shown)


Cargo.lock
```diff
@@ -11640,7 +11648,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash 2.1.1",
  "rustls",
- "socket2 0.6.3",
+ "socket2 0.5.6",
```
lxfind: are these downgrades auto-generated?

mystenmark: no idea how this happened, fixed

lxfind (approved):

PR #27516, opened 2026-07-30
sui-indexer-alt-reader: chunk oversized LedgerGrpcReader batch requests

evan-wall-mysten:
## Description

`LedgerGrpcReader` (used by GraphQL and jsonrpc for ledger gRPC/KV-RPC access) built a single unchunked `BatchGetTransactions`/`BatchGetObjects` request per `DataLoader` batch, which could exceed the ledger service's own 200/1000 batch-size caps and surface as an opaque gRPC error.

Introduces a `ChunkedLoader<K>` trait: `LedgerGrpcReader` implements it per key type with `chunk_size()` (the limit) and `load_chunk()` (the actual gRPC call); a provided `load_chunked()` method splits keys into chunks, dispatches `load_chunk` concurrently per chunk, and merges the results. A single generic `impl<K> Loader<K> for LedgerGrpcReader` (where `Self: ChunkedLoader<K>`) forwards to `load_chunked`, covering every chunked key type at once. `CheckpointKey`'s loader already fans out per-key rather than batching, so it keeps its own ordinary `Loader<CheckpointKey>` impl and never implements `ChunkedLoader`.

## Test plan

New chunking tests in `sui-indexer-alt-reader` (`events`/`transactions`/`objects`) construct oversized key sets against a mock ledger service and assert requests are split to the batch limit.

## Release notes

- [ ] Nothing user-facing changed.

### DIFF
--- a/crates/sui-indexer-alt-reader/src/ledger_grpc_reader.rs
+++ b/crates/sui-indexer-alt-reader/src/ledger_grpc_reader.rs
@@ -1,10 +1,14 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::HashMap;
+use std::hash::Hash;
 use std::time::Duration;
 
 use anyhow::Context;
 use async_graphql::dataloader::DataLoader;
+use async_graphql::dataloader::Loader;
+use futures::future::try_join_all;
 use prometheus::Registry;
 use sui_rpc::Client;
 use sui_rpc::proto::sui::rpc::v2 as grpc;
@@ -49,6 +53,68 @@ pub struct LedgerGrpcReader {
     timeout: Option<Duration>,
 }
 
+/// Maximum number of transaction digests `LedgerGrpcReader` will put in a single
+/// `BatchGetTransactions` call, matching the ledger gRPC/KV-RPC service's own hard cap.
+pub(crate) const MAX_BATCH_GET_TRANSACTIONS: usize = 200;
+
+/// Maximum number of object keys `LedgerGrpcReader` will put in a single `BatchGetObjects`
+/// call, matching the ledger gRPC/KV-RPC service's own hard cap.
+pub(crate) const MAX_BATCH_GET_OBJECTS: usize = 1000;
+
+/// Implemented by `LedgerGrpcReader` for each key type whose `Loader::load` needs to stay under
+/// the ledger service's batch-size limit — `DataLoader`'s own `max_batch_size` is only a dispatch
+/// trigger, not a hard cap on how many keys reach a single `Loader::load` call. `load_chunk`
+/// supplies the raw, single-chunk fetch and `chunk_size` supplies the limit; `load_chunked`'s
+/// default body splits `keys` into chunks, dispatches `load_chunk` concurrently per chunk, and
+/// merges the results.
+///
+/// `pub`, not `pub(crate)`: it's referenced through `LedgerGrpcReader`'s blanket `Loader<K>` impl
+/// below (`type Value = <Self as ChunkedLoader<K>>::Value`), and that impl is part of
+/// `LedgerGrpcReader`'s public interface since both the type and `Loader` are public.
+#[async_trait::async_trait]
+pub trait ChunkedLoader<K>
+where
+    K: Send + Sync + Hash + Eq + Clone + 'static,
+{
+    type Value: Send + Sync + Clone + 'static;
+    type Error: Send + Sync + Clone + 'static;
+
+    fn chunk_size(&self) -> usize;
+
+    async fn load_chunk(&self, keys: &[K]) -> Result<HashMap<K, Self::Value>, Self::Error>;
+
+    async fn load_chunked(&self, keys: &[K]) -> Result<HashMap<K, Self::Value>, Self::Error>
+    where
+        Self: Sync,
+    {
+        let limit = self.chunk_size();
+
+        let mut results = HashMap::new();
+        for batch in try_join_all(keys.chunks(limit).map(|chunk| self.load_chunk(chunk))).await? {
+            results.extend(batch);
+        }
+        Ok(results)
+    }
+}
+
+/// Covers every key type `LedgerGrpcReader` implements [`ChunkedLoader`] for. Coherent because
+/// `Self` (`LedgerGrpcReader`) is a concrete local type — only `K` is generic — unlike a bare
+/// `impl<K, T: ChunkedLoader<K>> Loader<K> for T`, which the orphan rules reject since neither
+/// `Loader` (foreign) nor `T` (an uncovered generic) is local.
+#[async_trait::async_trait]
+impl<K> Loader<K> for LedgerGrpcReader
+where
+    K: Send + Sync + Hash + Eq + Clone + 'static,
+    Self: ChunkedLoader<K>,
+{
+    type Value = <Self as ChunkedLoader<K>>::Value;
+    type Error = <Self as ChunkedLoader<K>>::Error;
+
+    async fn load(&self, keys: &[K]) -> Result<HashMap<K, Self::Value>, Self::Error> {
+        self.load_chunked(keys).await
+    }
+}
+
 impl LedgerGrpcArgs {
     pub fn new(
         statement_timeout_ms: Option<u64>,
@@ -90,7 +156,7 @@ impl LedgerGrpcReader {
         Ok(Self { client, timeout })
     }
 
-    pub fn as_data_loader(&self) -> DataLoader<Self> {
+    pub(crate) fn as_data_loader(&self) -> DataLoader<Self> {
         DataLoader::new(self.clone(), tokio::spawn)
     }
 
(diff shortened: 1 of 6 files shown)


crates/sui-types/src/limits.rs
tpham-mysten, line 1: Not sure if `sui-types` is the right place for these constants as it is the core protocol/blockchain types crate. Maybe `sui-rpc` is better as the 3 crates already depend on it?

crates/sui-kv-rpc/src/v2/get_object.rs
```diff
@@ -16,7 +16,7 @@ use sui_types::storage::ObjectKey;
 use crate::PackageResolver;
 use crate::render::object_to_response;
 
-pub const MAX_BATCH_REQUESTS: usize = 1000;
+pub const MAX_BATCH_REQUESTS: usize = sui_types::limits::MAX_BATCH_GET_OBJECTS;
```
tpham-mysten: nit: is this still necessary?

crates/sui-indexer-alt-reader/src/ledger_grpc_reader.rs
```diff
@@ -207,3 +212,160 @@ impl Default for LedgerGrpcArgs {
...
+pub(crate) async fn chunked_load<'k, K, V, F>(
+    keys: &'k [K],
+    limit: usize,
+    load_batch: F,
+) -> Result<HashMap<K, V>, Error>
+where
+    K: Eq + Hash,
+    F: Fn(&'k [K]) -> BoxFuture<'k, Result<HashMap<K, V>, Error>>,
+{
+    if keys.is_empty() {
```
tpham-mysten: nit: this is probably not necessary

crates/sui-indexer-alt-reader/src/ledger_grpc_reader.rs
```diff
@@ -207,3 +212,160 @@ impl Default for LedgerGrpcArgs {
...
+/// `max_batch_size` only affects when concurrently-coalescing keys are
+/// dispatched, not how many a single bulk call contains), so every
+/// `LedgerGrpcReader` loader needs this to stay under the ledger service's
+/// batch-size limit.
+///
+/// `load_batch` returns a boxed future (rather than a bare `impl Future`) so
+/// its lifetime can be tied directly to `keys`' — a generic `Fn(&[K]) -> Fut`
+/// bound can't otherwise express that every chunk borrows from the same
+/// slice.
+pub(crate) async fn chunked_load<'k, K, V, F>(
```
tpham-mysten: optional nit: `load_in_chunks` might read more natural as an action on the call site?

crates/sui-rpc-api/src/grpc/v2/ledger_service/get_transaction.rs
```diff
@@ -113,12 +113,7 @@ pub fn batch_get_transactions(
         FieldMaskTree::from(read_mask)
     };
 
-    if digests.len() > MAX_BATCH_REQUESTS {
-        return Err(RpcError::new(
-            tonic::Code::InvalidArgument,
-            format!("number of batch requests exceed limit of {MAX_BATCH_REQUESTS}"),
-        ));
-    }
+    crate::batch_limits::validate_batch_size(digests.len(), MAX_BATCH_REQUESTS)?;
```
tpham-mysten: tiny nit: import the crate to make it cleaner?

crates/sui-rpc-api/src/batch_limits.rs
```diff
@@ -0,0 +1,34 @@
...
+use crate::RpcError;
+
+/// Validate that a batch request's size does not exceed `max`. Shared by every
+/// batch-oriented handler (`BatchGetTransactions`, `BatchGetObjects`) so the
+/// limit check and error shape stay identical across them.
+pub fn validate_batch_size(len: usize, max: usize) -> Result<(), RpcError> {
+    if len > max {
+        return Err(RpcError::new(
+            tonic::Code::InvalidArgument,
+            format!("number of batch requests exceed limit of {max}"),
```
tpham-mysten: nit: Include the actual len as well?

crates/sui-types/src/limits.rs
```diff
@@ -0,0 +1,12 @@
...
+
+/// Maximum number of digests that can be requested in a single
+/// `BatchGetTransactions` call (gRPC `LedgerService`, KV-RPC, and the
+/// GraphQL indexer's underlying transaction/event batch loaders).
+pub const MAX_BATCH_GET_TRANSACTIONS: usize = 200;
+
+/// Maximum number of objects that can be requested in a single
+/// `BatchGetObjects` call (gRPC `LedgerService`, KV-RPC, and the GraphQL
+/// indexer's underlying object batch loader).
+pub const MAX_BATCH_GET_OBJECTS: usize = 1000;
```
nickvikeras: IMHO this is overkill and creates unnecessary coupling. Keep in mind these limits are independent operational tuning knobs and might need to drift in the future. I.e. we might find that it makes sense to support bigger limits in archival than on the full node, but that GraphQL tail latency is best with smaller batches. 

It also could create a footgun because if we change the constant and deploy both services, they could end up out of sync for a brief period of time and create a spike of failures depending on rollout order. I think we currently roll out both services concurrently so if you increased the limit, graphql could start requesting bigger batches than the old grpc instances supported.

tpham-mysten (approved):
Overall looking good. Just a few nits

crates/sui-types/src/limits.rs
evan-wall-mysten: I thought this was the simplest way of fixing the current problem and showing the dependency using code instead of a comment that is easier to ignore. I agree it restricts being able to change it, but that limit still existed before in a less easy to identify way.

crates/sui-kv-rpc/src/v2/get_object.rs
evan-wall-mysten: Removed — good catch, it was just an alias at that point. Call sites now reference `sui_types::limits::MAX_BATCH_GET_TRANSACTIONS`/`MAX_BATCH_GET_OBJECTS` directly.

crates/sui-types/src/limits.rs
amnn: +1 -- what about the crate where @wlmyng put the shared cursor stuff?

crates/sui-types/src/limits.rs
amnn: I agree -- it would be better to make this configurable as part the ledger grpc reader configuration.

crates/sui-kv-rpc/src/v2/get_object.rs
evan-wall-mysten: Replaced these constants with the `limits::` ones.

crates/sui-types/src/limits.rs
evan-wall-mysten: The shared cursor is in `sui-rpc-cursor` which does not seem like a good place either or at least the name is not good

crates/sui-indexer-alt-reader/src/ledger_grpc_reader.rs
```diff
@@ -207,3 +229,160 @@ impl Default for LedgerGrpcArgs {
...
     }
 }
+
+/// Splits `keys` into chunks of at most `limit`, fetches each chunk
+/// concurrently via `load_batch`, and merges the results. `DataLoader` does
+/// not itself bound how many keys reach a single `Loader::load` call (its
+/// `max_batch_size` only affects when concurrently-coalescing keys are
+/// dispatched, not how many a single bulk call contains), so every
+/// `LedgerGrpcReader` loader needs this to stay under the ledger service's
+/// batch-size limit.
```
wlmyng: huh, that's kind of unfortunate

wlmyng: Wonder if we should have a wrapper struct over the `Loader` that manages this chunking. On `LedgerGrpcReader` we have `fn max_batch_get_transactions`, `fn max_batch_get_objects`, which get set on the `impl Loader<K: Key> for LedgerGrpcReader`, where each `fn load` is implemented with `load_in_chunks ... load_batch()` with some `self.max_*()` 


I'm thinking of some
```
/// Per key-type batch-size limit. `Chunked` uses this to split oversized key sets before they reach the inner `Loader::load`
trait BatchLimit

/// `DataLoader` `max_batch_size` is only a dispatch trigger. Enforce backend batch limits at the `Loader::load` boundary
pub(crate) struct Chunked<L>(pub L);

impl<K, L> Loader<K> for Chunked<L> {
    async fn load() {
        let chunks = try_join_all(
            keys.chunks(self.0.batch_limit())
            .map(|chunk| self.0.load(chunk))
        )
    }
}
```

usage would be like
```
impl BatchLimit<TransactionKey> for LedgerGrpcReader {
    fn batch_limit(&self) -> usize { self.max_batch_get_transactions }
}
```
with the `fn load` implementation exactly what it looked like before the chunking change

The main justification is just, since every location of ledger service reader needs to do this, can we reduce that repetition, and helps centralize limits to key mapping

wlmyng (commented):
Also wonder if there could be an upstream change for dataloader to also chunk. Though this wouldn't be useful if diff ledger service calls have diff limits

Non-blocking suggestion: with the `DataLoader<Chunked...`, we just need an `impl BatchLimit` for each key, otherwise identical behavior to current in main

crates/sui-indexer-alt-reader/src/ledger_grpc_reader.rs
evan-wall-mysten: I made a slight change to this because not all keys (`CheckpointKey`) require chunking.

crates/sui-types/src/limits.rs
evan-wall-mysten: Moved this to https://github.com/MystenLabs/sui/pull/27539

nickvikeras (approved):

crates/sui-types/src/limits.rs
evan-wall-mysten: moved this to https://github.com/MystenLabs/sui/pull/27539

PR #27523, opened 2026-07-31
Implement implicitly read system objects with a blocking wait in execution

lxfind:
## Description

Instead of unwinding the VM, recording a retry request, discarding effects, and re-enqueuing the certificate, execution now calls `ObjectStore::load_implicitly_read_system_object`; the writeback cache waits for the requested version through `NotifyRead::read_one_blocking` when it is not yet local.

Blocked executions release their execution-driver permit before parking, and transactions that mutate implicitly-read system objects use a dedicated permit pool so a writer cannot be queued behind readers waiting on that writer. The PR also records implicit system-object reads in effects as `ReadOnlyRoot` unchanged consensus objects so checkpoint execution, replay, and follow-up consumers can reconstruct the exact versions.

The mechanism remains inert until a consumer performs an implicit system-object read. The follow-up #27524 moves object-funds withdraw sufficiency into the VM and is the first real consumer.

## Test plan

- Direct coverage in this PR: `mysten-common` tests cover the blocking primitive and execution-permit release behavior; `sui-core::test_load_implicitly_read_system_object` verifies the writeback cache blocks until the target version is committed, returns immediately when already available, and releases the execution permit while blocked.
- Follow-up coverage in #27524: object-funds e2e tests execute the first consumer path with exact-version accumulator reads, verify the accumulator root is recorded as `ReadOnlyRoot`, exercise same-commit unsettled withdrawals, and catch up a new fullnode from the recorded effects. The deterministic assertion that a wait occurred remains in the direct unit test above.

Validated with:

```text
SUI_SKIP_SIMTESTS=1 cargo nextest run --no-capture -p mysten-common execution_permit oneshot
SUI_SKIP_SIMTESTS=1 cargo nextest run --no-capture -p sui-core test_load_implicitly_read_system_object
```

The #27524 stack was also validated with its object-funds simtests, transactional test, `sui-core object_funds`, all-workspace clippy, formatting, and snapshot commands.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [x] JSON-RPC: `SuiTransactionBlockEffects.shared_objects` is derived from the effects' accessed consensus objects, which can now include implicitly read system objects such as the accumulator root once a consumer activates. Consumers that treat the field as declared shared inputs must filter against the transaction inputs; legacy `sui-replay` has been updated accordingly.
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -151,6 +174,37 @@ impl<'backing> TemporaryStore<'backing> {
...
+            SystemObjectVersionRequirements::Exact(versions) => {
+                let object = self
+                    .store
+                    // If this transaction needs to read an implicit system object,
+                    // the version must be assigned from consensus.
+                    .load_implicitly_read_system_object(
+                        object_id,
+                        versions.get(object_id).unwrap(),
+                    );
+                // Record the read at `required_version` (which is what the transaction depends
```
mystenmark: `required_version` doesn't appear anywhere else in this function so its not 100% clear what this comment is referring to.

crates/sui-types/src/transaction.rs
```diff
@@ -1901,6 +1893,16 @@ impl TransactionKind {
         )
     }
 
+    pub fn mutates_implicitly_read_system_object(&self) -> bool {
+        self.shared_input_objects().any(|obj| {
+            let mutates = match obj.mutability {
+                SharedObjectMutability::Immutable => false,
+                SharedObjectMutability::Mutable | SharedObjectMutability::NonExclusiveWrite => true,
+            };
+            mutates && crate::IMPLICITLY_READ_SYSTEM_OBJECTS.contains(&obj.id)
```
mystenmark: can you move this predicate to a function `is_implicity_read_system_object()`

crates/sui-core/src/execution_driver.rs
```diff
@@ -88,30 +93,17 @@ pub async fn execution_process(
...
-        // the semaphore in this context. The scope measures time blocked waiting for a
-        // permit, i.e. how saturated execution concurrency is.
-        let permit = {
-            let _scope = monitored_scope("ExecutionDriver::acquire_permit");
-            limit.acquire_owned().await.unwrap()
+        let limit = if certificate
+            .data()
+            .transaction_data()
+            .kind()
+            .mutates_implicitly_read_system_object()
```
mystenmark, line 99: i see that this is more precise than simply checking if it is a system tx, but is this necessary?

crates/sui-core/src/execution_cache/writeback_cache.rs
```diff
@@ -495,6 +497,48 @@ macro_rules! check_cache_entry_by_latest {
...
+        self.object_notify_read.read_one_blocking(
+            "load_implicitly_read_system_object",
+            &key,
+            |_key| self.object_exists_by_key(object_id, version).then_some(()),
+        );
+        self.metrics
+            .implicit_system_object_read_wait_latency
+            .with_label_values(&[object_id.to_string().as_str()])
+            .observe(wait_start.elapsed().as_secs_f64());
+        ObjectCacheRead::get_object_by_key(self, object_id, version).unwrap_or_else(|| {
```
mystenmark: why can't we use the return value of `read_one_blocking` above?

lxfind: `NotifyRead` returns `()`, instead of the value. To make this return the value there are quite some cascading interface changes. I can make that, but maybe not in this PR.

crates/sui-core/src/execution_driver.rs
lxfind: It will become necessary when we come to forwarding addresses, where all mutations are user transactions.

mystenmark (approved):

crates/sui-core/src/execution_driver.rs
```diff
@@ -121,14 +113,38 @@ pub async fn execution_process(
         let blocking_span = execution_span.clone();
```
lxfind, line 112: **Releasing the execution permit does not free the blocking thread; writer can be starved** -- Confidence: 80/100 | Importance: 75/100

(found by: Codex, Claude)
- **File**: `crates/sui-core/src/execution_driver.rs` lines 113-166 (wait site in `crates/sui-core/src/execution_cache/writeback_cache.rs` lines 525-529)
- **Problem**: When a reader blocks in `read_one_blocking`, it drops its semaphore permit but stays parked on a `spawn_blocking` thread. The freed permit lets the dispatch loop admit another reader, which can park too. The number of parked threads is therefore bounded by the number of dispatched readers, not by `execution_concurrency`. `SuiRuntimes` does not set `max_blocking_threads`, so Tokio's default of 512 applies. If the pool fills with parked readers waiting for accumulator version V, the settlement transaction that produces V cannot get a blocking thread. Its dedicated semaphore does not help because the bottleneck is OS threads, not permits.
- **Impact**: Execution deadlock on a validator or fullnode. Mitigating factor noted by Claude: the scheduler already gates coin-reservation transactions on the assigned accumulator root version, so the first consumer in #27524 should rarely reach the blocking path. The risk is a future consumer not gated by the scheduler, or a large catch-up backlog on a syncing fullnode. The stated design goal, that writers are never queued behind readers, is not actually achieved.
- **Suggestion**: Simplest fix: keep the normal-pool permit held while blocked. The dedicated writer semaphore already guarantees writers can obtain a permit, and holding the reader permit bounds parked threads to `execution_concurrency`, well below the blocking-pool limit. Alternatively, add a separate occupancy semaphore sized below Tokio's blocking-thread limit that a reader keeps while parked, and reserve capacity for writers. Either way, consider generalizing the scheduler gate so any transaction that may implicitly read a system object waits for the assigned version before dispatch, which keeps blocking as a rare fallback. Add a test with a deliberately small blocking pool that parks every allowed reader and proves a writer still starts.

---

crates/sui-replay/src/replay.rs
```diff
@@ -802,7 +802,7 @@ impl LocalExec {
                 &tx_info.executed_epoch,
                 tx_info.epoch_start_timestamp,
                 checked_input_objects,
-                std::collections::BTreeMap::new(),
+                sui_types::base_types::SystemObjectVersions::default(),
```
lxfind: **Legacy `sui-replay` not adapted to implicit reads** -- Confidence: 100/100 | Importance: 55/100

(found by: Codex, Claude)
- **File**: `crates/sui-replay/src/replay.rs` lines 805 and 1002; also `crates/sui-json-rpc-types/src/sui_transaction.rs` line 1064
- **Problem**: Both legacy replay execution paths pass `SystemObjectVersions::default()`. In `Exact` mode the temporary store unwraps the accumulator version, so replaying any transaction that performs an implicit read panics. Separately, JSON-RPC `shared_objects` is now derived from all accessed consensus objects, and legacy replay still copies every entry into `shared_object_refs` and asserts the count equals the transaction's declared shared inputs. An effects-only accumulator-root entry breaks that assertion. The PR description says legacy replay was updated, but the diff only swaps the empty map for the empty struct.
- **Impact**: Inert until a consumer lands in #27524, then legacy replay of withdrawal transactions fails the input-count assertion or panics. `sui-replay-2` is correct.
- **Suggestion**: Filter JSON `shared_objects` against the transaction's declared shared-input IDs when building replay inputs. Build `SystemObjectVersions` via `from_effects` at both call sites, mirroring `sui-replay-2` and the backtest tool. If legacy replay is not meant to support implicit reads, say so in the PR and fail with a clear error instead of a panic.

---

crates/sui-types/src/storage/object_store_trait.rs
```diff
@@ -13,6 +13,25 @@ pub trait ObjectStore {
...
+        version: ConsensusObjectVersion,
+    ) -> Object {
+        self.get_object_by_key(object_id, version.version)
+            .unwrap_or_else(|| {
+                panic!(
+                    "system object {object_id} not found at required version {}",
+                    version.version
+                )
+            })
+    }
```
lxfind, line 28: **`ObjectStore::load_implicitly_read_system_object` leaks execution policy into generic storage, and its doc contradicts the default impl** -- Confidence: 88/100 | Importance: 55/100

(found by: Codex; doc mismatch also found by Claude)
- **File**: `crates/sui-types/src/storage/object_store_trait.rs` lines 16-33
- **Problem**: The generic `ObjectStore` trait now carries an execution-specific concept with two very different behaviors behind one method: the default does an exact lookup and panics on a miss, while the writeback-cache override can block and indirectly release an execution-driver permit. The doc comment says other stores "can fall back to the latest version to ensure a result is always available", but the default implementation panics instead, and exact-version reads are consensus-critical so a latest-version fallback would be non-deterministic.
- **Impact**: Crosses the abstraction boundary between storage and execution scheduling, contrary to the repo's no-abstraction-leaks rule. Future store implementors get misleading guidance and could introduce non-deterministic reads in a context that requires an exact version.
- **Suggestion**: Move the operation onto a narrow execution-facing resolver with separate explicit exact and latest APIs. Keep blocking and permit coordination in the `sui-core` adapter. `TemporaryStore` already knows whether it needs `Exact` or `Latest`, so that policy does not belong in generic object storage. At minimum, reword the doc to match the panic-on-miss behavior.

---

lxfind (commented):

---

PR #27524, opened 2026-07-31
Check object funds withdraw sufficiency during execution

lxfind:
## Description

Second half of the stacked object-funds change, on top of #27523: object-funds withdrawal sufficiency is checked inside the Move VM at withdrawal time instead of by the post-execution `ObjectFundsChecker` wait/re-execute path. Protocol v137 enables `check_object_funds_withdraw_in_execution` on devnet only.

The native reads the settled object balance through the temporary store at the transaction accumulator version, relying on #27523 to block the implicit system-object read when that version is not yet local, and discounts unsettled withdrawals from earlier transactions in the same consensus commit. Insufficiency is a deterministic framework abort surfaced as `MoveAbort(funds_accumulator, E_OBJECT_FUNDS_INSUFFICIENT)`.

The first settled-balance read for each `(owner, type)` pair is charged a fixed 184 internal gas units. This matches the fixed work of the equivalent dynamic-field lookup: 52 for hashing, 52 for borrowing, and 80 for loading the fixed-size accumulator field contents. The existing transaction-local `ObjectFundsAvailable` state records when that balance has been queried, so later withdrawals for the same pair reuse the running balance and are not charged for another cold read. Withdrawals fully covered by deposits earlier in the same transaction also avoid the store read and its charge.

Known caveat carried over from the old PR: isolated replay does not reconstruct in-commit unsettled withdrawals (`unsettled = 0`; see the TODO in `crates/sui-replay-2/src/execution.rs`), which can diverge from the original execution. Mainnet enablement is blocked on resolving that.

## Test plan

- Successful and insufficient-funds gRPC simulations return the accumulator root object at exactly the version referenced by `ReadOnlyRoot` in their effects. Both cases fail without implicit-read tracking and pass with the fix.

- Cold-read metering follows the actual storage-access decision for each `(owner, type)` pair. A zero withdrawal or a withdrawal covered by transaction-local funds needs no settled-balance read. The first withdrawal that exceeds those funds takes the cold path and is charged 184; after that balance is marked queried, later withdrawals reuse the running balance and do not take or pay for another cold read, including when the cached balance is insufficient. Protocol snapshots verify that the cost is absent through v136 and set to 184 in v137.
- A vault funded with 1,000 units accepts withdrawals within its available balance, including zero and the exact remaining balance. A 5,000-unit withdrawal and the second of two 600-unit withdrawals fail specifically with `E_OBJECT_FUNDS_INSUFFICIENT`. The same boundary is verified for both simulation and committed execution.
- Deposits earlier in the same transaction are immediately available to later withdrawals: depositing 500 before withdrawing 1,400 from a 1,000-unit vault succeeds, while withdrawing 1,600 fails. Depositing 700 into an empty vault and then withdrawing 600 succeeds and leaves a net balance of 100. This verifies command ordering and netting within one execution.
- When two withdrawals use the same accumulator-root version in one consensus commit, unsettled withdrawals are deducted from the balance visible to later execution. From a settled balance of 1,000, two 600-unit withdrawals produce exactly one success and one insufficiency abort; two 400-unit withdrawals both succeed; withdrawing the exact remaining 200 then succeeds. A newly started fullnode catches up all resulting effects, exercising exact-version reads during replay and catch-up.
- With the feature enabled, a successful withdrawal records the accumulator root as a `ReadOnlyRoot` dependency. With it disabled, the same withdrawal succeeds without that dependency and continues through the legacy object-funds checker path, preserving pre-upgrade effects behavior.
- Regression cases attempt unbacked `u64::MAX` SUI and custom-coin withdrawals combined with gas smashing, gas refunds, accumulator merges, and gas-coin merges. Each transaction aborts at the object-funds check before poison accumulator writes are emitted; no value is credited, and the network can still reconfigure afterward. Each regression also runs with the new check explicitly disabled, preserving the `CoinBalanceOverflow` and custom-coin merge-cap assertions for the legacy path still used by mainnet and testnet.
- Concurrent stress covers owned, party-owned, and shared vaults while seven workers perform randomized transfers and expected overdraw attempts. All workers complete and the network reconfigures afterward, exercising the execution-time check under contention across ownership modes.

## Release notes

- [x] Protocol: Protocol v137 enables `check_object_funds_withdraw_in_execution` on devnet only. Object-funds withdrawal sufficiency is checked inside the Move VM at the point of withdrawal, insufficient withdrawals abort with a `sui::funds_accumulator` Move abort, and the first settled-balance read for each `(owner, type)` pair costs 184 internal gas units.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -209,13 +223,55 @@ impl<'a> ObjectRuntime<'a> {
...
+        amount: U256,
+    ) -> ObjectFundsSufficiency {
+        let key = (owner.into(), type_.clone());
+        let mut entry = self
+            .state
+            .object_funds_available
+            .get(&key)
+            .copied()
+            .unwrap_or_default();
+        if entry.available < amount && !entry.queried {
```
tnowacki: Per discussion, we might to charge more on query, if you look at the dynamic field caching code, you can see how we have done this in other cases (we more or less just need to return a second return value with that information)

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -209,13 +255,69 @@ impl<'a> ObjectRuntime<'a> {
...
+            {
+                Ok(balance) => balance,
+                Err(e) => {
+                    return ObjectFundsSufficiency::LoadError(e.to_string());
+                }
+            };
+            let Some(available) = entry.available.checked_add(U256::from(settled_available)) else {
+                return ObjectFundsSufficiency::LoadError(
+                    "object funds available balance overflow".to_string(),
+                );
```
tnowacki: I think we should just return a Move error here for overflow. If we think such a thing is safe to do. 
This will just prevent potential edge cases if/when we ever open up funds_accumulators for non-balance cases. (just for larger integer types it becomes easier to hit this case, depending on what we allow)

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -119,6 +120,51 @@ pub(crate) struct ObjectRuntimeState {
...
+}
+
+impl ObjectFundsAvailable {
+    fn needs_store_read(&self, amount: U256) -> bool {
+        self.available < amount && !self.queried
+    }
+}
+
+#[cfg(test)]
+mod object_funds_available_tests {
```
tnowacki: Nit I really don't like tests like this just sitting in the middle of the file.
If we need these, let's make a unit test module for the object runtime

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -119,6 +120,51 @@ pub(crate) struct ObjectRuntimeState {
...
+    /// amount that deposits increase and reservations decrease.
+    /// We lazily query the store the first time when the available balance is not sufficient
+    /// to serve a reservation. In case where we first deposit into an object, and then reserve
+    /// a smaller amount, we will not need to query the store at all.
+    available: U256,
+    /// Whether a query to the store has been made.
+    queried: bool,
+}
+
+impl ObjectFundsAvailable {
```
tnowacki: Nit: This file separates all of the type declarations from the impls for a very particular type of code navigation. 
So this should be moved above `ObjectRuntimeState` (at least immediately above it, potentially farther), and the `impl` block should be moved below into the relative position

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -25,6 +29,11 @@ use crate::{
 const E_OVERFLOW: u64 = 0;
 const E_ADDRESS_BALANCE_NOT_ENABLED: u64 = 1;
 
+#[derive(Clone)]
+pub struct ReserveObjectFundsForWithdrawalCostParams {
+    pub cold_read_cost: Option<InternalGas>,
+}
```
tnowacki, line 42: While the function isn't public, there should be gas parameters for this function in general so we have a knob to turn, even if that value is just `1`

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -134,3 +143,60 @@ pub fn withdraw_from_accumulator_address(
...
+        let cold_read_cost = context
+            .extensions()
+            .get::<NativesCostTable>()?
+            .reserve_object_funds_for_withdrawal_cost_params
+            .cold_read_cost
+            .ok_or_else(|| {
+                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR).with_message(
+                    "reserve_object_funds_for_withdrawal cold read gas cost is not set".to_string(),
+                )
+            })?;
```
tnowacki: The option should be resolved outside of here, so it shouldn't be possible to hit this at all (see table generation in sui-execution/latest/sui-move-natives/src/lib.rs)

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -134,3 +143,60 @@ pub fn withdraw_from_accumulator_address(
...
+    }
+
+    let obj_runtime: &mut ObjectRuntime = context.extensions_mut().get_mut()?;
+    match obj_runtime.check_object_funds_sufficiency(owner, &ty_tag, limit) {
+        ObjectFundsSufficiency::Sufficient => Ok(NativeResult::ok(context.gas_used(), smallvec![])),
+        ObjectFundsSufficiency::Insufficient => Ok(NativeResult::err(
+            context.gas_used(),
+            E_OBJECT_FUNDS_INSUFFICIENT,
+        )),
+        ObjectFundsSufficiency::LoadError(msg) => Err(PartialVMError::new(
```
tnowacki, line 223: If it simplifies things, it is fine if this is just an invariant violation passed back from the function, we do similar things with other object_runtime API (the similar thing being that any `Err` returned is always an invariant violation)

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -1067,6 +1078,30 @@ impl RuntimeObjectResolver for TemporaryStore<'_> {
...
+        let settled = AccumulatorRootValue::load(self, Some(required_version), owner, type_)?
+            .and_then(|value| value.as_u128())
+            .unwrap_or(0);
+
+        let unsettled = match self.unsettled_object_funds {
+            Some(reader) => reader.get_unsettled_object_withdraw(
+                &AccumulatorRootValue::get_field_id(owner, type_)?,
+                required_version,
+            ),
+            None => 0,
```
tnowacki: Under what cases is this `None`

crates/sui-types/src/funds_accumulator.rs
```diff
@@ -23,6 +23,19 @@ pub const RESOLVED_WITHDRAWAL_STRUCT: (&AccountAddress, &IdentStr, &IdentStr) =
 );
 pub const WITHDRAWAL_OWNER_FUNC_NAME: &IdentStr = ident_str!("withdrawal_owner");
 pub const WITHDRAWAL_SPLIT_FUNC_NAME: &IdentStr = ident_str!("withdrawal_split");
+pub const E_OBJECT_FUNDS_INSUFFICIENT: u64 = 2;
+
+pub fn is_object_funds_insufficient_abort(
+    status: &crate::execution_status::ExecutionFailureStatus,
+) -> bool {
```
tnowacki, line 30: Do we still need this? I remember it being a thing in earlier versions of the PR

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4569,6 +4580,14 @@ impl ProtocolConfig {
                     cfg.package_original_package_id_impl_cost_per_byte =
                         Some(package_read_cost_per_byte);
                 }
+                134 => {
+                    if chain != Chain::Mainnet && chain != Chain::Testnet {
+                        cfg.feature_flags.check_object_funds_withdraw_in_execution = true;
+                    }
+                    // Equivalent to the fixed portion of a dynamic-field lookup (52 + 52) plus
+                    // loading the 80-byte accumulator field contents at one gas unit per byte.
```
tnowacki, line 4731: Where is the per byte configuration?

crates/sui-framework/packages/sui-framework/sources/funds_accumulator.move
```diff
@@ -79,10 +79,14 @@ public(package) fun redeem<T: store>(withdrawal: Withdrawal<T>, _: internal::Per
     withdraw_impl(owner, value)
 }
 
-// Allows for creating a withdrawal from an object
-// (note no internal check needed since this will be gated at redemption)
-// Does not abort even if the value is greater than the amount in the object, unless we keep track
-// at each withdraw from object, we need to check it again at redeem so this seems fine?
+// Reserves funds from an object for a given limit.
```
tnowacki, line 91: We should add comments to balance.move and coin.move that gives the condition under which object withdrawals fail, e.g. `/// Aborts if the withdrawal specifies more funds than are currently available`

crates/sui-core/src/accumulators/design_docs/object_funds_checking.md
```diff
@@ -25,7 +35,7 @@ must be re-executed in a way that produces a deterministic failure.
 
 ```
   ┌───────────────┐     ┌───────────────────┐     ┌────────────────────┐
-  │ Execute       │     │ ObjectFundsChecker │     │ Commit Effects     │
+  │ Execute       │     │ ObjectFundsCheckerDEPRECATED │     │ Commit Effects     │
```
tnowacki: nit, the cute little diagram/box is borked

crates/sui-adapter-transactional-tests/tests/address_balances/object_funds_withdraw.move
```diff
@@ -0,0 +1,93 @@
...
+public fun withdraw_to(vault: &mut Vault, amount: u64, recipient: address) {
+    let w = balance::withdraw_funds_from_object<SUI>(&mut vault.id, amount);
+    let bal = balance::redeem_funds<SUI>(w);
+    balance::send_funds<SUI>(bal, recipient);
+}
+
+//# run test::obj_vault::new --sender A
+
+//# programmable --sender A --inputs 1000 object(2,0)
+//> 0: SplitCoins(Gas, [Input(0)]);
```
tnowacki, line 43: nit, could you add one line comment between the `//#` and the `//>` describing that is expected? It makes it easier to review the `.snap`

I'm also going to take a look at some point at improving the comment pairing

crates/sui-benchmark/src/lib.rs
```diff
@@ -227,26 +227,21 @@ impl ExecutionEffects {
...
+        matches!(
+            status,
+            ExecutionStatus::Failure(ExecutionFailure {
+                error: ExecutionErrorKind::InsufficientFundsForWithdraw,
+                ..
+            })
+        ) || matches!(
+            status,
+            ExecutionStatus::Failure(ExecutionFailure { error, .. })
+                if sui_types::funds_accumulator::is_object_funds_insufficient_abort(error)
```
tnowacki, line 243: How is this information used?

crates/sui-types/src/funds_accumulator.rs
lxfind: This is a helper for metrics and tests

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -134,3 +143,60 @@ pub fn withdraw_from_accumulator_address(
...
+
+    // Before check_object_funds_withdraw_in_execution is enabled, object funds withdrawals are not
+    // checked in execution. They are checked post-execution, and potentially retried if insufficient.
+    let (check_in_execution, needs_cold_read) = {
+        let obj_runtime: &ObjectRuntime = context.extensions().get()?;
+        let check_in_execution = obj_runtime
+            .protocol_config
+            .check_object_funds_withdraw_in_execution();
+        let needs_cold_read = check_in_execution
+            && obj_runtime.object_funds_sufficiency_needs_store_read(owner, &ty_tag, limit);
```
mystenmark: nit: the function uses the terminology `store_read` while the variable calls it `cold_read`.

I think `store_read` is better since a read from the store may or may not be cold

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -119,6 +120,51 @@ pub(crate) struct ObjectRuntimeState {
...
+}
+
+#[derive(Clone, Copy, Default)]
+struct ObjectFundsAvailable {
+    /// Current known available balance for this object balance account: a single running
+    /// amount that deposits increase and reservations decrease.
+    /// We lazily query the store the first time when the available balance is not sufficient
+    /// to serve a reservation. In case where we first deposit into an object, and then reserve
+    /// a smaller amount, we will not need to query the store at all.
+    available: U256,
```
mystenmark: why u256?

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -119,6 +120,51 @@ pub(crate) struct ObjectRuntimeState {
     settlement_output_sui: u64,
     accumulator_merge_totals: BTreeMap<(AccountAddress, TypeTag), u128>,
     accumulator_split_totals: BTreeMap<(AccountAddress, TypeTag), u128>,
+    object_funds_available: BTreeMap<(AccountAddress, TypeTag), ObjectFundsAvailable>,
+}
+
+#[derive(Clone, Copy, Default)]
+struct ObjectFundsAvailable {
```
mystenmark: suggestion for making this structure clearer:

```
struct ObjectFundsAvailable {
   pending_available: U256,
   settled_available: Option<U256>,
}
```

doing it this way would avoid the need to sum settled and pending balances into a single field, which seems nice.

i don't feel very strongly about this

crates/sui-protocol-config/src/lib.rs
lxfind: The accumulator account is of fixed size, so I don't think we need a per byte configuration. We should just always use a fixed total cost for each load for simplicity.

crates/sui-benchmark/src/lib.rs
lxfind: This is used for metric purpose so that when we benchmark things we know that we have covered this case. I added a TODO there to distinguish between address / object funds insufficiency so that we have coverage for both.

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
lxfind: A `U256` `pending_available` won't be able to represent pending withdraws though.

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
lxfind: We probably don't need u256 for object balance today. But I cannot say that for sure in the future when we have some kind of generic object funds?

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
lxfind: I checked a few others and this seems to be the universal way everywhere else (i.e. unwrap with an invariant violation upon acccess)

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
lxfind: I think returning an enum and make error decisions in the end is easier to reason about: we won't mix invariant violations with other errors such as insufficient funds or overflow.

crates/sui-framework/packages/sui-framework/sources/coin.move
```diff
@@ -165,6 +165,8 @@ public fun put<T>(balance: &mut Balance<T>, coin: Coin<T>) {
 // === Address Balance <-> Coin utility functions ===
 
 /// Redeem a `Withdrawal<Balance<T>>` and create a `Coin<T>` from the withdrawn Balance<T>.
+/// For an object withdrawal, the transaction aborts if the withdrawal specifies more funds than
+/// are currently available to the object.
```
tnowacki: Could you change this to match the style of other aborts messages? 
Two good examples:
https://github.com/MystenLabs/sui/blob/b13595fd7a99a501ddf793fcad099c2ac060d013/crates/sui-framework/packages/sui-framework/sources/dynamic_field.move#L37
https://github.com/MystenLabs/sui/blob/b13595fd7a99a501ddf793fcad099c2ac060d013/crates/sui-framework/packages/sui-framework/sources/coin.move#L324

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -93,6 +95,18 @@ pub struct RuntimeResults {
     pub settlement_output_sui: u64,
 }
 
+#[derive(Clone, Copy, Default)]
+struct ObjectFundsAvailable {
+    /// Current known available balance for this object balance account: a single running
+    /// amount that deposits increase and reservations decrease.
+    /// We lazily query the store the first time when the available balance is not sufficient
+    /// to serve a reservation. In case where we first deposit into an object, and then reserve
+    /// a smaller amount, we will not need to query the store at all.
```
tnowacki: 🤖 
Could you make this  comment a bit more succinct? Or just generally readable.
Maybe
```
Current known available balance. 
Keeps a running total available based on sends/redeems in this transaction. 
The first time an insufficient balance is reached, we must then query the object store.
```
(sorry I'm looking at the specific commit and suggestions aren't available)

crates/sui-execution-backtest/src/execute.rs
```diff
@@ -481,6 +481,9 @@ fn run_execution(
             ctx.epoch_start_timestamp_ms,
             input_objects,
             system_object_versions,
+            // TODO: When backtesting object funds withdraws, we will need unsettled balance
+            // information to reproduce the original execution.
+            &sui_types::accumulator_root::EmptyUnsettledObjectFunds,
```
tzakian, line 486: Yea -- this is a known limitation with the execution backtest (basically we don't recreate IFFW errors). Would be nice to fix at some point, but not super important.

crates/sui-framework-snapshot/bytecode_snapshot/134/0x0000000000000000000000000000000000000000000000000000000000000002
tzakian, line 1: My guess is this was added for testing reasons, but this and the other change under this crate should be backed out before landing.

crates/sui-framework-snapshot/manifest.json
```diff
@@ -2850,7 +2850,7 @@
     ]
   },
   "134": {
-    "git_revision": "2a0becb2fcc6989e492981104af67f62f2c9511a",
+    "git_revision": "1de968fbc01db97da5d50c00ffeccb6a46b03f18",
```
tzakian: ditto here.

crates/sui-types/src/storage/mod.rs
```diff
@@ -198,6 +207,10 @@ pub trait RuntimeObjectResolver: BackingPackageStore {
         child_version_upper_bound: SequenceNumber,
     ) -> SuiResult<Option<Object>>;
 
+    fn object_available_balance(&self, _owner: SuiAddress, _type_: &TypeTag) -> SuiResult<u128> {
+        Ok(u128::MAX)
```
tzakian: This feels like a very dangerous blanket definition for this function. 

Thoughts on not providing a default implementation for this and/or making it zero instead of `u128::MAX`?

external-crates/move/crates/move-vm-runtime/src/runtime/telemetry.rs
```diff
@@ -612,7 +612,13 @@ impl TransactionTelemetryContext {
...
 impl Drop for Timer {
     fn drop(&mut self) {
-        if !self.reported {
+        // Skip the assertion while unwinding. Execution can be unwound through the VM by a
+        // panic raised outside of it (e.g. under the simulator, a parked blocking thread is
+        // aborted via `panic_any` when its node is killed), which skips the `report_time`
+        // call that would have marked the timer reported. Asserting here would then be a
+        // panic inside a destructor during cleanup, which Rust escalates to a non-unwinding
+        // panic and aborts the process - turning an orderly unwind into a hard crash.
+        if !self.reported && !std::thread::panicking() {
```
tzakian, line 621: This looks fine to me, but cc @cgswords for viz

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -209,13 +231,67 @@ impl<'a> ObjectRuntime<'a> {
...
+        type_: &TypeTag,
+        amount: U256,
+    ) -> ObjectFundsSufficiency {
+        let key = (owner.into(), type_.clone());
+        let mut entry = self
+            .state
+            .object_funds_available
+            .get(&key)
+            .copied()
+            .unwrap_or_default();
```
tzakian: I really dislike `Default`, and I even more dislike it where relying on default values (and them not being what we expect) could be catastrophic.

Could you either:
* Remove default and make what the "empty object funds available" is an explicit function (and feel free to clippy-away the stupid clippy message that will probably raise); or
* If you really prefer to keep default, add a small little unit test that just asserts the `ObjectFundsAvailable::default()` produces the values that we expect.

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -209,13 +231,67 @@ impl<'a> ObjectRuntime<'a> {
...
     }
 
+    pub fn check_object_funds_sufficiency(
+        &mut self,
+        owner: SuiAddress,
+        type_: &TypeTag,
+        amount: U256,
+    ) -> ObjectFundsSufficiency {
+        let key = (owner.into(), type_.clone());
+        let mut entry = self
```
tzakian: why not do this over an actual `Entry` from the entry API? Then you don't need to do the replacing `insert`ion at 277?

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -209,13 +231,67 @@ impl<'a> ObjectRuntime<'a> {
...
+        &self,
+        owner: SuiAddress,
+        type_: &TypeTag,
+        amount: U256,
+    ) -> bool {
+        self.state
+            .object_funds_available
+            .get(&(owner.into(), type_.clone()))
+            .copied()
+            .unwrap_or_default()
```
tzakian: see above rant on `Default` :p

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -442,6 +518,23 @@ impl<'a> ObjectRuntime<'a> {
...
                     self.state.accumulator_merge_totals.insert(key, new_total);
+                    if self
+                        .protocol_config
+                        .check_object_funds_withdraw_in_execution()
+                    {
+                        let entry = self
+                            .state
+                            .object_funds_available
+                            .entry((target_addr, target_ty.clone()))
+                            .or_default();
```
tzakian: Image

sui-execution/latest/sui-move-natives/src/object_runtime/unit_tests.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use super::ObjectFundsAvailable;
+use move_core_types::u256::U256;
+
+#[test]
+fn store_read_is_needed_only_when_uncached_balance_is_insufficient() {
+    let empty = ObjectFundsAvailable::default();
```
tzakian: if you wanted to keep Default, I would assert the values of `empty` here.

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
tzakian: +1 we should at the very least have a base cost that we charge, even if that is nominal.

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -146,3 +158,62 @@ pub fn withdraw_from_accumulator_address(
...
+
+pub fn reserve_object_funds_for_withdrawal(
+    context: &mut NativeContext,
+    mut ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(ty_args.len() == 1);
+    debug_assert!(args.len() == 2);
+
+    let ty_tag = context.type_to_type_tag(&safe_unwrap!(ty_args.pop()))?;
```
tzakian: We should have a base charge that we always charge, whether a cold read is needed or not. 

It should be enough to cover the work needed before determining if a cold read is required, and I'd probably base it off of/use the same value as the dynamic field base cost.

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -146,3 +158,62 @@ pub fn withdraw_from_accumulator_address(
...
     Ok(NativeResult::ok(context.gas_used(), smallvec![withdrawn]))
 }
+
+pub fn reserve_object_funds_for_withdrawal(
+    context: &mut NativeContext,
+    mut ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(ty_args.len() == 1);
+    debug_assert!(args.len() == 2);
```
tzakian: I know this doesn't follow the conventions of the other functions, but I want us to move over to using `safe_assert!` for these instead, so if you could use those here that would be amazing.

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -146,3 +158,62 @@ pub fn withdraw_from_accumulator_address(
...
+    let needs_cold_read = {
+        let obj_runtime: &ObjectRuntime = context.extensions().get()?;
+        if !obj_runtime
+            .protocol_config
+            .check_object_funds_withdraw_in_execution()
+        {
+            // Before check_object_funds_withdraw_in_execution is enabled, object funds withdrawals are not
+            // checked in execution. They are checked post-execution, and potentially retried if insufficient.
+            return Ok(NativeResult::ok(context.gas_used(), smallvec![]));
+        }
```
tzakian, line 182: IMO I'd probably move this protocol check up to the top of this function, (just after the asserts, but before the type tag conversion). Then you can charge base gas immediately before doing any further work.

crates/sui-types/src/storage/mod.rs
lxfind: I will fix this properly once https://github.com/MystenLabs/sui/pull/27804 merges.

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -146,3 +161,68 @@ pub fn withdraw_from_accumulator_address(
...
+        if obj_runtime.object_funds_sufficiency_needs_store_read(owner, &ty_tag, limit) {
+            let cold_read_cost = cost_params.cold_read_cost.ok_or_else(|| {
+                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR).with_message(
+                    "reserve_object_funds_for_withdrawal cold read gas cost is not set".to_string(),
+                )
+            })?;
+            native_charge_gas_early_exit!(context, cold_read_cost);
+        }
+        (ty_tag, limit, owner)
+    };
```
lxfind, line 213: **`reserve_object_funds_for_withdrawal` omits `check_accumulator_type_bounds`** -- Confidence: 92/100 | Importance: 55/100

(found by: Claude, Codex)
- **File**: `sui-execution/latest/sui-move-natives/src/funds_accumulator.rs` lines 196-213
- **Problem**: The sibling natives `add_to_accumulator_address` / `withdraw_from_accumulator_address` call `check_accumulator_type_bounds` (abort `E_ACCUMULATOR_TYPE_TOO_LARGE`, bound `max_accumulator_type_nodes`, introduced in v133 as an accumulator DoS mitigation) before touching accumulator state. The new reservation native goes straight from `type_to_type_tag` to `get_field_id` (BCS + hash of the full `TypeTag`) and a child-object store read for a fixed 184 gas, regardless of type node count. Because `Withdrawal` has `drop`, a caller can discard it, so the redeem native that does enforce the bound need never run.
- **Impact**: Types exceeding the 16-node accumulator limit reach type-dependent hashing and a storage lookup while paying a fixed charge, bypassing an explicit protocol bound. The practical exposure is limited — work is still bounded by the general `max_type_nodes` VM limit, the field lookup for an oversized type will miss (available = 0) and abort with `E_OBJECT_FUNDS_INSUFFICIENT` anyway, and the flag is off on mainnet/testnet — but the abort semantics are inconsistent (type-bound error vs insufficient-funds error depending on path) and the invariant should hold before the flag is widened.
- **Suggestion**: Right after computing `ty_tag`, call `check_accumulator_type_bounds(protocol_config, &ty_tag)` and return `E_ACCUMULATOR_TYPE_TOO_LARGE` before `object_funds_sufficiency_needs_store_read` / the cold-read charge. Add a test with an oversized type covering both discarded and redeemed withdrawals, asserting no root read and no cold-read charge.

---

crates/sui-e2e-tests/tests/address_balance_tests.rs
```diff
@@ -2162,23 +2162,51 @@ async fn test_multiple_deposits_merged_in_effects() {
...
+}
+
+fn object_funds_in_execution_test_env() -> TestEnvBuilder {
+    TestEnvBuilder::new().with_proto_override_cb(Box::new(|_, mut cfg| {
+        cfg.set_enable_object_funds_withdraw_for_testing(true);
+        cfg.set_check_object_funds_withdraw_in_execution_for_testing(true);
+        cfg
+    }))
+}
+
```
lxfind, line 2199: **Overflow-guard e2e tests lost their flag-off coverage** -- Confidence: 90/100 | Importance: 55/100

(found by: Claude)
- **File**: `crates/sui-e2e-tests/tests/address_balance_tests.rs` lines 2192-2199 (and the five tests now using `object_funds_in_execution_test_env()`)
- **Problem**: `..._poison_pill`, `..._custom_coin_capped`, `..._conservation_overflow_single_withdrawal`, `..._gas_refund_poison_pill`, and `..._merge_into_gas_coin` were all converted to run with `check_object_funds_withdraw_in_execution = true`, where the first native withdraw aborts before any legacy guard is reached. `git grep CoinBalanceOverflow crates/sui-e2e-tests` now returns nothing: the per-key supply guard (`check_accumulator_amounts_representable`), the per-key `u64::MAX` merge cap, and the cross-key total-SUI bound have no e2e coverage with the flag off.
- **Impact**: The flag is off on mainnet and testnet in v135, so those guards remain the binding protection against exactly the poison-pill shapes these tests were written for. The guards still exist (the PR only adds a TODO) but are now untested on the configuration production actually runs.
- **Suggestion**: Keep the original flag-off variants — parameterize the env builder or duplicate each test with `set_check_object_funds_withdraw_in_execution_for_testing(false)` and the original `CoinBalanceOverflow` / `MovePrimitiveRuntimeError` assertions — until the flag is enabled everywhere and the guards are deleted.

---

lxfind (commented):
## Automated Review: 2 Actionable Finding(s)

2 issue(s) posted as inline comments on the relevant code.

---

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -209,13 +237,64 @@ impl<'a> ObjectRuntime<'a> {
...
+                }
+            };
+            let Some(available) = entry.available.checked_add(U256::from(settled_available)) else {
+                return ObjectFundsSufficiency::Overflow;
+            };
+            entry.available = available;
+            entry.queried = true;
+        }
+        if entry.available >= amount {
+            entry.available -= amount;
```
tnowacki, line 280: I'm surprised this didn't hit our clippy warnings (for not explicitly using checked/saturating/wrapping). But I suppose its fine due to the check above

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -442,6 +521,23 @@ impl<'a> ObjectRuntime<'a> {
...
+                        let entry = self
+                            .state
+                            .object_funds_available
+                            .entry((target_addr, target_ty.clone()))
+                            .or_insert_with(ObjectFundsAvailable::init);
+                        entry.available = entry
+                            .available
+                            .checked_add(U256::from(amount as u128))
+                            .ok_or_else(|| {
+                            PartialVMError::new(StatusCode::ARITHMETIC_ERROR)
```
tnowacki, line 542: Ah fuckin hell. Did not see this above either. So it is fine I suppose for now.
But in general, we should _not_ be doing this. Native calls need to through native aborts/errors. The only VMErrors returned should be invariant violations or limits. 

Something maybe worth fixing some day

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
tnowacki: Seems reasonable, though I don't think it really matters

crates/sui-protocol-config/src/lib.rs
```diff
@@ -4644,7 +4658,12 @@ impl ProtocolConfig {
 
                     if chain != Chain::Mainnet && chain != Chain::Testnet {
                         cfg.feature_flags.allowed_proposers = true;
+                        cfg.feature_flags.check_object_funds_withdraw_in_execution = true;
                     }
+                    cfg.reserve_object_funds_for_withdrawal_cost_base = Some(52);
+                    // Equivalent to the fixed portion of a dynamic-field lookup (52 + 52) plus
+                    // loading the 80-byte accumulator field contents at one gas unit per byte.
```
tnowacki, line 4731: 🎉

crates/sui-core/src/accumulators/unsettled_object_withdrawals.rs
```diff
@@ -80,6 +103,51 @@ impl UnsettledObjectWithdrawals {
...
+        chain_identifier: ChainIdentifier,
+    ) -> bool {
+        if accumulator_running_max_withdraws.is_empty() {
+            return false;
+        }
+        let address_funds_reservations: BTreeSet<_> = tx_data
+            .process_funds_withdrawals_for_execution(chain_identifier)
+            .into_keys()
+            .collect();
+        let updates: BTreeMap<_, _> = effects
```
mystenmark: there doesn't seem to be any reason to use a `BTreeMap` instead of a `Vec<(_, _)>` here

crates/sui-core/src/authority.rs
```diff
@@ -2115,20 +2122,62 @@ impl AuthorityState {
...
+                        let recorded = self
+                            .unsettled_object_withdrawals
+                            .record_object_funds_withdraws(
+                                certificate.transaction_data(),
+                                &effects,
+                                &inner_temp_store.accumulator_running_max_withdraws,
+                                accumulator_version,
+                                self.chain_identifier,
+                            );
+                        if recorded {
```
mystenmark: nit: i think we can just move this code inside `record_object_funds_withdraws` and get rid of the bool return val

crates/sui-core/src/authority.rs
```diff
@@ -3790,12 +3844,13 @@ impl AuthorityState {
     async fn init_object_funds_checker(&self) {
         let epoch_store = self.epoch_store.load();
         // TODO: Once we enable object funds checking during execution, we will no longer need to initialize the object funds checker here.
-        if self.node_role(&epoch_store).runs_consensus()
-            && epoch_store.protocol_config().enable_object_funds_withdraw()
-        {
+        let protocol_config = epoch_store.protocol_config();
+        let needs_checker = protocol_config.enable_object_funds_withdraw()
+            && self.node_role(&epoch_store).runs_consensus();
+        if needs_checker {
```
mystenmark: i think this edit is unnecessary and will just pollute blame

crates/sui-types/src/storage/mod.rs
mystenmark: looks like it landed, agreed that this needs to be fixed

external-crates/move/crates/move-vm-runtime/src/runtime/telemetry.rs
mystenmark: note that this is a no-op in release builds due to `panic=abort`

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -1066,6 +1077,27 @@ impl RuntimeObjectResolver for TemporaryStore<'_> {
...
+    /// checkpoint that have not settled yet.
+    /// This function is expected never to fail; an error indicates an invariant violation.
+    fn object_available_balance(&self, owner: SuiAddress, type_: &TypeTag) -> SuiResult<u128> {
+        let required_version = self
+            .load_implicitly_read_system_object(&SUI_ACCUMULATOR_ROOT_OBJECT_ID)
+            .version();
+
+        let settled = AccumulatorRootValue::load(self, Some(required_version), owner, type_)?
+            .and_then(|value| value.as_u128())
+            .unwrap_or(0);
```
mystenmark, line 1218: i don't understand how this is safe - what guarantees that the latest version exists?

lxfind: This is reading from the `required_version`, which is guaranteed to exist because we block read the root at this version above.

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
```diff
@@ -209,13 +237,64 @@ impl<'a> ObjectRuntime<'a> {
...
+        &self,
+        owner: SuiAddress,
+        type_: &TypeTag,
+        amount: U256,
+    ) -> bool {
+        self.state
+            .object_funds_available
+            .get(&(owner.into(), type_.clone()))
+            .copied()
+            .unwrap_or_else(ObjectFundsAvailable::init)
```
tzakian, line 297: Thank you! ❤️

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
tzakian: Should this just be an invariant violation? 

If you are able to exceed u256 max here that seems like it's breaking an invariant of funds accumulators, and not just an arithmetic error if that makes sense?

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
```diff
@@ -26,6 +35,12 @@ const E_OVERFLOW: u64 = 0;
 const E_ADDRESS_BALANCE_NOT_ENABLED: u64 = 1;
 const E_ACCUMULATOR_TYPE_TOO_LARGE: u64 = 4;
 
+#[derive(Clone)]
+pub struct ReserveObjectFundsForWithdrawalCostParams {
+    pub base_cost: Option<InternalGas>,
```
tzakian, line 40: Thank you! ❤️

sui-execution/latest/sui-move-natives/src/funds_accumulator.rs
tzakian: Same, but IMO worth adding just for paranoia 🤷🏻

lxfind (commented):
## Automated Review: 5 Actionable Finding(s)

5 issue(s) posted as inline comments on the relevant code.

---

crates/sui-replay/src/replay.rs
```diff
@@ -803,6 +803,11 @@ impl LocalExec {
                 tx_info.epoch_start_timestamp,
                 checked_input_objects,
                 sui_types::base_types::SystemObjectVersions::default(),
+                // TODO: Replaying a transaction that withdrew object funds needs the unsettled
+                // withdrawals that earlier transactions in the same consensus commit had
+                // accumulated at execution time. Until those are reconstructed, replaying such a
+                // transaction can diverge from the original execution.
+                &sui_types::accumulator_root::EmptyUnsettledObjectFunds,
```
lxfind: **Legacy replay panics on cold object-funds withdrawals** -- Confidence: 98/100 | Importance: 75/100

(found by: Codex)
- **File**: `crates/sui-replay/src/replay.rs` lines 805-810 and 1007-1012
- **Problem**: Both legacy replay paths still pass `SystemObjectVersions::default()` into execution. With the flag on, a withdrawal that is not covered by transaction-local deposits calls `load_implicitly_read_system_object` for the accumulator root, and the exact-version path unwraps the accumulator version. With a default `SystemObjectVersions` that version is absent, so execution panics. The new comment in replay only discusses the documented divergence from missing unsettled withdrawals, but execution never reaches that point.
- **Impact**: `sui-replay` and sandbox replay cannot replay ordinary v137 object withdrawals on any chain where the flag is on. That breaks an important debugging and recovery tool for exactly the class of transactions this PR introduces, and the breakage will spread to testnet and mainnet when the flag is enabled there.
- **Suggestion**: Reconstruct `SystemObjectVersions` from the expected effects' `ReadOnlyRoot` entry, following the `SystemObjectVersions::from_effects` pattern already used in `sui-replay-2`. If the RPC effects representation is insufficient for the sandbox, persist the required versions in sandbox state.

---

crates/sui-types/src/storage/mod.rs
```diff
@@ -198,6 +207,13 @@ pub trait RuntimeObjectResolver: BackingPackageStore {
         child_version_upper_bound: SequenceNumber,
     ) -> SuiResult<Option<Object>>;
 
+    fn object_available_balance(&self, _owner: SuiAddress, _type_: &TypeTag) -> SuiResult<u128> {
+        // TODO: It is dangerous to return u128::MAX as a default implementation.
+        // We will properly introduce a trait for this so that it is not too crazy to not have
+        // a default implementation.
+        Ok(u128::MAX)
+    }
```
lxfind: **Fail-open default on `RuntimeObjectResolver::object_available_balance`** -- Confidence: 100/100 | Importance: 70/100

(found by: Codex, Claude)
- **File**: `crates/sui-types/src/storage/mod.rs` lines 190-215, and the blanket impls at lines 537-610
- **Problem**: Object-funds semantics are added to the generic `RuntimeObjectResolver` trait with a default implementation that returns `u128::MAX`. Only `TemporaryStore` overrides it. The existing blanket impls for `Arc<S>`, `&S`, and `&mut S` forward `read_child_object` and `get_object_received_at_version` but were not updated to forward the new method, so a `TemporaryStore` reached through any of those wrappers gets the default. `ObjectFundsSufficiency` is also execution policy placed in the generic storage layer. The author's own TODO on the default acknowledges the danger, and this pattern conflicts with the repository's explicit "no abstraction leaks" requirement.
- **Impact**: Today the object runtime receives the state view directly, so the concrete impl is hit and behavior is correct. Any direct `ObjectRuntime` user, test harness, wrapper layer, or future execution path that forgets the override silently disables the sufficiency check and authorizes effectively unlimited withdrawals. There is no compile error, test failure, or log line when that happens.
- **Suggestion**: Forward `object_available_balance` in the three blanket impls now. Replace the `u128::MAX` default with something that cannot pass silently, such as returning an `ExecutionInvariantViolation` error or making it a required method on a small dedicated object-funds reader trait injected into `TemporaryStore` or `ObjectRuntime`, similar to `UnsettledObjectFundsRead`. Give the test-scenario store an explicit override. Keep `ObjectFundsSufficiency` local to the runtime and native layer.

---

sui-execution/src/latest.rs
```diff
@@ -212,6 +217,9 @@ impl executor::Executor for Executor {
                 store,
                 input_objects,
                 SystemObjectVersionRequirements::Latest,
+                // Dev-inspect and dry-run results are never committed, so they do not need to
+                // account for unsettled withdrawals from other transactions.
+                &EmptyUnsettledObjectFunds,
```
lxfind, line 222: **Simulation ignores same-version unsettled withdrawals** -- Confidence: 92/100 | Importance: 60/100

(found by: Codex)
- **File**: `sui-execution/src/latest.rs` lines 216-222 and 246-252
- **Problem**: Both simulation modes pass `EmptyUnsettledObjectFunds`. Being non-committing does not make unsettled withdrawals irrelevant: the latest accumulator root can still exclude earlier withdrawals that executed against that same root version and have not settled yet.
- **Impact**: A dry-run or simulation can report success after an earlier same-commit transaction has already consumed the available object balance, while committed execution correctly aborts. That weakens one of the stated motivations for moving the check into execution, namely honest simulation results without a second pass.
- **Suggestion**: Thread an unsettled reader through `dev_inspect_transaction` and the simulation APIs. Production fullnodes and simulacrum should pass their live `UnsettledObjectWithdrawals`. Isolated callers such as replay and genesis can deliberately keep the empty implementation.

---

sui-execution/latest/sui-move-natives/src/object_runtime/unit_tests.rs
```diff
@@ -0,0 +1,25 @@
...
+    };
+    assert!(!deposited.needs_store_read(U256::from(500u64)));
+    assert!(deposited.needs_store_read(U256::from(501u64)));
+
+    let queried = ObjectFundsAvailable {
+        available: U256::from(0u64),
+        queried: true,
+    };
+    assert!(!queried.needs_store_read(U256::from(1u64)));
+}
```
lxfind, line 25: **No test pins exact native gas charging** -- Confidence: 98/100 | Importance: 52/100

(found by: Codex)
- **File**: `sui-execution/latest/sui-move-natives/src/object_runtime/unit_tests.rs` lines 7-25
- **Problem**: The added unit test verifies only the read/no-read predicate. Protocol snapshots confirm the cold-read cost is configured as 184, but no test asserts that the native actually charges the base cost once and the cold-read cost exactly once on the first shortfall. The transactional snapshots round computation to the same 1,000,000-unit bucket and cannot detect a missing, repeated, or misplaced internal charge.
- **Impact**: A consensus-relevant gas-metering regression could pass the current suite and only appear as different out-of-gas behavior near a budget boundary.
- **Suggestion**: Add exact native gas-use tests covering zero withdrawal, transaction-local coverage, first cold read, a subsequent cached withdrawal, cached insufficiency, and pre-v137 behavior.

---

crates/sui-e2e-tests/tests/address_balance_tests.rs
```diff
@@ -2162,23 +2162,51 @@ async fn test_multiple_deposits_merged_in_effects() {
...
+/// Regression shape for an old accumulator overflow bug: an unbacked object-sourced `u64::MAX`
+/// withdrawal would be redeemed and then combined with an address-balance gas-smash Merge to the
+/// same `(sender, Balance<SUI>)` key. With in-execution object-funds checking, the native withdraw
+/// aborts before the poison accumulator write is emitted.
 #[sim_test]
-async fn test_accumulator_merge_overflow_poison_pill() {
-    let mut test_env = TestEnvBuilder::new().with_num_validators(1).build().await;
+async fn test_accumulator_merge_overflow_poison_pill_blocked_by_object_funds_check() {
+    let mut test_env = object_funds_in_execution_test_env()
```
lxfind, line 2206: **Flag-off poison-pill regression coverage was removed** -- Confidence: 85/100 | Importance: 55/100

(found by: Claude)
- **File**: `crates/sui-e2e-tests/tests/address_balance_tests.rs` lines 2192-2206
- **Problem**: The five poison-pill regression tests were rewritten to enable `check_object_funds_withdraw_in_execution` and assert the new native abort. Nothing now exercises the flag-off path where `check_accumulator_amounts_representable` and the per-key merge cap are the binding guards. A repo-wide search finds no remaining test that references `CoinBalanceOverflow`.
- **Impact**: Mainnet and testnet run v137 with the flag off, so the representability guard is the only protection on production networks for the exact transaction shapes these tests were written to lock in. That production path just lost all end-to-end coverage. The guard carries a TODO to remove it from future executor versions, which makes coverage more important while it is live, not less.
- **Suggestion**: Keep the original five tests with `set_check_object_funds_withdraw_in_execution_for_testing(false)` and their original `CoinBalanceOverflow` and arithmetic-error assertions, alongside the new flag-on variants.

---

sui-execution/latest/sui-move-natives/src/object_runtime/mod.rs
lxfind: I added a TODO at the top of this function, because it emits quite a few of these errors for similar reasons. I think if we want to allow larger values we would want to fix all of them properly.

mystenmark (approved):

crates/sui-types/src/funds_accumulator.rs
```diff
@@ -23,6 +23,19 @@ pub const RESOLVED_WITHDRAWAL_STRUCT: (&AccountAddress, &IdentStr, &IdentStr) =
 );
 pub const WITHDRAWAL_OWNER_FUNC_NAME: &IdentStr = ident_str!("withdrawal_owner");
 pub const WITHDRAWAL_SPLIT_FUNC_NAME: &IdentStr = ident_str!("withdrawal_split");
+pub const E_OBJECT_FUNDS_INSUFFICIENT: u64 = 2;
```
lxfind: **Insufficient-funds abort reuses code 2, colliding with `EOwnerMismatch`** -- Confidence: 100/100 | Importance: 60/100

(found by: Agy, Codex)
- **File**: `crates/sui-types/src/funds_accumulator.rs` line 26; emitted at `sui-execution/latest/sui-move-natives/src/funds_accumulator.rs` line 218; collides with `crates/sui-framework/packages/sui-framework/sources/funds_accumulator.move` line 28
- **Problem**: The new native aborts with `E_OBJECT_FUNDS_INSUFFICIENT`, a raw `u64` constant set to 2. The Move module `sui::funds_accumulator` already assigns explicit code 2 to `EOwnerMismatch` through an `#[error(code = 2)]` constant. The two raw abort values differ at the VM level because the Move constant is emitted in the packed clever-error representation, but JSON-RPC and GraphQL normalize that clever-error code back to 2. Both failures therefore surface to clients with the same module and error code identity.
- **Impact**: Clients, SDKs, and indexers that classify Move aborts by module and normalized code cannot tell an owner mismatch from insufficient object funds. A developer who hits the new abort and looks up code 2 in the module source finds the owner-mismatch constant and debugs the wrong problem. The stable numeric code introduced for this protocol behavior is undermined from the start. Agy additionally argued that genuine `EOwnerMismatch` aborts would trip `is_object_funds_insufficient_abort`, inflating the `in_execution_check_result` metric and misclassifying benchmark results. Codex's more detailed account indicates the raw values differ, so that path is unlikely to misfire and the confusion is confined to normalized representations. The flag is devnet-only today, so the code can still be changed cheaply.
- **Suggestion**: Allocate the next unused module code, currently 5, define a matching raw `u64` constant with an error string in `funds_accumulator.move`, update the Rust constant and the native, and add a test asserting that the normalized public representation of the two errors is distinct.

---

crates/simulacrum/src/epoch_state.rs
```diff
@@ -195,6 +206,19 @@ impl EpochState {
...
+                system_object_versions.get(&SUI_ACCUMULATOR_ROOT_OBJECT_ID)
+        {
+            self.unsettled_object_withdrawals
+                .record_object_funds_withdraws(
+                    transaction.transaction_data(),
+                    &effects,
+                    &inner_temp_store.accumulator_running_max_withdraws,
+                    accumulator_version.version,
+                    self.chain_identifier,
+                );
```
lxfind: **Simulacrum never clears settled entries from `UnsettledObjectWithdrawals`** -- Confidence: 99/100 | Importance: 52/100

(found by: Codex)
- **File**: `crates/simulacrum/src/epoch_state.rs` lines 209-220; checkpoint path at `crates/simulacrum/src/lib.rs` line 350
- **Problem**: Simulacrum's `EpochState` now records every successful positive object withdrawal into `UnsettledObjectWithdrawals`, but its checkpoint creation path never calls `commit_accumulator_versions`. Production performs that cleanup after checkpoint execution in `crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs` at line 431. Recording is also not gated on `check_object_funds_withdraw_in_execution`, so flag-off simulator runs accumulate entries that are never read.
- **Impact**: The two nested maps retain an entry for every withdrawn account and accumulator-version pair until the epoch state is replaced. Long-running simulation, synthetic ingestion, or fuzz workloads that create many checkpoints consume steadily increasing memory even though old versions can no longer affect execution. It also splits lifecycle ownership between `EpochState`, which owns the data, and `Simulacrum::create_checkpoint`, which performs settlement.
- **Suggestion**: After the settlement and barrier transactions for a simulated checkpoint execute, collect the settled root input versions and call `commit_accumulator_versions`, mirroring production. Gate recording on the feature flag and add a test asserting that checkpoint creation empties the settled entries.

---

lxfind (commented):
## Automated Review: 2 Actionable Finding(s)

2 issue(s) posted as inline comments on the relevant code.

---

crates/sui-framework-snapshot/manifest.json
```diff
@@ -2908,5 +2908,35 @@
...
+      {
+        "name": "DeepBook",
+        "path": "crates/sui-framework/packages/deepbook",
+        "id": "0x000000000000000000000000000000000000000000000000000000000000dee9"
+      },
+      {
+        "name": "Bridge",
+        "path": "crates/sui-framework/packages/bridge",
+        "id": "0x000000000000000000000000000000000000000000000000000000000000000b"
+      }
```
lxfind: **v137 framework bytecode snapshot is stale and recorded off-process** -- Confidence: 100/100 | Importance: 65/100

(found by: Claude, Codex)
- **File**: `crates/sui-framework-snapshot/manifest.json` lines 2911-2939; `crates/sui-framework-snapshot/bytecode_snapshot/137/`
- **Problem**: The v137 snapshot was generated at `f1ac3bc98e`, a pre-rebase branch commit, before the final commit changed the insufficient-funds abort code from 2 to 5 and added the `EObjectFundsInsufficient` constant to `funds_accumulator.move`. Per Claude, the snapshotted `0x2` package lacks that constant. Per Codex, the snapshotted framework also contains 67 modules where the current package has 68. Both symptoms are consistent with a snapshot cut before the rebase. Because 137 equals `ProtocolVersion::MAX`, `check_framework_change_with_protocol_upgrade` compares the snapshot against the built-in framework and fails. Both reviewers reproduced the failure locally:
  ```
  FAIL sui-framework-snapshot::compatibility_tests compatibility_tests::check_framework_change_with_protocol_upgrade
  "The current framework differs the latest bytecode snapshot. Did you forget to upgrade protocol version?"
  ```
  Separately, every existing bytecode snapshot was added by a release commit, for example `a4158b3fb4 Sui v1.79.0 Framework Bytecode snapshot`. v137 is not cut yet: the latest release branch `sui-v1.79.0-release` has `MAX_PROTOCOL_VERSION = 136`.
- **Impact**: CI fails as-is. If the snapshot is merely regenerated and kept, it freezes the v137 framework on `main`. Every other in-flight framework change targeting v137 will trip the same test until v1.80.0 is cut, and the manifest will permanently reference a commit that never lands on `main`.
- **Suggestion**: Remove `bytecode_snapshot/137/` and the `"137"` manifest entry and let the release process record the snapshot. If there is a reason to keep it, regenerate it from the final commit so the compatibility test passes.

---

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -1184,6 +1203,30 @@ impl RuntimeObjectResolver for TemporaryStore<'_> {
...
+            .ok_or(SuiErrorKind::ExecutionInvariantViolation)?
+            .version();
+
+        let settled = AccumulatorRootValue::load(self, Some(required_version), owner, type_)?
+            .and_then(|value| value.as_u128())
+            .unwrap_or(0);
+
+        let unsettled = self.unsettled_object_funds.get_unsettled_object_withdraw(
+            &AccumulatorRootValue::get_field_id(owner, type_)?,
+            required_version,
```
lxfind, line 1222: **Cold read hashes the accumulator dynamic-field key twice but charges for one** -- Confidence: 96/100 | Importance: 60/100

(found by: Codex)
- **File**: `sui-execution/latest/sui-adapter/src/temporary_store.rs` lines 1216-1222
- **Problem**: In the `ObjectFundsResolver` implementation on `TemporaryStore`, `AccumulatorRootValue::load` derives the field ID via `DynamicFieldKey::into_id_with_bound`, which BCS-serializes the key and hashes it. The following `get_field_id` call for the `get_unsettled_object_withdraw` lookup repeats the same serialize-and-hash on the same key. The configured 184-unit cold charge explicitly models one 52-unit hash, one 52-unit borrow, and the 80-byte field load, not two hashes.
- **Impact**: Every first cold read for a distinct `(owner, type)` pair performs an extra unmetered dynamic-field hash. A transaction can repeat this across many distinct pairs, so the implementation consumes more CPU than the protocol gas model prices. The gap widens with large type tags, which the previously reported absence of `check_accumulator_type_bounds` in this native leaves unbounded.
- **Suggestion**: Compute the `AccumulatorObjId` once with `get_field_id`, load the settled value through `load_by_id::<U128>`, and reuse the ID for the unsettled lookup. Alternatively, have `load` return the derived ID alongside the value.

---

crates/sui-adapter-transactional-tests/tests/address_balances/object_funds_withdraw.snap
```diff
@@ -0,0 +1,146 @@
...
+mutated: object(0,0), object(2,0)
+accumulators_written: (object(3,0), A, sui::balance::Balance<sui::sui::SUI>, Merge, 0), (object(3,1), fake(2,0), sui::balance::Balance<sui::sui::SUI>, Split, 0)
+gas summary: computation_cost: 1000000, storage_cost: 2287600,  storage_rebate: 2264724, non_refundable_storage_fee: 22876
+
+task 4, lines 37-39:
+//# programmable --sender A --dry-run --inputs object(2,0) 1 @A
+// Withdrawing one MIST should fail before the vault has an object balance.
+//> 0: test::obj_vault::withdraw_to(Input(0), Input(1), Input(2));
+Error: Transaction Effects Status: MoveAbort(MoveLocation { module: ModuleId { address: sui, name: Identifier("funds_accumulator") }, function: 10, instruction: 0, function_name: Some("reserve_object_funds_for_withdrawal") }, 2) in command 0
+Execution Error: MoveAbort(MoveLocation { module: ModuleId { address: sui, name: Identifier("funds_accumulator") }, function: 10, instruction: 0, function_name: Some("reserve_object_funds_for_withdrawal") }, 2) in command 0
```
lxfind: **Transactional snapshot still expects abort code 2** -- Confidence: 100/100 | Importance: 55/100

(found by: Claude, Codex)
- **File**: `crates/sui-adapter-transactional-tests/tests/address_balances/object_funds_withdraw.snap` lines 33-34, 61-62, 69-70, 88-89, 122-123
- **Problem**: Same root cause as issue 1. The final commit changed `E_OBJECT_FUNDS_INSUFFICIENT` and the Move `#[error(code = 5)]` constant from 2 to 5, but the recorded snapshot still expects `Abort Code: 2` at every insufficient-funds site. Claude reproduced the mismatch locally:
  ```
  -... reserve_object_funds_for_withdrawal (function index 10) at offset 0, Abort Code: 2
  +... reserve_object_funds_for_withdrawal (function index 10) at offset 0, Abort Code: 5
  test run_test::address_balances/object_funds_withdraw.move ... FAILED
  ```
  Codex could not start the transactional runner in its sandbox because genesis setup was blocked from binding a local port, but confirmed the checked-in code-2 expectation is inconsistent with the current code-5 constant.
- **Impact**: CI failure. The code is correct; the recorded expectation is stale.
- **Suggestion**: Re-record with a scoped `cargo insta test -p sui-adapter-transactional-tests` and accept only this file's diff.

---

lxfind (commented):

---

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -1184,6 +1203,30 @@ impl RuntimeObjectResolver for TemporaryStore<'_> {
...
 
+impl ObjectFundsResolver for TemporaryStore<'_> {
+    /// Loads the object balance at the required version and subtracts withdrawals from the same
+    /// checkpoint that have not settled yet.
+    /// This function is expected never to fail; an error indicates an invariant violation.
+    fn object_available_balance(&self, owner: SuiAddress, type_: &TypeTag) -> SuiResult<u128> {
+        let required_version = self
+            .load_implicitly_read_system_object(&SUI_ACCUMULATOR_ROOT_OBJECT_ID)
+            .ok_or(SuiErrorKind::ExecutionInvariantViolation)?
+            .version();
```
lxfind, line 1214: **Simulation responses omit the accumulator root object after a cold read** -- Confidence: 100/100 | Importance: 60/100

(found by: Codex)
- **File**: `sui-execution/latest/sui-adapter/src/temporary_store.rs` lines 1211-1214
- **Problem**: The cold-read path in the `ObjectFundsResolver` implementation loads the accumulator root through `load_implicitly_read_system_object`, which records a `ReadOnlyRoot` dependency in effects. The simulation wrapper `TrackingBackingStore::load_implicitly_read_system_object` forwards that read without calling `track_object`, unlike its other read methods. `simulate_transaction` then identifies the root as a required object, cannot find it among the collected objects, and silently omits it. Codex reproduced this with direct VM probes for both successful and insufficient reservations. A flag-off control run showed no mismatch.
- **Impact**: gRPC simulations that request transaction objects return an incomplete object set. The effects reference the root, but its object data is missing, so the effects renderer cannot populate the root's object type. This is the first code path that exercises an implicit system-object read inside simulation, which is why the gap surfaces with this PR.
- **Suggestion**: Track the returned object inside `TrackingBackingStore::load_implicitly_read_system_object`, matching its other read methods. Add a simulation assertion that every `ReadOnlyRoot` reference in effects has a corresponding object in the returned set.

---

lxfind (commented):

---

PR #27528, opened 2026-07-31
allowances

manolisliolios:
## Description 

Adds allowances

## Test plan 

---

## Release notes

- [x] Protocol: adds `allowances` feature
- [x] Nodes (Validators and Full nodes): adds `allowances` support
- [x] gRPC: adds `allowances` support
- [ ] JSON-RPC: 
- [x] GraphQL: adds `allowances` support
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    if (elapsed > 0 && day < anchor_day.min(days_in_month(year, month))) {
+        elapsed = elapsed - 1;
+    };
+    elapsed / (months as u64)
+}
+
+/// Civil (year, month, day) in UTC — month and day 1-based, so the epoch is
+/// `(1970, 1, 1)` — via Hinnant's `civil_from_days`; unsigned-only works
+/// because chain timestamps are never pre-epoch.
+public(package) fun civil_from_ms(timestamp_ms: u64): (u64, u64, u64) {
```
mystenmark: this seems like something that should be widely available and not re-implemented by every package that needs it? otoh maybe we don't want to open that can of worms

manolisliolios: Yeah, had similar thinking. Currently intentionally `public(package)` to avoid having to make that decision as part of this feature

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+#[error(code = 14)]
+const ENameTooLong: vector<u8> = b"Name exceeds the 128-byte limit";
+#[error(code = 15)]
+const EZeroLifetimeCap: vector<u8> = b"Lifetime cap must be greater than zero";
+#[error(code = 16)]
+const EBadTimeWindow: vector<u8> = b"Expiration must be after the start time";
+#[error(code = 17)]
+const ENoExpiration: vector<u8> = b"Allowance must have an expiration or a rate limit";
+#[error(code = 18)]
+const ENotEnabled: vector<u8> = b"Allowances are not enabled";
```
tnowacki: You should be able to use string literals here. So `""` instead of `b""`

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+}
+
+// === Spend ===
+//
+// Every spend path consumes limits and redeems in one step: returning a bare
+// `Withdrawal<T>` would let limits be consumed without funds actually moving.
+
+/// Signer path: the tx sender must be the spender. (A non-balance spend would
+/// require access to `funds_accumulator::Permit<T>`, so `Balance`-only for now.)
+public fun spend_balance<C>(
```
tnowacki: hmmm. or this this more `balance_spend`?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    clock: &Clock,
+    ctx: &TxContext,
+): Balance<C> {
+    assert!(self.app.is_none(), EHasApp);
+    balance::redeem_funds(self.consume(w, clock, ctx))
+}
+
+/// App path: authorized by `Permit<A>` (matching the allowance's `app`); the
+/// tx must still come from the spender.
+public fun spend_balance_as_app<C, A>(
```
tnowacki: `app_balance_spend`? 
`spend_balance_from_app`? `with_app`? Something about "as_ap" doesn't sit right to me

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+const ENoExpiration: vector<u8> = b"Allowance must have an expiration or a rate limit";
+#[error(code = 18)]
+const ENotEnabled: vector<u8> = b"Allowances are not enabled";
+
+const MAX_NAME_LENGTH: u64 = 128;
+const MS_PER_DAY: u64 = 86_400_000;
+
+/// Created by the core for a declared allowance source. Only the bound
+/// allowance's spend paths can unpack it. Dropping it is fine: funds only
+/// move on redemption.
```
tnowacki:
```suggestion
/// Created by the runtime for a declared allowance source. Enables the holder
/// of an `Allowance` to withdraw the underlying asset `T`. Note that the 
/// `Withdrawal` itself is not exposed to ensure usage against the `Allowance`
/// can be precisely tracked. 
```
What is "the core"? What are all these things? I feel that the comment doesn't really convey how this is used or the limitations

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+public struct AllowanceWithdrawal<phantom T: store> has drop {
+    allowance: ID,
+    inner: Withdrawal<T>,
+}
+
+/// Delegated authority to withdraw `T` from `funder`'s balance, within limits.
+/// A shared object (discoverable + revocable); the spending tx references it by id.
```
tnowacki:
```suggestion
/// Always kept as a shared object.
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+public struct Allowance<phantom T> has key {
+    id: UID,
+    funder: address,
+    /// Always `Some` in the first release.
+    /// `Option` so app-bound allowances can later go keyless.
```
tnowacki:
```suggestion
    /// The spender of the allowance. 
    /// While it is currently always set, in the future this may become optional
    /// to allow for keyless app-bound withdrawals.
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    spender: Option<address>,
+    /// When set, spends need the app's `Permit` on top of the spender's
+    /// signature, and only the app's module can rotate the spender.
```
tnowacki:
```suggestion
    /// When set, requires the app's `Permit`
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    app: Option<TypeName>,
+    /// `None` = no lifetime total; at least one of cap / rate limit must be
+    /// set. Amounts are `u256` (matching `Withdrawal.limit`); times are ms.
```
tnowacki:
```suggestion
    /// An optional lifetime cap on withdrawals using this allowance. 
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    lifetime_cap: Option<u256>,
+    /// Cumulative spend, compared against `lifetime_cap`.
```
tnowacki:
```suggestion
    /// Total cumulative spend from this allowance
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    current_spend: u256,
+    /// Inclusive activation time; `None` = active on issue.
```
tnowacki:
```suggestion
    /// Optional activation start time.
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    start_timestamp_ms: Option<u64>,
+    expiration_timestamp_ms: Option<u64>,
```
tnowacki, line 104:
```suggestion
    /// Optional expiration time. After which, the allowance cannot be used. 
    expiration_timestamp_ms: Option<u64>,
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    rate_limit: Option<RateLimit>,
```
tnowacki, line 107:
```suggestion
    /// INSERT HELPFUL WORDS HERE
    rate_limit: Option<RateLimit>,
```
not exactly sure what to say here yet since I don't fully understand the `RateLimit`

tnowacki: but something should be said

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    /// Off-chain label, at most 128 bytes; not consulted by any check.
```
tnowacki:
```suggestion
    /// A label not used in any check. It is intended for off-chain usage or bookkeeping.
    /// The `name` can be at most 128 bytes.
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    name: String,
+}
+
+/// Revocation for an allowance, sent to the funder at issuance (key-only, non-transferrable).
+/// Also used for discoverability (funder -> allowances)
+public struct AllowanceCap<phantom T> has key {
```
tnowacki, line 116: Will this cap always be used for Revocation? If so, maybe `RevocationCap`? Or `AllowanceRevocationCap`?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    id: UID,
+    allowance: ID,
+}
+
+/// At most `limit` per window. Windows only roll forward, so boundaries never
+/// drift: `FixedWindow` periods tile absolute time from the Unix epoch, while
+/// `CalendarWindow` windows follow the anniversary of the first charge.
```
tnowacki:
```suggestion
/// `RateLimit` enables control over the amount withdrawn during a given period of time. Boundaries are set from fixed intervals starting at Unix time 0 (the Unix epoch)
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+public enum RateLimit has copy, drop, store {
+    FixedWindow {
```
tnowacki:
```suggestion
    /// Withdrawals are limited within fixed intervals, repeating every `period_ms` seconds
    Periodic {
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        period_ms: u64,
+        limit: u256,
+        spent: u256,
+        window_start_ms: u64,
+    },
+    CalendarWindow {
+        months: u8,
```
tnowacki: Total number of months? Months left? Months that have passed?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        limit: u256,
+        spent: u256,
+        /// Anniversary anchor; stamped by the first successful charge (0 until then).
+        first_charge_ms: u64,
+        /// Windows are numbered from the anchor (0 = first). This is the one
+        /// `spent` accumulated in; a spend landing in a later one resets it.
+        window: u64,
+    },
+}
```
tnowacki, line 139: Should `limit` and `spent` be shared in struct? or do you see Rate Limits that won't use those fields?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        limit: u256,
+        spent: u256,
+        window_start_ms: u64,
+    },
+    CalendarWindow {
+        months: u8,
+        limit: u256,
+        spent: u256,
+        /// Anniversary anchor; stamped by the first successful charge (0 until then).
+        first_charge_ms: u64,
```
tnowacki: I don't really understand how this is used

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        /// Windows are numbered from the anchor (0 = first). This is the one
+        /// `spent` accumulated in; a spend landing in a later one resets it.
+        window: u64,
```
tnowacki: Why do we need this?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+}
+
+/// At most `limit` per month, from the first charge.
+public fun monthly(limit: u256): RateLimit { calendar_window(1, limit) }
+
+/// At most `limit` per quarter, from the first charge.
+public fun quarterly(limit: u256): RateLimit { calendar_window(3, limit) }
+
+/// At most `limit` per year, from the first charge.
+public fun yearly(limit: u256): RateLimit { calendar_window(12, limit) }
```
tnowacki: I worry these take up a lot of naming real estate without having a `rate_limit` prefix or suffix

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        lifetime_cap,
+        start_timestamp_ms,
+        expiration_timestamp_ms,
+        rate_limit,
+        ctx,
+    );
+}
+
+/// Like `new`, but also binds the controlling app `A` (see `Allowance.app`).
+entry fun new_for_app<T, A>(
```
tnowacki: This feels like a bit of an issue that `A` has no say in the creation of Allowances that use `A` as its "app". 
It makes it really hard for the app to verify potentially critical information (like the rate limit) is correctly configured.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+
+/// At most `limit` per quarter, from the first charge.
+public fun quarterly(limit: u256): RateLimit { calendar_window(3, limit) }
+
+/// At most `limit` per year, from the first charge.
+public fun yearly(limit: u256): RateLimit { calendar_window(12, limit) }
+
+// `entry`, not `public`: issuance must be an explicit PTB command, so a contract
+// cannot create an allowance funded by the caller inside some other call.
+
```
tnowacki, line 198:
```suggestion
/// Creates a new `Allowance` with the `TxContext::sender` as the the allowance's `funder`. 
/// This is kept as `entry` to ensure the sender approves and is aware of the allowance's creation. 
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        option::none(),
+        lifetime_cap,
+        start_timestamp_ms,
+        expiration_timestamp_ms,
+        rate_limit,
+        ctx,
+    );
+}
+
+/// Like `new`, but also binds the controlling app `A` (see `Allowance.app`).
```
tnowacki:
```suggestion
/// Like `new`, but also binds the controlling app `A` (see `Allowance.app` for more details).
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
tnowacki: Like we might want to require a `Permit<A>`, but exposing such a thing is definitely a bit risky. 

I wonder if we want a builder pattern of some search, where just starting the Builder is limited to being entry? 
I understand this will create a lot more PTB commands but do we care?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        /// Anniversary anchor; stamped by the first successful charge (0 until then).
+        first_charge_ms: u64,
+        /// Windows are numbered from the anchor (0 = first). This is the one
+        /// `spent` accumulated in; a spend landing in a later one resets it.
+        window: u64,
+    },
+}
+
+/// App authorization for the `_as_app` endpoints. A separate type so the
+/// allowance API has its own authorization type instead of `internal::Permit`.
```
tnowacki:
```suggestion
/// Enables app authorization for the `_as_app` endpoints.
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+public struct Permit<phantom A>() has drop;
+
+/// Only `A`'s module can create `internal::Permit<A>`, so only it can build this.
```
tnowacki: I would look to `scratch` for how we should be documenting Permits

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        expiration_timestamp_ms,
+        rate_limit,
+        ctx,
+    );
+}
+
+// === Spend ===
+//
+// Every spend path consumes limits and redeems in one step: returning a bare
+// `Withdrawal<T>` would let limits be consumed without funds actually moving.
```
tnowacki: I'm not sure it is helpful to have unattached comments for this module

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    _: Permit<A>,
+    w: AllowanceWithdrawal<Balance<C>>,
+    clock: &Clock,
+    ctx: &TxContext,
+): Balance<C> {
+    self.assert_app<Balance<C>, A>();
+    balance::redeem_funds(self.consume(w, clock, ctx))
+}
+
+/// Possession of the matching cap is what authorizes revocation; no signer check.
```
tnowacki:
```suggestion
/// Revokes an allowance, removing the ability to spend
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+public fun revoke<T>(self: Allowance<T>, cap: AllowanceCap<T>) {
```
tnowacki:
```suggestion
public fun revoke<T>(self: AllowanceCap<T>, allowance: Allowance<T>) {
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        current_spend: 0,
+        start_timestamp_ms,
+        expiration_timestamp_ms,
+        rate_limit,
+    };
+    let cap = AllowanceCap<T> {
+        id: object::new(ctx),
+        allowance: allowance.id.to_inner(),
+    };
+    transfer::transfer(cap, ctx.sender());
```
tnowacki: Why not return the cap?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    let Allowance {
+        id,
+        ..,
+    } = self;
+    id.delete();
+    cap_id.delete();
+}
+
+/// App-only: rotate the spender key without the funder reissuing.
+public fun rotate_spender<T, A>(self: &mut Allowance<T>, _: Permit<A>, new_spender: address) {
```
tnowacki: Why app only?

tnowacki: (and depending on the answer the comment could use some work)

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    self.assert_app<T, A>();
+    self.spender = option::some(new_spender);
+}
+
+// TODO: update endpoints for altering limits, expiration, etc.
```
tnowacki: What is the plan for this?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+
+fun assert_app<T, A>(self: &Allowance<T>) {
+    assert!(self.app.is_some(), ENoApp);
+    assert!(*self.app.borrow() == type_name::with_defining_ids<A>(), EWrongApp);
```
tnowacki: do we need separate asserts for this? Would it be fine to just do `is_some_and`?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+}
+
+/// Policy checks + accounting shared by all spend paths. The spender gate
+/// lives here; the app / no-app split stays with the callers.
```
tnowacki:
```suggestion
/// Central logic for policy checks and accounting, including the check of the spender. 
/// NB: Any app checks must be done beforehand by the caller. 
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+fun consume<T: store>(
+    self: &mut Allowance<T>,
+    w: AllowanceWithdrawal<T>,
+    clock: &Clock,
+    ctx: &TxContext,
+): Withdrawal<T> {
+    let AllowanceWithdrawal { allowance, inner } = w;
+    assert!(allowance == self.id.to_inner(), EWrongAllowance);
+    assert!(self.spender.contains(&ctx.sender()), ENotSpender);
+    // Signing already verified the funder; defense in depth against a core bug.
```
tnowacki:
```suggestion
    // Defense-in-depth check as this should already be verified at signing.
```

tnowacki: Again, not sure what "core" is

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    assert!(inner.owner() == self.funder, EWrongFunder);
+    let amount = inner.limit();
+    let now = clock.timestamp_ms();
+
+    self.start_timestamp_ms.do_ref!(|start_timestamp_ms| {
+        assert!(now >= *start_timestamp_ms, ENotStarted);
+    });
+
+    self.expiration_timestamp_ms.do_ref!(|expiration_timestamp_ms| {
+        assert!(now <= *expiration_timestamp_ms, EExpired);
```
tnowacki:
```suggestion
        assert!(now < *expiration_timestamp_ms, EExpired);
```
?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    });
+
+    self.lifetime_cap.do_ref!(|lifetime_cap| {
+        assert!(self.current_spend + amount <= *lifetime_cap, EExceedsLifetimeCap);
+    });
+
+    self.current_spend = self.current_spend + amount;
```
tnowacki: Thoughts on swapping these two? 
So just updating the spend and then checking the limit?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+}
+
+/// At most `limit` per window. Windows only roll forward, so boundaries never
+/// drift: `FixedWindow` periods tile absolute time from the Unix epoch, while
+/// `CalendarWindow` windows follow the anniversary of the first charge.
+public enum RateLimit has copy, drop, store {
+    FixedWindow {
+        period_ms: u64,
+        limit: u256,
+        spent: u256,
```
tnowacki, line 130: Doc comments for all these (including those below)

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+        window_start_ms: u64,
```
tnowacki: I find this a bit confusing, and probably needs a doc comment to explain that this is the most recent period start window under which there was spending

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+                        *window_start_ms = *window_start_ms + elapsed_periods * *period_ms;
+                        *spent = 0;
+                    };
+                    assert!(*spent + amount <= *limit, EExceedsRateLimit);
+                    *spent = *spent + amount;
+                },
+                RateLimit::CalendarWindow { months, limit, spent, first_charge_ms, window } => {
+                    if (*first_charge_ms == 0) {
+                        // The first successful charge anchors the windows; an
+                        // aborted spend unwinds the stamp.
```
tnowacki:
```suggestion
                        // The first successful charge anchors the windows
```
We don't need to comment about the semantics of the language (otherwise we would have comments on every line 😉 )

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+                        *first_charge_ms = now;
```
tnowacki: Worth setting spent 0 here? Or asserting it is 0?

tnowacki: And maybe for the fields like window and such?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+                    } else {
+                        let k = elapsed_windows(*first_charge_ms, now, *months);
+                        if (k > *window) {
+                            *window = k;
+                            *spent = 0;
+                        };
+                    };
+                    assert!(*spent + amount <= *limit, EExceedsRateLimit);
+                    *spent = *spent + amount;
```
tnowacki: Same as above, might be easier to do the math then assert `spent <= limit`
Though Im fine with it either way

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+                },
+            },
+        );
+
+    inner
+}
+
+/// How many whole `months`-month windows have elapsed since `anchor_ms` —
+/// equivalently, the number of the window `now_ms` falls in (0 = first).
```
tnowacki: I find this comment generally confusing and am not sure how I would reword it

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+/// because chain timestamps are never pre-epoch.
+public(package) fun civil_from_ms(timestamp_ms: u64): (u64, u64, u64) {
+    let z = timestamp_ms / MS_PER_DAY + 719468;
+    let era = z / 146097;
+    let doe = z % 146097;
+    let yoe = (doe - doe / 1460 + doe / 36524 - doe / 146096) / 365;
+    let doy = doe - (365 * yoe + yoe / 4 - yoe / 100);
+    let mp = (5 * doy + 2) / 153;
+    let day = doy - (153 * mp + 2) / 5 + 1;
+    let (year, month) = if (mp < 10) (era * 400 + yoe, mp + 3) else (era * 400 + yoe + 1, mp - 9);
```
tnowacki: 🪄 
Some comments here might be helpful

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    (year, month, day)
+}
+
+public(package) fun days_in_month(year: u64, month: u64): u64 {
+    if (month == 2) {
+        if (is_leap_year(year)) 29 else 28
+    } else if (month == 4 || month == 6 || month == 9 || month == 11) 30
```
tnowacki:
```suggestion
    if (month == 2) (if (is_leap_year(year)) 29 else 28)
    else if (month == 4 || month == 6 || month == 9 || month == 11) 30
    else 31
```
Also maybe consider a match with guards? if it makes it easier to read

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    name: String,
+    spender: address,
+    app: Option<TypeName>,
+    lifetime_cap: Option<u256>,
+    start_timestamp_ms: Option<u64>,
+    expiration_timestamp_ms: Option<u64>,
+    rate_limit: Option<RateLimit>,
+    ctx: &mut TxContext,
+) {
+    assert!(sui::protocol_config::is_feature_enabled(b"enable_allowances"), ENotEnabled);
```
tnowacki: Do we need this here?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    // we do not allow unlimited allowances (TODO: Do we?)
```
tnowacki: Let's decide on this before landing

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    assert!(lifetime_cap.is_some() || rate_limit.is_some(), ENoLimit);
+    assert!(expiration_timestamp_ms.is_some() || rate_limit.is_some(), ENoExpiration);
```
tnowacki, line 485: I'm confused by these. Why these limitations?

manolisliolios: We want to ensure there are no unlimited / unbounded allowances (to protect users)

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: The current agreement is we don't

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: It's kinda magic, should I link to this maybe? 

https://howardhinnant.github.io/date_algorithms.html#civil_from_days

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: should be fine to chain this

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: The idea is that when you bound an allowance to an app, the app is responsible for rotating & dictating who the spender is. 

For individual (non-app) allowances, there's no built-in thing for this. They can leverage address aliases to do this.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: The 2 use cases are discoverability + revocation; I'd probably lean into the latter (`AllowanceRevocationCap`), but I honestly chose `AllowanceCap` for simplicity, as it also serves the discovery case.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: my previous view of the stack 😆

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: hm the main point flagged here is that we're never giving away a plain `Withdrawal`. We give an `AllowanceWithdrawal` to make sure that **we always spend** what's tracked in rate limits!

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: Yeah, true point. I'll suffix those.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: This cap is soulbound (cannot be owned by anyone besides the owner of the allowance) so I don't really see a good reason to return it, do you?

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: Kinda? The move code might be released to testnet / mainnet before the feature is enabled so without that you could have creation of allowances that are not usable

crates/sui-framework/packages/sui-framework/sources/allowance.move
tzakian: Might be worth adding in the comment how the lifetimes of the `AllowanceCap` and `Allowance` are related (i.e., both exist or don't exist, can one exist without the other? etc.)

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+/// drift: `FixedWindow` periods tile absolute time from the Unix epoch, while
+/// `CalendarWindow` windows follow the anniversary of the first charge.
+public enum RateLimit has copy, drop, store {
+    FixedWindow {
+        period_ms: u64,
+        limit: u256,
+        spent: u256,
+        window_start_ms: u64,
+    },
+    CalendarWindow {
```
tzakian: I'm trying to understand the difference between the `CalendarWindow` and `FixedWindow`. It seems like the only  difference is a CalendarWindow takes into account leap years and the like, whereas a FixedWindow doesn't. 

However the CalendarWindow adds a decent amount of complexity, and I guess what I'm wondering as well, is can there just be one rate limit layout that encompasses all of this behavior and/or is it sufficient to only have a FixedWindow?

crates/sui-framework/packages/sui-framework/sources/allowance.move
tzakian: How does this handle leap seconds? 

Also is this super necessary? I guess this is getting back to my window question higher up where, if all it was was a fixed window w.r.t, unix epoch ms?

crates/sui-framework/packages/sui-framework/sources/allowance.move
tzakian: please link it!

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+
+public(package) fun days_in_month(year: u64, month: u64): u64 {
+    if (month == 2) {
+        if (is_leap_year(year)) 29 else 28
+    } else if (month == 4 || month == 6 || month == 9 || month == 11) 30
+    else 31
+}
+
+/// Gregorian rule: every 4th year, except centuries not divisible by 400.
+fun is_leap_year(year: u64): bool {
```
tzakian, line 468: Ditto same here and `days_in_month` link plz 🙏🏻

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,417 @@
...
+    start_timestamp_ms: Option<u64>,
+    expiration_timestamp_ms: Option<u64>,
+    rate_limit: Option<RateLimit>,
+    ctx: &mut TxContext,
+) {
+    assert!(sui::protocol_config::is_feature_enabled(b"enable_allowances"), ENotEnabled);
+    // we do not allow unlimited allowances (TODO: Do we?)
+    assert!(lifetime_cap.is_some() || rate_limit.is_some(), ENoLimit);
+    assert!(expiration_timestamp_ms.is_some() || rate_limit.is_some(), ENoExpiration);
+    assert!(name.length() <= MAX_NAME_LENGTH, ENameTooLong);
```
tzakian, line 486: Nit for consistency might be better to be a `<` (upper bounds should either all be < or <=, and I generally prefer <, but that's personal preference).

tzakian (commented):
One overarching thing to look at is consistency around upper bounds as well, e.g., there are a couple places where some bounds are < and other upper bounds are <=. IMO picking one interpretation of upper bound (across different fields) would probably be good.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: from what I read, the stance is to ignore their existence (which is fine especially in the context of what we're doing here)

Screenshot 2026-08-06 at 10 34 55

++ as above, I think we really need expression in months for this

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: When I shared the design around, the repeating concern was around expressing "monthly" subscriptions which would not really be possible with pure timestamp periods (and any other way was just a hacky way around this). That's because months are not fixed 30 days + change across years as you flag. 

The ask could be as simple as... I want to charge on the 1st of each month, which `FixedWindow` would struggle to solve! 

I find this essential & think it contributes to a good experience here (I buy the tradeoff of 50-100 loc)

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: My thinking is starting without modification APIs (you can always issue + revoke, and set boundaries dynamically to create new cycles).

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: True; switched to `balance_spend`, and `app_balance_spend` for the app flow.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: that's done now!

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: done deal (with small re-ordering to make public(package) go first)

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: This was refactored as we chatted yday, to enable two flows, allowing the app to both inspect & require approval for creating the allowance

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: This is all not possible in the new revision (anchor_ms == first_charge) and is now "Option"

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: this is how it flows now!

crates/sui-framework/packages/sui-framework/sources/allowance.move
tzakian: Yea, it makes sense/seems reasonable to keep that as a feature so 👍🏻 main thing was just looking at the complexity it got me wondering, but I think it does make sense for the exact reasons you mention.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+#[error(code = 15)]
+const ENoExpiration: vector<u8> = "Allowance must have an expiration or a rate limit";
+#[error(code = 16)]
+const ENotEnabled: vector<u8> = "Allowances are not enabled";
+
+const MAX_NAME_LENGTH: u64 = 128;
+const MS_PER_DAY: u64 = 86_400_000;
+
+/// Created by the runtime (adapter).
+///
```
tnowacki: I don't think any dev reading this will fully understand what "runtime" means

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+/// The inner `Withdrawal` is never handed out on its own: every spend path
+/// consumes limits and redeems in one step, so an allowance's limits can never
+/// be charged without the funds actually moving.
```
tnowacki:
```suggestion
/// Created via a PTB Argument.
///
/// The inner `Withdrawal` is full contained within this module, and cannot
/// be accessed directly. This ensures a n allowance's limits can never
/// be charged without the funds actually moving.
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+    id: UID,
+    allowance: ID,
+}
+
+/// A recurring cap on withdrawals, applied on top of any lifetime cap.
+///
+/// An enum so that a limit with different mechanics -- a sliding window, or a
+/// continuously refilling bucket -- can be added as a variant later, rather
+/// than as a layout change to allowances that already exist.
+public enum RateLimit has copy, drop, store {
```
tnowacki, line 124: Why is this an enum? I am imagining for upgradability?
While I am generally okay with adding variants to existing enums in the framework. Not everyone seems convinced. Though this is a very good use case potentially

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+        limit: u256,
+        /// Amount spent so far within a window.
+        spent: u256,
+        /// Start of the first window, stamped by the first successful charge.
+        /// Every later window is counted from here.
+        anchor_ms: Option<u64>,
+        /// Which window `spent` is accumulated in, numbered from the anchor
+        /// (0 = first). A spend landing in a later one resets `spent`.
+        index: u64,
+        /// How long a window lasts.
```
tnowacki:
```suggestion
        /// The defining period of time for this rate limit
```
A bit strange to define window with window

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+        window: Window,
+    },
+}
+
+/// How windows are counted from a rate limit's anchor.
```
tnowacki:
```suggestion
/// The defining time period for a `RateLimit::Windowed`.
```
Seems fine to just have back pointer here of sorts

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+public struct Settings has drop, store {
+    /// The address whose balance is debited by spends against this allowance.
+    funder: address,
+    /// The spender of the allowance.
+    /// While it is currently always set, in the future this may become optional
+    /// to allow for keyless app-bound withdrawals.
+    spender: Option<address>,
+    /// When set, requires the app's `SpendPermit` to spend, and only that app
+    /// can rotate the spender or issue the allowance in the first place.
+    app: Option<TypeName>,
```
damirka, line 97: I think I get where this is coming from but I'm not convinced that `app` is the right name for it.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+    /// An optional lifetime cap on withdrawals using this allowance.
+    /// Amounts are `u256`, matching `Withdrawal.limit`.
+    lifetime_cap: Option<u256>,
+    /// Optional activation time, in milliseconds. Inclusive.
+    start_timestamp_ms: Option<u64>,
+    /// Optional expiration time, in milliseconds. After which, the allowance
+    /// cannot be used.
+    expiration_timestamp_ms: Option<u64>,
+    /// An optional recurring cap, applied on top of `lifetime_cap`. At least
```
damirka: Got confused by `cap`, maybe use a synonym?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+    /// one of the two must be set, so no allowance is unbounded.
+    rate_limit: Option<RateLimit>,
+    /// A label not used in any check. It is intended for off-chain usage or
+    /// bookkeeping. The `name` can be at most 128 bytes.
```
damirka: With or without length? 127 then?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+    },
+}
+
+/// How windows are counted from a rate limit's anchor.
+public enum Window has copy, drop, store {
+    /// Windows of exactly `period_ms`.
+    Periodic { period_ms: u64 },
+    /// Windows of `months` civil (UTC) months, renewing on the anchor's
+    /// day-of-month at 00:00 UTC.
+    Calendar { months: u8 },
```
damirka: Is there an option to have days? Seems oddly specific. Perhaps, a better name for the variant can do.
Maybe calling it `Monthly(u8)` works?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+/// A `SettingsPermit<A>` authorizes changing what an allowance bound to `A` *is*
+/// -- issuing one, or rotating its spender. It is issued from an
+/// `internal::Permit<A>`.
+///
+/// Kept distinct from `SpendPermit` so an app can hand out the right to spend
+/// without also handing out the right to reconfigure, and vice versa.
+public struct SettingsPermit<phantom A>() has drop;
+
+/// Issues a `SpendPermit<A>` from the privileged `internal::Permit<A>`.
+public fun spend_permit<A>(_: internal::Permit<A>): SpendPermit<A> {
```
damirka, line 165: Yay! A world of permits!!!!

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+public fun periodic_rate_limit(period_ms: u64, limit: u256): RateLimit {
+    // A zero period resets the window on every spend; a zero limit spends nothing.
+    assert!(period_ms > 0 && limit > 0, EBadRateLimit);
+    new_rate_limit(limit, Window::Periodic { period_ms })
+}
+
+/// At most `limit` per `months` civil (UTC) months, counted from the first
+/// charge: windows renew on its day-of-month at 00:00 UTC, clamped to shorter
+/// months (a Jan 31 anchor renews Feb 28, then Mar 31).
+public fun calendar_rate_limit(months: u8, limit: u256): RateLimit {
```
damirka, line 186: Yeah, I do believe that `montly_rate_limit` works better in this case.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+    assert!(months > 0 && limit > 0, EBadRateLimit);
+    new_rate_limit(limit, Window::Calendar { months })
+}
+
+/// At most `limit` per month, from the first charge.
+public fun monthly_rate_limit(limit: u256): RateLimit { calendar_rate_limit(1, limit) }
```
damirka, line 191: And I finally realised what you meant by "calendar", lol. Perhaps, the enum should be Months and then you can keep "monthly", "quarterly" and "yearly"?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+
+/// At most `limit` per quarter, from the first charge.
+public fun quarterly_rate_limit(limit: u256): RateLimit { calendar_rate_limit(3, limit) }
+
+/// At most `limit` per year, from the first charge.
+public fun yearly_rate_limit(limit: u256): RateLimit { calendar_rate_limit(12, limit) }
+
+// === Issuance ===
+
+/// Creation is always `entry` to ensure only top-level PTB
```
damirka:
```suggestion
/// Creation is private `entry` to prevent creation in arbitrary packages. 
```

damirka: Maybe there's a better way to put it, also this function doesn't really have a meaningful description, only reasoning for fun modifiers.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+        start_timestamp_ms,
+        expiration_timestamp_ms,
+        rate_limit,
+        name,
+    );
+    AllowanceProposal(settings)
+}
+
+/// Accepts a proposal on `A`'s behalf, creating and sharing the allowance.
+public fun issue<T, A>(proposal: AllowanceProposal<T>, _: SettingsPermit<A>, ctx: &mut TxContext) {
```
damirka, line 251: If it accepts, should it be called "accept_proposal"?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+    w: AllowanceWithdrawal<Balance<C>>,
+    clock: &Clock,
+    ctx: &TxContext,
+): Balance<C> {
+    self.assert_app<Balance<C>, A>();
+    balance::redeem_funds(self.consume(w, clock, ctx))
+}
+
+/// Revokes an allowance, removing the ability to spend
+public fun revoke<T>(self: AllowanceCap<T>, allowance: Allowance<T>) {
```
damirka, line 284: Should the order of arguments be swapped?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+
+// === Package ===
+
+/// Civil (year, month, day) in UTC -- month and day 1-based, so the epoch is
+/// `(1970, 1, 1)`. This is Howard Hinnant's `civil_from_days`, where the
+/// derivation of every constant here is documented:
+/// https://howardhinnant.github.io/date_algorithms.html#civil_from_days
+///
+/// The unsigned-only form works because chain timestamps are never pre-epoch.
+public(package) fun civil_from_ms(timestamp_ms: u64): (u64, u64, u64) {
```
damirka, line 358: Shame framework can't depend on "date" package. 😆

damirka: Have we considered moving this logic to "clock"?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+    funder: address,
+    spender: address,
+    app: Option<TypeName>,
+    lifetime_cap: Option<u256>,
+    start_timestamp_ms: Option<u64>,
+    expiration_timestamp_ms: Option<u64>,
+    rate_limit: Option<RateLimit>,
+    name: String,
+): Settings {
+    assert!(sui::protocol_config::is_feature_enabled(b"enable_allowances"), ENotEnabled);
```
damirka:
```suggestion
    assert!(sui::protocol_config::is_feature_enabled("enable_allowances"), ENotEnabled);
```

crates/sui-framework/packages/sui-framework/tests/allowance_tests.move
```diff
@@ -0,0 +1,933 @@
...
+const MS_PER_DAY: u64 = 86_400_000;
+
+// Assertions are behavioral: a check holds iff the spend succeeds or aborts
+// as expected. The test `Clock` starts at 0.
+
+// === Test harness ===
+
+/// Wraps a test body in a scenario (begun as `$sender`) with a clock, and
+/// cleans both up after.
+macro fun test_tx($sender: address, $body: |&mut Scenario, &mut Clock|) {
```
damirka: This isn't really a tx, perhaps, just `test!`?

damirka (commented):
Well done! Left a couple notes on comments and naming. Implementation is decent!

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: @tnowacki we got a conflict of opinions here lol! I had it allowance first + swapped it after todd recommended the other way around. I feel fine both ways tbh.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: yes absolutely, deferring this to a post-allowance discussion / decision!!

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+    );
+    share_new<T>(settings, ctx);
+}
+
+/// Proposes an allowance bound to the controlling app `A`.
+///
+/// Unlike `new`, this does not create anything: `A`'s module must accept the
+/// proposal via `issue`, so an app always gets a say in the allowances that
+/// name it and can vet the configuration.
+entry fun propose_for_app<T, A>(
```
tnowacki, line 228: Maybe we should have a propose for normal use cases too? I mean yeah I don't necessarily see the utility, but something to consider.

crates/sui-framework/packages/sui-framework/sources/allowance.move
tnowacki: Why soulbound?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,546 @@
...
+    _: SettingsPermit<A>,
+    new_spender: address,
+) {
+    self.assert_app<T, A>();
+    self.settings.spender = option::some(new_spender);
+}
+
+// === Accessors ===
+
+public fun settings<T>(self: &Allowance<T>): &Settings { &self.settings }
```
tnowacki: I might suggest prepending the type name here, and then using a `use fun` for the name you want. 
It is going to make this a bit less chaotic if/when things are added.

```suggestion
public use fun allowance_settings as Allowance.settings;
public fun allowance_settings<T>(self: &Allowance<T>): &Settings { &self.settings }
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: Yeah, for upgradeability, e.g. we might wanna introduce rolling window variants in the future for instance! 

We got @tzakian who wants to help us if/when we need to do enum upgrades so I feel calm here!

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: To protect users: I don't think an address should ever be blocked from revoking their own allowances.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: We could always add it in the future if we find a good reason to!

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: I ended up keeping consistency per type:

- caps/limits use `<=`: name length, lifetime usage of an allowance, rate limit usage
- time boundaries use `<`: that keeps the naming true (e..g if we used `<=`, `expiration_timestamp_ms` would be a lie, since expiration would really be `expiration_timestamp_ms + 1`

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: I actually think `issue` feels right as it is the call that ends up issuing the allowance. I'll fix the comment instead to make it less missleading.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: Agreed, I prefixed nested structs and added the appropriate use funs!

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: Renamed this to `CalendarMonths` which is the most explicit way imo.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: I don't see how `app` is not a good fit here. It's meant to cover the case where... an app... is in control of the allowance

amnn: I think I see @damirka's point: at the level of core Sui abstractions we're used to dealing with packages/modules/functions, but not really apps, which sit at a higher abstraction level.

Looking at just this declaration, I can guess that the capability lies with either the module or the package that the type comes from, but I don't really know which, `module_: Option<TypeName>` or `packge: Option<TypeName>` would feel more natural to me.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,547 @@
...
+    spender: Option<address>,
+    /// When set, requires the app's `SpendPermit` to spend, and only that app
+    /// can rotate the spender or issue the allowance in the first place.
+    app: Option<TypeName>,
+    /// An optional lifetime cap on withdrawals using this allowance.
+    /// Amounts are `u256`, matching `Withdrawal.limit`.
+    lifetime_cap: Option<u256>,
+    /// Optional activation time, in milliseconds. Inclusive.
+    start_timestamp_ms: Option<u64>,
+    /// Optional expiration time, in milliseconds. After which, the allowance
```
amnn: Seeing as you mention it for `start_timestamp_ms` -- inclusive or exclusive?

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,547 @@
...
+    /// cannot be used.
+    expiration_timestamp_ms: Option<u64>,
+    /// An optional recurring limit, applied on top of `lifetime_cap`. At least
+    /// one of the two must be set, so no allowance is unbounded.
+    rate_limit: Option<RateLimit>,
+    /// A label not used in any check. It is intended for off-chain usage or
+    /// bookkeeping. The `name` contents can be at most 128 bytes.
```
amnn: Would be nice to explain why the limit exists.

crates/sui-framework/packages/sui-framework/sources/allowance.move
amnn: > We got [@tzakian](https://github.com/tzakian) who wants to help us if/when we need to do enum upgrades so I feel calm here!

The issue is not "nobody's willing to implement enum upgrades", it's "we don't collectively agree whether they are a good idea or not". Making type resolution non-referentially transparent has implications up and down the stack (execution, indexing, RPCs, off-chain codegen etc).

Assuming that you will have in-place type  evolution when it becomes necessary without having that discussion seems like it's putting the cart before the horse.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,547 @@
...
+public enum Window has copy, drop, store {
+    /// Windows of exactly this many milliseconds.
+    PeriodicMs(u64),
+    /// Windows of this many civil (UTC) months, renewing on the anchor's
+    /// day-of-month at 00:00 UTC.
+    CalendarMonths(u8),
+}
+
+/// A proposal that can only be issued by app `A`.
+public struct AllowanceProposal<phantom T>(Settings) has drop;
```
amnn, line 150:
```suggestion
public struct AllowanceProposal<phantom A>(Settings) has drop;
```

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: I didn't realize there's broader disagreement on the subject. There are no assumptions that we have this today. The assumption is that we can get to it if we have a good reason.

Are the concerns around how we would design this, or on the overall concept of adding variants to an enum?

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: Exclusive; will update the comment!

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: My main reasons are:
1. Keep the allowance object size bounded (since it's fetched / checked in signing)
2. Keep the name length bounded so that clients (UIs) can render it with a known cap


Not sure if it's worth going so deep in the move comments

crates/sui-framework/packages/sui-framework/sources/allowance.move
amnn: It may be me that needs to catch up but the last time I was paying attention the concerns were around making type resolution non-referentially transparent (making it unsafe to evaluate `resolve(type)` once and then replace that operation with its result). That goes beyond this specific use of enums, or even adding enum variants to generally making changes to type layouts.

This is a good reason to maybe table it again for our weekly sync!

crates/sui-framework/packages/sui-framework/sources/allowance.move
tnowacki: Aye yeah I think it makes sense this way "The cap revokes the allowance". To me it makes sense flipped from this since that would be "The allowance revokes itself with permission of the cap"

crates/sui-framework/packages/sui-framework/sources/allowance.move
tnowacki: I am not sure what other name to do but `app`. Packages/module are not the right abstraction since there could be more than one type that is used over multiple allowances. 
We sometimes would use the term "witness" for this, which feels a bit off here since the type itself is never materialized. 

This is kind of a new thing as it is one of the first patterns to leverage `Permit` in this way. 

If we wanted a more imperatively styled name here, I think `permit` would be fine (even though its really the inner permit type?). And I'm not sure what sort of declarative name we would use other than "app" but only because it is so general

crates/sui-framework/packages/sui-framework/sources/allowance.move
tnowacki: We should still have some description here.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,547 @@
...
+        start_timestamp_ms,
+        expiration_timestamp_ms,
+        rate_limit,
+        name,
+    );
+    share_new<T>(settings, ctx);
+}
+
+/// Proposes an allowance bound to the controlling app `A`.
+///
```
tnowacki: Similarly we might want some description here other than just "it does the thing it does"

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,547 @@
...
+        allowance: allowance.id.to_inner(),
+    };
+    transfer::transfer(cap, funder);
+    transfer::share_object(allowance);
+}
+
+// === Test-only ===
+
+/// Stands in for the runtime-issued reservation; the protocol creates these,
+/// Move code cannot.
```
tnowacki: I have no idea what this means

crates/sui-framework/packages/sui-framework/sources/allowance.move
tnowacki: These are different things though yeah? Maybe we need T and A?

tnowacki: Or am I lost

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,547 @@
...
+}
+
+/// Revocation for an allowance, sent to the funder at issuance (soulbound).
+/// Also used for discoverability (funder -> allowances).
+///
+/// Created with the allowance and destroyed with it: `revoke` consumes both, so
+/// neither outlives the other.
+public struct AllowanceCap<phantom T> has key {
+    id: UID,
+    allowance: ID,
```
tnowacki, line 118: Do we have/need an accessor for this?

crates/sui-framework/packages/sui-framework/sources/allowance.move
tnowacki: I would comment this on every bound/limit in the various structs. That way you don't have to go hunting for the usages to check

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: We need `T` (just checked). `T` is the asset type, A is the permit (inner) type and not added in the Proposal right now (as it's checked against the `app: Option<TypeName>`). 

We could add it for more type safety of proposals.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: Just added one, not insanely useful but doesn't hurt either.

crates/sui-framework/packages/sui-framework/sources/allowance.move
manolisliolios: Added on all.

crates/sui-framework/packages/sui-framework/sources/allowance.move
```diff
@@ -0,0 +1,549 @@
...
+    let Allowance { id, .. } = allowance;
+    id.delete();
+    cap_id.delete();
+}
+
+/// Rotates the spender key without the funder reissuing the allowance.
+///
+/// App-only: for an app-bound allowance the app dictates who the spender is. Non-app allowances
+/// rotate through address aliases instead.
+public fun rotate_spender<T, A>(
```
tzakian, line 296: nit: test to try and `rotate_spender` on a non-app allowance?

tzakian (approved):

PR #27529, opened 2026-07-31
[2/n] allowances: rust types + RPC

manolisliolios:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-types/src/transaction.rs
```diff
@@ -222,10 +229,28 @@ impl FundsWithdrawalArg {
...
+    ) -> Self {
+        Self {
+            reservation: Reservation::MaxAmountU64(amount),
+            type_arg: WithdrawalTypeArg::Balance(balance_type),
+            withdraw_from: WithdrawFrom::Allowance { funder, allowance },
+        }
+    }
+
+    /// The account debited by this withdrawal, a pure function of the tx:
+    /// allowance sources declare the funder (verified at signing).
```
tnowacki:
```suggestion
    /// The account debited by this withdrawal
```

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,128 @@
...
+    &SUI_FRAMEWORK_ADDRESS,
+    ALLOWANCE_MODULE_NAME,
+    ALLOWANCE_WITHDRAWAL_STRUCT_NAME,
+);
+
+/// BCS mirror of `std::type_name::TypeName`.
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct MoveTypeName {
+    pub name: String,
+}
```
tnowacki: Why is this in Allowance? If we need it, we should make a separate file (if one doesn't already exist here or in external-crates)

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// BCS mirror of the Move enum `sui::allowance::RateLimit`.
```
tnowacki:
```suggestion
/// mirror of the Move enum `sui::allowance::RateLimit`.
```
I'm  not sure I like this "BCS mirror" terminology. It is the Rust mirror, yes. And it is BCS compatible. But it is not a BCS mirror to me, that would just be a Vec<u8>

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,128 @@
...
+        limit: U256,
+        spent: U256,
+        first_charge_ms: u64,
+        window: u64,
+    },
+}
+
+/// BCS mirror of the Move struct `sui::allowance::Allowance<T>`.
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct Allowance {
```
tnowacki: We need all this in Rust? I thought the rate limits and such were just checked in Move?

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,128 @@
...
+            && s.module.as_ident_str() == ALLOWANCE_MODULE_NAME
+            && s.name.as_ident_str() == ALLOWANCE_STRUCT_NAME
+            && s.type_params.len() == 1
+    }
+}
+
+/// Sign-time view of an `Allowance<T>`. The spender can rotate, so never
+/// reuse a resolution across transactions.
+#[derive(Debug, Clone)]
+pub struct ResolvedAllowance {
```
tnowacki, line 94: Is this the right file for this? This isn't a Move equivalent type?

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,128 @@
...
+    if !Allowance::is_allowance(&tag) {
+        return Err(invalid(format!(
+            "declared allowance {id} is not a sui::allowance::Allowance"
+        )));
+    }
+    if !object.owner.is_shared() {
+        return Err(invalid(format!("allowance {id} is not a shared object")));
+    }
+    let allowance: Allowance = bcs::from_bytes(move_obj.contents())
+        .map_err(|e| invalid(format!("failed to deserialize allowance {id}: {e}")))?;
```
tnowacki: We could just walk the payload for the field, like we do for `Coin`, if we want something simpler. 

Maybe @tzakian has some deserialization/visitor thoughts

crates/sui-types/src/allowance.rs
manolisliolios: Hm yes they are, I mainly did that to properly deserialize the struct. Maybe we could pick fields with visitor?

crates/sui-types/src/allowance.rs
manolisliolios: This is a simplified type that's used in sign checks. 

The only reason I define the full-type above is to deserialize the struct, but I sense there's a better way to pick the fields we need from your comment below

crates/sui-types/src/allowance.rs
tzakian: Yea, this to me seems like the perfect case for a visitor -- you only want to extract a couple fields from the larger value. 

In order to use the visitor for this though you'll need to have the type layout for the value, so you'll either need to get that somehow, or manually write out the type layout of the `Allowance` type (but it has the benefit of not deserializing the whole blob, but only the parts that you care about).

crates/sui-types/src/sui_sdk_types_conversions.rs
```diff
@@ -1285,6 +1285,13 @@ impl From<crate::transaction::CallArg> for Input {
                     crate::transaction::WithdrawFrom::Sponsor => {
                         sui_sdk_types::WithdrawFrom::Sponsor
                     }
+                    crate::transaction::WithdrawFrom::Allowance { .. } => {
+                        // TODO(allowances): needs a WithdrawFrom::Allowance variant in
+                        // sui-sdk-types.
+                        unimplemented!(
```
tzakian: nit: todo!?

crates/sui-types/src/rpc_proto_conversions.rs
```diff
@@ -2885,6 +2885,9 @@ impl From<crate::transaction::FundsWithdrawalArg> for FundsWithdrawal {
         message.set_source(match value.withdraw_from {
             crate::transaction::WithdrawFrom::Sender => Source::Sender,
             crate::transaction::WithdrawFrom::Sponsor => Source::Sponsor,
+            // TODO(allowances): needs a Source::Allowance variant (with funder and
+            // allowance id fields) in the sui.rpc.v2 protos.
+            crate::transaction::WithdrawFrom::Allowance { .. } => Source::Unknown,
```
tzakian: todo!?

manolisliolios: Will add that to sdks repo before we land this! 

update: PR is being tracked here: https://github.com/MystenLabs/sui-rust-sdk/pull/295

crates/sui-types/src/allowance.rs
manolisliolios: Moved to base types where the struct type for `TypeName` is also defined.

crates/sui-indexer-alt-graphql/src/api/types/transaction_kind/programmable/inputs/balance_withdraw.rs
```diff
@@ -48,6 +48,11 @@ pub enum WithdrawFrom {
     Sender,
     /// The funds are withdrawn from the sponsor's account.
     Sponsor,
+    /// The funds are withdrawn from an allowance funder's account, with the
+    /// transaction sender as the allowance's spender.
+    // TODO(allowances): expose the funder and allowance id, which requires
+    // `withdraw_from` to become a union rather than an enum.
+    Allowance,
```
amnn: Where did we get to on the possibility of one day supporting withdrawing from an allowance on the sponsor's behalf rather than the sender's?

crates/sui-json-rpc-types/src/sui_transaction.rs
```diff
@@ -2524,6 +2527,11 @@ pub enum SuiWithdrawalTypeArg {
 pub enum SuiWithdrawFrom {
     Sender,
     Sponsor,
+    #[serde(rename_all = "camelCase")]
```
amnn: nit: is this necessary? Won't it be handled by the top-level `rename_all`?

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,297 @@
...
+    let funds_type = tag
+        .type_params
+        .into_iter()
+        .next()
+        .expect("checked by is_allowance");
+
+    let layout = A::MoveTypeLayout::Struct(Box::new(Allowance::layout(funds_type.clone())));
+    // Move wrote these bytes and the checks above pinned the type, so a decode failure means
+    // `Allowance::layout` has drifted from the Move source. Not a panic: this runs at signing.
+    let Ok(Some((funder, spender))) = Extractor::deserialize_value(
```
amnn: I definitely got a kick out of seeing the visitor framework being used in all its glory here, but for this use case it _may_ be simpler to create the Rust equivalent types to `bcs::from_bytes` into and then just pick out what you want, or to use `Extractor` twice, rather than introduce an intermediate `SettingsVisitor`:

```rust
let Ok(Some(funder)) = Extractor::deserialize_value(
    move_obj.contents(),
    &layout,
    &mut AddressVisitor,
    vec![E::Field("settings"), E::Field("funder")],
) else {
  // ...
};

let Ok(Some(spender)) = Extractor::deserialize_value(
  move_obj.contents(),
  &layout,
  &mut OptionVisitor(&mut AddressVisitor),
  vec![E::Field("settings"), E::Field("spender")],
) else {
  // ...
};
```

mystenmark: agreed - it seems simpler to just maintain a parallel type and rely on BCS deserialization.

crates/sui-indexer-alt-graphql/src/api/types/transaction_kind/programmable/inputs/balance_withdraw.rs
manolisliolios: Addressed on #27819 , added the pieces that would allow us to introduce this, if we ever wanted to!

crates/sui-framework-tests/tests/allowance_layout.rs
```diff
@@ -0,0 +1,32 @@
...
+use move_core_types::annotated_value as A;
+use move_core_types::language_storage::TypeTag;
+use sui_framework::BuiltInFramework;
+use sui_types::allowance::Allowance;
+use sui_types::balance::Balance;
+use sui_types::gas_coin::GAS;
+use sui_types::in_memory_storage::InMemoryStorage;
+
+#[test]
+fn allowance_layout_matches_framework() {
```
tzakian: I like this test. Actually, I like this test so much I think we should do this for all the other types that we have defined in a similarish way on the Rust side (but that's for a future PR or PRs).

crates/sui-types/src/allowance.rs
tzakian: Yea, I sadly am going to have to agree 😢 

This is impressive...but yea `bcs::from_bytes` is the way to go here.

manolisliolios: Moved back to `bcs::from_bytes`, visitors get no love allowance!

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub current_spend: U256,
+}
+
+/// Mirror of the Move struct `sui::allowance::Settings`
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct Settings {
+    pub funder: SuiAddress,
+    pub spender: Option<SuiAddress>,
+    /// `std::type_name::TypeName` wraps a single string, so it decodes as one.
+    pub app: Option<String>,
```
tnowacki, line 46: I wonder if we should have a TypeName Rust type... not sure if it would really do anything, but it is interesting to consider maybe.

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Rust bindings for `sui::allowance`.
+//!
+//! Signing validates a tx's declared (funder, allowance) source against the
+//! loaded object and reserves against the funder; execution creates the
+//! `AllowanceWithdrawal<T>` that only the allowance's spend paths can unpack.
```
tnowacki: 🙃 
These are indeed words. Not entirely sure what they mean but they are words.

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub settings: Settings,
+    pub current_spend: U256,
+}
+
+/// Mirror of the Move struct `sui::allowance::Settings`
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct Settings {
+    pub funder: SuiAddress,
+    pub spender: Option<SuiAddress>,
+    /// `std::type_name::TypeName` wraps a single string, so it decodes as one.
```
tnowacki:
```suggestion
    /// `Option<std::type_name::TypeName>`
```
Is maybe fine? Idk but the ", so it decodes as one." annoys me a bit

crates/sui-types/src/allowance.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub fn is_allowance(s: &StructTag) -> bool {
+        s.address == SUI_FRAMEWORK_ADDRESS
+            && s.module.as_ident_str() == ALLOWANCE_MODULE_NAME
+            && s.name.as_ident_str() == ALLOWANCE_STRUCT_NAME
+            && s.type_params.len() == 1
+    }
+}
+
+/// Sign-time view of an `Allowance<T>`. The spender can rotate, so never
+/// reuse a resolution across transactions.
```
tnowacki:
```suggestion
/// Sign-time view of an `Allowance<T>`. 
/// NB: The spender can rotate from transaction to transaction
```
I kind of see the need for the comment, but I'm not really sure what lunatic system would attempt to reuse these resolutions across transactions lol. 
Everything in Sui is per-transaction in terms of authentication, so it seems a bit strange to call out.

crates/sui-types/src/transaction.rs
```diff
@@ -204,6 +204,13 @@ pub enum WithdrawFrom {
     Sender,
     /// Withdraw from the sponsor of the transaction (gas owner).
     Sponsor,
+    /// Withdraw from `funder`'s balance under the given allowance. Declaring
+    /// the funder keeps the debited account derivable from the tx alone;
+    /// signing verifies it against the (immutable) object.
+    Allowance {
```
tnowacki:
```suggestion
    /// Withdraw from `funder`'s balance under the given `Allowance`
```

amnn (approved):
GraphQL changes look good to me!

tzakian (approved):

PR #27530, opened 2026-07-31
[3/n] allowances: sign-time checks

manolisliolios:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-types/src/transaction.rs
```diff
@@ -2949,6 +2951,11 @@ pub trait TransactionDataAPI {
         chain_identifier: ChainIdentifier,
     ) -> BTreeMap<AccumulatorObjId, u64>;
 
+    /// Validates each `WithdrawFrom::Allowance` against its loaded input
+    /// object: declared funder, spender gate, funds type. Execution then
+    /// trusts the declaration (the funder is immutable); policy lives in Move.
+    fn validate_allowance_withdrawals(&self, input_objects: &InputObjects) -> UserInputResult<()>;
```
tnowacki: Do we need a new function for this? Where is this called?
Is there a way of just integrating it with the other, pre-existing withdrawal checks to make sure it isn't skipped?

crates/sui-types/src/transaction.rs
```diff
@@ -3153,6 +3160,70 @@ impl TransactionDataAPI for TransactionDataV1 {
...
+            let WithdrawFrom::Allowance { funder, allowance } = &withdraw.withdraw_from else {
+                continue;
+            };
+            let resolved = match resolved_allowances.entry(*allowance) {
+                Entry::Occupied(entry) => entry.into_mut(),
+                Entry::Vacant(entry) => {
+                    let object = objects_by_id.get(allowance).ok_or_else(|| {
+                        UserInputError::InvalidWithdrawReservation {
+                            error: format!(
+                                "declared allowance {allowance} not found among the tx inputs"
```
tnowacki: nit
```suggestion
                                "specified allowance {allowance} not found among the tx inputs"
```

crates/sui-types/src/transaction.rs
```diff
@@ -3153,6 +3160,70 @@ impl TransactionDataAPI for TransactionDataV1 {
...
+                            ),
+                        }
+                    })?;
+                    entry.insert(parse_allowance_object(object)?)
+                }
+            };
+            if &resolved.funder != funder {
+                return Err(UserInputError::InvalidWithdrawReservation {
+                    error: format!(
+                        "Declared funder {funder} does not match the funder of \
```
tnowacki:
```suggestion
                        "Specified funder {funder} does not match the funder of \
```

crates/sui-types/src/transaction.rs
```diff
@@ -3153,6 +3160,70 @@ impl TransactionDataAPI for TransactionDataV1 {
...
+                });
+            }
+            if resolved.spender != Some(self.sender()) {
+                return Err(UserInputError::InvalidWithdrawReservation {
+                    error: format!(
+                        "Transaction sender is not the spender of allowance {allowance}"
+                    ),
+                });
+            }
+            if resolved.funds_type != withdraw.type_arg.to_type_tag() {
```
tnowacki: Should we just make a ResolveReservation (or whatever it is called) and check equality?

crates/sui-types/src/transaction.rs
```diff
@@ -3153,6 +3160,70 @@ impl TransactionDataAPI for TransactionDataV1 {
...
+            .collect();
+        // Parses are cached per object, but every withdrawal is checked: two
+        // may source the same allowance with different declared funders.
+        let mut resolved_allowances: BTreeMap<ObjectID, ResolvedAllowance> = BTreeMap::new();
+        for withdraw in allowance_withdrawals {
+            let WithdrawFrom::Allowance { funder, allowance } = &withdraw.withdraw_from else {
+                continue;
+            };
+            let resolved = match resolved_allowances.entry(*allowance) {
+                Entry::Occupied(entry) => entry.into_mut(),
```
tnowacki: Is this not `entry(...).or_insert_with(...)`?

crates/sui-types/src/transaction.rs
```diff
@@ -3153,6 +3160,70 @@ impl TransactionDataAPI for TransactionDataV1 {
...
+        let objects_by_id: BTreeMap<ObjectID, &Object> = input_objects
+            .iter()
+            .filter_map(|input| Some((input.id(), input.as_object()?)))
+            .collect();
+        // Parses are cached per object, but every withdrawal is checked: two
+        // may source the same allowance with different declared funders.
+        let mut resolved_allowances: BTreeMap<ObjectID, ResolvedAllowance> = BTreeMap::new();
+        for withdraw in allowance_withdrawals {
+            let WithdrawFrom::Allowance { funder, allowance } = &withdraw.withdraw_from else {
+                continue;
```
tzakian: Debug assert false and/or return an error here? or `filter_map` on line 3166?

crates/sui-types/src/transaction.rs
tzakian, line 3295: Does this need to be a mutable reference? It doesn't seem like it's mutated anywhere below.

crates/sui-types/src/transaction.rs
manolisliolios: Used `filter_map`, liked that more!

crates/sui-types/src/transaction.rs
manolisliolios: I think we'd need a `entry().or_try_insert_with()` API for this to be doable

manolisliolios: It doesn't, but both sides of the match return mutable refs in this API. I casted it to immutable so it's explicit that we are not mutating anyhow.

crates/sui-types/src/transaction.rs
mystenmark: initially I agreed with @tnowacki but I think this is okay - this is a new, distinct check from the existing sufficient-balance checks and it doesn't necessarily make sense to combine the two.

a better name might help though? maybe `check_allowance_inputs_are_present`?

manolisliolios: I'll rename to `check_allowance_inputs`, as it does not only check presence of those (it also validates the funder + spender). 

And agreed, I tried to move this with other withdrawal checks but it overcomplicated things:
1. This does not really do any sufficiency checks -- it's a spend eligibility check (so more like an auth check)
2. balance sufficiency checks I think are done before inputs have loaded (so the change would be a lot beefier to change this)
3. we explicitly want this check to run only on signing (not post-execution). It's a valid scenario for a spender to be rotated mid-way and get rejected in move (I have tests that pin this behaviour higher in the stack)

tzakian (approved):

PR #27535, opened 2026-07-31
[Block Streaming] Gate observer block stream subscription while commit-lagging

akichidis:
## Description 

When an observer node falls behind (or starts mid-epoch), it catches up via commit sync while the live block stream keeps delivering blocks that the observer service downloads, fully verifies and then rejects because the local commit index lags the quorum commit index too much. Every block in the catch-up window is transferred and verified twice, wasting bandwidth and CPU on both the observer and the serving peer, which can saturate the peer.

Suspend the subscription inside the subscription loop itself: when commit lag is detected on a received stream item, i.e. exactly when the observer service would start rejecting streamed blocks, the stream is dropped, and reconnecting is held off until commit sync has brought the local commit index within one commit-sync batch of the quorum commit index. The band between the suspension threshold (5 batches) and the resume threshold (1 batch) is hysteresis preventing flapping.

While suspended, the quorum commit index stays effectively frozen: with no blocks streamed, no new commit votes arrive from the peer, and the blocks fetched by commit sync only carry votes for commits at or below the already-known quorum commit index, so they cannot advance it. Commit sync therefore converges on a fixed target rather than chasing a moving one. On resume, the subscription re-derives its highest rounds from DagState, so the peer backfills blocks broadcast during the gap, and the freshly streamed higher-round blocks carry the votes that refresh the quorum commit index. If that refresh reveals the node is still too far behind, the subscription is suspended again — so a deep catch-up proceeds as a bounded cycle of suspend, catch up, briefly resubscribe, re-suspend.

## Test plan 

CI/PT/simtest

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/observer_subscriber.rs
+++ b/consensus/core/src/observer_subscriber.rs
@@ -31,6 +31,15 @@ use crate::{
 /// The subscription is gated at `COMMIT_LAG_MULTIPLIER` (5) batches of lag, where the
 /// observer service starts rejecting streamed blocks anyway; the band between the two
 /// thresholds is hysteresis preventing rapid gate flapping.
+///
+/// The value must stay within `1..=COMMIT_LAG_MULTIPLIER`:
+/// - `>= 1` is load-bearing for liveness: commit sync never fetches partial batches, so
+///   it can leave up to `commit_sync_batch_size - 1` commits of residual lag that only
+///   streamed blocks can close. With `0`, the resume condition (zero lag) would be
+///   unreachable and the subscription would stay gated forever.
+/// - `<= COMMIT_LAG_MULTIPLIER`, or the subscription resumes while still commit-lagging
+///   and the in-stream lag check immediately drops it again, churning reconnects until
+///   commit sync shrinks the lag below the gate threshold.
 const RESUBSCRIBE_LAG_BATCHES: u32 = 1;
 
 /// How often a gated subscription re-evaluates commit lag.
--- a/consensus/core/src/observer_subscriber.rs
+++ b/consensus/core/src/observer_subscriber.rs
@@ -43,9 +43,12 @@ const GATE_CHECK_INTERVAL: Duration = Duration::from_secs(1);
 /// While the local commit index lags the quorum commit index too much, streamed blocks
 /// would be rejected by the observer service and re-fetched later via commit sync, so the
 /// subscription drops its stream and holds off reconnecting until commit sync has nearly
-/// caught up, saving the bandwidth and verification on both ends. While gated, the
-/// quorum commit index keeps advancing from the commit votes observed on commit-synced
-/// blocks.
+/// caught up, saving the bandwidth and verification on both ends. While gated, the quorum
+/// commit index stays effectively frozen: commit sync only fetches up to the already-known
+/// quorum commit index, so its observed commit votes cannot push it further. Instead, each
+/// resubscription refreshes the quorum commit index from the freshly streamed blocks, and
+/// if the node is still too far behind, the gate re-engages — so a deep catch-up proceeds
+/// as a bounded cycle of gate, catch up, briefly resubscribe, re-gate.
 pub(crate) struct ObserverSubscriber<C: ObserverNetworkClient, S: ObserverNetworkService> {
     context: Arc<Context>,
     network_client: Arc<C>,
@@ -187,7 +190,7 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C, 
             // `tasks` keep running. Dropping `subscription` also ends the `&mut tasks` borrow,
             // so the block handlers in `tasks` can be awaited below.
             let subscription = Self::run_subscription(
-                context,
+                context.clone(),
                 network_client,
                 observer_service,
                 commit_vote_monitor,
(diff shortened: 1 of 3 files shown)


consensus/core/src/observer_subscriber.rs
```diff
@@ -27,9 +27,25 @@ use crate::{
     task::{join_and_propagate_panic, reap_finished_task, shutdown_join_set},
 };
 
+/// Number of commit-sync batches of lag below which a gated subscription is resumed.
+/// The subscription is gated at `COMMIT_LAG_MULTIPLIER` (5) batches of lag, where the
+/// observer service starts rejecting streamed blocks anyway; the band between the two
+/// thresholds is hysteresis preventing rapid gate flapping.
+const RESUBSCRIBE_LAG_BATCHES: u32 = 1;
```
arun-koshy, line 43: I actually hit an issue while building the minimal block flow. If you reject live blocks without counting their commit votes then you will have sync issues. So maybe worth documenting that this has to be >= 1 . 

One question, is reconnecting alone enough to refresh the quorum commit index during long catch-ups, or would a lightweight side channel for commit votes be worth it? My read is reconnect and refresh is fine, but flagging it just incase

consensus/core/src/observer_subscriber.rs
```diff
@@ -352,6 +381,62 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
...
+                gated = true;
+                info!(
+                    "Gating block stream subscription: local commit index {} lags quorum commit index {}, blocks will arrive via commit sync",
+                    local_commit_index, quorum_commit_index,
+                );
+                context
+                    .metrics
+                    .node_metrics
+                    .observer_subscription_gated
+                    .set(1);
```
arun-koshy: nit: if the subscription task is reset while gated the metrics can show that the node is gated even though its healthy

consensus/core/src/observer_subscriber.rs
akichidis: > I actually hit an issue while building the minimal block flow. If you reject live blocks without counting their commit votes then you will have sync issues. So maybe worth documenting that this has to be >= 1 .

Did you hit an issue with this PR or the minimal block flow in general? Yes, we definitely need this to be >= 1, otherwise if `commit_syncer` does't fill up to the `quorum_commit_index` (equal) we'll never be able to resume. Good point on documenting that. I might be a bit more forgiving and add some more buffer for reassurance.

> One question, is reconnecting alone enough to refresh the quorum commit index during long catch-ups, or would a lightweight side channel for commit votes be worth it? My read is reconnect and refresh is fine, but flagging it just incase

So the quorum commit index should get advanced by reconnecting, receiving the new blocks and making then the decision either to shutdown again (if it's too far) , or keep the connection open (if we have nearly caught up). Now, I see that my PR has some confusing comments and I'll fix those, because they aren't accurate. So yes, the reconnect alone should be fine. An alternative I could think of would be to have some timeout to allow periodically a few blocks getting received , just to make sure we have some additional liveness guarantees. 

Having a lightweight side channel could also do , but I would see this as a more overall upgrade of how we advance our commit quorum index - ex we could have the round prober querying for the commit index as well.

arun-koshy: Minimal block flow in general, not this PR. I had an issue where blocks stopped feeding the commit vote monitor and sync stalled because commit votes are only counted after it has been verified but you cant verify a minimal block until its fully inflated, which is why the >= 1 coupling jumped out at me. The new docs capture it well. Agree reconnect-and-refresh is sufficient given the resume guarantee. The round-prober idea seems like the right shape if we ever want the quorum index live during long catch-ups.

consensus/core/src/observer_subscriber.rs
```diff
@@ -280,18 +323,25 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
...
                                 commit_vote_monitor.quorum_commit_index(),
                             )
                         });
-                        if let Some(handler) = &randomness_signature_handler
-                            && !is_commit_lagging
-                        {
+                        if is_commit_lagging {
+                            retries = 0;
+                            backoff.reset();
+                            break 'stream;
```
arun-koshy, line 341: One leftover comment from AI review was that this line is not tested in your tests. So if you remove it the tests still pass because the fake stream you setup hangs up by itself after it sends 10 items. Maybe worth adding a test variant where the fake stream never ends so that the gate engages and hangs up the line here.

arun-koshy (approved):

consensus/core/src/observer_subscriber.rs
```diff
@@ -352,6 +402,68 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
...
+    // sync only fetches up to it, so its observed votes cannot push it further. It is
+    // refreshed from streamed blocks after the subscription resumes; if that reveals the
+    // node is still too far behind, the stream is dropped and the gate re-engages.
+    // Returns false when the node is shutting down.
+    async fn wait_while_commit_lagging(
+        context: &Context,
+        commit_vote_monitor: &CommitVoteMonitor,
+        dag_state: &Weak<parking_lot::RwLock<DagState>>,
+    ) -> bool {
+        let mut gated = false;
```
mwtian: nit: I think `suspended` is more common in our codebase.

mwtian (approved):
This part of PR description seems inconsistent with code and comment: `While gated, the quorum commit index keeps advancing from the commit votes carried by commit-synced blocks, so both the resume decision and commit sync scheduling stay live.`

Also, comment on the start and stop behavior of commit sync due to commit vote not arriving when streaming is stopped?

akichidis:
> This part of PR description seems inconsistent with code and comment: `While gated, the quorum commit index keeps advancing from the commit votes carried by commit-synced blocks, so both the resume decision and commit sync scheduling stay live.`
> 
> Also, comment on the start and stop behavior of commit sync due to commit vote not arriving when streaming is stopped?

@mwtian you are right - apologies for this. This was an artifact from when I've originally opened the PR which was obviously wrong. I'll fix. I can comment further on the behaviour of the commit sync yes

PR #27536, opened 2026-07-31
[consensus] Read store first in commit sync fetch_blocks to avoid I/O under the DagState lock

akichidis:
## Description 

BlockSyncService::fetch_blocks served requested blocks via DagState::get_blocks, which performs its store fallback while the caller holds the DagState read lock. Commit-sync serving reads up to max_blocks_per_fetch (1000) blocks per request, so bursts of fetches from lagging peers held the read lock across large storage reads. Since the lock is write-preferring, this delays Core's writes on the consensus hot path and queues later readers behind them.

For commit sync requests (no fetch_after_rounds), invert the read order: read the requested blocks from the store directly with no lock held, then look up only the store misses via DagState. Store misses are the recently accepted blocks not yet flushed, which are served from the recent blocks cache (eviction only removes blocks below the eviction round); only blocks that do not exist at all fall through to a second, small store read under the lock, and honest requesters never ask for those.

Below is the utilisation of the Core thread of a node when a peer is making a series of requests to fetch blocks during commit sync (today is the case of the block streaming clients). After this change the utilisation remains flat and no noticeable change on it.

Screenshot 2026-07-31 at 13 47 39

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/block_sync_service.rs
+++ b/consensus/core/src/block_sync_service.rs
@@ -72,6 +72,12 @@ impl BlockSyncService {
         if block_refs.is_empty() && (fetch_missing_ancestors || fetch_after_rounds.is_empty()) {
             return Err(ConsensusError::InvalidFetchBlocksRequest("When no block refs are provided, fetch_after_rounds must be provided and fetch_missing_ancestors must be false".to_string()));
         }
+        if fetch_missing_ancestors && fetch_after_rounds.is_empty() {
+            return Err(ConsensusError::InvalidFetchBlocksRequest(
+                "When fetch_missing_ancestors is true, fetch_after_rounds must be provided"
+                    .to_string(),
+            ));
+        }
         if !fetch_after_rounds.is_empty()
             && fetch_after_rounds.len() != self.context.committee.size()
         {
@@ -111,7 +117,7 @@ impl BlockSyncService {
         // targets recently accepted blocks, capped at max_blocks_per_sync, which are
         // almost always in the DagState cache; reading the store first would only add a
         // wasted store read per request.
-        let mut blocks = if fetch_after_rounds.is_empty() && !fetch_missing_ancestors {
+        let mut blocks = if fetch_after_rounds.is_empty() {
             self.read_blocks(&block_refs)?
         } else {
             self.dag_state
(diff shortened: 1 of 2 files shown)


consensus/core/src/block_sync_service.rs
```diff
@@ -103,14 +104,23 @@ impl BlockSyncService {
...
-            .flatten()
-            .collect::<Vec<_>>();
+        // Get the requested blocks first. Commit sync requests (block refs only, no
+        // fetch_after_rounds, no fetch_missing_ancestors) read old committed blocks, up
+        // to max_blocks_per_fetch of them, so they read the store first to avoid holding
+        // the DagState lock across large storage reads. Every other request shape
+        // targets recently accepted blocks, capped at max_blocks_per_sync, which are
+        // almost always in the DagState cache; reading the store first would only add a
+        // wasted store read per request.
+        let mut blocks = if fetch_after_rounds.is_empty() && !fetch_missing_ancestors {
```
arun-koshy: I think the only check needed here is fetch_after_rounds.is_empty(). That's the only pattern that can reach this read with up to max_blocks_per_fetch refs (non-empty rounds truncate refs to max_blocks_per_sync above, and empty refs read nothing).

consensus/core/src/block_sync_service.rs
```diff
@@ -203,6 +215,33 @@ impl BlockSyncService {
...
+    /// Reads blocks by refs like `DagState::get_blocks`, but reads the store first so no
+    /// storage I/O happens while holding the DagState lock: commit sync requests read up
+    /// to `max_blocks_per_fetch` blocks, and a read lock held across large store reads
+    /// delays Core's writes on the consensus hot path. Blocks missing from the store are
+    /// the most recently accepted ones that are not yet flushed, which are found in
+    /// DagState's in-memory cache; only blocks that do not exist at all fall through to
+    /// `get_blocks`' internal store read, and honest requesters never ask for those.
+    /// Blocks found nowhere are dropped from the result.
+    fn read_blocks(&self, block_refs: &[BlockRef]) -> ConsensusResult<Vec<VerifiedBlock>> {
+        let mut blocks = self.store.read_blocks(block_refs)?;
```
arun-koshy, line 238: For commit sync there was a metric tracking the store read count in dag state for get blocks but we wont have that as we are directly going to store here, may be worth adding a counter?

arun-koshy (approved):

consensus/core/src/block_sync_service.rs
akichidis: Good point, let me add something here

consensus/core/src/block_sync_service.rs
akichidis: @arun-koshy I have added an explicit validation check , which should in practice already hold true for the current paths, but I believe it will make the expectations clearer.

PR #27537, opened 2026-07-31
[graphql-stream] resolve subscription payloads concurrently [16/n]

tpham-mysten:
## Description

Resolve subscription payloads concurrently rather than one at a time. A single `max_concurrent_resolutions` config drives both the resolve window (async-graphql's `subscription_resolution_concurrency`, now [merged](https://github.com/amnn/async-graphql/pull/2)) and the backfill scan page, so a batch's matches resolve within the concurrency budget and coalesce their content reads into one `KvLoader` round trip.

## Test plan

- e2e parity test (`test_transaction_subscription_live_backfill_parity`): the same transactions resolve identically whether delivered live or through the backfill scan, plus the existing streaming/ordering/resume subscription tests.
- Batching e2e tests: `test_transaction_subscription_backfill_batches_matches` (a deep backfill packs matches into multi-edge payloads) and `test_transaction_subscription_live_batches_per_checkpoint` (each live payload is one checkpoint's matches).
- Unit tests for the scan retry: recovers within the budget, gives up once it is exhausted.

---

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425
   - #26495
   - #26731
   - #26714
   - #27140
   - #27537

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/lib.rs
+++ b/crates/sui-indexer-alt-graphql/src/lib.rs
@@ -317,7 +317,9 @@ pub async fn start_rpc(
     pg_pipelines: Vec<String>,
     registry: &Registry,
 ) -> anyhow::Result<Service> {
-    let rpc = RpcService::new(args, version, schema(), registry);
+    let schema = schema()
+        .subscription_resolution_concurrency(config.subscription.max_concurrent_resolutions);
+    let rpc = RpcService::new(args, version, schema, registry);
     let metrics = rpc.metrics();
 
     // Create gRPC full node client wrapper. If left unconfigured, the client will not be stored in
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1814,7 +1814,7 @@ dependencies = [
 [[package]]
 name = "async-graphql"
 version = "7.0.1"
-source = "git+https://github.com/amnn/async-graphql?branch=v7.0.1-react-18-graphiql-4#7be9351958ef7ebcbb7fed2f45dd3e9c92398df1"
+source = "git+https://github.com/amnn/async-graphql?branch=v7.0.1-react-18-graphiql-4#5224a1c72098070b817cfc401f84e8a6c336c83f"
 dependencies = [
  "async-graphql-derive",
  "async-graphql-parser",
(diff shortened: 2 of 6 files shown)


crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -274,8 +275,11 @@ impl Default for RpcArgs {
 }
 
 /// The GraphQL schema this service will serve, without any extensions or context added.
-pub fn schema() -> SchemaBuilder<Query, Mutation, Subscription> {
+pub fn schema(
+    subscription_config: &SubscriptionConfig,
+) -> SchemaBuilder<Query, Mutation, Subscription> {
     Schema::build(Query::default(), Mutation, Subscription)
+        .subscription_resolution_concurrency(subscription_config.max_concurrent_resolutions)
```
amnn: Do you need to set that here, or could you set that outside this function? I.e. from the caller of `schema` that actually uses it to execute transactions?

That way, you could avoid having to update all the other callers that are generating the schema for other purposes.

crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -317,7 +321,7 @@ pub async fn start_rpc(
     pg_pipelines: Vec<String>,
     registry: &Registry,
 ) -> anyhow::Result<Service> {
-    let rpc = RpcService::new(args, version, schema(), registry);
+    let rpc = RpcService::new(args, version, schema(&config.subscription), registry);
```
amnn: Per the comment above:

```suggestion
    let schema = schema().subscription_resolution_concurrency(
        config.subscription.max_concurrent_resolutions
    );
    
    let rpc = RpcService::new(args, version, schema, registry);
```

amnn (approved):
Thanks @tpham-mysten, looks good but take a look at  the comment to avoid the churn on the `schema` function.

crates/sui-indexer-alt-graphql/src/lib.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/lib.rs
tpham-mysten: SG. Done

tpham-mysten (commented):
Thanks @amnn! I have applied your suggestion on the `schema` builder

PR #27539, opened 2026-07-31
sui-indexer-alt-reader: make LedgerGrpcReader's batch-chunking size configurable

evan-wall-mysten:
## Description

sui-rpc-api and sui-kv-rpc each hardcode their own 200/1000 batch-request caps as local constants (unchanged here). `LedgerGrpcReader`'s chunk size (introduced in #27516) was previously a fixed constant, so GraphQL had no way to tune its own downstream batch sizes independently.

`LedgerGrpcReader` now accepts a configurable batch size at construction, clamped so it can only be set lower than its built-in default, never higher — a larger value would just be rejected server-side. GraphQL's `Limits` config plumbs this override through; jsonrpc keeps the built-in default.

## Test plan

Covered by existing tests in `sui-indexer-alt-reader`'s chunking suite (`events`/`transactions`/`objects`).

---

## Release notes

- [ ] Nothing user-facing changed.

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/config.rs
+++ b/crates/sui-indexer-alt-graphql/src/config.rs
@@ -7,6 +7,8 @@ use std::time::Duration;
 
 use serde::Deserialize;
 use serde::Serialize;
+use sui_indexer_alt_reader::ledger_grpc_reader::MAX_BATCH_GET_OBJECTS;
+use sui_indexer_alt_reader::ledger_grpc_reader::MAX_BATCH_GET_TRANSACTIONS;
 use sui_name_service::NameServiceConfig;
 use sui_protocol_config::Chain;
 use sui_protocol_config::ProtocolConfig;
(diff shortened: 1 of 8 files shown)


nickvikeras:
This doesn't really address my comment from the last PR - https://github.com/MystenLabs/sui/pull/27516#discussion_r3686038742

If we are going to couple all of these together, the constant shouldn't live in sui-types. The crate where Will put the cursor stuff would make more sense. But I still think it's a footgun that is going to hurt whoever tries to change the limits.

amnn (commented):
Will leave this with @nickvikeras to review/accept. It looks like this still needs changes to move the defaults out of `sui-types` which I agree is not the right home for this.

evan-wall-mysten:
> Will leave this with @nickvikeras to review/accept. It looks like this still needs changes to move the defaults out of `sui-types` which I agree is not the right home for this.

Yeah I broke this out of the original PR so the original could be merged, but did not yet update it to remove the shared config.

nickvikeras (approved):

PR #27544, opened 2026-07-31
sui-framework: add package version configuration

tzakian:
## Description 

Add the PackageConfig singleton and per-package version forbid-list settings, with the corresponding Rust types and compiler/verifier plumbing.

## Test plan 

CI + Move unit tests for the module. Integration tests + final plumbing is added in the PR stacked on top of this one.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-framework/packages/sui-framework/sources/object.move
```diff
@@ -40,6 +40,9 @@ const SUI_RANDOM_ID: address = @0x8;
 /// The hardcoded ID for the singleton DenyList.
 const SUI_DENY_LIST_OBJECT_ID: address = @0x403;
 
+/// The hardcoded ID for the singleton PackageConfig.
+const SUI_PACKAGE_CONFIG_OBJECT_ID: address = @0x426;
```
tzakian, line 44: Open to suggestions on the address for this

tnowacki: Where did the number come from?

crates/sui-framework/packages/sui-framework/sources/package_config.move
```diff
@@ -0,0 +1,273 @@
...
+            if (old_value.is_some()) *flags = old_value.destroy_some();
+            assert!(is_supported_bitset_version(*flags), EUnsupportedBitsetVersion);
+            *flags = *flags | VERSION_FORBIDDEN;
+        },
+        ctx,
+    );
+}
+
+/// Forbid a historical version of the package controlled by `cap`.
+public fun forbid_version(
```
tnowacki: nit, can you order these with `public` > `public(package)` > private > `#[mode(test]]`

crates/sui-framework/packages/sui-framework/sources/package_config.move
```diff
@@ -0,0 +1,273 @@
...
+    if (!package_config.per_package_metadata_exists(original_id)) return false;
+    let config = package_config.borrow_per_package_config(original_id);
+    let flags = config.read_setting_for_next_epoch<_, _, u64>(VersionForbiddenKey(version));
+    if (flags.is_none()) return false;
+    is_version_forbidden(flags.destroy_some())
+}
+
+// Test-only APIs for unit tests
+
+#[test_only]
```
tnowacki: nit 
```suggestion
#[mode(test)]
```

crates/sui-framework/packages/sui-framework/sources/package_config.move
```diff
@@ -0,0 +1,273 @@
...
+) {
+    let current_version = package::version(cap);
+    assert!(version > 0 && version < current_version, EInvalidVersion);
+    forbid_version_impl(package_config, package::original_package_id(cap), version, ctx);
+}
+
+/// Forbid all historical versions in the inclusive range `[start, end]`.
+public fun forbid_version_range(
+    package_config: &mut PackageConfig,
+    cap: &UpgradeCap,
```
tnowacki: Do you think we should have a wrapper for `forbid(0, package::version(cap) - 1)`?

crates/sui-framework/packages/sui-framework/sources/package_config.move
```diff
@@ -0,0 +1,273 @@
...
+const VERSION_FORBIDDEN: u64 = 1;
+
+/// A shared singleton that stores per-package configuration metadata.
+public struct PackageConfig has key {
+    id: UID,
+}
+
+/// The capability used to write to package configs. Ensures that per-package `Config`s are
+/// modified only by this module.
+public struct PackageConfigCap() has drop;
```
tnowacki, line 29: Oof. Forgot this still exists. 
We should really switch this over to a permit system for the Config

crates/sui-framework/packages/sui-framework/sources/package_config.move
```diff
@@ -0,0 +1,273 @@
...
+const EInvalidVersion: u64 = 1;
+/// The start of a version range is greater than its end.
+const EInvalidVersionRange: u64 = 2;
+/// The forbid-list bitset schema version is not supported by this implementation.
+const EUnsupportedBitsetVersion: u64 = 3;
+
+/// The high eight bits identify the bitset schema version.
+const BITSET_VERSION_SHIFT: u8 = 56;
+/// Schema version zero's bit indicating that the package version is forbidden.
+const VERSION_FORBIDDEN: u64 = 1;
```
tnowacki: I'm a bit confused about the motivation here. If we needed to update these settings, wouldn't we just have a new key? Like `VersionForbiddenV2`? And if we needed extra flags, we could just issue a new key?

crates/sui-framework/packages/sui-framework/sources/package_config.move
```diff
@@ -0,0 +1,273 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// On-chain package configuration, keyed by original package ID.
+///
+/// The configuration hierarchy is:
+/// ```
+/// PackageConfig
+///   └── PackageMetadataKey(original package ID)
+///         └── Config<PackageConfigCap>
```
tnowacki, line 10: Do we need a separate config per package?
Totally fine as is, but the reason we did it this way for DenyList was for potential caching concerns. It made it a bit easier to hardcode specific token/currency configurations. I am not sure if we need it per package like this, or if we need some different scheme, but I figured it was worth pointing out

crates/sui-framework/packages/sui-framework/sources/package_config.move
tzakian: old habits die hard...

crates/sui-framework/packages/sui-framework/sources/package_config.move
tzakian: Yea, totally what we should do (for some reason it didn't cross my mind :'( )

I'm just gonna back out the schema stuff (but will keep the u64 bitset) and if we need to version it we'll do it with keys.

crates/sui-framework/packages/sui-framework/sources/object.move
tzakian: Inspired by `0x403` I went looking at HTTP error codes, and this one seemed like the most apropos https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/426

PR #27545, opened 2026-07-31
Add package version forbid-list enforcement

tzakian:
## Description 

Create PackageConfig at epoch boundaries when the feature is enabled, reject forbidden resolved package versions at signing.

Need to make the sui-sdk-types changes and then pull those in on top of this (will be later in this stack).

## Test plan 

CI + transactional tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-types/src/rpc_proto_conversions.rs
```diff
@@ -2691,6 +2691,9 @@ impl From<crate::transaction::EndOfEpochTransactionKind> for EndOfEpochTransacti
                 .with_kind(Kind::AuthenticatorStateExpire),
             K::RandomnessStateCreate => message.with_kind(Kind::RandomnessStateCreate),
             K::DenyListStateCreate => message.with_kind(Kind::DenyListStateCreate),
+            K::PackageConfigCreate => {
+                todo!("PackageConfigCreate is not yet represented in the RPC API")
```
tzakian, line 2707: Will add these to the sdk types and do that whole dance on top of this PR

crates/sui-types/src/package_config.rs
```diff
@@ -0,0 +1,131 @@
...
+    version: u64,
+    object_store: &dyn ObjectStore,
+    cur_epoch: Option<EpochId>,
+) -> bool {
+    get_per_package_config(original_id, object_store)
+        .and_then(|config| read_version_forbid_flags(&config, version, object_store, cur_epoch))
+        .is_some_and(flags_forbid_version)
+}
+
+fn read_config_setting<K, V>(
```
tnowacki: Is this not pre-existing anywhere?

crates/sui-types/src/package_config.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// Returns whether `version` is forbidden for `original_id`. Missing configuration is allowed.
+pub fn is_version_forbidden(
+    original_id: ObjectID,
+    version: u64,
+    object_store: &dyn ObjectStore,
+    cur_epoch: Option<EpochId>,
+) -> bool {
+    get_per_package_config(original_id, object_store)
+        .and_then(|config| read_version_forbid_flags(&config, version, object_store, cur_epoch))
```
tnowacki, line 112: Everyone okay with this from a perf standpoint? Tis a lot of object queries added to each transaction

crates/sui-transaction-checks/src/lib.rs
```diff
@@ -829,4 +841,49 @@ mod checked {
...
+                pt,
+                backing_store,
+            )
+        else {
+            return Ok(());
+        };
+
+        for original_id in execution_original_ids {
+            let Some((_, version)) = resolved_packages.get(&original_id) else {
+                continue;
```
tnowacki, line 874: I continue to find this set+map extremely vexing

crates/sui-types/src/package_config.rs
tzakian: I think it's fine from a perf standpoint (and is only paid at signing time). 

If perf becomes an issue, we could do some bloom filter magic possibly to minimize the number of "false lookups" possibly. But I'd prefer to hold off on that unless we need to do it.

crates/sui-adapter-transactional-tests/tests/programmable/package_version_forbid_list_linkage.move
```diff
@@ -0,0 +1,121 @@
...
+
+public fun stamp(s: &mut S) { s.v = 3 }
+
+public fun ping() {}
+
+// Forbid v2 as well.
+//# run sui::package_config::forbid_version --args object(0x426) object(1,1) 2 --sender A
+
+// A forbidden type package remains allowed when no package code is executed from it.
+//# run Generic::g::use_type --type-args DepV2::d::B --sender A
```
tzakian, line 96: @tnowacki This is an example of why we need the "execution" set. `DepV2` is forbidden, but can still be used as a type argument.

crates/sui-types/src/package_config.rs
tzakian: Yea we have this in the deny list and probably should just reuse that and factor out config helpers possibly into a separate module -- let me take a look and see.

PR #27546, opened 2026-07-31
[jsonrpc-alt] Uniform address balance spoofing

wlmyng:
## Description 

  On jsonrpc-alt, we pretend an address balance is a coin object, spoofing an ObjectRef whose digest carries (epoch, amount). The two read paths fabricated this ref independently and consequently disagreed:

- `suix_getCoins` encoded the ref with `ParsedObjectRefWithdrawal::encode()`, but sourced the amount from the consistent store's get_balance() while taking the version from the live accumulator object. This means if consistent-store lagged, a stale amount gets encoded against a live version.
- `sui_getObject/sui_multiGetObjects` hashed the fabricated coin object, which is hashed for the object digest, which would result in a different value.

The new `AddressBalanceCoin` centralizes this spoofing. Two entrypoints by_owner (getCoins) and by_object_id (getObject) differ in how the accumulator field ID is obtained (derived from `(owner, coin_type)` or unmasking the request ID.)

Now, both read paths load the live accumulator object and derive everything from the loaded contents for uniformity.

## Test plan 

Validated with test:
```

    running 1 test
    test test_get_object_ab_coin_ref_agrees_with_get_coins ... FAILED

    failures:

    failures:
        test_get_object_ab_coin_ref_agrees_with_get_coins

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 9 filtered out; finished in 12.52s
    
  stderr ───

    thread 'test_get_object_ab_coin_ref_agrees_with_get_coins' (67172055) panicked at crates/sui-indexer-alt-e2e-tests/tests/jsonrpc_address_balance_coin_tests.rs:356:5:
    assertion `left == right` failed: getObject must return the encoded withdrawal digest, not a hash of the synthesized object
      left: o#HMHZtRUoY6fW19JB4Nee6XXXEdoPMKMtDyeLHd6TdkfX
     right: o#3px89oUYY7nzA43vQbiM2wcJbwPoNrLAwYsjnoFR9iVH

  Cancelling due to test failure: 
────────────
     Summary [  12.599s] 1 test run: 0 passed, 1 failed, 9 skipped
        FAIL [  12.598s] sui-indexer-alt-e2e-tests::jsonrpc_address_balance_coin_tests test_get_object_ab_coin_ref_agrees_with_get_coins
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-jsonrpc/src/api/objects/response.rs
+++ b/crates/sui-indexer-alt-jsonrpc/src/api/objects/response.rs
@@ -50,25 +50,25 @@ pub(super) async fn live_object(
     object_id: ObjectID,
     options: &SuiObjectDataOptions,
 ) -> Result<SuiObjectResponse, RpcError> {
-    let object = if let Some(object) = load_live(ctx, object_id)
+    if let Some(object) = load_live(ctx, object_id)
         .await
         .context("Failed to load latest object")?
     {
-        return Ok(SuiObjectResponse::new_with_data(
+        Ok(SuiObjectResponse::new_with_data(
             object_data_with_options(ctx, object, options).await?,
-        ));
+        ))
     } else if let Some(coin) = AddressBalanceCoin::by_object_id(ctx, object_id)
         .await
         .context("Failed to resolve address balance object")?
     {
-        return Ok(SuiObjectResponse::new_with_data(
+        Ok(SuiObjectResponse::new_with_data(
             coin.into_sui_object_data(ctx, options).await?,
-        ));
+        ))
     } else {
-        return Ok(SuiObjectResponse::new_with_error(
+        Ok(SuiObjectResponse::new_with_error(
             SuiObjectResponseError::NotExists { object_id },
-        ));
-    };
+        ))
+    }
 }
 
 /// Fetch the necessary data from the stores in `ctx` and transform it to build a response for a
--- a/crates/sui-indexer-alt-jsonrpc/src/api/objects/response.rs
+++ b/crates/sui-indexer-alt-jsonrpc/src/api/objects/response.rs
@@ -33,8 +33,8 @@ use sui_types::object::Object;
 use tokio::join;
 
 use crate::context::Context;
+use crate::data::AddressBalanceCoin;
 use crate::data::load_live;
-use crate::data::try_resolve_address_balance_object;
 use crate::error::InternalContext;
 use crate::error::RpcError;
 use crate::error::rpc_bail;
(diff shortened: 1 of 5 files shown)


crates/sui-indexer-alt-jsonrpc/src/data/address_balance_coins.rs
```diff
@@ -112,11 +125,25 @@ pub(crate) async fn try_resolve_address_balance_object(
         return Ok(None);
     }
 
+    let epoch = super::latest_epoch(ctx).await?;
```
amnn: Make this request concurrently with the `load_live`.

crates/sui-indexer-alt-jsonrpc/src/api/objects/response.rs
amnn: It would be better to move this up into the first branch, otherwise the control flow is kind of confusing/lop-sided here.

crates/sui-indexer-alt-jsonrpc/src/api/objects/response.rs
```diff
@@ -55,11 +55,17 @@ pub(super) async fn live_object(
...
+    } else if let Some(coin) = try_resolve_address_balance_object(ctx, object_id)
         .await
         .context("Failed to resolve address balance object")?
     {
-        object
+        // The fabricated object only renders the coin's contents; (version, digest) must come
+        // from the encoded withdrawal ref so `getObject` agrees with `suix_getCoins` and returns
+        // a spendable reference.
+        let mut data = object_data_with_options(ctx, coin.contents, options).await?;
+        data.version = coin.object_ref.1;
```
amnn: Isn't the version already set correctly to the accumulator version?

crates/sui-indexer-alt-jsonrpc/src/data/address_balance_coins.rs
```diff
@@ -19,6 +20,17 @@ use sui_types::object::Owner;
...
 
+/// Struct to support a view of an address balance as a coin.
+///
+/// An address balance is an entry folded into a balance accumulator. However, consumers of of
+/// jsonrpc-alt expect `ObjectRef`s, so the implementation spoofs one using a masked ID, the
+/// accumulator object's version, and the digest field repurposed as a carrier for (epoch, amount).
+pub(crate) struct AddressBalanceCoin {
+    /// Fabricated `Coin` object to render the balance as if it were a coin.
+    pub(crate) contents: Object,
+    pub(crate) object_ref: ObjectRef,
```
amnn: Let's just capture the `digest` here, if that's all we need?

amnn (approved):
Thanks for fixing @wlmyng -- the fix looks good, but could be cleaned up a bit (see inline comments). Is there also a way to use the same code path for the object load and the coin load paths to avoid the discrepancy by construction?

crates/sui-indexer-alt-jsonrpc/src/api/objects/response.rs
wlmyng: Yes, for the live loaded object. Although this comment made me realize that a discrepancy could arise on `get_coins` path. This is because the address balance is fetched from consistent store, but the version comes from the live accumulator object. It's possible for the two to drift. So I think for correctness, both read paths need to read the live accumulator object.

Address balances is still useful for the getBalance/ getAllBalances path

PR #27549, opened 2026-08-03
reduce log spam in notify_read

alex-mysten:
## Description

Three private testnet validators filled their disks with logs, and ~70% of the collective ~2.4TB was this one `warn!`. Every blocked read logs its own keys on every 10s tick, so a stall is narrated independently by each waiter: on `sgp-ptn3-val-33`, 99.6% of an 836GB syslog was this line and the two callers it fronts. 

After this change, we log one blocked read per task every 30s and a sampled set of 32 keys that blocked read is waiting on. This provides information to debug issues while preventing repeated log spam. Worst case is now ~30 lines/min across all 15 task names, versus unbounded. We now log within the same task, removing the shared state that crossed the task boundary.

## Test plan

UTs added.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/mysten-common/src/sync/notify_read.rs
+++ b/crates/mysten-common/src/sync/notify_read.rs
@@ -5,17 +5,14 @@ use crate::ZipDebugEqIteratorExt;
 use crate::debug_fatal;
 
 use futures::future::{Either, join_all};
-use mysten_metrics::spawn_monitored_task;
 use parking_lot::Mutex;
 use parking_lot::MutexGuard;
 use std::collections::HashMap;
-use std::collections::HashSet;
 use std::collections::hash_map::DefaultHasher;
 use std::future::Future;
 use std::hash::{Hash, Hasher};
 use std::mem;
 use std::pin::Pin;
-use std::sync::Arc;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 use std::task::{Context, Poll};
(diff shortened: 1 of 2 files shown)


crates/mysten-common/src/sync/notify_read.rs
```diff
@@ -166,80 +170,69 @@ impl<K: Eq + Hash + Clone + Unpin + std::fmt::Debug + Send + Sync + 'static, V:
...
+        loop {
+            tokio::select! {
+                values = &mut join => return values,
+                _ = interval.tick() => {
+                    let elapsed_secs = start_time.elapsed().as_secs();
+
+                    // Deduplicate logging by task name. When many reads are blocked,
+                    // logs will sample blocked reads per task and the keys we are
+                    // waiting on for those reads.
+                    if self.throttle_stall_log(task_name) {
```
mystenmark, line 211: this seems fine, but we could also move the throttling functionality into a macro, to make it easy to re-use elsewhere. there are some crates that do this already (e.g. https://docs.rs/throttled-tracing/latest/throttled_tracing/) but they can't handle the `task_name` key as far as i can tell. Up to you, just a suggestion

alex-mysten: This one would be better suited: https://crates.io/crates/tracing-throttle
but I'd rather ship this as is. Not convinced the 20-some lines of log-throttling-related code here justifies a new dependency, and the strategy here of throttling by task name is not likely to be reusable.

mystenmark (approved):

PR #27551, opened 2026-08-03
fix(bridge): update observed action sequence metrics

grandpig:
## Description 

Fix `bridge_last_observed_actions_seq_num` not being updated by the bridge watchers.

Both the Sui gRPC watcher and the Ethereum watcher previously called `with_label_values(...)` without setting the returned gauge. As a result, labeled metric series were initialized at zero and never reflected the sequence number of the observed action.

This change sets the gauge to `action.seq_number()` for the corresponding
`chain_id` and `action_type`.

## Test plan 

- Extended the Sui gRPC watcher test to verify the observed action sequence metric.
- Extended the Ethereum watcher test to verify the observed action sequence metric.
- Ran the relevant `sui-bridge` orchestrator tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-bridge/src/orchestrator.rs
+++ b/crates/sui-bridge/src/orchestrator.rs
@@ -125,10 +125,12 @@ where
 
                 // Convert to action using the same flow as JSON-RPC watcher
                 if let Some(mut action) = bridge_event.try_into_bridge_action() {
-                    metrics.last_observed_actions_seq_num.with_label_values(&[
-                        action.chain_id().to_string().as_str(),
-                        action.action_type().to_string().as_str(),
-                    ]);
+                    let chain_id = action.chain_id().to_string();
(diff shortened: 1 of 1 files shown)


grandpig:
@bmwill Hi! Would you mind taking a look at this small bridge watcher fix when you have a chance? The watchers currently create the labeled gauge without updating its value; this sets it to the observed action sequence number and adds regression coverage for both the Sui and Ethereum watcher paths. Thanks!

grandpig:
@bmwill Could you please take a look when you have a chance? Thanks!

PR #27556, opened 2026-08-03
sui-rpc-api: emit start frames for unfiltered transactions and events

nickvikeras:
## Description

When implementing filtered subscriptions, we decided to include a starting position as the initial frame. We intended to include this for all subscriptions other than the older unfiltered SubscribeCheckpoints API to avoid breaking client expectations there. There was a bug that caused the server to also omit the start frame for unfiltered txn/event subscriptions, which this PR fixes.

### DIFF
--- a/crates/sui-rpc-api/src/subscription/matcher.rs
+++ b/crates/sui-rpc-api/src/subscription/matcher.rs
@@ -223,7 +223,9 @@ struct Subscriber {
     guard: SubscriptionLifecycleGuard,
     /// Checkpoints processed since this subscriber last received any frame.
     checkpoints_since_frame: u32,
-    /// Whether a filtered subscriber still needs its initial progress frame.
+    /// Whether this subscriber still needs its initial progress frame. Enabled
+    /// for every transaction and event subscriber and filtered checkpoint
+    /// subscribers.
     needs_start_frame: bool,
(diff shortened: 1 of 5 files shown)


bmwill (approved):

crates/sui-rpc-api/src/subscription/matcher.rs
```diff
@@ -335,13 +340,14 @@ impl SubscriptionMatcher {
...
 
-    /// Match `checkpoint` against every subscription and deliver one
-    /// [`SubscriptionUpdate`] per subscriber that either matched or is due a
-    /// progress frame after `interval` checkpoints without one. Subscribers
-    /// whose channel is full or closed are dropped; returns how many departed
-    /// this way. `keys` holds the checkpoint's pre-extracted dimension keys
-    /// (see [`extract_checkpoint_keys`]); a key space must be present
-    /// whenever this matcher holds a filtered subscriber in that space.
+    /// Match `checkpoint` against every subscription. A subscriber may receive
+    /// its one-time start progress frame before a match from the same
```
wlmyng, line 344: Is it not the case that for some filtered subscription whose first observed checkpoint  is > 0, it should receive the start progress frame first, before the payload frame?

wlmyng (approved):
lgtm

PR #27557, opened 2026-08-04
sui-indexer-alt-reader: Remove direct Bigtable access

evan-wall-mysten:
## Description

`LedgerGrpcReader` has reached feature parity with `BigtableReader` for all the KV lookups and watermark reporting that `sui-indexer-alt-graphql` and `sui-indexer-alt-jsonrpc` need (batch chunking, metrics, decoding-size config, balance changes), so neither service needs direct GCP Bigtable credentials/access anymore - both rely entirely on the Ledger gRPC service (or Postgres). With no more callers, `BigtableReader` and its `KvLoader`/`TransactionContents` variants are removed from `sui-indexer-alt-reader`.

## Test plan

Verified `cargo check --workspace` builds cleanly. Covered by existing unit and e2e tests in `sui-indexer-alt-reader`, `sui-indexer-alt-jsonrpc`, and `sui-indexer-alt-graphql`, which exercise the Ledger gRPC/Postgres paths both services now use exclusively.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [x] JSON-RPC: `sui-indexer-alt-jsonrpc` no longer accepts `--bigtable-instance`, `--bigtable-project`, `--bigtable-app-profile-id`, or `--bigtable-watermark-pipeline`. Operators must use `--ledger-grpc-url` instead.
- [x] GraphQL: `sui-indexer-alt-graphql` no longer accepts `--bigtable-instance`, `--bigtable-project`, `--bigtable-app-profile-id`, or `--bigtable-watermark-pipeline`. Operators must use `--ledger-grpc-url` instead.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-graphql/README.md
```diff
@@ -16,12 +16,13 @@ GraphQL can run with access to just a postgres database, written to by
...
 
-If GraphQL is given access to credentials for Bigtable, it will look there to
-answer key-value queries (fetching an object by its ID and version, a
-checkpoint by its sequence number, or a transaction by its digest), instead of
-the relevant postgres tables.
+If GraphQL is given a Ledger gRPC service URL (`--ledger-grpc-url`), it will
+look there to answer key-value queries (fetching an object by its ID and
+version, a checkpoint by its sequence number, or a transaction by its
+digest), instead of the relevant postgres tables. This service is typically
+backed by an archival store (e.g. Bigtable).
```
wlmyng: since we only have two backends today, full node or archival store, we could just list them out

wlmyng (commented):
Is it not possible for us to remove direct bigtable access outright, and have jsonrpc-alt configured with KvLoader(LedgerService)?

evan-wall-mysten:
> Is it not possible for us to remove direct bigtable access outright, and have jsonrpc-alt configured with KvLoader(LedgerService)?

I saw we were going to remove sui-indexer-alt-jsonrpc in a few months so I thought I would leave it alone until then. I am fine removing it though if we are making changes to it still. @amnn is it ok to remove direct bigtable access from sui-indexer-alt-jsonrpc?

crates/sui-indexer-alt-graphql/README.md
evan-wall-mysten: Good call, done - spelled out both backends explicitly.

crates/sui-indexer-alt-reader/src/kv_loader.rs
```diff
@@ -46,40 +44,21 @@ use crate::pg_reader::PgReader;
...
-    #[arg(long)]
-    pub bigtable_project: Option<String>,
-
-    /// App profile ID to use for Bigtable client. If not provided, the default profile will be used.
-    #[arg(long)]
-    pub bigtable_app_profile_id: Option<String>,
-
     /// Maximum gRPC decoding message size for KV responses, in bytes.
     ///
     /// Applies to both Bigtable and Ledger gRPC readers.
```
amnn: nit: a remaining reference to `Bigtable`.

amnn (commented):
Is it feasible to remove all the uses of Bigtable Reader first (e.g. from JSONRPC-alt) and then clean everything up in one shot? It seems like that would be less churn-y and we could end up cleaning more from the reader crate in one shot that way.

evan-wall-mysten:
> Is it feasible to remove all the uses of Bigtable Reader first (e.g. from JSONRPC-alt) and then clean everything up in one shot? It seems like that would be less churn-y and we could end up cleaning more from the reader crate in one shot that way.

Yeah, I squashed https://github.com/MystenLabs/sui/pull/27568 into this PR because both depend on https://github.com/MystenLabs/sui-operations/pull/8679.

crates/sui-indexer-alt-reader/src/kv_loader.rs
evan-wall-mysten: Removed after combining with the other PR.

amnn (approved):
Accepting pending the ops change, thanks @evan-wall-mysten

PR #27564, opened 2026-08-04
[graphql-stream] per-subscriber delivery throttle for subscriptions [17/n]

tpham-mysten:
## Description

A per-subscriber delivery throttle for GraphQL subscriptions. The SSE handler paces each payload by its cost (`cost / rate` seconds), so a subscriber's sustained delivery stays within `per_subscriber_max_output_nodes_per_second` output nodes per second (default ~off, an operator opts in by lowering it). A payload's cost is its output nodes plus a query-depth surcharge. The throttle lives in the handler, so it wraps every subscription type identically. 

## Test plan

Unit tests cover the cost model and `Throttle::wrap` pacing (deterministic via tokio's paused clock). `throttle_subscription.rs` adds e2e for checkpoint pacing/ordering, the disable path, richer-payload cost, and the transaction backfill path.

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425
   - #26495
   - #26731
   - #26714
   - #27140
   - #27537
   - #27564

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/extensions/query_limits/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/extensions/query_limits/mod.rs
@@ -42,12 +42,13 @@ mod visitor;
 pub(crate) struct QueryDepth(Arc<AtomicU32>);
 
 impl QueryDepth {
-    pub(crate) fn set(&self, depth: u32) {
-        self.0.store(depth, Ordering::Relaxed);
-    }
-
     pub(crate) fn get(&self) -> u32 {
         self.0.load(Ordering::Relaxed)
+    }
+
+    #[cfg(test)]
+    pub(crate) fn new_for_test(depth: u32) -> Self {
+        Self(Arc::new(AtomicU32::new(depth)))
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/sui-indexer-alt-graphql/src/extensions/query_limits/mod.rs
```diff
@@ -193,6 +209,12 @@ impl Extension for QueryLimitsCheckerExt {
         self.metrics.input_nodes.observe(input.nodes as f64);
         self.metrics.output_nodes.observe(output.nodes as f64);
 
+        // Stash the validated depth so the subscription handler can add its depth surcharge to each
+        // payload's throttle cost.
+        if let Some(depth) = ctx.data_opt::<QueryDepth>() {
+            depth.set(input.depth);
+        }
```
amnn: nit: 

```suggestion
        if let Some(QueryDepth(depth)) = ctx.data_opt() {
            depth.store(input.depth, Ordering::Relaxed);
        }
```

I would personally opt for this, and then only expose `get`. I see that `QueryDepth::set` is used for testing, for which I would suggest a `new_for_test` instead of exposing a setter, which should not be broadly available.

amnn (approved):
I think this needs a rebase, but the change in the top commit looks good!

crates/sui-indexer-alt-graphql/src/extensions/query_limits/mod.rs
tpham-mysten: SG. Updated.

tpham-mysten (commented):
Thanks @amnn! I have rebased and applied the suggestion!

PR #27566, opened 2026-08-04
sui-proxy: fix pending_epoch_change parser for struct type

chris-gorham:
The on-chain Hashi `CommitteeSet.pending_epoch_change` field is `Option<PendingEpochChange>` (a struct with an `epoch` field), but the parser assumed `Option<u64>`. When the field is non-null during epoch handoff, the JSON renderer outputs a struct object which fails to parse as a string-encoded u64, breaking hashi allowlist refreshes on all sui-proxy testnet pods every ~90s.

Replace `json_option_u64` with `json_pending_epoch` which handles null, struct (extracting the `epoch` field), and bare string u64 (backward compat).

## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-proxy/src/peers.rs
+++ b/crates/sui-proxy/src/peers.rs
@@ -707,11 +707,25 @@ fn json_u64(v: &JsonValue, name: &str) -> Result<u64> {
         .with_context(|| format!("parsing {name} as u64"))
 }
 
-/// Parse a Move `Option<u64>`: JSON null = None, bare value = Some.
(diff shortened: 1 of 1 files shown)


julio-ml:
Good one, only fix the rustfmt, because it silently skips tests

julio-ml (approved):

PR #27567, opened 2026-08-04
Make periodic compaction explicit and configurable

nickvikeras:
## Description

RocksDB implicitly schedules 30-day periodic compaction for filtered, leveled, block-based column families. Pin that behavior for the RPC bitmap column families and expose an optional node override so reclamation does not depend on RocksDB auto-tuning.

## Test plan

Unit tests cover option layering, day-to-second conversion, bitmap-CF targeting, and the two-pass merge reclamation behavior.

---

## Release notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Fullnode operators can override the embedded RPC store's bitmap periodic-compaction interval. Omission retains 30 days; zero disables periodic compaction.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-config/src/node.rs
+++ b/crates/sui-config/src/node.rs
@@ -1331,6 +1331,17 @@ pub struct AuthorityStorePruningConfig {
         skip_serializing_if = "Option::is_none"
     )]
     pub periodic_compaction_threshold_days: Option<usize>,
+    /// Optional periodic-compaction interval override for the embedded
+    /// RPC store's transaction and event bitmap SSTs, in days. When
+    /// omitted, the RPC store's RocksDB configuration uses its 7-day
+    /// default. Zero disables periodic compaction; positive values are
+    /// the SST-age interval.
+    ///
+    /// Expired merge-written buckets may need one interval to
+    /// materialize and another to be filtered, so this is not a
+    /// wall-clock deletion SLA.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub rpc_store_bitmap_periodic_compaction_days: Option<u64>,
     /// number of epochs to keep the latest version of transactions and effects for
     #[serde(skip_serializing_if = "Option::is_none")]
     pub num_epochs_to_retain_for_checkpoints: Option<u64>,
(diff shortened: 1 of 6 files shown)


amnn (commented):
Can you split the two changes up (at least as separate commits in the PR)? The configuration change makes sense to me and seems simple enough, but most of this PR seems to actually be about changing the way the pruning floor atomic is initialised and propagated, and I need some more context on that:

- Why is it `Schema::cfs`' responsibility to initialise the `OpenContext`?
- Can the type be defaulted to `()` to avoid some boiler-plate?

nickvikeras:
> Can you split the two changes up (at least as separate commits in the PR)? The configuration change makes sense to me and seems simple enough, but most of this PR seems to actually be about changing the way the pruning floor atomic is initialised and propagated, and I need some more context on that:
> 
> * Why is it `Schema::cfs`' responsibility to initialise the `OpenContext`?
> * Can the type be defaulted to `()` to avoid some boiler-plate?

Looks like associated type defaults are not stable yet - https://github.com/rust-lang/rust/issues/29661

> Why is it `Schema::cfs`' responsibility to initialise the `OpenContext`?

This is cleaning up a footgun. Previously the pruning floor was a global variable, so this code was not safe to use in contexts where there were multiple DBs in a single process. To fix this, we need to have one Atomic floor per DB.

`Schema::cfs` is where the [compaction filter callback is initialized](https://github.com/MystenLabs/sui/blob/48e76d2d9b461a72b35eac897915dd2ab42b2fc6/crates/sui-rpc-store/src/schema/transaction_bitmap.rs?plain=1#L82-L100), and that callback needs to capture the pruning floor Atomic. An alternative to initializing it here would be to change the API so the context is passed in as an argument rather than initialized and returned. i.e.

```rust
let context = S::create_open_context();
let cfs = S::cfs(&resolver, &context);
// Open RocksDB.
let schema = S::open(&db, context);
```

I don't really have a strong preference one way or another. Let me know if that makes sense and if you prefer this alternative or have another idea.

Edit - moved this part of the PR to https://github.com/MystenLabs/sui/pull/27570

amnn (approved):
Thanks -- will take a look and respond on the other PR! This one looks good.

PR #27570, opened 2026-08-04
Scope bitmap pruning floors to each RPC store

nickvikeras:
## Description

The bitmap compaction filters previously shared a process-global pruning floor, allowing one RPC-store database or a cleared store to expose another database lifecycle's floor. Scope that state to each opened database and initialize it from the persisted pruning watermark.

Stacked on #27567; this PR contains the DB-local pruning-floor change only.

## Test plan

Tests cover independent databases, reopen initialization, clear-and-rebuild reset, restore floor publication, and production-shaped bitmap compaction.

---

## Release notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Embedded RPC-store bitmap pruning state is isolated to each database. No operator action is required.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-consistent-store/src/batch.rs
+++ b/crates/sui-consistent-store/src/batch.rs
@@ -64,11 +64,18 @@
 //! }
 //!
 //! impl Schema for MySchema {
-//!     fn cfs(opts: &sui_consistent_store::CfOptionsResolver) -> Vec<sui_consistent_store::CfDescriptor> {
-//!         vec![sui_consistent_store::CfDescriptor::new("items", opts.options("items"))]
+//!     type OpenContext = ();
+//!
+//!     fn cfs(
+//!         opts: &sui_consistent_store::CfOptionsResolver,
+//!     ) -> (Vec<sui_consistent_store::CfDescriptor>, Self::OpenContext) {
+//!         (
+//!             vec![sui_consistent_store::CfDescriptor::new("items", opts.options("items"))],
+//!             (),
+//!         )
 //!     }
 //!
-//!     fn open(db: &Db) -> Result<Self, OpenError> {
+//!     fn open(db: &Db, (): Self::OpenContext) -> Result<Self, OpenError> {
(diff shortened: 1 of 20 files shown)


amnn (commented):
I think we could simplify the `Schema` trait (simpler than it was before this change) and get what we want:

```rust
trait Schema {
    fn open(path: &Path, opts: &CfOptionsResolver) -> Result<(Db, Self), OpenError>;
}

impl Db {
    pub fn open_cfs(
        path: &Path, 
        opts: &CfOptionsResolver,
        cfs: Vec<CfDescriptor>
    ) -> Result<Self, OpenError>;
}
```

We make it the `Schema`'s responsibility to open the DB with the appropriate `cfs` which means it is now free to do whatever other state initialisation it wants to do around that.

nickvikeras:
> I think we could simplify the `Schema` trait (simpler than it was before this change) and get what we want:
> 
> ```rust
> trait Schema {
>     fn open(path: &Path, opts: &CfOptionsResolver) -> Result<(Db, Self), OpenError>;
> }
> 
> impl Db {
>     pub fn open_cfs(
>         path: &Path, 
>         opts: &CfOptionsResolver,
>         cfs: Vec<CfDescriptor>
>     ) -> Result<Self, OpenError>;
> }
> ```
> 
> We make it the `Schema`'s responsibility to open the DB with the appropriate `cfs` which means it is now free to do whatever other state initialisation it wants to do around that.

ty that's better. I plumbed `snapshot_capacity` through as it's own argument to `Schema::open` rather than tacking it onto `CfOptionsResolver` because `CfOptionsResolver` is currently scoped to RocksDB-level options, whereas snapshot_capacity is a higher-level consistent store config.

amnn (approved):
LGTM, thanks @nickvikeras !

PR #27379, opened 2026-07-21
docs: add additional payments documentation 

jessiemongeon1:
adds a few additional docs on payment models and address aliases, does not add any new code snippets or examples (reuses some existing ones from  address balances/gasless transfers docs)

PR #27573, opened 2026-08-04
docs: add gas station page (TypeScript)

jessiemongeon1:
## Summary
Add `gas-station.mdx` with TypeScript examples for coin pool management, sponsored transaction server, and client integration.

- Use `@mysten-incubation/sponsor` SDK instead of coin management (`package.json:1`)
- **CRITICAL** (`server.ts:47`): Server signs attacker-controlled tx kind without inspection. Must reject GasCoin by value, enforce allowlist, dry-run before signing.

## Test plan
- [ ] Address critical security issue in server.ts before merge
- [ ] Visual review of gas station page

### DIFF
--- a/examples/onchain-finance/gas-station/src/agent-sponsor.ts
+++ b/examples/onchain-finance/gas-station/src/agent-sponsor.ts
@@ -1,0 +1,34 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import type express from 'express';
+
+// Placeholder types for the agent sponsorship example
+declare function verifyApiKey(apiKey: string): Promise<{ address: string; dailyGasBudget: number } | null>;
+declare function getAgentDailySpend(address: string): Promise<number>;
+
+const GAS_BUDGET = 10_000_000;
+
+// docs::#agent-sponsor
+async function handleAgentSponsor(req: express.Request, res: express.Response) {
+	const { txBytes, apiKey } = req.body;
+
+	// Verify agent identity
+	const agent = await verifyApiKey(apiKey);
+	if (!agent) {
+		res.status(401).json({ error: 'Invalid API key' });
+		return;
+	}
+
+	// Check per-agent daily budget
+	const todaySpend = await getAgentDailySpend(agent.address);
+	if (todaySpend + GAS_BUDGET > agent.dailyGasBudget) {
+		res.status(429).json({ error: 'Daily gas budget exceeded' });
+		return;
+	}
+
+	// ... same sponsorship flow as the /sponsor endpoint
+}
+// docs::/#agent-sponsor
+
+export { handleAgentSponsor };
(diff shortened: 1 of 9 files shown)


examples/onchain-finance/gas-station/src/agent-sponsor.ts
```diff
@@ -0,0 +1,49 @@
...
+		return;
+	}
+
+	// From here the flow is identical to /sponsor: validate the transaction
+	// kind against the allowlist, acquire a gas coin, set gas data, simulate,
+	// then sign.
+	try {
+		res.json(await sponsorTransaction(txBytes, agent.address));
+	} catch (error) {
+		res.status(400).json({ error: (error as Error).message });
```
hayes-mysten: this error handling looks overly broad for a 400 error

examples/onchain-finance/gas-station/src/client.ts
```diff
@@ -0,0 +1,100 @@
...
+        Authorization: `Bearer ${apiKey}`,
+        "Content-Type": "application/json",
+    },
+    body: JSON.stringify({ txBytes: toBase64(kindBytes), sender: user.toSuiAddress() }),
+});
+if (!response.ok) throw new Error(`Sponsorship rejected: ${response.status}`);
+
+const sponsored = (await response.json()) as Record<string, unknown>;
+if (
+    typeof sponsored.txBytes !== "string" ||
```
hayes-mysten: we should probably just use a schema validation library, this feels like a bad way to do validation

examples/onchain-finance/gas-station/src/client.ts
```diff
@@ -0,0 +1,100 @@
...
+    typeof sponsored.sponsorAddress !== "string" ||
+    typeof sponsored.gasCoinId !== "string" ||
+    typeof sponsored.digest !== "string"
+) {
+    throw new Error("Invalid gas station response");
+}
+
+// Verify everything the gas station added before crossing the signing boundary.
+const finalBytes = fromBase64(sponsored.txBytes);
+const finalData = TransactionDataBuilder.fromBytes(finalBytes);
```
hayes-mysten: we should probably be using the Transaction class not TransactionDataBuilder directly

hayes-mysten (changes requested):
I am going to hold off on reviewing this one, I think if we want to have a gas station example, it should be based on the new sponsorship sdk which is so much simpler than what we have here, allows parallel transactions, should be mostly stateless, and doesn't require complex coin management.

examples/onchain-finance/gas-station/src/agent-sponsor.ts
```diff
@@ -0,0 +1,65 @@
...
+	sender: string,
+): Promise<{
+	txBytes: string;
+	sponsorSignature: string;
+	sponsorAddress: string;
+	gasCoinId: string;
+	digest: string;
+}>;
+
+class ValidationError extends Error {
```
hayes-mysten, line 22: what is this? its not thrown anywhere... I think the intent below might be to handle validation errors from the SDK, but this is not the same thing

hayes-mysten (changes requested):
this still had all the gas pool and other outdated sponsorship patterns

PR #27574, opened 2026-08-04
docs: add P2P reference app page (TypeScript)

jessiemongeon1:
## Summary
Add `p2p-reference-app.mdx` with inline TypeScript examples for end-to-end payment flow: auth, sponsorship, transfers, and GraphQL history.

- Use GraphQL variables rather than string interpolation (`L243`)
- Use `balanceChangesJson` instead of paginated `balanceChanges` (`L248`)

## Test plan
- [ ] Address GraphQL variable and balanceChangesJson feedback
- [ ] Visual review of P2P reference page

### DIFF
--- a/docs/content/onchain-finance/p2p-reference-app.mdx
+++ b/docs/content/onchain-finance/p2p-reference-app.mdx
@@ -1,0 +1,336 @@
+---
+title: P2P Payments Reference App
+description: >-
+  End-to-end walkthrough of a peer-to-peer payments application on Sui that
+  stitches together authentication, gas sponsorship, token transfers, and
+  transaction history.
+sidebar_label: P2P Reference App
+keywords:
+  - P2P payments
+  - reference app
+  - peer-to-peer
+  - payments example
+  - zkLogin
+  - sponsored transactions
+  - transaction history
+  - payment flow
+goal:
+  description: >-
+    Reader can build a complete P2P payments app on Sui with auth, sponsored
+    gas, transfers, and transaction history
+  requires:
+    - has_frontmatter:
+        - title
+        - description
+        - keywords
+      label: Has required frontmatter fields
+    - min_words: 300
+      label: Needs more content depth
+    - pattern: '```'
+      min: 1
+      label: Has code examples
+    - has_questions: true
+      label: Needs questions for AI search visibility
+    - has_answer: true
+      label: Needs answer summary for AI citation
+builder_paths:
+  - path_id: p2p-payments
+    path_name: P2P Payments / Neomoney
+    step: P2P reference app
+    stage: End to End
+questions:
+  - How do I build a P2P payments app on Sui?
+  - What does an end-to-end payment flow look like on Sui?
+  - How do I combine zkLogin, sponsored gas, and transfers in one app?
+answer: >-
+  A P2P payments app on Sui combines zkLogin for wallet-free authentication,
+  sponsored transactions for gasless UX, address balance transfers for the
+  actual payment, and gRPC queries for transaction history, all composable
+  within a single programmable transaction block.
+---
+
+This walkthrough builds a peer-to-peer payments flow that lets users send stablecoins to each other without holding SUI or managing private keys. It stitches together four primitives:
+
+1. **Authentication:** zkLogin or wallet connection
+2. **Gas sponsorship:** a gas station pays the fee
+3. **Transfer:** address balance `send_funds` for the actual payment
+4. **Transaction history:** gRPC query for past payments
+
+Each section is self-contained. You can adopt individual pieces or wire the full flow.
+
+## Architecture
+
+```mermaid
+sequenceDiagram
+    participant U as User (browser)
+    participant F as Frontend (React)
+    participant G as Gas Station (Express.js)
+    participant N as Sui Node (gRPC)
+    U->>F: 1. zkLogin auth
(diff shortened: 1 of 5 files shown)


docs/content/onchain-finance/p2p-reference-app.mdx
```diff
@@ -0,0 +1,334 @@
...
+	transaction: finalBytes,
+	signatures: [userSig.signature, sponsorSignature],
+});
+
+// Confirm to the gas station so it can release the coin.
+// Always confirm, even on failure, because gas is still charged.
+const txResult = result.Transaction ?? result.FailedTransaction;
+const digest = txResult.digest;
+
+await fetch('https://your-gas-station.com/sponsor/confirm', {
```
hayes-mysten, line 188: what is this sponsorship flow based on? this looks different than enoki sponsorship and the newer address balance style funding @manolisliolios has been working on (eg stablestation)

docs/content/onchain-finance/p2p-reference-app.mdx
```diff
@@ -0,0 +1,334 @@
...
+
+Wait for finality and verify the transaction effects before showing success to the user.
+
+```typescript
+if (result.$kind === 'FailedTransaction') {
+	console.error('Payment failed:', result.FailedTransaction.effects.status.error);
+	// Show error UI
+} else {
+	// Wait for the transaction to be indexed
+	await client.waitForTransaction({ digest: result.Transaction.digest });
```
hayes-mysten, line 214: we should do `	await client.waitForTransaction({ result })` first, then check for FailedTransaction after so we always await

docs/content/onchain-finance/p2p-reference-app.mdx
```diff
@@ -0,0 +1,334 @@
...
+// Use balanceChangesJson (not paginated balanceChanges.nodes)
+// so all changes are returned in a single JSON array.
+import { SuiGraphQLClient } from '@mysten/sui/graphql';
+
+const gqlClient = new SuiGraphQLClient({
+	url: 'https://sui-testnet.mystenlabs.com/graphql',
+	network: 'testnet',
+});
+
+interface BalanceChange {
```
hayes-mysten, line 235: We should not be explicitly typing graphql query results like this, it should be using the typed `qraphql` helper the sdk exports that will type the query for you automatically

docs/content/onchain-finance/p2p-reference-app.mdx
```diff
@@ -0,0 +1,334 @@
...
+For higher-volume history queries or filtered views (by coin type, date range, or amount), consider building a [custom indexer](/develop/accessing-data/custom-indexer/custom-indexers) that listens for payment events and stores them in a queryable database.
+
+## Putting it together
+
+The full flow in a single function:
+
+```typescript
+async function sendPayment(
+	client: SuiGrpcClient,
+	signer: Keypair,
```
hayes-mysten, line 299: We should probably accept Signer instead of Keypair most places, since this will allow it to work with wallet signers, kms signers, etc without code changes

docs/content/onchain-finance/p2p-reference-app.mdx
```diff
@@ -0,0 +1,334 @@
...
+
+	// Submit (SDK auto-detects gasless eligibility with gRPC)
+	const result = await client.signAndExecuteTransaction({ transaction: tx, signer });
+
+	if (result.$kind === 'FailedTransaction') {
+		throw new Error(`Payment failed: ${result.FailedTransaction.effects.status.error}`);
+	}
+
+	// Wait for indexing
+	await client.waitForTransaction({ digest: result.Transaction.digest });
```
hayes-mysten, line 322: some comment about re-ordering awaits

examples/onchain-finance/p2p-reference/src/history.ts
```diff
@@ -0,0 +1,73 @@
...
+				digest: string;
+				effects?: { balanceChangesJson?: BalanceChange[] };
+			}>;
+		};
+	};
+}
+
+const { data } = await gqlClient.query<HistoryQueryResult>({
+	variables: { sender: senderAddress },
+	query: `query P2PPaymentHistory($sender: SuiAddress!) {
```
hayes-mysten, line 35: same query typing comment here

hayes-mysten: We also have transaction history in grpc now, we should consider if we really want to mix clients in our examples like this now

examples/onchain-finance/p2p-reference/src/sponsor.ts
```diff
@@ -0,0 +1,54 @@
...
+// Submit with both signatures
+const result = await client.executeTransaction({
+	transaction: finalBytes,
+	signatures: [userSig.signature, sponsorSignature],
+});
+
+// Confirm to the gas station so it can release the coin.
+// Always confirm, even on failure, because gas is still charged.
+const txResult = result.Transaction ?? result.FailedTransaction;
+await fetch('https://your-gas-station.com/sponsor/confirm', {
```
hayes-mysten, line 42: I am not sure about this sponsorship pattern, I don't think this is really how any real gas stations will work (we don't have a standard, so this could be built, but its probably not the flow we would recommend anyone use)

jessiemongeon1:
## Style Guide Audit

PR #27575, opened 2026-08-04
docs: add x402 pay-per-request page (TypeScript)

jessiemongeon1:
## Summary
Add `x402-pay-per-request.mdx` with TypeScript examples for HTTP 402 payment flow, challenge-signature binding, server and client.

- Update SDK dependency (`package.json` flagged as "old sdk")

## Test plan
- [ ] Update SDK version
- [ ] Visual review of x402 page


hayes-mysten (commented):
where does this x402 implementation come from?  this looks very different than what I would recommend, and also feels out of line with how the official implementations work

docs/content/onchain-finance/x402-pay-per-request.mdx
```diff
@@ -0,0 +1,207 @@
...
+This command sends a probe request, inspects the 402 response headers, and reports whether your server's challenge is well-formed and discoverable.
+
+## Production considerations
+
+### Digest store
+
+Replace `InMemoryDigestStore` with a persistent store backed by Redis or Postgres. The store must guarantee that each transaction digest is accepted at most once, even across server restarts and horizontal scaling. The SDK provides a store interface you can implement with any backend:
+
+```ts
+import { RedisDigestStore } from '@suimpp/mpp/server';
```
hayes-mysten, line 169: do we know who is behind this and do we want to standardize on this?  I think recommending something like this in the official docs elevates it significantly and positions it as the recommended solution.  I have not been watching this space closely, but I had not heard of this before now

hayes-mysten (changes requested):
This is a payment protocol that uses an 402 http status code, but it is NOT the x402 standard as far as I can tell

PR #27576, opened 2026-08-04
docs: add regulated payments page (TypeScript)

jessiemongeon1:
## Summary
Add `regulated-payments.mdx`

No TS-specific review comments from #27379.

## Test plan
- [ ] Visual review of regulated payments page
- [ ] Visual review of fiat integration page


examples/regulated-payments/src/batch-screening.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { Transaction } from '@mysten/sui/transactions';
+import { SuiGrpcClient } from '@mysten/sui/grpc';
+import { Ed25519Keypair } from '@mysten/sui/keypairs/ed25519';
+
+const client = new SuiGrpcClient({
+	baseUrl: 'https://fullnode.mainnet.sui.io:443',
+	network: 'mainnet',
+});
+const adminKeypair = new Ed25519Keypair();
```
hayes-mysten: we should maybe show fromSecretKey with an env variable here, this will never work since this creates a random unfunded address

examples/regulated-payments/src/off-ramp.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { Transaction } from '@mysten/sui/transactions';
+import { SuiGrpcClient } from '@mysten/sui/grpc';
+import { Ed25519Keypair } from '@mysten/sui/keypairs/ed25519';
+
+const client = new SuiGrpcClient({
+	baseUrl: 'https://fullnode.mainnet.sui.io:443',
+	network: 'mainnet',
+});
+const userKeypair = new Ed25519Keypair();
```
hayes-mysten: same thing here

examples/regulated-payments/src/verify-deposit.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const client = new SuiGrpcClient({
+	baseUrl: 'https://fullnode.mainnet.sui.io:443',
+	network: 'mainnet',
+});
+const userSuiAddress = '0xUSER_ADDRESS';
+
+// docs::#verify-deposit
+// Check balance after webhook notification.
+const { balance } = await client.core.getBalance({
```
hayes-mysten: we don't need to use the .core.* methods, the top level methods are fine for user code (`core` is mostly for SDKs that need to work across grpc/jsonrpc/graphql)

examples/regulated-payments/src/webhook-handler.ts
```diff
@@ -0,0 +1,65 @@
...
+	const signature = req.headers.get('X-Provider-Signature');
+	if (!verifyWebhookSignature(req.body, signature, providerPublicKey)) {
+		return new Response('Invalid signature', { status: 401 });
+	}
+
+	const payload = await req.json();
+
+	// Step 2: Verify onchain state.
+	const txDigest = payload.transactionDigest;
+	const txResult = await suiClient.core.getTransaction({
```
hayes-mysten: same here

examples/regulated-payments/src/webhook-handler.ts
```diff
@@ -0,0 +1,65 @@
...
+		include: { effects: true },
+	});
+
+	if (txResult.$kind !== 'Transaction') {
+		await complianceLogger.warn('Transaction failed onchain', payload);
+		return new Response('Onchain verification failed', { status: 422 });
+	}
+
+	// Verify the recipient received at least the expected amount.
+	const { balance } = await suiClient.core.getBalance({
```
hayes-mysten: and here

examples/regulated-payments/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "name": "regulated-payments-examples",
+  "version": "0.0.1",
+  "private": true,
+  "description": "Runnable TypeScript examples for regulated payments and fiat integration docs.",
+  "type": "module",
+  "scripts": {
+    "build": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@mysten/sui": "2.22.1"
```
hayes-mysten, line 11: We could consider using pnpm catalog versioning so that all examples reference a single version from the workspace package.json

hayes-mysten (approved):

examples/regulated-payments/src/webhook-handler.ts
```diff
@@ -0,0 +1,65 @@
...
+		return new Response('Onchain verification failed', { status: 422 });
+	}
+
+	// Verify the recipient received at least the expected amount.
+	const { balance } = await suiClient.getBalance({
+		owner: payload.recipientAddress,
+		coinType: payload.coinType,
+	});
+
+	if (BigInt(balance.balance) < BigInt(payload.amount)) {
```
hayes-mysten, line 53: this is kinda bad accounting.  I don't really see the intent behind this, but I am guessing the idea is something like the web hook tells us about transactions give us money and we want to verify that the transaction actually gets that.

This is checking if the receiving account as the amount being sent which may be true before any transaction was set.  It probably should be loading the transaction and looking at its balance changes in the effects instead?

jessiemongeon1:
## Style Guide Audit

PR #27579, opened 2026-08-04
Make mutex table acquisition execution-aware

lxfind:
## Description

Follow up to #27509. A duplicate transaction execution can wait on the per-transaction mutex while retaining an execution permit and, under msim, parking the active blocking-thread quantum. This can prevent the lock holder from making progress.

Make the existing `LockTable::acquire_lock` release any installed execution permit only after its uncontended fast path fails, and retry with `yield_blocking` under msim. Non-execution callers retain the same API, and permit release remains a no-op for them. The execution-driver test stabilization previously included here is now #27678.

## Test plan

`LockTable` tests cover uncontended permit retention, contended native/msim acquisition, and progress for distinct keys in the same shard under both native and msim.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:

### DIFF
--- a/crates/sui-storage/src/mutex_table.rs
+++ b/crates/sui-storage/src/mutex_table.rs
@@ -15,6 +15,9 @@ use tokio::task::JoinHandle;
 use tokio::time::Instant;
 use tracing::info;
 
+use mysten_common::sync::execution_permit::release_execution_permit;
+#[cfg(test)]
+use mysten_common::sync::execution_permit::set_execution_permit;
 use mysten_metrics::spawn_monitored_task;
 
 type OwnedMutexGuard<T> = ArcMutexGuard<parking_lot::RawMutex, T>;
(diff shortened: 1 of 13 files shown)


crates/sui-storage/src/mutex_table.rs
```diff
@@ -271,6 +272,14 @@ impl<K: Hash + Eq + Send + Sync + 'static, L: Lock + 'static> LockTable<K, L> {
     }
 }
 
+impl<K: Hash + Eq + Send + Sync + 'static> LockTable<K, Mutex<()>> {
+    /// Acquires a lock from a blocking execution thread, releasing its execution
+    /// permit if acquisition is contended.
+    pub fn acquire_lock_blocking(&self, k: K) -> MutexGuard {
```
mystenmark: the existing method was already blocking so i'm not sure we need a new name here? We can just add the yielding behavior to the old method.

crates/sui-storage/src/mutex_table.rs
```diff
@@ -237,8 +240,27 @@ impl<K: Hash + Eq + Send + Sync + 'static, L: Lock + 'static> LockTable<K, L> {
         }
     }
 
+    /// Acquires the lock for `k`. If acquisition is contended, releases the current
+    /// execution permit before blocking. Under msim, contended acquisition must run
```
mystenmark: it may be worth noting that trying to take a contended lock in an async thread is bad in release builds as well. simtest should expose such situations, but readers should understand that when that happens, its a bug, not a defect in msim

mystenmark (approved):

PR #27580, opened 2026-08-05
[graphql-stream] Resumable event subscriptions on a shared scan-then-live driver [18/n]

tpham-mysten:
## Description

Add resumption to the events subscription, mirroring the transactions subscription: it backfills the events matching a filter after a resume point through the scanning API, then hands off to the live stream with no gap and no duplicate. A subscriber resumes with `after` (an opaque cursor), `filter.afterCheckpoint`, or both applied together, in which case delivery resumes from whichever is later.

   In doing so, the scan-then-live machinery that was specific to transactions is extracted into a single driver (`scan_then_live.rs`) generic over a `Subscribable` trait, so transactions and events share one implementation rather than duplicating it. The alpha ledger gRPC reader the backfill relies on is now required at start-up when streaming is enabled, instead of surfacing as a request-time error.

## Test plan

Ports the transactions resume/handoff/error e2e coverage to events: resume-backfill-then-live, resume-by-cursor, empty-backfill handoff, exactly-once across the seam, live/backfill parity, and the invalid-cursor / mutually-exclusive-resume / checkpoint-filter error cases. The full `graphql_subscription` e2e suite passes.

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425
   - #26495
   - #26731
   - #26714
   - #27140
   - #27537
   - #27564
   - #27580

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
@@ -6,16 +6,23 @@
 //! `TestClusterBuilder` validator, neither of which works inside the simulator's
 //! deterministic runtime.
 
+use std::collections::BTreeMap;
 use std::collections::HashSet;
 use std::time::Duration;
 
 use serde_json::Value;
 use serde_json::json;
+use sui_types::base_types::ObjectID;
 use tokio_stream::StreamExt;
 
 use crate::testing::SubscriptionTestCluster;
 use crate::testing::emit_event_harness;
 use crate::testing::graphql_redactions;
+use crate::testing::wait_for_matching_item;
+
+/// How long to let the validator advance so a resumed subscription's live receiver pins past the
+/// pre-subscription events, forcing them through the backfill scan.
+const BACKFILL_SETTLE: Duration = Duration::from_secs(5);
 
 fn event_value(item: &Value) -> Option<&str> {
     item["data"]["events"]["node"]["contents"]["json"]["value"].as_str()
@@ -23,6 +30,84 @@ fn event_value(item: &Value) -> Option<&str> {
 
 fn event_bcs(item: &Value) -> Option<&str> {
     item["data"]["events"]["node"]["eventBcs"].as_str()
+}
+
+/// The emitted `value` as the single-element list `wait_for_matching_item` expects.
+fn event_values(item: &Value) -> Vec<&str> {
+    event_value(item).into_iter().collect()
+}
+
+/// The `after` cursor string from a delivered event edge.
+fn event_cursor(item: &Value) -> String {
+    item["data"]["events"]["cursor"]
+        .as_str()
+        .expect("edge missing cursor")
+        .to_string()
+}
+
+/// The `{ "pkg": ... }` variables shared by the `type`-filtered subscription queries.
+fn pkg_var(package_id: &ObjectID) -> Option<Value> {
+    Some(json!({ "pkg": package_id.to_string() }))
+}
+
+/// The rich event-subscription query under a given `filter` predicate: live by default, or resuming
+/// from a checkpoint (backfill) when `after_checkpoint` is set. The node is deliberately broad so the
+/// live/backfill parity test catches any field that resolves differently between the two paths.
+fn event_query(filter: &str, after_checkpoint: Option<u64>) -> String {
+    let resume = after_checkpoint
+        .map(|c| format!("afterCheckpoint: {c},"))
+        .unwrap_or_default();
+    format!(
+        r#"subscription($pkg: SuiAddress!) {{
+            events(filter: {{ {resume} {filter} }}) {{
+                cursor
+                node {{
+                    eventBcs
+                    sequenceNumber
+                    sender {{ address }}
+                    contents {{ type {{ repr }} json }}
+                    transaction {{
+                        digest
+                        sender {{ address }}
+                        kind {{ __typename }}
+                        gasInput {{ gasBudget gasPrice }}
+                    }}
+                }}
+            }}
+        }}"#
+    )
+}
+
+/// Take the next `n` event payloads in arrival order.
+async fn take_events(
+    stream: &mut (impl tokio_stream::Stream<Item = Value> + Unpin),
+    n: usize,
+) -> Vec<Value> {
+    stream.take(n).collect().await
+}
+
+/// Collect the `node` of each expected event (keyed by emitted `value`) from a subscription,
+/// returned in `expected` order so the comparison is stable.
+async fn collect_event_nodes(
+    stream: &mut (impl tokio_stream::Stream<Item = Value> + Unpin),
+    expected: &[String],
+) -> Vec<Value> {
+    let mut by_value: BTreeMap<String, Value> = BTreeMap::new();
+    while by_value.len() < expected.len() {
+        let item = stream
+            .next()
+            .await
+            .expect("stream ended before all expected events");
+        let node = item["data"]["events"]["node"].clone();
+        let value = node["contents"]["json"]["value"]
+            .as_str()
+            .expect("event missing value")
+            .to_string();
+        if expected.contains(&value) {
+            by_value.insert(value, node);
+        }
+    }
+    expected.iter().map(|v| by_value[v].clone()).collect()
 }
 
 #[tokio::test]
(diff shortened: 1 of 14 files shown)


crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -0,0 +1,401 @@
...
+        wait_for_pipelines_catching_up_at(end_cursor.position.checkpoint(), watermarks_rx).await?;
+
+        // A forward page keeps `items` in ascending order, one delivered edge each.
+        let mut items = Vec::with_capacity(result.items.len() + 1);
+        for item in &result.items {
+            let token =
+                CursorToken::decode(&item.cursor).context("Failed to decode scan item cursor")?;
+            items.push(Scanned {
+                checkpoint: token.position.checkpoint(),
+                edge: Some(S::build_edge(scope, item)?),
```
wlmyng: in this generic driver, when you call `build_edge`, you already have the cursor component of the edge. And, `cursor_from_end(token: CursorToken) -> Self::Cursor`, so it should be correct to just use the token in hand

I think the trait just needs a conversion from scanned item into graphql node

fn node(scope: &Scope, payload: &Self::ScanItem) -> Result<Self::Item, RpcError>;

then you can use the `transaction_from_stream_item` or `event_from_stream_item` 

```
  for item in &result.items {
      let token = CursorToken::decode(&item.cursor).context("Failed to decode scan item cursor")?;
      let checkpoint = token.position.checkpoint();
      let cursor = S::Cursor::try_from(token)
          .context("Unexpected position in scan item cursor")?
          .encode_cursor();
      items.push(Scanned {
          checkpoint,
          edge: Some(Edge::new(cursor, S::node(scope, &item.payload)?)),
      });
  }
```

tpham-mysten: Yeah it's cleaner. I have applied the suggestion!

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -0,0 +1,406 @@
...
+//! not-yet-indexed store and resolve to null permanently, since the stream never revisits.
+//!
+//! # Cursors and anomalies
+//!
+//! Both phases mint the same opaque cursor, so a client resumes from any delivered one. A gap between
+//! the phases, or a subscriber that falls behind the broadcast buffer, disconnects with
+//! `reconnect_error`, and the client reconnects and resumes from its last cursor.
+//!
+//! What varies per feed (which fields match, how cursors wrap, which scan RPC to call) is supplied
+//! by the [`Subscribable`] implementation for that feed; everything here is shared.
```
wlmyng: Maybe we can reduce the text a bit, paraphrasing:

```
Scan-then-live subscription driver operates in two phases:
1. backfill scans indexed data to catch up, uses same pagination path 
2. read from checkpoint buffer from checkpoint broadcast

Within `handoff_threshold`, driver subscribes to the checkpoint broadcast, so that checkpoint reading is handed off at handoff + 1

```

maybe include coverage marker piece

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -0,0 +1,406 @@
...
+    package_store: Arc<StreamingPackageStore>,
+    resolver_limits: sui_package_resolver::Limits,
+    mut watermarks_rx: watch::Receiver<Arc<Watermarks>>,
+    filter: S::Filter,
+    mut after: Option<S::Cursor>,
+    checkpoint_lo: u64,
+    scan_page_size: usize,
+) -> impl Stream<Item = Result<Scanned<S>, RpcError>> {
+    stream! {
+        // Finalized, indexed data: fields resolve lazily through the index. cvat is None (uniform
```
wlmyng: cvat is checkpoint viewed at? seems we only have this short-hand once

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -0,0 +1,406 @@
...
+        let scope = Scope::for_backfilled_transactions(package_store, resolver_limits);
+        let limits = PageLimits {
+            default: scan_page_size as u32,
+            max: scan_page_size as u32,
+        };
+
+        // The scan applies both resume channels together: `after` (the precise `options.after`
+        // position) and the checkpoint window `[checkpoint_lo, ..]`, so the effective start is
+        // whichever is later. Open above: the caller stops the scan once the handoff is covered.
+        let cp_bounds = checkpoint_lo..=u64::MAX;
```
wlmyng: im actually not sure how the doc comment relates to L225? I think it's describing mechanisms handled in `scan_page` or elsewhere? suppose we could explain why rhs is u64::MAX (don't know exactly but at some point hand over to stream which is indefinite)

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -0,0 +1,406 @@
...
+/// Retry policy for the backfill scan: jittered exponential backoff, giving up after 60s (past which
+/// a failure is unlikely to be a transient blip).
+fn scan_backoff() -> ExponentialBackoff {
+    ExponentialBackoff {
+        initial_interval: Duration::from_millis(100),
+        max_interval: Duration::from_secs(5),
+        max_elapsed_time: Some(Duration::from_secs(60)),
+        ..Default::default()
+    }
+}
```
wlmyng, line 329: is it possible to establish some relationship between these values and graphql/ subscription-specific configs?

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -0,0 +1,406 @@
...
+        filter: &Self::Filter,
+    ) -> Result<Vec<Edge<String, Self::Item, EmptyFields>>, RpcError>;
+}
+
+/// One backfill item: a matching `edge`, or a coverage marker (`edge: None`) whose `checkpoint` is
+/// the fully-scanned frontier.
+struct Scanned<S: Subscribable> {
+    checkpoint: u64,
+    edge: Option<Edge<String, S::Item, EmptyFields>>,
+}
```
wlmyng: what about an enum `Match { checkpoint, edge } / Covered` so it's self-documenting

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -0,0 +1,406 @@
...
+    fn build_node(scope: &Scope, payload: &Self::ScanItem) -> Result<Self::Item, RpcError>;
+
+    /// The filter-matching edges in one live checkpoint, in delivery order.
+    fn matching_edges(
+        checkpoint: &Arc<ProcessedCheckpoint>,
+        package_store: &Arc<StreamingPackageStore>,
+        resolver_limits: &sui_package_resolver::Limits,
+        filter: &Self::Filter,
+    ) -> Result<Vec<Edge<String, Self::Item, EmptyFields>>, RpcError>;
+}
```
wlmyng, line 92: I suppose you could argue that Subscribable is really a Backfill half + Live half (backfill = scan + build_node, live = matching_edges)

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
```diff
@@ -1,429 +1,76 @@
...
-                    warn!(error = ?e, "scan_grpc failed, retrying");
-                    tokio::time::sleep(delay).await;
-                }
-                None => return Err(e),
-            },
+        let mut edges = Vec::new();
+        for tx in &checkpoint.transactions {
+            if !filter.matches(&tx.contents) {
+                continue;
+            }
```
wlmyng, line 67: i recall something about using bloom filters / bitmaps in memory instead of the check we do today, is that still a thing eventually?

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
```diff
@@ -157,110 +141,81 @@ impl Subscription {
...
+        // for checkpoints that arrive during the handoff so the receiver does not lag.
+        let handoff_threshold = config.broadcast_buffer as u64 / 2;
+
+        if filter.at_checkpoint.is_some() || filter.before_checkpoint.is_some() {
+            return Err(bad_user_input(Error::CheckpointBoundsUnsupported));
+        }
+
+        let after_checkpoint = filter.after_checkpoint.map(u64::from);
+
+        Ok(subscribe::<Event>(
```
wlmyng: why not just pass config into `subscribe`, since at least rn we derive the same config values?

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -321,18 +338,23 @@ impl Event {
...
-        request.start_checkpoint = Some(*cp_bounds.start());
-        // `cp_bounds` end is inclusive; the request bound is exclusive.
-        request.end_checkpoint = Some(cp_bounds.end().saturating_add(1));
+        request.start_checkpoint = match cp_bounds.start_bound() {
+            Bound::Included(&s) => Some(s),
+            Bound::Excluded(&s) => Some(s.saturating_add(1)),
+            Bound::Unbounded => None,
+        };
+        request.end_checkpoint = match cp_bounds.end_bound() {
+            Bound::Included(&e) => Some(e.saturating_add(1)),
```
wlmyng, line 347: Note to self to handle the duplicated checkpoint and query options setting

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
```diff
@@ -359,3 +444,342 @@ async fn test_event_subscription_recovers_from_upstream_disconnect() {
...
+    wait_for_matching_item(&mut stream, &["1000".to_string()], event_values).await;
+
+    // Phase 2: an event emitted after subscribing arrives via the live path.
+    emit_event_harness::emit_with_value(&mut cluster.validator, package_id, 2000).await;
+    wait_for_matching_item(&mut stream, &["2000".to_string()], event_values).await;
+}
+
+#[tokio::test]
+async fn test_event_subscription_empty_backfill_hands_off_to_live() {
+    let mut cluster = SubscriptionTestCluster::new_with_ledger_history().await;
```
wlmyng, line 506: rather than the blanket `BACKFILL_SETTLE` wait, is there a way we can poll graphql's streaming watermark 
which i think we can achieve by just subscribing to checkpoints
```
  let mut probe = cluster
      .subscribe("subscription { checkpoints { node { sequenceNumber transactions {
  nodes { digest } } } } }")
      .await;

  let digest = emit_event_harness::emit_with_value(&mut cluster.validator, package_id,
  1000).await;

  // Precondition, checked: graphql's broadcast has *delivered* the checkpoint
  containing
  // the emit, so network_tip() >= E 
  wait_for_matching_item(&mut probe, &[digest], checkpoint_tx_digests).await;
  drop(probe);

  let mut stream = cluster
      .subscribe_with_variables(&event_query("type: $pkg", Some(resume_from)),
  pkg_var(&package_id))
      .await;
```

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
```diff
@@ -359,3 +444,342 @@ async fn test_event_subscription_recovers_from_upstream_disconnect() {
...
+    tokio::time::sleep(Duration::from_secs(3)).await;
+    for v in [300u64, 400] {
+        emit_event_harness::emit_with_value(&mut cluster.validator, package_id, v).await;
+    }
+
+    // Every event arrives exactly once across the seam: a reorder or drop breaks the value order, a
+    // duplicate at the seam collapses the `eventBcs` distinct count.
+    let events = take_events(&mut stream, 4).await;
+    let values: Vec<&str> = events.iter().filter_map(event_value).collect();
+    assert_eq!(
```
wlmyng, line 559: i think this test is similarly probabilistic? Using a probe would force the events to land as you desire

wlmyng: given the shared testing with transactions, maybe a better home is in the scan_then_live?

We can mock a subscription broadcast, and the backfill side through `Subscribable`

The friction points are having to pass a concrete `AlphaLedgerGrpcReader` but this actually gets ignored

and we need to return ProcessedCheckpoint + StreamingPackageStore for live phase, but we can mock edges through `matching_edges`

wlmyng (approved):
impl looks good, some test improvement suggestions

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: Sounds good. I have updated it to an enum

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: Sure. Updated

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: Yeah you're right! I have updated the doc to explain u64::MAX

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
tpham-mysten: Good catch. We can do that.

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
tpham-mysten: Sounds good. I have made a helper called `wait_until_backfillable` and removed that `BACKFILL_SETTLE` wait

crates/sui-indexer-alt-graphql/src/api/subscription/transactions.rs
tpham-mysten: Yeah it's still on the table but lower priority and can be done after the launch

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
tpham-mysten: I have updated these to a deterministic probe (`wait_until_backfillable`), as in the comment above.

I looked at relocating into scan_then_live with mocks, but the handoff is emergent from real timing, so a faithful mock ends up about as complex as the e2e. And we'd be testing the mock rather than the real seam. So I'd prefer keeping it as-is.

We can revisit if it becomes a problem.

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: imo, these are internal fault-tolerance tuning, a different kind of value than `SubscriptionConfig`, which holds operator-facing policy (e.g. rate, concurrency, buffer, etc). Hence, I've kept them as local constants. Can revisit if an client ever needs to tune them.

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: That's a fair way to look at it. That said, I'd prefer keeping them in one trait. There's no live-only or backfill-only feed today, so I don't think the split earns any benefit yet, but it'd be an easy change if one ever comes up.

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: Sure done

tpham-mysten (commented):
Thanks @wlmyng!

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
```diff
@@ -25,6 +32,84 @@ fn event_bcs(item: &Value) -> Option<&str> {
...
+
+/// The `{ "pkg": ... }` variables shared by the `type`-filtered subscription queries.
+fn pkg_var(package_id: &ObjectID) -> Option<Value> {
+    Some(json!({ "pkg": package_id.to_string() }))
+}
+
+/// The rich event-subscription query under a given `filter` predicate: live by default, or resuming
+/// from a checkpoint (backfill) when `after_checkpoint` is set. The node is deliberately broad so the
+/// live/backfill parity test catches any field that resolves differently between the two paths.
+fn event_query(filter: &str, after_checkpoint: Option<u64>) -> String {
```
amnn: It would be nicer to use a variable for `filter` here to avoid the query interpolation. Not the biggest deal because this is test code, but still.

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
```diff
@@ -359,3 +444,347 @@ async fn test_event_subscription_recovers_from_upstream_disconnect() {
...
     assert_eq!(bcs_set.len(), 8, "eventBcs should be distinct across emits");
 }
+
+#[tokio::test]
+async fn test_event_subscription_ordering() {
+    let mut cluster = SubscriptionTestCluster::new().await;
+    let package_id = emit_event_harness::publish(&mut cluster.validator).await;
+
+    let mut stream = cluster
+        .subscribe_with_variables(&event_query("type: $pkg", None), pkg_var(&package_id))
```
amnn: I'm surprised this actually works from a typing perspective -- `type:` expects a string, but `$pkg` is marked as a `SuiAddress`.

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
```diff
@@ -359,3 +444,347 @@ async fn test_event_subscription_recovers_from_upstream_disconnect() {
...
+    let after = first_edge["cursor"]
+        .as_str()
+        .expect("backfill edge missing cursor")
+        .to_string();
+    assert!(["1000", "2000"].contains(&first_value.as_str()));
+    drop(stream);
+
+    // Subscription 2: resume via `after`. The next matching event must be the other one, proving the
+    // event at the cursor was skipped rather than re-delivered.
+    let query = format!(
```
amnn: ditto on using variables instead of string interpolation.

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
```diff
@@ -30,25 +29,17 @@ use crate::error::bad_user_input;
...
-    #[error("At most one of `after` or `afterCheckpoint` can be specified")]
-    MutuallyExclusiveResume,
-
-    #[error("Invalid `after` cursor: {0}")]
-    InvalidCursor(String),
-
-    #[error(
-        "Filtering by checkpoint (`afterCheckpoint`, `atCheckpoint`, `beforeCheckpoint`) is not supported for subscriptions"
-    )]
+    #[error("Filtering by `atCheckpoint` or `beforeCheckpoint` is not supported for subscriptions")]
```
amnn, line 42: Is this something we're going to relax later?

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -0,0 +1,408 @@
...
+
+impl<S: Subscribable> Scanned<S> {
+    /// The checkpoint this output sits at (the match's, or the covered frontier).
+    fn checkpoint(&self) -> u64 {
+        match self {
+            Scanned::Match { checkpoint, .. } | Scanned::Covered { checkpoint } => *checkpoint,
+        }
+    }
+}
```
amnn, line 113: nit: it's generally a good idea to keep constraints on the data type to an absolute minimum (and then just apply the constraints to the methods), which suggests a set-up like this:


```suggestion
/// One backfill scan output: either a matching edge, or a coverage marker whose `checkpoint` is the
/// fully-scanned frontier (advances the handoff even on a match-less page).
enum Scanned<I> {
    Match {
        checkpoint: u64,
        edge: Edge<String, I, EmptyFields>,
    },
    Covered {
        checkpoint: u64,
    },
}

impl<I> Scanned<I> {
    /// The checkpoint this output sits at (the match's, or the covered frontier).
    fn checkpoint(&self) -> u64 {
        match self {
            Scanned::Match { checkpoint, .. } | Scanned::Covered { checkpoint } => *checkpoint,
        }
    }
}
```

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -0,0 +1,408 @@
...
+
+        // No end cursor means nothing past the indexer tip is indexed yet; wait, then re-request. A
+        // range that was scanned but matched nothing still returns a boundary cursor, not `None`.
+        let Some(end_cursor) = result.last_cursor() else {
+            tokio::time::sleep(BACKFILL_POLL_INTERVAL).await;
+            continue;
+        };
+        let end_cursor = CursorToken::decode(end_cursor).context("Failed to decode scan cursor")?;
+
+        // Hold the page until both pipelines have indexed through its end.
```
amnn:
```suggestion
        // Hold the page until all pipelines have indexed through its end.
```

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
amnn: Perhaps could also be simplified to:

```rust
struct Scanned<I> {
  checkpoint: u64,
  edge: Option<Edge<String, I, EmptyFields>>,
}
```

amnn (approved):
Looks great, really like the `Subscribable` abstraction. Thanks @tpham-mysten !

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
tpham-mysten: Yes we will relax it later as discussed before

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: That was my original implementation and I changed to enum to differentiate 2 cases better.

By https://github.com/MystenLabs/sui/pull/27580#discussion_r3786635866

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
tpham-mysten: Sounds good. I have updated the test

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
tpham-mysten: Good catch! It actually works by accident: `impl_string_input!` names `TypeFilter`'s scalar `String`, so `type` is really a `String` in the schema. `$pkg: SuiAddress!` is also string-valued, and async-graphql
doesn't enforce the variable-vs-arg type check, so it slips through and `TypeFilter::from_str` parses it. 


Refactored to pass an `EventFilter!` variable instead, which avoids the mislabeled scalar.

crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/event_subscription.rs
tpham-mysten: SG. done

tpham-mysten (commented):
Thanks @amnn! I have resolved all comments

PR #27581, opened 2026-08-05
[sui-indexer-alt-graphql] Fix queries_in_flight gauge never decrementing on request cancellation

evan-wall-mysten:
## Description

`MetricsFuture` incremented `queries_in_flight` on every request but only decremented it when the request completed normally (the `Future::poll` `Ready` branch). The `PinnedDrop` path, which runs when a request's future is dropped before resolving (client disconnect, upstream proxy/LB timeout), logged the cancellation and bumped `queries_cancelled` but never decremented `queries_in_flight`, permanently inflating the gauge. In production this surfaced as a testnet `sui-indexer-alt-graphql` pod whose `queries_in_flight` got stuck at 24,034 after a burst of requests timed out and were cancelled, with the gauge never recovering.

## Test plan

Added `queries_in_flight_decrements_on_cancellation`, which drops a `MetricsFuture` before it resolves and asserts the gauge returns to 0 (confirmed it fails without the fix).

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


amnn (approved):
Nice find!

PR #8788, opened 2023-03-02
[ledger integration] add ledgerjs-hw-app-sui library as an SDK

williamrobertson13:
## Description 

This PR adds the `ledgerjs-hw-app-sui` library from Obsidian as an SDK under our repository since we'll be maintaining and owning it going forward. I changed around and added some project configurations (e.g. ESLint, Prettier, Vitest, etc.), but all of the code is pretty much strictly copy and pasted over. I confirmed with the Obsidian team that it's okay to move this code over and apply our software licenses.

There are some improvements and clean-up to be done in the SDK itself, but I figure this is a good starting point since it's at least in a working state. As follow-ups, we'll do some KTLO work to get the library in better shape and add more comprehensive test coverage (i.e., unit tests and E2E tests that interact with [Speculos](https://github.com/LedgerHQ/speculos)).

## Test Plan 
- All `npm` scripts run successfully

---
If your changes are not user-facing and not a breaking change, you can skip the following section. Otherwise, please indicate what changed, and then add to the Release Notes section as highlighted during the release process.

### Type of Change (Check all that apply)

- [ ] user-visible impact
- [ ] breaking change for a client SDKs
- [ ] breaking change for FNs (FN binary must upgrade)
- [ ] breaking change for validators or node operators (must upgrade binaries)
- [ ] breaking change for on-chain data layout
- [ ] necessitate either a data wipe or data migration

### Release notes
N/A

PR #27585, opened 2026-08-05
sui-indexer-alt-graphql: Switch to jemalloc global allocator

evan-wall-mysten:
## Description

Switches `sui-indexer-alt-graphql` to `tikv-jemallocator` as the global allocator, to avoid the
platform default allocator (glibc malloc) retaining freed heap pages instead of returning them to
the OS after a burst of concurrent allocations.

## Test plan

Load-tested `testnet-list-apis` (`sui-indexer-alt-graphql-list-apis`, `--enable-list-apis`) with
real GraphQL traffic (k6, 11,000-record Snowflake-sourced corpus of list-style
`transactions(filter: {...})` queries, flat 100 rps for 5 minutes) on `main` and on this PR's
commit, and compared [pod memory usage during/after each run]:

Blue: glibc (sharp drop is pod being destroyed)
Green: jemalloc
image

- **glibc (`main`)**: pod RSS climbs to ~18 GiB during the load test, then stays pinned at
  ~18 GiB for several minutes after the test ends - it only comes down when the pod is manually
  destroyed.
- **jemalloc (this PR)**: pod RSS peaks much lower, ~4.6 GiB, and is already declining on its own
  by the end of the observation window (no pod restart needed) - down to ~3.2 GiB.

This reproduces the production RSS-retention pattern live under real `list-apis` traffic, and
shows this PR's allocator switch resolving it.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/Cargo.toml
+++ b/crates/sui-indexer-alt-graphql/Cargo.toml
@@ -47,6 +47,7 @@ serde_json.workspace = true
 shared-crypto.workspace = true
 telemetry-subscribers.workspace = true
 thiserror.workspace = true
+tikv-jemallocator = { workspace = true, optional = true, features = ["profiling", "disable_initial_exec_tls"] }
 timeout-tracing.workspace = true
 tokio.workspace = true
 tokio-stream.workspace = true
@@ -89,4 +90,6 @@ tokio = { workspace = true, features = ["test-util"] }
 tracing-subscriber.workspace = true
 
 [features]
+default = ["jemalloc"]
+jemalloc = ["tikv-jemallocator"]
 staging = []
--- a/crates/sui-indexer-alt-graphql/Cargo.toml
+++ b/crates/sui-indexer-alt-graphql/Cargo.toml
@@ -77,6 +77,9 @@ sui-sdk-types.workspace = true
 sui-sql-macro.workspace = true
 sui-types.workspace = true
 
+[target.'cfg(not(target_env = "msvc"))'.dependencies]
+tikv-jemallocator = { workspace = true, optional = true }
+
 [dev-dependencies]
 bytes.workspace = true
 insta.workspace = true
@@ -88,4 +91,6 @@ tokio = { workspace = true, features = ["test-util"] }
 tracing-subscriber.workspace = true
 
 [features]
+default = ["jemalloc"]
+jemalloc = ["tikv-jemallocator"]
 staging = []
--- a/crates/sui-indexer-alt-graphql/Cargo.toml
+++ b/crates/sui-indexer-alt-graphql/Cargo.toml
@@ -47,6 +47,7 @@ serde_json.workspace = true
 shared-crypto.workspace = true
 telemetry-subscribers.workspace = true
 thiserror.workspace = true
+tikv-jemalloc-ctl = { workspace = true, optional = true }
 tikv-jemallocator = { workspace = true, optional = true, features = ["profiling", "disable_initial_exec_tls"] }
 timeout-tracing.workspace = true
 tokio.workspace = true
@@ -91,5 +92,5 @@ tracing-subscriber.workspace = true
 
 [features]
 default = ["jemalloc"]
-jemalloc = ["tikv-jemallocator"]
+jemalloc = ["tikv-jemallocator", "tikv-jemalloc-ctl"]
 staging = []
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15563,6 +15563,7 @@ dependencies = [
  "sui-types",
  "telemetry-subscribers",
  "thiserror 1.0.69",
+ "tikv-jemalloc-ctl",
  "tikv-jemallocator",
  "timeout-tracing",
  "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15563,6 +15563,7 @@ dependencies = [
  "sui-types",
  "telemetry-subscribers",
  "thiserror 1.0.69",
+ "tikv-jemallocator",
  "timeout-tracing",
  "tokio",
  "tokio-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15414,6 +15414,7 @@ dependencies = [
  "sui-types",
  "telemetry-subscribers",
  "thiserror 1.0.69",
+ "tikv-jemallocator",
  "tokio",
  "tokio-stream",
  "toml 0.7.4",
--- a/crates/sui-indexer-alt-graphql/src/main.rs
+++ b/crates/sui-indexer-alt-graphql/src/main.rs
@@ -28,6 +28,10 @@ static VERSION: &str = const_str::concat!(
     GIT_REVISION
 );
 
+#[cfg(all(not(target_env = "msvc"), feature = "jemalloc"))]
+#[global_allocator]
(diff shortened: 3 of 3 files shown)


evan-wall-mysten:
Blue: glibc (sharp drop is pod being destroyed)
Green: jemalloc
image

crates/sui-indexer-alt-graphql/Cargo.toml
```diff
@@ -47,6 +47,7 @@ serde_json.workspace = true
 shared-crypto.workspace = true
 telemetry-subscribers.workspace = true
 thiserror.workspace = true
+tikv-jemallocator = { workspace = true, optional = true, features = ["profiling", "disable_initial_exec_tls"] }
```
amnn: Should we adopt exactly `sui-node`'s approach here, which is to disable `jemalloc` on `msvc` builds?

https://github.com/MystenLabs/sui/blob/944e5f24dfe7cb61997bbbe35295a751b66d1914/crates/sui-node/Cargo.toml#L67-L68 

Also, this is a departure from what `sui-node` does, but the explicit features listed here are redundant, because these are already the features on the workspace dep:

```suggestion
tikv-jemallocator = { workspace = true, optional = true }
```

(Incidentally, it was surprising to me that we had profiling enabled all the time for ).

amnn (approved):
Thanks @evan-wall-mysten, this is a clearer repro for sure, and I'm happy to land this change, but it would still be good to understand:

- What the new ceiling is with jemalloc.
- Where the memory is going (4GiB is still quite a lot).
- How the system behaves over a longer time horizon (in particular, where does `jemalloc` eventually settle to in terms of RSS?)

crates/sui-indexer-alt-graphql/Cargo.toml
evan-wall-mysten: Good catch, done — moved `tikv-jemallocator` under `[target.'cfg(not(target_env = "msvc"))'.dependencies]` to match `sui-node`, and dropped the redundant `features` list since the workspace dep already declares them.

evan-wall-mysten:
## Profiling follow-up: dumps confirmed not working, root cause narrowed but not resolved

Added (this PR, latest 2 commits) a startup log of `opt.prof`/`opt.prof_active`/`opt.prof_gdump`
plus a periodic `prof.dump` background task, since `_RJEM_MALLOC_CONF=prof:true` alone produces
no heap dumps with no app-level dump trigger.

**Deployed and tested against testnet-list-apis (sui-operations PR #8788).** Result: `opt.prof`
reads `false` at runtime despite `_RJEM_MALLOC_CONF=prof:true,prof_gdump:true` being confirmed
byte-correct in `/proc/1/environ`. Ruled out, with hard evidence for each:

- Wrong env var name / prefix convention — confirmed `_RJEM_MALLOC_CONF` is correct for
  `tikv-jemalloc-sys` 0.5.4's default (prefixed) build.
- Malformed conf string — no `<jemalloc>:` parse-error lines anywhere in the pod's logs, and
  the env var's raw bytes are clean (no stray quoting).
- Cargo feature resolution — `cargo tree -e features` confirms `tikv-jemalloc-sys`'s `profiling`
  feature *is* correctly resolved for this binary target; no duplicate/conflicting jemalloc-sys
  versions in the dependency graph.
- `release-unwind` profile / LD_PRELOAD / read-only filesystem / wrong CWD / PID1 wrapper /
  `env::set_var` race in the startup path — all checked and ruled out.
- Stale warm build cache reusing a pre-profiling `tikv-jemalloc-sys` artifact — wired a
  `cache_family` override (sui-operations PR #8788) and forced a genuinely cold-cache rebuild on
  a fresh commit. **Still `opt.prof=false`.**

**Strongest remaining signal:** writing to the *live* `prof.active` mallctl (as opposed to the
always-present `opt.*` read-only reflections) returns jemalloc's "name/mib specifies an
unknown/invalid value" — i.e. ENOENT. This means the `prof.*` ctl subtree was never registered
at all, which only happens when `config_prof` (`--enable-prof`) was genuinely not compiled in —
despite Cargo correctly requesting it. A local build with an identical toolchain
(`rust:1.96.1-bullseye`, same `Cargo.toml`/`Cargo.lock` feature set) got as far as
`./configure --enable-prof` succeeding (`prof: 1` in its summary) before an unrelated QEMU
emulation crash stopped me from producing a runnable binary to confirm end-to-end — so I
couldn't rule out one remaining possibility: something in the *actual* MP3 build execution
(not just Cargo's feature resolution) differs from a plain `cargo build` with these flags.

Next step, if anyone wants to pick this up: get the raw MP3 build log for an
`sui-indexer-alt-graphql` build on this branch (the `build-status` skill / mp3 MCP tooling
wasn't available to me this session) and grep it for jemalloc's `./configure` summary — that
would show definitively what actually got compiled, vs. what Cargo *thinks* it requested.

evan-wall-mysten:
## Resolved: profiling now works end-to-end

Root cause found. `tikv-jemalloc-sys`'s `unprefixed_malloc_on_supported_platforms` feature
resolves **active** during this crate's actual `cargo build` for this binary — confirmed via a
verbose, single-threaded, single-crate `cargo build -p tikv-jemalloc-sys -vv` (so build-script
output can't interleave with anything else) run directly against a real mp3 build pod. This is
true even though:

- neither `sui-indexer-alt-graphql`'s nor the workspace root's `Cargo.toml` declares that
  feature anywhere, and
- `cargo tree -e features -i tikv-jemalloc-sys` does **not** predict it — so don't trust `cargo
  tree` alone for this crate/scenario; verified the discrepancy both locally and inside the real
  build container.

With that feature active, `tikv-jemalloc-sys`'s build.rs drops `--with-jemalloc-prefix=_rjem_`
from jemalloc's `./configure` invocation, so the built jemalloc reads the **unprefixed**
`MALLOC_CONF` env var, not `_RJEM_MALLOC_CONF` (which is what the `sui-node` "Memory profiling"
docs and the crate's normal prefixed-build convention would lead you to expect). Verified
empirically against the exact deployed image: `opt.prof` reads `true` with `MALLOC_CONF`, `false`
with `_RJEM_MALLOC_CONF`.

Fixed on the sui-operations side (deploy-time env var name only — no change needed here) in
[sui-operations#8788](https://github.com/MystenLabs/sui-operations/pull/8788). Redeployed
`testnet-list-apis` and confirmed real `jeprof.<pid>.<seq>.u<seq>.heap` dump files are now
produced continuously as traffic flows.

No code change needed in this PR beyond what's already here (the startup diagnostic log +
periodic `prof.dump` task added earlier in this thread) — those work correctly now that the
right env var is set at the deploy layer. Leaving the exact root cause here for anyone who hits
`opt.prof=false` unexpectedly on a `tikv-jemallocator`-based binary in this repo's build
pipeline in the future.

evan-wall-mysten:
> Thanks @evan-wall-mysten, this is a clearer repro for sure, and I'm happy to land this change, but it would still be good to understand:
> 
> * What the new ceiling is with jemalloc.
> * Where the memory is going (4GiB is still quite a lot).
> * How the system behaves over a longer time horizon (in particular, where does `jemalloc` eventually settle to in terms of RSS?)

I'll enable profiling in a follow up PR with this info.

PR #27586, opened 2026-08-05
ci: bind simulator protocol_config_override via env before shell

SashaMIT:
## Summary
- Bind `github.event.inputs.protocol_config_override` (plus `event_name` / `schedule`) through `env:` before the protocol-config override shell step in `simulator-nightly.yml`.
- Same class as GitHub’s documented script-injection guidance for interpolating workflow inputs into `run:` scripts.

## Test plan
- [ ] Manual dispatch with override still sets `SUI_PROTOCOL_CONFIG_CHAIN_OVERRIDE`
- [ ] Scheduled mainnet window still forces mainnet override

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/create-release-announce.yml
+++ b/.github/workflows/create-release-announce.yml
@@ -82,8 +82,8 @@ jobs:
         if: ${{ inputs.network == 'devnet' }}
         shell: bash
         env:
-          SUI_COMMIT: ${SUI_COMMIT}
-          NETWORK: ${NETWORK}
+          SUI_COMMIT: ${{ inputs.sui_commit }}
+          NETWORK: ${{ inputs.network }}
         run: |
           echo "${{ secrets.SUI_CREATE_RELEASE }}" | gh auth login --with-token
           echo "{\"sha\": \"${SUI_COMMIT}\", \"force\": true}" | gh api -X PATCH "/repos/MystenLabs/sui/git/refs/heads/framework/${NETWORK}" \
--- a/.github/workflows/create-release-announce.yml
+++ b/.github/workflows/create-release-announce.yml
@@ -75,7 +75,8 @@ jobs:
           echo "sui_release_version=${sui_release_version}" >> $GITHUB_ENV
 
           echo "${{ secrets.SUI_CREATE_RELEASE }}" | gh auth login --with-token
-          echo "{\"sha\": \"${SUI_COMMIT}\", \"force\": true}" | gh api -X PATCH "/repos/MystenLabs/sui/git/refs/heads/${NETWORK}" \
+          # Build the JSON payload with jq so input values are properly escaped.
+          jq -n --arg sha "$SUI_COMMIT" '{sha: $sha, force: true}' | gh api -X PATCH "/repos/MystenLabs/sui/git/refs/heads/${NETWORK}" \
             --header "Accept: application/vnd.github+json" --input -
 
       - name: Update framework/${{ inputs.network }} branch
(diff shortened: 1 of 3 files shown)


SashaMIT:
Follow-up pushed: also bound `create-release-announce.yml` (`sui_commit` / `network`) and `setup-sccache` `key-prefix` through `env:` before shell — same class as the simulator-nightly change.

.github/workflows/simulator-nightly.yml
```diff
@@ -130,16 +130,20 @@ jobs:
...
         run: |
           message=""
-          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
-            if [[ -n "${{ github.event.inputs.protocol_config_override }}" ]]; then
-              echo "Testing with protocol config override: ${{ github.event.inputs.protocol_config_override }}"
-              echo "SUI_PROTOCOL_CONFIG_CHAIN_OVERRIDE=${{ github.event.inputs.protocol_config_override }}" >> $GITHUB_ENV
+          if [[ "$EVENT_NAME" == "workflow_dispatch" ]]; then
+            if [[ -n "$PROTOCOL_CONFIG_OVERRIDE" ]]; then
+              echo "Testing with protocol config override: $PROTOCOL_CONFIG_OVERRIDE"
+              echo "SUI_PROTOCOL_CONFIG_CHAIN_OVERRIDE=${PROTOCOL_CONFIG_OVERRIDE}" >> $GITHUB_ENV
```
ebmifa: This removes direct shell interpolation, but a multiline override can still add extra entries to `$GITHUB_ENV`: every line after the first is parsed separately. Since this is security hardening, could we either restrict the override to its expected character/value set before writing it, or use the documented multiline delimiter form? It would also be useful for the manual-dispatch test to verify that this value reaches the remote simulator process—the later `tsh ssh ... nohup` command does not appear to forward `SUI_PROTOCOL_CONFIG_CHAIN_OVERRIDE` explicitly.

.github/workflows/create-release-announce.yml
```diff
@@ -67,33 +67,41 @@ jobs:
...
+        env:
+          SUI_COMMIT: ${{ inputs.sui_commit }}
+          NETWORK: ${{ inputs.network }}
         run: |
           export sui_release_version=$(cat Cargo.toml | grep '^version =' | tr -d '\"' | awk '{ print $3 }')
           echo "sui_release_version=${sui_release_version}" >> $GITHUB_ENV
 
           echo "${{ secrets.SUI_CREATE_RELEASE }}" | gh auth login --with-token
-          echo "{\"sha\": \"${{ inputs.sui_commit }}\", \"force\": true}" | gh api -X PATCH /repos/MystenLabs/sui/git/refs/heads/${{ inputs.network }} \
+          echo "{\"sha\": \"${SUI_COMMIT}\", \"force\": true}" | gh api -X PATCH "/repos/MystenLabs/sui/git/refs/heads/${NETWORK}" \
```
ebmifa: Passing `sui_commit` through `env:` prevents shell-source injection, but interpolating it into hand-built JSON still leaves JSON quoting/structure dependent on the input. Could we construct the payload with `jq -n --arg sha "$SUI_COMMIT" '{sha: $sha, force: true}'` (and do the same for the framework update and tag creation below), or validate that `SUI_COMMIT` is an expected commit SHA first?

ebmifa (commented):
The `env:` indirection is the right way to prevent GitHub expression values from being parsed as shell source. I found two remaining input-boundary issues worth tightening, and the simulator path would benefit from an end-to-end dispatch test.

PR #27589, opened 2026-08-05
[gql] Don't read the index from an execution-context scope

tpham-mysten:
## Description

#27140 relaxed the content-hydration cutoff so subscription backfills could hydrate from the store, but it also let an execution scope (`executeTransactionBlock` / `simulateTransaction`) read the store when resolving digest-only descendant nodes (e.g. an indexed `effects.dependencies` node), leaking indexed data into a context that should reflect only the executed result.

Gate the store read on `Scope::is_executed()`: execution scopes stay store-isolated, while indexed queries and streaming subscriptions hydrate as before.
   
## Test plan

e2e + unit tests

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
@@ -534,6 +534,11 @@ impl TransactionContents {
             });
         }
 
+        // Execution context is not backed by the index yet.
+        if self.scope.is_executed() {
+            return Ok(self.clone());
+        }
+
         let kv_loader: &KvLoader = ctx.data()?;
         let Some(transaction) = kv_loader
             .load_one_transaction(digest)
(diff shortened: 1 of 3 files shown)


amnn (commented):
Thanks for the fix @tpham-mysten -- I've also added a design topic for our meetings to discuss whether there is a better way to encapsulate these kinds of checks inside `Scope`.

amnn (approved):
Forgot to accept! Also, could you make sure this is picked to wherever the earlier change landed, if necessary?

tpham-mysten:
Thanks @amnn! We don't need a cherrypick here as #27140 was merged on July 31 after the branch cut.

PR #27592, opened 2026-08-06
Added support for tracing (local) dry-runs

awelc:
## Description 

This PR adds tracing support for `sui client ptb --dry-run` via an additional `--trace` flag. It produces artifacts similar to those created by the replay tool, most importantly the artifacts needed for trace debugging.

The traced dry-run executes locally, similarly to how the replay tool executes historical transactions. Unlike a regular dry-run on a full node, which executes against the latest chain state and may observe state mid-checkpoint, it executes against the latest completed checkpoint. This ensures that the epoch, protocol version, gas price, timestamp, packages, and object state form a consistent snapshot for local execution.

We chose the replay tool as a foundation because it already supports traced transaction execution, produces artifacts needed for trace debugging, and provides convenient data-store abstractions (but we had to add dry-run-specific input preparation and a checkpoint-bounded runtime store to support correct dry-run validation and checkpoint consistency)

## Test plan 

New tests have been added. All tests must pass. Also tested manually end-to-end


crates/sui-replay-2/src/dry_run/input_loader.rs
```diff
@@ -0,0 +1,682 @@
...
+        digest,
+        checked_inputs,
+        gas_status,
+        checkpoint_objects,
+        mock_gas_object: prepared.mock_gas_object,
+    })
+}
+
+/// Recheck address-owned inputs against the latest live state at the selected checkpoint.
+fn validate_live_owned_inputs<S>(
```
tzakian, line 161: Not sure I follow why this is needed in addition to the `check_transaction_input` checks? IMO it feels like this should share a lot/all with that and if not that's a good signal to re-evaluate probably

crates/sui-data-store/src/lib.rs
```diff
@@ -69,14 +75,80 @@ pub trait TransactionStore {
...
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct EpochData {
     pub epoch_id: u64,
     pub protocol_version: u64,
     pub rgp: u64,
     pub start_timestamp: u64,
 }
 
+/// Chain and epoch metadata for executing against state at a finalized checkpoint.
```
tzakian, line 86: This feels like the wrong place for this to live?

I get why you want it here, but the fact that it needs to exist here makes me a bit iffy about this...

tzakian (commented):
I think the main questions I have on this are the following: 
1. How important is it that we anchor to a consistent checkpoint view? My gut is that it's not super important, and my (possibly incorrect) read, is that if we relax things in this manner it makes the implementation much simpler and able to reuse a lot of the existing functionality.
2. If we do decide that we want to checkpoint anchor everything, I'd like to see if there is some way of unifying some things, as it feels like there's a lot of parallel and "almost the same but slightly dissimilar" place throughout -- so just enough to cause a lot of almost-duplicate-but-not-quite code which I'd prefer if we could avoid.

awelc:
Superseded by https://github.com/MystenLabs/sui/pull/27749

PR #27593, opened 2026-08-06
sui-indexer-alt-graphql: fix a request-timeout memory leak

evan-wall-mysten:
## Description

Investigated a slow memory-growth issue on `sui-indexer-alt-graphql`'s testnet-list-apis stack using throwaway diagnostic builds (a LeakSanitizer build on a nightly toolchain, and a heaptrack-based companion for the production allocator) — neither checked in, since they're trivial to recreate and their exact toolchain/version pins would likely be stale by the time they're needed again. That traced the leak to the third-party `timeout-tracing` crate: its custom `Waker::wake()` never freed its allocation on the consuming path.

Replaced the dependency with a small, local, corrected implementation (`sui_futures::timeout_trace`) rather than depending on a fix to an unmaintained, single-purpose crate. Following review, reworked it further: captured traces now flow through a bounded channel instead of being retained via `Arc`s that a long-pending branch would otherwise hold onto for its whole lifetime, and wake-forwarding was dropped entirely (justified in the module's doc comment — the one-shot final-poll usage here never needs it).

## Test plan

- `sui_futures::timeout_trace`: unit tests cover the success path, zero/one/multiple concurrently-pending traces, the `wake`/`wake_by_ref`/`drop`/`clone` `RawWakerVTable` contract, and specifically that a waker clone which is touched but never retained doesn't produce a false-positive trace (confirmed load-bearing: fails without the liveness check, passes with it).
- `sui-indexer-alt-graphql::extensions::timeout`: existing snapshot tests pass unchanged (confirms `Display` output is byte-identical through the crate boundary), plus a new test covering multiple concurrently-pending GraphQL fields timing out at once.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-futures/src/timeout_trace.rs
+++ b/crates/sui-futures/src/timeout_trace.rs
@@ -1,0 +1,339 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::error::Error;
+use std::fmt::Display;
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::sync::Mutex;
+use std::sync::Weak;
+use std::task::Context;
+use std::task::Poll;
+use std::task::RawWaker;
+use std::task::RawWakerVTable;
+use std::task::Waker;
+use std::time::Duration;
+
+use pin_project::pin_project;
+use tokio::time::Sleep;
+use tracing_error::SpanTrace;
+
+/// Wraps a future with a timeout that also captures [`SpanTrace`]s of whatever was still pending
+/// when it fired -- see [`timeout`].
+#[pin_project]
+pub struct TracedTimeout<Fut> {
+    #[pin]
+    deadline: Sleep,
+    #[pin]
+    inner: Fut,
+}
+
+/// Returned when `inner` hasn't finished within `duration`, carrying a [`SpanTrace`] for every
+/// await point of `inner` that was still pending at that moment. There can be more than one --
+/// e.g. when a caller resolves several sibling branches concurrently.
+#[derive(Debug)]
+pub struct TimeoutElapsed {
+    pub active_traces: Vec<Arc<SpanTrace>>,
+}
+
+/// Shared state for a single timeout's final poll: the traces captured so far, and the real
+/// waker to forward wake-ups to. Each clone's trace is only present in `active_traces` for as
+/// long as that clone is alive -- once it's dropped, its `Weak` naturally stops upgrading, so
+/// nothing needs to be nulled out by hand.
+struct TracingWakerInner {
+    active_traces: Mutex<Vec<Weak<SpanTrace>>>,
+    inner_waker: Waker,
+}
+
+/// A `Waker` that captures a [`SpanTrace`] every time it's cloned -- each clone marks a distinct
+/// still-pending await point -- and forwards actual wake-ups to `inner_waker`.
+struct TracingWaker {
+    inner: Arc<TracingWakerInner>,
+    /// This clone's own trace, kept alive for as long as this waker is (never read directly --
+    /// `active_traces` observes it through the `Weak` pushed alongside it). `None` for the
+    /// original waker, which isn't itself an await point.
+    _trace: Option<Arc<SpanTrace>>,
+}
+
+impl TracingWakerInner {
+    fn new(inner_waker: Waker) -> Arc<Self> {
+        Arc::new(Self {
+            active_traces: Mutex::new(Vec::with_capacity(4)),
+            inner_waker,
+        })
+    }
+
+    fn take_traces(&self) -> Vec<Arc<SpanTrace>> {
+        let traces = std::mem::take(&mut *self.active_traces.lock().unwrap());
+        traces
+            .into_iter()
+            .filter_map(|trace| trace.upgrade())
+            .collect()
+    }
+}
+
+impl TracingWaker {
+    const VTABLE: RawWakerVTable = RawWakerVTable::new(
+        Self::raw_clone,
+        Self::raw_wake,
+        Self::raw_wake_by_ref,
+        Self::raw_drop,
+    );
+
+    fn new_std_waker(inner: Arc<TracingWakerInner>) -> Waker {
+        let data = Box::into_raw(Box::new(Self {
+            inner,
+            _trace: None,
+        }));
+        // SAFETY: `data` was just obtained from `Box::into_raw`, so it's a unique, valid pointer
+        // to a `Box<Self>` -- the precondition every vtable function below relies on.
+        unsafe { Waker::new(data as *const (), &Self::VTABLE) }
+    }
+
+    fn clone_capturing_trace(&self) -> Box<Self> {
+        let trace = Arc::new(SpanTrace::capture());
+        self.inner
+            .active_traces
+            .lock()
+            .unwrap()
+            .push(Arc::downgrade(&trace));
+        Box::new(Self {
+            inner: self.inner.clone(),
+            _trace: Some(trace),
+        })
+    }
+
+    unsafe fn raw_clone(data: *const ()) -> RawWaker {
+        // SAFETY: see `new_std_waker`.
+        let this = unsafe { &*(data as *const Self) };
+        let cloned = this.clone_capturing_trace();
+        RawWaker::new(Box::into_raw(cloned) as *const (), &Self::VTABLE)
+    }
+
+    unsafe fn raw_wake(data: *const ()) {
+        // SAFETY: `wake` consumes the waker per the `RawWakerVTable` contract: forward the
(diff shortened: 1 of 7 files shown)


docker/sui-indexer-alt-graphql/Dockerfile.lsan
amnn, line 1: If it seems like this won't bitrot, ~~then a better home is `docker/sui-indexer-alt-graphql/Dockerfile.lsan`~~, but given the details here and the fact that we are not going to be building this image regularly it seems likely it would.

In which case we should not check this file in. Include the context for how the LSAN build was set-up in the PR description, and trust that if we need to do this in future, an agent will figure out how to set it up for whatever the state of the repo is then. The details of how it needed to be configured today are more likely to throw it off course than help.

**EDIT** apologies, I got confused and thought this was in `crates/sui-indexer-alt-graphql` initially.

crates/sui-futures/src/timeout_trace.rs
amnn, line 1: This module deserves a generous doc comment explaining how we're using the `Waker` contract to do what we want here (IIUC, we're relying on the fact that whenever a future returns `pending` it's supposed to clone and stash the waker given to it so we can piggy back off the clone to capture the trace that is currently pending).

crates/sui-futures/src/timeout_trace.rs
```diff
@@ -0,0 +1,339 @@
...
+/// e.g. when a caller resolves several sibling branches concurrently.
+#[derive(Debug)]
+pub struct TimeoutElapsed {
+    pub active_traces: Vec<Arc<SpanTrace>>,
+}
+
+/// Shared state for a single timeout's final poll: the traces captured so far, and the real
+/// waker to forward wake-ups to. Each clone's trace is only present in `active_traces` for as
+/// long as that clone is alive -- once it's dropped, its `Weak` naturally stops upgrading, so
+/// nothing needs to be nulled out by hand.
```
amnn: Also worth explaining what is keeping strong `Arc<SpanTrace>` alive here, to make the connection clear.

amnn (approved):
Another nice find! The implementation is tricky/delicate and so deserves a lot of docs, but it looks good to me, thanks @evan-wall-mysten.

I'd also avoid committing the Dockerfile for now, as it's unlikely to withstand the test of time.

crates/sui-futures/src/timeout_trace.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/// A `Waker` that captures a [`SpanTrace`] every time it's cloned -- each clone marks a distinct
+/// still-pending await point -- and forwards actual wake-ups to `inner_waker`.
+struct TracingWaker {
+    inner: Arc<TracingWakerInner>,
+    /// This clone's own trace, kept alive for as long as this waker is (never read directly --
+    /// `active_traces` observes it through the `Weak` pushed alongside it). `None` for the
+    /// original waker, which isn't itself an await point.
+    _trace: Option<Arc<SpanTrace>>,
```
amnn: One observation is that the amount of time between when you capture traces and take them is quite short, while in the case of a timeout, there's going to be a future that's been in a pending state for a long time and owns this `TracingWaker`, redundantly holding onto the `SpanTrace`.

To fix that, I think you need to:

- Store the captured traces in `active_traces` directly, when you clone.
- Have `TracingWaker` store a `Weak` reference to the inner struct.
- Have `TracingWaker` communicate with the outer waker through `Arc<()>`:

Something like this:

```rust
struct Inner {
    traces: Mutex<Vec<(SpanTrace, Weak<()>)>>,
    inner: Waker,
}

struct TracingWaker {
    inner: Weak<Inner>,
    signal: Arc<()>,
}
```

amnn: Also you could consider using an unbounded mpsc channel instead of a `Mutex<Vec<_>>` to hold the traces, as you are essentially building a concurrent queue here, and that's really what an `mpsc` channel is.

amnn: That approach would simplify other things too:

- You no longer need to keep a shared `inner` struct around.
- Your `TracingWaker` can be simpler -- it doesn't actually need to support signaling wake because at the point you're using it you know you're going to return a ready timeout response if the inner future doesn't return a ready response.

docker/sui-indexer-alt-graphql/Dockerfile.lsan
evan-wall-mysten: Removed, I thought it may have been useful because it took claude a few iterations to come up with one that works, but I agree it isn't that useful to keep because some of the versions are already out of date because they need to be compatible with an old version of `rustix` that was somehow being used over the most recent one.

crates/sui-futures/src/timeout_trace.rs
evan-wall-mysten: Removed based on your other comment.

crates/sui-futures/src/timeout_trace.rs
evan-wall-mysten: > Also you could consider using an unbounded mpsc channel instead of a Mutex<Vec<_>> to hold the traces, as you are essentially building a concurrent queue here, and that's really what an mpsc channel is.

Switched to a bounded channel that silently fails when trying to add over 64 spans. This would only matter if over 64 resolver methods timed out at once.

> That approach would simplify other things too:
>
> You no longer need to keep a shared inner struct around.

Removed the inner struct.

> Your TracingWaker can be simpler -- it doesn't actually need to support signaling wake because at the point you're using it you know you're going to return a ready timeout response if the inner future doesn't return a ready response.

I tried this, but the `only_retained_clone_leaves_a_trace` test fails
```
thread 'timeout_trace::tests::only_retained_clone_leaves_a_trace' panicked:
assertion `left == right` failed: "the touched-but-dropped clone shouldn't leave a trace; only the retained one should"
  left: 2
 right: 1
```

crates/sui-futures/src/timeout_trace.rs
```diff
@@ -0,0 +1,455 @@
...
+//!
+//! ## The technique
+//!
+//! Any `Future::poll` that returns [`Poll::Pending`] is required, by the `Waker` contract, to
+//! arrange for its task to be polled again later -- almost always by cloning the `Waker` it was
+//! given and stashing the clone somewhere (a timer wheel, an I/O reactor's readiness slot, a
+//! channel's waiter list, ...) so it can be woken once whatever it's waiting on is ready. In
+//! other words: *cloning the waker is what a future does to say "I'm going to sleep and someone
+//! needs to wake me up later"* -- it's the general-purpose signal for "this is a genuine await
+//! point".
```
amnn, line 15:
```suggestion
//! channel's waiter list, ...) so it can be woken once whatever it's waiting on is ready.
```

(Just AI doing AI docs things)

crates/sui-futures/src/timeout_trace.rs
```diff
@@ -0,0 +1,455 @@
...
+use std::task::Waker;
+use std::time::Duration;
+
+use pin_project::pin_project;
+use tokio::sync::mpsc;
+use tokio::time::Sleep;
+use tracing_error::SpanTrace;
+
+/// Capacity of the channel `TracingWaker` clones send captured traces down. Bounded rather than
+/// unbounded (the repo disallows `mpsc::unbounded_channel`); sends only ever happen synchronously
```
amnn, line 50: I think it would be fine for us to use an unbounded channel in this case (by the same logic that it was fine to use a `Vec<_>` before). It's possible to override the repo's restriction on a case-by-case basis.

crates/sui-futures/src/timeout_trace.rs
amnn: > I tried this, but the `only_retained_clone_leaves_a_trace` test fails

It looks like everything is working now (`TracingWaker` does not forward the wake, the test exists and is enabled, CI is passing) -- what was the issue and fix in the end?

evan-wall-mysten: Oh I was misinterpreting your comment then to remove the `Weak<()>` which was causing the test failure. If I addressed it then there is no issue.

crates/sui-futures/src/timeout_trace.rs
evan-wall-mysten: Updated and included an explanation why it is safe.

evan-wall-mysten:
Forgot to take off automerge. This addresses the remaining comments https://github.com/MystenLabs/sui/pull/27615/

crates/sui-futures/src/timeout_trace.rs
amnn: Ah no, by "signaling wake" I meant holding on to the inner waker to forward `wake_by_ref` calls to.

PR #27594, opened 2026-08-06
[regex-reference-safety] Even more borrow checker tests 

tnowacki:
## Description 

- Relaxed `VecImmBorrow` so that it can be used even if there are outstanding mutable references.
  - This was fixed by first calling `freeze`, which is what happens in the source language
- A lot more `mvir` tests
- Some proptests that attempt to make a concrete reference state and then check various properties against the graph
  - This is similar to the bgt tests but with proptests and some internal probing 
  - It is also a bit easier to be "exhaustive" in these tests
- Some additional hardening around canonicalize 

## Test plan 

- This is tests

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -612,7 +624,23 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
...
     pub fn canonicalize(&mut self, remapping: &BTreeMap<Ref, u32>) -> Result<()> {
+        let mut canonicalized = BTreeSet::new();
+        for r in self.nodes.keys() {
+            let r_canon = (*r).canonicalize(remapping)?;
+            ensure!(
+                canonicalized.insert(r_canon),
+                "remapping is not injective, two references map to {:?}",
+                r_canon
+            );
+        }
```
tnowacki, line 642: This hardening was previously only a debug_assert

external-crates/move/crates/move-bytecode-verifier/src/regex_reference_safety/mod.rs
```diff
@@ -356,10 +356,11 @@ fn execute_inner(
 
         Bytecode::VecImmBorrow(_) => {
             safe_assert!(safe_unwrap!(verifier.stack.pop()).is_non_ref());
-            let vec_ref = safe_unwrap!(verifier.stack.pop());
+            let vec_ref = safe_unwrap!(safe_unwrap!(verifier.stack.pop()).to_ref());
+            let imm_vec_ref = state.freeze_ref(offset, vec_ref, meter)?;
```
tnowacki, line 360: Without this, it would treat this as a call that could write to the reference. Since we know that is not the case, we can just freeze the reference first (which is always allowed in the new system). 

This "limitation" exists in the current borrow checker, but is not reachable through the compiler since the compiler will always insert the freeze.

external-crates/move/crates/bytecode-verifier-transactional-tests/tests/reference_safety/vec_unpack.mvir
```diff
@@ -0,0 +1,53 @@
...
+    _ = *move(r);
+    _ = move(a);
+    _ = move(b);
+    return;
+}
+
+}
+
+//# publish
+mvir 0x3::vunpack_through_ref_invalid {
```
tzakian: Nit: name shouldn't have `invalid` in it?

external-crates/move/crates/move-regex-borrow-graph/src/tests/graph_closure_tests.rs
```diff
@@ -0,0 +1,1638 @@
...
+proptest! {
+    #![proptest_config(ProptestConfig { cases: 512, ..ProptestConfig::default() })]
+
+    /// The closure/completeness property, checked after every step of a random valid history.
+    #[test]
+    fn graph_is_complete_against_the_concrete_model(
+        ops in prop::collection::vec(arb_raw_op(), 0..=14),
+    ) {
+        let mut model = Model::new();
+        model.new_root(true);
```
tzakian, line 521: should this just be calling `run_history`?

tzakian (approved):
so. many. tests 🪄

external-crates/move/crates/move-regex-borrow-graph/src/tests/graph_closure_tests.rs
tnowacki: It doesn't quite work because of invariants, but I did consolidate this

PR #27596, opened 2026-08-06
[graphql-stream] Extract per-item build_edge from build_grpc_connection

tpham-mysten:
## Description

`build_grpc_connection` inlines the construction of each edge while draining a bitmap-scan page. Factor that per-item step into a `build_edge` helper on both the transaction and event modules, so the subscription scan loop can build edges one at a time and reuse the same conversion. No behavior change.

## Test plan

Covered by the existing `build_grpc_connection_*` unit tests in the transaction and event modules.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


wlmyng (approved):
this motivates me to refactor the streamdpage -> edge/ connection story, i'll have something for this soon

tpham-mysten:
Thanks @wlmyng!

PR #27598, opened 2026-08-06
Populate allowed proposers from the transaction driver in simulate

mystenmark:
## Description

The simulate API now returns a `Validity` expiration naming the validators this node would submit to, so a client that signs and submits what simulate handed back gets a transaction nobody else can amplify into consensus. Sizing the set at 3 keeps it submittable when some of those validators are offline, and is exactly `MAX_UNPAID_ALLOWED_PROPOSERS`, so the transaction stays valid at the reference gas price.

The proposers come from the transaction driver's `ValidatorClientMonitor` — the same ranking `RequestRetrier` uses to pick submission targets, so the set names the validators the transaction was going to be sent to anyway. A narrow `ProposerSelector` trait in `sui-types` carries them to the RPC service, mirroring how `TransactionExecutor` already works and keeping `sui-rpc-api` free of a `sui-core` dependency.

The behavior can be disabled per node with the new `enable_simulate_allowed_proposers` NodeConfig flag (default on): when off, the proposer selector is simply not wired into the RPC service, and simulate leaves transactions unrestricted exactly as if no driver were present.

Coin-paid transactions are restricted too: they need no validity window of their own (the gas coin provides replay protection), so a `Validity` expiration is added only when a proposer set is actually available, and the transaction is otherwise returned exactly as resolved.

Before any latency has been observed the monitor's ranking is an arbitrary shuffle, so no proposer set is emitted — the address-balance paths fall back to `ValidDuring` and coin-paid transactions keep `None` — rather than pinning the transaction to three validators picked at random. A caller-supplied expiration is never overwritten, and no proposer set is offered on networks where the `allowed_proposers` protocol flag is off, since `validity_check` would reject the very transaction simulate handed back.

Proto and SDK conversions for `Validity` are lossless as of this PR, so the proposer set actually survives the trip out of simulate.

### Dependency bump

`sui-sdk-types`, `sui-crypto`, and `sui-rpc` move to `d71f9d6`, which picks up MystenLabs/sui-rust-sdk#293 (the `Validity` expiration and its wire format) and MystenLabs/sui-apis#31 (the proto). All three bump together — splitting them puts two versions of `sui_sdk_types` in the graph.

Also drops the stacked-on-#27454 note: that PR has landed, so this now targets `main` directly.

## Test plan

`proposer_selector_tests` cover the three properties the expiration depends on: no set before any observation, and otherwise a set that is bounded by `max`, strictly increasing, within the committee, and stamped with the current epoch — plus a committee smaller than `max`.

`validity_transaction_expiration_round_trips_through_proto` pins the conversion that used to drop the proposer set, with and without one.

Not covered: an end-to-end test that simulate itself emits `Validity`. It needs the driver's health-check loop to have published latencies, which does not happen within a short test.

---

## Release notes

- [ ] Protocol: 
- [x] gRPC: `SimulateTransaction` now returns a `VALIDITY` expiration carrying the validators allowed to propose the transaction, where it previously returned `VALID_DURING`. Clients that sign and submit the returned transaction unchanged need no action; clients that reconstruct the expiration themselves should carry the new `allowed_proposers` field through, or the transaction digest will not match.
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-core/src/transaction_driver/mod.rs
```diff
@@ -148,6 +155,36 @@ where
             .select_shuffled_preferred_validators(&authority_aggregator.committee, delta)
     }
 
+    /// The validators this node would prefer to submit to, as committee indices.
+    ///
+    /// These are the same targets `RequestRetrier` would pick, so a transaction restricted to them
+    /// names the validators it was going to be sent to anyway.
+    fn preferred_proposers_impl(&self, max: usize) -> Option<AllowedProposers> {
+        // Before any latency has been observed the ranking is an arbitrary shuffle, so pinning to
+        // it would be worse than leaving the transaction unrestricted.
```
alex-mysten, line 164: If we are considering penalizing unrestricted transactions, is it actually worse to select at random in the absence of data?

crates/sui-core/src/transaction_driver/unit_tests/proposer_selector_tests.rs
```diff
@@ -0,0 +1,98 @@
...
+/// the health check loop would.
+fn observe_latencies(driver: &TransactionDriver<MockAuthorityApi>) {
+    let auth_agg = driver.authority_aggregator().load_full();
+    for (i, validator) in auth_agg.committee.names().enumerate() {
+        driver
+            .client_monitor
+            .record_interaction_result(OperationFeedback {
+                authority_name: *validator,
+                display_name: auth_agg.get_display_name(validator),
+                operation: OperationType::SharedObjectFinality,
```
alex-mysten, line 55: Out of scope, but the naming tripped me up here, but confusing that a ping is `SharedObjectFinality`

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -131,7 +133,11 @@ pub fn simulate_transaction(
                 gasless_tx.gas_data_mut().price = 0;
                 gasless_tx.gas_data_mut().budget = 0;
                 // All gassless txns have to have a correct `ValidDuring` TransactionExpiration.
-                set_valid_during_transaction_expiration(service, &mut gasless_tx)?;
+                set_valid_during_transaction_expiration(
```
alex-mysten: perhaps we should change the name & comment?

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -468,31 +479,115 @@ fn mock_gas_storage_cost(
...
+/// is only touched when the caller left it unset and a proposer set is actually available;
+/// otherwise the transaction is returned exactly as resolved.
+fn restrict_transaction_proposers(
+    service: &RpcService,
+    protocol_config: &ProtocolConfig,
+    transaction: &mut sui_types::transaction::TransactionData,
+) -> Result<()> {
+    if !matches!(transaction.expiration(), TransactionExpiration::None) {
+        return Ok(());
+    }
```
alex-mysten, line 544: Are we penalizing callers that do provide a transaction expiration? ie should we merge or would that behavior be surprising?

mystenmark: i also wasn't sure about this. there's a good argument we should add allowed proposers to any transaction that doesn't specify them, but then there is no way for a user to say "i don't want allowed proposers".

I suppose in that case the recommendation could be for the user to simply remove the proposers after receiving the transaction back, but that sounds a bit argument.

@hayes or @bmwill might have thoughts

crates/sui-core/src/transaction_driver/mod.rs
akichidis: That would be my comment as well. I believe we should always populate the the allowed proposers (when config enabled) - even if those choices are not optimal - so we can more comfortably distinguish/penalise the ones who don't set it (and should not be coming from our full nodes)

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
```diff
@@ -468,31 +479,115 @@ fn mock_gas_storage_cost(
...
+    protocol_config: &ProtocolConfig,
+    current_epoch: u64,
+) -> Option<AllowedProposers> {
+    if !protocol_config.allowed_proposers() {
+        return None;
+    }
+    service
+        .proposer_selector
+        .as_ref()
+        .and_then(|selector| selector.preferred_proposers(MAX_ALLOWED_PROPOSERS))
```
akichidis, line 584: I might be missing something from the current design, but at which point we (or should?) take into account the SIP-45 paid amplification? I would expect to get the preferred proposers up to what the gas payment would allow to?

crates/sui-core/src/transaction_driver/mod.rs
mystenmark: my only concern here is that if some portion of stake is down, we could end up picking dead stake such that the transaction cannot be submitted. In the worst case this would be 1/3^3 = 3.7% of transactions.

On the other hand, a transaction with no allowed proposers might suffer a small latency increase.

Since both conditions will be only transient (as soon as the FN makes some successful submissions it will begin populating allowed proposers), I think it is better to have a small number of slightly slower transactions than it is  to have a small number of unsubmittable transactions.

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
mystenmark: I think anybody who is doing paid amplification can probably just set the proposers and gas price themselves, its very much a power-user feature.

crates/sui-core/src/transaction_driver/mod.rs
alex-mysten: Your reasoning makes sense. 

Just trying to unpack how these latencies are measured. We have `health_check` loop, which calls `update_cached_latencies` at each tick, which unconditionally writes cached_latencies for all validators, setting it to a known result from a `SharedObjectFinality` transaction, or to `MAX_LATENCY` if none is yet recorded. 

Separately, we have `run_latency_checks`, which fires a `ping` to each validator, after a period of jitter. The `ping` is a consensus round trip and is therefore recorded as a `SharedObjectFinality` operation. The next tick of the `health_check` loop would call `update_cached_latencies`, setting the measured latency accordingly.

It seems that there is a likely window between the two loops where health check has ticked before any ping has returned, setting all validators to MAX_LATENCY. Since cached_latencies is non-empty, your check here passes and we select random validators. Is that right?

mystenmark: Hmm, I didn't look too deeply into how the latency measurements are done, I was just trying to select whatever validators we are eventually going to submit to anyway. Will take a closer look

akichidis: > my only concern here is that if some portion of stake is down, we could end up picking dead stake such that the transaction cannot be submitted. In the worst case this would be 1/3^3 = 3.7% of transactions.

Even in this case  I would lean towards , as @alex-mysten said, just fallback to the random choice. The `select_shuffled_preferred_validators` basically is already doing that internally and return you the set of validators . This is mostly an issue that someone would deal with during the full node bootstrap where there are no scores and it should resolve fairly quickly , and it would be no different from what happens today anyways. Having said that we could even optimise further the client monitor in future work to close this gap.
Given the above I would personally accept this tradeoff knowing that a full node should always return back the allowed proposers filled in, rather than having to deal with those edge cases.

crates/sui-rpc-api/src/grpc/v2/transaction_execution_service/simulate/mod.rs
akichidis: Agreed, although I feel that we would have covered all the angles here if we were to respect that? Also probably quite low effort to adapt the code?

crates/sui-core/src/transaction_driver/mod.rs
```diff
@@ -148,6 +155,36 @@ where
...
+        }
+
+        let authority_aggregator = self.authority_aggregator.load();
+        let committee = &authority_aggregator.committee;
+        let mut proposers: Vec<u32> = self
+            .client_monitor
+            .select_shuffled_preferred_validators(committee, SELECT_LATENCY_DELTA)
+            .into_iter()
+            .filter_map(|name| committee.authority_index(&name))
+            .take(max)
```
alex-mysten, line 176: Do we intend for the fullnode to submit to all proposers concurrently with gas_price == rgp?

Separately, without `defer_unpaid_amplification` or another penalty, it seems that unrestricted tx are strictly better than tx with allowed proposers. Do we plan to reactivate unpaid amplification deferral, or impose a different penalty?

mystenmark: I think there are two options on the table:
1. Reactive unpaid amplification deferral, as you say
2. For un-restricted transactions, have a deterministic "fast proposer" + random delays on other validators. This would allow good latency in most cases, while causing most amplification attempts to be dropped by honest validators before submission to consensus

crates/sui-core/src/transaction_driver/mod.rs
mystenmark: I pushed a small change here which lets us prioritize validators that have answered health checks in the case where there is no ping data yet. I think this should put us in a pretty good place

crates/sui-core/src/validator_client_monitor/stats.rs
```diff
@@ -140,8 +140,12 @@ impl ClientObservedStats {
 
         let operation = OperationType::SharedObjectFinality;
         let Some(latency) = stats.average_latencies.get(&operation) else {
-            // For the target validator and operation type, no latency data has been recorded yet.
-            return MAX_LATENCY;
+            // No latency measurements yet, but health checks may have already told us the
+            // validator is unreachable. Rank known-unhealthy below merely-unmeasured, so that
+            // cold-start selection avoids validators we know are down.
+            let reliability_penalty = MAX_LATENCY
+                .mul_f64((1.0 - stats.reliability.get()) * self.config.reliability_weight);
```
alex-mysten, line 147: Nit: can't we deduplicate with the block below by computing this unconditionally?

We don't discriminate between otherwise healthy validators. Should we use health check latency as a second tiebreaker?

mystenmark: I think that would be nice but it would be a larger change - if we record health check latency it would end up being less than ping latency once that arrives. I kind of want to rewrite a lot of this code for this and other reasons, but this is probably not the PR.

alex-mysten: I meant as a tiebreaker between two validators at MAX_LATENCY. We could sort by (ping_latency, health_check_latency) with each adjusted for reliability.

Not blocking, just an idea

mystenmark: lets do that in another PR, its a good idea

alex-mysten (approved):

PR #27599, opened 2026-08-06
Submit proposer-restricted transactions only to their allowed proposers

mystenmark:
## Description

Third in the stack, on top of #27598 — review that first, and retarget this once it lands.

A transaction that names its proposers can only enter consensus through one of them: an honest validator outside the set will not propose it, and a byzantine one that does invalidates its own block. Submitting there is therefore never useful, so `RequestRetrier` drops those targets before any request goes out. The amplification factor is clamped to the size of the set for the same reason — asking for more targets than exist just runs the retrier dry.

This applies to submission only. Effects can be read from any validator, so the certifier's retrier stays unrestricted.

A set naming another epoch's committee is ignored here exactly as the validators ignore it, which leaves the transaction submittable anywhere rather than nowhere.

### Why no protocol-config gate

This is client-side submission policy, not consensus behavior: it changes only which validators receive a gRPC request, so there is no cross-validator consistency to protect. On a network where `allowed_proposers` is off, any transaction carrying a `Validity` expiration — the only variant that can name proposers — is rejected as `Unsupported` by every validator's `validity_check`, whatever its proposer set contains and wherever it is submitted. The driver honoring the restriction therefore changes nothing but which validators return that rejection. (The driver also has no protocol config to consult — it holds a `Committee`, not an epoch store — so a gate would add plumbing to distinguish outcomes that are already identical.)

Together with #27598 this closes the loop: simulate names the validators this node prefers, and the driver then submits only to them.

## Test plan

`test_allowed_proposers` checks that only the named validators survive into the retrier; `test_allowed_proposers_outside_committee` checks that an index naming nobody leaves nothing to submit to, rather than silently falling back to the whole committee.

---

## Release notes

- [ ] Protocol: 
- [x] Nodes (Validators and Full nodes): Full nodes now submit a proposer-restricted transaction only to the validators it names, instead of to any validator. No configuration change is needed; transactions without a proposer set are unaffected.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-core/src/transaction_driver/transaction_submitter.rs
```diff
@@ -65,10 +66,20 @@ impl TransactionSubmitter {
...
-        let amplification_factor = amplification_factor
-            .max(1)
-            .min(authority_aggregator.committee.num_members() as u64);
+        // Only the validators the transaction names may propose it, so submitting anywhere else
+        // would be wasted; a set naming another epoch's committee is ignored, exactly as the
+        // validators ignore it.
+        let allowed_proposers = request.transaction.as_ref().and_then(|tx| {
+            tx.transaction_data()
+                .expiration()
+                .allowed_proposers(authority_aggregator.committee.epoch())
```
akichidis, line 75: I don't remember, but do we perform at any earlier point any validation against the allowed proposers according to the SIP-45 checks?

akichidis: I would also propose adding a metric that counts the transactions with allowed proposers defined Vs the ones that don't.

mystenmark: Yes this is done during `validity_check`. good call on the metrics.

alex-mysten (approved):

PR #27602, opened 2026-08-06
[gql] generic `StreamConnection` and consolidate pagination logic on `Page`

wlmyng:
## Description 

Adds a `StreamConnection` with concrete types `EventConnection` and `TransactionConnection`. This enables `Page` to handle a `StreamPage` and return a `StreamConnection` generic on the output type. Barring the actual translation from the streamed item into a graphql node, everything else is shared pagination logic.

## Test plan 

Existing tests should cover

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
@@ -6,11 +6,6 @@ use std::sync::Arc;
 use anyhow::Context as _;
 use async_graphql::Context;
 use async_graphql::Object;
-use async_graphql::connection::Connection;
-use async_graphql::connection::CursorType;
-use async_graphql::connection::Edge;
-use async_graphql::connection::EmptyFields;
-use async_graphql::connection::PageInfo;
 use diesel::prelude::QueryableByName;
 use diesel::sql_types::BigInt;
 use itertools::Itertools;
@@ -55,6 +50,7 @@ use crate::api::types::transaction::Transaction;
 use crate::error::RpcError;
 use crate::extensions::query_limits;
 use crate::pagination::Page;
+use crate::pagination::StreamConnection;
 use crate::scope::Scope;
 use crate::task::watermark::Watermarks;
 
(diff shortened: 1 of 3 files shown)


crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -455,6 +416,14 @@ impl TryFrom<CursorToken> for EventToken {
     }
 }
 
+impl TryFrom<CursorToken> for CEvent {
```
tpham-mysten, line 418: This is nice!

tpham-mysten (approved):
Looking good. Thanks @wlmyng

crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -94,10 +90,7 @@ pub struct TransactionToken {
 pub type CTransaction = OpaqueCursor<TransactionToken>;
 
 /// Custom `Connection` for transactions to support partially-filled pages.
-pub(crate) struct TransactionConnection {
-    pub edges: Vec<Edge<String, Transaction, EmptyFields>>,
-    pub page_info: PageInfo,
-}
+pub(crate) type TransactionConnection = StreamConnection<Transaction>;
```
amnn, line 93: nit (optional): could we drop the alias now we have this abstraction?

crates/sui-indexer-alt-graphql/src/pagination.rs
```diff
@@ -330,6 +346,114 @@ impl<C: CursorType + Eq + PartialEq + Clone> Page<C> {
...
+    /// GraphQL cursor string in this page's cursor format.
+    fn encode_stream_cursor(bytes: &[u8]) -> Result<String, RpcError> {
+        let token = CursorToken::decode(bytes).context("Failed to decode stream cursor")?;
+        let cursor = C::try_from(token).context("Unexpected position in stream cursor")?;
+        Ok(cursor.encode_cursor())
+    }
+}
+
+#[Object(
+    concrete(name = "EventConnection", params(Event)),
```
amnn, line 405: TIL -- but could you not apply the same pattern that the built-in `Connection` type has for deciding the schema name?

wlmyng: Ah yeah, why didn't I make that connection... https://github.com/MystenLabs/sui/pull/27661

I suppose `concrete` is still useful for things that aren't straightforward to name but for these things yeah we can take a page directly from the stock `Connection`

amnn: so many puns 😅

PR #27604, opened 2026-08-07
[move-compiler] Fix cast optimization

tnowacki:
## Description 

- Fixes erroneous inlining of casts that always error at runtime

## Test plan 

- Added tests 

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [X] CLI: Fixes bug in Move compiler that would inline cast (`as`) operations that always error
- [ ] Rust SDK:
- [ ] Indexing Framework:
---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


cgswords (approved):
LGTM

PR #27611, opened 2026-08-07
[gql] Require ledger grpc for package resolver [1/n deprecating `kv_packages` on gql]

wlmyng:
## Description

Part of migrating off the Postgres `kv_packages` table. The package resolver fetches by storage ID, which identifies exactly one immutable package version. This can be answered by ledger gRPC (either backed by archival store or fullnode.)

Adds a `GrpcPackageStore` to alt-reader that does batch get objects.

Graphql requires `--ledger-grpc-url` configured.

`PackageCache` wraps a boxed store so jsonrpc-alt can continue to use `kv_packages`.

## Test plan

Existing tests should pass

## Release notes

- [ ] Protocol
- [ ] gRPC
- [ ] JSON-RPC
- [ ] GraphQL
- [ ] CLI
- [ ] Rust SDK

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/lib.rs
+++ b/crates/sui-indexer-alt-graphql/src/lib.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::Context as _;
+use anyhow::bail;
 use api::types::address::IAddressable;
 use api::types::move_datatype::IMoveDatatype;
 use api::types::move_object::IMoveObject;
@@ -47,7 +48,7 @@ use sui_indexer_alt_reader::fullnode_client::FullnodeArgs;
 use sui_indexer_alt_reader::fullnode_client::FullnodeClient;
 use sui_indexer_alt_reader::kv_loader::KvArgs;
 use sui_indexer_alt_reader::kv_loader::KvLoader;
-use sui_indexer_alt_reader::package_resolver::DbPackageStore;
+use sui_indexer_alt_reader::package_resolver::GrpcPackageStore;
 use sui_indexer_alt_reader::package_resolver::PackageCache;
 use sui_indexer_alt_reader::pg_reader::PgReader;
 use sui_indexer_alt_reader::pg_reader::db::DbArgs;
@@ -361,7 +362,11 @@ pub async fn start_rpc(
             .context("--ledger-grpc-url must be configured")?,
     );
 
-    let package_store = Arc::new(PackageCache::new(DbPackageStore::new(pg_loader.clone())));
+    // Package resolution is served exclusively by the ledger gRPC service (kv-rpc or fullnode).
+    let Some(reader) = &ledger_grpc_reader else {
+        bail!("Package resolution requires a ledger gRPC service; configure --ledger-grpc-url");
+    };
+    let package_store = Arc::new(PackageCache::new(Box::new(GrpcPackageStore::new(reader))));
 
     let system_package_task = SystemPackageTask::new(
         system_package_task_args,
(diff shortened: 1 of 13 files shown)


crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -367,7 +368,11 @@ pub async fn start_rpc(
         pg_loader.clone(),
     );
 
-    let package_store = Arc::new(PackageCache::new(DbPackageStore::new(pg_loader.clone())));
+    // Package resolution is served exclusively by the ledger gRPC service (kv-rpc or fullnode).
+    let Some(reader) = &ledger_grpc_reader else {
+        bail!("Package resolution requires a ledger gRPC service; configure --ledger-grpc-url");
```
tpham-mysten, line 367: This seems like a breaking deployment change. How do we plan to communicate this?

crates/sui-indexer-alt-reader/src/package_resolver.rs
```diff
@@ -110,3 +134,64 @@ impl Loader<PackageKey> for PgReader {
...
+
+        let batch_response = self
+            .batch_get_objects(request)
+            .await
+            .map_err(|e| store_err(GRPC_STORE, e))?;
+
+        for obj_result in batch_response.objects {
+            let Some(proto::get_object_result::Result::Object(object)) = obj_result.result else {
+                // Misses come back as per-object errors; leave the key unmapped so the store
+                // reports `PackageNotFound`.
```
tpham-mysten: This treats every per-object failure as a miss, so a genuine backend problem (`INTERNAL`, `UNAVAILABLE`, a deadline) surfaces to the caller as `PackageNotFound` rather than as a store error. That is probably confusing thing to debug from a GraphQL response. I think we should check for status to categorize the NotFound case particularly.

WDYT?

tpham-mysten (commented):
This change also impacts the logic in GraphQL subscription[ on package_eviction_task.rs](https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt-graphql/src/task/streaming/package_eviction_task.rs#L64).

For live streamed checkpoint, we are [caching](https://github.com/MystenLabs/sui/blob/main/crates/sui-indexer-alt-graphql/src/task/streaming/checkpoint_stream_task.rs#L492) the new published package in memory, and there is a background task `PackageEvictionTask` that is running every 5 mins to drain all items that are already indexed. As there was just kv_packages before, that `PackageEvictionTask` was checking the kv_packages watermark, which is not true anymore. We can fix that logic to just use ledger_service watermark as `ledger_grpc_reader` is already [required](https://github.com/MystenLabs/sui/blob/a6ef63025af9c8cec77af0dc830bf9237e4c2e9d/crates/sui-indexer-alt-graphql/src/lib.rs#L403) in Subscription.

crates/sui-indexer-alt-graphql/src/lib.rs
wlmyng: Initially, I thought that this shouldn't be a significant issue as the package resolver can be answered by ledger service on fullnode grpc. However, it's a little awkward if someone configures graphql to read from archival store instead ... I think we'll need to ask people to backfill the packages table

amnn (commented):
Does this need to be rebased over the changes from @evan-wall-mysten to make `ledger_grpc` configuration mandatory?

crates/sui-indexer-alt-graphql/src/lib.rs
wlmyng: I think I'll just wait for this stack https://github.com/MystenLabs/sui/pull/27666 to land

crates/sui-indexer-alt-reader/src/package_resolver.rs
wlmyng: Makes sense, I've split out not found from other errors

PR #27615, opened 2026-08-07
sui-futures: use an unbounded channel for capturing timeout traces

evan-wall-mysten:
## Description

Follow-up to #27593: `TracingWaker`'s trace channel was capped at 64 entries, silently dropping traces once a query had more concurrent branches pending than that. Switches it to an unbounded channel — safe here since sends only ever happen synchronously during the one poll call a `TracingWaker` is used for, and are always fully drained right after — with a `#[allow(clippy::disallowed_methods)]` exception for the repo's default ban on unbounded channels.

## Test plan

Covered by existing tests in `sui-futures::timeout_trace`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-futures/src/timeout_trace.rs
+++ b/crates/sui-futures/src/timeout_trace.rs
@@ -9,10 +9,7 @@
 //! Any `Future::poll` that returns [`Poll::Pending`] is required, by the `Waker` contract, to
 //! arrange for its task to be polled again later -- almost always by cloning the `Waker` it was
 //! given and stashing the clone somewhere (a timer wheel, an I/O reactor's readiness slot, a
-//! channel's waiter list, ...) so it can be woken once whatever it's waiting on is ready. In
-//! other words: *cloning the waker is what a future does to say "I'm going to sleep and someone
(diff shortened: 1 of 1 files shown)


evan-wall-mysten:
Forgot to disable automerge in https://github.com/MystenLabs/sui/pull/27593 this PR addresses the remaining comments from there.

crates/sui-futures/src/timeout_trace.rs
```diff
@@ -396,7 +381,8 @@ mod tests {
     /// borrowed the box, leaking it forever.
     #[test]
     fn wake_drops_the_waker_like_drop_does() {
-        let (sender, receiver) = mpsc::channel(TRACE_CHANNEL_CAPACITY);
+        #[allow(clippy::disallowed_methods)] // allow unbounded_channel()
```
amnn: nit: the comment here is redundant (ditto below)

amnn (approved):

PR #27635, opened 2026-08-10
[graphql-stream] Reject subscriptions that start too far ahead of the tip [19/n]

tpham-mysten:
## Description

A subscription can be asked to start from a checkpoint ahead of the tip (via `afterCheckpoint` or an `after` cursor). There is nothing to backfill ahead of the tip, so it just waits for the chain to reach that checkpoint before delivering from live, and a start far in the future holds the connection open indefinitely, so a client can cheaply tie up the service with subscriptions that only wait. This rejects a start more than `max_start_checkpoints_ahead_of_tip` (default 300, roughly a minute) past the tip, across the transaction, event, and checkpoint subscriptions. Follows up the discussion on #27580.

## Test plan

New e2e tests: each of the three subscriptions rejects a start set well past the tip with the expected error.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/checkpoint_subscription.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql_subscription/checkpoint_subscription.rs
@@ -509,6 +509,25 @@ async fn test_subscription_resume_with_after_checkpoint() {
 }
 
 #[tokio::test]
+async fn test_subscription_start_too_far_ahead_errors() {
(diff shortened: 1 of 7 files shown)


wlmyng (approved):
seems reasonable to enforce a cap

amnn (approved):
Nice!

PR #27645, opened 2026-08-10
[trace] Fixed handling of large integers in PTB execution traces

awelc:
## Description 

This PR fixed problems with handling large integers in PTB traces. It was discovered when trying to replay mainnet transaction with `HCvsgGqmWesDszy1dntMaZUvowH61cEmo8cgxXtwHj2k` digest and the concrete problem was in handling external events where `serde_json::json!` was failing (with an error) to materialized large integers into `serde_json::Value`. The solution is to materialize these into `serde_json::RawValue`.

Independently, when patching up the debugger code to reflect the Rust-side changes, it was discovered that it does not handle decoding of `u256` values correctly and this was fixed as well (with a separate trace debugger test added).

## Test plan 

All tests must pass (a new trace debugger test was added)


external-crates/move/crates/move-analyzer/trace-adapter/src/trace_utils.ts
```diff
@@ -606,7 +606,17 @@ export async function readTrace(
     const frameInfoStack: ITraceGenFrameInfo[] = [];
     for (const line of lineIterator) {
         eventCount++;
-        const event = JSON.parse(line) as JSONTraceEvent;
+        // Parse with a source-text reviver so that integer values outside JS's safe range
+        // (e.g. `u64`/`u128` Move values) are preserved exactly as their decimal string
+        // instead of being silently rounded to a double by a "regular"`JSON.parse`.
+        const event = JSON.parse(line, (_key: string, value: any, context?: { source?: string }) =>
```
tzakian, line 612: thanks JSON 👍🏻

external-crates/move/crates/move-trace-format/src/format.rs
```diff
@@ -179,7 +181,13 @@ pub enum TraceEvent {
         instruction: Box<String>,
     },
     Effect(Box<Effect>),
-    External(Box<serde_json::Value>),
+    // `serde_json::value::RawValue` implements neither `PartialEq` nor `Eq`, so the external
+    // payload is compared by its raw JSON text via `raw_value_eq`
+    External(#[educe(PartialEq(method(raw_value_eq)))] Box<serde_json::value::RawValue>),
```
tzakian: Do we need `PartialEq` or `Eq` at all? if we don't I'd prefer to just not derive it at all rather than pull in a dep.

tzakian (approved):
Main thing is to see if you can remove the derive/educe but otherwise LGTM

external-crates/move/crates/move-trace-format/src/format.rs
awelc: Good point! I should have checked... And indeed it does not look like it's needed so removed both `PartialEq` and `Eq` as well as the dependency

PR #27648, opened 2026-08-11
[move-compiler] Enable folding of constants in functions

tnowacki:
## Description 

- constant_fold would inline constants, but only within `const` declarations 
- This now also happens in functions 

## Test plan 

- Added tests 
---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [X] CLI: Move compiler now considers the value of constants when optimizing and folding expressions inside of functions.
- [ ] Rust SDK:
- [ ] Indexing Framework:

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


tnowacki:
Fixing #27605 after stack got screwed up

PR #27650, opened 2026-08-11
consensus: reset boot counter for simulated process restarts

mwtian:
## Description 

Fix a flakiness where the shared boot counter prevents amnesia recovery on validators restarted after crashing.

## Test plan 

CI

### DIFF
--- a/consensus/simtests/tests/consensus_tests.rs
+++ b/consensus/simtests/tests/consensus_tests.rs
@@ -19,7 +19,7 @@ mod consensus_tests {
     use consensus_core::NoopTransactionVerifier;
     use consensus_core::{
         BlockAPI, BlockStatus, CommitIndex, CommitRef, CommittedSubDag, Priority,
-        TransactionVerifier, ValidationError,
+        TransactionVerifier, ValidationError, storage::Store,
     };
     use consensus_simtests::node::{AuthorityNode, Config};
     use consensus_types::block::{BlockRef, BlockTimestampMs, TransactionIndex};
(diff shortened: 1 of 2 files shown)


consensus/simtests/tests/consensus_tests.rs
```diff
@@ -327,6 +327,62 @@ mod consensus_tests {
         load_handle.abort();
     }
 
+    // Restart one validator as a new process without its local store. The new process must
+    // recover its last own block from peers before it proposes another block.
```
arun-koshy, line 331: It looks like this assertion can pass via commit sync alone even if own block recovery were skipped. Nothing checks the restarted node proposes again after recovery. Worth also asserting the recovery half, i.e. that a post restart commit contains a block from this authority at a round above its pre restart proposals?

arun-koshy (approved):
worth adding a simtest that exercises the boot_counter > 0 restart path?

consensus/simtests/tests/consensus_tests.rs
akichidis: Good point. I guess on this test we can never be sure of its state - maybe the node has never proposed any block (low probability). Also if we don't do some explicit checks then maybe the `test_consensus_rolling_restarts_all_validators` already covers part of this logic?

mwtian: Renamed this to `test_consensus_amnesia_recovery`, and added assertion that new blocks get proposed and committed post recovery. In simtests, we don't need the test to verify amnesia recovery 100% of the runs. We just need to have high enough probability that the codepaths we are interested in are covered.

PR #27653, opened 2026-08-11
sui-indexer-alt-reader: drop Postgres KV fallback for ledger gRPC

evan-wall-mysten:
## Description

Follows #27557 — now that `LedgerGrpcReader` has full parity with Postgres for KV point-lookups (checkpoints, objects, transactions, events), drop the redundant Postgres-backed `Loader` impls and make `--ledger-grpc-url` required. This ripples into `sui-indexer-alt-jsonrpc` (shares the same `KvLoader`) and `sui start` localnet wiring, both of which relied on the Postgres fallback.

## Test plan

Manually verified `sui start --with-graphql --force-regenesis` boots and serves KV point-lookups purely through the new mandatory ledger gRPC wiring:
```
curl -s -X POST http://127.0.0.1:9125/graphql -H "Content-Type: application/json" \
  -d '{"query":"{ checkpoint(sequenceNumber: 0) { sequenceNumber digest timestamp } }"}'
```
returns the expected checkpoint. Otherwise covered by existing unit/e2e tests in `sui-indexer-alt-reader`, `sui-indexer-alt-graphql`, `sui-indexer-alt-jsonrpc`, and `sui-indexer-alt-e2e-tests`.

---

## Release notes

- [x] JSON-RPC: `--ledger-grpc-url` is now required to run `sui-indexer-alt-jsonrpc` — it no longer falls back to Postgres for key-value lookups (objects, checkpoints, transactions) when unset.
- [x] GraphQL: `--ledger-grpc-url` is now required to run `sui-indexer-alt-graphql`, for the same reason.

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/tests/graphql_execute_transaction_tests.rs
+++ b/crates/sui-indexer-alt-e2e-tests/tests/graphql_execute_transaction_tests.rs
@@ -125,12 +125,17 @@ impl GraphQlTestCluster {
 
         let fullnode_args = FullnodeArgs::new(validator_cluster.rpc_url().parse().unwrap());
 
+        let kv_args = KvArgs {
+            ledger_grpc_url: Some(validator_cluster.rpc_url().parse().unwrap()),
+            ..Default::default()
+        };
+
         // Start GraphQL server that connects directly to TestCluster's RPC
         let service = start_graphql(
             None, // No database - GraphQL will use fullnode RPC for executeTransaction
             fullnode_args,
             DbArgs::default(),
-            KvArgs::default(),
+            kv_args,
             ConsistentReaderArgs::default(),
             graphql_args,
             SystemPackageTaskArgs::default(),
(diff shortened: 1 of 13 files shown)


crates/sui-indexer-alt-graphql/Cargo.toml
```diff
@@ -67,6 +67,7 @@ sui-futures.workspace = true
 sui-indexer-alt-metrics.workspace = true
 sui-indexer-alt-reader.workspace = true
 sui-indexer-alt-schema.workspace = true
+sui-kvstore.workspace = true
```
wlmyng: Rather than having graphql directly take on this dependency, can we use the proto type from sui-rpc-api, or have alt-reader translate into some type and use that type instead?

wlmyng (approved):
lgtm, and from this another thing unearthed worth tackling eventually:

`TransactionContents` fused two ideas together, (1) which KV backend did this point lookup come from and (2) transaction data from external not from storage (executed, simulated, etc.)

Now that we've removed the `Pg` and `Bigtable` variants, to me it's strange to have `TransactionContents` bundle `LedgerGrpc` and `Executed` together, probably better to have these represented on graphql as an enum rather in the alt-reader

I think the story around `CheckpointedTransaction` and `ExecutedTransactionData` (sourced from the same proto::ExecutedTransaction) can be cleaned up eventually

I think this is related to stripping out the `KvLoader` newtype: on one hand it gives us benefits like not having to spell out `Arc<DataLoader<LedgerGrpcReader>>`, key-construction boilerplate, on the other we don't really need it anymore and graphql resolvers can consume the `CheckpointedTransaction` as it sees fit

Captured this in a linear issue

crates/sui-indexer-alt-graphql/Cargo.toml
evan-wall-mysten: Switched to use the proto.

PR #27655, opened 2026-08-11
consensus: add Mysticeti v3 transaction finalization

mwtian:
## Description

Add a separate transaction finalizer for Mysticeti V3.

The V3 finalizer:

- Uses signed votes from the next round for direct accept and reject decisions.
- Uses committed next-round voting evidence for depth-two indirect decisions.
- Records rejected transactions before execution and preserves them during recovery.

The Mysticeti V2 finalizer stays unchanged. V3 tests now use V3 committee thresholds. The FlexCommitter timestamp median now requires quorum stake and uses the parent-round quorum when leader stake is lower.

Scope note: This revision does not change vote-target selection or block verification. It asserts the existing invariant that proposer vote targets are above transaction-vote-tracker GC. Review threads on earlier effective-cutoff changes refer to code that is no longer in this PR.

## Test plan

- `SUI_SKIP_SIMTESTS=1 cargo nextest run --no-capture -p consensus-core`
  - 351 tests passed. 1 test was skipped.
- `cargo xclippy -p consensus-core -- -D warnings`
- `cargo test -p consensus-core commit_finalizer_v3`
  - 14 tests passed.
- Focused transaction-vote-tracker and V2/V3 proposer GC and cutoff tests.


consensus/core/src/transaction_vote_tracker.rs
```diff
@@ -142,12 +142,21 @@ impl TransactionVoteTracker {
...
-    pub(crate) fn get_own_votes(&self, block_refs: Vec<BlockRef>) -> Vec<BlockTransactionVotes> {
+    /// Retrieves own votes on peer block transactions and the effective vote cutoff.
+    ///
+    /// The returned cutoff covers blocks omitted by causal-history GC and blocks omitted by
+    /// vote-tracker GC. It is read under the same lock as the votes. Thus, later GC cannot make
+    /// the returned votes inconsistent with the returned cutoff.
+    pub(crate) fn get_own_votes(
+        &self,
+        block_refs: Vec<BlockRef>,
+        causal_history_cutoff_round: Round,
```
akichidis: Under the current code setup, where we always attempt propose after we have run the commit rule etc, at what scenario do we see the `gc_round >= block_ref.round` given that for a proposed block at round `R` we only count votes at round `R-1` ? Or is that more for bullet proofing in future cases where indeed we could propose for earlier rounds?

mwtian: Reverted most of tracker changes, but turned this skip into assert.

consensus/core/src/transaction_vote_tracker.rs
```diff
@@ -143,13 +143,17 @@ impl TransactionVoteTracker {
...
         for block_ref in block_refs {
-            if block_ref.round <= vote_tracker_state.gc_round {
-                continue;
-            }
+            assert!(
+                block_ref.round > vote_tracker_state.gc_round,
+                "Transaction vote target {} is at or below vote tracker GC round {}",
+                block_ref,
+                vote_tracker_state.gc_round,
+            );
```
akichidis: I believe this is valid/compatible even for the current(v2) version right?

consensus/core/src/commit_finalizer_v3.rs
```diff
@@ -0,0 +1,2231 @@
...
+        }
+    }
+
+    /// Returns the GC round for accept-vote evidence in a commit with leader round L.
+    ///
+    /// Consensus builds each committed sub-DAG before it advances GC. Leader rounds increase, and
+    /// consensus creates at most one commit for each leader round. Therefore, every commit before
+    /// the first depth-two decision uses a GC round at most one round above this value. Each first
+    /// vote is in a later round than its target. If the target is above this value, each first vote
+    /// is also above every GC round used to build the committed voting prefix.
```
akichidis: I am not sure I fully understand this.

consensus/core/src/commit_finalizer_v3.rs
```diff
@@ -0,0 +1,2231 @@
...
+                "direct_reject",
+            );
+        }
+    }
+
+    /// Returns the GC round for accept-vote evidence in a commit with leader round L.
+    ///
+    /// Consensus builds each committed sub-DAG before it advances GC. Leader rounds increase, and
+    /// consensus creates at most one commit for each leader round. Therefore, every commit before
+    /// the first depth-two decision uses a GC round at most one round above this value. Each first
```
akichidis: `above this value` , which value?

PR #27658, opened 2026-08-11
[simtest] enable one observer fullnode by default in simtests

akichidis:
## Description 

Consensus block streaming was only exercised by two dedicated observer tests. Every
sui-benchmark simtest cluster now also spawns an observer fullnode subscribed to the
first validator, so all scenarios — reconfig, restarts, crashes, forks — exercise the
observer, and each simulated-load test ends by asserting the observer caught up to the
RPC fullnode's highest executed checkpoint via block streaming (disable locally with
`SIM_STRESS_TEST_OBSERVER=0`).

First commit adds the test-infra (`TestClusterBuilder::with_observer_fullnode()`, peer
record derived from the network config instead of ~60 lines of manual wiring per test);
second commit flips the default.

Two non-obvious points:

- The cluster deliberately holds no handle to the observer (`observer_node()` looks it
  up in the swarm on demand): a held handle keeps a crashed instance's stores open, and
  the simulator-restarted node then panics re-acquiring its DB locks. Because of this
  the observer is *not* in the crash tests' keep-alive set — it gets killed and
  restarted like a validator, and the final catch-up assertion covers its recovery.
- `Swarm::validator_nodes()` filtered on `consensus_config.is_some()`, which
  misclassifies observer fullnodes (they carry a consensus config); it now filters on
  the intended node role.

## Test plan 

CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/tests/simtest.rs
+++ b/crates/sui-benchmark/tests/simtest.rs
@@ -276,17 +276,15 @@ mod test {
     /// rpc fullnode which are needed to run the benchmark.
     fn get_keep_alive_nodes(cluster: &TestCluster) -> HashSet<sui_simulator::task::NodeId> {
         let mut keep_alive_nodes = HashSet::new();
-        // The first fullnode in the swarm ins the rpc fullnode.
         keep_alive_nodes.insert(
             cluster
-                .swarm
-                .fullnodes()
-                .next()
-                .unwrap()
-                .get_node_handle()
-                .unwrap()
+                .fullnode_handle
+                .sui_node
                 .with(|n| n.get_sim_node_id()),
         );
+        // The observer fullnode is deliberately NOT kept alive: the cluster holds no
+        // handle to it, so the simulator can kill and restart it like any validator,
+        // and the final catch-up assertion covers its crash recovery.
         keep_alive_nodes.insert(sui_simulator::current_simnode_id());
         keep_alive_nodes
     }
@@ -1125,6 +1123,11 @@ mod test {
             ))
             .disable_fullnode_pruning()
             .with_synthetic_execution_time_injection();
+        // The standard setup includes an observer fullnode subscribed to the first
+        // validator, so that every scenario also exercises consensus block streaming.
+        if get_var("SIM_STRESS_TEST_OBSERVER", 1u8) != 0 {
+            builder = builder.with_observer_fullnode();
+        }
         if std::env::var("CHECKPOINTS_PER_EPOCH").is_ok() {
             eprintln!("CHECKPOINTS_PER_EPOCH env var is deprecated, use EPOCH_DURATION_MS");
         }
@@ -1486,6 +1489,7 @@ mod test {
         });
 
         if enable_surfer {
+            let surfer_cluster = test_cluster.clone();
             let surfer_task = tokio::spawn(async move {
                 // now do a sui-surfer test
                 let mut test_packages_dir = benchmark_move_base_dir();
@@ -1504,7 +1508,7 @@ mod test {
                     surf_strategy,
                     test_duration,
                     test_package_paths,
-                    test_cluster,
+                    surfer_cluster,
                     1, // skip first account for use by bench_task
                 )
                 .await;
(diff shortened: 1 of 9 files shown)


crates/sui-benchmark/tests/simtest.rs
```diff
@@ -1212,6 +1215,74 @@ mod test {
...
+                "Observer node failed to catch up to checkpoint {} via block streaming within \
+                 {:?}; highest finalized checkpoint was {:?}",
+                target_checkpoint,
+                catch_up_timeout,
+                read_observer_checkpoint(),
+            )
+        });
+
+        info!(
+            "Observer node caught up to checkpoint {} (target {}) via block streaming",
```
arun-koshy, line 1281: Can the node only catch up via block streaming or could it have caught up via state sync and thus mask that block streaming is not working?

crates/test-cluster/src/lib.rs
```diff
@@ -185,6 +185,27 @@ impl TestCluster {
...
+    /// keeps the running instance (and its open stores) alive, which prevents the
+    /// simulator from restarting the node after a crash. Acquire a fresh handle via
+    /// `Node::get_node_handle()` at the point of use and drop it promptly.
+    pub fn observer_node(&self) -> Option<&sui_swarm::memory::Node> {
+        self.swarm.observer_nodes().next()
+    }
+
+    /// Convenience method to start a new observer fullnode subscribed to the first
+    /// validator, which must have its observer server enabled.
+    pub async fn spawn_new_observer_node(&mut self) -> FullNodeHandle {
```
arun-koshy: is this needed? It doesnt seem to be called anywhere and returns the handle which seems to cause issues per your note above

crates/sui-benchmark/tests/simtest.rs
```diff
@@ -1450,6 +1522,10 @@ mod test {
             info!("Surfer disabled, running only benchmark task");
             bench_task.await.unwrap();
         }
+
+        // With load stopped, the observer (when present) must be able to reach the
+        // network's latest checkpoint via block streaming, in any scenario.
+        wait_for_observer_catch_up(&test_cluster, Duration::from_secs(120)).await;
```
arun-koshy, line 1528: If state-sync is an option for fallback that then may not be an issue.  But in tests that crash nodes including the observer node or the validator it is streaming from, it seems it can potentially lead to flakiness of the node catching up before the timeout

crates/test-cluster/src/lib.rs
```diff
@@ -1320,6 +1344,15 @@ impl TestClusterBuilder {
         self
     }
 
+    /// Adds a fullnode to the cluster that syncs as a consensus observer, subscribed to the
+    /// first validator. Unless a validator observer config is provided, the first validator
+    /// gets its observer server enabled. The observer is available via
+    /// `TestCluster::observer_node()`.
+    pub fn with_observer_fullnode(mut self) -> Self {
```
arun-koshy, line 1345: AI flagged - "If a test supplies a prebuilt network config via set_network_config, the validator-observer callback injected in build() is never applied (SwarmBuilder only uses it when it builds the network config itself), so this combination panics in observer_peer_record — same if a custom validator_observer_config doesn't enable index 0. Worth a line in this doc comment?"

arun-koshy (approved):

mwtian (approved):
Similar comments to Arun's, otherwise LGTM.

crates/test-cluster/src/lib.rs
akichidis: Ah yes this is a refactoring left over, thanks for flagging

crates/sui-benchmark/tests/simtest.rs
akichidis: The observer node has always state sync enabled, so in one way or another it will manage to catch up

PR #27659, opened 2026-08-11
Add package pause support

tzakian:
## Description 

Add the ability to pause and unpause a package across all versions.

## Test plan 

CI + additional tests

---

## Release notes

- [X] Protocol: Add support for being able to natively pause/unpause a package
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-transaction-checks/src/lib.rs
```diff
@@ -842,21 +842,23 @@ mod checked {
...
-        if !protocol_config.enable_package_version_forbid_list() {
+        let enforce_version_forbid_list = protocol_config.enable_package_version_forbid_list();
+        let enforce_global_pause = protocol_config.enable_package_global_pause();
+        if !enforce_version_forbid_list && !enforce_global_pause {
             return Ok(());
         }
 
         let TransactionKind::ProgrammableTransaction(pt) = transaction.kind() else {
             return Ok(());
         };
```
cgswords, line 858: This carveout seems odd to me.

tzakian: This only applies to/should only apply to PTBs/non-system transactions, so this seems like the right thing to me.

PR #27661, opened 2026-08-11
Use `name_type` to compute Connection name for `StreamConnection` at runtime

wlmyng:
## Description

`StreamConnection` required every node type to be registered by hand via `concrete(name = ..., params(...))`. Use `#[Object(name_type)]` with a `TypeName` impl instead — the same mechanism as the stock `Connection` — so each instantiation is named `{Node}Connection` automatically.

With naming handled by the generic, the `EventConnection` / `TransactionConnection` Rust type aliases no longer pull their weight, so they are removed in favor of writing `StreamConnection<Event>` / `StreamConnection<Transaction>` directly.

## Test plan

Schema type names are unchanged (`EventConnection`, `TransactionConnection`); covered by the existing schema snapshot test in `sui-indexer-alt-graphql`.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/query.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/query.rs
@@ -39,7 +39,6 @@ use crate::api::types::epoch::CEpoch;
 use crate::api::types::epoch::Epoch;
 use crate::api::types::event::CEvent;
 use crate::api::types::event::Event;
-use crate::api::types::event::EventConnection;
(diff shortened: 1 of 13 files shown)


amnn (approved):
Nice!

PR #27666, opened 2026-08-11
sui-indexer-alt-graphql: remove Postgres pagination for transactions and events

evan-wall-mysten:
## Description

Follow-up to #27653: `Transaction::paginate`/`Event::paginate` had a dual Postgres/ledger-gRPC implementation gated behind `--enable-list-apis`, but the gRPC List APIs (`ListTransactions`/`ListEvents`) already have full parity and `--ledger-grpc-url` is mandatory as of #27653. This removes `--enable-list-apis` and the flag, treating the gRPC List APIs as always enabled, and deletes the now-dead Postgres pagination/filtering code.

`available_range.rs`'s `Query.transactions`/`Query.events` pipeline mappings now point at the `ledger_grpc` watermark (the actual data source) instead of the Postgres pipelines they no longer read from.

This also depends on a fix to a pre-existing watermark race that this change would otherwise make far more visible: the consistency bound used to come from an unqualified `GetCheckpoint`, which can run ahead of the list-index pipelines the List APIs depend on.

**Cursor values change.** The removed Postgres path minted cursors with `checkpoint` hardcoded to `0` (it tracked position purely by `tx_sequence_number`/`ev_sequence_number`, never a checkpoint). The gRPC-only path now always encodes the real checkpoint each transaction/event was found in, so cursors for the same items serialize to different bytes than before (same `tx_sequence_number`/`ev_sequence_number`, non-zero `checkpoint`) — this is the source of every cursor diff in the updated e2e snapshots. Previously-issued cursors (with `checkpoint == 0`) still work: the gRPC scan already special-cases `checkpoint == 0` `before` cursors (substituting `u64::MAX` so the checkpoint window doesn't collapse to empty), since it had to interoperate with Pg-minted (and legacy JSON) cursors back when the gRPC path was opt-in.

## Test plan

Updated e2e test infra (`FullCluster`/`cluster_with_pipelines` call sites) to serve the List APIs by default where `transactions`/`events` connections are exercised. Ran the full `sui-indexer-alt-e2e-tests` suite (453 tests) and the transactional `graphql/**` suite (170 tests) — all pass deterministically. Existing snapshots were updated for two distinct reasons: cursor bytes changed as described above (same items/order, different `checkpoint` encoding), and the watermark fix corrected some previously-flaky (incomplete) recorded results.

## Release notes

Check each box that your changes affect. If none apply, leave all unchecked.
- [ ] Protocol: 
- [ ] Indexing Framework:
- [x] JSON-RPC: `--enable-list-apis` (and its alias `--experimental-query-apis`) is removed; the ledger gRPC List APIs are always used once `--ledger-grpc-url` is configured.
- [x] GraphQL: Same as above — `transactions`/`events` connections are served entirely by the ledger gRPC List APIs, with no Postgres fallback. Cursors returned by these connections now encode the real checkpoint instead of always `0`; previously-issued cursors remain valid.
- [ ] Rust SDK:


crates/sui-indexer-alt-graphql/src/api/types/transaction/mod.rs
```diff
@@ -335,54 +322,9 @@ impl Transaction {
...
-
-        page.paginate_results(
-            tx_digests(ctx, &tx_sequence_numbers).await?,
-            |(s, _)| TransactionToken::cursor(0, *s),
-            |(_, d)| Ok(Self::with_digest(scope.clone(), d)),
-        )
-        .map(Into::into)
+        query_limits::rich::debit(ctx)?;
+        let reader: &AlphaLedgerGrpcReader = ctx.data()?;
+        Self::paginate_grpc(reader, scope, page, filter).await
```
wlmyng, line 327: we should just inline `paginate_grpc` at this point

crates/sui-indexer-alt-graphql/src/api/types/event/mod.rs
```diff
@@ -191,66 +183,8 @@ impl Event {
...
-        .await?;
-
-        page.paginate_results(
-            events,
-            |(c, _)| EventToken::cursor(0, c.tx_sequence_number, c.ev_sequence_number),
-            |(_, e)| Ok(e),
-        )
-        .map(Into::into)
+        let reader: &AlphaLedgerGrpcReader = ctx.data()?;
+        Self::paginate_grpc(reader, scope, page, filter).await
```
wlmyng, line 187: same here lets inline

wlmyng (commented):
On the cursor change: yes it's true, but pagination on cursor should be able to proceed as normal, and the cursor internals should be opaque

Watermarking on ledger grpc: not an issue against archival as that is unpruned, but if against fullnode grpc, is that uniformly pruned or do different data sources follow different pruning retention?

amnn (approved):
This change looks good to go, modulo addressing @wlmyng's comments.

PR #27669, opened 2026-08-12
[graphql-stream-metrics] Add metrics for the subscription upstream checkpoint stream [21/n]

tpham-mysten:
## Description

The subscription upstream checkpoint stream had no metrics. Adds `SubscriptionMetrics` (prefix `graphql_subscription_`; `phase` = `live` or `recovery`):

- `upstream_connection_failures{code}` — failed upstream connect attempts, by gRPC code
- `upstream_disconnections` — an established stream dropped and reconnected
- `upstream_terminations{reason}` — task died permanently (`connect_error` / `stream_error`); alert on any increase
- `upstream_processed_checkpoints{phase}` — checkpoints decoded + broadcast (throughput)
- `upstream_latest_processed_checkpoint{phase}` — highest sequence number decoded (position)
- `upstream_processed_checkpoint_timestamp_lag{phase}` — wall-clock lag distribution, in seconds
- `upstream_latest_processed_checkpoint_timestamp_ms{phase}` — raw latest timestamp; stall = `time()*1000 - value` (keeps growing when frozen)

Recorded via one `ProcessedCheckpointMetricReporter::report(phase, seq, ts)` on both the live (`index_and_broadcast`) and recovery (`recover_gap`) paths, so recovery is visible under `{phase="recovery"}` without disturbing the live series.

Scope: upstream stream only. Per-subscriber metrics and methods (active subscriptions, delivery/throttle, the gap-recovery retry counter) follow in a subsequent PR.

## Test plan

Unit test covers the connect-failure gRPC-code labeling; existing gap-recovery and streaming tests still pass.

## Release notes

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/metrics/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/metrics/mod.rs
@@ -17,7 +17,6 @@ use prometheus::register_int_gauge_with_registry;
 
 mod subscription;
 
-pub use subscription::ProcessedCheckpointMetricReporter;
 pub use subscription::SubscriptionMetrics;
 
 /// Histogram buckets for the distribution of latency (time between receiving a request and sending
@@ -75,8 +74,6 @@ pub struct RpcMetrics {
     pub fields_received: IntCounterVec,
     pub fields_succeeded: IntCounterVec,
     pub fields_failed: IntCounterVec,
-
-    pub subscription: Arc<SubscriptionMetrics>,
 }
 
 impl RpcMetrics {
@@ -275,8 +272,6 @@ impl RpcMetrics {
                 registry,
             )
             .unwrap(),
-
-            subscription: Arc::new(SubscriptionMetrics::new(registry)),
         })
     }
 }
--- a/crates/sui-indexer-alt-graphql/src/metrics/mod.rs
+++ b/crates/sui-indexer-alt-graphql/src/metrics/mod.rs
@@ -15,6 +15,11 @@ use prometheus::register_int_counter_with_registry;
 use prometheus::register_int_gauge_vec_with_registry;
 use prometheus::register_int_gauge_with_registry;
 
+mod subscription;
+
+pub use subscription::ProcessedCheckpointMetricReporter;
+pub use subscription::SubscriptionMetrics;
+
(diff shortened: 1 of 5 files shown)


crates/sui-indexer-alt-graphql/src/metrics/mod.rs
```diff
@@ -70,6 +75,8 @@ pub struct RpcMetrics {
     pub fields_received: IntCounterVec,
     pub fields_succeeded: IntCounterVec,
     pub fields_failed: IntCounterVec,
+
+    pub subscription: Arc<SubscriptionMetrics>,
```
amnn: Nesting metrics like this is not a common pattern -- is there a reason to do it like this instead of including all the metrics inline here? If there is one, does it imply we should be factoring out more of these metrics into sub-sections?

amnn (approved):
Metrics themselves look good, but I'd consider looking again at the abstractions -- I don't see a strong motivation to introduce the two new types (subscription-specific metrics and the reporter), instead of keeping everything in `RpcMetrics` and putting the methods on there too:

- We typically don't introduce nested metric types, instead relying on sharing through `Arc`s to make it cheap to get metrics to where they need to be; or if we can split metrics up, then we split them up completely (initialise them separately and pass them in differently at the top).
- The reasons we've had in the past to introduce some kind of reporter type is usually that they either hold state, or they parametrize labels, neither of which is happening here.

I would suggest one of the following options, but it's not blocking and I'll leave it with you to decide:

- Collapse everything into `RpcMetrics`.
- Keep `SubscriptionMetrics` around, but fold the reporter type into it, and remove it as a field from `RpcMetrics` -- it's just a separate metric family that gets initialised and threaded in during service initialisation.

tpham-mysten:
Thanks @amnn! I'm going with option 2. I think Subscriptions are a large enough subsystem, with a growing set of metrics plus a few helper methods, that a dedicated SubscriptionMetrics object reads cleaner than folding everything flat into RpcMetrics.

amnn:
Makes sense to me too, after seeing the rest of your stack!

PR #27676, opened 2026-08-12
sui-kv-rpc: add a second, plaintext gRPC listener

evan-wall-mysten:
## Description

sui-kv-rpc's TLS is all-or-nothing today: one listener, either plaintext or TLS depending on whether `tls-cert`/`tls-key` are set. Trusted internal callers (e.g. sui-indexer-alt-jsonrpc reading over gRPC) shouldn't need to negotiate TLS just because the primary listener is TLS-secured for external traffic. Adds a `plaintext-address` config option that binds a second, unencrypted listener serving the same `LedgerService`, independent of the primary listener's TLS setting — so sui-kv-rpc can be run http-only, https-only, or both.

## Test plan

New end-to-end tests in `sui-indexer-alt-e2e-tests` connect a gRPC test client directly to the second listener: one confirms `GetServiceInfo` succeeds once the listener's cache warms up, the other executes and checkpoints a real transaction and confirms `GetTransaction` reads it back over that listener. New `config.rs` unit tests cover the default/parse/validate behavior of `plaintext-address`.

---

## Release notes

- [x] gRPC: `sui-kv-rpc` can now bind a second, unencrypted gRPC listener (`plaintext-address` config field) alongside its existing TLS listener, for trusted internal callers that shouldn't need to negotiate TLS.
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-e2e-tests/src/lib.rs
+++ b/crates/sui-indexer-alt-e2e-tests/src/lib.rs
@@ -126,6 +126,10 @@ pub struct OffchainCluster {
     /// The address the kv-rpc (LedgerService) server is listening on.
     kv_rpc_listen_address: SocketAddr,
 
+    /// The address kv-rpc's second, unencrypted listener is listening on, when
+    /// `OffchainClusterConfig::kv_rpc_plaintext_listener` is set.
+    kv_rpc_plaintext_listen_address: Option<SocketAddr>,
+
     /// Read access to BigTable.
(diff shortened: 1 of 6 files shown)


jessiemongeon1:
## Style Guide Audit

chris-gorham (approved):

PR #27685, opened 2026-08-13
graphql: add root balance lookups

amnn:
## Description 

Introduce `Query.multiGetBalances`, this allows us to fetch multiple balances, owned by different addresses in a single query, without having to dynamically generate a GraphQL query (we can create a single query, parametrized by variables).

I also noticed a couple of small issues with the comments related to balances, so I  cleaned those up too.

## Test plan 

Updated E2E tests:

```
$ cargo nextest run -p sui-indexer-alt-e2e-tests -- graphql/balances
```

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [x] GraphQL: Adds `Query.multiGetBalances` for fetching balances across multiple addresses in one query.
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/query.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/query.rs
@@ -29,6 +29,9 @@ use crate::api::scalars::uint53::UInt53;
 use crate::api::types::address;
 use crate::api::types::address::Address;
 use crate::api::types::address::AddressKey;
+use crate::api::types::balance;
+use crate::api::types::balance::Balance;
+use crate::api::types::balance::BalanceKey;
 use crate::api::types::checkpoint;
 use crate::api::types::checkpoint::CCheckpoint;
 use crate::api::types::checkpoint::Checkpoint;
@@ -334,11 +337,32 @@ impl Query {
         keys: Vec<AddressKey>,
     ) -> Result<Vec<Option<Address>>, RpcError<address::Error>> {
         let scope = self.scope(ctx)?;
-        try_join_all(
-            keys.into_iter()
-                .map(|k| Address::by_key(ctx, scope.clone(), k)),
-        )
-        .await
+        let addresses = keys
+            .into_iter()
+            .map(|k| Address::by_key(ctx, scope.clone(), k));
+
+        try_join_all(addresses).await
+    }
+
+    /// Fetch balances by their addresses and coin types.
+    ///
+    /// Each result includes the total balance, the balance held in coin objects, and the balance held in the address's balance accumulator. Returns a list that is guaranteed to be the same length as `keys`. If an address has no balance of a given type, all three values are zero for that key.
+    async fn multi_get_balances(
+        &self,
+        ctx: &Context<'_>,
+        keys: Vec<BalanceKey>,
+    ) -> Result<Vec<Balance>, RpcError<balance::Error>> {
+        let scope = self.scope(ctx)?;
+        let keys = keys
+            .into_iter()
+            .map(|key| (key.address.into(), key.coin_type.into()))
+            .collect();
+
+        // Query is only exposed at the request root or through Checkpoint.query, both of which
+        // set a checkpoint bound.
+        Ok(Balance::fetch_many(ctx, &scope, keys)
+            .await?
+            .context("Query scope is missing a checkpoint bound")?)
     }
 
     /// Fetch checkpoints by their sequence numbers.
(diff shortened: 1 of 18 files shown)


crates/sui-indexer-alt-graphql/src/api/query.rs
```diff
@@ -341,6 +344,24 @@ impl Query {
...
+        keys: Vec<BalanceKey>,
+    ) -> Result<Vec<Balance>, RpcError<balance::Error>> {
+        let scope = self.scope(ctx)?;
+        let balances =
+            try_join_all(keys.into_iter().map(|key| {
+                Balance::fetch_one(ctx, &scope, key.address.into(), key.coin_type.into())
+            }))
+            .await?;
+
+        Ok(balances.into_iter().flatten().collect())
```
tpham-mysten: flatten() drops None, and Balance::fetch_one can return None. So the documented "same length as `keys`" guarantee might not be hold?

crates/sui-indexer-alt-graphql/src/api/types/address.rs
```diff
@@ -88,7 +88,7 @@ pub(crate) enum AddressTransactionRelationship {
         name = "balance",
         arg(name = "coin_type", ty = "TypeInput"),
         ty = "Option<Result<Balance, RpcError<balance::Error>>>",
-        desc = "Fetch the total balance for coins with marker type `coinType` (e.g. `0x2::sui::SUI`), owned by this address.\n\nIf the address does not own any coins of that type, a balance of zero is returned.",
+        desc = "Fetch the balance for `coinType` (e.g. `0x2::sui::SUI`) owned by this address.\n\nThe result includes the total balance, the balance held in coin objects, and the balance held in the address's balance accumulator. If this address has no balance of that type, all three values are zero.",
```
tpham-mysten: nit: This doc only includes "all three values are zero", while `Address::balance` (line 289) includes the null when no checkpoint is set in scope". Look like a miss here?

crates/sui-indexer-alt-graphql/src/api/query.rs
```diff
@@ -341,6 +344,24 @@ impl Query {
...
+    ///
+    /// Each result includes the total balance, the balance held in coin objects, and the balance held in the address's balance accumulator. Returns a list that is guaranteed to be the same length as `keys`. If an address has no balance of a given type, all three values are zero for that key.
+    async fn multi_get_balances(
+        &self,
+        ctx: &Context<'_>,
+        keys: Vec<BalanceKey>,
+    ) -> Result<Vec<Balance>, RpcError<balance::Error>> {
+        let scope = self.scope(ctx)?;
+        let balances =
+            try_join_all(keys.into_iter().map(|key| {
```
tpham-mysten: optional nit: We can have a new line for try_join_all, which I think is cleaner and consistent with other multi_get methods

tpham-mysten (approved):
Looking good. Just a few nits and a question for `flatten()`

wlmyng (approved):

crates/sui-indexer-alt-graphql/src/api/query.rs
amnn: `Balance::fetch_one` only produces `None` if the `Scope` doesn't have a `checkpoint_bound` which should never happen under normal operation, but you're right that it would be clearer to check for that and produce a server error if we were in that situation.

Also each call to `Balance::fetch_one` counts as a rich query, which severely limits how many of those we can perform, so the `try_join_all` pattern was not a good fit here either.

For both of those reasons, I ended up implementing a broader refactor to make `Balance::fetch_many` support batching lookups across different addresses (the consistent store API already works this way).

This allows us to share the same underlying implementation across `IAddressable.multiGetBalances` and `Query.multiGetBalances`. That implementation produces a `Result<Option<Vec<Balance>>, ...>` so we can neatly treat `Ok(None)` as an internal error in the `Query` case or transpose it in the `IAddressable` case.

crates/sui-indexer-alt-graphql/src/api/types/address.rs
amnn: Updated, thanks!

crates/sui-indexer-alt-graphql/src/api/query.rs
amnn: Done! Also made the same change for `Query.multiGetAddresses`.

PR #27690, opened 2026-08-13
Tweak bigtable write controller

nickvikeras:
## Description

Last night I got paged because the mainnet json-rpc fallback bigtable cluster was lagging. A brief, large but short-lived latency spike caused the write controller to throttle writes down to 0.1 QPS and the controller got stuck there and I had to restart the indexers manually to recover.

The controller watches flow control feedback from bigtable that tells it by what factor it can increase its write QPS, and how long to wait after increasing it before increasing it _again_. The controller also tracks its baseline observed write latency and watches for large latency spikes as an additional guard, and won't increase write QPS if it detects latency has spiked. I added the latency guard because testing showed that operating on the bigtable feedback alone was causing dramatic high percentile _read_ latency spikes while autoscaling kicked in. This guard stabilized that. 

This PR addresses two failure modes thats I think caused the incident.

1. If the controller sees no writes for one `OBSERVATION_WINDOW` period (1s) it resets its local state variables to avoid applying stale feedback when no traffic is flowing through the system. If the controller has throttled down such that the QPS is less than 1.0, then we would always discard the growth feedback from the server before applying it (because it takes longer than 1s to dispatch a write). The fix is to wait until enough wall clock time has passed for at least one write to have had an opportunity to flow through the controller based on the current allowed QPS. This ensures the state is only reset when the system is truly idle, not throttled.

2. Previously if QPS dropped to the minimum of 0.1, it would take 75 minutes for the baseline to adjust because it evaluated 5 rpc samples at a time, and would wait for 90 evaluations of "elevated" latency before beginning to increase the baseline. I changed the code to instead start moving the baseline half-way towards the current observed latency over 10 minutes of wall clock time.

Here you can see the spike that caused the problem as well as the write throttling

image

image

## Test plan

I added simulator-style tests that reproduced the failure modes, then modified the controller to make the tests pass.


crates/sui-kvstore/src/bigtable/client/flow_control.rs
```diff
@@ -347,7 +368,14 @@ impl BatchWriteFlowController {
             return None;
         }
         if observation.rpc_starts == 0 {
-            state.restart_or_close_empty_observation(now);
+            // A low paced rate can make the fixed window empty even under sustained demand.
+            // Growth-only feedback remains valid through the next permit interval and one
+            // observation boundary. Conservative feedback restarts without including idle time.
+            if state.pending.growth_factor().is_none()
+                || elapsed >= state.growth_observation_timeout()
```
tpham-mysten, line 391: nit: Not sure if 

```
now >= state.next_permit_at + OBSERVATION_WINDOW
```

is cleaner? and we dont have to add a new method for growth_observation_timeout()

crates/sui-kvstore/src/bigtable/client/flow_control.rs
```diff
@@ -587,28 +616,30 @@ impl BatchWriteFlowController {
...
 
         let window_avg_micros = state.latency_total_micros as f64 / state.latency_samples as f64;
+        let evidence_duration = now
+            .saturating_duration_since(
+                state
+                    .latency_sampling_started_at
+                    .take()
+                    .expect("latency samples missing sampling start"),
+            )
+            .clamp(LATENCY_EVALUATION_PERIOD, state.baseline_evidence_cap());
```
wlmyng, line 669: this interaction between clamp and baseline_evidence_cap is so that idle time doesn't count as evidence?

crates/sui-kvstore/src/bigtable/client/flow_control.rs
```diff
@@ -347,7 +368,14 @@ impl BatchWriteFlowController {
...
         if observation.rpc_starts == 0 {
-            state.restart_or_close_empty_observation(now);
+            // A low paced rate can make the fixed window empty even under sustained demand.
+            // Growth-only feedback remains valid through the next permit interval and one
+            // observation boundary. Conservative feedback restarts without including idle time.
+            if state.pending.growth_factor().is_none()
+                || elapsed >= state.growth_observation_timeout()
+            {
+                state.restart_or_close_empty_observation(now);
+            }
```
wlmyng, line 394: relaxes so that growth-only observations aren't discarded

crates/sui-kvstore/src/bigtable/client/flow_control.rs
```diff
@@ -587,28 +616,30 @@ impl BatchWriteFlowController {
...
         state.next_latency_evaluation_at = now + LATENCY_EVALUATION_PERIOD;
 
         let condition =
             classify_write_latency(window_avg_micros, state.baseline_write_latency_micros);
         let latency_feedback = match condition {
-            WriteLatencyCondition::Severe => {
-                state.non_healthy_evals = state.non_healthy_evals.saturating_add(1);
-                Some(LatencyFeedback::Decrease)
-            }
+            WriteLatencyCondition::Severe => Some(LatencyFeedback::Decrease),
```
tpham-mysten, line 685: IIUC, `Severe` now contributes nothing to the baseline (it used to increment non_healthy_evals, so time above 3x counted toward unlocking Elevated adaptation). Doesn't that leave latency settling above 3x permanently unrecoverable, with every eval Severe, QPS pinned at MIN_QPS, and nothing left to move the baseline?

tpham-mysten (commented):
Thanks @nickvikeras for going through the PR over the call.

Looking great on the solution for problem 1, but I have a question for the second.

wlmyng (approved):
iiuc, the two changes come together so that flow control adjusts to elevated latency over time rather than latching permanently onto the initial "normal"

crates/sui-kvstore/src/bigtable/client/flow_control.rs
nickvikeras: I think previously severe counted as "non_healthy_evals" but it would block updating the baseline, so it would have been stuck with the old baseline forever with the old code too.

But we should definitely fix this. 

My intent here is for the bigtable RateLimitInfo messages to be the main source of truth while the latency is kind of a soft guard to slow down reacting to it. So if the "severe" threshold becomes the new baseline, I think we should just accept that in the controller and keep responding to feedback from bigtable.

nickvikeras: Fixed. Now the latency baseline should always adapt, and can only slow, but not stop the allowed QPS from increasing.

crates/sui-kvstore/src/bigtable/client/flow_control.rs
nickvikeras: I think that would work in the current call path because we never cancel the future after reserving the permit, but if someone added code to cancel futures while they are waiting for the permit (i.e. wrapping them in a timeout) we could get into a weird state where RPCs don't actually dispatch, but `state.next_permit_at` keeps advancing, so we end up keeping the feedback for far longer than intended.

crates/sui-kvstore/src/bigtable/client/flow_control.rs
tpham-mysten: SG. Thanks

PR #27694, opened 2026-08-13
[chore] try to fix windows build failing by switching to a beefier runner

stefan-mysten:
## Description 

In many recent PRs, the windows build fails with various errors. It's really unclear what's wrong from a first glance, so the first step is to try to bump the runner to `windows-ghcloud` to take advantage of the larger SSD and RAM. Not sure this fixes the build though, so we will have to see.

I switched it previously from `windows-ghcloud` to `windows-ghcloud-128` because the former was always not having runners and CI was often stuck at it.

## Test plan 

Existing CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -355,7 +355,7 @@ jobs:
     needs: [diff, rustfmt, clippy, snapshot-check]
     if: needs.diff.outputs.isRust == 'true'
     timeout-minutes: 45
-    runs-on: [ windows-ghcloud-128 ]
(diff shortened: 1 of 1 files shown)


stefan-mysten:
Uhm, still that runner is mostly blocked...

ebmifa (approved):

PR #27695, opened 2026-08-13
[consensus] slim block wire envelope behind a protocol flag

arun-koshy:
## Description

Prepares the subscription wire for slim (ancestor-compressed) blocks without changing today's bytes, gated by a new protocol flag `consensus_slim_block_propagation` — **off on every chain**.

The `block` field stays the only payload field. Flag off, it carries raw serialized `SignedBlock` bytes, **byte-identical to the current wire** — verified by an encoding test against the legacy message shape across present/empty field combinations. Flag on, the bytes hold a `SerializedBlockEnvelope` whose oneof distinguishes `Full` from `Slim`.

The envelope rides *inside* the bytes rather than as a message-level oneof so the same framing can later serve the observer batch stream, where blocks are a `repeated` field and protobuf permits no repeated oneof. Only the live subscription stream is framed: fetch, commit sync, latest-block, and the test-only unicast stay raw full blocks — they are the recovery paths slim reconstruction falls back to.

Both ends of a stream interpret the bytes by their own protocol config, shared through the epoch's protocol version; consensus restarts per epoch and old binaries cannot run a version that enables the flag. A payload that does not frame as an envelope is a **framing fault, not an invalid block**: the stream ends without resetting the subscriber's reconnect backoff, so a mismatched or hostile peer escalates through backoff instead of being rejected block by block indefinitely. Counted per peer and reason in `subscribe_stream_form_failures`.

Internally the serialized payload is one enum: `ExtendedSerializedBlock` carries `SerializedBlockForm` instead of parallel full/optional-slim fields, so both-or-neither states are unrepresentable. Nothing emits the slim form yet — the encoder lands with the codec — so a slim payload can only be a misbehaving peer: the subscriber drops and counts it, the authority service rejects it with `UnexpectedBlockForm`, and a slim form produced server-side while the flag is off fails the stream rather than fabricating a payload.

## Test plan

Byte-compat matrix pinning flag-off output to the legacy encoding (four present/empty field combinations, legacy field order); flag-on round trips for both forms including empty payloads; malformed envelopes (garbage, empty envelope, raw BCS bytes — which always fail, since a `SignedBlock`'s first byte reads as protobuf field 0); both config-mismatch directions, documented as unsupported and observable; a framing test driven through a real `ConsensusProtocolConfig` via the testing setter; slim-drop at the subscriber and `UnexpectedBlockForm` at the service. Byte-compat and slim-drop are mutation-verified.

No protocol-config snapshot changes (the flag is false and serialization skips false flags); OpenRPC snapshot regenerated. `MAX_PROTOCOL_VERSION` untouched — the version bump belongs to the PR that enables the flag.

Plan reviewed and implementation re-reviewed adversarially before push; both fixes from the second review are included.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


consensus/core/src/network/tonic_network.rs
```diff
@@ -142,18 +142,53 @@ impl ValidatorNetworkClient for TonicValidatorClient {
...
-            .filter_map(move |b| async move {
-                match b {
-                    Ok(response) => Some(ExtendedSerializedBlock {
-                        block: response.block,
-                        excluded_ancestors: response.excluded_ancestors,
-                    }),
-                    Err(e) => {
-                        debug!("Network error received from {}: {e:?}", peer);
-                        None
+            .filter_map(move |b| {
```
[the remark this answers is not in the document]
arun-koshy: Good catch — the reject path did bypass the throttle. Rather than reordering, I've dropped the rejection entirely: an unusable form now falls through as an empty block and is rejected downstream, which is both the pre-existing behaviour and inherently rate-limited. The loud rejection belongs with the decoder and its protocol flag in the follow-up.

consensus/core/src/network/tonic_network.rs
```diff
@@ -1374,11 +1401,25 @@ pub(crate) struct SubscribeBlocksRequest {
 
 #[derive(Clone, prost::Message)]
 pub(crate) struct SubscribeBlocksResponse {
-    #[prost(bytes = "bytes", tag = "1")]
-    block: Bytes,
```
mwtian, line 1378: Deleting fields in protobuf message is an incompatible change. We should use a `ConsensusProtocolConfig` flag (slim_block for example) to control the format here. Also, is it necessary to support both full and slim forms?

consensus/core/src/network/mod.rs
```diff
@@ -374,6 +374,9 @@ pub(crate) struct ExtendedSerializedBlock {
     pub(crate) block: Bytes,
     // Serialized BlockRefs that are excluded from the blocks ancestors.
     pub(crate) excluded_ancestors: Vec<Vec<u8>>,
+    /// The same block with its ancestor digests stripped, when the sender emitted
+    /// that form instead. Exactly one of `block` and `minimal` is populated.
+    pub(crate) minimal: Option<Bytes>,
```
mwtian: I guess sharing the `block` field with the full form will be too much churn? Is the plan to retire one of the fields eventually? And move it below the `block` field?

consensus/core/src/network/tonic_network.rs
```diff
@@ -147,10 +147,17 @@ impl ValidatorNetworkClient for TonicValidatorClient {
             .take_while(|b| futures::future::ready(b.is_ok()))
             .filter_map(move |b| async move {
                 match b {
-                    Ok(response) => Some(ExtendedSerializedBlock {
-                        block: response.block,
-                        excluded_ancestors: response.excluded_ancestors,
-                    }),
+                    Ok(response) => {
+                        let block = match response.form {
```
mwtian: Potentially, the serialized bytes will keep using the `block` field, and downstream will decide on how to parse it.

mwtian (commented):
The change should be backward compatible with existing wire format before rollout. One possibility is to reuse existing serialized block fields, but that would require using a protocol config flag. If a new field in the wire protocol is used, potentially no new protocol config flag would be needed. But there might be more churns on the usage side. What would be the cleaned path here?

consensus/core/src/network/tonic_network.rs
arun-koshy: On compatibility. The field leaves the Rust struct but not the wire, BlockForm::Full keeps tag 1, so a full block encodes exactly as block did.

Old readers decode new full-form messages, and new readers decode old ones. Two edges worth naming: an empty payload differs (Some(Full(empty)) writes 0a 00 where proto3 omits the field — doesn't arise for real blocks), and if a sender puts both tags in one message, an old reader takes tag 1 while a new one takes the last, so they disagree. But neither should be an issue.

The case that actually matters for rollout is an old binary receiving tag 3: it skips the unknown field, sees an empty block, and drops it. So compatibility rests on never emitting minimal blocks while pre-tag-3 binaries participate.

I had a test asserting the byte-identity above and dropped it as low-value before opening this. Happy to add it back.

On the flag. It exists as consensus_minimal_block_propagation, but it lives in the stacked #27702 alongside the encoder that can actually produce a minimal block. Here the minimal arm is unreachable because nothing populates it. If you'd rather it be unreachable by construction than by data, I can add the flag check at the emission site in this PR.

On whether both forms are needed. Both are needed by the design as it stands. In the prototype design a subscriber past the GC horizon can't rebuild a minimal block. Reconstruction is all-or-nothing across ancestors, so a small per-ancestor miss rate compounds. Those subscribers need full blocks until they catch up, and catch-up replay is always full for the same reason. The gate deciding this re-evaluates mid-stream, since a node that falls behind after subscribing would otherwise keep receiving blocks it can't use.

Mid-stream switching is what forces a per-message discriminator. The alternative is to fix the form per subscription — negotiate it in SubscribeBlocksRequest, or end the subscription and let the peer reconnect when the gate flips. That keeps one field and a smaller wire surface. I didn't take it because a resubscribe on every transition seemed heavier than a tag.

consensus/core/src/network/mod.rs
arun-koshy: By the time we're in this struct the decision is settled and the unused form is dead weight. So collapsing them is viable.

`enum SerializedBlockForm { Full(Bytes), Minimal(Bytes) }`

On retiring: the full form itself isn't going away based on current design, lagging subscribers and catch-up replay need it indefinitely, but these two parallel fields can become one. 

Before I make the change to the enum, just wanted to confirm this is what you were thinking as well?

consensus/core/src/network/tonic_network.rs
arun-koshy: The forms mix within a single subscription: catch-up replay is always full, live traffic may be minimal, and the gate can switch a stream between them mid-flight. So the protocol flag tells a receiver that minimal blocks are possible, not which form the message in front of it is.

Let me know if you don't agree with the proposed design of switching forms mid stream based on lag gate (draft [PR#27740](https://github.com/MystenLabs/sui/pull/27740))

akichidis (approved):
We've discussed offline about this. Unless we go for the alternative envelope approach, I am approving this as is.

consensus/core/src/network/mod.rs
arun-koshy: Done — they now share one field. ExtendedSerializedBlock carries a single block: SerializedBlockForm (Full | Slim) instead of the parallel full/optional-slim fields, so both-populated or neither is unrepresentable. Nothing left to retire or reorder.

consensus/core/src/network/tonic_network.rs
arun-koshy: This is exactly what we ended up doing. The bytes stay in block, and the receiver decides how to parse them by its own protocol config: flag off → raw block bytes, flag on → decode the envelope and match the form. No trial-parsing needed — the envelope's oneof tag is the discriminator. A payload that doesn't frame as an envelope ends the stream (counted per peer/reason) rather than being rejected block by block. Bonus of the envelope-in-bytes shape: it also works for the observer batch stream later, where blocks is repeated and protobuf has no repeated oneof.

consensus/core/src/network/tonic_network.rs
arun-koshy: Reworked along these lines — the field is no longer deleted, and the format is flag-controlled as you suggested. block stays the only wire field: with consensus_slim_block_propagation off (every chain), it carries raw SignedBlock bytes, and a test pins the flag-off framing to the identity, so deploying this is a wire no-op. With the flag on, the bytes hold a SerializedBlockEnvelope whose oneof distinguishes full from slim. Went with "slim" for the name too.

mwtian (approved):

consensus/core/src/network/mod.rs
```diff
@@ -371,15 +372,47 @@ pub(crate) use clients::{CommitSyncerClient, SynchronizerClient};
...
+    #[prost(bytes = "bytes", tag = "2")]
+    Slim(Bytes),
+}
+
+impl SerializedBlockEnvelope {
+    pub(crate) fn encode_form(form: SerializedBlockForm) -> Bytes {
+        Self { block: Some(form) }.encode_to_vec().into()
+    }
+
+    pub(crate) fn decode_form(bytes: &[u8]) -> Result<SerializedBlockForm, prost::DecodeError> {
```
mwtian: Is there a ConsensusError for this?

arun-koshy: Will add ConsensusError::MalformedBlockEnvelope

PR #27701, opened 2026-08-14
[trace-view] Added verification against onchain bytecode

awelc:
## Description 

This PR adds verification of locally built bytecode representing local sources against onchain bytecode used to generate a trace. The idea is to give users a signal in case bytecode in these two places does not match or it matching or not cannot be determined. This signal is represented by a warning so that users can proceed with debugging even in case bytecode does not match. The user experience in such case may be less ideal but the debugger has been [hardened](https://github.com/MystenLabs/sui/pull/26870) to handle this situation without crashing

## Test plan 

Tested various scenarios manually (no Sui CLI binary, old version of Sui CLI binary, matching bytecode, non-matching bytecode)

### DIFF
--- a/docs/content/references/ide/debugger.mdx
+++ b/docs/content/references/ide/debugger.mdx
@@ -152,7 +152,7 @@ Debugging a Move unit test is a two-step process:
       </div>
 
       :::info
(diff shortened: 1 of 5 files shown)


jessiemongeon1:
## Style Guide Audit

tzakian (approved):
Very cool!

PR #27702, opened 2026-08-14
[consensus] slim block codec

arun-koshy:
## Description

Strips a block's ancestor digests out of the subscription payload and rebuilds them at the receiver. Emission is gated by `consensus_slim_block_propagation`, off on every chain, so nothing changes until it is enabled.

A block's ancestors are 32 bytes each, and a receiver at the frontier already holds those blocks. The slim form carries ancestor authors in proposal order plus an override for the ones the receiver cannot infer: a non-parent round, a slot it cannot resolve uniquely, or an ancestor below the GC horizon. Everything else is looked up locally.

**Reconstruction is self-validating.** The rebuilt bytes must hash to the digest the author claimed before the block goes anywhere, and the signature is checked afterwards on the normal path — so a wrong or hostile digest costs a fetch, never a bad block. The rebuild must reproduce the author's *exact* serialization, since that is what the signature covers.

The encoding is produced once at proposal time, where every ancestor is already accepted, so the whole subscription fan-out shares it. Catch-up replay and the liveness re-broadcast stay on the full form.

Both directions scope their `DagState` guard to the part that needs it: only ancestor resolution reads local state, so cloning, serializing, and hashing the block happen outside the lock that accept and commit contend for.

A receiver that cannot rebuild a block drops it rather than passing it on to be rejected as a peer fault, since the usual cause is an ancestor it has not accepted yet. Failures are counted by peer and reason, and logged by kind: a malformed payload is the peer's fault and stays at info, an unresolvable ancestor is local and drops to debug.

### Not in this PR, both required before the flag can be enabled

- **No per-subscriber gate.** The flag gates emission globally, but nothing here decides full-vs-slim per peer: a subscriber trailing the horizon cannot rebuild anything and must be served full blocks. Observable rather than theoretical — setting the flag in `ConsensusProtocolConfig::for_testing` hangs the `authority_node` committee tests.
- **No recovery** for a block whose ancestors have not arrived yet; it is dropped and refetched by ordinary block sync.

Stacked on #27695 — review that first; this targets its branch.

## Test plan

Fourteen tests on the codec and resolver. The load-bearing ones: byte-identical round trip across all block versions; ancestors carried in proposal order rather than authority order (a decoder that sorted would pass every other test); the GC horizon boundary, where an ancestor sitting exactly on it keeps its digest; a unique-but-wrong local block caught as a digest mismatch; and twelve malformed inputs rejected without panicking.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/slim_block.rs
+++ b/consensus/core/src/slim_block.rs
@@ -25,7 +25,7 @@
 use std::sync::Arc;
 
 use bytes::Bytes;
-use consensus_config::{AuthorityIndex, Committee};
+use consensus_config::{AuthorityIndex, Committee, DIGEST_LENGTH};
 use consensus_types::block::{BlockDigest, BlockRef, Round};
 use parking_lot::RwLock;
 use prost::Message as _;
@@ -53,7 +53,7 @@ pub(crate) trait AncestorDigestResolver {
 struct SlimBlock {
     /// bcs(Block) built with ancestors = [].
     #[prost(bytes = "bytes", tag = "1")]
-    block_sans_ancestors: Bytes,
+    block: Bytes,
     /// Ancestor authors in exact proposal order (membership + order).
     #[prost(uint32, repeated, tag = "2")]
     ancestor_authors: Vec<u32>,
@@ -62,7 +62,7 @@ struct SlimBlock {
     overrides: Vec<AncestorOverride>,
     #[prost(bytes = "bytes", tag = "4")]
     signature: Bytes,
-    /// 32 B digest of the full serialized SignedBlock: rebuild proof + error distinction.
+    /// Digest of the full serialized SignedBlock: rebuild proof + error distinction.
     #[prost(bytes = "bytes", tag = "5")]
     claimed_block_digest: Bytes,
 }
@@ -127,37 +127,47 @@ impl DecodeError {
     }
 }
 
-/// Encodes a verified block into its slim wire form.
-///
-/// Digests are omitted only for slots the receiver can resolve safely: unique in the
-/// local DAG and either genesis or above `min_omittable_round` (the recent-cache
-/// horizon). Everything else gets an explicit digest override.
-pub(crate) fn serialize_slim(
+/// The explicit digest to put on the wire for `ancestor`, or `None` when the receiver is
+/// guaranteed to resolve the slot to this exact digest on its own.
+fn digest_override(
+    block: &VerifiedBlock,
+    ancestor: &BlockRef,
+    resolver: &impl AncestorDigestResolver,
+    min_omittable_round: Round,
+) -> Option<Bytes> {
+    let explicit = || Some(Bytes::copy_from_slice(&ancestor.digest.0));
+
+    // The author's own parent always carries an explicit digest. It is the only
+    // ancestor edge that chains within one author: a missing parent leaves every later
+    // block from that author undecodable too, since each needs the one before it.
+    // Explicit here, the block still decodes and reaches `block_manager`, whose
+    // suspension drives the normal missing-ancestor sync.
+    if ancestor.author == block.author() {
+        return explicit();
+    }
+    // Below the horizon the receiver may have collected the slot already; genesis is
+    // exempt because it resolves from the committee rather than the DAG.
+    if ancestor.round != GENESIS_ROUND && ancestor.round <= min_omittable_round {
+        return explicit();
+    }
+    // Safe to omit only where local state resolves the slot uniquely to this digest.
+    match resolver.digest_at_slot(Slot::from(*ancestor)) {
+        SlotDigest::Unique(digest) if digest == ancestor.digest => None,
+        _ => explicit(),
+    }
+}
+
+/// The override list for `block`, the only part of encoding that reads local state.
+fn ancestor_overrides(
     block: &VerifiedBlock,
     resolver: &impl AncestorDigestResolver,
     min_omittable_round: Round,
-) -> Result<Bytes, bcs::Error> {
-    let default_round = block.round().saturating_sub(1);
-    let mut ancestor_authors = Vec::with_capacity(block.ancestors().len());
+) -> Vec<AncestorOverride> {
+    let parent_round = block.round().saturating_sub(1);
     let mut overrides = vec![];
     for ancestor in block.ancestors() {
-        ancestor_authors.push(ancestor.author.value() as u32);
-
-        let round = (ancestor.round != default_round).then_some(ancestor.round);
-        // The author's own parent always carries an explicit digest. It is the only
-        // ancestor edge that chains within one author: a missing parent leaves every
-        // later block from that author undecodable too, since each needs the one before
-        // it. Explicit here, the block still decodes and reaches `block_manager`, whose
-        // suspension drives the normal missing-ancestor sync.
-        let is_own_parent = ancestor.author == block.author();
-        let can_omit_digest = !is_own_parent
-            && (ancestor.round == GENESIS_ROUND || ancestor.round > min_omittable_round);
-        let uniquely_resolvable = can_omit_digest
-            && matches!(
-                resolver.digest_at_slot(Slot::from(*ancestor)),
-                SlotDigest::Unique(digest) if digest == ancestor.digest
-            );
-        let digest = (!uniquely_resolvable).then(|| Bytes::copy_from_slice(&ancestor.digest.0));
+        let round = (ancestor.round != parent_round).then_some(ancestor.round);
+        let digest = digest_override(block, ancestor, resolver, min_omittable_round);
 
         if round.is_some() || digest.is_some() {
             overrides.push(AncestorOverride {
@@ -167,10 +177,25 @@ pub(crate) fn serialize_slim(
             });
         }
     }
+    overrides
+}
+
+/// Encodes a verified block into its slim wire form: the block with an empty ancestor
+/// vector, the ancestor authors in proposal order, and `overrides` for the slots the
+/// receiver cannot infer.
+fn serialize_slim(
+    block: &VerifiedBlock,
+    overrides: Vec<AncestorOverride>,
+) -> Result<Bytes, bcs::Error> {
+    let ancestor_authors = block
+        .ancestors()
+        .iter()
+        .map(|ancestor| ancestor.author.value() as u32)
+        .collect();
 
     let stripped = (**block).clone().with_ancestors(vec![]);
     let slim = SlimBlock {
-        block_sans_ancestors: bcs::to_bytes(&stripped)?.into(),
+        block: bcs::to_bytes(&stripped)?.into(),
         ancestor_authors,
         overrides,
         signature: block.signed_block().signature().clone(),
@@ -181,10 +206,14 @@ pub(crate) fn serialize_slim(
 
 /// A structurally validated envelope: everything checkable without reading local state.
 /// Fields stay private so no caller can assemble a `SignedBlock` that skipped the digest
-/// gate in [`rebuild_slim`].
+/// gate in [`rebuild_from_slim`].
 pub(crate) struct ParsedEnvelope {
-    slim: SlimBlock,
-    skeleton: Block,
+    /// The author's block with an empty ancestor vector, decoded once here so the
+    /// resolve and rebuild phases share the work.
+    stripped_block: Block,
+    ancestor_authors: Vec<u32>,
+    overrides: Vec<AncestorOverride>,
+    signature: Bytes,
     block_ref: BlockRef,
     claimed_digest: BlockDigest,
 }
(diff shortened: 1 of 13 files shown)


consensus/core/src/network/tonic_network.rs
```diff
@@ -1511,12 +1517,50 @@ pub(crate) fn chunk_blocks(blocks: Vec<Bytes>, chunk_limit: usize) -> Vec<Vec<By
...
+
     use super::*;
     use crate::{context::Clock, metrics::initialise_metrics};
     use consensus_config::{ConsensusProtocolConfig, Parameters, local_committee_and_keys};
     use consensus_types::block::BlockDigest;
     use prometheus::Registry;
 
+    /// The seam the codec's own tests cannot cross: server-side construction, the prost
+    /// round trip, and the receiver-side mapping. Collapsing the minimal arm here leaves
+    /// the subscriber's decoder unreachable and drops every block once the flag is on.
```
akichidis: Can we make this comment a bit more human friendly?

consensus/core/src/slim_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+    /// recoverable exactly when the slot is unique.
+    fn digest_at_slot(&self, slot: Slot) -> SlotDigest;
+}
+
+/// The `Minimal` arm of the block wire envelope.
+#[derive(Clone, prost::Message)]
+struct MinimalBlock {
+    /// bcs(Block) built with ancestors = [].
+    #[prost(bytes = "bytes", tag = "1")]
+    block_sans_ancestors: Bytes,
```
akichidis: what does `sans` stands for?  I think I would just go with `block` - the whole purpose of the `MinimalBlock` is to bear the block without explicut parents anyways

consensus/core/src/slim_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+            );
+        let digest = (!uniquely_resolvable).then(|| Bytes::copy_from_slice(&ancestor.digest.0));
+
+        if round.is_some() || digest.is_some() {
+            overrides.push(AncestorOverride {
+                author: ancestor.author.value() as u32,
+                round,
+                digest,
+            });
+        }
```
akichidis: I find the method bearing some complexity on its checks (lots of negative checking and stacking) . I am proposing a small refactoring:

```suggestion
        let round = (ancestor.round != default_round).then_some(ancestor.round);
        let digest = digest_override(block, ancestor, resolver, min_omittable_round);

        if round.is_some() || digest.is_some() {
            overrides.push(AncestorOverride {
                author,
                round,
                digest,
            });
        }
        .....
   }
/// The explicit digest to put on the wire for `ancestor`, or `None` when it can be
/// omitted — that is, when the receiver is guaranteed to resolve the slot to this
/// exact digest on its own.
fn digest_override(
    block: &VerifiedBlock,
    ancestor: &BlockRef,
    resolver: &impl AncestorDigestResolver,
    min_omittable_round: Round,
) -> Option<Bytes> {
    let explicit = || Some(Bytes::copy_from_slice(&ancestor.digest.0));

    // The author's own parent always carries an explicit digest. It is the only
    // ancestor edge that chains within one author: a missing parent leaves every
    // later block from that author undecodable too, since each needs the one before
    // it. Explicit here, the block still decodes and reaches `block_manager`, whose
    // suspension drives the normal missing-ancestor sync.
    if ancestor.author == block.author() {
        return explicit();
    }
    // Below the recent-cache horizon the receiver may no longer hold the slot;
    // genesis is exempt since it resolves from the committee, not the DAG.
    if ancestor.round != GENESIS_ROUND && ancestor.round <= min_omittable_round {
        return explicit();
    }
    // Safe to omit only when local state resolves the slot uniquely to this digest.
    match resolver.digest_at_slot(Slot::from(*ancestor)) {
        SlotDigest::Unique(digest) if digest == ancestor.digest => N
        _ => explicit(),
    }
}
```

consensus/core/src/minimal_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+pub(crate) trait AncestorDigestResolver {
+    /// What `slot` resolves to, counting genesis blocks at round 0. Only the three
+    /// outcomes are needed, never the full candidate list: an omitted digest is
+    /// recoverable exactly when the slot is unique.
+    fn digest_at_slot(&self, slot: Slot) -> SlotDigest;
+}
+
+/// The `Minimal` arm of the block wire envelope.
+#[derive(Clone, prost::Message)]
+struct MinimalBlock {
```
akichidis: Agreed on the offline discussion on the `Minimal` term. Unfortunately I don't have a much better proposal.

consensus/core/src/minimal_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+        .map_err(|e| DecodeError::Malformed(format!("prost decode: {e}")))?;
+
+    let claimed_digest: [u8; 32] = minimal
+        .claimed_block_digest
+        .as_ref()
+        .try_into()
+        .map_err(|_| DecodeError::Malformed("claimed_block_digest must be 32 bytes".into()))?;
+    let claimed_digest = BlockDigest(claimed_digest);
+
+    let skeleton: Block = bcs::from_bytes(&minimal.block_sans_ancestors)
```
akichidis: why not just `block` instead of `skeleton`?

consensus/core/src/slim_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+    #[prost(uint32, repeated, tag = "2")]
+    ancestor_authors: Vec<u32>,
+    /// Only the unusual ancestors: non-default round and/or explicit digest.
+    #[prost(message, repeated, tag = "3")]
+    overrides: Vec<AncestorOverride>,
+    #[prost(bytes = "bytes", tag = "4")]
+    signature: Bytes,
+    /// 32 B digest of the full serialized SignedBlock: rebuild proof + error distinction.
+    #[prost(bytes = "bytes", tag = "5")]
+    claimed_block_digest: Bytes,
```
akichidis, line 67: I would go with either `digest` or `block_digest`.

consensus/core/src/slim_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+        }
+        if let Some(round) = o.round
+            && round >= skeleton.round()
+        {
+            return Err(DecodeError::Malformed(format!(
+                "override round {round} not below block round"
+            )));
+        }
+        if let Some(digest) = &o.digest
+            && digest.len() != 32
```
akichidis: let's use a const instead of hardcoding `32`

consensus/core/src/slim_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+) -> Result<Vec<BlockRef>, DecodeError> {
+    let ParsedEnvelope {
+        minimal,
+        skeleton,
+        block_ref,
+        ..
+    } = parsed;
+    let block_ref = *block_ref;
+    let authors = &minimal.ancestor_authors;
+    let default_round = skeleton.round().saturating_sub(1);
```
akichidis: Maybe instead of `default_round` use something like `parent_round`?

consensus/core/src/minimal_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+        ..
+    } = parsed;
+    let block_ref = *block_ref;
+    let authors = &minimal.ancestor_authors;
+    let default_round = skeleton.round().saturating_sub(1);
+
+    // Each ancestor must resolve to exactly one digest. A slot the sender wrote
+    // explicitly already has one; an omitted slot must be unique in local state.
+    //
+    // Anything else falls back to fetching the full block. That is deliberate for
```
akichidis: is that still true or there are any attempts to fetch missing referrencies by slot position?

consensus/core/src/minimal_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+    // costs one full block, which is what propagation cost before this existed, and a
+    // validator equivocating often enough to matter is a problem for ancestor selection
+    // rather than for the codec.
+    let mut ancestors = Vec::with_capacity(authors.len());
+    let mut missing: Vec<Slot> = Vec::new();
+    for &author in authors {
+        let authority = committee
+            .to_authority_index(author as usize)
+            .expect("validated above");
+        let o = minimal.overrides.iter().find(|o| o.author == author);
```
akichidis: I would be a bit more descriptive here and rename `o` to `override`

consensus/core/src/minimal_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+            block_ref,
+            reason: FallbackReason::MissingAncestors(missing),
+        });
+    }
+    Ok(ancestors)
+}
+
+/// Reassembles the author's exact serialization from the resolved ancestors and accepts
+/// it only if it hashes to the digest the sender claimed. Touches no local state.
+pub(crate) fn rebuild_minimal(
```
akichidis: maybe `rebuild_from_minimal` ?

consensus/core/src/slim_block.rs
```diff
@@ -0,0 +1,1198 @@
...
+    pub(crate) fn encode(
+        &self,
+        block: &VerifiedBlock,
+        dag_state: &RwLock<DagState>,
+    ) -> Result<Bytes, bcs::Error> {
+        let dag_state = dag_state.read();
+        let resolver = LocalStateResolver {
+            genesis_digests: &self.genesis_digests,
+            dag_state: &dag_state,
+        };
```
akichidis: maybe the `dag_state` can be dropped, or put the code inside a block? No reason to hold the read lock during the serialisation

consensus/core/src/minimal_block.rs
arun-koshy: Resolved, went with Mingwei's suggestion of `slim`. The module, codec, protocol flag (`consensus_slim_block_propagation`), metrics, and PR titles are all renamed across the stack.

consensus/core/src/network/tonic_network.rs
arun-koshy: Fair and this one resolved itself: that comment and the tests it described were removed in the wire-format rework. The seam is now covered by the envelope round-trip tests in #27695, whose docs are plainer.

consensus/core/src/slim_block.rs
arun-koshy: Agreed renamed to `block`. You are right that the type already says it carries a block without ancestors, so the field does not need to restate it; the doc comment keeps the detail. (`sans` is just "without", but not worth the lookup.)

consensus/core/src/slim_block.rs
arun-koshy: Adopted, thanks for writing it out. The early returns read much better than the stacked negations, and each omission rule now sits next to the reason it exists.

consensus/core/src/minimal_block.rs
arun-koshy: Renamed, though to `stripped_block` rather than plain `block`: the rebuild phase has the finished block in scope too, and the point of this local is that it is not a complete block yet (empty ancestors, unverified).

consensus/core/src/slim_block.rs
arun-koshy: I would keep the `claimed_` prefix on this one, it is doing real work. The digest is unverified sender input, and the whole trust model is that the rebuilt bytes must hash to what the sender *claimed*; `block_digest` reads as though we already know it is the block digest. Happy to shorten to `claimed_digest` (which is what the parsed struct calls it) if the length is the issue.

consensus/core/src/slim_block.rs
arun-koshy: Done, `consensus_config::DIGEST_LENGTH` already exists (it is what `BlockDigest` is defined with), so both the length checks and their error messages use it now.

consensus/core/src/slim_block.rs
arun-koshy: Done, `parent_round` is the more accurate name anyway

consensus/core/src/minimal_block.rs
arun-koshy: Agreed on being descriptive, went with `ancestor_override`.

consensus/core/src/minimal_block.rs
arun-koshy: Done, it is `rebuild_from_slim` now after the rename, and the `from` does read better.

consensus/core/src/slim_block.rs
arun-koshy: Good catch, and fixed. Encoding now splits along the same seam as decoding: `ancestor_overrides` is the only part that reads local state, so `encode` scopes the DagState guard to that call and clones and serializes the block after dropping it.

PR #27703, opened 2026-08-14
consensus: add Mysticeti v3 block proposal

mwtian:
## Description

Create and verify BlockV3 proposals with a signed transaction-vote cutoff, then limit the number of transaction vote targets in a V3 block. Keep Mysticeti v2 proposal behavior unchanged. `enable_v3` is `false` in all production configurations, so both commits are dormant.

**`consensus: add Mysticeti v3 block proposal`**

The proposer creates a `BlockV3` with a signed transaction vote cutoff round, which is the DAG GC round. The cutoff is the highest round for which the block carries no implicit accept votes. V1 and V2 report the genesis round, because they have no cutoff. The block verifier checks the cutoff and the transaction votes of a V3 block: the cutoff is below the block round, each vote target is above the cutoff and below the block round, targets are unique, each target has at least one reject index, reject indices increase, and the indices stay below the transaction limit.

**`consensus: limit v3 transaction vote targets`**

A V3 block can hold at most `gc_depth * committee size` vote targets. The proposer removes complete rounds of targets, oldest round first, and moves the signed cutoff round to cover every removed target. The block verifier applies the same limit before it scans the targets. V2 proposals keep all their vote targets, because V2 blocks have no signed cutoff round.

The transaction vote tracker limit is separate in #27739.

## Prerequisite before v3 is enabled

`CommitFinalizer` must use the signed cutoff round of a V3 block. It still infers the vote coverage of a block from commit GC rounds, so it can count an implicit accept vote for a target which the proposer removed above the GC round and at or below the signed cutoff. There is a `TODO` on `truncate_transaction_votes()` in `consensus/core/src/proposer.rs`.

## Test plan

- `SUI_SKIP_SIMTESTS=1 cargo nextest run -p consensus-core --lib` (334 passed, 1 skipped)
- `cargo xclippy -p consensus-core` (no warnings)
- `cargo fmt --all`

New tests:

- `block.rs`: the shared target limit, and the limit for a GC depth of zero.
- `proposer.rs`: the removal rules, which include no targets, targets at or below the GC round, exactly the limit, one target above the limit, and more targets than the limit in one round.
- `block_verifier.rs`: all block versions verify with v3 enabled, the vote target limit, the transaction count limit, the reserved index limit, and each invalid transaction vote condition.
- `core.rs`: the proposed V3 cutoff follows the GC round, a V3 proposal removes targets above the limit, and a V2 proposal keeps all its targets.

### DIFF
--- a/consensus/core/src/block_verifier.rs
+++ b/consensus/core/src/block_verifier.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use bytes::Bytes;
-use consensus_types::block::{BlockRef, TransactionIndex};
+use consensus_types::block::{BlockRef, NUM_RESERVED_TRANSACTION_INDICES, TransactionIndex};
 use std::{collections::BTreeSet, sync::Arc};
 
 use crate::{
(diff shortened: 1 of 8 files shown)


consensus/core/src/proposer.rs
```diff
@@ -189,6 +190,13 @@ impl ValidatorProposer {
                             AncestorState::Include => {
                                 trace!("Found ancestor {ancestor} with INCLUDE state for round {clock_round}");
                             }
+                            AncestorState::Exclude(score)
```
akichidis, line 201: Why is this change needed as part of this PR? Could we do it on a separate one? 

Also, I understand that we want to include to help with the certification of it etc, but isn't this supposed still a poor ancestor choice for the network?

consensus/core/src/block_verifier.rs
```diff
@@ -146,11 +146,100 @@ impl SignedBlockVerifier {
             });
         }
 
-        let batch: Vec<_> = block.transactions().iter().map(|t| t.data()).collect();
+        if self.context.protocol_config.enable_v3() {
```
akichidis: Nit: probably reject a received V3 block when `enable_v3 == false` ? It shouldn't happen under normal circumstances, so that's why I don't consider it super important

consensus/core/src/block_verifier.rs
```diff
@@ -146,11 +146,100 @@ impl SignedBlockVerifier {
...
+            .max_num_transactions_in_block()
+            .min(
+                TransactionIndex::MAX
+                    .saturating_sub(NUM_RESERVED_TRANSACTION_INDICES)
+                    .into(),
+            ) as u16;
+
+        let mut vote_targets = BTreeSet::new();
+        for votes in transaction_votes {
+            if votes.block_ref.round >= block.round() {
```
akichidis, line 220: Let's add a check that `votes.block_ref.author` is also within the committee bounds.

akichidis (approved):

consensus/core/src/proposer.rs
mwtian: When there are many leaders per round, this logic becomes necessary. And I believe it helps the existing logic as well. I can guard this with enable_v3().

mwtian: Guarded it with `enable_v3()`, so v2 ancestor selection is unchanged by this PR: without v3 the leader slots are empty here and an EXCLUDE leader stays excluded, as before.

It is needed for v3 because a v3 proposal waits for the leader of the round, so dropping that leader from the ancestors would stall the wait and delay the commit. You are right that it is still a poor ancestor otherwise, which is why the override is limited to the leader slots the proposer is waiting for.

Restored the original v2 `test_smart_ancestor_selection`, and added `test_smart_ancestor_selection_keeps_v3_leader` for the v3 rule.

consensus/core/src/block_verifier.rs
mwtian: Added a block version check before this: the block version must match the protocol config of the epoch, so a V3 block is rejected when `enable_v3 == false`, and a V1 or V2 block is rejected when it is true. Every authority in an epoch uses the same protocol config, so an honest block always matches.

New error: `ConsensusError::UnexpectedBlockVersion`. `test_v3_validates_all_block_versions` became `test_block_version_matches_protocol_config` and now covers both directions.

consensus/core/src/block_verifier.rs
mwtian: Added the check.

For the record, an invalid authority index in a vote target is tolerated today rather than mishandled: the target `BlockRef` is only used as a map key, in `TransactionVoteTracker` (`votes` map) and in `CommitFinalizer` (`BlockState::reject_votes`). Vote aggregation indexes the committee by the *voting* block author, which is already validated, so an out of range target author cannot panic there or move stake. Such an entry simply never matches a real block, stays inside the vote target limit, and is removed by GC on round.

So the check rejects the block early instead of accepting junk state.

arun-koshy (approved):

PR #27704, opened 2026-08-14
Fix controller observation bug

nickvikeras:
## Description

There was another controller clamp down that didn't recover automatically due to a huge read spike in testnet.

I believe this is due a bug in the controller.

The controller only increases the limit if there is actual write "demand" at least 80% of that limit. This prevents the limit from increasing to the maxmimum possible value when the indexer is in steady state and limited by chain execution throughput. If we didn't have this utilization check, big table would keep telling the indexer its fine to increase the rate, the indexer would increase its limit, but the chain wouldn't actually produce more data to test the limit. Then the limit would reach a very high value and when a backfill started, it would be too high.

When closing the observation window we check the number of requests that started in that window and divide that by the length of the window to compute a QPS, then we compare that to the limit.

The bug was that if an RPC start caused us to close the observation window, that RPC start wouldn't be counted in the window (or the next window, it was just not observed).

So at very low QPS, we never actually were able to measured >80% utilization and got stuck.

I also increased the min QPS from 0.1 to 1.0, which I think is more reasonable. I think 0.1 was originally copied from the java client, but for the Sui indexing use case, it's probably too low. If we clamp down 0.1 due to a load spike it would take about 18 minutes to ramp back up to steady state whereas if we clamp down to 1, it would take just ~1 minute.

## Test plan

New TDD-style unit tests to reproduce the bug. Once I deploy this, I plan to do a quick load test to reproduce the issue and verify writes clamp down again but dont get stuck.

### DIFF
--- a/crates/sui-kvstore/src/bigtable/client/flow_control.rs
+++ b/crates/sui-kvstore/src/bigtable/client/flow_control.rs
@@ -21,7 +21,7 @@ const DEFAULT_PERIOD: Duration = Duration::from_secs(10);
 const OBSERVATION_WINDOW: Duration = Duration::from_secs(1);
 const LATENCY_EVALUATION_PERIOD: Duration = Duration::from_secs(10);
 const INITIAL_QPS: f64 = 10.0;
-const MIN_QPS: f64 = 0.1;
+const MIN_QPS: f64 = 1.0;
 const MAX_QPS: f64 = 100_000.0;
 const MIN_FACTOR: f64 = 0.7;
 const MAX_FACTOR: f64 = 1.3;
@@ -232,6 +232,22 @@ impl ControllerState {
             .expect("observed feedback window disappeared")
             .started_at = now;
     }
+    fn restart_or_close_empty_observation_if_ready(&mut self, now: Instant) {
+        let Some(observation) = self.observation.as_ref() else {
+            return;
+        };
+        let elapsed = now.saturating_duration_since(observation.started_at);
+        if elapsed < OBSERVATION_WINDOW || observation.rpc_starts != 0 {
+            return;
+        }
+
+        // A low paced rate can make the fixed window empty even under sustained demand.
+        // Growth-only feedback remains valid through the next permit interval and one
+        // observation boundary. Conservative feedback restarts without including idle time.
+        if self.pending.growth_factor().is_none() || elapsed >= self.growth_observation_timeout() {
+            self.restart_or_close_empty_observation(now);
+        }
+    }
 
     fn reserve_permit(&mut self, now: Instant) -> PermitReservation {
         let permit_interval = self.permit_interval();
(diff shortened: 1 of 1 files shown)


crates/sui-kvstore/src/bigtable/client/flow_control.rs
tpham-mysten, line 221: With the new min QPS of 1, I think this will always be LATENCY_EVALUATION_PERIOD?

Do we need to modify this value?

crates/sui-kvstore/src/bigtable/client/flow_control.rs
```diff
@@ -351,8 +374,9 @@ impl BatchWriteFlowController {
                     .lock()
                     .expect("flow-control state mutex poisoned");
                 let now = Instant::now();
-                let completed_observation = Self::finish_observation_if_ready(&mut state, now);
+                state.restart_or_close_empty_observation_if_ready(now);
                 let admitted = state.complete_reservation(reservation);
+                let completed_observation = Self::finish_observation_if_ready(&mut state, now);
```
tpham-mysten, line 381: Want to check my understanding of 377-379: this call has to precede `complete_reservation` so it can reset started_at and keep idle time out of the window (otherwise the window closes with observed_start_qps ~ 0 and reanchors to the floor), while finish_observation_if_ready has to
follow it, which is the fix in this PR. Is that right?

If so, worth a line or two of comment here the way you did for the pre-sleep block at 353. The ordering is invisible right now, and merging these or narrowing the lock scope would look like a safe refactor.

tpham-mysten (approved):
The fix and increasing min QPS to 1.0 makes sense. Just some clarification comments

crates/sui-kvstore/src/bigtable/client/flow_control.rs
nickvikeras: Yeah thats right. I can add a comment 👍

crates/sui-kvstore/src/bigtable/client/flow_control.rs
nickvikeras: You are correct that with the current tuning it doesn't matter, but I kinda think it makes sense to keep it in case someone ever raises MIN_WINDOW_SAMPLES or lowers MIN_QPS.

tpham-mysten: Sure. Maybe leave a comment saying the intention to leave it as-is?

crates/sui-kvstore/src/bigtable/client/flow_control.rs
```diff
@@ -232,6 +232,22 @@ impl ControllerState {
...
+            return;
+        }
+
+        // A low paced rate can make the fixed window empty even under sustained demand.
+        // Growth-only feedback remains valid through the next permit interval and one
+        // observation boundary. Conservative feedback restarts without including idle time.
+        if self.pending.growth_factor().is_none() || elapsed >= self.growth_observation_timeout() {
+            self.restart_or_close_empty_observation(now);
+        }
+    }
```
wlmyng, line 250: I think we might benefit from this method being named "expire_idle_observation" or something

But I think what's emerging though is, we basically want a window to be long enough to contain a few permit slots, and we're trying to disambiguate empty as an artifact of pacing or from reduced demand/ some state-change induced on boundary

What about making the window  length a function of the rate?
```
  const MIN_OBSERVATION_WINDOW: Duration = Duration::from_secs(1);  
  const MIN_WINDOW_PERMITS: u32 = 2;

  impl ControllerState {
      fn observation_window(&self) -> Duration {
          (self.permit_interval() * MIN_WINDOW_PERMITS).max(MIN_OBSERVATION_WINDOW)
      }
  }
```
So then the `*_if_ready` can become:
```
  fn restart_expired_empty_observation(&mut self, now: Instant) {
      let Some(observation) = self.observation.as_ref() else { return };
      if observation.rpc_starts != 0 { return; }
      if now.saturating_duration_since(observation.started_at) <
  self.observation_window() {
          return;
      }
      self.restart_or_close_empty_observation(now);
  }

```

PR #27709, opened 2026-08-14
[graphql-stream] Support previousTransaction with cached streamed transactions [20/n]

tpham-mysten:
## Description

On a live subscription, `Object.previousTransaction` resolved only its digest: the contents (kind, sender, effects) came back empty. The lookup falls through to the KV backend, which lags the streamed tip, so a transaction that was just streamed is not yet readable there.

The fix caches each streamed checkpoint's transactions in memory (`StreamedTransactionStore`), consulted before the KV backend so a just-streamed `previousTransaction` resolves from memory and hands off to the backend once it catches up. This mirrors the existing `StreamedPackageStore`, so the shared cache-and-evict logic is pulled into a generic `StreamedStore`, and both stores are now flushed by a single eviction task keyed on each one's backing watermark.

## Test plan

- e2e + unit tests

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425
   - #26495
   - #26731
   - #26714
   - #27140
   - #27537
   - #27564
   - #27580
   - #27635
   - #27709

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-graphql/src/api/types/transaction_effects.rs
```diff
@@ -586,6 +587,20 @@ impl EffectsContents {
...
+        // checkpoint-bounded read only references transactions the backend already has.
+        if self.scope.checkpoint_viewed_at().is_none()
+            && let Some(streaming_transactions) = ctx.data_opt::<Arc<StreamedTransactionStore>>()
+            && let Some(contents) = streaming_transactions.get(&digest)
+        {
+            return Ok(Self {
+                scope: self.scope.clone(),
+                contents: Some(contents),
+            });
+        }
```
wlmyng: I think it's worth moving `checkpoint_viewed_at` on `Scope` into `DataSource`:
```
  enum DataSource {
      Indexed { viewed_at: u64 },   // normal query path
      Backfill,                     // indexed reads, deliberately unbounded
      Executed { execution_objects: ExecutionObjectMap },
      Streamed { checkpoint: Arc<ProcessedCheckpoint> },
  }
```

so that fetch bodies can match on the datasource and decide where to retrieve data accordingly

```
  match scope.data_source() {
      Executed { .. } => Ok(None),               
      Streamed { .. } | Backfill => {
          if let Some(c) = streamed_store.get(&digest) { return Ok(Some(c)); }
          kv.load_one_transaction(digest).await
      }
      Indexed { viewed_at } => {
          let tx = kv.load_one_transaction(digest).await?;
          Ok(tx.filter(|t| t.cp_sequence_number() <= viewed_at))  
      }
  }
```

At least so that the logic in L578-602 are bundled together, as the current scope (execution, stream, or indexed data) impacts where we can read from, and given that Executed and Streamed already hold some relevant data.

Building on this, what do you think about having some `Arc<StreamedOverlays>` that bundles all the streamed and cached data, such that `CheckpointStreamTask` writes to it through `index_and_broadcast`, eviction task drains it, and these overlays get handed to the resolvers as we do with `package_store` today through `matching_edges`:

```
S::matching_edges(checkpoint, &overlays, ...)
```

Then we can slap it into scope
```
Scope::for_streamed_checkpoint(overlays, limits, checkpoint)
    → package_store field = overlays.packages
    → DataSource::Streamed { checkpoint, overlays }
```

culminating in
```
  Streamed { overlays, .. } => overlays.transactions.get(&digest)
      .map_or_else(|| kv.load(digest), ...),
  Backfill => kv.load(digest), 
```

tpham-mysten: > I think it's worth moving checkpoint_viewed_at on Scope into DataSource:

Yes, this is what we discussed at standup. But I think it's orthogonal: we can get the match style on the existing enum, without moving `checkpoint_viewed_at`, like:

```rust
match scope.data_source() {
    DataSource::Executed { .. } => Ok(None),
    DataSource::Streamed { .. } => {
        if let Some(c) = streamed_store.get(&digest) { return Ok(Some(c)); }
        kv.load_one_transaction(digest).await
    }
    DataSource::Indexed => match scope.checkpoint_viewed_at() {
        None => {                         // backfill: unbounded indexed read
            if let Some(c) = streamed_store.get(&digest) { return Ok(Some(c)); }
            kv.load_one_transaction(digest).await
        }
        Some(viewed_at) => {              // bounded query
            let tx = kv.load_one_transaction(digest).await?;
            Ok(tx.filter(|t| t.cp_sequence_number() <= viewed_at))
        }
    },
}
```

> At least so that the logic in L578-602 are bundled together, as the current scope (execution, stream, or indexed data) impacts where we can read from, and given that Executed and Streamed already hold some relevant data.

I get the motivation for this, and I think it's more a style/standard call than a correctness one, so whichever we pick, we should use everywhere, and it can go in a follow-up PR.

Here's how I see the two:

**match on `DataSource`**
- Pro: makes it clear the mode is what decides where we read from, and it's exhaustive, so a new variant forces us to handle it.
- Con: the `kv.load_one_transaction` and the streamed-store check get repeated in several cases.

**`if`-chain (current)**
- Pro: reads as a step-by-step lookup: try the nearest place we might already have it (already hydrated, then on the scope, then the streamed store), then fall back to KV. Same flow in every mode; the mode only affects two steps (is the store usable, does the KV read need a cutoff). It's the usual "do we have it, else fetch it" shape we use in a lot of other lookups.
- Con: the mode's effect is spread across a few guards instead of one place, so "mode decides source" is implicit.

I slightly lean 60:40 to the `if`-chain, since the main flow is this step-by-step lookup and the mode only tweaks two steps, so a match ends up repeating the same logic arms. But it's close and mostly taste, so happy to bring it to standup. Either way it doesn't block this PR.

> Building on this, what do you think about having some Arc<StreamedOverlays> that bundles all the streamed and cached data, such that CheckpointStreamTask writes to it through index_and_broadcast, eviction task drains it, and these overlays get handed to the resolvers as we do with package_store today through matching_edges:

Yeah I think it's nice wrapper. I have implemented that ~`StreamedOverlays`~

Actually, I found the name `StreamedOverlays` is a bit confusing as it's basically an object holding shared cached store. WDYT about `StreamedCaches`?

tpham-mysten (commented):
Thanks @wlmyng!

crates/sui-indexer-alt-graphql/src/api/types/object.rs
```diff
@@ -1224,19 +1224,28 @@ impl Object {
...
+                        });
+
+                if let Some(object) = in_memory {
+                    return Ok(Some(object));
                 }
+
+                Ok(kv_loader
+                    .load_one_object(self.super_.address.into(), version.into())
+                    .await
+                    .context("Failed to fetch object contents")?)
```
amnn: nit: this fits the pattern of what we had here before (makes the cascade a bit clearer, perhaps?)

```suggestion
                let scope = &self.super_.scope;
                if let some(executed) = scope
                    .execution_output_object(self.super_.address.into(), version.into())
                {
                    Ok(Some(executed.clone()))
                } else if let Some(streamed) = scope
                    .streamed_object_store()
                    .and_then(|s| s.get(self.super_.address.into(), version.into())
                {
                    Ok(Some(streamed.clone())
                } else {
                    Ok(kv_loader
                        .load_one_object(self.super_.address.into(), version.into())
                        .await
                        .context("Failed to fetch object contents")?)
                }
```

crates/sui-indexer-alt-graphql/src/task/streaming/streamed_store.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// The value held under `key`, if any.
+    pub(crate) fn get(&self, key: &K) -> Option<V> {
+        self.entries.get(key).map(|entry| entry.value.clone())
+    }
+
+    /// Drop entries introduced at a checkpoint at or below `indexed_checkpoint`. A key re-introduced
+    /// at a later checkpoint survives, since its entry records that later checkpoint.
+    pub(crate) fn evict_up_to(&self, indexed_checkpoint: u64) {
+        self.entries
+            .retain(|_, entry| entry.checkpoint > indexed_checkpoint);
```
amnn, line 42: This is not a new issue, but just observing that this access pattern is not particularly lock friendly. No action needed for now, but something to watch out for.

amnn (approved):
Looks good!

crates/sui-indexer-alt-graphql/src/task/streaming/streamed_store.rs
tpham-mysten: Sure ack!

crates/sui-indexer-alt-graphql/src/api/types/object.rs
tpham-mysten: Sure done!

PR #27711, opened 2026-08-14
[trace-view] Added binary version check during bytecode verification

awelc:
## Description 

Since we already utilized Sui CLI binary in the debugger extension, we might just as well check if it is of the right version to run bytecode verification.

## Test plan 

Tested manually to see if the old version is detected properly


tzakian (approved):
Seems reasonable to me!

PR #27714, opened 2026-08-15
ledger-history: lane-neutral intra-tx coordinate + Position accessors

wlmyng:
## Description

Stacked on #27713. Makes the intra-transaction coordinate machinery lane-neutral so ListPackages can reuse it instead of adding another hand-rolled copy:

- `EventPosition` → `IntraTxCoordinate` (with `IntraTxScanBounds`, `ResolvedIntraTxRange`, and method renames to match). Rust-side vocabulary only — the wire `Position::Events` variant and the packed `event_seq` encoding are untouched. The coordinate's field is still named `event_index`; renaming it is deferred to a compiler-assisted pass (many sites are textually ambiguous with the wire field of the same name).
- `Position::scalar()` / `Position::intra_tx()` accessors on the cursor type, with decode delegating to them — a new per-transaction-indexed lane extends `intra_tx()` rather than teaching another `match` about its variant.

## Test plan

No behavior change; covered by existing lib tests in `sui-rpc-api`/`sui-kv-rpc` and the ledger-history characterization snapshot suite (draft: verification round in progress).

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-kv-rpc/src/bigtable_client.rs
+++ b/crates/sui-kv-rpc/src/bigtable_client.rs
@@ -50,7 +50,7 @@ use sui_kvstore::RowFilter;
 use sui_kvstore::TransactionData;
 use sui_kvstore::TxSeqDigestData;
 use sui_rpc_api::RpcError;
-use sui_rpc_api::ledger_history::query_options::EventPosition;
+use sui_rpc_api::ledger_history::query_options::IntraTxCoordinate;
 use sui_types::digests::TransactionDigest;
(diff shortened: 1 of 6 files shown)


wlmyng:
Folded into #27713 — the endpoint-shape and lane-neutral commits review as one story; same commits, same verification.

PR #27721, opened 2026-08-17
graphql: add addressable derived object lookups

amnn:
## Description

Expose point and multi-get derived-object lookups on `IAddressable`. They work similarly to dynamic field lookups in that you can fetch an object given a parent address and a name that is either a type + bcs pair or a Display v2 literal. They differ in that a derived object is not version bounded by its parent.

## Test plan

Added `crates/sui-indexer-alt-e2e-tests/tests/graphql/addressable/derived_objects.move`, covering point and multi-get lookups, explicit version and checkpoint bounds, deleted and missing objects, and independence from the parent’s version or checkpoint bound.

Run it with:

```sh
cargo nextest run -p sui-indexer-alt-e2e-tests --test transactional_tests graphql/addressable/derived_objects.move
```

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [x] GraphQL: Adds `derivedObject` and `multiGetDerivedObjects` to `IAddressable`.
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/dynamic_field.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/dynamic_field.rs
@@ -55,7 +55,6 @@ use crate::api::types::transaction_object::TransactionObject;
 use crate::config::Limits;
 use crate::error::RpcError;
 use crate::error::bad_user_input;
-use crate::error::upcast;
 use crate::pagination::Page;
 use crate::pagination::StreamConnection;
 use crate::scope::Scope;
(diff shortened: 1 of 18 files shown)


crates/sui-indexer-alt-graphql/src/api/types/dynamic_field.rs
```diff
@@ -106,6 +105,45 @@ pub(crate) struct DynamicFieldName {
     pub(crate) literal: Option<String>,
 }
 
+/// Identifies a dynamic field or derived object under a parent object.
+///
+/// At most one of `version`, `rootVersion`, or `atCheckpoint` may be specified.
+#[derive(InputObject)]
```
tpham-mysten, line 111: Do we need to derive InputObject here?

tpham-mysten (approved):
Looking great!

crates/sui-indexer-alt-graphql/src/api/types/dynamic_field.rs
tpham-mysten: oh nvm, I see it was added to schema in #27743

PR #27712, opened 2026-08-15
sui-kv-rpc: remove enable-list-apis flag, always serve the List APIs

evan-wall-mysten:
## Description

Stacked on #27666: GraphQL's `transactions`/`events` pagination now always requires the ledger gRPC List APIs with no Postgres fallback, so any `sui-kv-rpc` instance backing it needs `enable-list-apis` on regardless. Removes the flag (and its `enable-experimental-query-apis` alias), making `ListCheckpoints`/`ListTransactions`/`ListEvents` and their `GetServiceInfo` watermark coverage unconditional.

## Test plan

Removed `test_list_apis_disabled_unless_enabled`, which asserted the now-removed disabled-by-default behavior. Ran `sui-kv-rpc`'s unit tests (124 tests) and the affected `sui-indexer-alt-e2e-tests` suites (`kv_rpc_tests`, `graphql_checkpoint_transactions_tests`, `graphql_transactions_query_tests`, `graphql_epoch_transactions_tests`, `graphql_bitmap_pagination_tests`, `graphql_available_range_tests`, `graphql_rich_query_limits`, `ledger_grpc_watermark_tests`) — all pass.

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [x] gRPC: `sui-kv-rpc`'s `--enable-list-apis` config field (and its `--enable-experimental-query-apis` alias) is removed; `ListCheckpoints`/`ListTransactions`/`ListEvents` are now always served.
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #27723, opened 2026-08-17
sui-indexer-alt-reader: merge AlphaLedgerGrpcReader into LedgerGrpcReader

evan-wall-mysten:
## Description

Stacked on #27712.

The gRPC List APIs (`ListTransactions`/`ListEvents`) are no longer alpha as of #27666 — they're the only path now, with `AlphaLedgerGrpcReader` opened as a second, separate gRPC connection to the same `ledger_grpc_url` alongside `LedgerGrpcReader`'s point-lookup connection. This folds `list_transactions`/`list_events` (and `StreamPage`/`PageItem`) directly into `LedgerGrpcReader`, deletes `AlphaLedgerGrpcReader`, and updates the GraphQL server to construct and register a single reader instead of two.

**Side effect:** the separate `graphql_alpha_ledger_grpc` metrics prefix is gone — List/Stream calls now report under the same `graphql_ledger_grpc` prefix as point lookups. Checked `dashboards/` for references to the old prefix; found none.

## Test plan

`cargo check`/`clippy` clean on `sui-indexer-alt-reader`, `sui-indexer-alt-graphql`, `sui-indexer-alt-jsonrpc`, `sui-indexer-alt-e2e-tests`. All 27 `sui-indexer-alt-reader` lib tests pass (including the moved `ledger_grpc_reader::tests::*` stream-draining tests). `cargo fmt` clean.

## Release notes

- [ ] Protocol: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-reader/src/ledger_grpc_reader.rs
```diff
@@ -58,6 +64,34 @@ pub struct LedgerGrpcReader {
...
+enum FrameKind<T> {
+    Frame {
+        payload: Option<T>,
+        cursor: Option<Bytes>,
+        end_reason: Option<grpc::QueryEndReason>,
+    },
+    /// A frame with none of the known fields set (unknown/future frame kind).
+    Unknown,
+}
+
```
amnn, line 94: nit: the constants are conventionally put before the types.

amnn (approved):
LGTM, thanks!

PR #27724, opened 2026-08-17
[sui-rpc-api][sui-kv-rpc] Rename `Event*` range elements to `IntraTx*`, two step resolution from cp range to scan range

wlmyng:
## Description

Initial consolidation of checkpoint-window resolution so a resolved scan is constructed one way and owns its lifecycle identically in sui-rpc-api and usi-kv-rpc.

Every `resolve_{tx,event}_range` special-cased the empty checkpoint window with an early return. However, this obfuscated the fact that it still performs a checkpoint-summary read to map cp to tx, and hand-built the same record through a second constructor `empty_at`. These parallel paths had to independently agree, which led to multiple copies and potential drift. One thing to flag is the entry_checkpoint instantiation on an empty window special case (where on descending, entry_checkpoint = terminal_checkpoint, while in the non-empty case it is `terminal_checkpoint - 1`. Consider an example where `[10, 10)` descending vs `[3, 10)` dsecending.)

This special case was eliminated once `checkpoint_to_tx_range` returned the corresponding `[t, t)` range for some empty `[c, c)` range.

  Commits

  - e9f503d — kv checkpoint_to_tx_range is empty-safe: [x,x) → [t,t), [0,0) → 0..0, and non-empty windows resolve both boundaries in one batched read. Callers now project checkpoint→tx range unconditionally; since the empty path always did a terminal lookup anyway, this costs nothing.
  - 9b87dd1 — remove the CheckpointRange middleman: every caller immediately resolved it, so ResolvedCheckpointRange::from_request now validates, clamps to the indexed tip, tightens by cursors, and attributes exhaustion in one call.
  - 7015b5f — lane-neutral renames: EventPosition / EventScanBounds / ResolvedEventRange → IntraTxCoordinate / IntraTxScanBounds / ResolvedIntraTxRange. Events are today's only per-transaction-indexed lane, not forever.
  - b9ef884 — ResolvedIntraTxRange::resolve(cp_range, tx_range, options): one constructor replacing empty_at plus the per-call-site struct literals, reducing the ways the record can be built.
  - 464e065 / 8cffd7e — move apply_cursor_bounds from QueryOptions onto ResolvedIntraTxRange / ResolvedRange. Pure move: the resolved record owns its lifecycle, in preparation for a generic range over Bound<T> in the next PR.

## Test plan

Existing tests pass.

New unit tests for `ResolvedIntraTxRange::resolve`
- empty windows produce the correct record in ascending or descending ordering
- entry/ terminal
- cursor application identical

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-kv-rpc/src/bigtable_client.rs
+++ b/crates/sui-kv-rpc/src/bigtable_client.rs
@@ -50,7 +50,7 @@ use sui_kvstore::RowFilter;
 use sui_kvstore::TransactionData;
 use sui_kvstore::TxSeqDigestData;
 use sui_rpc_api::RpcError;
-use sui_rpc_api::ledger_history::query_options::EventPosition;
+use sui_rpc_api::ledger_history::query_options::IntraTxCoordinate;
 use sui_types::digests::TransactionDigest;
 use sui_types::messages_checkpoint::CheckpointSequenceNumber;
 use sui_types::object::Object;
@@ -410,7 +410,7 @@ impl BigTableClient {
     pub(crate) fn event_wm_resolver(
         &self,
         direction: ScanDirection,
-    ) -> impl Fn(EventPosition) -> WmResolverFut + Send + 'static {
+    ) -> impl Fn(IntraTxCoordinate) -> WmResolverFut + Send + 'static {
         let client = self.clone();
         move |position| {
             let client = client.clone();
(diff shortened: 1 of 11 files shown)


wlmyng:
Superseded by a re-cut of this stack with cleaner PR boundaries (renames → behavior-preserving consolidation with characterization pins → wire-semantics alignment → unification → symbolic resume): #27751 → #27752 → #27753 → #27754 → #27755. Content is carried over commit-for-commit; the old branch stays as reference.

PR #27730, opened 2026-08-17
[gql] Refactor `Scope.checkpoint_viewed_at` into `DataSource::Indexed`

tpham-mysten:
## Description

`Scope` carried `checkpoint_viewed_at: Option<u64>` as a field independent of `data_source: DataSource`, but the two were only ever set in four valid combinations, and a `Some` checkpoint appeared exclusively in the indexed-query path (execution, streamed, and backfill were always `None`). This folds the value into `DataSource::Indexed { viewed_at }` and adds a `Backfill` variant for the unbounded indexed reads of a subscription's catch-up, so a checkpoint bound can no longer coexist with a streamed or executed source. `checkpoint_viewed_at()` now derives from the enum.

The pinning constructor is renamed `with_checkpoint_viewed_at` to `with_indexed_view`, since calling it re-scopes to `DataSource::Indexed`; the name makes that mode switch visible at the call sites. This is an internal refactor with no schema or behavior change.

## Test plan

Covered by existing tests in `sui-indexer-alt-graphql` (library unit tests and the `graphql_subscription` e2e suite).

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/balance.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/balance.rs
@@ -175,7 +175,7 @@ impl Balance {
             (Some(c), _) | (_, Some(c)) => c.0,
         };
 
-        let Some(scope) = scope.with_checkpoint_viewed_at(ctx, checkpoint) else {
+        let Some(scope) = scope.with_indexed_view(ctx, checkpoint) else {
             return Err(bad_user_input(Error::Future(checkpoint)));
         };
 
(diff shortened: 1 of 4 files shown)


tpham-mysten:
Actually I need to rethink on Subscription backfill scenario so close it for now

PR #27731, opened 2026-08-17
Implement balance changes for archival

nickvikeras:
## Description

Binance reported this as missing

## Test plan

Unit test

### DIFF
--- a/crates/sui-kv-rpc/src/render.rs
+++ b/crates/sui-kv-rpc/src/render.rs
@@ -130,6 +130,11 @@ pub(crate) fn render_full_checkpoint(
             format!("checkpoint {cp_seq} contents column missing"),
         )
     })?;
+    let include_balance_changes = read_mask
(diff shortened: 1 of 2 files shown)


crates/sui-kv-rpc/src/render.rs
```diff
@@ -146,6 +151,9 @@ pub(crate) fn render_full_checkpoint(
                     format!("transaction {} effects column missing", tx.digest),
                 )
             })?;
+            if let Some(transaction_balance_changes) = transaction_balance_changes.as_mut() {
+                transaction_balance_changes.push(tx.balance_changes);
+            }
```
wlmyng, line 156: balance_changes slipped through the merge-based path because the merge source type (full_checkpoint_content::ExecutedTransaction) doesn't carry it right? Might be worth introducing a helper that both transaction_to_response and render_full_checkpoint call

wlmyng (approved):

bmwill (approved):

PR #27733, opened 2026-08-17
docs: promote Consumer App Cookbook in navigation

jessiemongeon1:
## Summary

- Promotes the Consumer App Cookbook (zkLogin + Enoki + Seal) to a top-level Getting Started sidebar item, directly after "Hello, World!"
- Adds cross-links under zkLogin and Enoki in the Sui Stack sidebar
- Removes the duplicate entry from Example Apps > Digital Assets

## Why

Builder feedback indicates that zkLogin + sponsored transactions are documented as separate modules rather than positioned as the recommended default flow. The Consumer App Cookbook already exists and covers this pattern well, but it was buried 3 levels deep under Getting Started > Example Apps > Digital Assets. Builders missed it.

## Navigation changes

**Before:**
```
Getting Started
├── Hello, World!
├── Example Apps
│   ├── Move Patterns
│   ├── Frontend Apps
│   └── Digital Assets
│       └── Consumer App Cookbook  ← buried here
```

**After:**
```
Getting Started
├── Hello, World!
├── Consumer App Cookbook  ← promoted to top level
├── Example Apps
│   └── ...

Sui Stack
├── Enoki
│   ├── Enoki Docs
│   ├── Consumer App Cookbook  ← cross-linked
│   └── ...
├── zkLogin
│   ├── ...
│   └── Consumer App Cookbook  ← cross-linked
```

## Test plan

- [x] `npx docusaurus build` passes
- [ ] Verify sidebar renders correctly in all three locations
- [ ] Confirm the cookbook page loads from each link

### DIFF
--- a/docs/content/sidebars.js
+++ b/docs/content/sidebars.js
@@ -264,6 +264,7 @@ export default {
         'getting-started/onboarding/next-steps',
       ],
     },
+    'getting-started/examples/consumer-app-zklogin',
     {
       type: 'category',
       label: 'Example Apps',
@@ -293,7 +294,6 @@ export default {
             'getting-started/examples/lootbox-ctf',
             'getting-started/examples/merchant-ctf',
             'getting-started/examples/staking-ctf',
-            'getting-started/examples/consumer-app-zklogin',
             'getting-started/examples/defi-trading-zklogin',
           ],
         },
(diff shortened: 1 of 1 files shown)


jessiemongeon1:
Superseded by #27913.

This can no longer merge: #27878 moved the cookbook page on Sept 3, so `getting-started/examples/consumer-app-zklogin` no longer exists and Docusaurus throws on unknown sidebar doc IDs. The `Example Apps > Digital Assets` category this removes an entry from is also gone, and the page already sits under zkLogin on main, so that cross-link would duplicate an entry in its own category.

Beyond staleness, the diagnosis here was right but the lever was wrong. Grepping every mdx shows nothing links to the cookbook in prose, anywhere. `next-steps.mdx`, whose whole job is answering what to do after Hello, World!, never mentions it, so a reader can finish Getting Started without learning the page exists. That explains the missed-it feedback better than sidebar depth does.

#27913 adds prose entry points on the pages builders actually pass through, and leaves the sidebar alone.

PR #27734, opened 2026-08-18
[discovery] reduce mailbox pressure from redundant endpoint updates and synchronous peer-cache saves

aschran:
## Description

- Suppress duplicate endpoint updates in`EndpointManager`. Deduplicating the consensus path also stops redundant updates from tearing down the peer's gRPC channel and subscription task every round.
- Discovery forwards addresses only for records that win the merge.
- Peer-cache file saves are debounced to at most one per tick via a snapshot fingerprint and moved off the event loop onto `spawn_blocking`.
- New metrics: discovery mailbox depth, per-tick high-water mark, and capacity.

## Test plan

New unit tests cover the dedup and stale-version-drop semantics (including version bump on identical-content updates and clear-resets-version), LRU eviction at capacity, consensus updater replacement/buffered replay/failure paths, fingerprint change detection across trust transitions and culls, single-flight save debounce, and a mailbox-bound test asserting repeated identical gossip enqueues one batch instead of one per round. `cargo simtest -p sui-e2e-tests discovery` exercises peer-cache save/reload across restart.

---

## Release notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Fixes discovery mailbox overflow panics under load by suppressing redundant peer-address updates and debouncing peer-cache writes. Adds `discovery_mailbox_depth`, `discovery_mailbox_high_water`, and `discovery_mailbox_capacity` metrics.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-network/src/discovery/metrics.rs
+++ b/crates/sui-network/src/discovery/metrics.rs
@@ -6,7 +6,10 @@ use prometheus::{
     IntGauge, IntGaugeVec, Registry, register_int_gauge_vec_with_registry,
     register_int_gauge_with_registry,
 };
-use std::sync::Arc;
+use std::sync::{
+    Arc,
+    atomic::{AtomicU64, Ordering},
+};
 use tap::Pipe;
 
 #[derive(Clone)]
@@ -46,11 +49,44 @@ impl Metrics {
                 .set(code);
         }
     }
+
+    /// Folds a sampled mailbox depth into the high-water mark accumulated
+    /// since the last [`Self::publish_mailbox_depth`].
+    pub fn observe_mailbox_depth(&self, depth: u64) {
+        if let Some(inner) = &self.0 {
+            inner
+                .mailbox_high_water_accumulator
+                .fetch_max(depth, Ordering::Relaxed);
+        }
+    }
+
+    /// Publishes the current depth and the accumulated high-water mark, then
+    /// resets the accumulator for the next interval.
+    pub fn publish_mailbox_depth(&self, depth: u64) {
+        if let Some(inner) = &self.0 {
+            let high_water = inner
+                .mailbox_high_water_accumulator
+                .swap(0, Ordering::Relaxed)
+                .max(depth);
+            inner.mailbox_depth.set(depth as i64);
+            inner.mailbox_high_water.set(high_water as i64);
+        }
+    }
+
+    pub fn set_mailbox_capacity(&self, value: i64) {
+        if let Some(inner) = &self.0 {
+            inner.mailbox_capacity.set(value);
+        }
+    }
 }
 
 struct Inner {
     num_peers_with_external_address: IntGauge,
     active_p2p_address_source: IntGaugeVec,
+    mailbox_depth: IntGauge,
+    mailbox_high_water: IntGauge,
+    mailbox_high_water_accumulator: AtomicU64,
+    mailbox_capacity: IntGauge,
 }
 
 impl Inner {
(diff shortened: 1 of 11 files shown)


crates/sui-network/src/endpoint_manager.rs
```diff
@@ -76,11 +161,49 @@ impl EndpointManager {
...
         addresses: Vec<Multiaddr>,
+    ) -> SuiResult<()> {
+        self.update_endpoint_impl(endpoint, source, None, addresses)
+    }
+
+    /// Like [`Self::update_endpoint`], but silently drops the update if
+    /// `version` is not newer than the last accepted version for this
+    /// (endpoint, source).
+    ///
+    /// A source should be either consistently versioned or not. Today
```
alex-mysten: Should we enforce this?

crates/sui-network/src/discovery/metrics.rs
```diff
@@ -46,11 +49,44 @@ impl Metrics {
...
+    pub fn set_mailbox_capacity(&self, value: i64) {
+        if let Some(inner) = &self.0 {
+            inner.mailbox_capacity.set(value);
+        }
+    }
 }
 
 struct Inner {
     num_peers_with_external_address: IntGauge,
     active_p2p_address_source: IntGaugeVec,
```
alex-mysten, line 85: The existing metrics in this file do not exist in Grafana. Should they?

crates/sui-network/src/endpoint_manager.rs
```diff
@@ -22,8 +24,79 @@ pub struct EndpointManager {
...
+            return false;
+        };
+        if let (Some(version), Some(cached_version)) = (version, cached.version)
+            && version <= cached_version
+        {
+            return true;
+        }
+        if cached.addresses == addresses {
+            if version.is_some() {
+                cached.version = version;
```
alex-mysten, line 74: Surprising that a predicate mutates

crates/sui-network/src/endpoint_manager.rs
```diff
@@ -76,11 +161,49 @@ impl EndpointManager {
...
         match endpoint {
             EndpointId::P2p(peer_id) => {
+                let key = (peer_id, source);
+                let mut last_sent = self.inner.last_sent_p2p.lock().unwrap();
+                if addresses.is_empty() {
+                    // Always dispatch clears and forget the key: the event
+                    // loop's Chain fallback can re-install Discovery
+                    // addresses from known_peers_v2 behind this cache's back,
+                    // so "already cleared" is never proof they're still gone.
+                    last_sent.remove(&key);
```
alex-mysten, line 207: Should we disambiguate clears from absent data on last_sent?

crates/sui-network/src/discovery/metrics.rs
aschran: should be fixed now

crates/sui-network/src/endpoint_manager.rs
aschran: good idea

crates/sui-network/src/endpoint_manager.rs
aschran: can you elaborate? what's the benefit?

crates/sui-network/src/endpoint_manager.rs
aschran: refactored to be clearer

PR #27735, opened 2026-08-18
Only rate limit backfills

nickvikeras:
## Description

We have had a few blips where a read load spike causes the write rate limits to kick in while the cluster autoscales which leads to lag. This PR lets writes for pipelines who are ahead of the "tip of chain" bypass flow control. "Tip of chain" being defined as the latest checkpoint known to the ingestion client at indexer startup. These tip-of-chain writes are already paced by chain execution, and the additional load on the DB by simply keeping up with that pace in minimal. Plus most clients would rather accept some high percentile read latency for fresher data during these load spikes than make the opposite trade.

## Test plan

Given the dependency on real workloads and bigtable behavior, the easiest way to test it for real is to deploy it. We should stop seeing lag spikes during minor load spikes in prod when this is deployed.

### DIFF
--- a/crates/sui-kvstore/src/bigtable/client/mod.rs
+++ b/crates/sui-kvstore/src/bigtable/client/mod.rs
@@ -142,6 +142,7 @@ impl ChannelPrimer for BigtablePrimer {
 pub struct BigTableClient {
     table_prefix: String,
     client: BigtableInternalClient<AuthChannel<ChannelPool>>,
+    flow_control_disabled_client: BigtableInternalClient<AuthChannel<ChannelPool>>,
     client_name: String,
     metrics: Option<Arc<KvMetrics>>,
     flow_controller: Option<Arc<BatchWriteFlowController>>,
@@ -170,12 +171,17 @@ impl BigTableClient {
             None,
             bigtable_features_header(batch_write_flow_control),
         );
+        let flow_control_disabled_auth_channel =
+            auth_channel.with_features_header(bigtable_features_header(false));
         let client_name = client_name.to_string();
         let flow_controller = batch_write_flow_control
             .then(|| BatchWriteFlowController::new(client_name.clone(), None));
         Ok(Self {
             table_prefix: format!("projects/emulator/instances/{}/tables/", instance_id),
             client: BigtableInternalClient::new(auth_channel),
+            flow_control_disabled_client: BigtableInternalClient::new(
+                flow_control_disabled_auth_channel,
+            ),
             client_name,
             metrics: None,
             flow_controller,
@@ -263,19 +269,35 @@ impl BigTableClient {
             Some(token_provider),
             bigtable_features_header(batch_write_flow_control),
         );
-        let client = BigtableInternalClient::new(auth_channel).max_decoding_message_size(
-            max_decoding_message_size.unwrap_or(DEFAULT_MAX_DECODING_MESSAGE_SIZE),
-        );
+        let flow_control_disabled_auth_channel =
+            auth_channel.with_features_header(bigtable_features_header(false));
+        let max_decoding_message_size =
+            max_decoding_message_size.unwrap_or(DEFAULT_MAX_DECODING_MESSAGE_SIZE);
+        let client = BigtableInternalClient::new(auth_channel)
+            .max_decoding_message_size(max_decoding_message_size);
+        let flow_control_disabled_client =
+            BigtableInternalClient::new(flow_control_disabled_auth_channel)
+                .max_decoding_message_size(max_decoding_message_size);
         let flow_controller = batch_write_flow_control
             .then(|| BatchWriteFlowController::new(client_name.clone(), metrics.clone()));
         Ok(Self {
             table_prefix,
             client,
+            flow_control_disabled_client,
             client_name,
             metrics,
             flow_controller,
             app_profile_id,
         })
+    }
+
+    /// Clone this client with batch-write flow-control feature advertisement and local admission
+    /// disabled.
+    pub(crate) fn without_batch_write_flow_control(&self) -> Self {
+        let mut client = self.clone();
+        client.client = client.flow_control_disabled_client.clone();
+        client.flow_controller = None;
+        client
     }
 
     /// Fetch transactions with an optional column filter for partial reads.
@@ -2305,6 +2327,25 @@ mod tests {
         })
         .await
         .expect("successful writes took too long to isolate flow-control feedback")
+    }
+
+    #[tokio::test]
+    async fn clone_without_batch_write_flow_control_disables_controller() {
+        let mock = MockBigtableServer::new();
+        let (addr, _handle) = mock.start().await.unwrap();
+        let controlled = BigTableClient::new_for_host(
+            addr.to_string(),
+            "test".to_string(),
+            "flow-control",
+            true,
+        )
+        .await
+        .unwrap();
+
+        let flow_control_disabled = controlled.without_batch_write_flow_control();
+
+        assert!(controlled.flow_controller.is_some());
+        assert!(flow_control_disabled.flow_controller.is_none());
     }
 
     #[tokio::test]
--- a/crates/sui-kvstore/src/bigtable/client/mod.rs
+++ b/crates/sui-kvstore/src/bigtable/client/mod.rs
@@ -94,6 +94,42 @@ const DEFAULT_CHANNEL_TIMEOUT: Duration = Duration::from_secs(60);
 /// user-tunable knob.
 pub(crate) const MAX_TX_DIGESTS_PER_REQUEST: usize = 10_000;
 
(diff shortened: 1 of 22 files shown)


crates/sui-kvstore/src/handlers/bitmap/handler.rs
```diff
@@ -27,31 +27,44 @@ use sui_types::full_checkpoint_content::Checkpoint;
...
     }
 }
 
 impl BitmapBatch {
     pub(crate) fn clone_shards(&self) -> Vec<Arc<Vec<BitmapIndexValue>>> {
-        self.0.clone()
+        self.shards.clone()
+    }
+
+    pub(crate) fn requires_batch_write_flow_control(&self) -> bool {
```
tpham-mysten: nit: is_batch_write_flow_control_required? [or something similar to that]

As the other look like an action?

crates/sui-kvstore/src/handlers/handler.rs
```diff
@@ -42,6 +42,19 @@ pub struct BigTableHandler<P> {
...
+/// A BigTable entry paired with the checkpoint that produced it.
+pub struct CheckpointedEntry {
+    checkpoint: u64,
+    entry: Entry,
+}
+
+impl CheckpointedEntry {
+    fn new(entry: Entry, checkpoint: u64) -> Self {
+        Self { checkpoint, entry }
+    }
```
wlmyng, line 57: it seems that the BitmapHandler side already tracks some `max_cp` to stamp watermarks on the merged output. Might be worth changing the framework so `batch()` receives the checkpoint as a parameter, since the `batch()` call is made per-checkpoint.

crates/sui-kvstore/src/store/mod.rs
```diff
@@ -154,10 +164,31 @@ struct PipelineInitResult {
...
+        use_batch_write_flow_control: bool,
+    ) -> Result<()> {
+        let mut client = if use_batch_write_flow_control {
+            self.backfill_client.clone()
+        } else {
+            self.live_client.clone()
+        };
+        client.write_entries(table, entries).await
+    }
+}
```
wlmyng: i see, the BigtableHandler threads the `use_batch_write_flow_control` all the way through here, where we give the backfill_client or live_client aka `client.without_batch_write_flow_control()`

So it's the header that informs the BigTable server whether the client cooperates with rate limiting

crates/sui-kvstore/src/store/bitmap_committer/generation.rs
```diff
@@ -326,9 +326,10 @@ impl GenerationLoopState {
 mod tests {
     use std::collections::HashMap;
     use std::sync::Arc;
-    use std::sync::Mutex;
     use std::time::Instant;
 
+    use parking_lot::Mutex;
```
amnn: out of interest, why the change in Mutex impl?

crates/sui-kvstore/src/lib.rs
```diff
@@ -398,6 +398,7 @@ impl BigTableIndexer {
             registry,
         )
         .await?;
+        let latest_checkpoint_at_startup = indexer.latest_checkpoint_at_startup();
```
amnn: What I'd had in mind:

- Initialise the `IngestionClient` from `ClientArgs`, then you can query the latest checkpoint agains that without introducing a new API.
- This also means you can parametrize or decorate the `store` with this information to centralize rate limiting control there, rather than pass `latest_checkpoint_at_startup` into each pipeline's handler. The contract is just that the pipelines pass checkpointed records to the store to commit.
- Then you construct a new indexer instance with `Indexer::with_ingestion_clients`.

I may be missing a detail why that's not possible, but just offering in case that cleans things up.

crates/sui-kvstore/src/store/mod.rs
```diff
@@ -86,10 +87,18 @@ pub struct BitmapInitialWatermark {
...
+pub(crate) struct BigTableBatchWriter {
+    backfill_client: BigTableClient,
+    live_client: BigTableClient,
+}
+
 /// A Store implementation backed by BigTable.
 #[derive(Clone)]
 pub struct BigTableStore {
     client: BigTableClient,
+    batch_writer: BigTableBatchWriter,
```
amnn: How come there's a `client` sitting outside `batch_writer` as well as the one inside? Would it be possible to inline `live_client` and `backfill_client` into the store and connection types (with the existing outer client being the `live_client`?) Hard for me to tell whether `BigTableBatchWriter` is justifying itself.

crates/sui-kvstore/src/bigtable/client/mod.rs
```diff
@@ -142,6 +142,7 @@ impl ChannelPrimer for BigtablePrimer {
 pub struct BigTableClient {
     table_prefix: String,
     client: BigtableInternalClient<AuthChannel<ChannelPool>>,
+    flow_control_disabled_client: BigtableInternalClient<AuthChannel<ChannelPool>>,
```
tpham-mysten: Could we get away with a single client here and handle the enable/disable decision at the call site instead? Two clients for one boolean feels heavier than it needs to be.

tpham-mysten (approved):
Looks reasonable. Having a comment on if we can simplify 2 fields for client in BigTableClient

amnn (approved):
LGTM -- I think there might be a way to avoid threading this through each handler though (suggested in comments).

crates/sui-kvstore/src/store/bitmap_committer/generation.rs
nickvikeras: The LLM just decided to do it because the repo guidance says to prefer parking_lot over std::sync for cases where you are just immediately calling `.unwrap()`. It's kind of a distraction but it seemed good to do if that's what the repo guidance is 🤷‍♂️

crates/sui-kvstore/src/store/bitmap_committer/writer.rs
```diff
@@ -190,12 +192,18 @@ impl WriteContext {
     }
 
     async fn attempt_write_chunk(&self, chunk: Chunk) -> Attempt {
+        let use_batch_write_flow_control = chunk
+            .rows
+            .iter()
+            .any(|row| row.use_batch_write_flow_control);
```
wlmyng: took me a sec to figure this out, am i understanding correctly that:

1. the `bitmap/handler.rs` receives checkpoints from framework, which gets converted into batched bitmap values stamped with backfill or live
2. on handoff to committer, actually the batch is sharded up for parallelism, and shard workers handle things from here
3. each shard worker merges the shard into an accumulation, and for any changes, emits a row. Consists of the label, backfill or live, and the actual contents, which is the whole bitmap
4. in the writer.rs writer task, regroup rows into chunks and send one. Here, one backfill row pollutes the rest

crates/sui-kvstore/src/store/bitmap_committer/generation.rs
amnn: Sure why not 😅 I have yet to come across a scenario where I wouldn't unwrap the poison guard...

nickvikeras: ^This is what the LLM told me when I asked it the same question, and I took it's answer for granted. But I don't actually seeing parking_lot referenced in the monorepo Claude.md file 🤔

wlmyng (commented):
The bool name `use_batch_write_flow_control` describes the consequence, but the thing being propagated through the layers is a fact on whether the write was triggered by a pre-startup-checkpoint commit. The
  policy of backfill -> flow-controlled client is applied at the very bottom in the router. It might be worth representing this as a `{Commit/Write}Source { Backfill, Live }` with the `OR/any()` logic, which helps elucidate
  locations like `write_entries(table, entries, Source::Backfill)`

crates/sui-kvstore/src/store/bitmap_committer/generation.rs
nickvikeras: Apparently it came from omp's harness-level rust rules - https://github.com/can1357/oh-my-pi/blob/main/packages/coding-agent/src/discovery/builtin-rules/rs-parking-lot.md

I had assumed it came from Sui's Claude.md file, which is why I just let the agent do it initially.

crates/sui-kvstore/src/handlers/handler.rs
nickvikeras: I was trying to avoid a breaking framework change since to fix these lag spikes, but I agree it would be nice to have the checkpoint number in `batch()`.

crates/sui-kvstore/src/store/bitmap_committer/writer.rs
nickvikeras: Yup

crates/sui-kvstore/src/lib.rs
nickvikeras: This worked.

PR #27736, opened 2026-08-18
[sui-rpc-api][sui-kv-rpc] refactor `Resolved*Range.apply_cursor_bounds` in preparation for single range on `Bound<T>`

wlmyng:
## Description 

Reorganize the logic around applying cursor bounds, in preparation for a single range generic on the scan coordinates

the refactor reveals the following asymmetries:
1. `Range<u64> vs `Bound<IntraTxCoordinate>`
2. consequently, scalar range decodes eagerly and operates on plain arithmetic, while the latter `IntraTx` is symbolic (`Item` maps to `Excluded`, `Boundary` maps to `Included`)

## Test plan 

How did you test the new or updated feature?

---

## Release notes

Check each box that your changes affect. If none of the boxes relate to your changes, release notes aren't required.

For each box you select, include information after the relevant heading that describes the impact of your changes that a user might notice and any actions they must take to implement updates. 

- [ ] Protocol: 
- [ ] Nodes (Validators and Full nodes): 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

PR #27738, opened 2026-08-18
[sui-kv-rpc][sui-rpc-api] Unify Range resolution

wlmyng:
## Description

Third PR of the ledger-history convergence stack (#27724 → #27736 → this): the scalar (transactions/checkpoints) and intra-tx (events) scan lanes finish converging onto one resolved record and one cursor-application body.

`ResolvedRange`/`ResolvedIntraTxRange` are replaced outright by `ResolvedScan<P>` over `ScanBounds<P>`; the scalar endpoint resolvers drop `with_range` and their per-endpoint empty branches and use the same one-call `resolve` as events; scalar cursor tests that became structural twins of the intra-tx family are pruned (the lane-divergent ones — successor canonicalization, raw-echo divergence, `u64::MAX` saturation — are kept). A final commit settles the construction vocabulary: `resolve` takes the scan window in position space and applies cursor bounds itself, and the one remaining per-lane seam is `ScanCursor<P>` on `CursorToken` (`coordinate` + `resume_lo`) — how cursors read being the only thing the lanes actually disagree about.

**Wire change** (first commit; terminal no-item frames only, data frames unchanged): terminal cursors now echo the raw request cursor instead of a derived coordinate —
- descending stop-side terminals echo the cursor's own position rather than its successor (after-Item@n previously echoed n+1, now n; the scan interval still starts at n+1);
- an ascending after-Item cursor that empties the interval retains Item kind (previously normalized to Boundary).

All echoed tokens are resume-equivalent to their predecessors — resuming from either the old or new echo scans the same interval. (One wire-invisible ride-along in the consolidate commit: scalar descending-empty records normalize `entry = terminal` instead of the `end - 1` formula; empty terminal frames never read entry.)

Follow-up (deliberately not here): raw echo enables scalar resume going fully symbolic (uniform `resume_lo`, `+1` confined to the store edge) and anchoring the record on `Position`-variant payloads for verbatim cursor echo. Tracked separately.

## Test plan

Per-arm cursor unit tests in both lanes (tighten / entry-fold / empty / crossed / exact-edge, both cursor kinds), canonical empty-window records, and the `u64::MAX` cases in `query_options`. The resume-equivalence claim is pinned directly by `raw_after_item_echo_resumes_like_successor_boundary` (after-Item@n and Boundary@n+1 admit the identical interval), with the descending pair asserted inside `tx_after_cursor_pins_descending_terminal`. Endpoint behavior is covered by the existing `list_ledger_history` e2e suite.

## Release notes

- [ ] Protocol:
- [x] gRPC: In the v2 ledger-history `ListTransactions`/`ListEvents`/`ListCheckpoints` endpoints, terminal watermark frames now echo the client's raw cursor position and kind instead of a derived successor coordinate. Echoed cursors remain valid resume points; no client action needed.
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-kv-rpc/src/v2/list_checkpoints.rs
+++ b/crates/sui-kv-rpc/src/v2/list_checkpoints.rs
@@ -662,7 +662,7 @@ fn resolve_cp_range(
     options: &QueryOptions,
 ) -> ResolvedScan<u64> {
     let range = cp_range.range.clone();
-    ResolvedScan::<u64>::resolve(cp_range, range, options).apply_cursor_bounds(options)
+    ResolvedScan::<u64>::resolve(cp_range, range, options)
 }
 
 fn decode_checkpoint_row_key(key: &Bytes) -> Result<u64, RpcError> {
--- a/crates/sui-kv-rpc/src/v2/list_checkpoints.rs
+++ b/crates/sui-kv-rpc/src/v2/list_checkpoints.rs
@@ -30,7 +30,7 @@ use sui_rpc_api::RpcError;
 use sui_rpc_api::ledger_history::query_options::QueryOptions;
 use sui_rpc_api::ledger_history::query_options::RangeExhaustion;
 use sui_rpc_api::ledger_history::query_options::ResolvedCheckpointRange;
-use sui_rpc_api::ledger_history::query_options::ResolvedRange;
+use sui_rpc_api::ledger_history::query_options::ResolvedScan;
 use sui_rpc_api::ledger_history::watermark::ScanTerminal;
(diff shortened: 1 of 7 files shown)


wlmyng:
Superseded by a re-cut of this stack with cleaner PR boundaries (renames → behavior-preserving consolidation with characterization pins → wire-semantics alignment → unification → symbolic resume): #27751 → #27752 → #27753 → #27754 → #27755. Content is carried over commit-for-commit; the old branch stays as reference.

PR #27749, opened 2026-08-19
Added replay-based support for tracing dry runs

awelc:
## Description 

This PR adds tracing support for `sui client ptb --dry-run` via an additional `--trace` flag. It produces artifacts similar to those created by the replay tool, most importantly the artifacts needed for trace debugging.

The traced dry-run executes locally, similarly to how the replay tool executes historical transactions, but instead of a historical transaction being replayed, it's a transaction dry-ran on a full-node.

The results of full-node dry-run remain the source of truth and are reported to the user. Traced dry-run uses state fetched via GraphQL and its result may diverge from the full-node execution results. This is mitigated by comparing transaction effects between local execution and full-node execution and reporting potential discrepancies

## Test plan 

New unit tests have been added. All tests must pass. Also tested manually


crates/sui-data-store/src/stores/graphql.rs
```diff
@@ -269,6 +269,12 @@ impl DataStore {
     pub fn node(&self) -> &Node {
         &self.node
     }
+
+    /// Return the latest checkpoint sequence number indexed by the configured GraphQL endpoint.
+    pub fn latest_checkpoint_sequence_number(&self) -> Result<u64, Error> {
```
tzakian: It seems like this generally exposes an async interface, so I'm wondering if this should be async and the call and/or somewhere else should block on the async call rather than here?

crates/sui/src/client_ptb/ptb.rs
```diff
@@ -117,6 +121,32 @@ impl PTB {
...
+                program_metadata.dry_run_set,
+                "--trace can only be used together with --dry-run."
+            );
+            ensure!(
+                !program_metadata.preview_set,
+                "--trace cannot be combined with --preview."
+            );
+            ensure!(
+                !program_metadata.dev_inspect_set,
+                "--trace cannot be combined with --dev-inspect."
```
tzakian: but..should it? what if anything prevents being able to use it with dev-inspect?

crates/sui/src/client_ptb/ptb.rs
```diff
@@ -117,6 +121,32 @@ impl PTB {
...
+            );
+            ensure!(
+                !program_metadata.tx_digest_set,
+                "--trace cannot be combined with --tx-digest."
+            );
+            ensure!(
+                !program_metadata.serialize_unsigned_set && !program_metadata.serialize_signed_set,
+                "--trace cannot be combined with transaction serialization."
+            );
```
tzakian: for these "--trace cannot be combined with X" type of error messages, would it be better to just say "--trace can only be combined with XYZ" instead -- that way when we add another flag or the like we don't need to worry about adding that here as well.

crates/sui/src/client_ptb/ast.rs
```diff
@@ -121,6 +123,8 @@ pub struct ProgramMetadata {
     pub json_set: bool,
     pub tx_digest_set: bool,
     pub dry_run_set: bool,
+    /// Whether best-effort local tracing was requested for the fullnode dry-run.
+    pub trace_set: bool,
```
tzakian, line 126: IMo no need for the comment.

crates/sui-replay-2/src/replay_txn.rs
```diff
@@ -277,7 +325,7 @@ fn verify_txn_and_save_effects(
             .member(Artifact::ForkedTransactionEffects)
             .try_remove_artifact()?;
     }
-    Ok(())
+    Ok(effects == expected_effects)
```
tzakian: nit: I'd lift this up into an `let tx_forked = ;` and have the check on line 302 use that as well.

crates/sui-replay-2/src/simulation.rs
```diff
@@ -0,0 +1,552 @@
...
+    coin_reservation::ParsedDigest,
+    digests::TransactionDigest,
+    effects::{TransactionEffects, TransactionEffectsAPI},
+    object::{MoveObject, OBJECT_START_VERSION, Object, Owner},
+    supported_protocol_versions::ProtocolConfig,
+    transaction::{TransactionData, TransactionDataAPI},
+};
+
+// Keep this in sync with `DEV_INSPECT_GAS_COIN_VALUE` in `sui-core/src/authority.rs`.
+const SIMULATION_GAS_COIN_VALUE: u64 = 1_000_000_000_000_000_000;
```
tzakian: ..we should just use that const then? It's public and we have access to `sui-core` in this crate.

crates/sui-replay-2/src/simulation.rs
```diff
@@ -0,0 +1,552 @@
...
+    );
+    Ok(())
+}
+
+/// Reconstruct the exact transaction executed by the fullnode.
+///
+/// For a request without gas payment, the fullnode executes with a deterministic synthetic gas
+/// coin while returning the original payment-free transaction data. Insert the same coin locally
+/// and require the resulting digest to match the effects; any other mismatch is rejected.
+fn reconstruct_fullnode_transaction(
```
tzakian, line 252: why not have this call `ensure_supported_transation`? That way if this function passes than you're gtg as far as checks?

crates/sui-replay-2/src/simulation.rs
```diff
@@ -0,0 +1,552 @@
...
+        digest,
+        artifact_path,
+        read_checkpoint,
+        effects_match: outcome.effects_match,
+        trace_generated: outcome.trace_generated,
+        mock_gas_id,
+    })
+}
+
+fn clear_generated_artifacts(path: &Path) -> Result<()> {
```
tzakian: Can/should this live on the artifact manager?

tzakian (commented):
Overall liking it and much nicer than the checkpoint-pinned one! 

Left some comments, but please ping on Slack with any questions or when you want me to take another pass.

crates/sui-replay-2/src/simulation.rs
awelc: Makes sense! Moved!

crates/sui-replay-2/src/simulation.rs
awelc: I did not want to pull the entire `sui-core` dependency to the replay tool's crate, but thinking about it more, the tool is largely used as part of Sui CLI anyway where this dependency is already pulled, and having the extra heft for the replay tool's dev builds should be OK. Changed!

crates/sui-replay-2/src/simulation.rs
awelc: Makes sense. The way it's structured is a historical artifact from the "base" [PR](https://github.com/MystenLabs/sui/pull/27592) where it made more sense, but really does not make sense here. Fixed!

crates/sui-replay-2/src/replay_txn.rs
awelc: Done!

crates/sui/src/client_ptb/ast.rs
awelc: OK!

crates/sui-data-store/src/stores/graphql.rs
awelc: Good point. It's not super elegant and, indirectly, also makes scheduling trace execution in `ptb.rs` a bit awkward as well. Re-worked a few things to be more idiomatic with how all this should work with tokio

crates/sui/src/client_ptb/ptb.rs
awelc: Another artifact of this being a "child" of another [implementation](https://github.com/MystenLabs/sui/pull/27592)... In this on we can indeed trace `dev-inspect` as well!

crates/sui/src/client_ptb/ptb.rs
awelc: Done!

crates/sui-replay-2/src/artifacts/manager.rs
```diff
@@ -353,3 +437,50 @@ impl ArtifactMember<'_, '_> {
         Ok(())
     }
 }
+
+#[cfg(test)]
```
tzakian, line 441: not entirely sure if we need these tests, but I guess it doesn't hurt to have them 🤷🏻

crates/sui-replay-2/src/artifacts/manager.rs
```diff
@@ -2,13 +2,16 @@
...
 use move_trace_format::format::{MoveTrace, MoveTraceReader};
 use std::{
+    fs,
     io::Write,
     path::{Path, PathBuf},
 };
-use sui_types::{effects::TransactionEffects, gas::GasUsageReport};
+use sui_types::{base_types::ObjectID, effects::TransactionEffects, gas::GasUsageReport};
+
+pub(crate) const BCODE_DIR: &str = "bytecode";
```
tzakian: Ok, you moved it so it's fair game to comment on it! 🥳 

Maybe rename this to a bit nicer name (looking at you `BYTECODE_DIR` 👀)

crates/sui-replay-2/Cargo.toml
```diff
@@ -28,6 +28,7 @@ prometheus.workspace = true
 serde.workspace = true
 similar.workspace = true
 sui-config.workspace = true
+sui-core.workspace = true
```
tzakian: oh, wow, I thought we already pulled this for replay-2 in but I was wrong!

My bad, I think in that case, revert back to what you had for the const for the dev inspect gas and remove this dep as not having a dep on sui-core is actually quite nice 🙇🏻

crates/sui-replay-2/src/artifacts/manager.rs
awelc: I'd say let's keep them as they indeed do no harm :-)

crates/sui-replay-2/src/artifacts/manager.rs
awelc: Renamed!

crates/sui-replay-2/Cargo.toml
awelc: Reverted!

PR #27752, opened 2026-08-19
refactor: consolidate ledger-history window resolution and cursor application 2/5

wlmyng:
## Description

Second of the stack (on top of the rename PR). Behavior-preserving consolidation of ledger-history window resolution and cursor application:

- `checkpoint_to_tx_range` becomes empty-safe on kv, so empty windows ride the same projection path as non-empty ones instead of a special-cased boundary lookup.
- `CheckpointRange::from_request` + `.resolve(options)` fuse into one `ResolvedCheckpointRange::from_request` call; the intermediate struct dissolves. Both backends keep their existing error precedence via a standalone `validate_checkpoint_bounds` called before read-mask/options parsing, exactly where the old validation sat — the fullnode handlers' private `ledger_read` copy of the same check is deleted in favor of the shared helper.
- Cursor application moves off `QueryOptions` onto the records, and both lanes' `apply_cursor_bounds` are restructured into per-arm `apply_after_cursor`/`apply_before_cursor` with the same shape.

## Test plan

- window past indexed tip = `LedgerTip`
- zero-width window is `CheckpointBound`
- cursors don't change already-empty `Resolved*Range`
- `resolve` orients entry/terminal checkpoints by ordering

Tests on `ResolvedRange` and `ResolvedIntraTxRange` around cursor tightening, after/before + ascending/descending.

Which cursor owns the terminal frame when both are present and applied, when one is rejected, when `before` overrides `after`.

Every commit compiles and passes the scoped suites individually.

## Release notes
- [ ] Protocol
- [ ] gRPC
- [ ] JSON-RPC
- [ ] GraphQL
- [ ] CLI
- [ ] Rust SDK

### DIFF
--- a/crates/sui-kvstore/src/bigtable/client/mod.rs
+++ b/crates/sui-kvstore/src/bigtable/client/mod.rs
@@ -1558,46 +1558,42 @@ impl BigTableClient {
         })
     }
 
-    /// Resolve inclusive checkpoint bounds to a tx_sequence_number range.
-    ///
-    /// Returns `[start_tx, end_tx)` where `start_tx` is the first tx in
-    /// `start_checkpoint` and `end_tx` is one past the last tx in `end_checkpoint`.
-    /// Reads checkpoint summaries from the checkpoints table.
+    /// Resolve a half-open checkpoint range to the tx_sequence_number range it spans. Checkpoint 0
+    /// resolves to tx_sequence_number 0. An empty checkpoint range can only be of the form [x, x)
+    /// and maps to an empty transaction range [t, t).
     pub async fn checkpoint_to_tx_range(
         &mut self,
         checkpoint_range: Range<u64>,
     ) -> Result<Range<u64>> {
         use crate::tables::checkpoints;
 
-        let start_checkpoint = checkpoint_range.start;
-        let end_checkpoint = checkpoint_range.end.saturating_sub(1);
+        let mut keys: Vec<u64> = [checkpoint_range.start, checkpoint_range.end]
+            .into_iter()
+            .filter(|&cp| cp > 0)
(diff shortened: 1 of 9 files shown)


nickvikeras (approved):

crates/sui-kvstore/src/bigtable/client/mod.rs
```diff
@@ -1558,46 +1558,42 @@ impl BigTableClient {
...
         checkpoint_range: Range<u64>,
     ) -> Result<Range<u64>> {
         use crate::tables::checkpoints;
 
-        let start_checkpoint = checkpoint_range.start;
-        let end_checkpoint = checkpoint_range.end.saturating_sub(1);
+        let mut keys: Vec<u64> = [checkpoint_range.start, checkpoint_range.end]
+            .into_iter()
+            .filter(|&cp| cp > 0)
+            .map(|cp| cp - 1)
```
amnn: Could be


```suggestion
            .filter_map(|&cp| cp.checked_sub(1))
```

crates/sui-kvstore/src/bigtable/client/mod.rs
```diff
@@ -1558,46 +1558,42 @@ impl BigTableClient {
...
-        let end_summary = end_cps
-            .first()
-            .and_then(|cp| cp.summary.as_ref())
-            .context("checkpoint summary not found for end bound")?;
-        let end_tx = end_summary.network_total_transactions;
+        let boundary = |cp: u64| -> Result<u64> {
+            if cp == 0 {
+                return Ok(0);
+            }
+            let lookup_cp = cp - 1;
```
amnn: Same deal, checked sub.

PR #27755, opened 2026-08-19
[sui-rpc-api] symbolic resume for the scalar lanes 5/5

wlmyng:
## Description

While applying an after cursor, the intra-tx path converts an after-Item cursor into `Excluded(coordinate)`, while the scalar u64 path eagerly computes `Included(coordinate + 1)`. This PR makes both uniformly symbolic, and pushes the `+1`/`u64::MAX` handling to `ScanBounds<u64>::to_range`. This shrinks the `ScanCursor` trait to `coordinate()`.

In practice, the fetch interval is invariant: `to_range` projects `Excluded(n)` and `Included(n + 1)` to the same `Range<u64>`. However, the bounds are also read by two frame decisions, which now see the raw coordinate instead of the successor:

1. `ScanBounds::is_empty` checks whether a cursor emptied the window and thus should set  `CursorBound`. `Included(n + 1)` vs `Excluded(n + 1)` reads as empty, but `Excluded(n)` vs `Excluded(n + 1)` does not, so exactly-adjacent bounds no longer set  `CursorBound` eagerly at resolution — the zero-row drain produces the frame instead. This is the current event/intra-tx behavior.
2. The admission comparison `lower_bound_gte` used to see `Included(n + 1)`, which ties a window bound at the same coordinate and admits the cursor; `Excluded(n)` ranks below the tie, so a cursor coinciding with the window's own bound is now rejected outright.

Wire-visible changes outlined below, over window `[10, 20)`:

| scenario | old → new | via |
|---|---|---|
| asc, after = Item@19 (window's last item) | CursorBound at the request cursor → CheckpointBound at the window end | 1 |
| desc, after = Item@9 (one below window) | CursorBound at the cursor → the window's natural frame | 2 |
| desc, after = Item@14, before = Item@15 | frame at before@15 → after@14 (the before's emptying claim no longer fires, so the after's stop-side stamp stands) | 1 |
| desc, after = Item@19 | unchanged — the stop-side stamp never asks about emptiness | — |
| asc, after = Item@14, before = Item@15 | unchanged — the asc stop side (before) stamps and claims the same coordinate | — |

Beyond-window cursors still empty at resolution, and `u64::MAX` handling is unchanged
(the saturation now lives in `to_range`).

## Test plan

- `to_range_collapses_excluded_lo_at_successor`
- `raw_after_item_echo_resumes_like_successor_boundary`
- `tx_after_item_at_window_edge_defers_to_drain`
- `tx_after_item_at_window_edge_descending_keeps_stamped_terminal`
- `tx_after_item_below_window_start_descending_is_noop`
- `test_list_transactions_after_before_adjacent_empty_results` (pre-flip preservation pin)
- `test_list_transactions_after_last_item_ends_at_window_bound` (`cargo simtest -p sui-e2e-tests list_ledger_history`)

## Release notes
- [ ] Protocol
- [ ] gRPC
- [ ] JSON-RPC
- [ ] GraphQL
- [ ] CLI
- [ ] Rust SDK

### DIFF
--- a/crates/sui-e2e-tests/tests/rpc/v2/ledger_service/list_ledger_history.rs
+++ b/crates/sui-e2e-tests/tests/rpc/v2/ledger_service/list_ledger_history.rs
@@ -992,6 +992,50 @@ async fn test_list_transactions_unfiltered_and_sender_filter() {
     );
 }
 
+/// (after = item n, before = item n + 1) serves nothing and ends with a CursorBound frame from the
+/// ascending stop side (before cursor.)
+#[sim_test]
+async fn test_list_transactions_after_before_adjacent_empty_results() {
+    let cluster = new_cluster().await;
+    let sender = cluster.get_address_0();
+    let tx = transfer_self(&cluster, sender).await;
+    let (start, end) = checkpoint_range(&[&tx]);
+
+    let mut client = new_ledger_client(&cluster).await;
+
+    let mut req = ListTransactionsRequest::default();
+    req.read_mask = Some(FieldMask::from_paths(["digest"]));
+    req.start_checkpoint = Some(start);
+    req.end_checkpoint = Some(end);
+    req.options = Some(query_options(100));
+    let baseline = list_transactions_result(&mut client, req).await;
+    assert!(baseline.end);
+    assert!(
+        baseline.transactions.len() >= 2,
+        "need at least two transactions for adjacency windows"
+    );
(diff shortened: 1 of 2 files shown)


nickvikeras:
looks good to me. Thanks again for cleaning this up. Once you have handled the empty-range discrepancy we discussed over slack lmk and I'll take one last look.

PR #27756, opened 2026-08-19
[graphql-stream-metrics] Add per-subscriber subscription metrics [22/n]

tpham-mysten:
## Description

Builds on the upstream-stream metrics from #27669 (1/n) with the per-subscriber view. A `SubscriptionLifecycleGuard` (RAII, held for the lifetime of each subscription stream) records, by subscription type, how many are active, how many opened, how long they live, and how and why they end. It also records per-payload delivery: a delivered counter and a server-side delivery-lag histogram, observed once per payload in the live loop after each yield (so the lag reflects that payload's own throttle pacing).

Metrics added, all labelled by subscription type:

- `graphql_subscription_active_subscriptions` (gauge)
- `graphql_subscription_opened` (counter)
- `graphql_subscription_terminations` (counter, by reason)
- `graphql_subscription_duration_seconds` (histogram)
- `graphql_subscription_payloads_delivered` (counter)
- `graphql_subscription_payload_delivery_checkpoint_timestamp_lag` (histogram)

## Test plan

Unit tests for the guard: the active gauge and opened counter on construction, and on drop the gauge decrement, a duration sample, and a `client_closed` termination when no reason was set; a set reason overriding that default; and per-payload delivery counting plus one lag observation per call.

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
@@ -42,6 +42,8 @@ use crate::task::streaming::CheckpointBroadcaster;
 use crate::task::streaming::ProcessedCheckpoint;
 use crate::task::streaming::StreamedCaches;
 use crate::task::streaming::SubscriptionBroadcast;
+use crate::task::streaming::SubscriptionLifecycleGuard;
+use crate::task::streaming::SubscriptionTerminationReason;
 use crate::task::streaming::broadcast_error;
 use crate::task::streaming::reconnect_error;
 use crate::task::streaming::wait_for_pipelines_catching_up_at;
@@ -55,6 +57,9 @@ const BACKFILL_POLL_INTERVAL: Duration = Duration::from_millis(100);
 /// GraphQL type (transactions, events) supplies its cursor, filter, and matching, and the free
 /// functions here drive the shared machinery over it.
 pub(super) trait Subscribable {
+    /// The subscription kind, used to label this feed's aggregate subscriber metrics.
+    const SUBSCRIPTION_TYPE: &'static str;
+
     /// The GraphQL node delivered in each edge.
     type Item: OutputType + Send + 'static;
     /// The opaque cursor minted for resumption.
@@ -137,6 +142,7 @@ pub(super) fn subscribe<S: Subscribable>(
     let handoff_threshold = config.broadcast_buffer as u64 / 2;
 
     stream! {
+        let guard = SubscriptionLifecycleGuard::new(S::SUBSCRIPTION_TYPE, broadcast.metrics());
         let mut pending_receiver = None;
         let mut handoff: Option<u64> = None;
         let mut last_checkpoint: Option<u64> = None;
@@ -159,6 +165,7 @@ pub(super) fn subscribe<S: Subscribable>(
                 let scanned = match scanned {
                     Ok(scanned) => scanned,
                     Err(e) => {
+                        guard.terminate(SubscriptionTerminationReason::BackfillError);
                         yield Err(e);
                         return;
                     }
@@ -179,6 +186,7 @@ pub(super) fn subscribe<S: Subscribable>(
                             break;
                         }
                         yield Ok(edge);
+                        guard.record_backfill_delivered();
                     }
                     // A coverage marker (its `checkpoint` is the fully-scanned frontier): stop once
                     // it has covered the handoff.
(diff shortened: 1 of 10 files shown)


crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
```diff
@@ -83,9 +85,10 @@ impl Subscription {
         let caches = caches.clone();
         let resolver_limits = limits.package_resolver();
 
-        let stream = broadcast
-            .clone()
-            .subscribe(start_from, fetcher.clone(), config);
+        let stream =
+            broadcast
+                .clone()
+                .subscribe(start_from, fetcher.clone(), config, metrics.clone());
```
amnn: Can we initialize `SubscriptionBroadcast` with its metrics instead? Saves passing it through every time.

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -159,6 +168,7 @@ pub(super) fn subscribe<S: Subscribable>(
                 let scanned = match scanned {
                     Ok(scanned) => scanned,
                     Err(e) => {
+                        guard.set_termination_reason(SubscriptionTerminationReason::BackfillError);
```
amnn: Rather than offering a way to set the termination reason to be handled when the `guard` drops, we could consider providing a `terminate` function which consumes the guard early? Should work in most cases here.

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -289,26 +303,34 @@ fn live<S: Subscribable>(
...
                             return;
                         }
                     }
                     // Deliver each matching item as its own payload, ordered within the checkpoint.
                     // Empty checkpoints yield nothing.
-                    let edges = S::matching_edges(&checkpoint, &caches, &resolver_limits, &filter)?;
+                    let edges = S::matching_edges(&checkpoint, &caches, &resolver_limits, &filter)
+                        .inspect_err(|_| {
+                            guard.set_termination_reason(SubscriptionTerminationReason::Error);
+                        })?;
```
amnn: This is the only case where the `terminate` suggestion might complicate things, but it can be addressed like so:


```suggestion
                    let edges = match S::matching_edges(&checkpoint, &caches, &resolver_limits, &filter) {
                        Ok(edges) => edges,
                        Err(e) => {
                            guard.terminate(SubscriptionTerminationReason::Error);
                            return Err(e);
                        
                    };
```

crates/sui-indexer-alt-graphql/src/task/streaming/lifecycle.rs
```diff
@@ -0,0 +1,230 @@
...
+use prometheus::IntCounter;
+use prometheus::IntCounterVec;
+use prometheus::IntGauge;
+use tokio::sync::broadcast;
+use tokio::time::Instant;
+
+use crate::metrics::SubscriptionMetrics;
+
+/// The kind of subscription, used to label the aggregate subscriber metrics.
+#[derive(Clone, Copy, strum::IntoStaticStr)]
```
amnn: Consider making the associated constant a `&'static str` to avoid quite a bit of machinery/boilerplate.

crates/sui-indexer-alt-graphql/src/lib.rs
```diff
@@ -456,7 +462,8 @@ pub async fn start_rpc(
         .data(consistent_reader)
         .data(pg_loader)
         .data(kv_loader)
-        .data(package_store);
+        .data(package_store)
+        .data(subscription_metrics);
```
amnn: If `SubscriptionBroadcast` owns metrics can we drop this from the context?

amnn (approved):
Could look into removing the dependency on `strum`, but looks good.

crates/sui-indexer-alt-graphql/src/task/streaming/lifecycle.rs
tpham-mysten: Sure done

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: SG. Done

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: Yeah, it's more complex but not end of the world tho.

crates/sui-indexer-alt-graphql/src/api/subscription/mod.rs
tpham-mysten: Sure we can!

crates/sui-indexer-alt-graphql/src/lib.rs
tpham-mysten: Yes we can

tpham-mysten (commented):
Thanks @amnn! I have dropped `strum` and moved `SubscriptionMetrics` into the broadcast.

PR #27759, opened 2026-08-19
Tighten gc_round invariant checks in FlexCommitter's build_commit

lexfrl:
## Description 

`build_commit`'s gc_round check ran once, in aggregate, after the DFS finished. This moves the leader check to before `dag_state` is mutated, drops the now-redundant aggregate check, and folds the ancestor `is_committed` pre-check into `set_committed`'s existing check-and-set.

## Test plan 

Added `try_commit_multi_leader_shared_ancestor_visited_once`, covering shared ancestors discovered by multiple leaders within one `build_commit` call — not covered by existing tests, whose shared ancestors are always genesis blocks excluded before `dag_state` is touched. Full `consensus-core` suite passes (337 tests).

### DIFF
--- a/consensus/core/src/flex_committer.rs
+++ b/consensus/core/src/flex_committer.rs
@@ -332,7 +332,9 @@ impl FlexCommitter {
             }
             committed_leaders
         };
-        assert!(!committed_leaders.is_empty(), "No committed leaders found");
+        if committed_leaders.is_empty() {
+            panic!("No committed leaders found");
+        }
         self.report_decided_prefix_metrics(commit_leader_round);
 
         let mut dag_state = self.dag_state.write();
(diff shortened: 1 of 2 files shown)


consensus/core/src/flex_committer.rs
```diff
@@ -332,7 +332,9 @@ impl FlexCommitter {
             }
             committed_leaders
         };
-        assert!(!committed_leaders.is_empty(), "No committed leaders found");
+        if committed_leaders.is_empty() {
```
mwtian: Can we revert this change?

mwtian (commented):
lgtm except one comment!

consensus/core/src/flex_committer.rs
lexfrl: Sure! I just got a little stuck on this assert while reading - it seemed like a double negative to me

mwtian (approved):

PR #27761, opened 2026-08-19
[Block Streaming] subscribed peer fail over

akichidis:
## Description 

When a node is configured as observer client a list of peers is allowed to be provided. At the moment the first one is always being used for the direct connection to stream blocks. The rest are used for live and commit block synchronisation. 

This PR is changing the mechanics so in case of failure during block streaming, the node is not attempting to reconnect only to the first peer, but will try to connect to the rest ones - via a round robin protocol. 

This is particularly important especially for a full node that is consensus-sync enabled as there is strong dependency on the observer peer server. Although the full node will be able to make state sync progress, as it keeps checkpoint sync, it will effectively become slower.

**Edge case**
The new peer that is switched over might be either quite ahead in quorum, or behind.

If it's too far ahead, commit sync will step in and quickly fill in the gaps
If it's too far behind, then there might be a block delivery overlap and this will eventually normalise 

If the new peer serves blocks with missing causal history for the client, then the client will need to sync them via the scheduled synchronizer. If this doesn't matter for the commit progress, then GC steps in and cuts off the synchronization stress that might be put it.

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/consensus/core/src/observer_subscriber.rs
+++ b/consensus/core/src/observer_subscriber.rs
@@ -13,7 +13,7 @@ use parking_lot::Mutex;
 use tokio::{
     sync::oneshot,
     task::{JoinError, JoinHandle, JoinSet},
-    time::{Instant, sleep, sleep_until, timeout},
+    time::{Instant, sleep, timeout},
 };
 use tracing::{debug, info, warn};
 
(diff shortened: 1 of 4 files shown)


consensus/core/src/observer_subscriber.rs
```diff
@@ -308,15 +399,34 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
                 let _scope = monitored_scope("ObserverSubscriberStreamConsumer");
 
                 let next_item = tokio::select! {
+                    biased;
```
mwtian, line 402: I had ran into unexpected behavior with biased, which you may be able to find from git history. YMMV.

mwtian (approved):

consensus/core/src/observer_subscriber.rs
```diff
@@ -308,15 +399,34 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
...
+                                retries += 1;
+                                break 'stream;
+                            }
+                        }
+                    }
                 };
 
                 match next_item {
                     Some(item) => {
+                        last_progress = Instant::now();
```
mwtian, line 429: [P1] Do not count auxiliary-only items as block progress

ObserverService merges randomness signatures into this stream as items with an empty `blocks` vector. An observer can relay these signature items while its local block and commit path has stopped. Each item completes the timeout, and this line resets `last_progress`. The subscriber can therefore stay on this peer forever, although it receives no blocks. Signature items do not update CommitVoteMonitor, so commit sync cannot learn a later quorum commit index. Please reset block progress only after a non-empty block batch, preferably after an accepted new block, and add a signature-only test.

consensus/core/src/observer_subscriber.rs
```diff
@@ -308,15 +399,34 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
...
                         Self::handle_task_result(result);
                         continue 'stream;
                     }
-                    item = blocks.next(), if tasks.len() < max_parallel_tasks => item,
+                    // The stream is only polled (and the stall timer only runs) while there is
+                    // handler capacity: a quiet stream under handler backpressure is a local
+                    // processing problem, not a peer stall. `timeout` polls the stream before
+                    // checking the timer, so an already-buffered item wins over an expired
+                    // deadline.
+                    item = timeout(STREAM_STALL_TIMEOUT, blocks.next()), if tasks.len() < max_parallel_tasks => {
```
mwtian, line 412: [P2] Keep the stall deadline across handler completions

This timeout is created inside the select loop. If `tasks.join_next()` wins, `continue` drops this timeout, and the next loop starts a new 60 s timeout. A batch can leave several handlers in progress. If those handlers finish less than 60 s apart, they can delay detection of a silent stream by almost 60 s for each handler. The biased select makes this result deterministic when both branches are ready. Please keep one deadline outside this loop. Reset it only on real block progress. If local backpressure must pause the deadline, pause it only while the handler set is full.

consensus/core/src/observer_subscriber.rs
```diff
@@ -244,8 +270,33 @@ impl<C: ObserverNetworkClient, S: ObserverNetworkService> ObserverSubscriber<C,
...
+        let mut peer_index: usize = 0;
+        // When the current peer last delivered a block (or became current), shifted forward
+        // past commit-lag suspensions so suspended time does not count against the peer.
+        let mut last_progress = Instant::now();
 
         'subscription: loop {
+            // Rotate to the next configured peer when the current one has delivered nothing
+            // for the whole failure budget despite the retries below. Sticky: rotation only
+            // moves forward on failure, never back to an earlier peer that recovered.
+            if peers.len() > 1 && last_progress.elapsed() >= STREAM_STALL_TIMEOUT {
```
mwtian, line 282: [P2] Enforce one peer-wide failure budget

This check runs only before a retry. A retry that starts just before 60 s can still wait about 10 s for backoff, 30 s to establish the stream, and 60 s for the first item. The next rotation check can then occur about 160 s after the last progress. The existing unreachable-peer test does not assert the rotation time and already exceeds 60 s under its virtual clock. Please use one absolute peer deadline and limit each backoff, setup, and stream wait to the remaining time. Continue to move the deadline forward only for the explicit commit-lag suspension.

PR #27767, opened 2026-08-20
[move][move-compiler][sui-lint] Add a lint to match the new PTB restrictions

cgswords:
## Description

Follow-up to #27451: the `uncallable_function` lint already warns on the parameter-side `TxContext` rules (by-value, duplicate `&mut`, `&mut` mixed with `&`), so this adds the missing return-position rule.

## Test plan

New true-positive, true-negative, and suppression cases in the `uncallable_functions` linter tests.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/external-crates/move/crates/move-compiler/src/sui_mode/linters/uncallable_function.rs
+++ b/external-crates/move/crates/move-compiler/src/sui_mode/linters/uncallable_function.rs
@@ -122,7 +122,7 @@ simple_visitor!(
         if matches!(&fdef.visibility, Visibility::Public(_)) || fdef.entry.is_some() {
             const RETURN_NOTE: &str = "Due to restrictions in PTB execution, 'TxContext' may \
                 never appear in the return type of a function callable from a transaction, by \
-                value or by reference. This function will not be callable on Sui";
+                value or by reference. This function will not be callable from PTBs on Sui";
             for ret_ty in return_position_types(&signature.return_type) {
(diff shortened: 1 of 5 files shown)


external-crates/move/crates/move-compiler/src/sui_mode/linters/uncallable_function.rs
```diff
@@ -114,6 +115,40 @@ simple_visitor!(
...
             }
         }
+
+        // `TxContext` can never appear in return position for a function callable from a
+        // transaction. Only public and entry functions can be called that way; private helpers
+        // returning references derived from a `TxContext` parameter are fine
+        if matches!(&fdef.visibility, Visibility::Public(_)) || fdef.entry.is_some() {
+            const RETURN_NOTE: &str = "Due to restrictions in PTB execution, 'TxContext' may \
+                never appear in the return type of a function callable from a transaction, by \
+                value or by reference. This function will not be callable on Sui";
```
tnowacki:
```suggestion
                value or by reference. This function will not be callable from PTBs on Sui";
```
We should soften the language a bit, at least for public functions. Since you can call them, just only from other Move functions

tnowacki (approved):
LGTM, just a bit of massaging for the language maybe

PR #27771, opened 2026-08-20
graphql: serve balanceChanges purely from the ledger gRPC transaction lookup [1/3 remove Postgres `tx_balance_changes`]

wlmyng:
## Description

Since #27653 made ledger gRPC the only KV path, and `tx_balance_changes` was only used on graphql as a kv lookup for balance changes, the `Postgres` path is unreachable. Strip this out, along with the `Stored`/`Native` balance-change variants. Stop registering the `tx_balance_changes` pipeline.

## Test plan

No behavior change (the deleted fallback was unreachable); covered by existing graphql e2e balance-change tests and the pipeline-mapping snapshot test.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/types/balance_change.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/balance_change.rs
@@ -17,12 +17,6 @@ pub(crate) struct BalanceChange {
     pub(crate) content: GrpcBalanceChange,
 }
 
-impl BalanceChange {
-    pub(crate) fn from_grpc(scope: Scope, content: GrpcBalanceChange) -> Self {
-        Self { scope, content }
-    }
-}
-
 /// Effects to the balance (sum of coin values per coin type) of addresses and objects.
 #[Object]
 impl BalanceChange {
--- a/crates/sui-indexer-alt-graphql/src/api/types/balance_change.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/types/balance_change.rs
@@ -1,16 +1,9 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::str::FromStr;
-
 use anyhow::Context as _;
 use async_graphql::Object;
-use sui_indexer_alt_reader::kv_loader::BalanceChangeContents as KvBalanceChangeContents;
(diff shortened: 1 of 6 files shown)


crates/sui-indexer-alt-reader/src/kv_loader.rs
```diff
@@ -380,20 +372,10 @@ impl TransactionContents {
...
-                    .map(|c| BalanceChangeContents::Grpc(c.clone()))
-                    .collect(),
-            ),
-            Self::LedgerGrpc(txn) => Some(
-                txn.balance_changes
-                    .iter()
-                    .map(|c| BalanceChangeContents::Grpc(c.clone()))
-                    .collect(),
-            ),
+            Self::ExecutedTransaction(tx) => tx.balance_changes.clone(),
```
tpham-mysten: nit: is there a way to not clone here as it's just returning an internal reference for both?

crates/sui-indexer-alt-graphql/src/api/types/balance_change.rs
```diff
@@ -1,58 +1,25 @@
...
-            KvBalanceChangeContents::Grpc(grpc) => Self {
-                scope,
-                content: BalanceChangeContents::Grpc(grpc),
-            },
-            KvBalanceChangeContents::Native(native) => Self {
-                scope,
-                content: BalanceChangeContents::Native(native),
-            },
-        }
+    pub(crate) fn from_grpc(scope: Scope, content: GrpcBalanceChange) -> Self {
```
tpham-mysten: nit: not sure if we still need this instead of constructing BalanceChange directly

tpham-mysten (approved):
Looking very neat now!

crates/sui-indexer-alt-reader/src/kv_loader.rs
wlmyng: yeah we can return a slice instead of the vec

crates/sui-indexer-alt-graphql/src/api/types/balance_change.rs
wlmyng: fair enough

PR #27772, opened 2026-08-20
jsonrpc-alt: serve balance changes from the ledger gRPC transaction lookup [2/3 remove `tx_balance_changes`]

wlmyng:
## Description

`getTransactionBlock` already fetches the transaction through ledger gRPC, which returns balance changes alongside it, so the separate Postgres `tx_balance_changes` lookup is redundant. Convert the protos to `SuiBalanceChange` the same way the write path does. Since balance changes now always accompany a found transaction, `BalanceChangesNotFound` (and its retry, which covered the pg pipeline lagging the kv store) is unreachable and removed.

Stacked on #27771 (2/3).

## Test plan

Covered by existing e2e tests in `sui-indexer-alt-e2e-tests/tests/jsonrpc/transactions`.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-jsonrpc/src/api/transactions/error.rs
+++ b/crates/sui-indexer-alt-jsonrpc/src/api/transactions/error.rs
@@ -15,9 +15,6 @@ pub(super) enum Error {
     #[error("Pagination issue: {0}")]
     Pagination(#[from] crate::paginate::Error),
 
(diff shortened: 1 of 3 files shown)


tpham-mysten (approved):
Nice!

PR #27776, opened 2026-08-20
sui-indexer-alt-graphql: Replace avoidable heap allocations with stack allocation

evan-wall-mysten:
## Description

A sweep of `sui-indexer-alt-graphql` for heap allocations that could be stack allocations turned up two patterns:

- Several places built a `Vec<u8>` from a fixed-size `SuiAddress` just to bind it into a query (a Diesel `.eq()` filter, and several raw-SQL `{Bytea}` binds via the `sui_sql_macro::query!` macro) — Diesel already supports `[u8; N]` directly for `Binary`/`Bytea` columns, so these can pass the address as a stack array instead. This removed the GraphQL `SuiAddress` scalar's `into_vec()` entirely, since all 9 of its call sites fit this pattern; replaced with a `to_inner()` accessor mirroring the one already on `sui_types::base_types::SuiAddress`.
- `package_eviction_task.rs` used `Box::pin` on a stream that's created once, never moved, and never shared — `pin!` produces the same `Pin<&mut _>` on the stack instead.

## Test plan

Covered by existing tests in `sui-indexer-alt-graphql` (228 passing); all call sites are otherwise behavior-identical, verified via `cargo check`/`clippy`.

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/api/scalars/sui_address.rs
+++ b/crates/sui-indexer-alt-graphql/src/api/scalars/sui_address.rs
@@ -55,8 +55,8 @@ impl ScalarType for SuiAddress {
 impl SuiAddress {
     pub(crate) const ZERO: Self = Self([0u8; SUI_ADDRESS_LENGTH]);
 
-    pub fn into_vec(self) -> Vec<u8> {
-        self.0.to_vec()
+    pub(crate) fn to_inner(self) -> [u8; SUI_ADDRESS_LENGTH] {
+        self.0
     }
 }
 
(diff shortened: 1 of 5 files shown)


amnn (approved):
LGTM, but I suspect this is not going to make a huge dent in our allocation pressure.

PR #27782, opened 2026-08-20
mysten-metrics, mysten-common, sui-indexer-alt-metrics: Replace avoidable heap allocations with stack allocation

evan-wall-mysten:
## Description

Continuing the dependency sweep from #27780/#27781.

- `InflightGuardFuture` (mysten-metrics) and `MonitoredScopeFuture` (mysten-metrics) each wrapped their inner future in `Box::pin` purely to make `Self: Unpin`, even though the wrapped future is only ever polled in place by the wrapper's own `Future` impl and never moved after construction. Switched both to `pin-project-lite`'s structural pinning, matching the identical pattern already used in `mysten-network`'s `RequestLogBody`. Verified against every call site in the repo (`sui-core`'s consensus-submission path, `sui-benchmark`'s driver).
- `NotifyRead`'s shard table (mysten-common) is a `Vec` built from a hardcoded `0..255` range; the shard count is a compile-time constant, so it's now `[Mutex<HashMap<K, Registrations<V>>>; 255]` via `std::array::from_fn`.
- `DbConnectionStatsCollector::desc` (sui-indexer-alt-metrics) is a `Vec` literal that always holds exactly the same 7 metric descriptors; changed to a fixed-size array.

## Test plan

Covered by existing tests (mysten-metrics/mysten-common combined suite, 58 passing); downstream consumers (`sui-core`, `sui-benchmark`) verified to still compile. All changes verified via `cargo check`/`clippy --all-targets -D warnings` across the changed crates plus their consumers.

## Release notes

Check each box that your changes affect.

- [ ] Protocol:
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9545,6 +9545,7 @@ dependencies = [
  "futures",
  "once_cell",
  "parking_lot 0.12.3",
+ "pin-project-lite",
  "prometheus",
  "prometheus-closure-metric",
  "scopeguard",
--- a/crates/mysten-common/src/sync/notify_read.rs
+++ b/crates/mysten-common/src/sync/notify_read.rs
@@ -67,8 +67,10 @@ const MAX_SAMPLED_KEYS: usize = 32;
 pub const CHECKPOINT_BUILDER_NOTIFY_READ_TASK_NAME: &str =
     "CheckpointBuilder::notify_read_executed_effects";
 
+const NUM_SHARDS: usize = 255;
+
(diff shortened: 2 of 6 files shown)


evan-wall-mysten:
Splitting into 3 separate PRs, one per crate, per review request.

PR #27787, opened 2026-08-21
[graphql-stream] Cap concurrent subscriptions with a configurable max_subscribers limit [23/n]

tpham-mysten:
## Description

Subscriptions currently open without any aggregate limit, so a burst of subscribers can drive server CPU and memory unbounded. This adds a configurable `max_subscribers` (default 1024): each subscription claims a slot from a shared semaphore held for its lifetime by the lifecycle guard, and a subscription opened while all slots are taken is refused with a `resource_exhausted` error so the client can retry later. Refusals are counted by `graphql_subscription_rejected{type, reason}`.

Guard creation moves out of the driver `stream!` bodies up to the resolver, so admission happens once at connect time before any work begins.

## Test plan

New `rejects_at_capacity` unit test covers admission, the at-capacity refusal and its metric, and slot release on drop. Existing subscription lib tests cover the guard threaded through the drivers.

## Stack
   - #26019
   - #26094
   - #26117
   - #26170
   - #26194
   - #26202
   - #26414
   - #26453
   - #26476
   - #26487
   - #26425
   - #26495
   - #26731
   - #26714
   - #27140
   - #27537
   - #27564
   - #27580
   - #27635
   - #27709
   - #27669 
   - #27756 
   - #27787

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/sui-indexer-alt-graphql/src/task/streaming/lifecycle.rs
+++ b/crates/sui-indexer-alt-graphql/src/task/streaming/lifecycle.rs
@@ -1,14 +1,38 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::sync::Arc;
+
 use prometheus::Histogram;
 use prometheus::IntCounter;
 use prometheus::IntCounterVec;
 use prometheus::IntGauge;
+use tokio::sync::OwnedSemaphorePermit;
+use tokio::sync::Semaphore;
 use tokio::sync::broadcast;
 use tokio::time::Instant;
 
+use crate::error::RpcError;
+use crate::error::resource_exhausted;
 use crate::metrics::SubscriptionMetrics;
+
+/// Server-wide cap on the number of concurrent subscriptions, as a pool of admission slots.
+/// Registered once in the schema data; each admitted subscription's guard holds one slot for its
+/// lifetime, and a subscription opened while all slots are taken is rejected.
+#[derive(Clone)]
+pub(crate) struct SubscriberLimit(Arc<Semaphore>);
+
+impl SubscriberLimit {
+    pub(crate) fn new(max_subscribers: usize) -> Self {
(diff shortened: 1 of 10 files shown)


crates/sui-indexer-alt-graphql/src/task/streaming/lifecycle.rs
```diff
@@ -72,18 +96,36 @@ pub(crate) struct SubscriptionLifecycleGuard {
...
+    /// Admit a subscription: claim a concurrency slot from `subscriber_limit`, or return an
+    /// at-capacity error (counting a rejection) when the server is already full. The returned guard
+    /// holds the slot and all per-subscriber metric handles for the subscription's lifetime.
+    pub(crate) fn new(
+        subscription_type: SubscriptionType,
+        metrics: &SubscriptionMetrics,
+        subscriber_limit: &SubscriberLimit,
+    ) -> Result<Self, RpcError> {
         let label = subscription_type.metric_label();
+        let Some(permit) = subscriber_limit.try_acquire() else {
```
amnn, line 98: Should we allow a request to wait for some short period of time to get in? Otherwise people will just hammer the service with repeated requests.

amnn (approved):
Would consider allowing requests to wait a little bit for a semaphore token, but broadly looks good, thanks @tpham-mysten

tpham-mysten:
Thanks @amnn! 

Doing this on the GraphQL server side is probably not the right place, mainly because it's tricky to decide which request we should hold. From the infra side, I think the per-IP rate limit should be partially sufficient to prevent it.

amnn:
I think if we take the view that we should not be hitting this limit under normal operation, and we can rely on health checks and auto-scaling, then I can get behind this (but then we should probably set a more generous upper limit at this layer and leave the more conservative limit for the gateway to enforce).

PR #27788, opened 2026-08-21
[7/n] allowances: deny-list checks cover the funder

manolisliolios:
## Description

Add cross-owner checks for denylist. That'd cover both sponsor + allowances spend checks

## Test plan

New transactional tests

---

## Release notes

- [ ] Protocol:
- [ ] Nodes:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-adapter-transactional-tests/tests/deny_list_v2/allowance_deny_combinations.move
+++ b/crates/sui-adapter-transactional-tests/tests/deny_list_v2/allowance_deny_combinations.move
@@ -1,0 +1,209 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Signing checks every address a regulated coin is debited from, per coin type:
+// the sender for everything, plus the funder of each allowance withdrawal. A
+// global pause on the coin type covers allowance withdrawals like anything else.
+// A: issuer, holds both deny caps. B, D: funders. C: spender and sender.
+
+//# init --accounts A B C D --addresses test=0x0
+
+//# publish --sender A
+module test::regulated_coin1 {
+    use sui::coin;
+
+    public struct REGULATED_COIN1 has drop {}
+
+    #[allow(deprecated_usage)]
+    fun init(otw: REGULATED_COIN1, ctx: &mut TxContext) {
+        let (mut treasury_cap, deny_cap, metadata) = coin::create_regulated_currency_v2(
+            otw,
+            9,
+            b"RC",
+            b"REGULATED_COIN",
+            b"A new regulated coin",
+            option::none(),
+            true,
+            ctx
+        );
+        let coin = coin::mint(&mut treasury_cap, 10000, ctx);
+        transfer::public_transfer(coin, tx_context::sender(ctx));
+        transfer::public_transfer(deny_cap, tx_context::sender(ctx));
+        transfer::public_freeze_object(treasury_cap);
+        transfer::public_freeze_object(metadata);
+    }
+}
+
+module test::regulated_coin2 {
+    use sui::coin;
+
+    public struct REGULATED_COIN2 has drop {}
+
+    #[allow(deprecated_usage)]
+    fun init(otw: REGULATED_COIN2, ctx: &mut TxContext) {
+        let (mut treasury_cap, deny_cap, metadata) = coin::create_regulated_currency_v2(
+            otw,
+            9,
+            b"RC",
+            b"REGULATED_COIN",
+            b"A new regulated coin",
+            option::none(),
+            false,
+            ctx
+        );
+        let coin = coin::mint(&mut treasury_cap, 10000, ctx);
+        transfer::public_transfer(coin, tx_context::sender(ctx));
+        transfer::public_transfer(deny_cap, tx_context::sender(ctx));
+        transfer::public_freeze_object(treasury_cap);
+        transfer::public_freeze_object(metadata);
+    }
+}
+
+//# programmable --sender A --inputs object(1,5) object(1,10) 1000 @B @C @D
+// Fund address balances: RC1 for B, C and D; RC2 for B.
+//> 0: SplitCoins(Input(0), [Input(2), Input(2), Input(2)]);
+//> 1: sui::coin::send_funds<test::regulated_coin1::REGULATED_COIN1>(NestedResult(0,0), Input(3));
+//> 2: sui::coin::send_funds<test::regulated_coin1::REGULATED_COIN1>(NestedResult(0,1), Input(4));
+//> 3: sui::coin::send_funds<test::regulated_coin1::REGULATED_COIN1>(NestedResult(0,2), Input(5));
+//> 4: SplitCoins(Input(1), [Input(2)]);
+//> 5: sui::coin::send_funds<test::regulated_coin2::REGULATED_COIN2>(Result(4), Input(3));
+
+//# create-checkpoint
+
+//# programmable --sender B --inputs b"" @C vector[1000u256] vector[] vector[99999999999999]
+// B issues an RC1 allowance to C.
+//> 0: std::option::none<sui::allowance::RateLimit>();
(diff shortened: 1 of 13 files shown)


crates/sui-types/src/deny_list_v1.rs
```diff
@@ -95,21 +125,25 @@ pub(crate) fn input_object_coin_types_for_denylist_check(
...
-    ) else {
-        return Ok(());
-    };
-    if count == 0 {
+    let addresses: BTreeSet<SuiAddress> =
+        addresses_by_coin_type.values().flatten().copied().collect();
+
+    // TODO: Add a limit of addresses? Our current boundary is 10 (due to max withdrawals)
+    // but might be worth making this explicit.
+    let denied_anywhere: BTreeSet<SuiAddress> = addresses
```
tzakian, line 152: At least worth adding a debug assert to document this.

crates/sui-types/src/deny_list_v1.rs
```diff
@@ -95,21 +125,25 @@ pub(crate) fn input_object_coin_types_for_denylist_check(
 
 fn check_deny_list_v1_impl(
     deny_list: PerTypeDenyList,
-    address: SuiAddress,
-    coin_types: BTreeSet<String>,
+    addresses_by_coin_type: BTreeMap<String, BTreeSet<SuiAddress>>,
```
tzakian, line 146: The logic of this function gets weird I believe if all of the `BTreeSet` values are all empty (for a non-empty map).

IMO it is worth at the very least adding a debug assert that at least one set is non-empty (or if I'm reading this right, you should be able to state that all BTreeSet's are non-empty). If you can raise an invariant violation or the like if this ever is the case that's even better than the debug assert (IMO).

crates/sui-types/src/deny_list_v2.rs
```diff
@@ -115,24 +115,28 @@ impl MoveTypeTagTrait for GlobalPauseKey {
...
-    coin_types.extend(funds_withdraw_types);
-
-    for coin_type in coin_types {
+    let addresses_by_coin_type = addresses_by_coin_type_for_denylist_check(
+        sender,
+        input_objects,
+        receiving_objects,
+        funds_withdrawals,
+    );
+    for (coin_type, addresses) in addresses_by_coin_type {
```
tzakian, line 130: Ditto here above this line -- debug assert/invariant violate if any `addresses` is empty I think

crates/sui-types/src/deny_list_v1.rs
```diff
@@ -48,17 +48,22 @@ pub struct PerTypeDenyList {
 }
 
 /// Checks coin denylist v1 at signing time.
-/// It checks that none of the coin types in the transaction are denied for the sender.
+/// It checks that none of the coin types in the transaction are denied for an address they are
+/// debited from: the sender, plus the funder of each withdrawal, which an allowance lets differ
+/// from the sender.
```
tnowacki:
```suggestion
/// It checks that none of the coin types in the transaction are denied for an address they are
/// debited from. This includes the sender, the sponsor, and any allowance's funding source.
```

crates/sui-types/src/deny_list_v1.rs
```diff
@@ -68,7 +73,32 @@ pub fn check_coin_deny_list_v1(
             return Ok(());
         }
     };
-    check_deny_list_v1_impl(deny_list, sender, coin_types, object_store)
+    check_deny_list_v1_impl(deny_list, addresses_by_coin_type, object_store)
+}
+
+/// A map of (coin_type -> addresses) to be checked for deny-list.
+/// Includes objects (input, receiving) + fund withdrawals.
+/// Sender is included for all withdrawal types (sponsor / allowances).
```
tnowacki:
```suggestion
/// Generates a map of (coin_type -> addresses) to be checked for deny-list.
/// Includes the sender's input objects, and the withdrawals from address balances (which may have funding sources other than the sender)
```

crates/sui-types/src/deny_list_v1.rs
tnowacki: +1

crates/sui-types/src/deny_list_v1.rs
```diff
@@ -95,21 +125,25 @@ pub(crate) fn input_object_coin_types_for_denylist_check(
...
-        &address,
-    ) else {
-        return Ok(());
-    };
-    if count == 0 {
+    let addresses: BTreeSet<SuiAddress> =
+        addresses_by_coin_type.values().flatten().copied().collect();
+
+    // TODO: Add a limit of addresses? Our current boundary is 10 (due to max withdrawals)
+    // but might be worth making this explicit.
```
tnowacki: @lxfind any thoughts here?

crates/sui-adapter-transactional-tests/tests/deny_list_v2/allowance_deny_combinations.move
```diff
@@ -0,0 +1,209 @@
+// Copyright (c) Mysten Labs, Inc.
```
tnowacki, line 1: Do we have any sponsor cases?

crates/sui-types/src/deny_list_v1.rs
manolisliolios: added to the fn that gathers the data as we discussed

crates/sui-adapter-transactional-tests/tests/deny_list_v2/allowance_deny_combinations.move
manolisliolios: There's no way to exercise sponsor spends today in any way so it's not here. The current changes are future-proof for when that's added.

If we want to exercise it, we'll need to parse sponsor withdrawal arguments in the test runner, only to show sign-time failures

crates/sui-types/src/deny_list_v1.rs
```diff
@@ -68,7 +72,36 @@ pub fn check_coin_deny_list_v1(
...
+        let addresses = addresses_by_coin_type.entry(coin_type).or_default();
+        addresses.insert(sender);
+        addresses.insert(funder);
+    }
+    debug_assert!(
+        addresses_by_coin_type.values().all(|a| !a.is_empty()),
+        "every set contains at least the sender"
+    );
+    addresses_by_coin_type
 }
```
lxfind, line 120: **Shared deny-list helper is owned by `deny_list_v1` and imported by v2** -- Confidence: 98/100 | Importance: 72/100

(found by: Codex)
- **File**: `crates/sui-types/src/deny_list_v1.rs` lines 78-105; `crates/sui-types/src/deny_list_v2.rs` lines 7-9 and 124-129
- **Problem**: The new `addresses_by_coin_type_for_denylist_check` helper encodes version-neutral policy (which addresses are relevant for which coin type), yet it lives in the v1 module, and v2 imports it. This extends the existing pattern where v2 already depended on the v1-owned coin-type helper. Claude's review observed the same structure and described it as consistent with the prior pattern rather than flagging it, so this is a design judgment, not a correctness bug.
- **Impact**: v2's signing semantics are lifecycle-coupled to the legacy v1 module. Refactoring or removing v1 could inadvertently alter v2, and the ownership of shared policy is obscured. Functionally the code is correct today.
- **Suggestion**: Move the address-grouping helper, the pre-existing `input_object_coin_types_for_denylist_check`, and ideally the shared deny-list constants into a neutral crate-private module (for example `coin_deny_list_common`), and have both v1 and v2 depend on that module.

---

lxfind (commented):

---

crates/sui-types/src/deny_list_v1.rs
lxfind: Do we need ever need a tighter bound than max withdrawal for this?

tzakian (approved):
👍🏻

PR #27791, opened 2026-08-21
protocol-config: disable defer_unpaid_amplification; reconcile v134/v135 with release branches

mystenmark:
## Description

Disables `defer_unpaid_amplification` in a way that is compatible with current deployments: testnet already runs v134 as released from the 1.78 branch, and mainnet is on v133 (1.77 branch), so the disable must keep per-(version, chain) config identical across every binary that can serve a network at that version.

- v134: the pre-existing content (original_package_id costs, consensus block limits) is gated to `chain != Mainnet`, and mainnet's v134 carries only the deferral disable — matching the backfilled v134 on the 1.77 branch.
- v135: applies the original v134 content on mainnet and disables the flag everywhere — matching the new v135 on the 1.78 branch.
- v136 (new): the previously pending v135 content (`ptb_tx_context_restrictions`, `allowed_proposers` on devnet), which is not in any release branch.

The framework bytecode snapshot recorded at the 1.78 release cut moves from v134 to v135, mirroring the same move on the 1.78 branch: mainnet takes v134 from the 1.77 branch with the old framework, so the native-bearing framework ships universally at v135. This also keeps the mainnet split-cluster check working once origin/mainnet supports v134.

Companion release-branch PRs: MystenLabs/sui#27792 (1.77, merged — deploying to mainnet), MystenLabs/sui#27794 (1.78). All v134 snapshots verified byte-identical to origin/mainnet and all v135 snapshots byte-identical to #27794.

## Test plan

Verified the `(v134, Mainnet)` snapshot and all v135 snapshots are byte-identical to the corresponding snapshots in the 1.77/1.78 companion PRs, and that the released testnet/devnet v134 snapshots are unchanged.

---

## Release notes

- [x] Protocol: Adds protocol versions 135 and 136. `defer_unpaid_amplification` is disabled on mainnet at v134 and on all networks from v135; v135 also applies v134's original_package_id costs and consensus block limits on mainnet. v136 carries the previously pending v135 features.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


ebmifa (approved):

crates/sui-framework-snapshot/manifest.json
```diff
@@ -2849,7 +2849,7 @@
       }
     ]
   },
-  "134": {
+  "135": {
     "git_revision": "2a0becb2fcc6989e492981104af67f62f2c9511a",
```
tzakian, line 2853: This git revision is pointing to the wrong place now I think? But this may need to be a multi-step land in order to get it right possibly.

tzakian (approved):

PR #27803, opened 2026-08-21
[move][move-compiler] Match compilation improvements

cgswords:
## Description 

This PR fixes and improves a number of match-related issues:

1. ICE when a type-parameter field hit head position on the match fringe with a discriminating sibling behind it (#26788)
2. Exponential default-tree duplication for all-wild enum columns (fixed by the same fix as 1).
3. Compiler panic on a non-bool literal in a bool column (`Wrap { b: 5 }`) in both exhaustiveness and the IDE missing-arms report.
4. Compiler panic (None.unwrap()) on a foreign variant name saturating an enum column.
5. Spurious ICE diagnostic on legal generic match subjects (match (t: T)) in IDE mode.
6. Latent panic in the datatype dispatch for non-datatype `Apply` types (tuples).
7. No longer manufacture spurious non-exhaustive error from dropped ErrorPat rows.

In addition, I improved the code as follows:

1. `MatchTree` reshaped to be more honest
2. Dispatch on datatype_kind (exhaustive match) instead of `tyargs`-presence / `is_struct`
3.`counterexample_rec` mirrors the compilation mechanism now.

## Test plan 

Plenty of new tests, including a few I should fix later.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-compiler/src/typing/match_analysis.rs
```diff
@@ -385,92 +424,28 @@ fn find_counterexample_impl(
...
-                .into_iter()
-                .collect::<BTreeSet<_>>();
-
-            let ctors = matrix.first_variant_ctors();
-            for ctor in ctors.keys() {
-                unmatched_variants.remove(ctor);
-            }
-            if unmatched_variants.is_empty() {
-                for (ctor, (ploc, arg_types)) in ctors {
+        match context.info().datatype_kind(&mident, &datatype_name) {
```
cgswords, line 427: Git really messed this diff up, but I promise I just indented the two existing blocks under the switch

tnowacki (commented):
Looks fine to me, but maybe @tzakian would be better suited to review

external-crates/move/crates/move-compiler/src/shared/matching.rs
```diff
@@ -154,6 +154,23 @@ impl PatternArm {
...
+            match &pat.pat.value {
+                TP::Binder(_, _) | TP::Wildcard => true,
+                TP::At(_, inner) => bindable_only(inner),
+                TP::Variant(..)
+                | TP::BorrowVariant(..)
+                | TP::Struct(..)
+                | TP::BorrowStruct(..)
+                | TP::Literal(_)
+                | TP::ErrorPat => false,
+                TP::Constant(_, _) | TP::Or(_, _) => unreachable!(),
```
tzakian: Nit: ICE + unreachable? or `ice_assert`?

external-crates/move/crates/move-compiler/src/shared/matching.rs
```diff
@@ -565,6 +582,12 @@ impl PatternMatrix {
         })
     }
 
+    /// Returns true if each row's head pattern is a binder or wildcard
+    /// (e.g., none discriminate the subject)
+    pub fn first_column_binders_only(&self) -> bool {
```
tzakian: I don't love this name, but not sure I can suggest a better one, so I'll toss out a couple options, but feel free to to pick one/a different one or just keep this if you want.

other options:
* `all_binders_in_first_column`
* `first_column_all_binders` << my fave

tzakian (approved):
Couple little nits to consider but LGTM

external-crates/move/crates/move-compiler/src/shared/matching.rs
cgswords: Done

external-crates/move/crates/move-compiler/src/shared/matching.rs
cgswords: Done

PR #27805, opened 2026-08-22
[1.78][alt] protocol-config: disable defer_unpaid_amplification on mainnet in v134

mystenmark:
## Description

Alternate to MystenLabs/sui#27794, for the scenario where the deferral disable ships to mainnet directly in a 1.78 release instead of via a 1.77 patch release first.

Mainnet is on v133, so `(v134, Mainnet)` is still mutable: the flag is disabled inside the existing v134 block under a `Chain::Mainnet` guard, leaving the released `(v134, Testnet)` config untouched. Mainnet picks up the full v134 (original_package_id costs, consensus block limits, deferral disable) when it upgrades on 1.78 binaries. No new protocol version, no gating of the existing v134 content.

**Mutually exclusive** with MystenLabs/sui#27792 (1.77) and MystenLabs/sui#27794, which define a different `(v134, Mainnet)`. If this lands, those must not ship, and main (MystenLabs/sui#27791) needs the matching minimal scheme (mainnet-guarded disable folded into its v134 block, no v135/v136 restructure). Note testnet keeps the flag enabled until a later version disables it there.

## Test plan

Snapshot diff is exactly one line: `defer_unpaid_amplification` removed from `Mainnet_version_134`; Testnet/devnet v134 snapshots unchanged.

---

## Release notes

- [x] Protocol: Disables `defer_unpaid_amplification` on mainnet in protocol version 134 on the 1.78 branch (testnet's released v134 config is unchanged).
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-protocol-config/src/lib.rs
+++ b/crates/sui-protocol-config/src/lib.rs
@@ -377,6 +377,7 @@ const MAINNET_USDB: &str =
 //              Bound type nodes in accumulators.
 // Version 134: Add `package::original_package_id` and its native costs.
 //              Reduce the consensus block transaction count and payload limits.
+//              Disable defer_unpaid_amplification on mainnet.
 
 #[derive(Copy, Clone, Debug, Hash, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
 pub struct ProtocolVersion(u64);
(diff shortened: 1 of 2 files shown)


mystenmark:
Closing: the 1.77 approach shipped (#27792 merged, 1.77.x deploying to mainnet), so this alternate is obsolete. #27794 tracks the 1.77 protocol config scheme instead.

PR #27810, opened 2026-08-24
[feat] Add ML-DSA-65 account signature scheme (flag 0x08)

mahdi-mysten:
## Description

Adds ML-DSA-65 (FIPS 204) as an account signature scheme (`0x07`), gated off on every network.

**Types**

* Adds `SignatureScheme::MLDSA65` with boxed `PublicKey`/`CompressedSignature` variants appended last to preserve BCS indices.
* Uses `flag || sig || pk` envelopes (5,262 bytes), Ed25519-style `flag || seed` keystore blobs, and `blake2b256(flag || pk)` addresses.
* Uses crypto from `fastcrypto-pq`.
* gRPC/GraphQL wire types and JSON-RPC schema are intentionally deferred to the dedicated RPC PR after the [sui-rust-sdk](https://github.com/mystenlabs/sui-rust-sdk) flag PR. The new variants use `schemars(skip)` and don't affect serde/BCS.

**Protocol gating**

* `mldsa65_auth`: rejects plain ML-DSA-65 signatures as `Unsupported` unless enabled.
* `accept_mldsa65_in_multisig`: rejects committees containing ML-DSA members unless enabled, following the existing zkLogin/passkey multisig flags. Kept separate because multisig signing/verification support lands in the multisig PR.
* Both flags are default-off with no protocol version arm, so there are no protocol snapshot/version changes. Devnet enablement will follow separately.

**Follow-ups**

* sui-rust-sdk flag PR, then gRPC/GraphQL RPC PR.
* Devnet enablement.
* Mnemonic derivation and multisig signing. zkLogin keeps its ephemeral scheme classical.

## Test plan

* `cargo simtest -p sui-e2e-tests mldsa`: covers enabled/disabled flags, bad signatures, and hybrid Ed25519 + ML-DSA multisig in both formats.
* `cargo test -p sui-types --lib mldsa65`: covers golden vectors, round-trips, invalid lengths, `0x07` parsing, zkLogin ephemeral rejection, and BCS index stability.
* `cargo test -p sui-protocol-config`, `-p sui-types --test format`, `-p sui-open-rpc --test generate-spec`, `-p sui-indexer-alt-graphql`, `-p sui keytool`, `-p sui-keys`.
* `cargo check --workspace --all-targets`, clippy, and fmt clean.

---

## Release notes

* [x] Protocol: Adds `mldsa65_auth` and `accept_mldsa65_in_multisig`, both default-off. ML-DSA-65 signatures and committee members remain `Unsupported` until enabled.
* [ ] Nodes (Validators and Full nodes):
* [ ] gRPC:
* [ ] GraphQL:
* [x] CLI: `sui keytool` supports importing/exporting `mldsa65` keys in `suiprivkey` form. Mnemonic derivation is not yet supported.
* [ ] Rust SDK:
* [ ] Indexing Framework:

### DIFF
--- a/crates/sui-e2e-tests/tests/mldsa_tests.rs
+++ b/crates/sui-e2e-tests/tests/mldsa_tests.rs
@@ -1,0 +1,215 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use fastcrypto::traits::{KeyPair as _, ToFromBytes};
+use fastcrypto_pq::mldsa65::MLDSA65KeyPair;
+use shared_crypto::intent::{Intent, IntentMessage};
+use std::net::SocketAddr;
+use sui_core::authority_client::AuthorityAPI;
+use sui_macros::sim_test;
+use sui_protocol_config::ProtocolConfig;
+use sui_test_transaction_builder::TestTransactionBuilder;
+use sui_types::base_types::SuiAddress;
+use sui_types::crypto::{Signature, SuiKeyPair};
+use sui_types::effects::TransactionEffectsAPI;
+use sui_types::error::{SuiErrorKind, SuiResult, UserInputError};
+use sui_types::messages_grpc::SubmitTxRequest;
+use sui_types::signature::GenericSignature;
+use sui_types::transaction::{Transaction, TransactionData};
+use test_cluster::{TestCluster, TestClusterBuilder};
+
+fn mldsa_keypair() -> SuiKeyPair {
+    SuiKeyPair::MLDSA65(MLDSA65KeyPair::generate(&mut rand::thread_rng()))
+}
+
+/// Submits to one validator, so `Ok` means that validator's signature
+/// verification accepted the transaction.
+async fn execute_tx(tx: Transaction, test_cluster: &TestCluster) -> SuiResult {
+    test_cluster
+        .authority_aggregator()
+        .authority_clients
+        .values()
+        .next()
+        .unwrap()
+        .authority_client()
+        .submit_transaction(
+            SubmitTxRequest::new_transaction(tx),
+            Some(SocketAddr::new([127, 0, 0, 1].into(), 0)),
+        )
+        .await
+        .map(|_| ())
+}
+
+/// Fund the keypair's address and build an unsigned transfer from it.
+async fn funded_transfer(test_cluster: &TestCluster, kp: &SuiKeyPair) -> TransactionData {
+    let sender = SuiAddress::from(&kp.public());
+    let rgp = test_cluster.get_reference_gas_price().await;
+    let gas = test_cluster
+        .fund_address_and_return_gas(rgp, Some(20000000000), sender)
+        .await;
+    TestTransactionBuilder::new(sender, gas, rgp)
+        .transfer_sui(None, SuiAddress::ZERO)
+        .build()
+}
+
+fn sign(tx_data: TransactionData, kp: &SuiKeyPair) -> Transaction {
+    let intent_msg = IntentMessage::new(Intent::sui_transaction(), tx_data);
+    let sig = Signature::new_secure(&intent_msg, kp);
+    Transaction::from_generic_sig_data(intent_msg.value, vec![GenericSignature::Signature(sig)])
+}
+
+/// Tests of the enabled path turn the flag on explicitly rather than depending
+/// on which protocol versions carry it; `test_mldsa65_feature_deny` pins it off.
+fn enable_mldsa() -> sui_protocol_config::OverrideGuard {
+    ProtocolConfig::apply_overrides_for_testing(|_, mut config| {
+        config.set_mldsa65_auth_for_testing(true);
+        config
+    })
+}
+
+fn enable_mldsa_in_multisig() -> sui_protocol_config::OverrideGuard {
+    ProtocolConfig::apply_overrides_for_testing(|_, mut config| {
+        config.set_mldsa65_auth_for_testing(true);
+        config.set_accept_mldsa65_in_multisig_for_testing(true);
+        config
+    })
+}
+
+/// A funded 1-of-2 committee with Ed25519 and ML-DSA-65 members, and a
+/// transfer signed by the Ed25519 member alone, in both multisig formats.
+async fn hybrid_committee_txs(test_cluster: &TestCluster) -> [Transaction; 2] {
+    use sui_types::crypto::get_key_pair;
+    use sui_types::multisig::{MultiSig, MultiSigPublicKey};
+    use sui_types::multisig_legacy::{MultiSigLegacy, MultiSigPublicKeyLegacy};
+
+    let ed_kp = SuiKeyPair::Ed25519(get_key_pair().1);
+    let pks = vec![ed_kp.public(), mldsa_keypair().public()];
+    let multisig_pk = MultiSigPublicKey::new(pks.clone(), vec![1, 1], 1).unwrap();
+    let legacy_pk = MultiSigPublicKeyLegacy::new(pks, vec![1, 1], 1).unwrap();
+    let sender = SuiAddress::from(&multisig_pk);
+
+    let rgp = test_cluster.get_reference_gas_price().await;
+    let gas = test_cluster
+        .fund_address_and_return_gas(rgp, Some(20000000000), sender)
+        .await;
+    let tx_data = TestTransactionBuilder::new(sender, gas, rgp)
+        .transfer_sui(None, SuiAddress::ZERO)
+        .build();
+    let intent_msg = IntentMessage::new(Intent::sui_transaction(), tx_data);
+    let sig: GenericSignature = Signature::new_secure(&intent_msg, &ed_kp).into();
+    [
+        GenericSignature::MultiSig(MultiSig::combine(vec![sig.clone()], multisig_pk).unwrap()),
+        GenericSignature::MultiSigLegacy(MultiSigLegacy::combine(vec![sig], legacy_pk).unwrap()),
+    ]
+    .map(|generic| Transaction::from_generic_sig_data(intent_msg.value.clone(), vec![generic]))
+}
+
+#[sim_test]
+async fn test_mldsa65_transfer_executes() {
+    let _guard = enable_mldsa();
+    let test_cluster = TestClusterBuilder::new().build().await;
+    let kp = mldsa_keypair();
+    let tx_data = funded_transfer(&test_cluster, &kp).await;
+    let (effects, _) = test_cluster
+        .execute_transaction_return_raw_effects(sign(tx_data, &kp))
+        .await
+        .unwrap();
+    assert!(effects.status().is_ok(), "{:?}", effects.status());
+}
+
+#[sim_test]
+async fn test_mldsa65_feature_deny() {
+    // With the flag pinned off, the transaction is refused as Unsupported
+    // before signature verification runs.
+    let _guard = ProtocolConfig::apply_overrides_for_testing(|_, mut config| {
+        config.set_mldsa65_auth_for_testing(false);
+        config
+    });
+    let test_cluster = TestClusterBuilder::new().build().await;
+    let kp = mldsa_keypair();
+    let tx_data = funded_transfer(&test_cluster, &kp).await;
+    let err = execute_tx(sign(tx_data, &kp), &test_cluster)
+        .await
+        .unwrap_err();
+    assert!(matches!(
+        err.as_inner(),
+        SuiErrorKind::UserInputError {
+            error: UserInputError::Unsupported(..)
+        }
+    ));
+}
+
+#[sim_test]
+async fn test_mldsa65_wrong_signer_fails() {
+    let _guard = enable_mldsa();
+    let test_cluster = TestClusterBuilder::new().build().await;
+    let kp = mldsa_keypair();
+    let other = mldsa_keypair();
+    let tx_data = funded_transfer(&test_cluster, &kp).await;
+    // Signed by a different ML-DSA key than the sender's: past the feature
+    // gate, rejected at signature verification.
+    let err = execute_tx(sign(tx_data, &other), &test_cluster)
+        .await
+        .unwrap_err();
+    assert!(
+        matches!(err.as_inner(), SuiErrorKind::SignerSignatureAbsent { .. }),
+        "must fail verification, not the feature gate: {err:?}"
+    );
+}
+
+#[sim_test]
+async fn test_mldsa65_tampered_signature_fails() {
+    let _guard = enable_mldsa();
+    let test_cluster = TestClusterBuilder::new().build().await;
+    let kp = mldsa_keypair();
+    let tx_data = funded_transfer(&test_cluster, &kp).await;
+    let intent_msg = IntentMessage::new(Intent::sui_transaction(), tx_data);
+    let sig = Signature::new_secure(&intent_msg, &kp);
+
+    // Flip one byte inside the signature body. The envelope still parses
+    // (flag, length and pk intact), so the failure is verification proper.
+    let mut bytes = sig.as_ref().to_vec();
+    bytes[100] ^= 0x01;
+    let tampered = GenericSignature::from_bytes(&bytes).unwrap();
+    let tx = Transaction::from_generic_sig_data(intent_msg.value, vec![tampered]);
+    let err = execute_tx(tx, &test_cluster).await.unwrap_err();
+    assert!(
+        matches!(err.as_inner(), SuiErrorKind::InvalidSignature { .. }),
+        "must fail verification, not the feature gate: {err:?}"
+    );
+}
+
+#[sim_test]
+async fn test_mldsa65_multisig_member_denied() {
+    // `mldsa65_auth` alone does not admit ML-DSA committee members: that is
+    // `accept_mldsa65_in_multisig`, off here.
+    let _guard = enable_mldsa();
+    let test_cluster = TestClusterBuilder::new().build().await;
+    for tx in hybrid_committee_txs(&test_cluster).await {
+        let err = execute_tx(tx, &test_cluster).await.unwrap_err();
+        assert!(
+            matches!(
+                err.as_inner(),
+                SuiErrorKind::UserInputError {
+                    error: UserInputError::Unsupported(..)
+                }
+            ),
+            "expected the member gate to refuse the committee: {err:?}"
+        );
+    }
+}
+
+#[sim_test]
+async fn test_mldsa65_multisig_member_accepted_when_enabled() {
+    // With the multisig flag on, hybrid committees can transact on their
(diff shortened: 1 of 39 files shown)


crates/sui-indexer-alt-graphql/src/api/types/user_signature/multisig.rs
```diff
@@ -152,32 +152,33 @@ impl From<&sui_types::multisig::MultiSigPublicKey> for MultisigCommittee {
...
+                sui_sdk_types::ZkLoginPublicIdentifier::try_from(z.to_owned())
+                    .map(|id| ZkLoginPublicIdentifier {
+                        iss: Some(id.iss().to_owned()),
+                        address_seed: Some(id.address_seed().to_string()),
+                    })
+                    .unwrap_or_default(),
+            ))
         }
+        // TODO: add once the GraphQL API defines MLDSA65.
+        PublicKey::MLDSA65(_) => None,
```
amnn: The optional type here is for GraphQL schema backwards compatibility reasons. Pleae introduce the GraphQL type associated with the name signature scheme and convert this back into a `From` impl before landing.

mahdi-mysten: Good catch @amnn, thanks!
Added a first-class `Mldsa65PublicKey` type, extended the union (appended, so it stays additive), and converted back to the From impl. Regenerated `schema.graphql`, `staging.graphql`, etc for both feature configurations.
Let me know if that works.

crates/sui-indexer-alt-graphql/src/api/types/user_signature/multisig.rs
```diff
@@ -51,6 +51,7 @@ pub(crate) enum MultisigMemberPublicKey {
     Secp256r1(Secp256r1PublicKey),
     Passkey(PasskeyPublicKey),
     ZkLogin(ZkLoginPublicIdentifier),
+    MLDSA65(MLDSA65PublicKey),
```
amnn: We tend to use CamelCase even for acronyms (see e.g. the other ones in this list), but otherwise this looks good, thanks!

```suggestion
    Mldsa65(Mldsa65PublicKey),
```

amnn: Or perhaps you want it to be `MlDsa65`, up to you!

crates/sui-types/src/rpc_proto_conversions.rs
```diff
@@ -1679,22 +1679,35 @@ impl From<&crate::crypto::ZkLoginPublicIdentifier> for ZkLoginPublicIdentifier {
...
+            S::MultiSig => Ok(Self::Multisig),
+            S::ZkLoginAuthenticator => Ok(Self::Zklogin),
+            S::PasskeyAuthenticator => Ok(Self::Passkey),
+            S::MLDSA65 => Err(value),
         }
     }
 }
 
+/// Proto3 enums are open: schemes without a proto variant are carried as their raw flag byte.
+// TODO: drop once sui.rpc.v2 defines MLDSA65.
```
joyqvq: you can open a pr in https://github.com/MystenLabs/sui-rust-sdk to add it!

crates/sui-types/src/unit_tests/crypto_tests.rs
```diff
@@ -138,3 +139,146 @@ proptest! {
 
 
 }
+
+// Same seed as the pq-sigs-ts "digest-msg" interop vector.
+const MLDSA65_TEST_SEED: [u8; 32] = [2; 32];
```
joyqvq, line 145: do we need to use this seed for all tests? maybe have one golden test, and the rest of the tests should use random seed

crates/sui-types/src/unit_tests/crypto_tests.rs
```diff
@@ -138,3 +139,146 @@ proptest! {
...
+    let bytes = sig.as_ref().to_vec();
+    assert_eq!(bytes.len(), 5262);
+    let parsed = Signature::from_bytes(&bytes).unwrap();
+    assert_eq!(parsed, sig);
+    assert_eq!(parsed.as_ref(), &bytes[..]);
+}
+
+#[test]
+fn mldsa65_verify_secure_intent_message() {
+    let kp = mldsa65_test_keypair();
```
joyqvq: use random keypair and below

crates/sui-types/src/sui_sdk_types_conversions.rs
```diff
@@ -1234,21 +1239,32 @@ impl TryFrom<crate::crypto::Signature> for SimpleSignature {
                     )?,
                 }
             }
+            crate::crypto::Signature::MLDSA65SuiSignature(_) => {
+                return Err(SdkTypeConversionError(
```
joyqvq, line 1243: you can open a pr in https://github.com/MystenLabs/sui-rust-sdk to add it!

good idea land that first and upstream it here

crates/sui-types/src/unit_tests/crypto_tests.rs
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
```
joyqvq, line 2: do we plan to add a multisig test to check for mldsa accepted vs not? e.g. https://github.com/MystenLabs/sui/blob/7b0a3a58bdd9fc1f418cdb3812ff7efac7f1a495/crates/sui-e2e-tests/tests/multisig_tests.rs#L1146

also need the protocol config changes and check feature flag in both the single sig verifier and also the multisig verifier

crates/sui/src/unit_tests/keytool_tests.rs
```diff
@@ -108,6 +109,13 @@ async fn test_flag_in_signature_and_keypair() -> Result<(), anyhow::Error> {
                 );
                 assert!(pk1.flag() == Secp256r1SuiSignature::SCHEME.flag())
             }
+            Signature::MLDSA65SuiSignature(_) => {
```
joyqvq, line 121: this arm is not exercised - you would need to import the keystore on line 79

crates/sui/src/unit_tests/keytool_tests.rs
joyqvq, line 145: add mldsa to this test too too for the file write/read

joyqvq:
> As with zkLogin and passkeys, this lands the types first so the BCS positions can be reviewed independently. Transaction verification behind a protocol flag, mnemonic derivation, and multisig support (including the CompressedSignature variant)) will follow in separate PRs.

the protocol flag for single sig and multisig to allow mldsa for devnet must be part of this pr upon landing. otherwise it will go out on a release and enable for all networks. 

for multisig flag see example: https://github.com/MystenLabs/sui/blob/7b0a3a58bdd9fc1f418cdb3812ff7efac7f1a495/crates/sui-types/src/multisig.rs#L140-L145

for single sig flag see: https://github.com/MystenLabs/sui/blob/7b0a3a58bdd9fc1f418cdb3812ff7efac7f1a495/crates/sui-types/src/transaction.rs#L4052-L4060

the mnemonics and key derivation can wait for the next pr since they are client side changes. same for anything related to keystore file too.

joyqvq:
cc @amnn @mahdi-mysten discussed in slack that we should remove the diff for jsonrpc bc its deprecated. 

re wire type on grpc/graphql, please remove from this pr. this pr should be focused on sui core protocol changes with validator code changes only, including feature flags.

then you can open and land the sui-rust-sdk pr and open another pr in sui just for the sui grpc/graphql. since the support them are orthogonal to sui protocol's support on this.

crates/sui-types/src/rpc_proto_conversions.rs
mahdi-mysten: Will open the PR this week in parallel. Keeping the raw-flag fallback here so this PR doesn't block on that merge; it collapses in the follow-up that bumps the sui-sdk-types rev.

crates/sui-types/src/unit_tests/crypto_tests.rs
mahdi-mysten: Not necessarily, I will fix this in the next push. The idea will be to use the fixed seed now backs only the golden address test; eight call sites generate a fresh keypair.

crates/sui-types/src/unit_tests/crypto_tests.rs
mahdi-mysten: Yes, staged to keep this PR types-only: the verification PR is ready (protocol flag, single-sig verifier gate, e2e for both flag states), and the multisig PR adds CompressedSignature::MLDSA65 plus acceptance in the multisig verifier wired to the same flag, with accepted-vs-not tests following the pattern you linked. I was planning to keep the PR easy to read and review

crates/sui/src/unit_tests/keytool_tests.rs
mahdi-mysten: Good catch, will work on it.

crates/sui-types/src/rpc_proto_conversions.rs
joyqvq: whats the point of adding the todos and the schema change at all? this pr should just contain the core changes it keeps this pr focused easy to review.

crates/sui-types/src/unit_tests/crypto_tests.rs
joyqvq: > I was planning to keep the PR easy to read and review

thats not the logical split of the pr. to keep it easy to review is to leave out all the client side changes ie sui keystores and the rpc changes. this pr should contain sui protocol changes only which has to include the feature flags. otherwise you will end up releasing this to all networks without gating. 

> the verification PR is ready

also where is this?

mahdi-mysten: Agreed on the split. Lets remove gRPC/GraphQL wire types from this PR. Quick note, Json RPC was part of the CI and i havent touched that part and already mentioned in the description that it is deprecated. 

You are right, I’ll keep the verification work here, including the `mldsa65_auth` feature gate, `from_bytes` gating, and simtest coverage, so ML-DSA can’t ship ungated.

> also where is this?

I have them on my local, planned to push and open a new PR but will bring them all here. Tnx!

crates/sui-types/src/crypto.rs
```diff
@@ -267,6 +283,11 @@ pub enum PublicKey {
     Secp256r1(Secp256r1PublicKeyAsBytes),
     ZkLogin(ZkLoginPublicIdentifier),
     Passkey(Secp256r1PublicKeyAsBytes),
+    // Boxed because the 1,952-byte pk; Box is transparent to BCS.
+    // Schema-only skip (serde unaffected): keeps the generated openrpc spec
+    // unchanged, since deprecated JSON-RPC will never serve ML-DSA.
+    #[schemars(skip)]
```
mahdi-mysten, line 289: the openrpc spec diff is gone (the ML-DSA variants are schema-skipped since deprecated jsonrpc will never serve them; serde/BCS unaffected).

mahdi-mysten:
@joyqvq restructured this PR based on your comments. Quick summary of what's now in it:

**Removed**

* gRPC/GraphQL wire types: moving these to a separate RPC PR after the sui-rust-sdk flag PR, which is much cleaner anyway.
* jsonrpc: `#[schemars(skip)]` keeps the new variants out of the OpenRPC spec (serde/BCS unaffected). The only lines left in that snapshot are the auto-generated `sui_getProtocolConfig` flag-map example, which every new feature flag touches and it was out of my hand.

**Added (protocol gating)**

* `mldsa65_auth`: plain ML-DSA-65 signatures are rejected as `Unsupported` unless this is set.
* `accept_mldsa65_in_multisig`: same for committees containing an ML-DSA member, following `accept_zklogin_in_multisig` / `accept_passkey_in_multisig`. Kept separate because multisig verification only checks the members who signed; the ML-DSA compressed-signature and verify arms come with the multisig PR, which enables both flags. Flagging it here so we're on the same page early.
* e2e tests under simtest: enabled/disabled flags, wrong signer, tampered signature, and a hybrid Ed25519 + ML-DSA multisig in both formats (denied with the flag off, executing on the Ed25519 signature with it on).

Both flags are default-off with no version arm, so there are no protocol snapshot or version changes and ML-DSA stays gated off everywhere. Devnet enablement would be a small follow-up under whatever protocol version is current at that point, which avoids racing main's version bumps (135 to 137 since I branched). If you'd rather have the version arm in this PR, I can rebase and add it.

@amnn didnt touch `From` impl for GraphQL. The ML-DSA arm is unreachable by construction: `MultisigCommittee::from`, its only caller, renders ML-DSA members as `null` until the dedicated GraphQL PR adds the member type. Will work on this later.

crates/sui-protocol-config/src/lib.rs
```diff
@@ -774,6 +772,10 @@ struct FeatureFlags {
     #[serde(skip_serializing_if = "is_false")]
     passkey_auth: bool,
 
+    // Enable ML-DSA-65 (FIPS 204) account signatures.
+    #[serde(skip_serializing_if = "is_false")]
+    mldsa65_auth: bool,
+
```
joyqvq, line 778: you need to actually enable them here? https://github.com/MystenLabs/sui/blob/eac911a4e2bea5faa795672a715e6ffa0c19fb9c/crates/sui-protocol-config/src/lib.rs#L4615-L4621 also there maybe some diffs in snapshots tests need to regen
```
    cfg.feature_flags.mldsa65_auth = true;
    cfg.feature_flags.accept_mldsa65_in_multisig = true;
}
```

you also need to make bump it to 136 after rebasing to latest main and add it there

crates/sui-types/src/rpc_proto_conversions.rs
```diff
@@ -1809,27 +1821,27 @@ impl From<&crate::crypto::CompressedSignature> for MultisigMemberSignature {
...
             }
             crate::crypto::CompressedSignature::ZkLogin(_z) => {
                 //TODO
-                SignatureScheme::Zklogin
+                SignatureScheme::Zklogin as i32
             }
             crate::crypto::CompressedSignature::Passkey(_p) => {
                 //TODO
-                SignatureScheme::Passkey
+                SignatureScheme::Passkey as i32
```
joyqvq, line 1840: why they are i32 now?

crates/sui-types/src/rpc_proto_conversions.rs
```diff
@@ -1809,27 +1821,27 @@ impl From<&crate::crypto::CompressedSignature> for MultisigMemberSignature {
...
                 message.signature = Some(b.0.to_vec().into());
-                SignatureScheme::Secp256r1
+                SignatureScheme::Secp256r1 as i32
             }
             crate::crypto::CompressedSignature::ZkLogin(_z) => {
                 //TODO
-                SignatureScheme::Zklogin
+                SignatureScheme::Zklogin as i32
             }
             crate::crypto::CompressedSignature::Passkey(_p) => {
```
joyqvq, line 1838: CompressedSignature::MLDSA65 is still missing. how are you using it in multisig?

you also need to add it here https://github.com/MystenLabs/sui/pull/27810/changes#diff-c6de70b6e61851a12afce07640a17896557a780e0653c595af70137824c4d244L175

joyqvq: if you do plan to add the feature gate for multisig/mldsa in this pr, then you should implement it too. even though its disabled. currently the e2e test doesnt actually test the flag being enabled/disabled bc ML-DSA cannot actually sign as a multisig member yet.

crates/sui-e2e-tests/tests/mldsa_tests.rs
```diff
@@ -0,0 +1,215 @@
...
+            "expected the member gate to refuse the committee: {err:?}"
+        );
+    }
+}
+
+#[sim_test]
+async fn test_mldsa65_multisig_member_accepted_when_enabled() {
+    // With the multisig flag on, hybrid committees can transact on their
+    // classical members' signatures; ML-DSA members are covered by the
+    // compressed-signature support that enables this flag.
```
joyqvq, line 206: multisig test only puts an ML-DSA key in the committee; the actual signature is Ed25519 with threshold 1. this test should construct the mldsa sig inside multisig and check if its accepted. 

> ML-DSA members are covered by the compressed-signature

i dont think its covered. is there another pr? lets make sure its in this one

crates/sui-indexer-alt-graphql/src/api/types/user_signature/multisig.rs
```diff
@@ -142,7 +142,10 @@ impl From<&sui_types::multisig::MultiSigPublicKey> for MultisigCommittee {
                 pk.pubkeys()
                     .iter()
                     .map(|(public_key, weight)| MultisigMember {
-                        public_key: Some(MultisigMemberPublicKey::from(public_key)),
+                        // ML-DSA-65 members have no GraphQL type yet; render
+                        // them as null rather than reaching the `From` arm.
```
joyqvq, line 146: i think you can just do unreachable error (maybe change this to tryfrom instead of from so you can return error

amnn:
> @amnn didnt touch `From` impl for GraphQL. The ML-DSA arm is unreachable by construction: `MultisigCommittee::from`, its only caller, renders ML-DSA members as `null` until the dedicated GraphQL PR adds the member type. Will work on this later.

SGTM!

crates/sui-types/src/crypto.rs
joyqvq: #[schemars(skip)] not needed?

mahdi-mysten:
@joyqvq pushed the multisig signing and the rest of your comments, quick summary:

**Multisig**

* `CompressedSignature::MLDSA65` is in, with the `to_compressed`, `to_public_key` and `verify_claims` arms.
* `accept_mldsa65_in_multisig` is now checked in `MultiSig::verify_claims` through `VerifyParams`, same shape as passkey and zkLogin. Client side helpers pass `false` until the flag ships.
* Dropped the earlier membership gate in the compat check, so a hybrid committee keeps working on its classical members with the flag off. Only the ML-DSA member's signature is gated.
* e2e now has the ML-DSA member actually signing: alone with the flag off it is rejected in both the upgraded and legacy formats, and 2 of 2 with both members signing executes with the flag on. Unit test for the verify arm as well.

**Other comments**

* `set_scheme` is back, no more i32. The ML-DSA arms just leave the proto field unset until the RPC PR.
* GraphQL member key conversion is `TryFrom` now as you suggested, ML-DSA members render as `null`.
* Flag moved to `0x08` since zkLogin v2 takes `0x07`, goldens regenerated, title updated. This will be a huge conflict but we can defer this for later.

**Version arm**

* Still pending as Idk when we can move forward with this. I will rebase onto main and add both flags under the next version for non mainnet/testnet chains plus the three snapshots. Wanted to get the multisig part in front of you first and have your comments on.

crates/sui-types/src/crypto.rs
joyqvq: ah i see its needed by jsonrpc deprecation

PR #27816, opened 2026-08-24
docs: add the human.tech Wallet Protocol SDK and WaaP CLI to the tooling catalogue

hebbianloop:
Adds two `ToolCard`s: an embedded wallet SDK to **Building apps**, and the agent-facing CLI to **AI**.

Consolidated from two PRs into one, so this is the only change we are asking you to carry.

## Two surfaces, two audiences

| Card | Section | For |
|:---|:---|:---|
| **human.tech Wallet Protocol (WaaP) SDK** | Building apps | app developers adding an embedded wallet |
| **WaaP CLI** | AI | autonomous agents that transact |

They are genuinely different products with different install commands and different docs, which is why they sit in different sections rather than as one card. There is precedent for one org holding cards across sections.

## What the SDK is

White label embedded wallets. Users sign in with email, phone, a social account or biometrics, and get a wallet with no seed phrase. Keys are reconstructed in secure hardware, the user keeps ownership and can export at any time, and policies are enforced before a signature exists.

The account is **not scoped to one app**, so it follows the user, and it signs on Solana and EVM as well as Sui.

## What the CLI is

The headless wallet an autonomous agent uses. The agent holds **no key**, and spend caps are enforced server-side before a signature exists rather than by the agent's own code, so a prompt-injected agent has nothing to reach for. Grants are bound by recipient, chain, a cumulative ceiling and an expiry.

## History of this PR

It originally carried a single card describing the whole product agent-first, which narrowed an embedded wallet to an agent wallet in the one catalogue a Sui developer browses for an embedded wallet. That was our error and the split fixes it.

Your style guide audit flagged **line 1022, `on-chain` → `onchain`**. Fixed, and neither new card uses the hyphenated form.

## Related

A `sui-agent-wallets` skill is open at [MystenLabs/community-skills#1](https://github.com/MystenLabs/community-skills/pull/1).

Both cards are community-tagged. Not a Mysten or Sui Foundation product.

### DIFF
--- a/docs/content/getting-started/tooling.mdx
+++ b/docs/content/getting-started/tooling.mdx
@@ -550,6 +550,18 @@ Frontend toolkits, wallet integration, authentication, and app scaffolding.
 />
 
 <ToolCard
+  name="human.tech Wallet Protocol (WaaP) SDK"
+  description="White label embedded wallets with social login, biometrics and gas sponsorship. One account works across apps, and also signs on Solana and EVM."
+  notes="Non-custodial. Keys are reconstructed in secure hardware, the user keeps ownership and can export at any time. Policies are enforced before a signature exists, and WaaP Squid Mode adds decentralized signing across the Ika network."
+  install="npm install @human.tech/waap-sdk"
+  docs="https://docs.waap.human.tech/for-apps"
+  website="https://wallet.human.tech"
+  github="https://github.com/holonym-foundation/welcome.human.tech"
+  language="TypeScript"
+  community
+/>
+
+<ToolCard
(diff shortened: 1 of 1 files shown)


jessiemongeon1:
## Style Guide Audit

hebbianloop:
> ## Style Guide Audit
> All 1 file(s) pass the style guide audit.

thanks for checking! please note we realized that we should split the PR into two cards to cover the agentic tools + the embedded wallet tools

this is staged now and ready for review

jessiemongeon1 (approved):

PR #27820, opened 2026-08-25
docs(suins): document Pyth access token and refresh package ids

John-Atha:
## Description

Registering or renewing with SUI or NS now needs a price info object from a keyed Hermes endpoint. Add a Pyth price feeds section covering where the access token comes from and the two ways to get the object into a transaction.

Also repoint five stale package ids in the active constants tables against mainPackage in ts-sdks: mainnet core (the row pointed at the pricing package) and payments, plus testnet subnames, subnames proxy, and payments.

## Test plan

- Package ids checked against `mainPackage` in ts-sdks `main` (`fe68bf59`), both networks.
- Confirmed onchain via mainnet GraphQL: core `0x55800f20...` is version 6 (latest), payments `0xdbbf2339...` is v2, upgraded 2026-08-18, receiving traffic since 2026-08-24.

### DIFF
--- a/docs/content/sui-stack/suins/developer.mdx
+++ b/docs/content/sui-stack/suins/developer.mdx
@@ -50,13 +50,13 @@ The core package cannot change, nor can the core SuiNS object.
             <TabItem label="Packages" value='packages'>
 | Kind | Package ID |
 |------|-----------|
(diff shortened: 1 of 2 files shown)


jessiemongeon1:
## Style Guide Audit

jessiemongeon1 (approved):

PR #27824, opened 2026-08-25
[docs] fix dApp Kit ImportContent 404s on wallet pages

jessiemongeon1:
## Summary

- Fix `dapp-kit-next` → `dapp-kit` in all `ImportContent` source paths on the self-custody and slush wallet pages. The directory was renamed in the [ts-sdks repo](https://github.com/MystenLabs/ts-sdks), breaking all GitHub source imports with "GitHub fetch failed: 404"
- Remove a stray backtick on `self-custody.mdx` line 113 that prevented the first `ImportContent` (for `createDAppKit`) from rendering as an MDX component

**Affected pages:**
- https://docs.sui.io/onchain-finance/asset-custody/wallets/self-custody
- https://docs.sui.io/onchain-finance/asset-custody/wallets/slush

### DIFF
--- a/docs/site/src/shared/components/ImportContent/index.tsx
+++ b/docs/site/src/shared/components/ImportContent/index.tsx
@@ -180,7 +180,7 @@ export default function ImportContent({
       setGhErr(null);
       try {
         const branchName = branch || "main";
-        const path = String(source || "").replace(/^\.\/?/, "");
+        const path = String(source || "").replace(/^\.\/?/, "").replace(/^\/+/, "");
         const url = `https://raw.githubusercontent.com/${org}/${repo}/${branchName}/${path}`;
         const headers: Record<string, string> = {};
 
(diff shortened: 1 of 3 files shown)


jessiemongeon1:
## Style Guide Audit

docs/site/src/shared/components/ImportContent/index.tsx
```diff
@@ -180,7 +180,7 @@ export default function ImportContent({
       setGhErr(null);
       try {
         const branchName = branch || "main";
-        const path = String(source || "").replace(/^\.\/?/, "");
+        const path = String(source || "").replace(/^\.\/?/, "").replace(/^\/+/, "");
```
stefan-mysten, line 183: Why this change? Is it because below the /${path} already includes the forward slash?

stefan-mysten (approved):

docs/site/src/shared/components/ImportContent/index.tsx
jessiemongeon1: Yea, in some URLs it renders it as `//`

PR #27826, opened 2026-08-25
move-unit-test: add --package-size option to bound the package arena size

cgswords:
## Description

Adds a `--package-size <MB>` option to `sui move test` (and `move test`) that bounds the package arena packages are loaded into during test runs, so arena-limit behavior can be exercised without rebuilding the VM.

The arena size moves from a hardcoded constant into `VMRuntimeLimitsConfig` as an `Option<u64>`, fed on-chain by a new `package_arena_size_in_bytes` protocol config value (set to the existing 10MB in v136; the VM falls back to its local default when unset). `VMConfig::new_for_test` is the only other place that sets it.

## Test plan

Smoke-tested with a scratch package: default and `--package-size 20` pass, `--package-size 1` fails with `PACKAGE_ARENA_LIMIT_REACHED` (1MB is too small to load the stdlib). Otherwise covered by existing move-vm-runtime, move-unit-test, and sui-protocol-config suites.

## Release notes

- [x] Protocol: Adds `package_arena_size_in_bytes` (10MB, matching prior hardcoded behavior) to protocol version 136, bounding the arena a package is loaded into by the Move VM.
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: `sui move test` now accepts `--package-size <MB>` to bound the maximum size of a loaded package during unit tests.
- [ ] Rust SDK:
- [ ] Indexing Framework:


sui-execution/latest/sui-adapter/src/adapter.rs
```diff
@@ -60,6 +60,7 @@ mod checked {
                 vector_len_max: protocol_config.max_move_vector_len(),
                 max_value_nest_depth: protocol_config.max_move_value_depth_as_option(),
                 hardened_otw_check: protocol_config.hardened_otw_check(),
+                ..Default::default()
```
tzakian: no, please no. No no no no no no _NO_ NO **_NO_**

Default has no place here. **_NO PLACE_**.

Plz add a protocol config field for this, and use that here. You can either set it from the beginning to 10MB, or set it in the protocol config (and the `unwrap_or(DEFAULT_ARENA_SIZE)` will cover that until it activates.

tzakian (approved):
LGTM!

sui-execution/latest/sui-adapter/src/adapter.rs
cgswords: Removed

PR #27829, opened 2026-08-26
Add native vector splice, keep, slice, and reverse operations

dariorussi:
## Description

Move bytecode can only express bulk vector operations one element at a time, making common operations disproportionately expensive in gas and validator CPU. 
Add the selected native surface: `splice`, `keep`, `slice`, and `reverse`. 
`append`, `drain`, and `truncate` remain Move wrappers over the general primitives, and affected functions in `std::vector`, `std::ascii`, `std::string`, and `sui::bcs` reuse them. 
Gas parameters activate at protocol version 138.

## Test plan

- Move tests cover the new APIs, invalid ranges, ability constraints, and affected wrappers.
- VM tests cover primitive and container representations, deep-copy/drop behavior, exact gas formulas, and out-of-gas-before-mutation behavior.

## Release notes

- [x] Protocol: Adds native vector `splice`, `keep`, `slice`, and `reverse` operations and their gas schedule at protocol version 138.


crates/sui-framework/packages/move-stdlib/sources/vector.move
```diff
@@ -73,9 +73,10 @@ public fun reverse<Element>(v: &mut vector<Element>) {
 }
 
 /// Pushes all of the elements of the `other` vector into the `lhs` vector.
-public fun append<Element>(lhs: &mut vector<Element>, other: vector<Element>) {
-    other.do!(|e| lhs.push_back(e));
-}
+public native fun append<Element>(lhs: &mut vector<Element>, other: vector<Element>);
+
+/// Retain the first `n` elements of `v`.
+public native fun truncate<T: drop>(v: &mut vector<T>, n: u64);
```
tnowacki: Nit, since this is new it should maybe go at the end of the `public` section.

crates/sui-framework/packages/move-stdlib/sources/vector.move
```diff
@@ -73,9 +73,10 @@ public fun reverse<Element>(v: &mut vector<Element>) {
 }
 
 /// Pushes all of the elements of the `other` vector into the `lhs` vector.
-public fun append<Element>(lhs: &mut vector<Element>, other: vector<Element>) {
-    other.do!(|e| lhs.push_back(e));
-}
+public native fun append<Element>(lhs: &mut vector<Element>, other: vector<Element>);
+
+/// Retain the first `n` elements of `v`.
```
tnowacki: What happens if `n > v.length()`?

crates/sui-framework/packages/move-stdlib/sources/vector.move
tnowacki, line 147: This can be rewritten in terms of reverse+truncate+reverse 
(though I'm fine if that is a follow up PR)

tnowacki, line 156: Similarly this is just a single call to `truncate` (though switch to that can be a follow up PR)

external-crates/move/crates/move-compiler/tests/move_2024/sui_mode/vector_truncate_requires_drop.move
```diff
@@ -0,0 +1,146 @@
+#[defines_primitive(vector)]
+module std::vector {
```
tnowacki: Do we really need this test?

tnowacki: I mean I'm fine with it but it seems a bit silly

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -2032,6 +2032,95 @@ impl VectorRef {
...
+            capacity: usize,
+        ) -> PartialVMResult<()> {
+            if lhs
+                .len()
+                .checked_add(rhs.len())
+                .is_none_or(|len| len > capacity)
+            {
+                return Err(partial_vm_error!(
+                    VECTOR_OPERATION_ERROR,
+                    "vector size limit is {capacity}",
```
tnowacki, line 2112: Is this an invariant violation? I'm confused

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -2032,6 +2032,95 @@ impl VectorRef {
...
+                )
+                .with_sub_status(VEC_SIZE_LIMIT_REACHED));
```
tnowacki, line 2114: Though I think this check is missing. We need to check that `lhs.len() + other.len()` is less than the max vector size

crates/sui-framework/packages/move-stdlib/sources/vector.move
tnowacki: I do wonder if we want an efficient form of `drain` to go with this, which is kind of like the non-`drop` version of this
```
public native fun drain<T>(v: &mut vector<T>, i: u64, j: u64): vector<T>
```

Still sort of spinning trying to think of what the right set of primitives here is

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
```diff
@@ -311,6 +311,113 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(ty_args.len() == 1);
+    debug_assert!(args.len() == 2);
+
+    native_charge_gas_early_exit!(context, gas_params.base);
+
+    let other = pop_arg!(args, Vector);
+    let lhs = pop_arg!(args, VectorRef);
+
+    let len = other.elem_len()? as u64;
```
tnowacki:
```suggestion
    let other_len = other.elem_len()? as u64;
```
nit, just a bit easier to follow. I Would assume that `len` is `lhs.len() + other.len()`

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
```diff
@@ -311,6 +311,113 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+    // TODO: define charges
+    const WORD_SIZE: u64 = 8;
+    native_charge_gas_early_exit!(context, InternalGas::new(WORD_SIZE.saturating_mul(len)));
```
tnowacki: What exactly is this word size charge? 
Do we just want another gas parameter for a per element charge?
```

File: external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs [hunk 8]
```diff
@@ -311,6 +311,113 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+    let new_len = checked_as!(pop_arg!(args, u64), usize)?;
+    let v = pop_arg!(args, VectorRef);
+    let ty = ty_args.safe_get(0)?;
+    let current_len = match v.len(ty).and_then(|value| match value {
+        Value::U64(len) => Ok(len),
+        _ => Err(partial_vm_error!(
+            UNKNOWN_INVARIANT_VIOLATION_ERROR,
+            "vector::length must return a u64"
+        )),
+    }) {
```
tnowacki, line 401: Is this what `elem_len` is?

external-crates/move/crates/move-vm-runtime/src/unit_tests/value_tests.rs
```diff
@@ -456,6 +456,366 @@ fn vector_push_back_and_pop_container_vec() -> PartialVMResult<()> {
     Ok(())
 }
 
+#[test]
+fn vector_append_prim_vec() -> PartialVMResult<()> {
```
tnowacki: Should we test all prims?

external-crates/move/crates/move-vm-runtime/src/unit_tests/value_tests.rs
```diff
@@ -456,6 +456,366 @@ fn vector_push_back_and_pop_container_vec() -> PartialVMResult<()> {
...
+    assert!(vr.len(&ty)?.equals(&Value::u64(2))?);
+
+    let second: Reference = VMValueCast::cast(vr.borrow_elem(1, &ty)?)?;
+    assert!(second.read_ref()?.equals(&Value::vector_u8([3]))?);
+
+    Ok(())
+}
+
+#[test]
+fn vector_append_all_primitive_specializations() -> PartialVMResult<()> {
```
tnowacki: Is this covered by `vector_append_prim_vec`?

external-crates/move/crates/move-vm-runtime/src/unit_tests/value_tests.rs
```diff
@@ -456,6 +456,366 @@ fn vector_push_back_and_pop_container_vec() -> PartialVMResult<()> {
...
+            Type::Address
+        )?,
+        2
+    );
+
+    Ok(())
+}
+
+#[test]
+fn vector_append_exceeds_capacity_errors() -> PartialVMResult<()> {
```
tnowacki: I still don't really understand this error

external-crates/move/crates/move-vm-runtime/src/unit_tests/value_tests.rs
```diff
@@ -456,6 +456,366 @@ fn vector_push_back_and_pop_container_vec() -> PartialVMResult<()> {
...
+        VectorSpecialization::Container,
+        [Value::vector_u8([3])],
+    )?)?;
+
+    assert!(vr.append(other, &Type::U8, 100).is_err());
+    Ok(())
+}
+
+#[test]
+fn vector_truncate_prim_vec() -> PartialVMResult<()> {
```
tnowacki: Is this covered by `vector_truncate_all_primitive_specializations`?

tnowacki (changes requested):
Overall looks good, but I think this capacity error is incorrect? And requesting  changes for that

external-crates/move/crates/move-compiler/tests/move_2024/sui_mode/vector_truncate_requires_drop.move
dariorussi: I have no idea, meaning I put them there asking myself the same question and decided to leave them waiting for feedback....

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
dariorussi: I left a TODO on both charging.
Here we have a couple of options (maybe more) under the assumption that append moves the vector portion and nothing else. No deep walk or traversal, just a memcpy on all cases with 256 being the biggest vector elem size we have.
- we use a vector elem size to charge, so we charge `elem_size() * len` with elem size being: 1, 2, 4, 8, 16, 32. That felt the best way but  I had but it introduced a new function (very simple) which I thought may not make everybody happy.
- use a default value for all and call it a day. Very simple and what it is here. 

Happy to hear opinions on what we would like to do and happy to add back the `elem_size` function

crates/sui-framework/packages/move-stdlib/sources/vector.move
dariorussi: moved down but before the macros (?)

crates/sui-framework/packages/move-stdlib/sources/vector.move
dariorussi: changed comment

crates/sui-framework/packages/move-stdlib/sources/vector.move
dariorussi: is it better that way? I have not done this yet and I'll take a closer look at costs

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
dariorussi: that is fair, this is combining the "overflow" with the "over vector size, which maybe should be different errors.
I have split the 2 errors to make it clear

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
dariorussi: from the code:
`lhs.len().checked_add(rhs.len())` -> `lhs.len()` + `rhs.len()` with overflow included
`is_none_or(|len| len > capacity)` -> deals with overflow or length bigger than capacity
is that what you were asking?

crates/sui-framework/packages/move-stdlib/sources/vector.move
dariorussi: this is something we talked about and part of a coming investigation. I think there is something there that could help a few scenarios, we'll see. 
Others should comment here if there are opinions

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
dariorussi: no, this is the impossible condition of vector length coming back as a non u64

external-crates/move/crates/move-compiler/tests/move_2024/sui_mode/vector_truncate_requires_drop.move
dariorussi: to be more clear. I realized tests that verified the correctness of the API were missing (well the type correctness so to speak), but it was not clear to me they should be added.
The tests themselves are interesting in that they would guarantee that a change to the API would break tests. However I am not sure we have those tests for any of the API that have specific abilities requirements. But this is a framework vector API so maybe more critical than others.
And I went back and forth with that.
Until I realized I'd ask for feedback and put the tests here for someone to make the call

external-crates/move/crates/move-vm-runtime/src/unit_tests/value_tests.rs
dariorussi: take a look and see if the changes are what you had in mind

external-crates/move/crates/move-vm-runtime/src/unit_tests/value_tests.rs
dariorussi: take a look and see if the changes are what you had in mind

external-crates/move/crates/move-vm-runtime/src/unit_tests/value_tests.rs
dariorussi: reworked the test a bit, also in light of the change to the error.
I think it should be more clear now. Please notice that capacity in the test is very low instead of the default `vector_len_max`

external-crates/move/crates/move-vm-runtime/src/unit_tests/value_tests.rs
dariorussi: take a look and see if the changes are what you had in mind

crates/sui-framework/packages/move-stdlib/sources/vector.move
dariorussi: actually thinking some more about this it may be a good addition to this PR?

dariorussi: and the more I think about it the more I think that is the API we want so I will make one

crates/sui-framework/packages/move-stdlib/sources/vector.move
tnowacki: I think we can wait to touch these until we have the move_range API we discussed yesterday.

external-crates/move/crates/move-compiler/tests/move_2024/sui_mode/vector_truncate_requires_drop.move
tnowacki: Yeah I really don't think we need this test. If you wanted to be very paranoid, I would rather somehow debug assert inside the native implementation for `truncate` if possible. 

There is nothing here you are really testing other than the compiler works the way it should work

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
tnowacki: I think I got a bit confused by the word. Should we replace `capacity` with `limit` or `size_limit` or something similar? 
When I hear "capacity" in terms of vectors, I immediately think of Rust's notion of [capacity](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.capacity), but I take it this is the limit from the vm config?

dariorussi:
changed all of this to go with a more proper set of primitives as discussed

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -1879,6 +1879,63 @@ pub const POP_EMPTY_VEC: u64 = NFE_VECTOR_ERROR_BASE + 2;
...
+fn checked_range<T>(v: &[T], i: usize, j: usize) -> PartialVMResult<&[T]> {
+    let len = v.len();
+    v.get(i..j).ok_or_else(|| {
+        partial_vm_error!(
+            VECTOR_OPERATION_ERROR,
+            "range [{i}, {j}) out of bounds for vector of length {len}",
+        )
+        .with_sub_status(INDEX_OUT_OF_BOUNDS)
+    })
+}
```
cgswords, line 1918: Why!? Why not just call `check_vector_range`? And if these are meant to be the same error, the error string should be a `const`. I don't know, this double-design feels bad.

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -1879,6 +1879,63 @@ pub const POP_EMPTY_VEC: u64 = NFE_VECTOR_ERROR_BASE + 2;
...
+
+/// Vec-to-Vec splice on pre-validated bounds: removes `v[i..j)`, inserts `other` at `i`,
+/// returns the removed elements.
+fn splice_impl<T>(v: &mut Vec<T>, i: usize, j: usize, other: Vec<T>) -> Vec<T> {
```
cgswords, line 1922: Well... not this.

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -2032,6 +2089,185 @@ impl VectorRef {
...
+            V::PrimVec(PV::VecU64(xs)) => keep_vec!(xs),
+            V::PrimVec(PV::VecU128(xs)) => keep_vec!(xs),
+            V::PrimVec(PV::VecU256(xs)) => keep_vec!(xs),
+            V::PrimVec(PV::VecBool(xs)) => keep_vec!(xs),
+            V::PrimVec(PV::VecAddress(xs)) => keep_vec!(xs),
+            V::Vec(items) => keep_vec!(items),
+        }
+
+        Ok(())
+    }
```
cgswords, line 2128:
```suggestion
  pub fn keep(&self, start: usize, end: usize, type_param: &Type) -> PartialVMResult<()> {
      let value = &mut *self.0.try_borrow_mut()?;
      check_elem_layout(type_param, value)?;
      let vec = value.vector_mut_ref()?;
      check_vector_range(start, end, vec.len())?;

      fn keep_range<T>(v: &mut Vec<T>, start: usize, end: usize) {
          v.truncate(end);
          v.drain(..start);
      }

      match_vec_ref_container!(
          (mut vec)
          prim r => keep_range(r, start, end);
          vec r => keep_range(r, start, end);
      );
      Ok(())
  }
```

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -2032,6 +2089,185 @@ impl VectorRef {
...
+            V::PrimVec(PV::VecU64(xs)) => reverse_vec!(xs),
+            V::PrimVec(PV::VecU128(xs)) => reverse_vec!(xs),
+            V::PrimVec(PV::VecU256(xs)) => reverse_vec!(xs),
+            V::PrimVec(PV::VecBool(xs)) => reverse_vec!(xs),
+            V::PrimVec(PV::VecAddress(xs)) => reverse_vec!(xs),
+            V::Vec(items) => reverse_vec!(items),
+        }
+
+        Ok(())
+    }
```
cgswords, line 2141:
```suggestion
  pub fn reverse(&self, type_param: &Type) -> PartialVMResult<()> {
      let value = &mut *self.0.try_borrow_mut()?;
      check_elem_layout(type_param, value)?;
      let vec = value.vector_mut_ref()?;
      match_vec_ref_container!(
          (mut vec)
          prim r => r.reverse();
          vec r => r.reverse();
      );
      Ok(())
  }
```

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -2032,6 +2089,185 @@ impl VectorRef {
...
+                slice_vec!(xs, |v| Value::PrimVec(PV::VecAddress(v)))
+            }
+            V::Vec(items) => Vector(Value::Vec(
+                checked_range(items, i, j)?
+                    .iter()
+                    .map(|m| m.copy_value())
+                    .collect(),
+            )),
+        })
+    }
```
cgswords:
```suggestion
  pub fn slice(&self, i: usize, j: usize, type_param: &Type) -> PartialVMResult<Vector> {
      let value = &*self.0.try_borrow()?;
      check_elem_layout(type_param, value)?;
      Ok(Vector(match value.vector_ref()?.0 {
          VectorMatch::PrimVec(prim_vec) => Value::PrimVec(map_prim_vec!(
              prim_vec,
              xs,
              checked_range(xs, i, j)?.to_vec()
          )),
          VectorMatch::Vec(items) => Value::Vec(
              checked_range(items, i, j)?
                  .iter()
                  .map(MemBox::copy_value)
                  .collect(),
          ),
      }))
  }
```

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
cgswords:
```
/// Vec-to-Vec splice on pre-validated bounds (`i <= j <= v.len()`): removes
/// `v[i..j)`, inserts `other` at `i`, returns the removed elements.
fn splice_impl<T>(v: &mut Vec<T>, i: usize, j: usize, mut other: Vec<T>) -> Vec<T> {
    // Whole-vector replacement: swap buffers, move nothing.
    if i == 0 && j == v.len() {
        return std::mem::replace(v, other);
    }

    // Same-length replacement: swap in place; `other`'s buffer becomes the result.
    if other.len() == j - i {
        v[i..j].swap_with_slice(&mut other);
        return other;
    }

    v.splice(i..j, other).collect()
}
```

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -2032,6 +2089,185 @@ impl VectorRef {
...
+    /// Removes `v[i..j)`, inserts all elements of `other` at position `i`, and returns the
+    /// removed elements. The vector grows or shrinks by `other.len() - (j - i)`.
+    pub fn splice(
+        &self,
+        i: usize,
+        j: usize,
+        other: Vector,
+        type_param: &Type,
+        capacity: u64,
+    ) -> PartialVMResult<Vector> {
```
cgswords: These all should return `Value`, not `Vector`, for consistency.

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
```diff
@@ -2032,6 +2089,185 @@ impl VectorRef {
...
+                    .iter()
+                    .map(|m| m.copy_value())
+                    .collect(),
+            )),
+        })
+    }
+
+    /// Removes `v[i..j)`, inserts all elements of `other` at position `i`, and returns the
+    /// removed elements. The vector grows or shrinks by `other.len() - (j - i)`.
+    pub fn splice(
```
cgswords, line 2165:
```
pub fn splice(&self, i: usize, j: usize, other: Vector, type_param: &Type, limit: u64)
      -> PartialVMResult<Vector>
  {
      let value = &mut *self.0.try_borrow_mut()?;
      check_elem_layout(type_param, value)?;
      check_elem_layout(type_param, &other.0)?;
      let vec = value.vector_mut_ref()?;
      check_vector_range(i, j, vec.len())?;
      let other_len = other.elem_len()?;
      let new_len = j
          .checked_sub(i)
          .and_then(|removed| vec.len().checked_sub(removed))
          .and_then(|kept| kept.checked_add(other_len))
          .ok_or_else(|| partial_vm_error!(UNKNOWN_INVARIANT_VIOLATION_ERROR, "vector length overflow during splice"))?;
      check_vector_limit(new_len, limit)?;

      let mismatch = || partial_vm_error!(UNKNOWN_INVARIANT_VIOLATION_ERROR, "vector::splice on mismatched vector representations");
      Ok(Vector(match (vec.0, other.0) {
          (VectorMatch::PrimVec(lhs), Value::PrimVec(rhs)) => Value::PrimVec(
              map_prim_vec_pair!(lhs, rhs, xs, ys, xs.splice(i..j, ys).collect(), mismatch())?,
          ),
          (VectorMatch::Vec(lhs), Value::Vec(rhs)) => Value::Vec(lhs.splice(i..j, rhs).collect()),
          _ => return Err(mismatch()),
      }))
  }
```

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
```diff
@@ -311,6 +311,262 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+ *   The VM gas meter separately charges the returned, removed vector by its deep abstract size.
+ *
+ **************************************************************************************************/
+#[derive(Debug, Clone)]
+pub struct SpliceGasParameters {
+    pub base: InternalGas,
+    pub per_elem: InternalGasPerArg,
+}
+
+pub(crate) fn splice_work(len: u64, i: u64, j: u64, n_in: u64) -> u64 {
```
cgswords:
```suggestion
pub(crate) fn splice_gas_count(len: u64, i: u64, j: u64, n_in: u64) -> u64 {
```

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
```diff
@@ -311,6 +311,262 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+    if i > j || j > len {
+        return 0;
+    }
```
cgswords: This logic is repeated so many times now...

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
```diff
@@ -311,6 +311,262 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(ty_args.len() == 1);
+    debug_assert!(args.len() == 1);
+
+    native_charge_gas_early_exit!(context, gas_params.base);
+
+    let v = pop_arg!(args, VectorRef);
+    let ty = ty_args.safe_get(0)?;
+    let len = match vector_len(&v, ty) {
```
cgswords: Why are we calling `vector_len` like this?

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
```diff
@@ -311,6 +311,262 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+    native_charge_gas_early_exit!(context, gas_params.base);
+
+    let end_arg = pop_arg!(args, u64);
+    let start_arg = pop_arg!(args, u64);
+    let end = checked_as!(end_arg, usize)?;
+    let start = checked_as!(start_arg, usize)?;
+    let v = pop_arg!(args, VectorRef);
+    let ty = ty_args.safe_get(0)?;
+
+    let len = match vector_len(&v, ty) {
```
cgswords: Why are we calling `vector_len` like this?

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
```diff
@@ -311,6 +311,262 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+    let j_arg = pop_arg!(args, u64);
+    let i_arg = pop_arg!(args, u64);
+    let j = checked_as!(j_arg, usize)?;
+    let i = checked_as!(i_arg, usize)?;
+    let v = pop_arg!(args, VectorRef);
+
+    // charge according to the moved elements (relocation)
+    let n_in = other.elem_len()? as u64;
+    let ty = ty_args.safe_get(0)?;
+    let len = match vector_len(&v, ty) {
```
cgswords: Why are we calling `vector_len` like this?

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
```diff
@@ -311,6 +311,262 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+
+pub fn make_native_splice(gas_params: SpliceGasParameters) -> NativeFunction {
+    Arc::new(
+        move |context, ty_args, args| -> PartialVMResult<NativeResult> {
+            native_splice(&gas_params, context, ty_args, args)
+        },
+    )
+}
+
+fn vector_len(v: &VectorRef, ty: &Type) -> PartialVMResult<u64> {
```
cgswords: This is private, but we end up boxing this up into a value just to unbox it or panic about it. This is a strange pattern and feels like an antipattern to me.

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
```diff
@@ -311,6 +311,262 @@ pub fn make_native_swap(gas_params: SwapGasParameters) -> NativeFunction {
...
+    // get arguments from the Move call frame
+    let other = pop_arg!(args, Vector);
+    let j_arg = pop_arg!(args, u64);
+    let i_arg = pop_arg!(args, u64);
+    let j = checked_as!(j_arg, usize)?;
+    let i = checked_as!(i_arg, usize)?;
+    let v = pop_arg!(args, VectorRef);
+
+    // charge according to the moved elements (relocation)
+    let n_in = other.elem_len()? as u64;
```
cgswords: No `as`

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
dariorussi: Pulled the error into a single `range_out_of_bounds(i, j, len)`.
See if that covers it, I like it more but I am not sure that was all

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
dariorussi: This was done so we mimic gas operation best. The `std::splice` performs element by element operations.
We'll talk more offline and, fwiw, I am fine either way, though I do believe what it's there is a better way

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
dariorussi: Nice. 
Liked the `truncate(end)` instead of `drain(end..)`.

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
dariorussi: done

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
dariorussi: done

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
dariorussi: done

external-crates/move/crates/move-vm-runtime/src/execution/values/values_impl.rs
dariorussi: I added a `map_prim_vec_pair!` next to `map_prim_vec!`, see if it's even close to what you had in mind

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
dariorussi: Renamed. `keep_work` became `keep_gas_count` too, so the pair stays consistent.

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
dariorussi: made a `valid_range(len, i, j)` shared by both?

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
dariorussi: `vector_len` is gone but see if it makes sense

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
dariorussi: `vector_len` is gone but see if it makes sense

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
dariorussi: `vector_len` is gone but see if it makes sense

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
dariorussi: Good call. 
Added `VectorRef::elem_len(&self, ty) -> PartialVMResult<usize>` and made `len` delegate to it, so no boxes and deleted `vector_len`. See what's there and let me know

external-crates/move/crates/move-vm-runtime/src/natives/move_stdlib/vector.rs
dariorussi: used `checked_as!`

crates/sui-framework/packages/move-stdlib/sources/ascii.move
```diff
@@ -78,15 +78,13 @@ public fun append(string: &mut String, other: String) {
 /// Insert the `other` string at the `at` index of `string`.
```
damirka, line 78: Should we maybe do this and other modules in a follow up?

crates/sui-framework/packages/move-stdlib/sources/vector.move
```diff
@@ -59,22 +59,12 @@ public native fun destroy_empty<Element>(v: vector<Element>);
...
-
-    let mut front_index = 0;
-    let mut back_index = len - 1;
-    while (front_index < back_index) {
-        v.swap(front_index, back_index);
-        front_index = front_index + 1;
-        back_index = back_index - 1;
-    }
-}
+public native fun reverse<Element>(v: &mut vector<Element>);
```
damirka, line 62: Whoa, this is cool!

crates/sui-framework/packages/move-stdlib/sources/vector.move
```diff
@@ -288,6 +269,38 @@ public fun flatten<T>(v: vector<vector<T>>): vector<T> {
     r
 }
 
+/// Retains only `v[start..end)`, moving the retained elements to the beginning of `v` and
+/// dropping all others.
+/// Aborts if `start > end` or `end > v.length()`.
+public native fun keep<T: drop>(v: &mut vector<T>, start: u64, end: u64);
```
tnowacki, line 275: 🚲 
I worry about the name of this function a bit, just in that it doesn't match the behavior API exactly of similar operations. Just that truncate, skip, take all work over "n" elements. But keep and drain work over ranges. 

🚲 🚲 🏠 🚲 🚲 
For maximal bikeshedding, I wonder if we should have 
`keep_range`
`drain_range`
`copy_range` (which is currently named `slice`)

PR #27830, opened 2026-08-27
[core] refactor simulate transaction in core

stefan-mysten:
## Description 

This PR refactors out the `simulate_transaction` from `AuthorityState` into its own module, as preparation for future work on reusing this code for `simulacrum`.

The best way to review this is to go through each commit, as each commit should be only minimal and logical changes, basically moving things around.

Stack

## Test plan 

Existing CI.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-core/src/authority.rs
```diff
@@ -1065,6 +1066,45 @@ pub struct AuthorityState {
     transaction_deny_config_manager: Arc<TransactionDenyConfigManager>,
 }
 
+/// Run deny list checks and process funds withdrawals before loading input objects.
```
amnn: Do you imagine this will eventually be used as a more public function? If not, it may not be worth factoring this out? I'm also not used to seeing private functions here (between type defs and impl blocks), but I'm also not sure of the conventions in this file.

crates/sui-core/src/authority/transaction_simulation.rs
```diff
@@ -0,0 +1,282 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use super::*;
+
+pub(super) struct SimulationEpochContext<'a> {
```
amnn: I would initially err on the side of having way too many parameters, and use the refactor to guide what should be grouped together and what their common thread is.

Names like "XContext" and "YServices" are really just thinly veiled "bags of things". Sometimes that's really what you need, but (a) I suspect it's too early to tell right now and (b) pre-grouping them like this may make it more difficult to regroup them later.

Also, I'm looking at many of these grouping structs, and their names all start with "Simulation..." but their content doesn't seem specific to simulation at all -- this seems like the same information execution would need regarding epochs, denials and backing stores. In some cases, I think there are already grouping traits that cover parts of `SimulationService`.

amnn (commented):
I think this is a reasonable place to start, thanks @stefan-mysten. In the inline comments I make an argument that you can and maybe should do even less to begin with, but it's not blocking.

I'd also like someone who works more regularly in the `sui-core` codebase to look and give the final stamp.

crates/sui-core/src/authority.rs
stefan-mysten: In some form, this will be needed by simulacrum too, but agreed that maybe it's too early to refactor it.

stefan-mysten: But also the simulate transaction function in the new module calls this function, so this must be out of `AuthorityState` for this to work.

crates/sui-core/src/authority/transaction_simulation.rs
stefan-mysten: Reworked this part, thank you!

crates/sui-core/src/authority.rs
```diff
@@ -2473,260 +2436,43 @@ impl AuthorityState {
...
-        let dev_inspect = checks.disabled();
-        if dev_inspect && self.config.dev_inspect_disabled {
+        if checks.disabled() && self.config.dev_inspect_disabled {
             return Err(SuiErrorKind::UnsupportedFeatureError {
                 error: "simulate with checks disabled is not allowed on this node".to_string(),
             }
             .into());
         }
 
-        // Reject coin reservations in gas payment when the execution engine
```
stefan-mysten, line 2484: This and below it's just moved to the new `transaction_simulation` module, in a `simulate_transaction` function.

crates/sui-core/src/authority.rs
```diff
@@ -2473,260 +2436,43 @@ impl AuthorityState {
...
+            .effective_config()
+            .load();
+        let epoch_data = epoch_store.epoch_start_config().epoch_data();
+        let suggested_gas_price = self
+            .congestion_tracker
+            .get_suggested_gas_prices(&transaction);
 
-        let (mut kind, signer, gas_data) = transaction.execution_parts();
-        let rewritten_inputs = rewrite_transaction_for_coin_reservations(
+        crate::transaction_simulation::simulate_transaction(
```
stefan-mysten, line 2455: Uhm, the way the diff shows this is unfortunate. Most of the logic is moved to the new module, and the right arguments are called into the function.

alex-mysten (approved):
Seems like a pretty straightforward 1:1 refactor. LGTM

tzakian (approved):

PR #27832, opened 2026-08-27
[adapter][hardening] Add additional hardening and linkage consistency checks and invariants

tzakian:
## Description 

Add a bunch of consistency and sanity checks/invariants around packages and publish/upgrade.

## Test plan 

CI

---

## Release notes

- [x] Protocol: Add additional invariant checks around packages.
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-protocol-config/src/lib.rs
```diff
@@ -4645,6 +4651,7 @@ impl ProtocolConfig {
                     if chain != Chain::Mainnet && chain != Chain::Testnet {
                         cfg.feature_flags.allowed_proposers = true;
                     }
+                    cfg.feature_flags.harden_linkage_consistency = true;
```
tnowacki, line 4654: I have something in the works to put up shortly after this. My dog put all new additions above the `if` here. Do we have style for that? 
Not that it matters I was just thinking about it for collisions reasons

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
```diff
@@ -154,6 +156,46 @@ pub struct DeserializedPackage {
...
+///
+/// NB: we presuppose that a function named `init` is the module's initializer. If it does not
+/// conform to the `init` signature requirements the entry points verifier rejects the publish
+/// later, failing the transaction as a whole.
+pub(crate) fn module_has_init(module: &CompiledModule) -> bool {
+    module.function_defs().iter().any(|func_def| {
+        let handle = module.function_handle_at(func_def.function);
+        module.identifier_at(handle.name) == INIT_FN_NAME
+    })
 }
```
tnowacki, line 199: I wonder if we should move this to some shared location, as we seem to keep writing this. Fine with it as it is though

tnowacki (approved):

crates/sui-protocol-config/src/lib.rs
tzakian: I don't think we do, I generally try (but am terrible at being consistent about) always adding at the end. 

That _usually_ helps with collisions, but it's a bit of a mixed bag. I might go update the protocol config skill actually.

sui-execution/latest/sui-adapter/src/static_programmable_transactions/loading/ast.rs
tzakian: Perhaps a `helpers.rs` or `utils.rs`? 😜

PR #27834, opened 2026-08-27
[ptbs] Add metering and limits for references in PTBs

tnowacki:
## Description 

- Returning references in PTBs is not yet enabled. 
- Before it is enabled, we will need metering/limits, which is done here

## Test plan 

- 👀 not yet enabled

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-protocol-config/src/lib.rs
```diff
@@ -4642,6 +4663,11 @@ impl ProtocolConfig {
                 136 => {
                     cfg.feature_flags.ptb_tx_context_restrictions = true;
 
+                    cfg.translation_per_live_reference_charge = Some(1);
+                    cfg.max_ptb_live_references = Some(64);
+                    cfg.max_ptb_returned_references = Some(16);
+                    cfg.max_ptb_total_returned_references = Some(256);
+
```
tzakian, line 4676: I see you have gone for neither the end, nor the beginning, but right smack-dab in the middle.

Bold move. Bold move...

external-crates/move/crates/move-regex-borrow-graph/src/collections.rs
```diff
@@ -80,6 +80,7 @@ impl<Loc: Copy, Lbl: Ord + Clone + fmt::Display> Graph<Loc, Lbl> {
         initial_refs: impl IntoIterator<Item = (K, Loc, /* is_mut */ bool)>,
     ) -> Result<(Self, BTreeMap<K, Ref>)> {
         let mut map = BTreeMap::new();
+        let canonical_reference_capacity = canonical_reference_capacity.min(512);
```
tzakian: Thoughts on pulling this number up into a constant since it is used here and in the `graph_map` and they should probably stay in sync?

tzakian (approved):
Couple little nits, but looks good!

crates/sui-protocol-config/src/lib.rs
tnowacki: 🧠 
I'm hoping this minimizes any conflicts lol

PR #27836, opened 2026-08-27
[graphql-stream] unify checkpoint subscription onto the scan-then-live driver, backed by ListCheckpoints [24/n]

tpham-mysten:
## Description

Checkpoint subscriptions had their own backfill and resume path, separate from the shared scan-then-live driver that transactions and events already use. This backs the checkpoint backfill with an unfiltered `ListCheckpoints` scan, one sequential read, instead of a `GetCheckpoint` per checkpoint, and then folds checkpoints onto the same `Subscribable` driver so all three feeds share one backfill-to-live seam, resume handling, the concurrency admission cap, and far-ahead-start rejection.

Key changes:
- Back the checkpoint backfill with an unfiltered `ListCheckpoints` scan (one sequential read) instead of a `GetCheckpoint` per checkpoint.
- Implement `Subscribable` for `Checkpoint`, so transactions, events, and checkpoints all run on the shared scan-then-live driver and share one resume path.
- Move deriving the resume point, rejecting a far-ahead start, and admitting the subscriber into the shared `subscribe`, so each resolver just fetches its context and delegates.
- `build_node` builds its own scope from the scanned payload, so a backfilled checkpoint resolves from its processed data in memory the same way a live one does.
- Remove the now-dead per-subscriber scan config (`per_subscriber_scan_max_qps`, `per_subscriber_scan_max_concurrent_fetches`).

## Test plan

Covered by the `graphql_subscription` e2e suite across transactions, events, and checkpoints, including resume by cursor and by `afterCheckpoint`, both-args-uses-max, far-ahead rejection, and long backfill.

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-indexer-alt-graphql/src/api/subscription/checkpoints.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Fully process the scanned checkpoint and carry it on the node, resolving against it in memory.
+    /// The scan supplied the data cheaply as one sequential read; processing it here is the same work
+    /// the live path does off the broadcast, so a backfilled node resolves identically to a live one.
+    fn build_node(
+        caches: &Arc<StreamedCaches>,
+        resolver_limits: &sui_package_resolver::Limits,
+        payload: &v2::Checkpoint,
+    ) -> Result<Self, RpcError> {
+        let processed = Arc::new(process_checkpoint(payload.clone())?);
```
amnn, line 63: Is the pattern of processing a checkpoint and building a scope universal, and if so, should it be moved into whatever calls `build_node`?

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
```diff
@@ -91,6 +99,13 @@ pub(super) trait Subscribable {
         resolver_limits: &sui_package_resolver::Limits,
         filter: &Self::Filter,
     ) -> Result<Vec<Edge<String, Self::Item, EmptyFields>>, RpcError>;
+
+    /// The checkpoint a resume cursor points at, letting the driver derive the resume point from an
+    /// `after` cursor without knowing the concrete cursor type.
+    fn cursor_checkpoint(cursor: &Self::Cursor) -> u64;
```
amnn: There's maybe something worth exploring here that `Self::Cursor` implements `TryFrom<CursorToken>` which already exposes a way to get a `checkpoint` from its `position`, so maybe we don't need to introduce this new API surface.

amnn (approved):
Nice clean up!

nickvikeras (approved):

crates/sui-indexer-alt-graphql/src/api/subscription/checkpoints.rs
tpham-mysten: It isn't universal. Transactions and events build Scope::for_indexed straight from the scanned payload with no checkpoint processing;

crates/sui-indexer-alt-graphql/src/api/subscription/scan_then_live.rs
tpham-mysten: Yes it's possible. I have added 3 new blocks to convert from CEvent / CTransaction / CCheckpoint back to `CursorToken`, so we can remove this `cursor_checkpoint` now

tpham-mysten (commented):
Thanks all for the review!

PR #27837, opened 2026-08-27
[core] Refactor transaction simulation api

stefan-mysten:
## Description 

This PR refactors two parts of the simulation logic in two clean commits (review them in that order, if possible):
- `TransactionInputLoader` which is a core concept and it's not available in simulacrum, thus a new trait called `SimulationInputLoader` is used with one function `read_objects_for_simulation`, and then the `TransactionInputLoader` type implements it. Simulacrum also implements this over its generic store in the next PR (see stack or #27838).
- existing `CoinReservationResolver` owns its storage through an `Arc<dyn RuntimeObjectResolver>`. This works for the authority side, but simulacrum owns a generic store and only has `&Store` to provide. Thus having a borrowed type so that we can do `let coin_resolver = BorrowedCoinReservationResolver::new(store)` on simulacrum side, because its store implements `RunTimeObjectResolver`.
```rust
struct BorrowedCoinReservationResolver<'a> {
    runtime_object_resolver: &'a dyn RuntimeObjectResolver,
}
impl<'a> BorrowedCoinReservationResolver<'a> {
    /// Create a coin-reservation resolver over borrowed storage.
    pub fn new(runtime_object_resolver: &'a dyn RuntimeObjectResolver) -> Self {
        Self {
            runtime_object_resolver,
        }
    }
}
```

Stack

## Test plan 

Existing CI

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-core/src/transaction_simulation.rs
+++ b/crates/sui-core/src/transaction_simulation.rs
@@ -13,7 +13,7 @@ use sui_config::{
 use sui_execution::Executor;
 use sui_transaction_checks::{check_dev_inspect_input, check_transaction_input};
 use sui_types::{
-    base_types::{EpochId, ObjectID},
+    base_types::{EpochId, ObjectID, ObjectRef},
     coin_reservation::{CoinReservationResolverTrait, ParsedDigest},
     digests::{ChainIdentifier, TransactionDigest},
     effects::TransactionEffectsAPI,
@@ -28,7 +28,10 @@ use sui_types::{
     storage::{
         BackingPackageStore, BackingStore, TrackingBackingStore, get_transaction_object_set,
     },
-    transaction::{ObjectReadResult, TransactionData, TransactionDataAPI, TxValidityCheckContext},
+    transaction::{
+        InputObjectKind, InputObjects, ObjectReadResult, ReceivingObjects, TransactionData,
+        TransactionDataAPI, TxValidityCheckContext,
+    },
     transaction_executor::{SimulateTransactionResult, TransactionChecks},
 };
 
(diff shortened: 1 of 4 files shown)


crates/sui-core/src/transaction_simulation.rs
```diff
@@ -38,11 +41,23 @@ use crate::{
         transaction_rewriting::rewrite_transaction_for_coin_reservations,
     },
     authority::{DEV_INSPECT_GAS_COIN_VALUE, pre_object_load_checks},
-    transaction_input_loader::TransactionInputLoader,
     transaction_outputs::unchanged_loaded_runtime_objects,
 };
 
-pub(crate) fn simulate_transaction(
+/// Load transaction inputs for simulation without preparing them for committed execution.
+pub trait SimulationInputLoader {
```
stefan-mysten, line 48: `simulacrum` has a generic store, so this trait allows us to load the objects there required for execution.

crates/sui-types/src/coin_reservation.rs
```diff
@@ -263,31 +308,42 @@ impl CoinReservationResolver {
...
+            self.runtime_object_resolver.as_ref(),
+            sender,
+            coin_reservation,
+            accumulator_version,
+        )
+    }
+}
+
+/// Borrow a runtime object resolver for coin-reservation lookups.
+pub struct BorrowedCoinReservationResolver<'a> {
```
stefan-mysten, line 321: `CoinReservationResolver::new` requires `Arc<dyn RuntimeObjectResolver>`, and `simulacrum` has only a reference to its store. This adapter allows us to do this in `simulacrum` code:
```rust
  pub(crate) fn simulate_transaction<S: SimulatorStore + Send + Sync>(
        &self,
        store: &S,
     [.... some other code here .... ] {
     let coin_resolver = BorrowedCoinReservationResolver::new(store);

}
```

crates/sui-types/src/coin_reservation.rs
```diff
@@ -205,6 +205,73 @@ pub fn encode_object_ref(
         .encode(version, chain_identifier)
 }
 
+fn get_owner_and_type_for_object_impl(
```
stefan-mysten, line 208: These are basically moved from CoinReservationResolver impl such that we can call them from the new 
[Borrow impl](https://github.com/MystenLabs/sui/pull/27837/changes#diff-3706cd26efae5cc853c3894391c379bc8d46f621431d5b7fe6859a4bf61749b2R334-R335) below.

tzakian (approved):
Looks good to me!

alex-mysten (approved):

PR #27838, opened 2026-08-27
[simulacrum] Add transaction simulation

stefan-mysten:
## Description 

Wire up the `simulate_transaction` on top of the refactoring of the simulate transaction logic in `sui-core`. Note that `simulacrum` already depended on `sui-core` for `MockCheckpointBuilder`, so as a first step this seems ok.

Concretely, this PR implements:
- a `SimulationInputLoader` adapter backed by `SimulatorStore::read_objects_for_synchronous_execution.`                                                                                                                                                           
 - an `AccountFundsRead` adapter that reads address balances from accumulator objects.
 - uses a separate executor just for simulation, to avoid polluting the VM.
- the suggested gas price is RGP.

Stack
## Test plan 

New test in simulacrum to verify that simulate does not commit effects to the store.

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/simulacrum/src/epoch_state/simulation.rs
```diff
@@ -0,0 +1,137 @@
...
+    ) -> u128 {
+        AccumulatorValue::load_by_id::<U128>(self.0, version, *account_id)
+            .expect("simulator accumulator reads must succeed")
+            .map(|value| value.value)
+            .unwrap_or(0)
+    }
+}
+
+impl EpochState {
+    pub(crate) fn simulate_transaction<S: SimulatorStore + Send + Sync>(
```
stefan-mysten: Maybe this should live in the epoch_state.rs file instead.

crates/simulacrum/src/epoch_state/simulation.rs
```diff
@@ -0,0 +1,137 @@
...
+use sui_types::transaction::InputObjects;
+use sui_types::transaction::ReceivingObjects;
+use sui_types::transaction::TransactionData;
+use sui_types::transaction::TxValidityCheckContext;
+use sui_types::transaction_executor::SimulateTransactionResult;
+use sui_types::transaction_executor::TransactionChecks;
+
+use crate::SimulatorStore;
+
+use super::EpochState;
```
tzakian: why not `crate::EpochState`?

crates/simulacrum/src/epoch_state.rs
```diff
@@ -26,13 +26,17 @@ use sui_types::{
...
+
 pub struct EpochState {
     epoch_start_state: EpochStartSystemState,
     committee: Committee,
     protocol_config: ProtocolConfig,
     execution_metrics: Arc<ExecutionMetrics>,
     bytecode_verifier_metrics: Arc<BytecodeVerifierMetrics>,
     executor: Arc<dyn Executor + Send + Sync>,
+    /// Keeps arbitrary simulated packages out of committed execution's VM cache.
+    simulation_executor: Arc<dyn Executor + Send + Sync>,
```
tzakian, line 55: nice!

crates/simulacrum/src/epoch_state/simulation.rs
tzakian, line 1: Couple questions here:
1. Should this live under `epoch_state`? Feels like the wrong place for it to live. It living alongside the other modules feels better to me? But I realize this is a bit subjective.
2. The naming of the file (and related) is a bit unfortunate (though correct) since "simulate" inside of "simulacrum" is not confusing at all.... so may be worthwhile putting a top level comment here explaining what this module does (supports simulate_transaction/dev-inspect/dry-run on simulacrum).

stefan-mysten: I think you're right, I moved everything to EpochState. It was small enough that it made sense to do so.

crates/simulacrum/src/epoch_state/simulation.rs
stefan-mysten: Moved it.

tzakian (approved):
🎉

PR #27840, opened 2026-08-27
Schedule execution in causal order to prevent blocking deadlocks

mystenmark:
## Description

#27509 lets execution park its thread waiting for a value another execution will produce (undeclared dependencies: the clock, the accumulator root). With bounded concurrency, every slot can end up parked behind a writer that was never started. This PR makes that impossible by admitting transactions to execution in causal order:

- Every enqueued unit gets a **causal index** in enqueue order. Both sources (consensus handler, checkpoint executor) enqueue causally, so anything execution can wait for has a lower index. The scheduler now depends on enqueue order (`test_execution_with_dependencies`, which deliberately enqueued in reverse order, was updated).
- The driver admits when `in_flight < K || index == C+1`, where C is the watermark below which everything is done. The C+1 transaction can never block, so it is admitted even at the limit (one at a time, bounding in-flight at K+1): progress is guaranteed no matter how many admitted transactions are parked.
- Settlement transactions are index-less and admitted unconditionally: they can never block, and running them is what unblocks their waiters.
- Cross-source duplicate enqueues are rejected by an **accumulator-root-version watermark**: all transactions of a version are enqueued together, so a group at or below the watermark was already enqueued in full by the other source. Versionless groups (pre-accumulator epochs replayed from checkpoints, the end-of-epoch transaction) bypass dedup.
- The driver's semaphore and the #27509 execution-permit machinery (release-on-park hooks) are **deleted** — causal admission makes the deadlock they guarded against impossible. The `transaction_execution_delay` failpoint is retained, in the dispatch task where it has always been: it sits after admission, so if the epoch ends during the delay the slot drop still retires the causal index.
- Test configurations use a fixed K; the crashes-and-delays simtest randomizes it per node in 1..=8 via the `execution-concurrency-limit` fail point. Reachability asserts cover both admission branches, duplicate rejection, and actual thread-parking in `blocking_recv`. At each epoch boundary the scheduler checks that every index was retired, since a leaked index would silently disable the causal-next lane.

Full design rationale is in the module comments of `execution_driver.rs` and `execution_scheduler/causal_admission.rs`.

## Test plan

Unit tests in `execution_scheduler::causal_admission` cover the admission and dedup mechanics. All runs below are on this branch rebased on #27524, where execution blocks on implicit accumulator-root reads.

**Positive**
- Ran a 200-seed search of `test_composite_workload`: 200/200 passed.
- Reachability asserts show the tests exercise both admission paths (admission under the concurrency limit, and admission of the next transaction in causal order when the limit is full), rejection of duplicate version groups, and threads actually blocking in `blocking_recv`.
- The full `sui-core` unit suite (829/829), `object_funds_withdraw.move`, and the single-node benchmark smoke simtest pass.

**Negative**
- Removed the causal lane (patched `try_admit` to refuse admission at the limit) and ran 100 seeds: 4 seeds deadlocked and 1 livelocked, while the same binary with the lane passes 100/100. This proves the tests explore cases that would deadlock without the causal lane.
- Ran the same negative test before #27524: 100/100 passed and no thread ever blocked in `blocking_recv`. This shows no transaction blocked during execution before #27524, so those deadlock cases were unreachable until it landed.

**Found by the seed search**
- The search found a deadlock: a node that never receives a randomness round's signature lost the only executable copy of that round's update, because dedup rejected the checkpoint copy as a duplicate of the keyed enqueue that never resolves. Fixed by resolving keys from digest-carrying enqueues; the failing seed passes on the fix.
- The first full CI run found tests violating the enqueue contract: the test version-assignment helper claimed an accumulator version without ever settling, and directly executed withdrawals reached the driver without an index. The tests were changed to follow the contract.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:


crates/sui-core/src/execution_driver.rs
```diff
@@ -1,102 +1,238 @@
...
+                        info!("No more certificate will be received. Exiting executor ...");
+                        return;
+                    }
+                }
+                _ = causal_admission.changed() => {}
+                _ = &mut rx_execution_shutdown => {
+                    info!("Shutdown signal received. Exiting executor ...");
+                    return;
+                }
+            };
```
mystenmark: i think we can simplify this and make it closer to the original code. at the top of the loop:
1. recv from the channel, insert into heap
2. pop from heap and execute if possible

crates/sui-core/src/execution_driver.rs
```diff
@@ -168,16 +301,94 @@ pub async fn execution_process(
...
+                "Ignoring certificate from previous epoch."
+            );
+            waiting.pop().unwrap().0.retire(causal_admission);
+            continue;
+        }
+        {
+            let mut head = waiting.peek_mut().unwrap();
+            if !head.0.executed_checked {
+                head.0.executed_checked = true;
+                if authority.is_tx_already_executed(&digest) {
```
mystenmark: do we need this? the only way i can see how this could happen would be if we crash, and upon recovery begin re-executing a partially committed checkpoint. but in that case, can't we just let `try_execute_immediately` handle the already executed check?

crates/sui-core/src/execution_driver.rs
```diff
@@ -1,66 +1,224 @@
...
-                    txn_ready_time = pending_cert.stats.ready_time.unwrap();
-                    _executing_guard = pending_cert.executing_guard;
-                } else {
-                    // Should only happen after the AuthorityState has shut down and tx_ready_certificate
-                    // has been dropped by ExecutionScheduler.
-                    info!("No more certificate will be received. Exiting executor ...");
+        // Get the next admissible transaction, buffering arrivals meanwhile.
+        let (pending_cert, mut slot) = loop {
+            tokio::select! {
+                received = rx_ready_certificates.recv_many(&mut arrivals, RECV_BATCH_SIZE) => {
```
mystenmark: i think this would be simpler without recv_many

crates/sui-core/src/execution_driver.rs
```diff
@@ -168,13 +304,21 @@ pub async fn execution_process(
...
+                        }
                         authority
                             .metrics
                             .execution_driver_paused_transactions
                             .inc();
                     }
                 }
+                // Dropping the slot releases it and retires the causal index (unless a
+                // retry kept it alive), waking the driver once.
+                drop(slot);
```
mystenmark: do we need an explicit drop?

crates/sui-core/src/execution_driver.rs
```diff
@@ -1,66 +1,173 @@
...
+                let cert = waiting.pop().unwrap().0.0;
+                debug_assert!(
+                    cert.certificate
+                        .transaction_data()
+                        .kind()
+                        .is_accumulator_settle_tx(),
+                    "only settlement transactions may bypass causal admission"
+                );
+                return (cert, None);
+            }
```
mystenmark: settlement transactions should always bypass any causal restrictions (they don't have causal indices assigned so cannot respect them in any case).

however they should not violate concurrency constraints, if possible. enforcing concurrency does however require that the barrier is ordered after the settlements

crates/sui-benchmark/tests/simtest.rs
```diff
@@ -2175,13 +2178,15 @@ mod test {
...
+        // Congestion-induced cancellations vary seed-to-seed. The band was ~98-121
+        // over a 200-seed sweep with fixed execution concurrency; with the execution
+        // driver's concurrency limit randomized in test configurations (1..=8),
+        // low-concurrency draws reduce cluster throughput and with it the
+        // cancellation count (observed low 40s). A `> 20` bar still asserts that
+        // shared-object congestion control actually kicked in, with margin below
+        // the observed floor.
         assert!(
-            metrics_sum.cancellation_count > 50,
+            metrics_sum.cancellation_count > 20,
```
mystenmark: instead of the limit changes in this test, lets only have the concurrency randomization in the crashes_and_delays test, which doesn't do as many assertions on tx counts. then we can revert these changes

crates/sui-core/src/execution_driver.rs
```diff
@@ -1,71 +1,174 @@
...
-            _ = &mut rx_execution_shutdown => {
-                info!("Shutdown signal received. Exiting executor ...");
-                return;
+                }
             }
         };
+        let certificate = pending_cert.certificate;
+        let execution_env = pending_cert.execution_env;
+        let txn_ready_time = pending_cert.stats.ready_time.unwrap();
+        let _executing_guard = pending_cert.executing_guard;
```
lxfind, line 171: Should this remain in the task for proper accounting, which is still used by load control?

crates/sui-core/src/execution_scheduler/execution_scheduler_impl.rs
```diff
@@ -499,6 +514,12 @@ impl ExecutionScheduler {
         certs: Vec<(Schedulable, ExecutionEnv)>,
         epoch_store: &Arc<AuthorityPerEpochStore>,
     ) {
+        // Deduplicate and assign causal indices for the whole batch first, in enqueue
+        // order, for every unit - including keys whose transactions materialize only
+        // later. A key's transactions must run under the key's index: units enqueued
+        // after it may already be parked waiting for its outputs.
+        let certs = self.causal_admission.dedup_and_assign(certs);
```
lxfind: Could you reason about why this step belongs to the scheduler instead of consensus handling logic?

crates/sui-core/src/execution_scheduler/execution_scheduler_impl.rs
```diff
@@ -557,23 +578,63 @@ impl ExecutionScheduler {
...
+        // Resolve non-digest transaction keys. The consensus path enqueues keyed
+        // transactions (e.g. randomness updates) before their digest is known, and the
+        // usual resolver (RandomnessRoundReceiver) is best-effort per node - the
+        // signature may never arrive. The keyed copy owns its version group's causal
+        // index and dedup drops later enqueues of the group as duplicates, so a
+        // digest-carrying enqueue must resolve the key or the transaction is lost.
+        for (cert, _) in &certs {
+            let key = cert.key();
+            if !matches!(key, TransactionKey::Digest(_))
+                && epoch_store.insert_tx_key(key, *cert.digest()).is_err()
```
lxfind, line 591: Is this duplicate from a few other callsites of insert_tx_key? I don't fully understand this part.

crates/sui-core/src/execution_scheduler/causal_order.rs
```diff
@@ -0,0 +1,366 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Causal-order bookkeeping for execution scheduling.
```
lxfind, line 4: Should this file be named causal_admission?

crates/sui-core/src/execution_scheduler/causal_order.rs
```diff
@@ -0,0 +1,366 @@
...
+/// comment and `execution_driver.rs`.
+pub struct CausalAdmission {
+    inner: Mutex<AdmissionInner>,
+    /// Wakes the driver loop when the watermark or in-flight count changes.
+    notify: Notify,
+    /// Max transactions admitted for execution concurrently via the capacity branch (K).
+    concurrency_limit: usize,
+}
+
+struct AdmissionInner {
```
lxfind, line 53: Do we need to clean up this during epoch boundary? Or assert any remaining state?

crates/sui-types/src/transaction.rs
```diff
@@ -1968,11 +1968,6 @@ impl TransactionKind {
         )
     }
 
-    pub fn mutates_implicitly_read_system_object(&self) -> bool {
```
mystenmark: why was this deleted?

crates/sui-core/src/execution_scheduler/execution_scheduler_impl.rs
mystenmark: Not a duplicate - the three callsites cover disjoint paths. `RandomnessRoundReceiver` resolves the key when the signature arrives, but that's best-effort per node (the signature may never arrive), and `commit_certificate` resolves it after execution, which is too late for a placeholder that is still waiting. The case this handles is a node whose receiver never got the signature and whose only executable copy of the randomness update arrives via checkpoint. With enqueue dedup, that checkpoint copy is dropped as a duplicate of the keyed placeholder (which owns the version group's causal index), so unless the key is resolved here the placeholder never wakes and the transaction never executes. The seed search found exactly this deadlock.

That said, you were right that it didn't belong here: the checkpoint executor is the only source that enqueues digest-carrying copies of keyed transactions, and the certified checkpoint is what makes the key-to-digest mapping authoritative. Moved it there in 5e8280bcec, right before the unexecuted transactions are enqueued. `enqueue_transactions` now documents key resolution as a caller precondition and checks it in debug builds, since a violation is silent otherwise.

crates/sui-core/src/checkpoints/checkpoint_executor/mod.rs
```diff
@@ -984,6 +984,22 @@ impl CheckpointExecutor {
             ),
         );
 
+        // Resolve non-digest transaction keys (e.g. randomness updates). Consensus
+        // enqueues these as keyed placeholders before their digest is known, and the
+        // usual resolver (RandomnessRoundReceiver) is best-effort per node - the
+        // signature may never arrive. The placeholder owns its version group's causal
+        // index, so this digest-carrying enqueue must resolve the key or the
+        // transaction is never executed. The certified checkpoint makes the mapping
+        // authoritative.
```
mystenmark: this comment is confusing. simply explain that when the randomness txn arrives via checkpoint (rather than by constructing the txn locally) nothing else would resolve the key.

PR #27843, opened 2026-08-28
[move][move-decompiler] emit mut annotations, recover match guards, fix by-ref unpack rendering

cgswords:
## Description

Fixed 3 things:
1. We now derive when a `let` binding needs `mut` and add it
2. We now recover match guards in many (but not all) cases from the RHS if-else, which means a new match arm representation and processing arm guards.
3. Fixed up some by-ref variant unpack syntax

Smaller cleanup includes some work around annotation printing.

## Test plan

Plenty of new tests to exercise this behavior, specifically around match, plus some changes to existing test suite.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


external-crates/move/crates/move-decompiler/src/refinement/recover_match_guards.rs
```diff
@@ -0,0 +1,256 @@
...
+            }
+    }
+}
+
+// -------------------------------------------------------------------------------------------------
+// Recognition
+
+/// Recognize a guard-shaped arm body (see the module doc); `None` declines the arm.
+fn recognize(arm: &MatchArm) -> Option<Guarded<'_>> {
+    if arm.guard.is_some() || arm.fields.is_empty() {
```
tzakian, line 103: Fine for a limitation, but just to make sure -- 

This won't recover a match like
```
match (x) {
  _ if (cond1) => { ... },
  _ if (cond2) => { ... },
  _ if (cond3) => { ... },
}
```

And or non-binding structural matches coupled with guards right?

external-crates/move/crates/move-decompiler/src/needs_mut.rs
```diff
@@ -0,0 +1,732 @@
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Per-binder `mut` inference for the pretty printer. Bytecode carries no `mut`, so it is
```
tzakian, line 4: should we (do we?) do this for pattern variables today? If not possibly a good thing to add in the future and/or add tests for.

external-crates/move/crates/move-decompiler/tests/move/enums/guarded.snap
```diff
@@ -0,0 +1,88 @@
...
+            return l1
+        }
+    }
+
+    public fun tiered () {
+        let l6 = l0;
+        return         match(&l6: Order) {
+            Order::Market => {
+0u64            },
+            Order::Limit { qty: reg_6, price: reg_7 } if (        *reg_6 > 100u64) => {
```
tzakian, line 73: This formatting is a bit..rough.

Any way to fix it a smidge? Or "run an autoformatter on it"?

tzakian (approved):
Couple questions but otherwise looks reasonable to me 👍🏻

external-crates/move/crates/move-decompiler/src/refinement/recover_match_guards.rs
cgswords: It sure won't. Sadly.

external-crates/move/crates/move-decompiler/src/needs_mut.rs
cgswords: Match-arm binders go through `arm_needs_mut`; added a test case.

external-crates/move/crates/move-decompiler/tests/move/enums/guarded.snap
cgswords: Fixed!

PR #27854, opened 2026-08-31
[execution] Separate invariant check inputs from execution state

lxfind:
## Description

`TemporaryStore` mixed immutable check inputs, invariant execution state, and execution outcomes with different lifecycles, making reset behavior easy to get wrong.

- Move invariant-only execution state out of `TemporaryStore` and into `InvariantChecker`.
- Keep immutable post-execution check inputs solely owned by `TemporaryStore`; invariant checks read them from the store. Initialize normal-transaction inputs in the store constructor instead of through a separate setter.
- Add a dedicated genesis state update constructor so genesis callers do not supply unused input objects or transaction-only check inputs.
- Fix `into_bump_only` to preserve dynamically loaded object, package, and per-epoch metadata needed for replay while discarding execution outcomes and resetting invariant execution state.

## Test plan

Covered by existing tests in `sui-adapter-latest`.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -44,6 +45,46 @@ use sui_types::{SUI_SYSTEM_STATE_OBJECT_ID, TypeTag, is_system_package};
...
+    fn new(
+        transaction: Option<(&TransactionKind, &GasData, SuiAddress)>,
+        enable_gasless: bool,
+    ) -> Self {
+        let Some((transaction_kind, gas_data, transaction_signer)) = transaction else {
+            return Self {
+                is_genesis: true,
+                ..Default::default()
+            };
+        };
```
lxfind: **`None` transaction parameter silently means "genesis" and disables conservation checks** -- Confidence: 100/100 | Importance: 50/100

(found by: Claude, Codex)
- **File**: `sui-execution/latest/sui-adapter/src/temporary_store.rs` lines 64-73, 140-151
- **Problem**: `TemporaryStore::new` uses `None` for the `(&TransactionKind, &GasData, SuiAddress)` tuple to mean "this is genesis"; `PostExecutionCheckInputs::new(None)` consequently sets `is_genesis: true`, clears reservations, and omits declared-package checks. A privileged invariant-check exemption is encoded as missing data rather than an explicit semantic state, with nothing at the call site or in the API enforcing the contract.
- **Impact**: The only current `None` caller is `execute_genesis_state_update` (which never runs conservation checks anyway), so present behavior is correct. But any future caller constructing a store with `transaction: None` for a non-genesis purpose (a test harness, a new system-tx path) would silently disable SUI-conservation, balance-accumulator, and package-correspondence invariant checks — the exact defense-in-depth this PR is organizing.
- **Suggestion**: Make the contract explicit — replace the `Option` with an enum such as `PostExecutionCheckContext::{Transaction { ... }, GenesisStateUpdate}`, provide a separate genesis constructor, or at minimum document on `TemporaryStore::new` that `None` is genesis-only and disables conservation checks.

---

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -612,13 +650,12 @@ impl<'backing> TemporaryStore<'backing> {
...
+            loaded_per_epoch_config_objects,
+            post_execution_check_inputs,
             execution_results: ExecutionResultsV2::default(),
-            loaded_runtime_objects: BTreeMap::new(),
-            wrapped_object_containers: BTreeMap::new(),
-            runtime_packages_loaded_from_db: RwLock::new(BTreeMap::new()),
-            generated_runtime_ids: BTreeSet::new(),
-            loaded_per_epoch_config_objects: RwLock::new(BTreeSet::new()),
-            invariants: InvariantChecker::new(),
+            invariants: InvariantChecker::default(),
```
lxfind, line 689: **No targeted regression tests for the lifecycle boundaries this PR defines** -- Confidence: 100/100 | Importance: 55/100

(found by: Claude, Codex)
- **File**: `Entire PR` (transformations at `sui-execution/latest/sui-adapter/src/temporary_store.rs` lines 611-658)
- **Problem**: The central lifecycle changes have no direct tests. Nothing asserts that `drop_writes` retains immutable inputs while clearing all attempt-local invariant state, or that `into_bump_only` preserves loaded runtime objects, runtime-loaded packages, and per-epoch config reads while discarding outcomes — and the latter is the actual bug being fixed, not just a refactor.
- **Impact**: A future field addition or destructuring change could silently reintroduce stale ownership-authorization state or drop replay inputs again; ordinary successful-execution tests are unlikely to catch either. The v15 BumpOnly path in particular is unreleased and thus not exercised by production traffic.
- **Suggestion**: Add focused unit tests that seed every lifecycle category and assert field-by-field behavior across both transformations. For `into_bump_only`, drive a v15 BumpOnly exit (e.g. IFFW short-circuit after a receive/dynamic-field load) and assert the resulting `InnerTemporaryStore` retains replay metadata while effects contain only mutable-input version bumps.

---

lxfind (commented):
## Automated Review: 2 Actionable Finding(s)

2 issue(s) posted as inline comments on the relevant code.

---

sui-execution/latest/sui-adapter/src/temporary_store/invariants.rs
```diff
@@ -27,53 +25,31 @@ use mysten_common::debug_fatal;
...
 /// object ownership) as its API.
 ///
-/// It owns the transaction-derived inputs the checks need ([`InvariantInputs`], set once up
-/// front) and the PTB-emitted accumulator-event ranges accumulated during execution; the rest of
-/// the data the checks need (modified objects, accumulator events, settlement SUI, gas summary,
-/// ...) is read from the owning [`TemporaryStore`] passed to each check method.
+/// Immutable check inputs and execution results are read from the owning [`TemporaryStore`]
+/// passed to each check method.
+#[derive(Default)]
 pub(crate) struct InvariantChecker {
```
tzakian, line 48: Would a better name for this be `PostExecutionDynamicState` (or similar)?

I guess what I'm wondering, is would it be better to make a full split of there being two types:
* "static" state derived solely from transaction input(s);
* "dynamic" state derived from results of execution

Then have the invariant checker take both of these states, and the invariant checker would then operate over both of these?

E.g., the `InvariantChecker` type would be something like

```rust
struct InvariantChecker {
   post_execution_static_state: PostExecutionStaticState, // or PostExecutionInputState
   post_execution_dynamic_state: PostExecutionDynamicState, // wrapped objs, generated runtime ids etc.
}
```

or you could phrase it in terms of `PostExecutionPreState`, and `PostExecutionPostState` or the like as well possibly?

sui-execution/latest/sui-adapter/src/execution_engine.rs
```diff
@@ -241,6 +241,16 @@ pub(crate) mod checked {
...
+        // Apply the legacy gas-payment recovery before constructing the store so the gas charger
+        // and transaction-derived reservation inputs see the same final payment list.
+        if !bump_only_enabled(protocol_config.gas_model_version()) {
+            legacy::iffw_filter_address_balance_gas_payments(
+                &mut gas_data,
+                &execution_params,
+                protocol_config,
+            );
+        }
+
```
dariorussi, line 253: is it correct that this will disappear on execution version cut?
in other words this is legacy only right?

sui-execution/latest/sui-adapter/src/temporary_store.rs
```diff
@@ -92,24 +129,26 @@ pub struct TemporaryStore<'backing> {
...
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn new(
         store: &'backing dyn BackingStore,
         input_objects: InputObjects,
         receiving_objects: Vec<ObjectRef>,
         tx_digest: TransactionDigest,
         protocol_config: &'backing ProtocolConfig,
         cur_epoch: EpochId,
         _system_object_versions: BTreeMap<ObjectID, SequenceNumber>,
+        transaction: Option<(&TransactionKind, &GasData, SuiAddress)>,
```
dariorussi: when I was looking at some clean up, I felt those in TemporaryStore may be "all you need" so to speak. Just a comment

sui-execution/latest/sui-adapter/src/temporary_store.rs
dariorussi, line 689: thanks

dariorussi (approved):
I think this is another step forward, thanks!

sui-execution/latest/sui-adapter/src/execution_engine.rs
lxfind: I believe so

PR #27866, opened 2026-09-01
[sui][sui-executor] Improve execution timing behavior for generated commands

cgswords:
## Description

Replaced the execution timing structure with a more-robust one which handles out-of-bounds commands and command synthesis for timing.

## Test plan

Added a few new tests, and the others all remain green.

---

## Release notes

- [ ] Protocol:
- [ ] gRPC:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [ ] CLI:
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-types/src/unit_tests/execution_timings_tests.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Mysten Labs, Inc.
```
cgswords: I could take or leave these tests...

tnowacki (changes requested):
I think we can be a bit more surgical here, and I also think there is no real reason to move from `Vec<ExecutionTiming>` (or the slice) to a struct. Which would mean just changing `IndexedExecutionTimings` in latest. 

Which would mean the following approach I think:
1. protocol gate it (though we can check with @aschran)
2. just return from `into_coalesced` if the `original_command_len` is `0`
3. Debug assert that all indexes are `< original_command_len` (this  is just a more granular version of the existing assert)

We will also then need some e2e tests of some sort

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/interpreter.rs
```diff
@@ -553,3 +562,85 @@ impl IndexedExecutionTimings {
         coalesced
     }
 }
+
+#[cfg(test)]
+mod indexed_execution_timings_tests {
+    use super::IndexedExecutionTimings;
```
tnowacki: I don't really love these sort of tests floating around. If you think they are helpful. I would say we should at least break the `IndexedExecutionTimings` into its own file/module to keep things a bit cleaner.

tnowacki: I'm also fine not having these.

tnowacki (approved):

sui-execution/latest/sui-adapter/src/static_programmable_transactions/execution/interpreter.rs
cgswords: I agree, they were added during development but now we have stronger tests in the form of the transactional ones. Dropoed.

PR #27875, opened 2026-09-02
[Allowed Proposers] Stagger submission when proposers are undefined

akichidis:
## Description 

Introduces the `staggered_submission` component. As its name pretty much describes, it will help us stagger the transaction submission to consensus for transactions that have their `allowed proposers` field empty/undefined. The staggering will not be enabled by default, but it will get activated based on some signal trigger. This will allow us to move with more confidence while we rollout the allowed proposers feature knowing that clients still using the endpoint without defining any proposers will not abuse and amplify their transactions.

For the staggering we do perform a deterministic ordering of the committee members based on a seed that is derived using the transaction's gas objects. We could alternatively use the transaction's id (which is used for the gasless ones), but I did consider it more robust against double spend amplification attempts. 

The delay step `350ms` is chosen based on the current p50 commit latency, but we can tune it less/more aggressively

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

## Reviewer note: staggered submissions have their own, smaller permit pool

Non-staggered submissions keep the pre-existing structure exactly: the submit
permit is acquired from `submit_semaphore` before the race against the
processed waiter, and their behavior is unchanged.

Staggered submissions defer permit acquisition into the raced future, after
their hold — a held transaction must not occupy a scarce permit while sleeping
for up to `max_delay` — and draw from a dedicated pool sized
`max_pending_local_submissions / 4`. The stagger quota admits far more held
transactions (`max_pending / 4`, ~5000) than there are submit permits (~400),
so a wave of elapsed holds sharing the main FIFO semaphore could queue ahead of
restricted traffic; the dedicated pool makes the wave queue behind itself,
capping the staggered class at a quarter of submission concurrency. A staggered
transaction that commits elsewhere while sleeping or queued for a permit is
cancelled without ever being submitted (`Semaphore::acquire` is cancel-safe;
the wait gauge decrements on cancellation). `check_limits()` overload
signaling, FIFO ordering, and permit lifetime through the submit/status loop
are unchanged for existing traffic.

### DIFF
--- a/crates/sui-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/sui-core/src/authority/authority_per_epoch_store.rs
@@ -124,6 +124,7 @@ use crate::execution_cache::cache_types::CacheResult;
 use crate::fallback_fetch::do_fallback_lookup;
 use crate::module_cache_metrics::ResolverMetrics;
 use crate::signature_verifier::*;
+use crate::staggered_submission::StaggeredSubmission;
 use crate::stake_aggregator::{GenericMultiStakeAggregator, StakeAggregator};
 use sui_types::execution::ExecutionTimeObservationChunkKey;
 
@@ -424,6 +425,15 @@ pub struct AuthorityPerEpochStore {
 
     /// A cache which tracks recently finalized transactions.
     pub(crate) finalized_transactions_cache: FinalizedTransactionsCache,
+
+    /// Delays consensus submission of transactions without allowed proposers when
+    /// duplication is detected on the network. Shared by both submission paths (the
+    /// consensus adapter and the consensus transaction pool); per-epoch so the mode
+    /// resets across epoch boundaries along with the committee and ranks. Constructed
+    /// inactive: whatever drives activation (the commit-derived duplication signal,
+    /// once wired) must re-arm it after each reconfiguration — activation does not
+    /// survive an epoch change.
+    staggered_submission: StaggeredSubmission,
 
     /// The node's role for this epoch, derived from committee membership and
     /// the configured sync mode. Computed once at construction.
@@ -1061,6 +1071,7 @@ impl AuthorityPerEpochStore {
             tx_reject_reason_cache,
             submitted_transaction_cache,
             finalized_transactions_cache,
+            staggered_submission: StaggeredSubmission::new(),
             node_role: NodeRole::from_committee(&committee, &name, fullnode_sync_mode),
         });
 
@@ -1324,6 +1335,10 @@ impl AuthorityPerEpochStore {
     /// This validator's index in the current epoch's committee, if it is a member.
     pub fn own_committee_index(&self) -> Option<u32> {
         self.own_committee_index
+    }
+
+    pub fn staggered_submission(&self) -> &StaggeredSubmission {
+        &self.staggered_submission
     }
 
     /// Checks that the validator at committee index `proposer` is allowed to propose `tx` in
--- a/crates/sui-core/src/consensus_adapter.rs
+++ b/crates/sui-core/src/consensus_adapter.rs
@@ -49,6 +49,7 @@ use crate::authority::consensus_tx_status_cache::{
 use crate::checkpoints::CheckpointStore;
 use crate::consensus_handler::{SequencedConsensusTransactionKey, classify, tx_type_label};
 use crate::epoch::reconfiguration::{ReconfigState, ReconfigurationInitiator};
+use crate::staggered_submission::StaggeredSubmission;
 
 #[cfg(test)]
 #[path = "unit_tests/consensus_tests.rs"]
@@ -69,6 +70,7 @@ pub struct ConsensusAdapterMetrics {
     pub sequencing_in_flight_semaphore_wait: IntGauge,
     pub sequencing_in_flight_submissions: IntGauge,
     pub sequencing_best_effort_timeout: IntCounterVec,
+    pub sequencing_staggered_delay: Histogram,
     pub consensus_latency: Histogram,
     pub num_rejected_cert_in_epoch_boundary: IntCounter,
 }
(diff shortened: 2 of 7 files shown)


crates/sui-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -425,6 +426,15 @@ pub struct AuthorityPerEpochStore {
...
     pub(crate) finalized_transactions_cache: FinalizedTransactionsCache,
 
+    /// Delays consensus submission of transactions without allowed proposers when
+    /// duplication is detected on the network. Shared by both submission paths (the
+    /// consensus adapter and the consensus transaction pool); per-epoch so the mode
+    /// resets across epoch boundaries along with the committee and ranks. Constructed
+    /// inactive: whatever drives activation (the commit-derived duplication signal,
+    /// once wired) must re-arm it after each reconfiguration — activation does not
+    /// survive an epoch change.
+    staggered_submission: StaggeredSubmission,
```
akichidis, line 436: For now chose to keep it in the epoch store as its been shared both by the consensus adapter and the transaction pool . We can consider a better location (probably direct injection) once we deprecate the consensus adapter path (at least for the user transactions).

crates/sui-core/src/staggered_submission.rs
```diff
@@ -0,0 +1,1021 @@
...
+    hasher.update(epoch.to_le_bytes());
+    if gas_payment.is_empty() {
+        let mut digests: Vec<&TransactionDigest> = tx_digests.to_vec();
+        digests.sort();
+        for digest in digests {
+            hasher.update(digest.inner());
+        }
+    } else {
+        let mut ids: Vec<ObjectID> = gas_payment.to_vec();
+        ids.sort();
```
alex-mysten, line 357: Can this be gamed if I submit + amplify many nearly identical transactions that share a gas coin alongside a different dust coin per tx? They would each have different hashes and each would land on a different validator, but only one would succeed

crates/sui-core/src/staggered_submission.rs
```diff
@@ -0,0 +1,1021 @@
...
+        Self {
+            step: DEFAULT_STAGGER_STEP,
+            max_delay: DEFAULT_STAGGER_MAX_DELAY,
+            free_slots: MAX_UNPAID_ALLOWED_PROPOSERS,
+        }
+    }
+}
+
+/// Decides whether this validator should delay submitting a given user transaction to
+/// consensus, and by how much. Inactive by default;
```
alex-mysten, line 107: Do we expect this to turn off when we do not detect amplification?

crates/sui-core/src/staggered_submission.rs
```diff
@@ -0,0 +1,1021 @@
...
+        let Some(delay) = epoch_store
+            .staggered_submission()
+            .submission_delay(txs, epoch_store)
+        else {
+            return Ok(None);
+        };
+        if self
+            .held
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |held| {
+                (held < self.quota).then_some(held + 1)
```
alex-mysten, line 239: Could double-spend spam saturate this?

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -561,12 +634,44 @@ impl ConsensusAdapter {
...
+                }
+
+                let _permit: Option<SemaphorePermit> = match pre_acquired_permit {
+                    Some(permit) => Some(permit),
+                    None if is_system_message => None,
+                    // Staggered: the hold is over, take a permit from the class's own
+                    // smaller pool, so a wave of elapsed holds queues behind itself
+                    // instead of ahead of restricted traffic.
+                    None => Some(
+                        self.staggered_submit_semaphore
```
alex-mysten, line 663: A lot of touch points with staggered submission. I think we should either tie this to the pool only, or push it down to the queue, if that's feasible

crates/sui-core/src/consensus_adapter.rs
```diff
@@ -260,6 +300,10 @@ impl ConsensusAdapter {
             num_inflight_transactions,
             metrics,
             submit_semaphore: Arc::new(Semaphore::new(max_pending_local_submissions)),
+            staggered_submit_semaphore: Arc::new(Semaphore::new(
+                (max_pending_local_submissions / STAGGERED_SUBMIT_PERMITS_DIVISOR).max(1),
```
alex-mysten, line 304: What's the justification on this sizing? Feels small

crates/sui-core/src/consensus_adapter.rs
akichidis: Eventually consensus_adapter will become obsolete once we switch over to the transaction pool - at least on the user path - and this will be removed.  I would suggest to leave as is for now and not spend extra effort on it.

crates/sui-core/src/consensus_adapter.rs
akichidis: Good point - I've basically picked a number based on the assumption that the majority of the transactions will be with proposers defined. However, I think this number should be adjusted after we have more insight on from the mainnet metrics once the code has been fully deployed on our FNs .

crates/sui-core/src/staggered_submission.rs
akichidis: Yes , that's the mechanism. It is part of the https://github.com/MystenLabs/sui/pull/27882

crates/sui-core/src/staggered_submission.rs
akichidis: Yes, but pretty much in the way that double-spend could saturate today our existing consensus-adapter quotas. Keep in mind that double-spend should be non-incentivised by the downstream deferral mechanics in consensus_handler

PR #27882, opened 2026-09-02
[Allowed Proposers] enable staggered submission based on tx duplicates

akichidis:
## Description 

Enables the transaction staggered submission for the the transactions that have the allowed proposers undefined based on signal derived from commit output. The staggered submission component is now keeping a window of observations on transaction duplicates excess and total unique transactions. It calculates the percentage of tx duplication for the given window , and if above a threshold it activates the transaction staggering. It gets deactivated when duplication falls under a lower defined threshold.

## Test plan 

CI/PT

---

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:


crates/sui-protocol-config/src/lib.rs
```diff
@@ -1185,6 +1185,14 @@ struct FeatureFlags {
     #[serde(skip_serializing_if = "is_false")]
     allowed_proposers: bool,
 
+    // If true, validators derive the staggered-submission activation signal from commit
+    // output: unpaid duplication of transactions without allowed proposers arms staggered
+    // consensus submission in lockstep across honest validators. Requires
+    // `allowed_proposers` (accessed through the hand-written getter that asserts it).
+    #[serde(skip_serializing_if = "is_false")]
+    #[skip_protocol_config_accessor]
+    staggered_submission_signal: bool,
```
mystenmark, line 1194: this should all be done in NodeConfig since it isn't a protocol feature

akichidis: I've introduced a separate node config to act as a kill switch against the protocol config (I would suggest keeping the protocol config - it's mostly for sync rollout of the feature) .

PR #27884, opened 2026-09-02
docs: Rework accessing-data guides for readers with an existing project

jessiemongeon1:
## Description

`custom-indexer/build` and `grpc/using-grpc` both started from an empty directory: `cargo new`, `createdb`, `diesel setup`, `npm init`. 

`build` keeps the Sui-specific work and points readers who are genuinely starting cold at `examples/rust/basic-sui-indexer`, which is already the source of every code sample on the page. `using-grpc` is otherwise reference material, so only its client section changed.

## Release notes

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/docs/content/develop/accessing-data/custom-indexer/build.mdx
+++ b/docs/content/develop/accessing-data/custom-indexer/build.mdx
@@ -1,19 +1,19 @@
 ---
-title: Integrate a Custom Indexer
+title: Build a Custom Indexer
 description: >-
-  Add a custom indexer to an existing Rust service using the
-  `sui-indexer-alt-framework` module. Implement the `Processor` and `Handler`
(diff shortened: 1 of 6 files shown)


jessiemongeon1:
## Style Guide Audit

All 4 file(s) pass the style guide audit.

nickvikeras (approved):
I'm not really sure this change is necessary or helpful for anyone.  Even if someone is an engineer at a tradfi institution, they are probably building a new indexer service that would integrate into their larger architecture somehow.is

jessiemongeon1:
@nickvikeras Ok, will keep the build page as-is and add a separate integrate page so both workflows are supported

PR #27889, opened 2026-09-03
sui-benchmark: add large-transaction workload for block-size benchmarking

alex-mysten:
## Description

Adds a stress workload that submits transactions of a configurable size, so consensus throughput can be measured as a function of block size (`--large-transaction`, `--large-transaction-size-bytes`).

## Test plan

`test_simulated_load_large_transaction` runs the workload in isolation on an in-process cluster and asserts it generates load.

## Release notes

Check each box that your changes affect. If none apply, ignore this section. For each box you select, include information after the relevant checkbox that describes the impact of your changes that a user might notice and any actions they must take to implement updates.

- [ ] Protocol: 
- [ ] gRPC:
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] Indexing Framework:

### DIFF
--- a/crates/sui-benchmark/src/options.rs
+++ b/crates/sui-benchmark/src/options.rs
@@ -171,6 +171,13 @@ pub enum RunSpec {
         // relative weight of adversarial transactions in the benchmark workload
         #[clap(long, num_args(1..), value_delimiter = ',', default_values_t = [0])]
         adversarial: Vec<u32>,
+        // relative weight of large transactions (payload-only transactions of a configurable
+        // size, for block-size benchmarking)
+        #[clap(long, num_args(1..), value_delimiter = ',', default_values_t = [0])]
+        large_transaction: Vec<u32>,
+        // payload bytes attached to each large transaction (capped by max_tx_size).
+        #[clap(long, num_args(1..), value_delimiter = ',', default_values_t = [100_000])]
+        large_transaction_size_bytes: Vec<u64>,
         // relative weight of shared deletion transactions in the benchmark workload
         #[clap(long, num_args(1..), value_delimiter = ',', default_values_t = [0])]
         shared_deletion: Vec<u32>,
--- a/crates/sui-benchmark/src/workloads/workload_configuration.rs
+++ b/crates/sui-benchmark/src/workloads/workload_configuration.rs
@@ -25,6 +25,7 @@ use tracing::info;
 use super::adversarial::{AdversarialPayloadCfg, AdversarialWorkloadBuilder};
 use super::composite::{CompositeWorkloadBuilder, CompositeWorkloadConfig};
 use super::expected_failure::{ExpectedFailurePayloadCfg, ExpectedFailureWorkloadBuilder};
+use super::large_transaction::LargeTransactionWorkloadBuilder;
 use super::randomized_transaction::RandomizedTransactionWorkloadBuilder;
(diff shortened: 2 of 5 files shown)


crates/sui-benchmark/src/workloads/workload_configuration.rs
```diff
@@ -164,6 +170,7 @@ impl WorkloadConfiguration {
                             conflicting_transfer: conflicting_transfer[i],
                             composite: composite[i],
                         },
+                        large_transaction_size: large_transaction_size[i],
```
lxfind: **Multi-group runs panic when the new per-group flags are omitted** -- Confidence: 100/100 | Importance: 72/100

(found by: Codex)
- **File**: `crates/sui-benchmark/src/workloads/workload_configuration.rs` lines 164-173; `crates/sui-benchmark/src/options.rs` lines 176-180
- **Problem**: `--large-transaction` and `--large-transaction-size` are declared as vectors whose defaults hold a single element, yet `configure` indexes both vectors once per benchmark group. With `--num-of-benchmark-groups` greater than one and the new flags omitted, the second group indexes past the end of the default vector and panics, even though the large-transaction weight for every group would be zero.
- **Impact**: Multi-group benchmark command lines that were valid before this PR now abort with an index-out-of-bounds panic unless the operator adds both new flags with one value per group. The new workload is not even enabled in those runs. The existing per-group vector flags appear to be consumed the same way, so a general fix would normalize all of them, but the new flags are what turn a previously working command into a panic.
- **Suggestion**: Distinguish omitted values from explicitly supplied ones. When omitted, replicate the default for every group; when supplied, validate the vector length against the group count and return a clear error instead of indexing blindly. Add a two-group CLI regression test that omits both new flags.

---

lxfind (commented):

---

crates/sui-benchmark/src/options.rs
```diff
@@ -171,6 +171,13 @@ pub enum RunSpec {
         // relative weight of adversarial transactions in the benchmark workload
         #[clap(long, num_args(1..), value_delimiter = ',', default_values_t = [0])]
         adversarial: Vec<u32>,
+        // relative weight of large transactions (payload-only transactions of a configurable
+        // size, for block-size benchmarking)
+        #[clap(long, num_args(1..), value_delimiter = ',', default_values_t = [0])]
+        large_transaction: Vec<u32>,
+        // payload bytes attached to each large transaction (capped by max_tx_size).
+        #[clap(long, num_args(1..), value_delimiter = ',', default_values_t = [100_000])]
+        large_transaction_size: Vec<u64>,
```
akichidis: I would suggest something like `larget_transaction_size_bytes` so it becomes clear to the user the unit.

akichidis (approved):

crates/sui-benchmark/src/workloads/workload_configuration.rs
alex-mysten: True, supplying a new flag would break callers that specify multiple benchmark groups. However, in sui and sui-operations we do not have any multi group callers, and this change follows convention. I'm introducing a stress pattern soon that will use multiple bench groups, will address this in a followup.

crates/sui-benchmark/src/options.rs
alex-mysten: Done! Thanks

PR #27892, opened 2026-09-03
[graphql] Pool gRPC connections to the ledger service

tpham-mysten:
## Description

A ledger reader opened a single gRPC connection to the ledger service (sui-kv-rpc). Backfill runs many concurrent list calls, and one HTTP/2 connection's flow-control window and single driver task cap that throughput well below the KV-RPC pod's capacity. In-cluster benchmarks show pooling a few connections per replica reaches the per-pod ceiling.

This extracts a shared `ClientPool` that round-robins calls across N independent connections, uses it in both the streaming (`AlphaLedgerGrpcReader`) and point-get (`LedgerGrpcReader`) readers, and exposes the size as `--ledger-grpc-pool-size` (default 4).

## Test plan

New unit tests cover the pool's round-robin selection, wraparound, shared cursor across clones, and size clamping. Existing reader behavior is unchanged.

## Release notes

- [x] Indexing Framework: adds a `--ledger-grpc-pool-size` flag (default 4) controlling how many gRPC connections each ledger reader opens to the ledger service.

### DIFF
--- a/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
+++ b/crates/sui-indexer-alt-reader/src/alpha_ledger_grpc_reader.rs
@@ -10,19 +10,18 @@ use bytes::Bytes;
 use futures::Stream;
 use futures::StreamExt;
 use prometheus::Registry;
-use sui_rpc::Client;
 use sui_rpc::proto::sui::rpc::v2 as proto;
 use sui_rpc::proto::sui::rpc::v2::ExecutedTransaction;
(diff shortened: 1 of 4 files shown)


amnn (commented):
Does connection pooling belong in `sui_rpc`, or even the Rust SDK?

tpham-mysten:
Thanks @amnn for the catch! Yes I think adding the support to sui_rpc::Client directly is better. I have closed this PR and sent https://github.com/MystenLabs/sui-rust-sdk/pull/311

